PR #33, opened 2022-12-07
Add python tests with coverage workflow

lewijacn:
Signed-off-by: Tanner Lewis <lewijacn@amazon.com>

### Description
This change is to execute our python unit tests on opening of a PR and updates to a PR. It will also include code coverage to be added as a comment to the PR

### Testing
Verified workflow is successful on personal fork
* https://github.com/lewijacn/opensearch-migrations/actions/runs/3641127326/jobs/6146687419

### Items of Note
Two things I have noticed which are not blockers but may warrant minor additions in future
* Code coverage should be uploaded to the issue without token on codecov/codecov-action@v3 for public repos but this was not the case for my personal fork. This does not fail the Upload Coverage Report step as can be seen, but if this occurs for our migrations repo as well we may need a codecov repo upload token to work properly. I have seen other repos in OS which have included this and some that have not.
* Although using the same trigger `on: [push, pull_request]` seen in other OS projects, my personal fork would run this workflow twice whenever an open PR received an additional commit (once for both triggers). If this is the case, I have a fix in mind to limit this but other repos do not appear to have additional logic for this.

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -1,0 +1,31 @@
+name: python-tests
+
+on: [push, pull_request]
+
+jobs:
+  test-linux:
+    strategy:
+      matrix:
+        python-version: ['3.10']
+        os: [ubuntu-latest]
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        working-directory: ./upgrades
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install Dependencies for Framework and Test Coverage
+        run: |
+          python -m pip install -r requirements.txt coverage pytest-cov
+      - name: Run Tests with Coverage
+        run: |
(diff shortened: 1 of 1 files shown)


lewijacn:
> When I read the actually workflow results I saw this error in the coverage upload:
> 
> ```
> [2022-12-07T17:06:32.992Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Could not find a repository, try using repo upload token', code='not_found')}
> ```
> 
> I'm pretty sure that's the issue you were talking about in the "notes" section of your PR description. So it sounds like we don't actually know if this works. Am I correct in guessing the plan is to try pushing the change and just circling back if it doesn't?

Right, I'm not for sure whether coverage will get uploaded to PR as we expect. Luckily it will not fail the step and we may need to circle back with a small fix

chelma:
> Right, I'm not for sure whether coverage will get uploaded to PR as we expect. Luckily it will not fail the step and we may need to circle back with a small fix

Hmm, OK - works for me.  Will approve.

chelma (approved):
Approving PR, with the understanding we might need to circle back and fix the coverage upload.

okhasawn (approved):

PR #68, opened 2023-01-06
Add sample bwc expectations

lewijacn:
Signed-off-by: Tanner Lewis <lewijacn@amazon.com>

### Description
Add sample BWC expectations to build up knowledge base

### Issues Resolved
None

### Testing
N/A

### Check List

### DIFF
--- a/knowledge_base/consistent-index-settings.json
+++ b/knowledge_base/consistent-index-settings.json
@@ -1,0 +1,4 @@
+{
+  "id": "consistent-index-settings",
+  "description": "An index should have the same core index settings and associated data values before and after a migration or upgrade."
+}
--- a/knowledge_base/index-shrink-after-upgrade-consistent-documents.json
+++ b/knowledge_base/index-shrink-after-upgrade-consistent-documents.json
@@ -1,0 +1,4 @@
+{
+  "id": "index-shrink-after-upgrade-consistent-documents",
+  "description": "An index should have the same documents when its number of shards are reduced after the index is migrated or upgraded"
+}
--- a/knowledge_base/index-shrink-before-upgrade-consistent-documents.json
+++ b/knowledge_base/index-shrink-before-upgrade-consistent-documents.json
@@ -1,0 +1,4 @@
+{
+  "id": "index-shrink-before-upgrade-consistent-documents",
+  "description": "An index should have the same documents when its number of shards are reduced before the index is migrated or upgraded"
+}
--- a/knowledge_base/new-index-replica-before-upgrade-consistent-documents.json
+++ b/knowledge_base/new-index-replica-before-upgrade-consistent-documents.json
@@ -1,0 +1,4 @@
+{
+  "id": "new-index-replica-before-upgrade-consistent-documents",
+  "description": "An index and its replica should contain the same documents when the replica is created before a migration or upgrade"
+}
--- a/knowledge_base/consistent-document-count.json
+++ b/knowledge_base/consistent-document-count.json
@@ -1,4 +1,4 @@
 {
     "id": "consistent-document-count",
-    "description": "There should be the same number of documents in the each index before and after a migration or upgrade."
+    "description": "There should be the same number of documents in each index before and after a migration or upgrade."
 }
--- a/knowledge_base/new-index-replica-after-upgrade-consistent-documents.json
+++ b/knowledge_base/new-index-replica-after-upgrade-consistent-documents.json
@@ -1,0 +1,4 @@
+{
+  "id": "new-index-replica-after-upgrade-consistent-documents",
+  "description": "An index and its replica should contain the same documents when the replica is created after a migration or upgrade"
+}


knowledge_base/consistent-index-settings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "id": "consistent-index-settings",
+  "description": "An index should have the same core index settings and associated data values before and after a migration or upgrade."
```
chelma, line 3: Is this actually true, or are we just asserting it as true in a "come at me, bro" sense and seeing if it sticks?  Happy either way, and it will be fascinating to see how it plays out.

The concern comes in what, specifically, we're checking for as I can see these settings changing (either the options available or how they're stored) as versions change.  This is probably a case of "unknown unknowns" since we haven't delved into how different versions express this, so maybe it does make sense not to tie this expectation to specific versions and see how it plays out...

Curious on your thoughts @gregschohn

knowledge_base/index-shrink-before-upgrade-consistent-documents.json
```diff
@@ -0,0 +1,4 @@
+{
+  "id": "index-shrink-before-upgrade-consistent-documents",
```
chelma, line 2: Following @mikaylathompson's convention w/ the DOY bug expectation, maybe this should live in the same JSON file as `index-shrink-after-upgrade-consistent-documents`?  Are they two sides of the same coin?  Alternatively, they may actually be different different coins, and their flipside is that for certain versions it's NOT true...

I'm still trying to figure out what the right structure here is, but it feels like we do need the ability to express different scenarios for a given general topic.  In this case, have `knowledge_base/index_shrink_consistent_documents.json` contain both the "happy-path" and "sad-path" expectations for `index-shrink-after-upgrade-consistent-documents` and `index-shrink-before-upgrade-consistent-documents`.

Also curious about your thoughts on this one @gregschohn

knowledge_base/new-index-replica-before-upgrade-consistent-documents.json
```diff
@@ -0,0 +1,4 @@
+{
+  "id": "new-index-replica-before-upgrade-consistent-documents",
```
chelma, line 2: Same comment - should this be in the same file as its partner?  Probably...

chelma (commented):
Good finds; these expectations raised some questions for us to discuss/explore.

knowledge_base/consistent-index-settings.json
lewijacn: This is definitely a leading expectation as I am not aware of how this is handled in all versions. This is specifically targeted at index settings which I have at least some confidence on a couple values here as they are included in the bwc tests which makes no distinction really between versions. My plan was identifying this expectation as the core list of expectations that remains consistent and as we find specific settings that change or get added we could create specific expectations for that setting if that makes sense. But this is definitely a good one to think about as we will have similar issues

knowledge_base/index-shrink-before-upgrade-consistent-documents.json
lewijacn: They are essentially the same test, with the shrink either happening before the upgrade or after followed by our assertions.

I think I like grouping these together, my only drawback is like you mentioned if we find there is a break for one of both of these expectations we would need to add that here as well but I think that is reasonable. 

There is another side of this that may be `index-expand` which would be increasing the shards which I currently think makes sense to have a separate grouping for.

gregschohn (commented):
As stated in the inline comments, we need to start to define how we're constructing ids.  Since the implementation is separate, the id itself needs to very clearly identify how we're establishing context and what validations we're running.  That could be in the README of the knowledge_base.

What does "same documents" mean?  What if documents were malformed (as stated above)?  Presumably, those won't be returned.

This starts to hint at where a compositional aspect of expectations can give us a lot more leverage.  If this expectation had a contextual qualifier - for any set of documents well-formed for version X that have been added exactly once w/out updates, we expect this...  That would allow us to test the expectation on any dataset that provided such a dependency.  That doesn't sound like much, but that normalization could make the system much more expressive and find unexpected issues.

knowledge_base/consistent-index-settings.json
gregschohn: I've got some thoughts below - but I think that we should add some informative indicators to show the confidence in our assertions - from - maybe/hopeful thinking/I'd like to we've tested this elsewhere a million times.  

Another axis for informative purposes might be sensitivity.  That would probably be baked into the description.  Those additional axes/fields can grow later & we don't need to discuss that here.

knowledge_base/index-shrink-before-upgrade-consistent-documents.json
gregschohn: Additional thoughts on this are below.  Quick answer - this problem has the potential to explode, especially as we pick up more nuance and stricter interpretations.  I think that decomposing our expectations into reusable building blocks would help.

I like the spirit of grouping that @lewijacn brings up, but doubt that expectations will be able to be grouped over just a single axis.  This is a good spot where tags could come in.  On the other hand, the building blocks from above might also be able to serve as a grouping tag (documents with geospatial boundaries; clusters that were in the process of being upgraded in place when a [new|old] node failed over]).

I would say that my answer to @chelma's question of structure would be to distill expectations to have only a single responsibility.  At that point, you would at least have a major axis that could serve as a primary axis for grouping, even if it had mixin/dependency behaviors from other building blocks.  

Does that make any sense?

knowledge_base/consistent-index-settings.json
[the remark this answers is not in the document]
lewijacn: I mentioned slightly in Chris's comment above, but my original intent was for this to be fields in the settings portion of an index (excluding mappings, aliases, etc.) that were consistent across all versions. Initially my intention was to put all index settings here with a low level of confidence, and then as we actually ran this across versions, we would pull out settings that change/removed and give them their own expectation to clearly be able to list out how a particular setting might change

knowledge_base/new-index-replica-before-upgrade-consistent-documents.json
```diff
@@ -0,0 +1,4 @@
...
+  "description": "An index and its replica should contain the same documents when the replica is created before a migration or upgrade"
```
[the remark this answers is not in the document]
lewijacn: I would probably agree with this. As I was writing these expectations it was often the thought that I could easily change a small piece of a given expectation and probably have another valuable expectation to add as well so the level of detail can be desirable. 

I think the tricky part of this is how much detail is appropriate for the expectation model, as I accommodate more and more details the sentence structure begins to be not as pleasant and having a standard DSL starts to look a lot like the robot tests themselves and/or creates a little larger gap for new committers to add to.

I'd very interested to see an example of your thoughts @gregschohn for an informal DSL which sounds like it could potentially be a nice middle ground? I had played with doing a sample model that I for whatever reason gravitated toward SQL-like but didn't feel much better than the robot tests themselves to understand
```
GIVEN
CLUSTER source_cluster
CLUSTER target_cluster
INDEX test-index ON source_cluster 
	WITH DATA well-formed random number of documents n 
	WITH SETTING zero replicas

WHEN
test-index CHANGES SETTING one replicas
source_cluster UPGRADES|MIGRATES to target_cluster

THEN
Each DOCUMENT IN INDEX test-index ON target_cluster ON node with primary MATCHES
Each DOCUMENT IN INDEX test-index ON target_cluster ON node with replica
```

knowledge_base/index-shrink-after-upgrade-consistent-documents.json
```diff
@@ -0,0 +1,4 @@
+{
+  "id": "index-shrink-after-upgrade-consistent-documents",
```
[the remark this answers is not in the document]
lewijacn: This is a good point on the document comparison and likely why I see several cases of the BWC simply checking that the document count is the same without actually checking the content of the documents. Maybe down the road but it seems like we could potentially have a library for document comparison that accounts for known changes and could overlook specific differences that other expectations will call out

PR #69, opened 2023-01-12
Refactoring robot tests to start using shared REST client + Removing dependency on OpenSearch Python client

okhasawn:
### Description
-Remove references to opensearch python client and dependency
-Setup keywork mapping for unclear functions
-Add more functions to be available for use by the robot framework

* Category (Enhancement, New feature, Bug fix, Test fix, **Refactoring**, Maintenance, Documentation)

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
+++ b/cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
@@ -82,6 +82,7 @@ class RESTClientDefault(RESTClientBase):
         params = {"pretty": "true"}
 
         return ops.perform_post(rest_path=rest_path, params=params)
+
     def create_snapshot(self, port: int, repo: str, snapshot_id: str) -> ops.RESTResponse:
         """
         Create a snapshot of the cluster into a repo by synchronously blocking on completion
--- a/cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
+++ b/cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
@@ -2,8 +2,6 @@ import json
 
 from cluster_migration_core.clients.rest_client_base import RESTClientBase
 import cluster_migration_core.clients.rest_ops as ops
-#import requests
-#import RestResponse
 
 
 class RESTClientDefault(RESTClientBase):
@@ -48,7 +46,7 @@ class RESTClientDefault(RESTClientBase):
 
     def create_an_index(self, port: int, index: str) -> ops.RESTResponse:
         """
-        Post a single document to an index
+        Creates an index
         """
         rest_path = ops.RESTPath(port=port, suffix=f"{index}")
         params = {"pretty": "true"}
--- a/cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
+++ b/cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
@@ -2,6 +2,8 @@ import json
 
 from cluster_migration_core.clients.rest_client_base import RESTClientBase
 import cluster_migration_core.clients.rest_ops as ops
+#import requests
+#import RestResponse
 
 
 class RESTClientDefault(RESTClientBase):
@@ -41,7 +43,17 @@ class RESTClientDefault(RESTClientBase):
         """
         rest_path = ops.RESTPath(port=port, suffix=f"{index}/_doc/{doc_id}")
         params = {"pretty": "true"}
+
         return ops.perform_get(rest_path=rest_path, params=params)
+
+    def create_an_index(self, port: int, index: str) -> ops.RESTResponse:
+        """
+        Post a single document to an index
+        """
+        rest_path = ops.RESTPath(port=port, suffix=f"{index}")
+        params = {"pretty": "true"}
+
+        return ops.perform_post(rest_path=rest_path, params=params)
 
     def post_doc_to_index(self, port: int, index: str, doc: dict) -> ops.RESTResponse:
         """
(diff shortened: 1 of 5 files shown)


okhasawn:
The test that is failing as part of the workflow is actually passing on a local run.

cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
```diff
@@ -53,6 +63,24 @@ def post_doc_to_index(self, port: int, index: str, doc: dict) -> ops.RESTRespons
 
         return ops.perform_post(rest_path=rest_path, data=json.dumps(doc), headers=headers, params=params)
 
+    def count_doc_in_index(self, port: int, index: str) -> ops.RESTResponse:
```
chelma, line 66: Very minor nitpick: mind changing this to `count_docs_in_index()` (e.g. pluralizing doc)?

upgrades/upgrade_testing_framework/robot_lib/OpenSearchRESTActions.py
```diff
@@ -1,34 +1,44 @@
...
-from opensearchpy import OpenSearch
+import cluster_migration_core.core.versions_engine as ev
+import cluster_migration_core.clients as clients
 
 DEFAULT_TEMP_STORAGE = "/tmp/utf/robot_data.json"
 
 
-# Change this library to use REST requests instead of Python client
-class OpenSearchClientLibrary(object):
+class CouldNotRetreiveAttributeFromRESTResponse(Exception):
```
chelma, line 10: Typo: `Retreive`

upgrades/upgrade_testing_framework/robot_lib/OpenSearchRESTActions.py
```diff
@@ -1,34 +1,44 @@
...
 
     def count_documents_in_index(self, index_name: str) -> int:
-        return self._os_client.count(index=index_name).get('count')
+        response = self._rest_client.count_doc_in_index(port=self._port, index=index_name).response_json
+        count_key = "count"
+        try:
+            count = response[count_key]
+            return count
+        except CouldNotRetreiveAttributeFromRESTResponse(count_key, response):
+            print('Please enter a valid attribute from RESTResponse')
```
chelma, line 38: Hmmm, not sure I understand your intent here.  Can you clarify?  I actually didn't know you even could supply an instantiated object to an `except` clause like this.  Also, curious why we're using a print statement instead of Python's logging facilities.

upgrades/upgrade_testing_framework/robot_lib/OpenSearchRESTActions.py
```diff
@@ -1,34 +1,44 @@
...
+        super().__init__(f"Could not find key '{key}' in the RESTResponse json: {response}")
+
+
+class OpenSearchRESTActions(object):
 
     def __init__(self, host="localhost", port=9200, temp_storage_location=DEFAULT_TEMP_STORAGE):
-        os_client = OpenSearch(hosts=[{'host': host, 'port': port}])
-        self._os_client = os_client
+        #This is temporary, this will be provided by the executor
+        engine_version = ev.EngineVersion(ev.ENGINE_OPENSEARCH, 1, 3, 6)
```
chelma, line 19: Hmm, is there a reason we didn't pipe this through in this PR?  We're using the same set of actions against both ES 7.10.2 and OS 1.3.6 Clusters in our current workflow, right?  This feels a bit too hacky to me.

lewijacn: @chelma I didn't want to bog Omar down with routing the EngineVersion from the steps to the actual actions here and have a PR to raise which will remove this.

@okhasawn I think I'd be a little more comfortable with removing the engine version lines here and just using  `self._rest_client = RESTClientDefault()` until my change goes in and reworks this

cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
```diff
@@ -41,8 +41,18 @@ def get_doc_by_id(self, port: int, index: str, doc_id: str) -> ops.RESTResponse:
...
         rest_path = ops.RESTPath(port=port, suffix=f"{index}/_doc/{doc_id}")
         params = {"pretty": "true"}
+
         return ops.perform_get(rest_path=rest_path, params=params)
 
+    def create_an_index(self, port: int, index: str) -> ops.RESTResponse:
+        """
+        Creates an index
+        """
+        rest_path = ops.RESTPath(port=port, suffix=f"{index}")
```
lewijacn, line 51: As a tip I don't think you necessarily need the f string `f"{index}"` here as index is already a string and is not joined with anything, `index` by itself should work as well

upgrades/upgrade_testing_framework/robot_lib/OpenSearchRESTActions.py
okhasawn: Ah, thanks for pointing this out, I had written the whole try catch block in a different way which ended up being reverted when I was undoing some of the code I wrote while testing a few other things.
But the idea is to throw an exception when `count_key != "count"` and provide the value of count_key and the response gotten from the REST client.
As for the print statement, it's also part of the things that were undone, I do realize though that it's considered bad practice and it's better to use python logging at least for categorizing the messages, So thanks for pointing that out as well :)

lewijacn (commented):
Left some minor comments, looks like its on the right path 👍

cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
okhasawn: Will do!

cluster_migration_core/cluster_migration_core/clients/rest_client_default.py
okhasawn: Ah I see, thanks! I'll address this.

lewijacn (approved):

chelma (approved):
LGTM!

PR #101, opened 2023-02-09
Logs for requests/responses are valid json objects

okhasawn:
Signed-off-by: Omar Khasawneh <okhasawn@amazon.com>

### Description
This change will cause the log file entries to be valid json objects for each of the logs and responses.
Logs include: Requests (timestamp, uri, method, body, h̶e̶a̶d̶e̶r̶s̶) and Responses (response time, body, h̶e̶a̶d̶e̶r̶s̶, status code)

Example log entry:

`
 { "request": { "timestamp":1676307415, "uri":"\/noldor\/_doc\/1", "method":"PUT", "body":"{\"name\": \"Finwe\"}" }, "response": {"response_time_ms":556, "body":"{\"_index\":\"noldor\",\"_type\":\"_doc\",\"_id\":\"1\",\"_version\":1,\"result\":\"created\",\"_shards\":{\"total\":2,\"successful\":1,\"failed\":0},\"_seq_no\":0,\"_primary_term\":1}", "status_code": 201 } }
`

The entries will still be valid json objects when non-english characters are part of a request/response, and will show as unicode characters.

An issue with this log-format (that can't be avoided per my research about it), is that the body of a request/response will still be included within it's respective headers section, which is why this task was created [Migrations-992](https://opensearch.atlassian.net/browse/MIGRATIONS-992)

It's important to capture all known and unknown headers, and the only way to do that is with the variable `%hr`, which will have the body at index 0 due to this line `http-request capture req.body id 0`, which is how we can capture the body.

The headers variables (`%hr %hs`) will contain certain characters that break the ability to read the data, so it's important to be able to json-escape the headers, haven't found a way to do that yet with HAProxy, so reverting to not logging the headers for now and created a task ([Migrations-993](https://opensearch.atlassian.net/browse/MIGRATIONS-993)) to investigate that. 

### Issues Resolved
Partially https://opensearch.atlassian.net/browse/MIGRATIONS-953

### Check List


cluster_traffic_capture/cluster_traffic_capture/gen_haproxy_cfg.py
```diff
@@ -71,7 +71,9 @@ def _get_fe_config_haproxy(haproxy_port: int, mirror: bool) -> str:
     declare capture request len 80000
     declare capture response len 80000
     http-request capture req.body id 0
```
mikaylathompson, line 73: Can you 
1/ add a JIRA issue about removing the body from the headers block,
2/ add a comment here explaining the issue with the headers in the body + linking to/mentioning that issue?

cluster_traffic_capture/cluster_traffic_capture/gen_haproxy_cfg.py
```diff
@@ -71,7 +71,9 @@ def _get_fe_config_haproxy(haproxy_port: int, mirror: bool) -> str:
...
-    log-format Request-URI:\\ %[capture.req.uri]\\nRequest-Method:\\ %[capture.req.method]\\nRequest-Body:\\ %[capture.req.hdr(0)]\\nResponse-Body:\\ %[capture.res.hdr(0)]
+    http-request capture req.hdrs len 512
```
mikaylathompson, line 74: Why 512 characters? Is the body included as part of this 512? if so, it seems very possible that the body will push out all of the actual headers

mikaylathompson:
Does your example above come from a real request? 

It doesn't have any headers except for the body, so it's tough to tell 1/ whether it's working as expected, and 2/ what the formatting of the headers looks like. Could you add more examples that show that?

cluster_traffic_capture/cluster_traffic_capture/gen_haproxy_cfg.py
```diff
@@ -71,7 +71,9 @@ def _get_fe_config_haproxy(haproxy_port: int, mirror: bool) -> str:
...
+    log-format '{{ "request": {{ "timestamp":%Ts, "uri":"%[capture.req.uri,json('utf8ps')]", "method":"%[capture.req.method,json('utf8ps')]", "body":"%[capture.req.hdr(0),json('utf8ps')]", "headers": "%hr" }}, "response":  {{"response_time_ms":%Tr, "headers":"%hs", "body":"%[capture.res.hdr(0),json('utf8ps')]", "status_code": %ST }} }}'
```
mikaylathompson, line 75: Is there a reason you're not json-escaping the headers as well? I'm not positive whether it's necessary or not (I think it depends on what kind of characters are legal to include in headers) but I wasn't sure if there was a specific reason you're not doing it

okhasawn:
> Does your example above come from a real request?
> 
> It doesn't have any headers except for the body, so it's tough to tell 1/ whether it's working as expected, and 2/ what the formatting of the headers looks like. Could you add more examples that show that?

My mistake on that part, that example is from before I added the second commit to this PR, so that's not considered a valid example, I'll update that, thanks for catching that!

cluster_traffic_capture/cluster_traffic_capture/gen_haproxy_cfg.py
okhasawn: I did have it 4096 at one point but didn't think that was necessary, now I think otherwise as I just ran a test and the body is in fact included in as part of the 512. Any suggested length?

mikaylathompson:
Can you also clarify this comment in the text of the PR:
>  It's important to capture all known and unknown headers, and the only way to do that is with the variable %hr and by HAProxy's default, it includes the body.

It's not by HAProxy's default, it's because of the `http-request capture req.body id 0` line that "captures" the body and places it at index 0 of the headers. The problem is that we haven't yet found another way to capture the body.

cluster_traffic_capture/cluster_traffic_capture/gen_haproxy_cfg.py
chelma: +1

cluster_traffic_capture/cluster_traffic_capture/gen_haproxy_cfg.py
chelma: I don't think we ever want to truncate here.  So, how about ~1MB (1048576 characters) for all the capture mechanisms.  We'll need to either find a better solution long term or scream this limit loud and clear to users.

chelma (changes requested):

okhasawn:
Due to not being able to json-escape the headers in the log-format, I'll be removing the headers entirely until I'm able to do so. The reason being is that the headers sometimes contain certain characters that break our ability to read the entire entry.

A task to investigate that for a solution that has been created: https://opensearch.atlassian.net/browse/MIGRATIONS-993

mikaylathompson (approved):

chelma (approved):
I think these changes will be overwritten by @mikaylathompson's upcoming PR

PR #104, opened 2023-02-13
Extend log line length limits

mikaylathompson:

### Description
We realized we were encountering an issue where log lines were being truncated. The default haproxy limit is 1024 characters [0], and once that is addressed, the default for syslog is a limit of 2048 characters [1].  Both of these can be increased to about 64K (64*1024) characters.

While we're very possibly going to hit further issues with that down the road, for now this limit is longer than any of the logs we're generating while running the Opensearch Benchmark workloads (max seems to be about 20k characters).

[0] https://cbonte.github.io/haproxy-dconv/1.7/configuration.html#3.1-log
[1] https://support.hpe.com/hpesc/public/docDisplay?docId=c03271127

I also threw in two python code linting fixes.

### Testing
Manually tested with my docker setup.

### Check List


chelma (approved):
LGTM, thanks!

okhasawn (approved):

PR #111, opened 2023-02-17
MIGRATIONS-963 Pipe AWS Creds through to container

chelma:
### Description
The HAProxy container needs AWS Credentials and other context in order to offload the logs.  When the containers are run in an AWS-context, these will be provided by Fargate/ECS/whatever natively.  For our demo, we need to pipe them in manually but preferably in a manner similar to how they will be available in an AWS-context.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-963

### Testing
Manually confirmed availability of ENV variables inside the demo's Primary HAProxy container.  

### Check List


gregschohn (approved):

cluster_traffic_capture/demo_haproxy.py
```diff
@@ -47,6 +47,38 @@ def main():
     if verbose:
         logging.root.setLevel(logging.DEBUG)
 
+    # =================================================================================================================
+    # Pull Necessary State
+    # =================================================================================================================
+    print("Pulling AWS credentials from ENV variables...")
```
mikaylathompson, line 53: not a blocker, but this would be a nice thing to put behind a flag. We'll likely still want to run this without providing aws credentials (and therefore not offloading to S3).

cluster_traffic_capture/demo_haproxy.py
```diff
@@ -159,6 +191,12 @@ def subshell_print(message: str):
         ports=[dfc.PortMapping(HAPROXY_INTERNAL_PORT, HAPROXY_PRIMARY_PORT)],
         volumes=[],
         ulimits=[Ulimit(name='memlock', soft=-1, hard=-1)],
+        env_kv={
+            "AWS_REGION": aws_region,
+            "AWS_ACCESS_KEY_ID": aws_access_key_id,
+            "AWS_SECRET_ACCESS_KEY": aws_secret_access_key,
+            "AWS_SESSION_TOKEN": aws_session_token
+        },
```
mikaylathompson, line 199: I'm not sure I have a great understanding of when you use the `env_kv` vs the passthrough -- why are this in the `env_kv` instead of passthrough?

mikaylathompson (approved):
Some questions, but not blocking.

cluster_traffic_capture/demo_haproxy.py
chelma: The current Docker SDK we're using doesn't invoke CLI commands via the shell and so it can't do passthrough unfortunately.  Therefore, we need to use the K/V approach.  I left the passthrough code I wrote in the PR because I think the path forward is to move away from the Python Docker SDK and just invoke the CLI manually using our own code.

cluster_traffic_capture/demo_haproxy.py
chelma: Trying not to invest too heavily in this script as a lot of it will be either deleted or re-written when we move away from HAProxy.  For the time being, you just need to have the ENV variables defined; they don't need to be valid. In that case, the container will spin up but CloudWatch Agent won't be able to upload.

chelma:
Merging, per @mikaylathompson

PR #110, opened 2023-02-15
Logging Plugin & Traffic Replayer POC work

gregschohn:
### Description
    POC code.... 
    1) TransportReplayer: Crude and inefficient replayer that can take logfiles generated by ElasticSearch/OpenSearch instances that were using the logging-transport-netty4 plugin.
    2) Wire logging for OpenSearch/ElasticSearch: A crude network plugin for OpenSearch that replaces and extends the transport-netty4 module.  This plugin uses log4j to emit channel status updates and the contents of the input/output that is going through the http channel.  To build, it can be dropped into the OpenSearch project or the ElasticSearch one (though namespaces will need to be updated) and built from there.
    3) Simple Dockerfile to setup an ES7.10 cluster with the logging netty plugin & it's requisite configuration changes.

### Issues Resolved
MIGRATIONS-971

### Testing
Unit tests for the traffic replayer are included.  None have been done for the plugin yet.

### Check List
- [ ] New functionality includes testing
  - [ ] All tests pass, including unit test, integration test and doctest
- [ sortof ] New functionality has been documented
- [ x ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #113, opened 2023-02-28
POC Logging Plugin & Replayer

mikaylathompson:

### Description
This PR is a follow up to #110. See that issue for more info on the plugin.
As for the replayer, it is very much a work-in-progress, but it successfully does the following:
- Read in logged request/response files from a directory
- Replay the requests packet-by-packet
- Log the results in the form of "triple" jsons -- {"request" , "primaryResponse", "shadowResponse"} -- for the traffic comparator, with latencies for both responses.

There are definitely a lot of bugs and TODOs -- we're hoping to surface a lot of those in this PR.

### Issues Resolved
[MIGRATIONS-970](https://opensearch.atlassian.net/browse/MIGRATIONS-970)

### Testing
A lot of manual testing, but it is woefully lacking on unit and integration tests.

### Check List


TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -78,26 +87,49 @@ public static void main(String[] args) throws IOException, InterruptedException
...
             AtomicBoolean waiter = new AtomicBoolean(true);
-            packetHandler.finalizeRequest(summary->{
+            var requestResponseTriple = new RequestResponseResponseJSONTriple(requestResponsePacketPair);
+            packetHandler.finalizeRequest(summary-> {
                 System.err.println("Summary(#"+(nReceived++)+")="+summary);
+                try {
+                    requestResponseTriple.addShadowResponse(summary.getReceiptTimeAndResponsePackets().map(entry -> entry.getValue()).collect(Collectors.toList()),
+                            summary.getResponseDuration());
+                } catch (IOException e) {
+                    e.printStackTrace();
```
gregschohn, line 128: You should add a very noisy log statement here too that also says something like - SWALLOWING EXCEPTION, XXXXXX, TODO, etc.  TODO in the code is helpful too.
That's basic programming tip... if I have a byte array, what could possibly go wrong?  Why would the block above ever NEED to throw an IOException.  If it's an internal error that should never happen, that should be handled further down the stack - this code shouldn't be concerned with it & likely doesn't have the means to deal with it, since any exceptions wouldn't be something that this code has any control over.

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -110,7 +142,7 @@ private void writeToSocketAndClose(RequestResponsePacketPair requestResponsePack
             }
         } catch (IOException e) {
             throw new RuntimeException(e);
-        } catch (IPacketToHttpHandler.InvalidHttpStateException e) {
+        } catch (org.opensearch.migrations.replay.IPacketToHttpHandler.InvalidHttpStateException e) {
```
gregschohn, line 145: import

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -193,6 +233,43 @@ public Stream<byte[]> getResponseDataStream() {
         }
     }
 
+    public static class RequestResponseResponseJSONTriple {
```
gregschohn, line 236: Move this to it's own class.  See the comments above about how this should be associated w/ the write callbacks, not the replayer.  That coupling makes the code less general and will make unit tests harder to author.

It's also worth noting that this is a stopgap solution that was put in place to expedite things for the traffic comparison engine.  Longer term, I would like to see the output of this to be as detailed as the data we have from the original dumps (notice that we could look to share the same netty WireHandler if we really wanted to).

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -193,6 +233,43 @@ public Stream<byte[]> getResponseDataStream() {
...
+        final JSONObject primaryResponseData;
+        JSONObject shadowResponseData;
+
+        private JSONObject jsonFromHttpData(List<byte[]> data) throws IOException {
+            ByteArrayOutputStream requestStream = new ByteArrayOutputStream();
+            for(var d: data) {
+                requestStream.write(d);
+            }
+            String[] splitMessage = requestStream.toString(Charset.defaultCharset()).split("\r\n\r\n", 2);
+            JSONObject message = HTTP.toJSONObject(splitMessage[0].replaceAll("\r\n", "\n"));
```
gregschohn, line 247: A comment or some self documenting code would be helpful.  Consider a couple final static values for 0 & 1 to be HEADERS and BODY respectively.

gregschohn, line 247: message should be called something else.  For the first 8 readings, I kept thinking that it was an HTTP message.

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -245,6 +323,44 @@ public void consumeLogLinesForFile(Path p, Consumer<ReplayEntry> replayEntryCons
...
+            JSONObject meta = new JSONObject();
+            meta.put("request", triple.requestData);
+            meta.put("primaryResponse", triple.primaryResponseData);
+            meta.put("shadowResponse", triple.shadowResponseData);
+
+            try {
+                bufferedOutputStream.write(meta.toString().getBytes(StandardCharsets.UTF_8));
+                bufferedOutputStream.write("\n".getBytes(StandardCharsets.UTF_8));
+            } catch (IOException e) {
+                e.printStackTrace();
```
gregschohn, line 354: Never do this, as per above - at least put lots of loud logging around it.  Eaten exceptions are a MAJOR cause of errors in java code.

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn, line 247: Why are you replacing \r\n with \n?  If this was a difference in behaviors between the two clusters, it would go unnoticed & could be rather fatal, especially if there's wonky code on the client side (presume that somebody will have a hand-built mobile client making a call).

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -245,6 +323,44 @@ public void consumeLogLinesForFile(Path p, Consumer<ReplayEntry> replayEntryCons
         }
     }
 
+    private static String formatPacketsAsBase64String(String prefix, ArrayList<byte[]> packets) {
```
gregschohn, line 326: Is this used?  I was wondering what the prefix was, but I don't see a reference to it in the PR & the code shows as new.

TrafficReplayer/src/org/opensearch/migrations/replay/NettyPacketToHttpSender.java
```diff
@@ -8,31 +8,26 @@
...
 
 import java.io.IOException;
-import java.io.ObjectInputStream;
-import java.net.Socket;
 import java.net.URI;
+import java.time.Instant;
 import java.util.function.Consumer;
 
-public class NettyPacketToHttpHandler implements IPacketToHttpHandler {
+public class NettyPacketToHttpSender implements org.opensearch.migrations.replay.IPacketToHttpHandler {
```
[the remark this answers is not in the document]
mikaylathompson: Not intentional. Fixing.

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -78,26 +87,49 @@ public static void main(String[] args) throws IOException, InterruptedException
...
     }
 
     private static String aggregateByteStreamToString(Stream<byte[]> bStream) {
         return new String(Bytes.concat(bStream.toArray(byte[][]::new)), Charset.defaultCharset());
     }
 
     static int nReceived = 0;
-    private void writeToSocketAndClose(RequestResponsePacketPair requestResponsePacketPair) {
+    private void writeToSocketAndClose(RequestResponsePacketPair requestResponsePacketPair,
+                                       Consumer<RequestResponseResponseJSONTriple> onResponseReceivedCallback) {
```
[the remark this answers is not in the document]
mikaylathompson: Okay, yep, that makes sense.

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
mikaylathompson: Nope, not used. Thanks for catching. It's new in the scope of this PR because it went with the original approach of outputting everything as base64 strings, but not used in the latest commits.

mikaylathompson:
updated code in response to comments here, but due to commit signing issues in #110 , it needed to be on a new branch. Opened here: https://github.com/opensearch-project/opensearch-migrations/pull/117

PR #116, opened 2023-03-04
Streaming traffic replay

gregschohn:
### Description
* Category - Enhancement
* Why these changes are required? - To facilitate E2E streaming.
* What is the old behavior before changes and new behavior after changes? - before, this worked on files until the end of the file was reached.  Now it can read from stdin until stdin is closed.  Directory support is now gone.  It's the caller's responsibility to convert many files into a single stream.

### Issues Resolved
MIGRATIONS-1020

### Testing
Updated the unit test.  Tested this manually w/ stdin and a file.  Directory support is now gone.

### Check List
- [ x ] New functionality includes testing
  - [ x ] All tests pass, including unit test, integration test and doctest
- [ x ] New functionality has been documented
- [ x ] Commits are signed per the DCO using --signoff

### DIFF
--- a/TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
+++ b/TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
@@ -17,6 +17,9 @@ import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.Comparator;
@@ -37,6 +40,7 @@ public class TrafficReplayer {
             Pattern.compile("^\\[(.*),(.[0-9]*)\\]\\[.*\\]\\[.*\\] \\[(.*)\\] \\[id: 0x([0-9a-f]*), .*\\] " +
                     "(ACTIVE|FLUSH|INACTIVE|READ COMPLETE|REGISTERED|UNREGISTERED|WRITABILITY|((READ|WRITE)( ([0-9]+):(.*))))");
     final static int TIMESTAMP_GROUP        = 1;
+    final static int TIMESTAMP_MS_GROUP     = 2;
     final static int CLUSTER_GROUP          = 3;
     final static int CHANNEL_ID_GROUP       = 4;
     final static int SIMPLE_OPERATION_GROUP = 5;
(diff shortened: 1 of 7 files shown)


TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -59,32 +49,51 @@ public TrafficReplayer(URI serverUri)
...
-        Path directoryPath;
-        try {
-            directoryPath = Paths.get(args[0]);
-        } catch (Exception e) {
-            exitWithUsageAndCode(3);
-            return;
-        }
+
+        CloseableStringStreamWrapper css = args.length > 1 ? getLogLineStreamFromFile(args[1]) :
+                getLogLineStreamFromInputStream(System.in);
```
mikaylathompson, line 92: I'm okay with this, but the convention that I'm used to is that `-`in the place of a file means to use stdin. That seems slightly more explicit and harder to mess up. Any reason you didn't go that route?

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -125,19 +134,9 @@ private void log(String s) {
...
-                        .sorted(Comparator.comparing(Path::toString));
-                sortedDirStream.forEach(filePath -> {
-                    try {
-                        consumeLogLinesForFile(filePath, replayEngine);
-                    } catch (IOException e) {
-                        throw new RuntimeException(e);
-                    }
-                });
-            }
+            logLinesStream.forEach(s->parseAndConsumeLine(replayEngine, s));
```
mikaylathompson, line 139: this is a nice simplification!

mikaylathompson (approved):
Looks good, with one question about the input format

TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: Other than gunzip, where have you seen that convention?  It looks like cat & grep will assume stdin if nothing is specified, BUT will also interpret '-' to be stdin.  It's a fine change to make, but I feel like the arg parsing in general will need to be eventually improved anyway here.

mikaylathompson: I think it's actually quite widespread.
https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html (second to last bullet point) is one source, "Many tools accept a bare hyphen, not associated with any option letter, as a pseudo-filename directing the application to read from standard input" [[source](http://www.catb.org/~esr/writings/taoup/html/ch10s05.html)] is a second, and [many non-authoritative stack overflow/github/google results](https://www.google.com/search?q=unix+convention+hyphen+as+stdin).

Totally fine with it being part of an argparsing improvement down the line though.

PR #119, opened 2023-03-17
Allow reading from stdin & writing to stdout.  Also pulled log4j2 and beust.jcommander

gregschohn:
I had to convert all existing System.err.printlns to appropriate log statements.  Same goes for removing rogue System.out lines since that may now be the "triples" output stream. To simplify handling the nature of the input stream, I've modeled input in the form of a java.util.stream.Stream (instead of an InputStream directly).  I've wrapped that Stream in a class that is Closeable so that callers can be responsible for closing it upon completion so that the underlying IOStream can be freed.

### Description
MIGRATIONS-1020 - allow this to read from stdin and write outputs to stdout

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1020

### Testing
Updated the unit test, but also tested manually by comparing that the output to stdout didn't contain spurious lines.  Line count matched when using a file & all lines began with '{"request'.

### Check List
  - [] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented (via command line args)

### DIFF
--- a/TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
+++ b/TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
@@ -1,7 +1,11 @@
 package org.opensearch.migrations.replay;
 
+import com.beust.jcommander.JCommander;
+import com.beust.jcommander.Parameter;
+import com.beust.jcommander.ParameterException;
 import com.google.common.io.CharSource;
 import com.google.common.primitives.Bytes;
+import lombok.extern.log4j.Log4j2;
 
 import java.io.BufferedOutputStream;
 import java.io.BufferedReader;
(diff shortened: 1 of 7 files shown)


TrafficReplayer/src/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -56,57 +60,106 @@ public TrafficReplayer(URI serverUri)
         packetHandlerFactory = new NettyPacketToHttpHandlerFactory(serverUri);
     }
 
-    private static void exitWithUsageAndCode(int statusCode) {
-        System.err.println("Usage: <log directory> <destination_url>");
-        System.exit(statusCode);
+
+    static class Parameters {
```
mikaylathompson, line 64: I do not understand the java syntax going on here at all, but this is definitely an improvement. Does having a "name" field mean that the name needs to be provided? (but the URI can be inferred as the first parameter?)

mikaylathompson (approved):

PR #134, opened 2023-04-01
Logging proxy

gregschohn:
### Description
This is a trivial proxy that uses netty to stand up an https or http endpoint to proxy traffic to an underlying host/port.  It ALSO includes support for reliably (I think) committing PUT/POST/DELETES to disk before proceeding with making the call to the downstream service.  This is done in a netty async friendly way and non0mutating calls are also logged but when the connection is closed.  

Logged data now stored in a protocol buffer format and there are some messages within it to assist with replay (specifically, "EndOfMessage", which includes some byte lengths for the first line and headers).

THERE ARE STILL A LOT OF THINGS LEFT TO DO HERE - but hopefully, this can set a path forward.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1038 and part of https://opensearch.atlassian.net/browse/MIGRATIONS-1039

### Testing
Only 2 and one (an E2E proxy one) is flaky.

### Check List


TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/InMemoryConnectionCaptureFactory.java
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.WeakHashMap;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.BiFunction;
+
+public class InMemoryConnectionCaptureFactory implements IConnectionCaptureFactory {
```
lewijacn, line 21: It may be helpful to call out this is only used for testing or specify if you are planning an additional purpose here

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ProxyChannelInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+                (httpRequest.method().equals(HttpMethod.POST) ||
+                        httpRequest.method().equals(HttpMethod.PUT) ||
+                        httpRequest.method().equals(HttpMethod.DELETE)));
+    }
+
+    @Override
+    protected void initChannel(SocketChannel ch) throws IOException {
+        var sslContext = sslEngineProvider != null ? sslEngineProvider.get() : null;
+        if (sslContext != null) {
+            ch.pipeline().addLast(new SslHandler(sslEngineProvider.get()));
```
lewijacn, line 46: Assuming the SslHandler decrypts the incoming data for the following handlers, do we do any re-encryption before we send to the source cluster?

lewijacn (approved):
I think there are a few things that could be done to improve the code quality here (which I believe we are planning to complete in the future) around readability, comments, error handling, unit tests that I didn't focus on in this PR given the malleable state we are currently in. From what I have understood conceptually things have made sense (minus the comments), but plan to advocate/make improvements as I gain a better understanding here

As a point of reference after adding a test dependency for protobuf (discussed in one of the comments) I was able to build these projects and have the tests pass.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/FrontsideHandler.java
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    @Override
+    public void channelActive(ChannelHandlerContext ctx) {
+        final Channel inboundChannel = ctx.channel();
+        // Start the connection attempt.
+        Bootstrap b = new Bootstrap();
+        b.group(inboundChannel.eventLoop())
+                .channel(ctx.channel().getClass())
+                .handler(new LoggingHandler(LogLevel.ERROR))
```
mikaylathompson, line 40: Netty novice question: this isn't a handler that's going to cause everything that comes through the channel to be logged, right?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/NettyScanningHttpProxy.java
```diff
@@ -0,0 +1,62 @@
...
+            throws InterruptedException {
+        InternalLoggerFactory.setDefaultFactory(JdkLoggerFactory.INSTANCE);
+        bossGroup = new NioEventLoopGroup(1);
+        workerGroup = new NioEventLoopGroup();
+        ServerBootstrap serverBootstrap = new ServerBootstrap();
+        try {
+            mainChannel = serverBootstrap.group(bossGroup, workerGroup)
+                    .channel(NioServerSocketChannel.class)
+                    //.handler(new LoggingHandler(LogLevel.INFO))
+                    //.childHandler(new HexDumpProxyInitializer(backsideHost, backsidePort))
```
mikaylathompson, line 41: nit: any reason to keep this commented out code?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ProxyChannelInitializer.java
```diff
@@ -0,0 +1,58 @@
...
+        }
+
+        var offloader = connectionCaptureFactory.createOffloader(ch.id().asShortText());
+        ch.pipeline().addLast(new LoggingHandler("PRE", LogLevel.WARN));
+        ch.pipeline().addLast(new LoggingHttpResponseHandler(offloader));
+        ch.pipeline().addLast(new ConditionallyReliableLoggingHttpRequestHandler(offloader,
+                this::shouldGuaranteeMessageOffloading));
+        //ch.pipeline().addLast(new ConditionallyReliableWireLoggingHandler(x->true, offloader));
+        //ch.pipeline().addLast(new LoggingHandler("POST", LogLevel.ERROR));
+        ch.pipeline().addLast(new FrontsideHandler(host, port));
```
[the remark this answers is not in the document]
mikaylathompson: You'll probably get a more full response from Greg, but this is my understanding from working on the replayer w/ netty:

>  It seems like they work together to relay messages to the underlying cluster

Yes, exactly.
The proxy has two functions (ignoring the whole kafka/logging thing, this is just a generic proxy): receiving requests and sending those same requests out to whoever's on the other end of the proxy (so Elasticsearch in this case). And then the same receiving/sending, but flipped around for the responses.

The frontside is the front from the perspective of someone outside of the cluster -- think front door. a request comes in and it hits the frontside. The backside is what's pushing out those requests to the Elasticsearch cluster (and then receiving the responses).

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * This Java source file was generated by the Gradle 'init' task.
+ */
```
mikaylathompson, line 3: Is this actually true? This is a generated file?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -0,0 +1,90 @@
...
+                System.err.println("Received shutdown signal.  Trying to shutdown cleanly");
+                proxy.stop();
+                System.err.println("Done stopping the proxy.");
+            } catch (InterruptedException e) {
+                System.err.println("Caught exception while shutting down: "+e);
+                throw new RuntimeException(e);
+            }
+        }));
+        while (true) {
+            Thread.sleep(60*1000);
```
mikaylathompson, line 86: What is this for? Why 1 minute?

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -0,0 +1,321 @@
...
+    @Override
+    public void addDeregisterEvent(Instant timestamp) throws IOException {
+
+    }
+
+    static abstract class BufRangeConsumer {
+        abstract void accept(byte[] buff, int offset, int len);
+    }
+//
+//    private void addByteBuf(ByteBuf buffer, BufRangeConsumer bufConsumer) {
```
mikaylathompson, line 184: Why is this commented out?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/HttpCaptureSerializerUtil.java
```diff
@@ -0,0 +1,52 @@
...
+    public static HttpProcessedState addHttpMessageIndicatorEvents(
+            IChannelConnectionCaptureSerializer trafficOffloader,
+            List<Object> parsedMsgs) throws IOException {
+        Instant timestamp = Instant.now();
+        for (var obj : parsedMsgs) {
+            if (obj instanceof LastHttpContent) {
+                trafficOffloader.commitEndOfHttpMessageIndicator(timestamp);
+                return HttpProcessedState.FULL_MESSAGE;
+            }
+            var decoderResultLoose = getDecoderResult(obj);
```
mikaylathompson, line 40: what does "loose" mean here?

mikaylathompson (approved):

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/FrontsideHandler.java
gregschohn: It DOES log everything on the channel to the log4j facilities - meaning with the current setup, stderr.  In a future commit, I've removed the handler - or commented it out entirely.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/NettyScanningHttpProxy.java
gregschohn: Second line should be deleted.  First line - judgment call.  When debugging netty, I find myself often having to add logging - so having the pattern "LoggingHandler" that I can search for & being able to uncomment it in about 2 seconds & making sure that it's in the right spot is very helpful.  It's easy to add LoggingHandlers at different places that aren't as helpful - this code would be better if it included a name for the LoggingHandler so that it was really clear.  First we should get comfortable keeping that commented code in place though.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
gregschohn: The statement is true, your comment is not ;).
It _WAS_ generated by gradle long ago, but has been changed, so the current contents have not been generated by Gradle.  I'll drop it in a future change.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
gregschohn: Here are comments that I'll be adding - let me know if it makes more sense now
```
        // This loop just gives the main() function something to do while the netty event loops
        // work in the background.
        while (true) {
            // TODO: The kill signal will cause the sleep to throw an InterruptedException,
            // which may not be what we want to do - it seems like returning an exit code of 0
            // might make more sense.  This is something to research - e.g. by seeing if there's
            // specific behavior that POSIX recommends/requires.
```

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/HttpCaptureSerializerUtil.java
gregschohn: Looser typing.

PR #139, opened 2023-04-15
Traffic replayer protobuf updates

gregschohn:
### Description
TrafficReplayer now consumes protobuf delimited streams of "TrafficStream" objects and sends those to the target cluster.  These changes also remove all of the logfile parsing/base64 code that had done this previously.
* New feature/Refactoring
* Performance & correctness - having a higher fidelity capture that is less computationally intensive to marshal will improve results and performance.
* We base64 encoded all contents, obfuscating them for the developer.  Now, w/ protobuf serialized objects, the payload and headers are passed untouched, meaning that they're visible w/out needing to pass them through additional parse and decoding stages.  Notice that there will be some binary values for protobuf metadata in between those strings, but there are also many off-the-shelf tools to deal with protocol buffer content.

Along the way, the protobuf definition files (.proto) have been moved into their own separate (java) library that the captureOffloader and trafficReplayer are each taking a dependency on. 

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1069

### Testing
Unit tests to verify that the aggregation of packets is working appropriately (TrafficReplayerTest.testReader())

### Check List


TrafficCapture/settings.gradle
```diff
@@ -8,7 +8,11 @@
  */
 
 rootProject.name = 'LoggingProxy'
-include('trafficCaptureProxyServer', 'captureOffloader', 'nettyWireLogging',
+include('captureOffloader',
+        'captureProtobufs',
+        'nettyWireLogging',
+        'trafficCaptureProxyServer',
+        'trafficReplayer',
```
mikaylathompson, line 15: Commented on this on your other PR, but I don't understand why `trafficReplayer` is being added to the `LoggingProxy` project.

mikaylathompson (approved):
The actual code on the replayer side dealing with the protobufs was quite easy to follow. LGTM

TrafficCapture/settings.gradle
gregschohn: For now, I'm combining them together for convenience.  As a developer, I really need to be able to build "world" with one command.  Putting all of the capture and replay logic into a single hierarchy makes that really simple.  It's a monorepo approach.  I'll (maybe begrudgingly) agree to move off of it when somebody has cycles to invest in another approach that affords me the same abilities.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -285,27 +245,41 @@ public ReplayEngine(Consumer<RequestResponsePacketPair> fullDataHandler) {
...
-                    accept(e);
-                    return;
+                runningList.addResponseData(timestamp, observation.getWrite().toByteArray());
+            } else if (observation.hasWriteSegment()) {
+                updateFirstResponseTimestamp = true;
+                throw new RuntimeException("Not implemented yet.");
+            } else if (observation.hasRequestReleasedDownstream()) {
+                updateFirstResponseTimestamp = true;
+            }
+            if (updateFirstResponseTimestamp && runningList != null) {
```
lewijacn, line 279: nit: We could combine this to a single if statement

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -285,27 +245,41 @@ public ReplayEngine(Consumer<RequestResponsePacketPair> fullDataHandler) {
...
-                    runningList = liveStreams.get(e.id);
-                } else if (runningList.lastOperation == Operation.Write) {
-                    publishAndClear(e.id);
-                    accept(e);
-                    return;
+                runningList.addResponseData(timestamp, observation.getWrite().toByteArray());
+            } else if (observation.hasWriteSegment()) {
+                updateFirstResponseTimestamp = true;
+                throw new RuntimeException("Not implemented yet.");
+            } else if (observation.hasRequestReleasedDownstream()) {
```
lewijacn, line 276: This is to signify that we have finished forwarding request to cluster, correct. Do we set this currently in the LoggingProxy?

lewijacn (approved):
Looks much cleaner 👍 . Left some comments for understanding

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: There are multiple branches that set updateFirstResponseTimestamp, hence the reason that it's in the "follow-up" area.  Please let me know if I'm missing something.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: Yes and no.  Here's the [commit ](https://github.com/opensearch-project/opensearch-migrations/commit/1c9e0e20c9dffd7dd440a80a96b6d8225dafb388) that created the new observation

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
lewijacn: This was to simply say that lines 279 and 280 could be combined as 280 is just a nested if statement

PR #144, opened 2023-04-21
[Index configuration tool] Unit tests and refactoring of main.py

kartg:
### Description

This PR houses two major changes:

1. `main.py` has been refactored into smaller modules - `search_endpoint.py` for request/response interactions with ES/OS endpoints, and a `utils.py` module for utility functions.
2. Unit tests have been added for all modules

A secondary change to note is that `index-configuration-tool` has been renamed to `index_configuration_tool` so it is a valid Python identifier, and `__init__.py` files have been added to create valid Python packages. Other minor changes:
* Tweaks to the Logstash parsing to better support multiple plugins per section, and fixes for number parsing.
* Changed a variable name in `main.py` for clarity to incorporate [this PR comment](https://github.com/opensearch-project/opensearch-migrations/pull/141#discussion_r1171772555).
* README file and a Dockerfile definition have been added

### Testing
Unit tests for all modules are now included, and passing. Current coverage:

```
$ python -m coverage report --omit="*/tests/*"

Name                      Stmts   Miss  Cover
---------------------------------------------
index_operations.py          29      2    93%
logstash_conf_parser.py      30      1    97%
main.py                      65      0   100%
utils.py                     12      0   100%
---------------------------------------------
TOTAL                       136      3    98%

```

### Check List


index_configuration_tool/main.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def validate_logstash_config(config: dict):
+    if "input" not in config or "output" not in config:
+        raise ValueError("Missing input or output data from Logstash config")
+    validate_plugin_config(config, "input")
+    validate_plugin_config(config, "output")
+
+
+def print_report(source: dict, target: dict) -> set:
```
mikaylathompson, line 62: Minor naming note: this does print a report, but the more important job seems to be determining which indices to create--true?
Maybe either break it into two functions with clear jobs or come up with a name that encompasses both jobs?

index_configuration_tool/main.py
```diff
@@ -0,0 +1,98 @@
...
+    print("Indices already created in target cluster (data will NOT be moved): " +
+          utils.string_from_set(identical_indices))
+    print("Indices to create: " + utils.string_from_set(indices_to_create))
+    return indices_to_create
+
+
+if __name__ == '__main__':
+    # Parse logstash config file
+    print("\n##### Starting index configuration tool... #####\n")
+    logstash_config = parser.parse(sys.argv[1])
```
mikaylathompson, line 85: Maybe a check for whether there is a `argv[1]` and a error/help message if not?

index_configuration_tool/search_endpoint.py
mikaylathompson, line 1: Can you explain the name of this file? It's effectively something like "index operations", right? I'm not sure how to understand search endpoint, so I'd recommend renaming, unless I'm just not understanding the connection.

index_configuration_tool/tests/test_logstash_parser.py
```diff
@@ -0,0 +1,76 @@
...
+
+    # Test input json should match loaded pickle data
+    def test_parser_happy_case(self):
+        actual = parse(LOGSTASH_TEST_INPUT_FILE)
+        test_diff = diff(self.test_data, actual)
+        # Validate that diff is empty
+        self.assertFalse(test_diff)
+
+    def test_bad_configs(self):
+        with self.assertRaises(lark.exceptions.UnexpectedToken):
```
mikaylathompson, line 33: If one of these raises an `UnexpectedToken` and the rest don't, this test would pass, right?

And currently, if the first raises an exception, do the rest actually execute? I would expect the code to stop running at that point.

Regardless, I think these should be separated to be in their own `with assertRaises` context, unless I'm misunderstanding something about how it works.

index_configuration_tool/tests/test_logstash_parser.py
```diff
@@ -0,0 +1,76 @@
...
+            logstash_parser.parse("")
+            # Sections should begin with type name
+            logstash_parser.parse("{}")
+            # Invalid type
+            logstash_parser.parse("bad {}")
+            # Valid type but without params
+            logstash_parser.parse("input")
+
+    def test_empty_config(self):
+        logstash_parser.parse("input {}")
```
mikaylathompson, line 44: To be clear, these are valid and should pass? Maybe worth a comment or naming the function as `test_empty_config_is_valid` or something -- they _look_ pretty invalid. 

Actually, these succeed in parsing, but would fail the `validate_logstash_config` call, right? So maybe `test_empty_config_can_be_parsed` or along those lines?

index_configuration_tool/tests/test_main.py
```diff
@@ -0,0 +1,172 @@
...
+
+class TestMain(unittest.TestCase):
+    # Run before each test
+    def setUp(self) -> None:
+        with open(LOGSTASH_EXPECTED_OUTPUT_FILE, "rb") as f:
+            self.loaded_logstash_config = pickle.load(f)
+
+    def test_get_auth(self):
+        # Empty input should produce empty auth tuple
+        self.assertFalse(main.get_auth(dict()))
```
mikaylathompson, line 53: oof, I don't love just testing whether the result is falsey for an empty tuple. Can you do `assert main.get_auth(dict()) == tuple()` to be a little more direct about what you're looking for?

But additionally, my understanding from the `get_auth` function is that it should _actually_ be returning None, not an empty tuple (which also evaluates to False). If that's the case, can you do `assert main.get_auth(dict()) is None` and update the comment?

index_configuration_tool/tests/test_main.py
```diff
@@ -0,0 +1,172 @@
...
+        # SSL enabled
+        test_config = create_plugin_config(["test"], True)
+        result = main.get_endpoint_info(test_config)
+        self.assertEqual("https://test/", result[0])
+        self.assertFalse(result[1])
+        # SSL disabled, invalid auth config
+        test_config = create_plugin_config(["test"], False, "test")
+        result = main.get_endpoint_info(test_config)
+        self.assertEqual("http://test/", result[0])
+        self.assertFalse(result[1])
```
mikaylathompson, line 84: what is the actual value here? is it None or False (or something else)?

index_configuration_tool/tests/test_main.py
```diff
@@ -0,0 +1,172 @@
...
+            "index1": {
+                "settings": {
+                    "bool_key": True,
+                    "num_key": 1
+                },
+                "mappings": {
+                    "str_key": "value"
+                }
+            },
+            "index2": {}
```
mikaylathompson, line 109: what does it mean for an index to not have any settings or mappings? Is this a possible state?

index_configuration_tool/tests/test_main.py
```diff
@@ -0,0 +1,172 @@
...
+        }
+        result = main.print_report(base_data, dict())
+        self.assertEqual(2, len(result))
+        self.assertTrue("index1" in result)
+        self.assertTrue("index2" in result)
```
mikaylathompson, line 114: more a clarification question: earlier you were using vanilla `assert`s -- is there any difference in the use cases or advantage to using `assertTrue`?

index_configuration_tool/tests/test_main.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_validate_plugin_config_unsupported_endpoints(self):
+        with self.assertRaises(ValueError):
+            # No supported endpoints
+            main.validate_plugin_config(BASE_DATA, TEST_KEY)
+
+    def test_validate_plugin_config_missing_host(self):
+        with self.assertRaises(ValueError):
+            # Missing hosts value
+            test_data = create_config_section({})
```
mikaylathompson, line 137: Minor thing for these next few tests, I would pull the `test_data = ...` line _out_ of the `with` block, so the `assertRaises` is very clear about what part you expect to raise errors.

index_configuration_tool/tests/test_main.py
```diff
@@ -0,0 +1,172 @@
...
+        main.validate_plugin_config(test_data, TEST_KEY)
+
+    def test_validate_logstash_config_missing_required_keys(self):
+        with self.assertRaises(ValueError):
+            # Empty input
+            main.validate_logstash_config(dict())
+            # input missing
+            main.validate_logstash_config({"output": ()})
+            # output missing
+            main.validate_logstash_config({"input": ()})
```
mikaylathompson, line 165: Same comment here as with the earlier `assertRaises` with multiple test cases -- these should be in separate `with assertRaises` blocks so that 1/ they're all run and 2/ we know if any of them fail.

index_configuration_tool/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class TestSearchEndpoint(unittest.TestCase):
+    def setUp(self) -> None:
+        self.test_data_without_internal = copy.deepcopy(TEST_DATA)
+        # Remove internal data
+        for i in self.test_data_without_internal.keys():
+            self.test_data_without_internal[i][SETTINGS_KEY].pop(INTERNAL_INDEX_KEY, None)
+
+    @responses.activate
```
mikaylathompson, line 44: This is cool! I haven't seen the `responses` library before, but it looks super helpful!

index_configuration_tool/tests/test_search_endpoint.py
```diff
@@ -0,0 +1,66 @@
...
+    def test_fetch_all_indices(self):
+        # Set up GET response
+        responses.get(TEST_ENDPOINT + "*", json=TEST_DATA)
+        # Now send request
+        index_data = search_endpoint.fetch_all_indices(TEST_ENDPOINT)
+        self.assertEqual(2, len(index_data.keys()))
+        # Test that internal data has been filtered
+        self.assertTrue(INTERNAL_INDEX_KEY in index_data[INDEX1_NAME][SETTINGS_KEY])
+        self.assertFalse(index_data[INDEX1_NAME][SETTINGS_KEY][INTERNAL_INDEX_KEY])
```
mikaylathompson, line 53: I'm not sure I understand this combo of test conditions:
we're testing that the key _is_ present, but the value is gone (or falsey --can we test explicitly for None if that's the expectation?) -- is that correct?  That doesn't match with my understanding of `pop`, it generally deletes the key as well as the value, so--if I'm reading this correctly--I'm surprised it passes.

index_configuration_tool/search_endpoint.py
```diff
@@ -0,0 +1,37 @@
...
+def fetch_all_indices(endpoint: str, optional_auth: Optional[tuple] = None) -> dict:
+    actual_endpoint = endpoint + __ALL_INDICES_ENDPOINT
+    resp = requests.get(actual_endpoint, auth=optional_auth)
+    # Remove internal settings
+    result = dict(resp.json())
+    for index in result:
+        for setting in __INTERNAL_SETTINGS_KEYS:
+            index_settings = result[index][SETTINGS_KEY]
+            if __INDEX_KEY in index_settings:
+                index_settings[__INDEX_KEY].pop(setting, None)
```
mikaylathompson, line 23: the second param of pop just sets a default return value for the function (so if `setting` wasn't in the dictionary, it would return `None` instead of the value of the `setting` key). In this case, you're not doing anythign with the returned value (you're just treating pop as delete), so you don't need a default value at all.

mikaylathompson (commented):
🎉 Yay tests! Improvements overall look good -- some python-y stuff. Can I put documentation on the wishlist for the next PR?

index_configuration_tool/tests/test_logstash_parser.py
kartg: Good point. I've changed this (and a similar test case in `test_main.py` to use a loop now.

index_configuration_tool/tests/test_logstash_parser.py
kartg: Changed to `test_empty_config_can_be_parsed` and added a comment

index_configuration_tool/tests/test_main.py
kartg: Fixed the assert to be strongly-typed, as suggested. I've also broken this out into separate test cases - one for the bad inputs (where we check for `is None`) and one for the valid input

index_configuration_tool/tests/test_main.py
kartg: The expected value here is `None`. Fixed to be strongly-typed, and added deeper checks for the with-auth cases

index_configuration_tool/tests/test_main.py
kartg: Yes, it's possible for an index to have no explicit mappings. Strictly speaking, the default minimum configuration (ignoring internal/transient entries) for an index is:
```
"mappings": {},
"settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1
}
```

However, I think these default configurations are irrelevant for the purposes of the tests in this file. I'm totally fine to use this configuration if you feel otherwise though.

index_configuration_tool/tests/test_main.py
kartg: I went down a rabbit hole on this, but here's the short version:

1. `self.assert*` is better than Python's `assert`
2. `self.assertEqual` is better than `self.assertTrue / self.assertFalse`

The reason for (1) is because the `self.assert*` methods return an output object that you can inspect for more granular assertions. Meanwhile, the `assert` statement would simply output an `AssertionError` with no other context.

The reason for (2) is because the failure message for `assertTrue / assertFalse` is simply `False is not True` which is not particularly useful, while `assertEqual` will actually print the mismatched values. See [this StackOverflow answer](https://stackoverflow.com/a/45947566) for more.

Keeping this in mind, i've updated the unit tests to use `assertEqual` wherever possible

index_configuration_tool/tests/test_search_endpoint.py
kartg: Fixed this test to be clearer. 

The purpose of these assertions is to verify that the `index` key is still present (since this contains other important settings such as `number_of_replicas`) but internal fields have been filtered out. Note that the [filtering logic](https://github.com/kartg/opensearch-migrations/blob/indexConfigurationTool/index_configuration_tool/search_endpoint.py#L23) doesn't `pop` the index key, but the individual internal settings within it.

Since the input data in this revision _only_ had internal keys, the `assertFalse` check was checking for an empty `dict`. I've added a non-internal key now to make this clearer and moved to a more strongly-typed assert.

index_configuration_tool/search_endpoint.py
kartg: Without the default value, the `dict` would throw a `KeyError` if the key is not found ([docs](https://python-reference.readthedocs.io/en/latest/docs/dict/pop.html_)) 😞

index_configuration_tool/main.py
kartg: 👍  Separated into `get_index_differences` and `print_report` now

index_configuration_tool/search_endpoint.py
kartg: Ah, I was attempting to use a generic name to show that this module could hit either ElasticSearch or OpenSearch. So I opted for "search", but I see how that can be confusing since the operations within are index-related rather than search-related. Renamed to `index_operations.py`

lewijacn (commented):
Looks good 👍 . Mainly nitpicks but would like to know about the comment on endpoint always having slash at end and some details about lark

index_configuration_tool/tests/logstash_test_input.conf
```diff
@@ -0,0 +1,40 @@
+input {
```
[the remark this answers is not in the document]
kartg: Moved input and output files to a `resources` directory under `tests`

index_configuration_tool/requirements.txt
```diff
@@ -0,0 +1,5 @@
+jsondiff>=2.0.0
+lark>=1.1.5
+requests>=2.28.2
+responses>=0.23.1
+coverage>=7.2.3
```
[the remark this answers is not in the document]
kartg: Moved development-only dependencies to `dev-requirements.txt` and updated the README

index_configuration_tool/README.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Docker
+
+Replace `<path-to-Logstash-config-file>` in the command below with the path to your Logstash config file.
+
+```shell
+confPath=<path-to-Logstash-config-file>; docker run -v $confPath:/tmp/conf.json -t kartg/index-configuration-tool /tmp/conf.json
+```
+
+## Development
```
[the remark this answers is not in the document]
kartg: Copied a part of the writeup in [the earlier PR](https://github.com/opensearch-project/opensearch-migrations/pull/141) as a short write-up at the bottom of the README, under `Development`

mikaylathompson (approved):

lewijacn (approved):

PR #146, opened 2023-04-25
[Traffic Capture] MIGRATIONS-1063: Add kafka producer offloader

lewijacn:
### Description
Introduces new Kafka producer offloader which will offload protobuf messages to one of two Kafka topics. The first topic is a mutating topic for mutating requests/responses. The second topic is for non-mutating requests/responses. The goal being to allow non-mutating requests to flow freely while "blocking" on mutating requests.

The LoggingProxy accepts a file path for a Kafka properties file, see README for an example, which will contain the broker endpoints that the offloader will communicate with.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1063

### Testing
Manual testing with both local Kafka cluster and MSK

### Check List


TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpRequestHandler.java
```diff
@@ -26,6 +27,8 @@ public ConditionallyReliableLoggingHttpRequestHandler(IChannelConnectionCaptureS
     @Override
     protected void channelFinishedReadingAnHttpMessage(ChannelHandlerContext ctx, Object msg, DefaultHttpRequest httpRequest) throws Exception {
         if (shouldBlockPredicate.test(httpRequest)) {
+            ctx.channel().attr(AttributeKey.valueOf("isBlocking")).set(true);
```
gregschohn, line 30: Please make isBlocking a final static constant.  Keep in mind that this is not the only time that flush() will be called (the serializer may do it on its own).

gregschohn: We should poke around, and this will be largely subject to change shortly as the HTTP decoder is rewritten, but the first line/headers of a message may be present, or could be present, after they've been processed.

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpResponseHandler.java
```diff
@@ -102,6 +104,12 @@ public void deregister(ChannelHandlerContext ctx, ChannelPromise promise) throws
 
     @Override
     public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
+        // Enable responses for blocking requests to have the same blocking metadata
+        Attribute<Object> attr = ctx.channel().attr(AttributeKey.valueOf("isBlocking"));
+        if (attr.get() != null) {
+            trafficOffloader.setIsBlockingMetadata(true);
+            attr.set(null);
```
gregschohn, line 111: I don't understand what is happening here.  What if the attribute was set to "false"?  Why are you setting the blocking flag once, but then unsetting it after that?  It seems like this conditional should never be in here.  Instead, when the metadata that you need is found, it should be sent to the serializer and that should have the responsibility to track it.  

As it is now, I'm presuming that the request's offloading no longer needs to block.  While that is true, that behavior 1) should not be decided here, nor is it.  2) You're not passing the "isBlocking" flag to indicate or control blocking behavior, but to route messages to the right topic and splitting a request/response pair between multiple topics is incorrect.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -40,13 +38,14 @@ private static Settings getSettings(@NonNull String configFile) {
...
-        return new FileConnectionCaptureFactory(traceLogsDirectory);
+    private static IConnectionCaptureFactory getConnectionCaptureFactory(String kafkaPropsPath) throws IOException {
+        //return new FileConnectionCaptureFactory("./traceLogs");
+        return new KafkaCaptureFactory(kafkaPropsPath);
     }
 
     public static void main(String[] args) throws InterruptedException, IOException {
 
-        String traceLogsDirectory = args[0];
+        String kafkaPropsPath = args[0];
```
gregschohn, line 48: We really need to invest in a command line parser for the short term so that we can easily keep the file logging in place (though it may not stay long term).  Longer term, all of the offloader libraries should be implemented as ServiceProviders.

TrafficCapture/captureKafkaOffloader/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+sourceSets {
+    main {
+        java {
+            srcDirs 'src/main/java'
+        }
+    }
+}
+dependencies {
+    api group: 'io.netty', name: 'netty-buffer', version: '4.1.89.Final'
+    implementation project(':captureOffloader')
```
gregschohn, line 20: I'd say, please pick one addressing format over the other, but we shouldn't probably pick one for the entire solution & I'm torn between the two.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -0,0 +1,95 @@
...
+                codedStreamToByteStreamMap.put(cos, baos);
+                return cos;
+            },
+            (offloaderInput) -> {
+                try {
+                    CodedOutputStream cos = offloaderInput.getCodedOutputStream();
+                    CaptureOutputStreamMetadata metadata = offloaderInput.getCaptureOutputStreamMetadata();
+                    ByteArrayOutputStream byteStream = codedStreamToByteStreamMap.get(cos);
+                    byte[] bytes = byteStream.toByteArray();
+                    byteStream.close();
```
gregschohn, line 70: close() can throw an IOException, though I honestly cannot think of why it would in this case.  If it did though, everything after this would not be run, including the removal from the map.

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/model/OffloaderInput.java
```diff
@@ -0,0 +1,24 @@
+package org.opensearch.migrations.trafficcapture.model;
+
+import com.google.protobuf.CodedOutputStream;
+
+import java.util.concurrent.CompletableFuture;
+
+public class OffloaderInput {
+
+    private CodedOutputStream codedOutputStream;
+    private CaptureOutputStreamMetadata captureOutputStreamMetadata;
```
gregschohn, line 10: See the comment above about inlining the contents of the metadata into this class.

gregschohn (commented):
Thanks Tanner!  This is a good next step!  
I've left a lot of comments around making the code both more performant and (I hope) more efficient.  Some of the work might make sense to tackle right away in this PR and other work might make more sense in the immediate future.

I would like to see some unit tests, but after some of the correctness issues get tightened up.  For example, will topic routing stay consistent for all flushes?

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -0,0 +1,95 @@
...
+        this(producerPropertiesPath, DEFAULT_MUTATING_TOPIC, DEFAULT_NON_MUTATING_TOPIC);
+    }
+
+    @Override
+    public IChannelConnectionCaptureSerializer createOffloader(String connectionId) throws IOException {
+        AtomicLong supplierCallCounter = new AtomicLong();
+        Map<CodedOutputStream, ByteArrayOutputStream> codedStreamToByteStreamMap = new HashMap<>();
+        return new StreamChannelConnectionCaptureSerializer(connectionId,
+            () -> {
+                // Potentially enhance to use java nio ByteBuffer here if we know size beforehand
```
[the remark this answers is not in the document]
lewijacn: Updated to use nio byte buffer. My concern here is that it seems expensive to shrink the initial byte array of 1MB to its actual finished size for each message. On the flip side we never have to do resizing which would be a drawback of the BAOS

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -293,6 +298,11 @@ public void commitEndOfHttpMessageIndicator(Instant timestamp)
         this.headersByteLength = -1;
     }
 
+    @Override
+    public void setIsBlockingMetadata(boolean isBlocking) {
```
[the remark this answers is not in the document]
lewijacn: Take a look at my update. I want to move to making the Serializer more generic and storing the metadata somewhere else (currently in an extension of the Serializer). I feel the change gets us closer there but there still some untangling that feels like needs to be done to make the Serializer generic enough that extensions of it have more control over things like the close handler. Probably a good spot to discuss how we may want to do this

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpResponseHandler.java
lewijacn: This section and the other section for attribute were more of a hack to get things in place for supporting multiple topics. I've removed them now and just let the serializer dictate this in a similar way. However, this piece is still not complete. I feel we need to add some mechanism for linking request and responses together through some UUID or other mechanism. Currently there doesn't seem to be a reliable way to tell what response is for which request. Let me know if you have thoughts on this.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
lewijacn: Agreed

lewijacn (commented):
@gregschohn thanks for comments, added some updates and replies, take a look

lewijacn:
After some discussion with Greg have decided to keep the Kafka producer limited to one topic for the time being. There are some complexities that come with matching requests to responses as they progress through the pipeline and ensuring that the request/response pair ends up on the same topic with the mindset of blocking only when we must. Along with this dealing with mutating and non-mutating requests competing for the same CodedOutputStream presents its own challenges to think through. We may come back to implementing multiple topics here in the near future but for the time being it seems worthwhile to look into other options that could achieve this goal (such as something like a Kafka process to split a topic) and keep our traffic capture as quick and as simple as we can

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicLong;
+
+@Slf4j
+public class KafkaCaptureFactory implements IConnectionCaptureFactory {
+
+    private static final String DEFAULT_MUTATING_TOPIC = "logging-mutating-topic";
+    private static final String DEFAULT_NON_MUTATING_TOPIC = "logging-non-mutating-topic";
+
+    private final Producer<String, byte[]> producer;
```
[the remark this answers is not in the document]
gregschohn: After a quick 10+ minute check, I don't see how one can interface with the Kafka client can produce messages directly using nio to serialize objects directly into the buffers that would be destined for the network.  I'm not sure that Kinesis supports this either.
I wouldn't be surprised to find SOME implementation that will work, but this can be filed away as a premature optimization for now.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -0,0 +1,88 @@
...
+                try {
+                    ByteBuffer byteBuffer = codedStreamToByteStreamMap.get(codedOutputStream);
+                    codedStreamToByteStreamMap.remove(codedOutputStream);
+                    ProducerRecord<String, byte[]> record = new ProducerRecord<>(topicNameForTraffic,
+                        String.format("%s_%d", connectionId, supplierCallCounter.incrementAndGet()),
+                        Arrays.copyOfRange(byteBuffer.array(), 0, byteBuffer.position()));
+                    // Used to essentially wrap Future returned by Producer to CompletableFuture
+                    CompletableFuture cf = new CompletableFuture<>();
+                    // Async request to Kafka cluster
+                    producer.send(record, handleProducerRecordSent(cf));
```
gregschohn, line 71: I've also noticed in the Kafka documentation that the Producer deals with buffering - can you confirm that handleProducerRecordSent will fire only after the message has been ack'ed by the broker (& not just buffered, or even sent)?

TrafficCapture/captureOffloader/src/test/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializerTest.java
```diff
@@ -99,8 +99,9 @@ public void testThatReadCanBeDeserialized() throws IOException, ExecutionExcepti
                     codedStreamToByteBuffersMap.put(rval, bytes);
                     return rval;
                 },
-                cos-> CompletableFuture.runAsync(() -> {
+            (codedOutputStream) -> CompletableFuture.runAsync(() -> {
                     try {
+                        CodedOutputStream cos = codedOutputStream;
```
gregschohn, line 104: What do you need this line for?  Looks superfluous.

TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/InMemoryConnectionCaptureFactory.java
```diff
@@ -43,8 +43,9 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
             var cos = CodedOutputStream.newInstance(baos);
             codedStreamToFileStreamMap.put(cos, baos);
             return cos;
-        }, (stream) -> CompletableFuture.runAsync(() -> {
+        }, (codedOutputStream) -> CompletableFuture.runAsync(() -> {
             try {
+                CodedOutputStream stream = codedOutputStream;
```
gregschohn, line 48: Is this necessary?

gregschohn (commented):
Wow - with just one topic, this is way simpler & the right way to go, at least for now!
I've left a couple comments on efficiency and correctness that I'd like to see addressed.
The correctness one (when has a message been sent) might involve tracking down some documents and adding citations along with possibly some tests.

gregschohn (approved):

PR #157, opened 2023-05-10
[Traffic Capture] Add multipart support for offloading messages

lewijacn:
### Description
Some incoming packets will be larger than what our offloading protocols (e.g. Kafka) can handle.  This change chops those messages into pieces (while retaining proper protobuf format) and sends them to the offloader. The offloader will provide the `CodedOutputStream` and thus can set the max desired size of the message which will be determined by the size of the `CodedOutputStream` ~~and can also provide the minimum chunk size which will limit the maximum number of data bytes that will be chunked~~. This functionality required minimal change for offloaders (Kafka, File, InMemory) and was mainly rooted in the Serializer.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1061

### Testing
Manual testing with local kafka cluster and file directory and unit tests

### Check List

### DIFF
--- a/TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
+++ b/TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
@@ -18,6 +18,7 @@ import java.util.Properties;
 import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
 
 @Slf4j
 public class KafkaCaptureFactory implements IConnectionCaptureFactory {
@@ -49,12 +50,11 @@ public class KafkaCaptureFactory implements IConnectionCaptureFactory {
     @Override
     public IChannelConnectionCaptureSerializer createOffloader(String connectionId) throws IOException {
         AtomicLong supplierCallCounter = new AtomicLong();
+        AtomicReference<CompletableFuture> aggregateCf = new AtomicReference<>(CompletableFuture.completedFuture(null));
         WeakHashMap<CodedOutputStream, ByteBuffer> codedStreamToByteStreamMap = new WeakHashMap<>();
-        return new StreamChannelConnectionCaptureSerializer(connectionId,
+        return new StreamChannelConnectionCaptureSerializer(connectionId, 100,
             () -> {
-                // Set ByteBuffer to Kafka max message size of 1MB with 1024 bytes of leeway temporarily until
-                // serializer has been updated
-                ByteBuffer bb = ByteBuffer.allocate(1024 * 1023);
+                ByteBuffer bb = ByteBuffer.allocate(1024 * 1024);
                 var cos = CodedOutputStream.newInstance(bb);
                 codedStreamToByteStreamMap.put(cos, bb);
                 return cos;
@@ -70,7 +70,9 @@ public class KafkaCaptureFactory implements IConnectionCaptureFactory {
                     CompletableFuture cf = new CompletableFuture<>();
                     // Async request to Kafka cluster
                     producer.send(record, handleProducerRecordSent(cf, recordId));
-                    return cf;
+                    // TODO more reliable way to cut off CF tree
+                    aggregateCf.set(aggregateCf.get().isDone() ? cf : CompletableFuture.allOf(aggregateCf.get(), cf));
+                    return aggregateCf.get();
                 } catch (Exception e) {
                     throw new RuntimeException(e);
                 }
--- a/TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
+++ b/TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
@@ -24,16 +24,18 @@ public class KafkaCaptureFactory implements IConnectionCaptureFactory {
     // Potential future optimization here to use a direct buffer (e.g. nio) instead of byte array
     private final Producer<String, byte[]> producer;
     private final String topicNameForTraffic;
+    private final int bufferSize;
 
-    public KafkaCaptureFactory(Producer<String, byte[]> producer, String topicNameForTraffic) {
+    public KafkaCaptureFactory(Producer<String, byte[]> producer, String topicNameForTraffic, int bufferSize) {
         // There is likely some default timeout/retry settings we should configure here to reduce any potential blocking
         // i.e. the Kafka cluster is unavailable
         this.producer = producer;
         this.topicNameForTraffic = topicNameForTraffic;
+        this.bufferSize = bufferSize;
     }
 
-    public KafkaCaptureFactory(Producer<String, byte[]> producer) {
-        this(producer, DEFAULT_TOPIC_NAME_FOR_TRAFFIC);
+    public KafkaCaptureFactory(Producer<String, byte[]> producer, int bufferSize) {
+        this(producer, DEFAULT_TOPIC_NAME_FOR_TRAFFIC, bufferSize);
     }
 
     @Override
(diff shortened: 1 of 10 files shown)


TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -197,23 +200,84 @@ private void addStringMessage(int observationFieldNumber, int dataFieldNumber,
...
+        // from a much larger message
+        int totalMessageBytesRequired = CodedOutputStreamSizeUtil.totalBytesNeededForMessage(timestamp,
+            segmentFieldNumber, segmentDataFieldNumber, segmentCountFieldNumber, 2, byteBuffer, numFlushesSoFar + 1);
+        int totalDataSurroundingBytesRequired = totalMessageBytesRequired - byteBuffer.capacity();
+
+        if (totalDataSurroundingBytesRequired + minDataChunkBytes >= getOrCreateCodedOutputStream().spaceLeft()) {
+            flushCommitAndResetStream(false);
+        }
+
+        if (byteBuffer.limit() == 0 || totalMessageBytesRequired <= getOrCreateCodedOutputStream().spaceLeft()) {
```
gregschohn, line 229: why if byteBuffer.limit() == 0?  You should have a unit test to pick that one up.  Maybe I'm misremembering, but couldn't that cause overflow of your output?  If you have no bytes left & a 0 byte buffer, there's still overhead, right?

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -197,23 +200,84 @@ private void addStringMessage(int observationFieldNumber, int dataFieldNumber,
...
+            segmentFieldNumber = TrafficObservation.WRITESEGMENT_FIELD_NUMBER;
+            segmentCountFieldNumber = WriteSegment.COUNT_FIELD_NUMBER;
+            segmentDataFieldNumber = WriteSegment.DATA_FIELD_NUMBER;
+        }
+
+        // The required bytes here are not optimizing for space and instead are calculated on the worst case estimate of
+        // the potentially required bytes for simplicity. This could leave ~5 bytes of unused space in the CodedOutputStream
+        // when considering the case of a message that does not need segments or the case of a smaller segment created
+        // from a much larger message
+        int totalMessageBytesRequired = CodedOutputStreamSizeUtil.totalBytesNeededForMessage(timestamp,
```
gregschohn, line 221: Since you modify this value in the while loop below, you should indicate that w/ an adjective like 'left'.  I'd also include something else like 'IncludingOverhead'

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -197,23 +200,84 @@ private void addStringMessage(int observationFieldNumber, int dataFieldNumber,
...
+
+        int dataCount = 0;
+        while(byteBuffer.position() < byteBuffer.limit()) {
+            int availableCOSSpace = getOrCreateCodedOutputStream().spaceLeft();
+            int chunkBytes = totalMessageBytesRequired > availableCOSSpace ? availableCOSSpace - totalDataSurroundingBytesRequired : byteBuffer.limit() - byteBuffer.position();
+            ByteBuffer bb = byteBuffer.slice(byteBuffer.position(), chunkBytes);
+            byteBuffer.position(byteBuffer.position() + chunkBytes);
+            addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, segmentCountFieldNumber, ++dataCount, timestamp, bb);
+            // 1 to N-1 chunked messages
+            if (byteBuffer.position() < byteBuffer.limit()) {
```
gregschohn, line 242: You'd also want to flush in situations where the next observation in was too big to fit in what's left.  If your max size is 1MB though, and the overhead is 20 bytes, that would be < 1 in 50K times - so it probably doesn't matter.

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -197,23 +200,84 @@ private void addStringMessage(int observationFieldNumber, int dataFieldNumber,
...
+
+        if (byteBuffer.limit() == 0 || totalMessageBytesRequired <= getOrCreateCodedOutputStream().spaceLeft()) {
+            addSubstreamMessage(observationFieldNumber, dataFieldNumber, timestamp, byteBuffer);
+            byteBuffer.position(byteBuffer.limit());
+        }
+
+        int dataCount = 0;
+        while(byteBuffer.position() < byteBuffer.limit()) {
+            int availableCOSSpace = getOrCreateCodedOutputStream().spaceLeft();
+            int chunkBytes = totalMessageBytesRequired > availableCOSSpace ? availableCOSSpace - totalDataSurroundingBytesRequired : byteBuffer.limit() - byteBuffer.position();
```
gregschohn, line 237: Upon working through the arithmetic a couple times, I think that this might be right.  I'd like to see some more exhaustive unit tests that work through a number of edge conditions.  Starting w/ a smaller & buffer size and incrementing it by 1 for a sizeable range might do just that.

gregschohn (commented):
Most of the questions are for discussion.  I don't understand the minDataSize value, nor do I think that the file or memory streams should be capped at 1MB.
One big blocker now though is on unit tests.  There's a lot of edge conditions in the arithmetic.  Dealing with byte buffers that may be pooled, especially when tossed in with some async activities, can create a number of issues as I've started to outline.  Tighten up the unit tests, checking the coverage on the new code would be helpful too to see if you've hit all of the edges/branches (both today & for any new ones).  Narrowly focused exhaustive testing seems apropos.

lewijacn (commented):
Addressing some comments still some more I need to follow up on

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -197,23 +200,84 @@ private void addStringMessage(int observationFieldNumber, int dataFieldNumber,
...
+            // 1 to N-1 chunked messages
+            if (byteBuffer.position() < byteBuffer.limit()) {
+                flushCommitAndResetStream(false);
+                totalMessageBytesRequired = totalMessageBytesRequired - chunkBytes;
+            }
+        }
+
+    }
+
+    private void addSubstreamMessage(int observationFieldNumber, int dataFieldNumber, int dataCountFieldNumber, int dataCount,
```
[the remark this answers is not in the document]
lewijacn: From looking at this there doesn't seem to be a simple division (which avoids a bunch of duplication) that makes sense as the order of writing kind of weaves them together. I think ultimately we could break out some functions and make this more logical but probably only makes sense if you have strong preference here

lewijacn (commented):
Following up on rest of the comments and adding improvements

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -70,7 +60,9 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
                     CompletableFuture cf = new CompletableFuture<>();
                     // Async request to Kafka cluster
                     producer.send(record, handleProducerRecordSent(cf, recordId));
-                    return cf;
+                    // A more desirable way to cut off our tree of cf aggregation should be investigated
+                    aggregateCf.set(aggregateCf.get().isDone() ? cf : CompletableFuture.allOf(aggregateCf.get(), cf));
```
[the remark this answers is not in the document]
gregschohn: The order of messages processed by Kafka.  Each one will fire independent of previous ones that were fired.  We'll only have the last close() call be aware of when all of the pieces have been committed (in some order).

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
gregschohn: Don't worry about it.  I was thinking about speculating ahead - like if any of the next messages would be too big too, then flush now before the next round.  However, I don't think that it's worth it to make that optimization at this point.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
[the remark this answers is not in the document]
lewijacn: Gotcha yes the Kafka messages are firing independent of each other but our CompletableFuture is synchronously expanding so that the final close() will be a future comprised of the previous ones

lewijacn:
Updated PR after some discussion with @gregschohn on min data chunks and setting buffer size. Adding a follow up task here to add some additional edge testing that is needed here: https://opensearch.atlassian.net/browse/MIGRATIONS-1097

gregschohn (approved):

PR #158, opened 2023-05-10
Migrations-1065 - Add Kafka Consumer - Replace CW in CDK with Kafka support

okhasawn:
### Description
This PR replaces the CloudWatch puller with a Kafka Consumer.

### Issues Resolved
[Migrations-1065](https://opensearch.atlassian.net/browse/MIGRATIONS-1065)

### Check List


TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        Properties properties = new Properties();
+        properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        properties.setProperty("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+        properties.setProperty("value.deserializer", "org.apache.kafka.common.serialization.ByteArrayDeserializer");
+        properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId);
+        properties.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+
+        KafkaConsumer<String, byte[]> consumer = new KafkaConsumer<>(properties);
+        final Thread mainThread = Thread.currentThread();
```
lewijacn, line 33: What is this used for?

deployment/docker/kafka-puller/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# syntax=docker/dockerfile:1
+
+FROM openjdk:17-jdk-slim AS cloner
+
+RUN apt-get update && apt-get install -y git
+
+# Be weary that this layer can be cached and not capture recent commits. Run docker build --no-cache to avoid this
```
lewijacn, line 7: nit: use aware instead of weary here

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        try {
+            consumer.subscribe(Collections.singleton(topic));
+
+            while (true) {
+                ConsumerRecords<String, byte[]> records =
+                        consumer.poll(Duration.ofMillis(100));
```
lewijacn, line 40: Will this poll only new records or all records?

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
```diff
@@ -0,0 +1,55 @@
...
+import java.time.Duration;
+import java.util.Collections;
+import java.util.Properties;
+
+public class KafkaPrinter {
+    private static final Logger log = LoggerFactory.getLogger(KafkaPrinter.class);
+
+    public static void main(String[] args) {
+        String bootstrapServers = args[1];
+        String groupId = "GroupID";
```
lewijacn, line 22: Could we name this something like `default-logging-group`? I know we aren't using right now as we only have a single consumer so not a big deal, but in future I imagine we may pass this in

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
```diff
@@ -0,0 +1,55 @@
...
+        String topic = args[2];
+
+        Properties properties = new Properties();
+        properties.setProperty(ConsumerConfig.BOOTSTRAP_SERVERS_CONFIG, bootstrapServers);
+        properties.setProperty("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+        properties.setProperty("value.deserializer", "org.apache.kafka.common.serialization.ByteArrayDeserializer");
+        properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId);
+        properties.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
lewijacn, line 30: What does this `AUTO_OFFSET_RESET_CONFIG` do?

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
```diff
@@ -0,0 +1,55 @@
...
+
+        try {
+            consumer.subscribe(Collections.singleton(topic));
+
+            while (true) {
+                ConsumerRecords<String, byte[]> records =
+                        consumer.poll(Duration.ofMillis(100));
+
+                for (ConsumerRecord<String, byte[]> record : records) {
+                    log.info(record.value().toString());
```
lewijacn, line 43: It seems like we are missing the part where we write this record delimited to stdout

deployment/docker/kafka-puller/run_app.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+bootstrap_server="$1"
+topic_name="$2"
+./gradlew build
+
+while true
+do
+  nc -v -l -p 9220 |  java -cp build/libs/KafkaPrinter.jar:"$(ls extracted/*/*/*.jar | tr \\n :)" org.opensearch.KafkaPrinter -o "$target_endpoint" "$topic_name"
```
lewijacn, line 9: Any issue running this java command locally? It would probably be good to have a comment describing what this java command is doing and maybe even better to store the result of the command to get the jars in a variable to use so that this line is a bit simpler

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -53,7 +53,9 @@ export class StackComposer {
         const openAccessPolicyEnabled = getContextForType('openAccessPolicyEnabled', 'boolean')
         const availabilityZoneCount = getContextForType('availabilityZoneCount', 'number')
         const migrationAssistanceEnabled = getContextForType('migrationAssistanceEnabled', 'boolean')
-        const sourceCWLogStreamARN = getContextForType('sourceCWLogStreamARN', 'string')
+        const MSKARN = getContextForType('MSKARN', 'string')
```
lewijacn, line 56: Could you add these three options to the Configuration Options table in the README. It will be helpful for others to be able to use

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -53,7 +53,9 @@ export class StackComposer {
...
+        const MSKBrokers = getContextForType('MSKBrokers', 'object')
+        const MSKTopic = getContextForType('MSKTopic', 'string')
```
lewijacn, line 58: Any issue when providing these options and having `migrationAssistanceEnabled` as true and running `cdk synth "*"`?

lewijacn (commented):
Added some comments, please reach out if things don't make sense

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
okhasawn: I had this code setup for multiple consumers in different threads, so this line was forgotten, I'll remove that.

deployment/docker/kafka-puller/Dockerfile
okhasawn: I saw Mikayla's comment on your PR earlier but forgot to change it after running into some weird gradle and docker issues.

deployment/docker/kafka-puller/run_app.sh
okhasawn: No issue running this locally. Great point, I'll add comments and store the result of the `ls` command.

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
okhasawn: Will do!

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
okhasawn: The first time it polls, it will pull all records, then it will start polling for new records.

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
okhasawn: This has to do with the poll command, having it set to "earliest" means it will poll all records the first time it's ran.

TrafficCapture/KafkaPrinter/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.7.1-bin.zip
```
lewijacn, line 3: Can you update this to 8.0.2 I don't think jdk 17 of your dockerfile image will work with this earlier version

deployment/docker/kafka-puller/run_app.sh
```diff
@@ -0,0 +1,14 @@
...
+
+bootstrap_server="$1"
+topic_name="$2"
+./gradlew build
+# The below variable will contain the classpath for all the jar dependencies.
+jars=$(ls extracted/*/*/*.jar | tr \\n :)
+
+while true
+do
+  nc -v -l -p 9220 |  java -cp build/libs/KafkaPrinter.jar:"$jars" org.opensearch.KafkaPrinter -o "$target_endpoint" "$topic_name"
```
lewijacn, line 11: Target_endpoint variable no longer exists and needs to be updated to bootstrap_server

deployment/docker/kafka-puller/run_app.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+bootstrap_server="$1"
+topic_name="$2"
+./gradlew build
```
lewijacn, line 5: Seems like we should be doing the build in the dockerfile not here

deployment/docker/kafka-puller/run_app.sh
lewijacn, line 11: Netcat is not piping in here but rather the std out of the jar should be piped to netcat and thus to the Replayer (which should be 9210), similar to:
`java -cp build/libs/KafkaPrinter.jar:"$jars" org.opensearch.KafkaPrinter -o "$bootstrap_server" "$topic_name" | nc -v localhost 9210`

deployment/docker/kafka-puller/run_app.sh
gregschohn: +1

gregschohn (approved):

PR #167, opened 2023-05-25
Include minor fixes identified from demo

lewijacn:
### Description
This change includes minor fixes that were identified when putting pieces together for a demo. Additional pieces that will be added elsewhere to get complete setup shown in demo include:
* Jupyter container CDK setup
* MSK/Kafka cluster deployed to AWS
* Zero Latency fix (to be added with/after PR for E2E from Greg)

### Issues Resolved
None

### Testing
Manual testing

### Check List

### DIFF
--- a/TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
+++ b/TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
@@ -59,10 +59,10 @@ public class KafkaPrinter {
         properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId);
         properties.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
         // Required for using SASL auth with MSK public endpoint
-        properties.setProperty("security.protocol", "SASL_SSL");
-        properties.setProperty("sasl.mechanism", "AWS_MSK_IAM");
-        properties.setProperty("sasl.jaas.config", "software.amazon.msk.auth.iam.IAMLoginModule required;");
-        properties.setProperty("sasl.client.callback.handler.class", "software.amazon.msk.auth.iam.IAMClientCallbackHandler");
+        //properties.setProperty("security.protocol", "SASL_SSL");
+        //properties.setProperty("sasl.mechanism", "AWS_MSK_IAM");
+        //properties.setProperty("sasl.jaas.config", "software.amazon.msk.auth.iam.IAMLoginModule required;");
+        //properties.setProperty("sasl.client.callback.handler.class", "software.amazon.msk.auth.iam.IAMClientCallbackHandler");
 
         KafkaConsumer<String, byte[]> consumer = new KafkaConsumer<>(properties);
 


TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
```diff
@@ -58,6 +58,11 @@ public static void main(String[] args) {
         properties.setProperty("value.deserializer", "org.apache.kafka.common.serialization.ByteArrayDeserializer");
         properties.setProperty(ConsumerConfig.GROUP_ID_CONFIG, groupId);
         properties.setProperty(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
+        // Required for using SASL auth with MSK public endpoint
+        //properties.setProperty("security.protocol", "SASL_SSL");
+        //properties.setProperty("sasl.mechanism", "AWS_MSK_IAM");
+        //properties.setProperty("sasl.jaas.config", "software.amazon.msk.auth.iam.IAMLoginModule required;");
+        //properties.setProperty("sasl.client.callback.handler.class", "software.amazon.msk.auth.iam.IAMClientCallbackHandler");
```
okhasawn, line 65: I guess this is the zero-latency fix that you're going to uncomment after Greg's E2E PR is merged?

okhasawn (commented):
LGTM. Left a quick question on the commented out code though.

deployment/docker/kafka-puller/run_app.sh
```diff
@@ -1,14 +1,23 @@
...
 do
-  java -cp build/libs/KafkaPrinter.jar:"$jars" org.opensearch.KafkaPrinter -b "$bootstrap_server" -t "$topic_name" | nc -v localhost 9210
+
+  # Verify netcat can connect to Replayer before process begins
+  while ! nc -v -w 1 localhost 9210
+  do
+    >&2 echo "Attempting to connect to Traffic Replayer..."
+    sleep 2
+  done
+  sleep 30
```
gregschohn, line 18: I don't understand this code.  Is there a way that you can do dependencies in Fargate?  You can (& should) in compose definitions.

gregschohn: Why is this sleep 30 seconds and the one below 7?  Why do you need any sleeps at all?

okhasawn (approved):

okhasawn:
Which changes address this point "Jupyter container CDK setup"?

deployment/docker/traffic-replayer/run_app.sh
```diff
@@ -10,22 +10,22 @@ do
...
   # Allow time for TC to begin listening after terminated connection
-  sleep 10
+  sleep 7
 
   touch triples.log
 
   # These commands look a bit backwards but ultimately will receive input from CW-Puller via netcat and then
   # process with the Replayer jar before ultimately printing the produced triples to stdout via tee and sending to TC via netcat
   # Note: Unbuffer has not been verified to actually be effective here. stdbuf was ruled out for its caveat with tee (https://linux.die.net/man/1/stdbuf)
   unbuffer tail -f triples.log | tee /dev/stderr | nc -v localhost 9220 &
```
gregschohn, line 25: Why are you using a file at all here?  Why not pipe directly?  Does unbuffer matter for the case where lots of messages are going through?

deployment/docker/traffic-replayer/run_app.sh
```diff
@@ -10,22 +10,22 @@ do
...
-  nc -v -l -p 9210 | java -jar trafficReplayer-uber.jar -o triples.log "$target_endpoint"
+  nc -v -l -p 9210 | java -jar trafficReplayer-uber.jar "$target_endpoint" -o triples.log --insecure
```
gregschohn, line 26: are you unbuffering the output to triples.log from the java process?

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
lewijacn: This is required properties to be able to communicate with an MSK public endpoint. By default I want to allow all kafka clusters but want to keep settings to be able to communicate with MSK if needed

lewijacn:
> Which changes address this point "Jupyter container CDK setup"?

We don't have a PR yet but we need to do some reconfiguring of the jupyter notebook container so that it is easier to get setup and run and will be a followup PR sometime

deployment/docker/traffic-replayer/run_app.sh
gregschohn: What happens if the uberjar has multiple mains() defined within different source jars?

gregschohn (approved):
I'm approving this, but it does feel like this might be better on a branch or separate patch.  The CDK changes have long term value (though the https->http one is problematic).  The other stuff seems like it might be better on a branch - but I know that the kafka puller and containers are in need of rework anyway.

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
okhasawn: Oh I see, I misunderstood the description of the PR which made me get to this assumption. I apologize.

deployment/docker/kafka-puller/run_app.sh
lewijacn: This is ugly code and needs to be reworked in future change. Sleeps here are a "hack" to allow time for the replayer and comparator to start up

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
```diff
@@ -111,7 +119,7 @@ export class MigrationAssistanceStack extends Stack {
             image: ContainerImage.fromDockerImageAsset(trafficReplayerImage),
             // Add in region and stage
             containerName: "traffic-replayer",
-            environment: {"TARGET_CLUSTER_ENDPOINT": "https://" + props.targetEndpoint + ":80"},
+            environment: {"TARGET_CLUSTER_ENDPOINT": "http://" + props.targetEndpoint + ":80"},
```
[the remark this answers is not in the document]
lewijacn: Yeah currently we are building Domains that allow http requests but this is not the default we will want long term

deployment/docker/traffic-replayer/run_app.sh
lewijacn: I believe at the time we were running into issue with having netcat receiving and sending in the same line. That may not be present now, but ideally we move away from netcat anyway

TrafficCapture/KafkaPrinter/src/main/java/org/opensearch/KafkaPrinter.java
lewijacn: Yes ideally I would like to have a config file here

deployment/docker/traffic-replayer/run_app.sh
lewijacn: We were using unbuffer as an alternative to stdbuf as it didn't work with tee commands. That said I don't know if we ever confirmed it was useful here and could possibly be removed. 

I've not ran into that issue, but goal is to move away from uberjar here.

PR #168, opened 2023-05-30
Replayer - Moving away from UberJar to classpath

okhasawn:
### Description
This PR removes the usage of an UberJar to run the replayer and instead uses the classpath.

### Check List

### DIFF
--- a/deployment/docker/traffic-replayer/run_app.sh
+++ b/deployment/docker/traffic-replayer/run_app.sh
@@ -4,6 +4,7 @@
 # Traffic Comparator (TC) as connections get setup and restarted
 
 target_endpoint="$1"
+jars=$(ls extracted/*/*/*.jar | tr \\n :)
 
 while true
 do
@@ -23,7 +24,7 @@ do
   # process with the Replayer jar before ultimately printing the produced triples to stdout via tee and sending to TC via netcat
   # Note: Unbuffer has not been verified to actually be effective here. stdbuf was ruled out for its caveat with tee (https://linux.die.net/man/1/stdbuf)
   unbuffer tail -f triples.log | tee /dev/stderr | nc -v localhost 9220 &
-  nc -v -l -p 9210 | java -jar trafficReplayer-uber.jar "$target_endpoint" -o triples.log --insecure
+  nc -v -l -p 9210 | java -cp build/libs/KafkaPrinter.jar:"$jars" org.opensearch.migrations.replay.TrafficReplayer "$target_endpoint" -o triples.log --insecure
 
   rm triple.log
   >&2 echo "Command has encountered error. Restarting now ..."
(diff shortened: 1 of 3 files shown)


deployment/docker/traffic-replayer/run_app.sh
```diff
@@ -23,7 +24,7 @@ do
   # process with the Replayer jar before ultimately printing the produced triples to stdout via tee and sending to TC via netcat
   # Note: Unbuffer has not been verified to actually be effective here. stdbuf was ruled out for its caveat with tee (https://linux.die.net/man/1/stdbuf)
   unbuffer tail -f triples.log | tee /dev/stderr | nc -v localhost 9220 &
-  nc -v -l -p 9210 | java -jar trafficReplayer-uber.jar "$target_endpoint" -o triples.log --insecure
+  nc -v -l -p 9210 | java -cp build/libs/KafkaPrinter.jar:"$jars" org.opensearch.migrations.replay.TrafficReplayer "$target_endpoint" -o triples.log --insecure
```
lewijacn, line 27: Shouldn't this be the TrafficReplayer jar instead of KafkaPrinter?

lewijacn (commented):
One comment around the jar used. Also any issue with gradle build here or building the docker image and running?

okhasawn:
> One comment around the jar used. Also any issue with gradle build here or building the docker image and running?

Since we were doing gradle assemble instead of gradle build, that is still working as intended and so does building the docker image and running it.

deployment/docker/traffic-replayer/run_app.sh
okhasawn: Thanks for catching this Tanner. Just updated it, was not included in the commit. Tested it was still working as intended after the update as well.

gregschohn (commented):
This patch has been subsumed by the latest changes to the mainline.  Take a look at the [build.gradle file for the dockerSolution project](https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/build.gradle).  
The newest approach doesn't need to do a zip and unzip.  That process is still like the uberJar step, where a lot of files that haven't changed are bound together (meaning that we're paying for things that we don't need - and invalidating a lot of caches along the way).  Both put us a step further away from being able to directly mount the build directory into the container for local testing too. 
The creation of the jarpath looks good and that's currently employed on the mainline's runApp for all of the java application docker images.

okhasawn:
> This patch has been subsumed by the latest changes to the mainline. Take a look at the [build.gradle file for the dockerSolution project](https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/build.gradle). The newest approach doesn't need to do a zip and unzip. That process is still like the uberJar step, where a lot of files that haven't changed are bound together (meaning that we're paying for things that we don't need - and invalidating a lot of caches along the way). Both put us a step further away from being able to directly mount the build directory into the container for local testing too. The creation of the jarpath looks good and that's currently employed on the mainline's runApp for all of the java application docker images.

Thank you Greg. Nice that at least we're avoiding the zip and unzip now.

okhasawn:
Closing this PR due to what's mentioned in the conversation.

PR #170, opened 2023-06-01
Capture proxy container with tls and command line args

gregschohn:
### Description

This change sets up an ES 7.10 image with search-guard so that a Capture Proxy image can be built upon that to use the TLS configs from the base image.  Additional command line arguments were used to be able to more easily manage starting the proxy in different contexts.

Search-Guard is installed (from a remote zip plugin) and is configured with the plugin's 'install_demo_configuration' script.
The elasticsearch.yml file is configured to support it coming up as a single node.

Additional changes are made to support our uses cases. Two simple scripts are added to the container to toggle TLS on/off via the configuration. TLS and HTTP auth are both disabled, as these are most useful for us at the moment.
Lastly, the image that is built from this docker file (which can also be done via gradle) is the base image for the capture proxy image. That gets us the same TLS configurations and JDK. We can choose then to run the logging proxy and Elasticsearch together in a single container, which already has all of the elasticsearch+search-guard artifacts, or to run them separately, using the base image for one container and the capture proxy image for the other.

To facilitate better dependency tracking of the 'external' docker images (those that aren't triggered by other gradle dependencies), I'm using a hash value of the Dockerfile's directory. The dependent images now use the base image of a specific tag that matches that hash value. That works well, but the docker compose file, nor anybody else, should ever have to know that tag, so two tags are created for each image, 'latest' and the hash value.

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/build.gradle
+++ b/TrafficCapture/dockerSolution/build.gradle
@@ -11,13 +11,13 @@ dependencies {
     implementation project(':trafficReplayer')
 }
 
-
 def containerServices = [
         "trafficCaptureProxyServer": "capture_proxy",
         "trafficReplayer": "traffic_replayer"
 ]
+
 containerServices.each { projectName, dockerImageName ->
-    def dockerBuildDir = "${project}/docker/${projectName}"
+    def dockerBuildDir = "build/docker/${projectName}"
     def artifactsDir = "${dockerBuildDir}/jars";
     task("copyArtifact_${projectName}", type: Copy) {
         dependsOn ":${projectName}:build"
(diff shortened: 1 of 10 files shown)


TrafficCapture/dockerSolution/build.gradle
```diff
@@ -89,6 +106,8 @@ dockerCompose {
 task buildAllDockerImages {
     dependsOn dockerBuildImage_trafficReplayer
     dependsOn dockerBuildImage_trafficCaptureProxyServer
+    dependsOn dockerBuildImage_elasticsearchWithSearchGuard
+    //dependsOn dockerBuildImage_
```
sumobrian, line 110: Don't think this is needed

TrafficCapture/dockerSolution/build.gradle
```diff
@@ -4,20 +4,57 @@ plugins {
     id "com.bmuschko.docker-java-application" version "9.3.1"
 }
 
+import java.security.MessageDigest
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+import com.bmuschko.gradle.docker.tasks.image.DockerTagImage
+
+def calculateHash(File file) {
```
mikaylathompson, line 11: Is this code still necessary? Or is it replaced by the function below?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -25,12 +29,85 @@
...
+        int maximumTrafficStreamSize = 1024*1024;
+        @Parameter(required = false,
+                names = {"--destinationHost"},
+                arity = 1,
+                description = "Hostname of the server that the proxy is capturing traffic for.")
+        String backsideHostname = "localhost";
+        @Parameter(required = true,
+                names = {"--destinationPort"},
+                arity = 1,
+                description = "Port of the server that the proxy is capturing traffic for.")
```
lewijacn, line 83: Why is this and line 88 the same?

TrafficCapture/dockerSolution/src/main/docker/elasticsearchWithSearchGuard/disableTlsConfig.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+sed -i "s/searchguard.ssl.http.enabled: *true/searchguard.ssl.http.enabled: false/g" "$@"
```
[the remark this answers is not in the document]
sumobrian: Nvm - Something to be considered when this is used for customers source cluster

sumobrian (approved):

mikaylathompson (approved):

PR #175, opened 2023-06-01
Migrations-1129 - Script for testing E2E solution.

okhasawn:
### Description
This PR's purpose is to add a script that runs tests which will verify the setup and our tools are working properly.

### Issue Solved
[Migrations-1129](https://opensearch.atlassian.net/browse/MIGRATIONS-1129)

### Check List


test/operations.py
```diff
@@ -0,0 +1,78 @@
...
+        else:
+            print('Created index successfully')
+            print(response.text)
+
+        pass
+
+    @staticmethod
+    def delete_index(endpoint, index_name, auth=None, data=None):
+        response = requests.delete(f'{endpoint}/{index_name}', auth=auth)
+        if response.status_code != 200:
```
sumobrian, line 19: I recommend using this library for status codes: https://docs.python.org/3/library/http.html#http.HTTPStatus

test/operations.py
```diff
@@ -0,0 +1,78 @@
...
+        else:
+            print('Created document successfully')
+            print(response.text)
+        pass
+
+
+
+def main():
+
+    username = '*'  # Enter master username and password
```
sumobrian, line 63: These should be pulled in from environment variables.

test/operations.py
```diff
@@ -0,0 +1,78 @@
...
+    password = '*'
+    auth = (username, password)
+    endpoint = '*:443'  # Replace * with domain endpoint
```
sumobrian, line 66: Same - add environment variable w/ readme on whats required.

sumobrian (changes requested):
Next steps: create unit tests that use this class and compare source and target cluster resutls.

test/operations.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            print('Created document successfully')
+            print(response.text)
+
+        return response
+
+
+def main():
+
+    username = os.getenv('USERNAME')
```
sumobrian, line 79: Add default

test/operations.py
```diff
@@ -0,0 +1,94 @@
+from http import HTTPStatus
+import requests
+import json
+import os
+
```
sumobrian, line 5: import logging facility and use info/warn/error levels

test/tests.py
```diff
@@ -0,0 +1,60 @@
...
+        self.assert_(response1.status_code == HTTPStatus.CREATED and response2.status_code == HTTPStatus.CREATED)
+
+        response1 = Operations.delete_document(endpoint1, index, doc_id)
+        response2 = Operations.delete_document(endpoint2, index, doc_id)
+        self.assert_(response1.status_code == HTTPStatus.OK and response2.status_code == HTTPStatus.OK)
+
+        response1 = Operations.delete_index(endpoint1, index)
+        response2 = Operations.delete_index(endpoint2, index)
+        # TODO: check comparator's results here, and add more logging to know where exactly the test fails, if it does.
+        self.assert_(response1.status_code == HTTPStatus.OK and response2.status_code == HTTPStatus.OK)
```
sumobrian, line 48: use assertEqual and try to eliminate compound checks. It's more difficult to debug.

test/operations.py
chelma: +1; print statements are usually the wrong answer.  Here's an example of a logger that you can use to get a lot more granularity over what/how things get emitted:

https://github.com/arkime/aws-aio/blob/main/manage_arkime/logging_wrangler.py#L39

test/operations.py
```diff
@@ -0,0 +1,94 @@
...
+
+class Operations:
```
chelma, line 7: Generally speaking, you only need a class if you're storing state.  In this case, we're not storing state so we're just using it as a namespace, which is fine... but a bit confusing.  Recommend you just turn these into module-level functions for clarity.

test/operations.py
```diff
@@ -0,0 +1,94 @@
...
+    @staticmethod
+    def create_index(endpoint, index_name, auth=None, data=None):
```
chelma, line 9: I don't see the `auth` and `data` arguments being used; is that something you're going to add later?

Also - I recommend you add type hints to your Python code.

test/tests.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class MyTestCase(unittest.TestCase):
+    def test_index(self):
+        endpoint1 = os.getenv('ENDPOINT_1', 'http://localhost:9200')
+        endpoint2 = os.getenv('ENDPOINT_2')
+        index = "my_index"
+        response1 = Operations.create_index(endpoint1, index)
+        response2 = Operations.create_index(endpoint2, index)
+        self.assert_(response1.status_code == HTTPStatus.OK and response2.status_code == HTTPStatus.OK)
```
chelma, line 20: Please separate these so we more easily know which expectation failed when the tests fail

test/operations.py
```diff
@@ -0,0 +1,94 @@
...
+        response = requests.put(f'{endpoint}/{index_name}', auth=auth)
+        if response.status_code != HTTPStatus.OK:
+            print('Failed to create index')
+            print(response.text)
+        else:
+            print('Created index successfully')
+            print(response.text)
```
chelma, line 16: I don't see what the value of these print statements (or log statements, if you did a replacement), since we're running this in a unit test framework that will print out a report of what went wrong with the test.  I can see some value in maybe emitting the `response.text` to STDOUT so that it's available if a test fails.

test/tests.py
```diff
@@ -0,0 +1,60 @@
...
+# This code is probably existing as a placeholder only for testing purposes,
+# final code here will be using comparator's results. Also, won't be sending request directly to target cluster,
+# and will be using the replayer instead, so we're able to check comparator's results once the replayer outputs
+# the triples log file.
+# TODO: Add endpoints to be read from environment variable.
+
+
+class MyTestCase(unittest.TestCase):
+    def test_index(self):
+        endpoint1 = os.getenv('ENDPOINT_1', 'http://localhost:9200')
```
chelma, line 15: In general, it isn't the best idea to pull test configuration during time-of-test; it's possible that the environment variable/files may change, for example, especially on long-running test suites.  It also makes is hard for maintainers to reason about the boundaries of these tests - they have to look at each and every test to figure out whether the test needs something from outside the pytest process (whether a file, an environment variable, etc).  For these reasons, I recommend pulling your out-of-process configuration for your tests in a single, unified location and then letting your tests refer to the in-memory state created by that piece of code.

Here's an example of how you can do that using pytest fixtures:
https://automationpanda.com/2023/01/12/passing-test-inputs-into-pytest/

chelma (changes requested):

okhasawn:
Addressed some of feedback on this, still some more work to do.

To run these tests, please the following command. `python -m unittest tests.py`

The following is an example output of a test run:

`(venv) okhasawn@bcd07411f8f4 test % python -m unittest tests.py
/Users/okhasawn/eng/amazon-python/venv/lib/python3.10/site-packages/urllib3/connectionpool.py:1045: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
  warnings.warn(
.....
Ran 5 tests in 3.607s

OK
`

test/tests.py
```diff
@@ -0,0 +1,104 @@
...
+        source_endpoint = os.getenv('SOURCE_ENDPOINT', 'https://localhost:9200')
+        target_endpoint = os.getenv('TARGET_ENDPOINT', 'https://localhost:29200')
+        username = os.getenv('username', 'admin')
+        password = os.getenv('password', 'admin')
+        auth = (username, password)
+        index = "my_index"
+
+        source_response = create_index(source_endpoint, index, auth)
+        self.assertEqual(source_response.status_code, HTTPStatus.OK)
+
```
sumobrian, line 37: I would expect some delay here or else you would get inconsistent results. Ideally this would be some sort of poll with timeout rather than a hard-coded sleep.

sumobrian: Also, when polling, print out an informational statement indicating how long you had to wait (in seconds and iterations). This information may be useful over time when comparing older runs with newer runs. There should be a maximum number of retries where, if this limit is reached, the test fails. The test can fail in two ways:

1. The timeout/retries were exceeded before an index was created.
2. An unexpected status code was returned.

sumobrian (changes requested):

test/tests.py
```diff
@@ -0,0 +1,104 @@
...
+        # Add a stall here maybe? Sometimes the check index function is performed before the delete request is replayed
+        # on the target cluster, so the check will find the index and return a code 200 instead of 404.
+
+        # TODO: check comparator's results here.
+
+        sleep(3)
+        target_response = check_index(target_endpoint, index, auth)
+        self.assertEqual(target_response.status_code, HTTPStatus.NOT_FOUND)
+
+    def test_document(self):
```
sumobrian, line 54: Can you add test numbers to identify the order you expect them to run. For example

`test_001_<name>`, `test_002_<name>`

test/tests.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def retry_request(request, args=(), max_attempts=10, delay=0.5, expectedStatusCode=None):
+    for attempt in range(1, max_attempts + 1):
+        try:
+            result = request(*args)
+            if result.status_code == expectedStatusCode:
+                return result
+            else:
+                logger.error(f"Status code returned: {result.status_code} did not"
```
sumobrian, line 22: Change this to a warning. This may not be an error until your max number of retries has been exceeded.

test/tests.py
```diff
@@ -0,0 +1,142 @@
...
+                             f" match the expected status code: {expectedStatusCode}")
+        except Exception:
+            logger.error(f"Trying again in {delay} seconds")
```
sumobrian, line 25: I would break out of this for loop in this case. This is an unknown exception and a retry will probably fail the same way. The reason you have this retry in is because of eventual consistency. Other exceptions should not be handled as a retry. Simply log error (as you are doing), then raise. Remove the sleep.

test/tests.py
sumobrian: Add sleep from below here. In your message, indicate the status code and how many seconds/ms you are sleeping for.

sumobrian (changes requested):

test/operations.py
```diff
@@ -0,0 +1,61 @@
+import requests
+import json
+from typing import Optional, Tuple
+
+
+def create_index(endpoint: str, index_name: str, auth: Optional[Tuple[str, str]] = None, data: Optional[dict] = None):
+    response = requests.put(f'{endpoint}/{index_name}', auth=auth, verify=False)
```
mikaylathompson, line 7: It doesn't look like `data` is ever used -- why is it a param for the function?

mikaylathompson: Same question for `check_index` and `delete_index`

test/operations.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def delete_document(endpoint: str, index_name: str, doc_id: str, auth: Optional[Tuple[str, str]] = None):
+    response = requests.delete(f'{endpoint}/{index_name}/_doc/{doc_id}', auth=auth, verify=False)
+
+    return response
+
+
+def create_document(endpoint: str, index_name: str, doc_id: str, auth: Optional[Tuple[str, str]] = None,
+                    data: Optional[dict] = None):
```
mikaylathompson, line 31: Can you elaborate on why this function has `data` as well? I expected `data` to be the document that was being created, but it actually doesn't seem to be used at all.

test/operations.py
```diff
@@ -0,0 +1,61 @@
...
+    }
+    url = f'{endpoint}/{index_name}/_doc/{doc_id}'
+    headers = {'Content-Type': 'application/json'}
+
+    response = requests.put(url, headers=headers, data=json.dumps(document), auth=auth, verify=False)
+
+    return response
+
+
+def check_document(endpoint, index_name, doc_id, auth=None):
```
mikaylathompson, line 44: minor: would "retrieve_document" be a better name for this? "check" sounds to me like it's going to return true/false about whether the document exists, but this just returns either the document itself or a 404, correct?

mikaylathompson: Also, I think you may have missed the type hints for this function

test/operations.py
```diff
@@ -0,0 +1,61 @@
...
+
+def get_document(endpoint: str, index_name: str, doc_id: str, auth: Optional[Tuple[str, str]] = None):
+    url = f'{endpoint}/{index_name}/_doc/{doc_id}'
+    headers = {'Content-Type': 'application/json'}
+
+    response = requests.get(url, headers=headers, auth=auth, verify=False)
+    document = response.json()
+    content = document['_source']
+
+    return content
```
mikaylathompson, line 61: Okay, I just read this function it does seem to be the "retrieve_document" I described above. In that case, I'm not sure what the purpose of `check_document` is. These two functions call the same API, right? The second just returns the original document?

I wonder if they should be combined into one function: return the document, or return None if it doesn't exist. Not sure if that's the right solution, but confused that they both exist.

test/tests.py
```diff
@@ -0,0 +1,143 @@
+from operations import create_index, check_index, create_document, delete_document, get_document, delete_index
+from operations import check_document
```
mikaylathompson, line 2: Looks like this is imported twice

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+from http import HTTPStatus
+import unittest
```
mikaylathompson, line 4: In the readme, it sounds like you're running the tests with `pytest` and you're installing `pytest` in the requirements file. Is there a reason you're using `unittest` instead of `pytest` here?

Either one works, but if you're just using `unittest`, we might as well remove the external dependency on `pytest`.

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+import logging
+import time
+import requests
+
+logger = logging.getLogger(__name__)
+
+# Tests will say which line the test failed at and what was the result of the execution, but better logging can be done.
+
+
+def retry_request(request, args=(), max_attempts=10, delay=0.5, expectedStatusCode=None):
```
mikaylathompson, line 15: non-blocking, but would be great to get type hints in this file too

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+    for attempt in range(1, max_attempts + 1):
+        try:
+            result = request(*args)
+            if result.status_code == expectedStatusCode:
+                return result
+            else:
+                logger.warning(f"Status code returned: {result.status_code} did not"
```
mikaylathompson, line 22: I think it would be helpful if this log statement included which request is being attempted.

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+                               f" match the expected status code: {expectedStatusCode}."
+                               f" Trying again in {delay} seconds.")
+                time.sleep(delay)
+        except Exception:
+            logger.error(f"All {max_attempts} attempts failed.")
+            raise Exception(f"Couldn't get the expected status code: {expectedStatusCode} even after attempting to"
```
mikaylathompson, line 28: I think particularly important to include which request it is in the exception message.

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def tearDown(self):
+        proxy_endpoint = os.getenv('SOURCE_ENDPOINT', 'https://localhost:9200')
+        username = os.getenv('username', 'admin')
+        password = os.getenv('password', 'admin')
+        auth = (username, password)
+        index = "my_index"
+        doc_id = '7'
+        delete_index(proxy_endpoint, index, auth)
+        delete_document(proxy_endpoint, index, doc_id, auth)
```
mikaylathompson, line 51: These setup and teardown methods are identical, right? It might make sense to pull the logic out to a new function and call it from both of these places.

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+        password = os.getenv('password', 'admin')
+        auth = (username, password)
+        index = "my_index"
+
+        proxy_response = create_index(proxy_endpoint, index, auth)
+        self.assertEqual(proxy_response.status_code, HTTPStatus.OK)
+
+        target_response = retry_request(check_index, args=(target_endpoint, index, auth),
+                                        expectedStatusCode=HTTPStatus.OK)
+        self.assertEqual(target_response.status_code, HTTPStatus.OK)
```
mikaylathompson, line 67: Out of curiosity, based on how the `retry_request` is written, wouldn't we have already failed with an exception on the previous line and never get to this one if the status code _isn't_ HTTPStatus.OK? Is there any case where we actually execute this line, but the status code isn't 200?

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+        delete_document(proxy_endpoint, index, doc_id, auth)
+
+    def test_001_index(self):
+        proxy_endpoint = os.getenv('PROXY_ENDPOINT', 'https://localhost:9200')
+        source_endpoint = os.getenv('SOURCE_ENDPOINT', 'http://localhost:19200')
+        target_endpoint = os.getenv('TARGET_ENDPOINT', 'https://localhost:29200')
+        username = os.getenv('username', 'admin')
+        password = os.getenv('password', 'admin')
+        auth = (username, password)
+        index = "my_index"
```
mikaylathompson, line 60: non-blocking, but nice code cleanliness option:

You can use your `setUp` method to take care of a lot of this boilerplate code. If you do 
```
self.proxy_endpoint = ...
...
self.auth = ...
self.index = ...
```
in your `setUp` method, then in each of these `test_00X..` methods, you can skip these first 7 lines and go straight to:
```
proxy_response = create_index(self.proxy_endpoint, self.index, self.auth)
```

It saves you a lot of repeated code, but also guarantees that if you change the name of the index, you only have to update it in one place and all of your tests will be updated automatically.

test/tests.py
```diff
@@ -0,0 +1,143 @@
...
+        proxy_endpoint = os.getenv('SOURCE_ENDPOINT', 'https://localhost:9200')
+        username = os.getenv('username', 'admin')
+        password = os.getenv('password', 'admin')
+        auth = (username, password)
+        index = "my_index"
+        doc_id = '7'
+        delete_index(proxy_endpoint, index, auth)
+        delete_document(proxy_endpoint, index, doc_id, auth)
+
+    def test_001_index(self):
```
mikaylathompson, line 53: I would love a human description of what happens in each of these tests. I can figure it out from the code, but a high level "this tests creates an index by calling the proxy, which causes ... It verifies that ..." would be very helpful.

test/tests.py
okhasawn: I preferred using `unittest` to implement the tests as these tests do not require any of the advantages that `pytest `provides over `unittest` and since its' way of running `setup and teardown` before _each and every_ test without having to explicitly mention that made the code look simpler and cleaner in my opinion.
BUT, the output of running the tests using `pytest` provides a lot more details in the way that unittest could not do, even with some extra configuration.

test/tests.py
okhasawn: Thanks for the feedback here, I just rewrote the `retry_request` function removing the whole exception part as I don't think it's needed due to the assert here. Also, following your feedback to improve the logging messages with more details got me thinking that the exception is not needed.

okhasawn:
Thanks for the excellent feedback @mikaylathompson!! I just pushed in 2 commits addressing _most_ of it, with what's left being mainly adding comments and improve the documentation.

mikaylathompson (commented):
Changes look great! Just a couple small things.

README.md
```diff
@@ -20,11 +20,40 @@ traffic, which can then be offloaded and replayed to other HTTP server(s).
...
+pytest tests.py
+```
+
+#### Notes 
+##### Ports Setup
+The test script, by default, uses the ports assigned to the containers in this
+[docker-compose file](TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml), so if the docker solution in
+it's current setup started with no issues, then the test script will run as is. If for any reason
+the user changed the ports in that file, they must also either, change the following environment variables:
+`PROXY_ENDPOINT`, `SOURCE_ENDPOINT`, `TARGET_ENDPOINT` and `JUPYTER_NOTEBOOK` respectively, or update the default value
```
sumobrian, line 45: I would state the default here and list them in a bulleted list.

README.md
```diff
@@ -20,11 +20,40 @@ traffic, which can then be offloaded and replayed to other HTTP server(s).
...
+for them in [tests.py](test/tests.py). Those are the only endpoints touched by this script.
+#### Clean Up
+The test script is implemented with a setup and teardown functions that are ran after 
+each and every test where additions made to the endpoints are deleted, *mostly* cleaning up after themselves, however,
+as we log all operations going through the proxy (which is capturing the traffic), those are only being 
+deleted after the docker solution is shut down.
```
sumobrian, line 51: nit: Couple of instances where docker is not capitalized should be updated.

test/tests.py
```diff
@@ -0,0 +1,142 @@
...
+
+# The following "retry_request" function's purpose is to retry a certain request for "max_attempts"
+# times every "delay" seconds IF the requests returned a status code other than what's expected.
+# So this "retry_request" function's arguments are a request function's name and whatever arguments that function
+# expects, and the status code the request function is expecting to get.
+def retry_request(request: Callable, args: Tuple = (), max_attempts: int = 10, delay: float = 0.5,
+                  expected_status_code: HTTPStatus = None):
+    for attempt in range(1, max_attempts + 1):
+
+        result = request(*args)
```
sumobrian, line 22: Where is the exception we discussed? Based on the demo conversation I thought you were going to add this back in and handle connection refused errors?

sumobrian (changes requested):

test/tests.py
okhasawn: I spoke with Mikayla after the demo conversation, we agreed that there are two valid point of views here:
- Adding the exception handling + make the first time each endpoint has a request made to retry-able.
- Acknowledge that the exception handling we discussed would not be necessary as the code would not be reached if the prerequisites (making sure all the containers are up) for the test script are followed properly.

I pushed the commit I had prepared which follows the second point of view, but later realized that it can still be helpful in cases such as the containers are up but ports are not set properly, therefore it can still be helpful to the user of this test. Pushing the commit with the exception handling now. It will handle all cases where any of the containers are down.

test/tests.py
```diff
@@ -0,0 +1,161 @@
...
+        except SSLError as e:
+            logger.error(f"Received exception: {e}. Unable to connect to server. Please check all containers are up"
+                         f"and ports are setup properly")
+            logger.warning(f"Trying again in {delay} seconds.")
+            time.sleep(delay)
+            continue
+    logger.error(f"All {max_attempts} attempts failed.")
+    logger.error(f"Couldn't get the expected status code: {expected_status_code} while making the request:"
+                 f"{request.__name__} using the following arguments: {args} ")
+    return None
```
sumobrian, line 46: If you are going to return None rather than raise the exception you should check for this case. Otherwise your tests below can fail with an attribute error.

okhasawn: Thanks for pointing this out. I didn't think it was an issue since if the tests do fail with an attribute error, the logs would mention that an exception had happened and would show the retry attempts, but now just realized that it this will give more clarity to the user, and clearer test results. Just pushed a commit to address that.

sumobrian (approved):

PR #179, opened 2023-06-01
Http json redrives

gregschohn:
### Description
The HttpJsonTransformer builds up a netty pipeline on an EmbeddedChannel.  Early in the process (after NettyDecodedHttpRequestHandler receives the parsed header), it might be determined that the incoming HTTP message doesn't need any kind of transformation.  At that point, previously all of the subsequent handlers were removed and we attempted to just pass the raw packets directly through.

This was problematic for messages that had bodies in the same ByteBuf as headers.  That meant that some of the contents were being processed still by the previous handler, which hadn't yet fired them through the pipeline.  When the handler was removed, the extra parsed bytes/HttpContent/etc were dropped and we picked up on the last packet boundary NOT consumed by the handlers.

This code attempts to give us a better way to back out of the transformation code and send the original packets exactly as we received them to the target.  There are other reasons that we might need to back off of the netty pipeline too - like if the request was completely garbled (it isn't a fair performance test if there's a DOS attack & we filter the attack away).

To support all of those cases, the HttpJsonTransformer now preserves all of the original incoming bytes for its lifetime.  If processing yields a NoContentException or we haven't finished running all of the data through the pipeline after we've flushed it, the pipeline is abandoned and we send the packets directly using the new `redriveWithoutTransformation` method.  To show the status of message processing, that is now included in the AggregateRawResponse - though we don't do much with it yet.

### Issues Resolved
No Jira

### Testing
Manual testing and new unit tests

### Check List
  - [x] All tests pass, including unit test

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Utils.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Utils.java
@@ -1,0 +1,19 @@
+package org.opensearch.migrations.replay;
+
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+public class Utils {
+    public static <A, B> Collector<A, ?, B>
+    foldLeft(final B seedValue, final BiFunction<? super B, ? super A, ? extends B> f) {
+        return Collectors.collectingAndThen(
+                Collectors.reducing(
+                        Function.<B>identity(),
+                        a -> b -> f.apply(b, a),
+                        Function::andThen),
+                finisherArg -> finisherArg.apply(seedValue)
+        );
+    }
+}
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformer.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformer.java
@@ -4,16 +4,40 @@ import io.netty.buffer.ByteBuf;
 import io.netty.channel.embedded.EmbeddedChannel;
 import lombok.extern.slf4j.Slf4j;
 import org.opensearch.migrations.replay.AggregatedRawResponse;
+import org.opensearch.migrations.replay.Utils;
 import org.opensearch.migrations.replay.datahandlers.IPacketToHttpHandler;
 import org.opensearch.migrations.transform.JsonTransformer;
 
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.CompletionException;
+import java.util.function.Function;
 
+/**
+ * This class implements a packet consuming interface by using an EmbeddedChannel to write individual
+ * packets through handlers that will parse the request's HTTP headers, determine what may need to
+ * be done with the message, and continue to parse/handle/serialize the contents according to the
+ * transformation and the headers present (such as for gzipped or chunked encodings).
+ *
+ * There will be a number of reasons that we need to reuse the source captured packets - both today
+ * (for the output to the comparator) and in the future (for retrying transient network errors or
+ * transformation errors).  With that in mind, the HttpJsonTransformer now keeps track of all of
+ * the ByteBufs passed into it and can redrive them through the underlying network packet handler.
+ * Cases where that would happen with this edit are where the payload wasn't being modified, nor
+ * were the headers changed.  In those cases, the entire pipeline gets removed and the "baseline"
+ * ones never get activated.  This class detects that the baseline (NettySendByteBufsToPacketHandlerHandler)
+ * never processed data and will re-run the messages that it already had accumulated.
+ *
+ * A bigger question will be how to deal with network errors, where some packets have been sent to
+ * the network.  If a partial response comes back, should that be reported to the user?  What if the
+ * error was due to transformation, how would we be able to tell?
+ */
 @Slf4j
 public class HttpJsonTransformer implements IPacketToHttpHandler {
+    private final RequestPipelineOrchestrator pipelineOrchestrator;
     private final EmbeddedChannel channel;
     /**
      * Roughly try to keep track of how big each data chunk was that came into the transformer.  These values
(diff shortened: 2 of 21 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestHandler.java
```diff
@@ -28,59 +27,79 @@ public class NettyDecodedHttpRequestHandler extends ChannelInboundHandlerAdapter
...
     }
 
+    private void handlePayloadNeutralTransformation(ChannelHandlerContext ctx,
+                                                    HttpRequest request,
+                                                    HttpJsonMessageWithFaultablePayload httpJsonMessage,
+                                                    RequestPipelineOrchestrator pipelineOrchestrator)
+            throws PayloadNotLoadedException
+    {
+        var pipeline = ctx.pipeline();
+        transformer.transformJson(httpJsonMessage);
```
lewijacn, line 69: What is this command actually doing as we don't directly check a response?

lewijacn: We do the transformation but may later completely disregard it, is this correct?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Utils.java
```diff
@@ -0,0 +1,19 @@
+package org.opensearch.migrations.replay;
+
+import java.util.function.BiFunction;
+import java.util.function.Function;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+public class Utils {
+    public static <A, B> Collector<A, ?, B>
+    foldLeft(final B seedValue, final BiFunction<? super B, ? super A, ? extends B> f) {
```
lewijacn, line 10: I think this would benefit from comment

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformer.java
```diff
@@ -4,57 +4,107 @@
...
+                        } else {
+                            throw new CompletionException(t);
+                        }
+                    } else {
+                        return CompletableFuture.completedFuture(v);
+                    }
+                }).thenCompose(Function.identity());
+    }
+
+    private CompletableFuture<AggregatedRawResponse> redriveWithoutTransformation(IPacketToHttpHandler packetConsumer) {
```
lewijacn, line 103: Is this just simply sending the bytebufs directly to NettyPacketToHttpHandler?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestHandler.java
```diff
@@ -28,59 +27,79 @@ public class NettyDecodedHttpRequestHandler extends ChannelInboundHandlerAdapter
...
+                pipeline.removeFirst();
+            }
+        } else if (headerFieldIsIdentical("content-encoding", request, httpJsonMessage) &&
+                headerFieldIsIdentical("transfer-encoding", request, httpJsonMessage)) {
+            log.info("There were changes to the headers that will be reformatted through this pipeline, " +
+                    "but the content (payload) doesn't need to be transformed.  Content Handlers are not " +
+                    "being added to the pipeline");
+            pipelineOrchestrator.addBaselineHandlers(pipeline);
+            ctx.fireChannelRead(httpJsonMessage);
+            pipelineOrchestrator.removeThisAndPreviousHandlers(pipeline, this);
```
lewijacn, line 83: Not relevant for right now, but would like to discuss at some point what the solution looks like if we avoid removing handlers and have some sort of skipping mechanism for handlers that we don't need or if there is any way to detect the needed handlers sooner. I think this story is a little easier to tell if the pieces aren't dynamically changing

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/RequestPipelineOrchestrator.java
```diff
@@ -77,30 +78,41 @@ void addContentParsingHandlers(ChannelPipeline pipeline, JsonTransformer transfo
             log.warn("Adding JSON handlers to pipeline");
             //  IN: Netty HttpRequest(2) + HttpJsonMessage(1) with headers + HttpContent(2) blocks
             // OUT: Netty HttpRequest(2) + HttpJsonMessage(2) with headers AND payload
+            addLoggingHandler(pipeline, "B");
             pipeline.addLast(new NettyJsonBodyAccumulateHandler());
             //  IN: Netty HttpRequest(2) + HttpJsonMessage(2) with headers AND payload
```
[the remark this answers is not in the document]
gregschohn: It isn't today, out of the box, because for C & D, the message flowing through is just an HttpJsonMessage, which is going to look the same at each stage because it isn't dumping the body.  We could make some further edits to elaborate.  That would be useful, but a bit more involved.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Utils.java
gregschohn: Kartik had asked for this too.  Here's what I put on it...
    /**
     * See https://en.wikipedia.org/wiki/Fold_(higher-order_function)
     */
I'm not sure that I can do a better job describing it than the image does.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformer.java
```diff
@@ -4,57 +4,107 @@
...
-    private NettySendByteBufsToPacketHandlerHandler getEndOfConsumptionHandler() {
-        var last = channel.pipeline().last();
-        return (NettySendByteBufsToPacketHandlerHandler) last;
+    private NettySendByteBufsToPacketHandlerHandler getOffloadingHandler() {
+        return Optional.ofNullable(channel).map(c -> (NettySendByteBufsToPacketHandlerHandler) c.pipeline()
+                        .get(RequestPipelineOrchestrator.OFFLOADING_HANDLER_NAME))
+                .orElse(null);
     }
 
     public CompletableFuture<Void> consumeBytes(ByteBuf nextRequestPacket) {
```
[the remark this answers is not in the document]
gregschohn: I think that Kartik found that one too & you can see that it's in PR-182 now.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformer.java
gregschohn: Yes, provided that the transformedPacketReceiver was a NettyPacketToHttpHandler.   It also makes sure that it does them so that packets won't be committed out of order.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestHandler.java
gregschohn: This was updated in PR-182 to reset the httpJsonMessage.
In _some_ cases we'll disregard it.  Line 82 will send it along as is, before a reset though.  That worked before because the transformations today won't replace the top-level object.  The signature indicates that that could happen, so you (& kartg) are right that it should be updated (and it is in a future commit)

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestHandler.java
gregschohn: This part of the story is easier to tell - and even writing the logic to just pass stuff through isn't terrible.  Where we'll suffer most might be in logging.  You'd see way more flowing through the handlers with logging turned up, even though the data was static.

PR #191, opened 2023-06-06
Update CDK setup to incorporate latest changes

lewijacn:
### Description
This change gets our CDK setup up to speed with the latest changes on main and should get us pretty close (if not all the way there) to tagging a demo-able setup.

### Issues Resolved
None

### Testing
Manual deployment to AWS verified

### Check List


TrafficCapture/trafficReplayer/docker/Dockerfile
gregschohn, line 1: Why have you moved these files into project directories?  The docker file doesn't have any bearing on the local source code, so it will be highly confusing at best for a developer to make some changes, build the docker file, and not get the changes into the containers that they're spinning up.

TrafficCapture/kafkaPrinter/docker/Dockerfile
gregschohn, line 1: This is only here for historical purposes, right?  This is something that you've used to deploy a Kafka->stdout->netcat->REPLAYER solution.  We should soon recycle this away and try leveraging the docker image built by the dockerSolution project (biggest advantage is that we don't need to build w/in the container).

gregschohn (commented):
From the description, I'm not sure what I would do to deploy the artifacts that are within this repo.  Could you please add a README in the deployment directory that shows some of the commands.  You can use pseudo-script with console interactions that are involved represented with placeholder comment-descriptions.

TrafficCapture/kafkaPrinter/docker/Dockerfile
lewijacn: Agree this is a temporary state, see my other comment

lewijacn:
> From the description, I'm not sure what I would do to deploy the artifacts that are within this repo. Could you please add a README in the deployment directory that shows some of the commands. You can use pseudo-script with console interactions that are involved represented with placeholder comment-descriptions.

Added some details in README

TrafficCapture/trafficReplayer/docker/Dockerfile
gregschohn: The dockerfile is in the same directory as the source code.  However, if I change the source code & then rebuild the docker file, my changes won't be there because we're just pulling from the repo.  That's not what I would expect to happen as a developer.

lewijacn: This should be modified to not use git but pull from local. Correct me if I'm wrong

PR #192, opened 2023-06-06
MIGRATIONS-1080 - Stop directly using an HttpObjectDecoder

gregschohn:
### Description
* **_Category_**: Bug fix
* _**Why these changes are required**_?: To fix bugs when packet boundaries may be anywhere within an HTTP request.
* **_What is the old behavior before changes and new behavior after changes?_**: The TrafficStream recording would be incomplete and/or we would indefinitely hang on requests to ConditionallyReliableLoggingHttpRequestHandler that should have been stopped for the last bytes before flushing.  The old code would get confused and the state model would break.

This change replaces direct usage an HttpObjectDecoder for an HttpRequestDecoder.

ByteBufs are now passed through an EmbeddedChannel that has an HttpRequestDecoder (SImpleHttpRequestDecoder to be specific, which is a bit lighter on memory retention) followed by a SimpleDecodedHttpRequestHandler. When the second handler gets notification that the HttpRequest and LastHttpContent objects are received, that triggers different callbacks. This allows the HttpRequestDecoder object to do the intra-line buffering that the Decoder by itself doesn't do. We're still able to use the decoder in a lightweight way by overriding createMessage() to use a new PassThroughHeaders as opposed to DefaultHttpHeaders(). Both of the tests within ConditionallyReliableLoggingHttpRequestHandlerTest now pass.

### Issues Resolved
MIGRATIONS-1080

### Testing
Unit tests pass.  :dockerSolution:composeUp shows tuples being output by the comparator w/ responses (spot-checked).

### Check List

### DIFF
--- a/TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
+++ b/TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
@@ -6,6 +6,7 @@ import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.embedded.EmbeddedChannel;
 import io.netty.handler.codec.http.DefaultHttpRequest;
 import io.netty.handler.codec.http.HttpMessage;
+import io.netty.handler.codec.http.HttpMessageDecoderResult;
 import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.codec.http.HttpRequest;
 import io.netty.handler.codec.http.HttpRequestDecoder;
@@ -129,12 +130,20 @@ public class LoggingHttpRequestHandler extends ChannelInboundHandlerAdapter {
 
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        var timestamp = Instant.now();
         var bb = (ByteBuf) msg;
-        trafficOffloader.addReadEvent(Instant.now(), bb);
+        trafficOffloader.addReadEvent(timestamp, bb);
 
         var httpProcessedState = parseHttpMessageParts(bb);
         if (httpProcessedState == HttpCaptureSerializerUtil.HttpProcessedState.FULL_MESSAGE) {
             var httpRequest = getHandlerThatHoldsParsedHttpRequest().resetCurrentRequest();
+            var decoderResultLoose = httpRequest.decoderResult();
+            if (decoderResultLoose instanceof HttpMessageDecoderResult) {
+                var decoderResult = (HttpMessageDecoderResult) decoderResultLoose;
+                trafficOffloader.addEndOfFirstLineIndicator(decoderResult.initialLineLength());
+                trafficOffloader.addEndOfHeadersIndicator(decoderResult.headerSize());
+            }
+            trafficOffloader.commitEndOfHttpMessageIndicator(timestamp);
             channelFinishedReadingAnHttpMessage(ctx, msg, httpRequest);
         } else {
             super.channelRead(ctx, msg);
--- a/TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
+++ b/TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
@@ -17,23 +17,22 @@ import org.opensearch.migrations.trafficcapture.IChannelConnectionCaptureSeriali
 
 import java.io.IOException;
 import java.time.Instant;
-import java.util.ArrayList;
 import java.util.List;
 
 @Slf4j
 public class LoggingHttpRequestHandler extends ChannelInboundHandlerAdapter {
     static class SimpleHttpRequestDecoder extends HttpRequestDecoder {
+        /**
+         * Override this so that the HttpHeaders object can be a cheaper one.  PassThruHeaders
+         * only stores a handful of headers that are required for parsing the payload portion
+         * of an HTTP Message.
+         */
         @Override
         public HttpMessage createMessage(String[] initialLine) throws Exception {
             return new DefaultHttpRequest(HttpVersion.valueOf(initialLine[2]),
                     HttpMethod.valueOf(initialLine[0]), initialLine[1]
                     , new PassThruHttpHeaders()
             );
-        }
-
-        @Override
-        public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
-            super.channelRead(ctx, msg);
         }
     }
 
@@ -52,6 +51,7 @@ public class LoggingHttpRequestHandler extends ChannelInboundHandlerAdapter {
         }
 
         public HttpRequest resetCurrentRequest() {
+            isDone = false;
             var old = currentRequest;
             currentRequest = null;
             return old;
@@ -61,7 +61,7 @@ public class LoggingHttpRequestHandler extends ChannelInboundHandlerAdapter {
     protected final IChannelConnectionCaptureSerializer trafficOffloader;
 
     protected final EmbeddedChannel httpDecoderChannel;
-    SimpleHttpRequestDecoder requestDecoder;
+    protected final SimpleHttpRequestDecoder requestDecoder;
 
 
     public LoggingHttpRequestHandler(IChannelConnectionCaptureSerializer trafficOffloader) {
@@ -73,19 +73,7 @@ public class LoggingHttpRequestHandler extends ChannelInboundHandlerAdapter {
         );
     }
 
-
-
-    protected HttpCaptureSerializerUtil.HttpProcessedState
-    onHttpObjectsDecoded(List<Object> parsedMsgs) throws IOException {
-        parsedMsgs.stream()
-                .filter(o -> o instanceof DefaultHttpRequest)
-                .map(o -> (DefaultHttpRequest) o)
-                .findAny()
-                .ifPresent(req -> getHttpRequestReceiverHandler().currentRequest = req);
-        return HttpCaptureSerializerUtil.addRelevantHttpMessageIndicatorEvents(trafficOffloader, parsedMsgs);
-    }
-
-    private HttpCaptureSerializerUtil.HttpProcessedState parseHttpMessageParts(ChannelHandlerContext ctx, ByteBuf msg) throws Exception {
+    private HttpCaptureSerializerUtil.HttpProcessedState parseHttpMessageParts(ByteBuf msg) throws Exception {
         var bb = msg;
         // Preserve ownership of this ByteBuf because the HttpRequestDecoder will take ownership of the
         // ByteBuf, releasing it once the data has been copied into its cumulation buffer.  However,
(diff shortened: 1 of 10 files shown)


TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -19,75 +23,85 @@
...
-         *
-         * @param ctx    the {@link ChannelHandlerContext} which this {@link ByteToMessageDecoder} belongs to
-         * @param buffer the {@link ByteBuf} from which to read data
-         * @param out    the {@link List} to which decoded messages should be added
-         * @throws Exception
-         */
         @Override
-        public void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out) throws Exception {
-            super.decode(ctx, buffer, out);
+        public HttpMessage createMessage(String[] initialLine) throws Exception {
```
lewijacn, line 27: nit: I know this is existing code but may be helpful to callout the benefit of using PassThruHttpHeaders here

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -19,75 +23,85 @@
...
+            currentRequest = null;
+            return old;
         }
     }
 
     protected final IChannelConnectionCaptureSerializer trafficOffloader;
 
-    protected final SimpleHttpRequestDecoder httpDecoder;
+    protected final EmbeddedChannel httpDecoderChannel;
+    SimpleHttpRequestDecoder requestDecoder;
```
lewijacn, line 64: Can this be `protected final` as well ?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -19,75 +23,85 @@
...
-        httpDecoder = new SimpleHttpRequestDecoder();
-        currentHttpRequest = null;
+        requestDecoder = new SimpleHttpRequestDecoder(); // as a field for easier debugging
+        httpDecoderChannel = new EmbeddedChannel(
+                requestDecoder,
+                new SimpleDecodedHttpRequestHandler()
+        );
     }
 
+
```
lewijacn, line 76: nit: extra lines here

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -157,6 +171,7 @@ public void channelWritabilityChanged(ChannelHandlerContext ctx) throws Exceptio
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         trafficOffloader.addExceptionCaughtEvent(Instant.now(), cause);
+        httpDecoderChannel.close();
```
lewijacn, line 174: Do we need to close on normal termination as well?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -19,75 +23,85 @@
...
-            httpDecoder.decode(ctx, bb, parsedMsgs);
-            rval = onHttpObjectsDecoded(parsedMsgs);
-            if (rval == HttpCaptureSerializerUtil.HttpProcessedState.FULL_MESSAGE || // got what we needed
-                    lastIdx == bb.readerIndex()) { // no progress
-                break;
-            }
-            parsedMsgs.clear();
-        }
-        log.trace("Resetting index");
+        httpDecoderChannel.writeInbound(bb);
```
lewijacn, line 96: +1 this method seems easier to follow now

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -19,75 +23,85 @@
...
         bb.resetReaderIndex();
-        return rval;
+        return getHttpRequestReceiverHandler().isDone ?
+                HttpCaptureSerializerUtil.HttpProcessedState.FULL_MESSAGE :
+                HttpCaptureSerializerUtil.HttpProcessedState.ONGOING;
+    }
+
+    private SimpleDecodedHttpRequestHandler getHttpRequestReceiverHandler() {
```
lewijacn, line 103: nit: the naming being different from type and method name threw me off a little

lewijacn (commented):
Overall this seems a bit easier to follow now. One of my main open questions is do we need this logic on the HttpResponseHandler as well? Left some minor comments as well

TrafficCapture/trafficCaptureProxyServer/src/test/resources/log4j2.properties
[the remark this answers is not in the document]
gregschohn: Thanks... Unintentional checkin. (the following are notes mostly for myself or anybody else that wants to think about how logging for unit tests should work).

However, in tweaking some code in response to later comments, I went back and forth between whether unit tests should output to stderr and stdout.  I'm not sure that stdout is superior from a developer perspective, especially since our tools (though not this one) output a lot to stdout for machine consumption.  

Either way, gradle won't show the output of stderr or stdout unless you have `test.testLogging.showStandardStreams=true` or run with gradle test with the `-i` option.  In either of those cases BOTH stdout and stderr will render - so it feels like keeping this the same as the main jar is the easiest way to go.  It also lets us test logging configurations.  I've also tested that the tests will pickup the log configuration in the project's jar file if there isn't one in the test jar artifact.

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
gregschohn: It shouldn't hurt, provided that we can identify what normal termination is - and I'm not 100% sure yet what exactly those cases are.  This close() is here not for resource management but because the connection and this handler should be in a terminal, non-recoverable state.  If we prematurely close the channel though and then have more data to write, that would be bad.  Given that the EmbeddedChannel doesn't hold any disposable resources, I'm fine letting the GC take the channel, pipeline, & handlers all out when this object gets finalized.  

I plan to address stress testing through many of the corner cases in the coming weeks, but it's hard to write representative tests without seeing some real traffic flow through the system first.

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
gregschohn: Done.  Thanks

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -73,19 +73,7 @@ public LoggingHttpRequestHandler(IChannelConnectionCaptureSerializer trafficOffl
...
-
-
-    protected HttpCaptureSerializerUtil.HttpProcessedState
-    onHttpObjectsDecoded(List<Object> parsedMsgs) throws IOException {
-        parsedMsgs.stream()
-                .filter(o -> o instanceof DefaultHttpRequest)
-                .map(o -> (DefaultHttpRequest) o)
-                .findAny()
-                .ifPresent(req -> getHttpRequestReceiverHandler().currentRequest = req);
-        return HttpCaptureSerializerUtil.addRelevantHttpMessageIndicatorEvents(trafficOffloader, parsedMsgs);
```
lewijacn, line 68: Don't we still need to call `addRelevantHttpMessageIndicatorEvents` ?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
lewijacn: Slightly, my inclination is to use `getSimpleDecodedHttpRequestHandler` but that might not be clear to others so I'm fine either way

TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/InMemoryConnectionCaptureFactory.java
```diff
@@ -48,6 +50,7 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
             CompletableFuture cf = closeHandler(codedStreamToByteBufferMap.get(codedOutputStream));
             codedStreamToByteBufferMap.remove(codedOutputStream);
             singleAggregateCfRef[0] = singleAggregateCfRef[0].isDone() ? cf : CompletableFuture.allOf(singleAggregateCfRef[0], cf);
+            cf.whenComplete((v,t)->onCaptureClosedCallback.run());
```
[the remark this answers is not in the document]
gregschohn: Quick answer, what was specified in the constructor.  In our case for the unit test, it decrements the countdownLatch upon each completion.  Notice that this class is just for test purposes (it was a bit weird to me to chain yet one more listener to bind this to the sending context).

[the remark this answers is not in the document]
lewijacn: Okay makes sense now, I didn't pick up this was being passed into the constructor

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
gregschohn: Yep - next version does that.  It will include an additional check in a unit test to make sure that the EOM is there for the requests.

gregschohn (commented):
To answer the question about "do we need this for responses".  See [MIGRATIONS-1114](https://opensearch.atlassian.net/jira/software/c/projects/MIGRATIONS/boards/1?modal=detail&selectedIssue=MIGRATIONS-1114).  Trying to do any kind of HTTP parsing and EOM tracking for responses is futile anyway.  I'm going to be undertaking 1114 to just put that responsibility in the replayer.

lewijacn (approved):
Thanks for updates

PR #193, opened 2023-06-07
Migrations-1112 - Specify supported Java version

okhasawn:
### Description
This PR specifies and locks the supported Java version to 11.

### Issues Resolved
[Migrations-1112](https://opensearch.atlassian.net/browse/MIGRATIONS-1112)

### Check List


TrafficCapture/dockerSolution/build.gradle
```diff
@@ -117,3 +117,9 @@ task buildDockerImages {
 }
 
 tasks.getByName('composeUp').dependsOn(tasks.getByName('buildDockerImages'))
+
+java {
+    toolchain {
+        languageVersion = JavaLanguageVersion.of(11) // Locks supported Java version to 11
```
gregschohn, line 123: This probably won't do much since the dockerSolution project doesn't build Java code. This would need to be added for each project. Is there a way to do this at a global level, for all the Gradle projects with something like Gradle settings?

okhasawn: You're right, this was failing when I run `./gradlew dockerSolution:build` if I had a different version specified, but I guess it's a habit of mine that I run build before running `./gradlew dockerSolution:composeUp` or down here.
Thanks Greg, I'll address this.

TrafficCapture/dockerSolution/build.gradle
```diff
@@ -116,4 +116,15 @@ task buildDockerImages {
     //dependsOn buildDockerImage_
 }
 
-tasks.getByName('composeUp').dependsOn(tasks.getByName('buildDockerImages'))
+tasks.getByName('composeUp').dependsOn(tasks.getByName('buildDockerImages').dependsOn('configureJavaVersion'))
+
+tasks.register('configureJavaVersion') {
```
gregschohn, line 121: You're not doing any configuration here - you're checking the configuration.  You really should be doing the configuration though, so the name is good, the implementation should do more (see the overall note, this isn't the right place to do it)

gregschohn (commented):
Take a look at https://github.com/opensearch-project/opensearch-migrations/pull/195.  If you want to approve and merge that, that works for me.  
The current solution will allow other projects to be compiled with newer versions.  Users might never know that they're using features that won't work for the containers that we're deploying to if they never run that single dockerSolution targets.  Worse, they could do builds with one version, then come in and change the version that they use for dockerSolution and end up with a frankenbuild with different jar versions.

TrafficCapture/dockerSolution/build.gradle
okhasawn: I see your point above about how if a user doesn't run the solution, they can still use a different version and can see how my solution is not valid since even though it only permits java 11 with the docker solution as a _whole_, separate tools can still be allowed to be built. I was somehow still focused on the docker solution and thought that this would be valid since it will technically cover each tool since it's hit before attempting to build them (in docker).
I think seeing #195 made me see the point clearly. Thanks for creating that! I'll go ahead and close this PR.

PR #203, opened 2023-06-15
Add Support for Proxy to connect to HTTPS Servers

lewijacn:
### Description
This change, similar to what has been done in the Traffic-Replayer [here](https://github.com/opensearch-project/opensearch-migrations/commit/7bb1eef8ea1f2a438f609e8aa1c03fb5c57ade2e#diff-b8c531f1b8f2cf3948ca4249f59868e7181dbc734e5a9f531b1ea32ae75f7641), allows the Logging Proxy to connect to a backside HTTPS server by creating an SslContext and allowing insecure connections if specified.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1105

### Testing
Docker compose local testing

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
@@ -78,7 +78,7 @@ public class Main {
         @Parameter(required = false,
             names = {"--insecureDestination"},
             arity = 0,
-            description = "Do not check the backside server's certificate")
+            description = "Do not check the destination server's certificate")
         boolean allowInsecureConnectionsToBackside;
         @Parameter(required = true,
                 names = {"--destinationUri"},
@@ -174,6 +174,8 @@ public class Main {
         }
     }
 
+    // Utility method for converting uri string to an actual URI object. Similar logic is placed in the trafficReplayer
+    // module: TrafficReplayer.java
     private static URI convertStringToUri(String uriString) {
         URI serverUri;
         try {
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
@@ -4,6 +4,9 @@ import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.ParameterException;
 import com.google.protobuf.CodedOutputStream;
+import io.netty.handler.ssl.SslContext;
+import io.netty.handler.ssl.SslContextBuilder;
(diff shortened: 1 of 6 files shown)


TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -73,15 +76,15 @@ static class Parameters {
                 description = "The maximum number of bytes that will be written to a single TrafficStream object.")
         int maximumTrafficStreamSize = 1024*1024;
         @Parameter(required = false,
-                names = {"--destinationHost"},
-                arity = 1,
-                description = "Hostname of the server that the proxy is capturing traffic for.")
-        String backsideHostname = "localhost";
+            names = {"--insecureDestination"},
+            arity = 0,
+            description = "Do not check the backside server's certificate")
```
gregschohn, line 81: s/backside/destination/.  You're introducing a new term (backside) when you've already defined it as destination.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -171,11 +174,46 @@ private static IConnectionCaptureFactory getConnectionCaptureFactory(Parameters
         }
     }
 
+    private static URI convertStringToUri(String uriString) {
```
gregschohn, line 177: It's probably worth at least calling out in a comment how this is similar to the replayer.

gregschohn (approved):
Approving this, but I'd like to verify that you've verified that the proxy can make HTTPs connections w/ the docker conpose solution.  The other comments are minor and can be addressed at your own pace.

lewijacn:
> 

This does work with using HTTPS and the docker solution with a couple caveats that may have already been expected. Mainly I wanted to wait until we add something for point 2. before enabling to avoid disruption for others

1.  The elastic search with searchguard Dockerfile needs to be modified to enable searchguard and use `enableTlsConfig.sh` instead of `disableTlsConfig.sh`
2. Auth header needs to be provided in call to proxy and will be passed to destination call. To make this easier for our local docker solution we may should open another task to allow passing in an auth header parameter similar to what we have done for the Traffic Replayer
3. Edit the docker compose file:

Following command for capture proxy:
 ```
/runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.proxyserver.Main  --kafkaConnection kafka:9092 --destinationUri  https://elasticsearch:9200 --insecureDestination --listenPort 9200 --sslConfigFile /usr/share/elasticsearch/config/proxy_tls.yml
```

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
lewijacn: Agreed adjusted

PR #211, opened 2023-06-23
Github Workflow for Gradle Build and Test

okhasawn:
### Description
This PR adds a github action for running `./gradlew build` and `./gradlew test` on the tools within the TrafficCapture directory. This will also add generating and uploading a coverage report for each tool (when applicable; has any tests)

This will also disable linting and pytest workflows for non-python directories.

### Issues Resolved
[Migrations-1152](https://opensearch.atlassian.net/browse/MIGRATIONS-1152)

### Testing
https://github.com/okhasawn/opensearch-migrations/actions/runs/5358402373/jobs/9720638919

### Check List

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -26,6 +26,11 @@ jobs:
       - name: Upload to Codecov
         uses: codecov/codecov-action@v3
         with:
-          file: ./TrafficCapture/*/build/reports/jacoco/test/jacocoTestReport.xml
+          files: TrafficCapture/captureKafkaOffloader/build/reports/jacoco/test/jacocoTestReport.xml,
+            TrafficCapture/captureOffloader/build/reports/jacoco/test/jacocoTestReport.xml,
+            TrafficCapture/KafkaPrinter/build/reports/jacoco/test/jacocoTestReport.xml,
+            TrafficCapture/nettyWireLogging/build/reports/jacoco/test/jacocoTestReport.xml,
+            TrafficCapture/trafficCaptureProxyServer/build/reports/jacoco/test/jacocoTestReport.xml,
+            TrafficCapture/trafficReplayer/build/reports/jacoco/test/jacocoTestReport.xml
           flags: unittests
-          fail_ci_if_error: true
+          fail_ci_if_error: false
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -24,9 +24,8 @@ jobs:
         working-directory: TrafficCapture
 
       - name: Upload to Codecov
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v3
         with:
           file: ./TrafficCapture/*/build/reports/jacoco/test/jacocoTestReport.xml
           flags: unittests
-          name: codecov-umbrella
           fail_ci_if_error: true
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -1,0 +1,32 @@
+name: Gradle Build and Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


.github/workflows/gradle.yml
```diff
@@ -0,0 +1,36 @@
+name: Gradle Build and Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
```
lewijacn, line 6: Could we have this name be more descriptive, something like `gradle-build-and-test`

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up JDK 11
+        uses: actions/setup-java@v2
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+
+      - name: Run Gradle Build
+        run: ./gradlew build
```
lewijacn, line 19: This will run tests as well. Using this command `./gradlew build -x test` will not run tests until the next step which is what I think we want

TrafficCapture/build.gradle
```diff
@@ -9,4 +10,15 @@ allprojects {
...
+
+    jacocoTestReport {
+        reports {
+            xml.required = true
+            xml.destination file("${buildDir}/reports/jacoco/test/jacocoTestReport.xml")
+        }
+    }
+}
+
+test {
```
lewijacn, line 22: What are lines 22-24 doing? I didn't see any usage of this

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,36 @@
...
+        working-directory: TrafficCapture
+
+      - name: Run Tests with Coverage
+        run: ./gradlew test jacocoTestReport
+        working-directory: TrafficCapture
+
+      - name: Upload to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          files: TrafficCapture/captureKafkaOffloader/build/reports/jacoco/test/jacocoTestReport.xml,
```
lewijacn, line 29: Do you know if its possible to use any pattern matching/regex here to have this as a single line so that it's easier to maintain in future

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,36 @@
...
+        with:
+          java-version: '11'
+          distribution: 'adopt'
+
+      - name: Run Gradle Build
+        run: ./gradlew build
+        working-directory: TrafficCapture
+
+      - name: Run Tests with Coverage
+        run: ./gradlew test jacocoTestReport
```
lewijacn, line 23: To give a little more detail we should probably add `--info`

TrafficCapture/build.gradle
okhasawn: They are causing the `jacocoTestReport` task to run at the end of the test task, which means that's when the coverage will be available.

.github/workflows/gradle.yml
okhasawn: That makes sense, thanks!

.github/workflows/gradle.yml
okhasawn: You're right, no need to attempt running the tests twice, and this'll provide more info about the tests during the test step not build. I'll update.

.github/workflows/gradle.yml
okhasawn: I know that it is possible to use wildcards, but not in the way I had hoped.
e.g `TrafficCapture/*/build/reports/jacoco/test/jacocoTestReport.xml` will NOT lead including all the files in our case.
One of the cases that would work is if, for example we had jacocoTestReport1, jacocoTestReport2, jacocoTestReport3 - then all these would be included if `full-path-here/jacocoTestReport*.xml` was used instead of each file respectively.

lewijacn (approved):

PR #221, opened 2023-06-29
[Index configuration tool] Support for disable_authentication key in Data Prepper YAML

kartg:
### Description
This change adds logic to read the "disable_authentication" key in the Data Prepper source and sink configuration. This flag was recently added to Data Prepper and removes the need to specify dummy username and password values in order to connect to an unauthenticated OpenSearch source/sink.
* Category: Enhancement

### Issues Resolved
N/A

### Testing
```
..........................
Ran 26 tests in 0.520s

OK
main.py                  92      0   100%
TOTAL                   142      2    99%
```

### Check List

### DIFF
--- a/index_configuration_tool/main.py
+++ b/index_configuration_tool/main.py
@@ -10,6 +10,7 @@ SUPPORTED_ENDPOINTS = ["opensearch", "elasticsearch"]
 SOURCE_KEY = "source"
 SINK_KEY = "sink"
 HOSTS_KEY = "hosts"
+DISABLE_AUTH_KEY = "disable_authentication"
 USER_KEY = "username"
 PWD_KEY = "password"
 INSECURE_KEY = "insecure"
@@ -28,7 +29,7 @@ def is_insecure(config: dict) -> bool:
 
 # TODO Only supports basic auth for now
 def get_auth(input_data: dict) -> Optional[tuple]:
-    if USER_KEY in input_data and PWD_KEY in input_data:
+    if not input_data.get(DISABLE_AUTH_KEY, False) and USER_KEY in input_data and PWD_KEY in input_data:
         return input_data[USER_KEY], input_data[PWD_KEY]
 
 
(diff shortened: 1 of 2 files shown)


index_configuration_tool/tests/test_main.py
```diff
@@ -20,13 +20,16 @@
...
+                         password: Optional[str] = None,
+                         disable_auth: Optional[bool] = None) -> dict:
     config = dict()
     config["hosts"] = host_list
     if user:
         config["username"] = user
     if password:
         config["password"] = password
+    if disable_auth is not None:
+        config["disable_authentication"] = disable_auth
```
okhasawn, line 32: I was gonna comment if this should be `is not False` but then I got the idea that you're only checking if it has a value at all first, then assigning that value (no matter what it is) if it exists to the `disable_authentication` config attribute (I might be using the wrong term here). Anyways, I somehow really like this if condition :)

okhasawn (approved):

index_configuration_tool/main.py
```diff
@@ -77,10 +78,13 @@ def validate_plugin_config(config: dict, key: str):
...
-        raise ValueError("Invalid auth configuration (no password for user) for endpoint: " + supported_endpoint[0])
-    elif PWD_KEY in plugin_config and USER_KEY not in plugin_config:
-        raise ValueError("Invalid auth configuration (Password without user) for endpoint: " + supported_endpoint[0])
+    if not plugin_config.get(DISABLE_AUTH_KEY, False):
+        if USER_KEY in plugin_config and PWD_KEY not in plugin_config:
+            raise ValueError("Invalid auth configuration (no password for user) for endpoint: " +
+                             supported_endpoint[0])
+        elif PWD_KEY in plugin_config and USER_KEY not in plugin_config:
+            raise ValueError("Invalid auth configuration (Password without user) for endpoint: " +
+                             supported_endpoint[0])
```
mikaylathompson, line 87: I want to make sure I understand the logic in this section:
1. Disable Auth key is present: any other combo is okay
2. Disable Auth key is _not_ present AND both user and pass are present: we're good, we have auth 
3. Disable Auth key is _not_ present AND only one of user and pass are present: not good, handled by this logic
4. Disable Auth key is _not_ present AND neither user nor pass is present: ??? Is this okay? Or is it handled elsewhere?

kartg: Good catch on 4 - if the disable auth key is not present, the logic in 2 and 3 should apply. I'll fix this in a follow-up PR

okhasawn: Ah, sorry I didn't catch that, Kartik. Thanks Mikayla!

kartg: https://github.com/opensearch-project/opensearch-migrations/pull/222

PR #223, opened 2023-06-30
[Index Configuration Tool] Adds support for writing to an output YAML file

kartg:
### Description
The Index Configuration Tool has been updated to now produce a Data Prepper pipeline YAML file as an output. The tool now expects two positional arguments - the input pipeline YAML file like before, and an output path for the generated YAML file. The output YAML pipeline file is a modified version of the input that adds an index inclusion configuration for the sink, specifying only those indices that were created by the tool.

This change also adds two optional command line flags:
- "--report" prints a human-readable report to the command line in addition to generating an output YAML file
- "--dryrun" skips creation of indices on the target cluster
---
* Category: Enhancement

### Issues Resolved
N/A

### Testing
```
.........................Wrote output YAML pipeline to: dummy
....
Ran 29 tests in 0.041s

OK
main.py                 112      0   100%
TOTAL                   162      2    99%
```

### Check List

### DIFF
--- a/index_configuration_tool/main.py
+++ b/index_configuration_tool/main.py
@@ -170,8 +170,7 @@ def run(args: argparse.Namespace) -> None:
     indices_to_create = diff[0]
     if indices_to_create:
         # Write output YAML
-        if not args.report:
-            write_output(dp_config, indices_to_create, args.output_file)
+        write_output(dp_config, indices_to_create, args.output_file)
         if not args.dryrun:
             index_data = dict()
             for index_name in indices_to_create:
@@ -189,8 +188,7 @@ if __name__ == '__main__':  # pragma no cover
         "of the pipeline YAML file is written. This version of the pipeline adds an index inclusion configuration " +
         "to the sink, specifying only those indices that were created by the index configuration tool.\nThis tool " +
         "can also print a report based on the indices in the source cluster, indicating which ones will be created, " +
-        "along with indices that are identical or have conflicting settings/mappings.\nIn case of the latter, no "
-        "action will be taken on the target cluster.",
+        "along with indices that are identical or have conflicting settings/mappings.",
         formatter_class=argparse.RawTextHelpFormatter
     )
     # Positional, required arguments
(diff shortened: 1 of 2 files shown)


index_configuration_tool/main.py
```diff
@@ -97,6 +100,21 @@ def validate_pipeline_config(config: dict):
...
+def write_output(yaml_data: dict, new_indices: set, output_path: str):
+    pipeline_config = next(iter(yaml_data.values()))
+    # Endpoint is a tuple of (type, config)
+    source_config = get_supported_endpoint(pipeline_config, SOURCE_KEY)[1]
+    source_indices = source_config.get(INDICES_KEY, dict())
+    included_indices = source_indices.get(INCLUDE_KEY, list())
+    for index in new_indices:
+        included_indices.append({INDEX_NAME_KEY: index})
+    source_indices[INCLUDE_KEY] = included_indices
+    source_config[INDICES_KEY] = source_indices
```
mikaylathompson, line 112: I'm not 100% on the semantics of the `get` operator in lines 107 and 108, but in other contexts dictionaries are basically always passed by reference in python, so this modifies the original in `yaml_data`. Is that intentional?

mikaylathompson: Okay, yes, in line 114 I see that it definitely is. Out of curiosity, why do you only operate on the first value in the yaml (line 104)?

index_configuration_tool/main.py
```diff
@@ -128,9 +146,9 @@ def print_report(index_differences: tuple[set, set, set]):  # pragma no cover
     print("Indices to create: " + utils.string_from_set(index_differences[0]))
 
 
-def run(config_file_path: str) -> None:
+def run(args: argparse.Namespace) -> None:
     # Parse and validate pipelines YAML file
-    with open(config_file_path, 'r') as pipeline_file:
+    with open(args.config_file_path, 'r') as pipeline_file:
```
mikaylathompson, line 151: worth checking that this value exists/throwing an intentional exception here? Or does argparse handle it for you?

mikaylathompson (approved):

index_configuration_tool/main.py
kartg: `config_file_path` is a positional argument, so `argparse` will ensure that it is present

index_configuration_tool/main.py
kartg: > Out of curiosity, why do you only operate on the first value in the yaml (line 104)?

The top level object in the Data Prepper pipeline YAML is a human-readable pipeline name (as the key) followed by the actual pipeline configuration (as the value). Since this logic only aims to add an `include` configuration to the `source` plugin, line 104 drops down a level to ignore the pipeline name.

PR #231, opened 2023-07-05
Migrations-1079 - Adding Benchmark Testing Environment

okhasawn:
### Description
This is a remake of PR #161, as this one uses a whole new method of creating the containers, similar to how the dockerSolution was implemented by @gregschohn, also following Greg's recommendations on PR #161.
This PR adds the ability to create an environment where JMeter tests are run repeatedly, and its' requests are sent to the capture proxy before hitting the webserver (nginx in our case).

To get the containers up, run `./gradlew trafficCaptureProxyServerTest:composeUp` from the TrafficCapture directory.

### Issues Resolved
[Migrations-1079](https://opensearch.atlassian.net/browse/MIGRATIONS-1079)

### Check List


TrafficCapture/testingEnvironment/build.gradle
```diff
@@ -0,0 +1,125 @@
...
+                        }
+                    }
+            )
+    return digest.digest().encodeHex().toString()
+}
+
+dependencies {
+    implementation project(':trafficCaptureProxyServer')
+    testImplementation project(':trafficCaptureProxyServerTest')
+}
```
gregschohn, line 29: This can fall away by merging these changes into the ...ProxyServerTest directory

TrafficCapture/testingEnvironment/build.gradle
```diff
@@ -0,0 +1,125 @@
...
+    id 'org.opensearch.migrations.java-library-conventions'
+    id "com.avast.gradle.docker-compose" version "0.16.12"
+    id "com.bmuschko.docker-java-application" version "9.3.1"
+}
+
+import java.security.MessageDigest
+import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
+import org.apache.tools.ant.taskdefs.condition.Os
+
+def calculateDockerHash(String projectName) {
```
gregschohn, line 11: This should be refactored into a common shared class and imported by both projects

TrafficCapture/testingEnvironment/build.gradle
```diff
@@ -0,0 +1,125 @@
...
+]
+
+// Create the static docker files that aren't hosting migrations java code from this repo
+dockerFilesForExternalServices.each { projectName, dockerImageName ->
+    task("buildDockerImage_${projectName}", type: DockerBuildImage) {
+        def hash = calculateDockerHash(projectName)
+        images.add("testenv/${dockerImageName}:$hash")
+        images.add("testenv/${dockerImageName}:latest")
+        inputDir = project.file("src/main/docker/${projectName}")
+    }
```
gregschohn, line 42: This should be refactored into a common shared class and imported by both projects

TrafficCapture/testingEnvironment/build.gradle
```diff
@@ -0,0 +1,125 @@
...
+}
+
+(javaContainerServices).forEach { projectName, dockerImageName ->
+    def dockerBuildDir = "build/docker/${projectName}"
+    task "buildDockerImage_${projectName}"(type: DockerBuildImage) {
+        dependsOn "createDockerfile_${projectName}"
+        inputDir = project.file("${dockerBuildDir}")
+        images.add("testenv/${dockerImageName}:${version}")
+        images.add("testenv/${dockerImageName}:latest")
+    }
```
gregschohn, line 111: This should be refactored into a common shared class and imported by both projects

TrafficCapture/testingEnvironment/src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+      - "9201:9201"
+    command: /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.proxyserver.Main --destinationUri  http://nginx:80 --listenPort 9201 --noCapture
+    depends_on:
+      - nginx
+
+  jmeter:
+    image: 'testenv/jmeter:latest'
+    networks:
+      - testing
+    command: /bin/sh -c "while sleep 10; do /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.JMeterLoadTest -p 9201 -d captureproxy; done"
```
gregschohn, line 22: Why do you want to run this within a loop?  The JMeter configuration would be able to loop as many times as you'd like it to.  Managing multiple processes in a container is less desirable than managing just one.

TrafficCapture/testingEnvironment/src/main/docker/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
+version: '3.7'
+services:
+  nginx:
+    image: 'testenv/nginx:latest'
```
gregschohn, line 4: nit: A slightly more descriptive name than 'nginx' might be helpful for others

TrafficCapture/testingEnvironment/src/main/docker/nginx/Dockerfile
```diff
@@ -43,8 +43,12 @@ RUN NGINX_ARGS=$(nginx -V 2>&1 | sed -n -e 's/^.*arguments: //p') \
 
 FROM nginx:${NGINX_VERSION}
 
-RUN apt-get update
-RUN apt-get install -y vim
+# Switch to noninteractive mode to avoid interactive prompts during package installation
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get upgrade -y && apt-get install -y --no-install-recommends vim \
```
gregschohn, line 49: Is this to install vim?  What does the process need vim for?  Please leave a comment, especially if it's just nice to have for somebody exec-ing a shell within the container.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -49,7 +49,7 @@ static class Parameters {
         @Parameter(required = false,
                 names = {"--noCapture"},
                 arity = 0,
-                description = "Directory to store trace files in.")
+                description = "If enabled, Does NOT store trace files in.")
```
gregschohn, line 52: ... s/store trace files/capture traffic to ANY sink/

TrafficCapture/trafficCaptureProxyServerTest/src/main/java/org/opensearch/migrations/trafficcapture/JMeterLoadTest.java
```diff
@@ -23,23 +26,49 @@
...
 public class JMeterLoadTest {
 
+    static class Parameters {
+        @Parameter(required = true,
+                names = {"-p", "--port"},
+                description = "Port number")
+        int backsidePort;
+        @Parameter(required = true,
+                names = {"-d", "--domain-name"},
+                description = "Domain name")
```
gregschohn, line 36: nit: s/domain-name/server-name/.

gregschohn (commented):
The term "testingEnvironment" is going to be very overloaded.  Instead of having a separate directory/project, I'd overlay everything that you have into the trafficCaptureProxyServerTest directory, merging the build.gradle files.

TrafficCapture/trafficCaptureProxyServerTest/src/main/java/org/opensearch/migrations/trafficcapture/JMeterLoadTest.java
```diff
@@ -23,23 +26,49 @@
...
+    static class Parameters {
+        @Parameter(required = true,
+                names = {"-p", "--port"},
+                description = "Port number")
+        int backsidePort;
+        @Parameter(required = true,
+                names = {"-s", "--server-name"},
+                description = "Server name")
+        String domainName;
+    }
```
gregschohn, line 38: http vs https will eventually be a concern too.  We could also take in a URL and parse those three components.  That's what the replayer or the proxy (IIRC) does, but it's weird because the path and query string wouldn't be allowed/used.

TrafficCapture/trafficCaptureProxyServerTest/src/main/docker/docker-compose.yml
```diff
@@ -1,6 +1,30 @@
 version: '3.7'
 services:
   webserver:
-    image: 'nginx-simple:latest'
+    image: 'migrations/nginx:latest'
```
gregschohn, line 4: We should probably choose a more descriptive name than nginx.  For this project, it is fine, but the image will live in a more general migrations space.  One might presume that it's a more general purpose nginx container, but it's instead highly tailored to this testing framework.  Something like nginx-perf-test-webserver.

TrafficCapture/buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
```diff
@@ -0,0 +1,91 @@
...
+            }
+            from { project.project(":${projectName}").configurations.findByName("runtimeClasspath").files }
+            from { project.project(":${projectName}").tasks.getByName('jar') }
+            into artifactsDir
+            include "*.jar"
+            duplicatesStrategy = 'WARN'
+        }
+    }
+
+    static def createDockerfile(Project project, String projectName, Map<String, String> baseImageProjectOverrides, Map<String, String> dockerFilesForExternalServices) {
```
gregschohn, line 47: This will need some eventual refactoring - but not for this task.  There are a lot of assumptions that we're building precisely 2 kinds of dockerfiles.  We also could be better off w/ static Dockerfiles for each of these.  Again, those changes would not be for this PR.

TrafficCapture/dockerSolution/build.gradle
```diff
@@ -1,9 +1,10 @@
 plugins {
     id 'org.opensearch.migrations.java-library-conventions'
     id "com.avast.gradle.docker-compose" version "0.16.12"
-    id "com.bmuschko.docker-java-application" version "9.3.1"
+    id 'com.bmuschko.docker-remote-api'
```
gregschohn, line 4: What is this required for?  I assume that the plugin is effectively being loaded by CommonUtils now.  Is this line required?  If so, can you document why and put a version on it?

okhasawn: The id 'com.bmuschko.docker-remote-api' plugin is necessary for the _remaining_ DockerBuildImage tasks to function properly. While this plugin is now declared in the `buildSrc/build.gradle` file, it's not automatically applied to the `trafficCaptureProxyServerTest/build.gradle` file and needs to be explicitly declared. As for the version number, adding it here causes a conflict because the plugin is already on the classpath due to including it in the `buildSrc/build.gradle` file. So to sum it up, this line IS required but W/O a specified version. The version is managed within the `buildSrc/build.gradle` file.

TrafficCapture/trafficCaptureProxyServerTest/src/main/docker/docker-compose.yml
okhasawn: Oh I thought you meant the name of the container, not the image. I'll update that as you recommended.

TrafficCapture/dockerSolution/build.gradle
gregschohn: Thanks.  I understand the version.  Thanks for the reference.
For my own future understanding, the plugin isn't imported in the common code, just some java imports from it are and the remote-api plugin is a better fit to pickup the remaining other pieces unique for this grade file.

TrafficCapture/trafficCaptureProxyServerTest/build.gradle
```diff
@@ -40,38 +57,66 @@ dependencies {
     }
 }
 
+CommonConfigurations.applyCommonConfigurations(project)
 configurations {
     all {
-        resolutionStrategy.dependencySubstitution {
-            substitute module('org.apache.xmlgraphics:batik-codec') using module('org.apache.xmlgraphics:batik-all:1.15')
-        }
+        exclude group: 'org.apache.logging.log4j', module: 'log4j-slf4j-impl'
```
gregschohn, line 63: What do you need this line for?  What are you excluding it from?

TrafficCapture/trafficCaptureProxyServerTest/build.gradle
```diff
@@ -40,38 +57,66 @@ dependencies {
...
+
+javaContainerServices.each(createContainerTasks)
+
+(javaContainerServices).forEach { projectName, dockerImageName ->
+    def dockerBuildDir = "build/docker/${projectName}"
+    task "buildDockerImage_${projectName}"(type: DockerBuildImage) {
+        dependsOn "createDockerfile_${projectName}"
+        inputDir = project.file("${dockerBuildDir}")
+        images.add("migrations/${dockerImageName}:${version}")
+        images.add("migrations/${dockerImageName}:latest")
```
gregschohn, line 105: Note for myself.  This is identical to code within dockerSolution, but it seems fair to replicate it since it is short and because a couple policies are embedded within it (like the dual image tags) that could vary by project.

TrafficCapture/trafficCaptureProxyServerTest/src/main/docker/docker-compose.yml
```diff
@@ -1,6 +1,30 @@
...
+      - "9201:9201"
+    command: /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.proxyserver.Main --destinationUri  http://webserver:80 --listenPort 9201 --noCapture
+    depends_on:
+      - webserver
+
+  jmeter:
+    image: 'migrations/jmeter:latest'
+    networks:
+      - testing
+    command: /bin/sh -c "while sleep 10; do /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.JMeterLoadTest -p 9201 -s captureproxy; done"
```
gregschohn, line 22: Please kill the while part.  If you want, you can put a tail -f /dev/null in at the end to keep the container alive to make it easier to run additional commands or process files.  I'd like to know when a run has been done and have whatever time I need to process the results across the other containers (manually or automatically).

TrafficCapture/trafficCaptureProxyServerTest/src/main/java/org/opensearch/migrations/trafficcapture/JMeterLoadTest.java
gregschohn: On this note, are the tests run to test HTTPS or HTTP?  HTTPS should be the default.  I can't tell from the HttpSampler interface how to set it other than by specifying a fullUrl, which we aren't doing.

okhasawn: The HttpSampler has a setProtocol method, which by default is set to HTTP, added a parameter for that in next commit.

TrafficCapture/trafficCaptureProxyServerTest/build.gradle
okhasawn: I needed to use it to prevent classpath issues caused by multiple logging implementations, which was occuring due to mulptiple dependencies pulling in this Log4j to SLF4J bridge as a transitive dependency.
I would get this exception without it: `Exception in thread "main" java.lang.NoSuchMethodError: 'void org.apache.logging.slf4j.Log4jLoggerFactory.<init>(org.apache.logging.slf4j.Log4jMarkerFactory)'`

gregschohn: If I understand correctly, JMeter and the ProxyCapture project are both bringing in log4j components?  Do you know what version you're fixing on?

gregschohn (approved):
Approved with one comment

Issue #165, opened 2023-05-24
[Historical Data Migration] Single Docker image for historical data migration


kartg:
relates #162 

Currently, the historical data migration process consists of two disconnected tools:
1. The [index configuration tool](https://github.com/opensearch-project/opensearch-migrations/tree/main/index_configuration_tool) to compare source and target clusters and setup target indices correctly
2. The data migration tool (currently [datastash](https://github.com/opensearch-project/opensearch-migrations/tree/main/datastash))

Typically, these tools will always be executed in sequence. The primary goal of this task is to set up a Docker image (and possibly a corresponding Docker compose file) that integrates the two tools together into a data migration pipeline. This will include a step where the index configuration tool updates the Data Prepper config file to only migrate data for the subset of indices it operated on.

Issue #163, opened 2023-05-24
[Historical Data Migration] Integrate Data Prepper Docker image


kartg:
relates #162 

* Change the datastash Docker compose file to use Data Prepper images instead of Logstash
* Write a Data Prepper configuration file to replace the Logstash one 
* Test the data migration pipeline

sumobrian:
Tracking through Jira - Epic https://opensearch.atlassian.net/browse/MIGRATIONS-1154

PR #234, opened 2023-07-07
[Index Configuration Tool] Unified Dockerfile for ICT and Data Prepper

kartg:
### Description

This change updates the Dockerfile definition for index_configuration_tool (ICT) to use Data Prepper as the base image. This will allow us to run (as a single Docker image) the index_configuration_tool Python logic to configure indices on the target cluster before kicking off Data Prepper execution to move data over. Note that since these steps must occur in sequence, the child Docker image overrides the parent's CMD definition and must manually invoke the data-prepper executable.

This is a POC implementation, so there are some limitations:

* The data migration pipeline does not shut down automatically once all data has been migrated. This is pending platform support in Data Prepper - https://github.com/opensearch-project/data-prepper/issues/2944
* If a `_bulk` call fails during data migration, the index-creation steps performed by ICT are not rolled back. These must be manually removed before running the Docker image again. This is necessary because ICT errs on the side of caution - the current logic does not validate `doc_count` for identical indices so we may end up incorrectly overwriting documents on the target cluster. Even if the `doc_count` is identical, it would require a much deeper check to verify if the contents of the index are identical.
* If there are no indices to migrate, the Docker container should ideally just shut down without running Data Prepper. Currently, Data Prepper is still kicked off but fails with an ugly stack-trace. This is because the Dockerfile simply defines two commands in sequence.
* There is scope to simplify the input mechanism for the unified Docker image. Today, it expects a Data Prepper YAML file, but this would require an understanding of the DP input format. Ideally, the input to the tool should be far simpler - possibly user-interactive and smartly deriving expected inputs based on endpoints. 
* Today, if an index is present on both source and target clusters but differs in settings/mappings, it is classified as a “conflict” and data is not migrated for it. However, it may be safe to migrate data that only differ in settings and not mappings, since these are unlikely to behave differently. We would still need to first incorporate a check for `doc_count` though (like with identical indices above)

### Issues Resolved
relates #165 and #163 

### Testing
```
.........................Wrote output YAML pipeline to: dummy
....
Ran 29 tests in 0.043s

OK
main.py                 112      0   100%
TOTAL                   162      2    99%

```

### Check List

### DIFF
--- a/index_configuration_tool/Dockerfile
+++ b/index_configuration_tool/Dockerfile
@@ -6,7 +6,8 @@ RUN apk update
 RUN apk add --no-cache python3 py-pip
 RUN pip install --user -r requirements.txt
 
-WORKDIR /code
+ENV ICT_CODE_PATH /code
+WORKDIR $ICT_CODE_PATH
 # Copy only source code
 COPY ./*.py .
 
@@ -14,4 +15,4 @@ COPY ./*.py .
 ENV PATH=/root/.local:$PATH
 
 # make sure you include the -u flag to have our stdout logged
-CMD python -u ./main.py -r $DATA_PREPPER_PATH/pipelines/pipelines.yaml $DATA_PREPPER_PATH/pipelines/pipelines.yaml; $DATA_PREPPER_PATH/bin/data-prepper
+ENTRYPOINT python -u ./main.py -r $ICT_CODE_PATH/input.yaml $DATA_PREPPER_PATH/pipelines/pipelines.yaml; $DATA_PREPPER_PATH/bin/data-prepper
--- a/index_configuration_tool/Dockerfile
+++ b/index_configuration_tool/Dockerfile
@@ -1,15 +1,12 @@
-# Multistage build - 1st stage
-FROM python:3.11 AS builder
+FROM opensearch-data-prepper:2.4.0-SNAPSHOT
 COPY requirements.txt .
 
 # Install dependencies to local user directory
+RUN apk update
+RUN apk add --no-cache python3 py-pip
 RUN pip install --user -r requirements.txt
 
-# Multistage build - 2nd stage
-FROM python:3.11-slim
 WORKDIR /code
-# Copy only required dependencies
-COPY --from=builder /root/.local /root/.local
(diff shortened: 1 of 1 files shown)


index_configuration_tool/Dockerfile
```diff
@@ -1,20 +1,17 @@
...
-COPY --from=builder /root/.local /root/.local
 # Copy only source code
 COPY ./*.py .
 
 # update PATH
 ENV PATH=/root/.local:$PATH
 
 # make sure you include the -u flag to have our stdout logged
-ENTRYPOINT [ "python", "-u", "./main.py" ]
+CMD python -u ./main.py -r $DATA_PREPPER_PATH/pipelines/pipelines.yaml $DATA_PREPPER_PATH/pipelines/pipelines.yaml; $DATA_PREPPER_PATH/bin/data-prepper
```
mikaylathompson, line 17: This is more of a me-and-docker question, but why switch from ENTRYPOINT to CMD?

kartg: > This is more of a me-and-docker question, but why switch from ENTRYPOINT to CMD?

Good question! I was using `CMD` because the [base Data Prepper Dockerfile](https://github.com/kartg/data-prepper/blob/main/release/docker/Dockerfile) uses `CMD`, but it turns out [ENTRYPOINT in the child image overrides the parent CMD](https://stackoverflow.com/a/49031590). Updated.

> This seems to be largely because we're reading from and writing to the same YAML file, and then running with that YAML file. Would the situation be "safer" if we e.g. renamed pipelines.yaml to original-pipelines.yaml, used that as the source for ./main.py but output to pipelines.yaml and then ran data-prepper? Would that ensure that only the correct indices get migrated?

Agree that this is a reasonable mitigation. I've changed the path for the input YAML file for ICT to not live in the data-prepper `pipelines` directory. While this mitigates this bug, the Docker image still runs the commands in sequence, which causes Data Prepper to start but fail because no pipeline configuration is found. I'm tracking a fix for this in https://opensearch.atlassian.net/browse/MIGRATIONS-1218

mikaylathompson (approved):

PR #239, opened 2023-07-18
Merge Kafka Puller into Traffic Replayer

lewijacn:
### Description
This change merges the kafka puller into the replayer, and by doing so needed to accomplish the following items:
* Adjust kafka consumer logic to work inside the replayer and update tests
* Remove references to the kafka puller and its directory
* Adjust docker solution to remove kafka puller and update replayer arguments
* Adjust copilot to remove kafka puller and update replayer arguments

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1149

### Testing
Unit tests and manual deployment tests

### Check List


TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumerTest.java
```diff
@@ -0,0 +1,204 @@
...
+            Assertions.assertEquals(substreamCounts.size(), numTrafficStreams);
+        });
+
+        AtomicInteger foundStreamsCount = new AtomicInteger(0);
+        trafficStream.forEach(stream -> {
+            log.trace("Stream has substream count: " + stream.getSubStreamCount());
+            Assertions.assertTrue(stream instanceof TrafficStream);
+            Assertions.assertEquals(stream.getSubStreamCount(), substreamCounts.get(foundStreamsCount.get()));
+            foundStreamsCount.getAndIncrement();
+            if (foundStreamsCount.get() == numTrafficStreams) {
```
gregschohn, line 110: same comments as above

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+        if (brokers == null || topic == null || groupId == null) {
+            throw new RuntimeException("To enable a Kafka traffic source, the following parameters are required " +
+                "[--kafka-traffic-brokers, --kafka-traffic-topic, --kafka-traffic-group-id]");
+        }
+        var kafkaProps = buildKafkaProperties(brokers, groupId, enableMSKAuth, propertyFilePath);
+        return new KafkaProtobufConsumer(new KafkaConsumer<>(kafkaProps), topic, behavioralPolicy);
+    }
+
+    public static Properties buildKafkaProperties(String brokers, String groupId, boolean enableMSKAuth, String propertyFilePath) {
```
gregschohn, line 52: there are 3 string parameters.  Consider renaming this to include an indication as to what the order of the arguments should be.  buildKafkaParametersWithBrokerAndGroupId.  Even as I say that, I don't really like it - but I get nervous that brokers and groupId can be easily transposed (sometimes I'll make wrapper classes for added type safety).

gregschohn (commented):
I'm fine with most of the changes here except for deleting the KafkaPrinter project.  It doesn't have a place in a production release, but it can be exceptionally valuable even today as a debug tool.  I'm thinking of the case where the replayer chokes up.  I'd love to hop onto the box/container and run this Kafka dumper program to output the Kafka stream to a log so that I can test the replayer later on.  Let me know what you think about leaving the project in place (but deleting the container and doing all other refactoring)

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ITrafficCaptureSource.java
```diff
@@ -0,0 +1,11 @@
...
+
+import org.opensearch.migrations.trafficcapture.protos.TrafficStream;
+
+import java.io.Closeable;
+import java.util.stream.Stream;
+
+public interface ITrafficCaptureSource extends Closeable {
+
+    Stream<TrafficStream> supplyTrafficFromSource();
+}
```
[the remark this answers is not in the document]
lewijacn: We should discuss on this, I feel like I'm not picking up the intent here very well

lewijacn:
> I'm fine with most of the changes here except for deleting the KafkaPrinter project. It doesn't have a place in a production release, but it can be exceptionally valuable even today as a debug tool. I'm thinking of the case where the replayer chokes up. I'd love to hop onto the box/container and run this Kafka dumper program to output the Kafka stream to a log so that I can test the replayer later on. Let me know what you think about leaving the project in place (but deleting the container and doing all other refactoring)

If this is still valuable we shouldn't remove. My expectation was that we could leverage this with some debug enabled logic to achieve the same but it does feel a bit clunky as I think about it. My current stance is we should keep but if we are going to use it as a utility it should be integrated a little better

gregschohn (approved):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficCaptureSourceFactory.java
```diff
@@ -0,0 +1,31 @@
...
+import org.opensearch.migrations.replay.kafka.KafkaBehavioralPolicy;
+import org.opensearch.migrations.replay.kafka.KafkaProtobufConsumer;
+
+import java.io.FileInputStream;
+import java.io.IOException;
+
+@Slf4j
+public class TrafficCaptureSourceFactory {
+
+    public static ITrafficCaptureSource createTrafficCaptureSource(TrafficReplayer.Parameters appParams) throws IOException {
```
gregschohn, line 13: nit: Since this takes the Parameters object & Parameters is a construct of the command line main class, this could e folded to within TrafficReplayer too.  It helps that it's really short.  I can't imagine another use case where I'd want this functionality other than from the command line where multiple options were supplied.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaBehavioralPolicy.java
```diff
@@ -9,7 +9,7 @@
 public class KafkaBehavioralPolicy {
 
     // Default behavior to log and ignore invalid records
```
gregschohn, line 11: Can you supply some more detailed docs now that the behavior has changed?  What should a user expect if an Exception is returned?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficCaptureInputStream.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Objects;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+import java.util.stream.Stream;
+
+@Slf4j
+public class TrafficCaptureInputStream implements ITrafficCaptureSource {
```
gregschohn, line 14: ...InputStream makes me expect to see a read() call that would return some data.  In this case, you're supplying data from read() calls.  You could call this TrafficCaptureSourceFromInputStream or something else.  As it is, it fits in the mold of FileInputStream, BufferedInputStream, etc.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaBehavioralPolicy.java
lewijacn: Added a proper javadoc

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficCaptureInputStream.java
lewijacn: Renamed to hopefully be clear and not as wordy

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficCaptureSourceFactory.java
lewijacn: I probably agree with you right now, my thought here is that when more sources are added in the future, it would be nice to contain this factory source logic here

PR #241, opened 2023-07-25
Fix MIGRATIONS-1268 (capturing encrypted, rather than plaintext data)

gregschohn:
The fix itself to NettyPacketToHttpConsumer is trivial.  I've spent more efforts on getting a unit test setup to be able to do more live testing with TLS. I've intentionally used client and server implementations different from netty since the real world will have clients that aren't just netty.

### Description
* **Category**: Bug fix
* **Why these changes are required**?  We were capturing encrypted responses rather than plaintext ones, making it impossible to compare them or reason with them in any way.
* **What is the old behavior before changes and new behavior after changes?** 
 - Before: all response packets were encrypted when the target server was using TLS
 - Now: response pieces are first decrypted before being added to the packet responess.

### Issues Resolved
MIGRATIONS-1268

### Testing
Unit test

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
@@ -85,7 +85,7 @@ public class NettyPacketToHttpConsumer implements IPacketFinalizingConsumer<Aggr
         responseBuilder = AggregatedRawResponse.builder(Instant.now());
         b.group(eventLoopGroup)
                 .channel(NioSocketChannel.class)
-                .handler(new BacksideSnifferHandler(responseBuilder))
+                .handler(new ChannelDuplexHandler())
                 .option(ChannelOption.AUTO_READ, false);
         String host = serverUri.getHost();
         int port = serverUri.getPort();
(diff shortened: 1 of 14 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -133,6 +133,7 @@ public NettyPacketToHttpConsumer(NioEventLoopGroup eventLoopGroup, URI serverUri
     }
 
     private void addHttpTransactionHandlersToPipeline(ChannelPipeline pipeline) {
+        pipeline.addLast(new BacksideSnifferHandler(responseBuilder));
```
lewijacn, line 136: I've been trying to understand why BacksideSnifferHandler is a ChannelDuplexHandler when we are only acting on channelRead, would this make it justifiable to be a ChannelInboundHandlerAdapter or is there something else happening here?

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumerTest.java
```diff
@@ -0,0 +1,123 @@
...
+        var headers = Map.of(
+                "Content-Type", "text/plain",
+                "Funtime", "checkIt!",
+                "Content-Transfer-Encoding", "chunked");
+        var payloadBytes = SERVER_RESPONSE_BODY.getBytes(StandardCharsets.UTF_8);
+        return new SimpleHttpResponse(headers, payloadBytes, "OK", 200);
+    }
+
+    @Test
+    public void testHttpResponseIsASuccessfullyCaptured()
```
lewijacn, line 103: nit: remove 'A'

lewijacn (approved):

PR #245, opened 2023-07-28
Capture connection close events and react to them

gregschohn:
### Description
Emit the close observation when a channel is unregistered and handle that close observation within the replayer.
In the replayer, when the close event is encountered, it's presumed that NO OTHER traffic observations should follow. If any ARE found, we'll emit an error message.
I had to rework a couple other assumptions in CapturedTrafficToHttpTransactionAccumulator, like returning whether an observation was a close or not.

* **Category** Enhancement
* **Why these changes are required**? To send requests to the target cluster in a more timely manner
 - Some requests would hold for 30 seconds (expiration period) even if the request's sockets had closed
 - Upon getting a close observation, the connection/accumulation will be considered to be expired and requests and responses will be flushed.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1183

### Testing
Unit testing and dockerSolution

### Check List


TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -112,7 +112,8 @@ private void writeObservationTag(int fieldNumber) throws IOException {
                 getWireTypeForFieldIndex(TrafficObservation.getDescriptor(), fieldNumber));
     }
 
-    private void beginWritingObservationToCurrentStream(Instant timestamp, int captureTag, int captureClosureSize) throws IOException {
+    private void beginSubstreamObservationToCurrentStream(Instant timestamp, int captureTag, int captureClosureSize) throws IOException {
```
lewijacn, line 115: Isn't substream and observation synonmous, this throws me off a bit. Maybe `writeSubstreamBeginningToCurrentStream` and `writeSubstreamBeginningWithTagToCurrentStream` for later one? Even then the ToCurrentStream could possibly just be implied since everything in this file is working off of the current stream

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Accumulation.java
```diff
@@ -47,6 +47,9 @@ public String toString() {
         final StringBuilder sb = new StringBuilder("Accumulation{");
         sb.append("rrPair=").append(rrPair);
         sb.append(", state=").append(state);
+        if (trafficStreamsSorter.hasPending()) {
```
lewijacn, line 50: Any reason to only output this with pending? Could it be useful to know the offset or items even if the accumulation is caught up. I could also see this being unnecessary noise so I think I'm fine for now until I find a case where I actually want this regardless

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/TrafficReplayerTest.java
```diff
@@ -149,10 +154,36 @@ public void testReader() throws IOException, URISyntaxException {
         Assertions.assertTrue(byteArrays.stream().allMatch(ba->ba.size()==2));
     }
 
+    @Test
+    public void testMissingStreamCausesWarning() throws URISyntaxException, IOException, ExecutionException, InterruptedException {
```
lewijacn, line 158: Could you highlight where the missing stream gets introduced? It wasn't immediately obvious to me where we had cut out or removed a traffic stream then saw we create a traffic stream with a higher chunk number

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -90,22 +101,41 @@ private static String summarizeTrafficStream(TrafficStream ts) {
 
     public void accept(TrafficStream yetToBeSequencedTrafficStream) {
         log.trace("Got trafficStream: " + summarizeTrafficStream(yetToBeSequencedTrafficStream));
-        var accum = liveStreams.getOrCreateWithoutExpiration(
-                yetToBeSequencedTrafficStream.getNodeId(),
-                yetToBeSequencedTrafficStream.getConnectionId());
+        var partitionId = yetToBeSequencedTrafficStream.getNodeId();
```
lewijacn, line 104: nit: Calling this partition makes me think of Kafka topic partitions

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -215,9 +257,14 @@ public void close() {
...
 
     private void fireAccumulationsCallbacks(Accumulation accumulation) {
+        if (accumulation.trafficStreamsSorter.hasPending()) {
+            onTrafficStreamMissingOnExpiration.accept(accumulation);
+        }
         switch (accumulation.state) {
             case NOTHING_SENT:
-                handleEndOfRequest(accumulation);
+                if (!handleEndOfRequest(accumulation)) {
+                    return;
```
lewijacn, line 266: What does this return achieve? Wouldn't we just complete this method after this switch case anyway?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -90,22 +101,41 @@ private static String summarizeTrafficStream(TrafficStream ts) {
...
+                                log.error("Got a traffic observation AFTER a Close observation for the stream " +  ts);
+                                return true;
+                            }
+                            var didTerminate = CONNECTION_STATUS.CLOSED ==
+                                    addObservationToAccumulation(ts.getNodeId(), accum, o);
+                            terminated.set(didTerminate);
+                            return didTerminate;
+                        })
+                        .takeWhile(b->!b)
+                        .forEach(b->{}));
```
lewijacn, line 121: Curious what this forEach line is doing?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Accumulation.java
gregschohn: It would just be empty, so when you see an Accumulation w/out this part, you can infer that trafficStreamsSorter has no elements within it.

lewijacn: The offset count would still be shown, but I don't really see a need to have that right now

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
lewijacn: I'm not sure I follow this but its not a blocker for me either way

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
lewijacn: Okay this makes more sense now that we are juggling a bit of processing this stream as well as reserving the ability to terminate early if the connection is closed. Part of me thinks a simple while loop instead of a stream would have been cleaner here, but its not a blocker for me

PR #254, opened 2023-08-03
Add Kafka offloader bug fixes and minor improvements

lewijacn:
### Description
This changes introduces a set of improvements, these incude:
* Addressing the Kafka offloader RecordTooLargeException by adding a needed overhead allowance for record key bytes and other Kafka metadata required bytes. Included unit tests
* Adjusting the Kafka offloader flush handler to complete its future exceptionally if an error is encountered instead of completing normally
* Changed how Kafka properties file argument is used to act as an additional Kafka client customization mechanism rather than an all encompassing properties file
* Altered the Kafka offloader to not keep track of its own integer for number of traffic streams and instead have the CaptureSerializer pass the traffic stream index to create more harmony between these two and make debugging between the proxy and replayer have a consistent id (connection-id.traffic stream index)
* Adding some default Kafka producer client configuration to lower some timeout values. From what I have seen the retry logic for this Kafka client is robust enough to use for now, and several guides recommend adjusting the timeouts for your use case as needed and not adjusting the retries. As far as the timeout values go, these are still a work in progress. Realistically, this could vary depending on the user's tolerance but the main concern right now is around handling cases where Kafka goes down and we want to impact the normal user's traffic as little as possible as Kafka requests fail. This is more of an issue for mutating requests where we will block responses to users until the Kafka offload future is completed. Ideally we can hand the response back and still block in another way to ensure mutating requests happen properly, but some more investigation needed here.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1264
https://opensearch.atlassian.net/browse/MIGRATIONS-1262
https://opensearch.atlassian.net/browse/MIGRATIONS-1213

### Testing
Unit testing and manual docker compose testing

### Check List


TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -78,14 +81,14 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
 
     private Callback handleProducerRecordSent(CompletableFuture cf, String recordId) {
```
gregschohn, line 82: Can you please add a comment referencing some of the guidance that you saw (from your description of the CR).  As a developer, this is where I might naturally want to put retry code.  Understanding what exceptions can happen here and which ones (all?) have been retried will be really helpful to future devs.

TrafficCapture/captureKafkaOffloader/src/test/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactoryTest.java
```diff
@@ -13,23 +21,73 @@
...
+        final var referenceTimestamp = Instant.now(Clock.systemUTC());
+
+        int messageSize = 1024*1024;
+        MockProducer<String, byte[]> producer = new MockProducer<>(true, new StringSerializer(), new ByteArraySerializer());
+        KafkaCaptureFactory kafkaCaptureFactory =
+            new KafkaCaptureFactory(TEST_NODE_ID_STRING, producer, messageSize);
+        IChannelConnectionCaptureSerializer serializer = kafkaCaptureFactory.createOffloader(connectionId);
+
+        StringBuilder sb = new StringBuilder();
+        // Create over 1MB packet
```
gregschohn, line 56: Could you put a self-documenting assertion in place to show the total size of the string.  With the escapes, it's very hard to find out the size of the text that's being appended every time.  It will also help in case anybody changes the json in the future.

TrafficCapture/captureKafkaOffloader/src/test/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactoryTest.java
```diff
@@ -13,23 +21,73 @@
...
+        byte[] fakeDataBytes = sb.toString().getBytes(StandardCharsets.UTF_8);
+        var bb = Unpooled.wrappedBuffer(fakeDataBytes);
+        serializer.addReadEvent(referenceTimestamp, bb);
+        CompletableFuture future = serializer.flushCommitAndResetStream(true);
+        future.get();
+        for (ProducerRecord<String, byte[]> record : producer.history()) {
+            int recordSize = calculateRecordSize(record, null);
+            Assertions.assertTrue(recordSize <= messageSize);
+            int largeIdRecordSize = calculateRecordSize(record, connectionId + ".9999999999");
+            Assertions.assertTrue(largeIdRecordSize <= messageSize);
```
gregschohn, line 69: I'm having difficulty following this.  What are you trying to check here?  It looks like messageSize should be a maxAllowableMessageSize, but I can't figure out what edge condition you're trying to exercise by adding a bigger connectionId.

TrafficCapture/captureKafkaOffloader/src/test/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactoryTest.java
```diff
@@ -13,23 +21,73 @@
...
+        }
+        bb.release();
+        producer.close();
+    }
+
+    /**
+     * This size calculation is based off the Kafka client 3.5 version request size validation check done from the Producer side
```
gregschohn, line 76: The client 3.5 comment is making me nervous.  If the packagte were to use a different client, would we still be able to test the same functionality of our code?

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -280,6 +280,9 @@ private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant
             bb = bb.slice();
             byteBuffer.position(byteBuffer.position() + chunkBytes);
             addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, segmentCountFieldNumber, ++dataCount, timestamp, bb);
+            if (getOrCreateCodedOutputStream().spaceLeft() < (availableCOSSpace - chunkBytes - trafficStreamOverhead)) {
+                log.warn("Writing segment substream to CodedOutputStream took more space than max calculated, this should be investigated");
```
gregschohn, line 284: If this is a programmer error, we should have a full assertion or thrown exception instead of a warning message.  This would be a critical error, blocking the current observations from being sent, right?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpRequestHandler.java
```diff
@@ -24,13 +23,11 @@ protected void channelFinishedReadingAnHttpMessage(ChannelHandlerContext ctx, Ob
...
                 if (t != null) {
                     log.warn("Got error: " + t.getMessage());
-                    ctx.close();
-                } else {
-                    try {
-                        super.channelFinishedReadingAnHttpMessage(ctx, msg, httpRequest);
-                    } catch (Exception e) {
-                        throw new RuntimeException(e);
-                    }
+                }
```
gregschohn, line 26: This should definitely be a policy that the user drives.  There should be at least two modes on offloader failures - proceed and block traffic.  We might want to get more granular by having different policies for mutable vs non-mutating calls too.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -39,6 +40,7 @@
 public class Main {
 
     private final static String HTTPS_CONFIG_PREFIX = "plugins.security.ssl.http.";
+    private final static String DEFAULT_KAFKA_CLIENT_ID = "KafkaLoggingProducer";
```
gregschohn, line 43: Kafka isn't very helpful in this string.  HttpCaptureProxyProducer would be better.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -170,24 +155,33 @@ private static IConnectionCaptureFactory getConnectionCaptureFactory(Parameters
...
+            kafkaProps.put(ProducerConfig.MAX_BLOCK_MS_CONFIG, 3000);
+
+            if (params.kafkaPropertiesFile != null) {
+                try {
+                    kafkaProps.load(new FileReader(params.kafkaPropertiesFile));
+                } catch (IOException e) {
+                    log.error("Unable to locate provided Kafka producer properties file path: " + params.kafkaPropertiesFile);
+                    throw e;
+                }
+            }
```
gregschohn, line 175: Can you write a test to show/document how precedence will work for these?  I would have expected to load the file BEFORE loading the other properties so that the more granular options could override the file rather than the other way around.

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
lewijacn: There are cases where this could be a critical error and cases where it's not. In the cases where we are filling up the CodedOutputStream as much as we can and we didn't calculate enough size we will already hit a no space available in CodedOutputStream error as a result.  There are other cases where we could potentially write more than we thought but would be pretty much untraceable since there was also enough space in the CodedOutputStream to handle. This pattern could go undetected unless we got a message that was close to the bounds.

Because of this I don't want to throw an exception since we could very well handle it, and if not the CodedOutputStream will throw an exception. I've added another check like this to a normal message that doesn't need segmented

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpRequestHandler.java
lewijacn: Originally thought I would quickly add this behavioral policy in but think it requires a little more refinement. For now adding a comment in code and created a JIRA: https://opensearch.atlassian.net/browse/MIGRATIONS-1276

TrafficCapture/captureKafkaOffloader/src/test/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactoryTest.java
```diff
@@ -13,23 +21,76 @@
...
+        bb.release();
+        producer.close();
+    }
+
+    /**
+     * This size calculation is based off the KafkaProducer client request size validation check done when Producer
+     * records are sent. This validation appears to be consistent for several versions now, here is a reference to
+     * version 3.5 at the time of writing this: https://github.com/apache/kafka/blob/3.5/clients/src/main/java/org/apache/kafka/clients/producer/KafkaProducer.java#L1030-L1032.
+     * It is, however, subject to change which may make this test scenario more suited for an integration test where
+     * a KafkaProducer does not need to be mocked.
```
gregschohn, line 80: Thanks for the callout.  Do you have a follow up task for it?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpRequestHandler.java
gregschohn: Shouldn't you be closing the context?  This feels like it could be a bad resource leak.

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
gregschohn: Hmmmm - so you're saying that your overage will make this a softer landing in many cases.
OK - let me turn this around.  If a user sees this, how can it be investigated?  The log message doesn't give me any useful information.  Can you at least put the sizes, maybe some ids (in case there are additional previous logs)

gregschohn:
I have 2 material comments that I'd like to see addressed.  1) are you sure that you can drop the ctx.close() call and 2) can you add a few more details to the investigation log statement so that users have something more to go on.

TrafficCapture/captureKafkaOffloader/src/test/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactoryTest.java
lewijacn: Sure have adjusted

TrafficCapture/captureKafkaOffloader/src/test/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactoryTest.java
lewijacn: A follow up task for an integration test to cover this right?

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
lewijacn: Sure good callout, I have added some additional details to this log line

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpRequestHandler.java
lewijacn: The problem I face is that if I close the context our pipeline ends, so it would be equivalent with saying if our Kafka offload fails we won't send the request to the source cluster.

gregschohn (approved):

PR #238, opened 2023-07-18
Fix TLS handshake

gregschohn:
### Description
Better reliability for the replayer to make send TLS traffic to target clusters!
* **_Category_** Bug fix
* **_Why these changes are required?_** Previously, some data was sent unencrypted and wasn't properly parsed or handled by the target cluster
* **_What is the old behavior before changes and new behavior after changes?_**
 * Old results - about 87-98% accuracy on geonames and log OpenSearch benchmark datasets.
 * New results - about 100% accuracy on geonames and log OpenSearch benchmark datasets.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1190 

### Testing
dockerSolution

### Check List
- [ ] New functionality includes testing
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #256, opened 2023-08-08
Add optional connection pooling to the destination server for the Capture Proxy

gregschohn:
### Description
* Category: Enhancement/New feature
* Why these changes are required? To improve the performance of the proxy/reduce latency of incoming client calls (by reducing backend setup time)
* What is the old behavior before changes and new behavior after changes?  With the --destinationConnectionPoolSize argument, some calls will be able to remove the need to wait for TCP and TLS handshaking, dropping the latency that the client observes.

Some of the other work that happened here was wiring command line parameters. A slightly larger bit of code deals with a couple extra levels of hierarchy between the BacksideConnectionPool and the ExpiringSubstitutableItemPool. One level deals with different EventLoops (threads) via a FastThreadLocal (netty's version of ThreadLocal) and the second is a HashMap that organizes different pools for each of those threads by channel class, since we need to make sure that we're returning the right kinds of channels.

The underlying cache allows callers to retrieve objects that were created within a window, while expiring objects that hadn't been pulled within that window. Once objects are given to the caller (via getAvailableOrNewItem), they are removed from the cache. Upon items leaving the cache, new items creation is begun to backfill the old items. If there are no ready items to be retrieved, the oldest future to build an item is instead returned. If there are no "in progress" items being built, then an item is supplied directly (the future from the a new invocation of the supplier goes straight back to the user). Creation order is maintained to try to minimize the perceived latency for callers.

Lastly, I'm not sure that the SSL handshaking to the destination server would always completely finish before we tried to write (or read) data on the channel. This was an issue for the Replayer (see https://github.com/opensearch-project/opensearch-migrations/pull/238). I think that the solution here, where a new ChannelFuture is bound to the SSLHandler's handshakeFuture() is a much cleaner approach.

This cache allows callers to retrieve objects that were created within a window, while expiring objects that hadn't been pulled within that window.  Once objects are given to the caller (via getAvailableOrNewItem), they are removed from the cache.  Upon items leaving the cache, new items creation is begun to backfill the old items.  If there are no ready items to be retrieved, the oldest future to build an item is instead returned.  If there are no "in progress" items being built, then an item is supplied directly (the future from the a new invocation of the supplier goes straight back to the user).  Creation order is maintained to try to minimize the perceived latency for callers.

The unit test for this class uses some concurrency primitives to mitigate the effects of time, but I'm not controlling the time used for expiration, so some of the expiration checks, which aren't as critical, could be unreliable when in a debugger.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1215

### Testing
Unit tests

### Check List


mikaylathompson:
A few high level questions:
>  If there are no ready items to be retrieved, the oldest future to build an item is instead returned.
I want to make sure I understand what this is saying. By "no ready items" you're saying that none of the items here have _completed_ being built, but they're in progress. You're returning the oldest in progress one because it's presumably the closest to completion? What is the tradeoff between returning a future vs. making the caller wait (either blocking or via retries) for a complete one to be ready?

> If there are no "in progress" items being built, then an item is supplied directly.
Compared to the last one, this kind of sounds like "if nothing's ready, just make one". By "an item is supplied directly" are you saying that the a fully complete object is provided, or a future once one is started? If the former, is the caller blocked until it's ready? Once the item is ready, are there any ways in which this is this different from the usual pool case? (basically, what work is "directly" doing in this sentence?)

gregschohn:
For the first question: When I said, the oldest in-progress one is returned, we're actually returning the future.  We won't be blocking, but for the customer to actively use the value, they will have to wait somehow (probably via addListener chaining).  Blocking isn't an option in a netty eventloop unless you want to deadlock and/or have it throw exceptions.

For the second question, again, it's a future.  The future will have been produced the same way that the in-progress and ready items were constructed.  The only difference will be that it likely isn't as far along.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPool.java
```diff
@@ -0,0 +1,304 @@
...
+ * if no items are available within the cache that are ready to go.
+ *
+ * This class does not use locking.  Instead, it is assumed that one of these will be created for
+ * each netty event loop.
+ */
+@Slf4j
+public class ExpiringSubstitutableItemPool<F extends Future<U>, U> {
+
+    private static class Entry<F> {
+        Instant timestamp;
```
mikaylathompson, line 34: to verify: timestamp is the instant when creation begins, or when the future is completed? Which one matters for when a connection (or generic object) should expire?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPool.java
```diff
@@ -0,0 +1,304 @@
...
+        }
+    }
+
+    public static class PoolClosedException extends RuntimeException { }
+
+    public static class Stats {
+        @Getter
+        private long nItemsCreated;
+        @Getter
+        private long nExpiredItems;
```
mikaylathompson, line 57: very minor nit, but this would match the previous item better if it were `nItemsExpired`

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPool.java
```diff
@@ -0,0 +1,304 @@
...
+
+        public long getTotalGets() {
+            return nHotGets+nColdGets;
+        }
+
+        public Duration averageWaitTime() {
+            if (getTotalGets() == 0) {
+                return Duration.ZERO;
+            }
+            return totalWaitTimeForCallers.dividedBy(getTotalGets());
```
mikaylathompson, line 87: We should be cautious with this. My understanding is that if we have a situation where one item was called 1 second after being created, and 1000 other items have been created and never called, we'd report a wait time of 1 second, because wait time is only updated when an item is given to a caller.

I don't think it's a bad metric to have, especially if you contexturalize it with `totalItemsCreated` and `totalExpiredItems`, but we shouldn't rely on it being an accurate gauge on its own.

mikaylathompson: (The same thing can happen with build time if a bunch of futures are just building indefinitely)

mikaylathompson (approved):

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPool.java
gregschohn: Timestamps for expiration purposes are relative to when the item supplier returns (so the other way around from what you asked about).

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPool.java
gregschohn: fixed - thx

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPool.java
gregschohn: This is only relevant for the items that were provided to the user.  Hot and cold gets are collectively increased by 1 when getAvailableOrNewItem() is called.  totalItemsCreated() will be incremented each time that we supply a new item, including items that may be expired or those that are still pending.

Please let me know if I've got some other issue in these calculations.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/Main.java
```diff
@@ -96,6 +97,11 @@ static class Parameters {
                 arity = 1,
                 description = "Exposed port for clients to connect to this proxy.")
         int frontsidePort = 0;
+        @Parameter(required = false,
+                names = {"--numThreads"},
+                arity = 1,
+                description = "How many threads netty should create in its event loop group")
+        int numThreads;
```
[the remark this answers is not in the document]
gregschohn: now it does - I was also setting the numThreads on the boss group rather than the worker group, which has also been corrected.

PR #264, opened 2023-08-15
[Fetch Migration] Monitoring script for Data Prepper

kartg:
## Description
This change adds a `monitor.py` Python script that can monitor the running Data Prepper pipeline and shut it down once the target document count has been reached (as reported by Data Prepper's `documentsSuccess` Prometheus metric) and an idle pipeline is detected.
* Category: New feature

---

### Unit Tests

```
endpoint_info.py          6      0   100%
index_operations.py      35      0   100%
main.py                 121      1    99%
monitor.py               52      0   100%
TOTAL                   227      1    99%
```

### Integration Test

1. Setup step (see the Index Configuration Tool [readme](https://github.com/opensearch-project/opensearch-migrations/tree/main/FetchMigration)):

```
python index_configuration_tool/main.py -r <input_pipeline_yaml> <output_yaml_path>
```

2. Data Prepper container kickoff:

```
docker run -p 4900:4900 -v <output_yaml_path>:/usr/share/data-prepper/pipelines/pipelines.yaml opensearch-data-prepper:2.4.0-SNAPSHOT
```

3. After Data Prepper server has started, monitoring process kickoff:

```
python index_configuration_tool/monitor.py https://localhost:4900 <target_doc_count_from_report>
```

4. Verified that the monitoring process invoked the `/shutdown` endpoint once the target doc count was reached

---

## Limitations
* Though the monitoring script correctly stops the Data Prepper server, the Docker container does not terminate. The current theory is that this is due to https://github.com/opensearch-project/data-prepper/issues/3141
* Currently the endpoint auth information is hard-coded. The fix for this is tracked in JIRA task [MIGRATIONS-1285](https://opensearch.atlassian.net/browse/MIGRATIONS-1285)
* The monitoring script only tracks Data Prepper metrics to identify shutdown state. It does not check the target cluster to validate that the target document count has been reached. Since index readers need to refresh to make the newly ingested documents searchable, there is often a delay (~5 minutes from my local testing) in the document counts reflecting in the `_cat/indices` count of the target cluster. Handling this scenario is tracked in JIRA task [MIGRATIONS-1228](https://opensearch.atlassian.net/browse/MIGRATIONS-1228)

---

### Check List

### DIFF
--- a/FetchMigration/index_configuration_tool/monitor.py
+++ b/FetchMigration/index_configuration_tool/monitor.py
@@ -20,9 +20,13 @@ def shutdown_pipeline(endpoint: EndpointInfo):
     requests.post(shutdown_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl)
 
 
-def fetch_prometheus_metrics(endpoint: EndpointInfo) -> List[Metric]:
+def fetch_prometheus_metrics(endpoint: EndpointInfo) -> Optional[List[Metric]]:
     metrics_endpoint = endpoint.url + __PROMETHEUS_METRICS_ENDPOINT
-    response = requests.get(metrics_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl)
+    try:
+        response = requests.get(metrics_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl)
+        response.raise_for_status()
+    except requests.exceptions.RequestException:
+        return None
     # Based on response headers defined in Data Prepper's PrometheusMetricsHandler.java class
     metrics = response.content.decode('utf-8')
     # Collect generator return values into list
@@ -55,17 +59,20 @@ def run(args: argparse.Namespace, wait_seconds: int) -> None:
     endpoint = EndpointInfo(args.dp_endpoint, default_auth, False)
     prev_no_partitions_count = 0
     terminal = False
-    target_doc_count = int(args.target_count)
     while not terminal:
+        # If the API call fails, the response is empty
         metrics = fetch_prometheus_metrics(endpoint)
-        success_docs = get_metric_value(metrics, __DOC_SUCCESS_METRIC)
-        rec_in_flight = get_metric_value(metrics, __RECORDS_IN_FLIGHT_METRIC)
-        no_partitions_count = get_metric_value(metrics, __NO_PARTITIONS_METRIC)
-        terminal = check_if_complete(success_docs, rec_in_flight, no_partitions_count,
-                                     prev_no_partitions_count, target_doc_count)
+        if metrics is not None:
+            success_docs = get_metric_value(metrics, __DOC_SUCCESS_METRIC)
+            rec_in_flight = get_metric_value(metrics, __RECORDS_IN_FLIGHT_METRIC)
+            no_partitions_count = get_metric_value(metrics, __NO_PARTITIONS_METRIC)
+            terminal = check_if_complete(success_docs, rec_in_flight, no_partitions_count,
+                                         prev_no_partitions_count, args.target_count)
+            if not terminal:
+                # Save no_partitions_count
+                prev_no_partitions_count = no_partitions_count
+
         if not terminal:
-            # Save no_partitions_count
-            prev_no_partitions_count = no_partitions_count
             time.sleep(wait_seconds)
     # Loop terminated, shut down the Data Prepper pipeline
     shutdown_pipeline(endpoint)
@@ -75,9 +82,9 @@ if __name__ == '__main__':  # pragma no cover
     # Set up parsing for command line arguments
     arg_parser = argparse.ArgumentParser(
         prog="python monitor.py",
-        description="Monitoring process for a running Data Prepper pipeline.\n" +
-                    "The first input is the Data Prepper URL endpoint.\n" +
-                    "The second input is the target doc_count for termination.",
+        description="""Monitoring process for a running Data Prepper pipeline.
+        The first input is the Data Prepper URL endpoint.
+        The second input is the target doc_count for termination.""",
         formatter_class=argparse.RawTextHelpFormatter
     )
     # Required positional arguments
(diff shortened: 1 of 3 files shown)


kartg:
The [codecov diff hit](https://app.codecov.io/gh/opensearch-project/opensearch-migrations/pull/264?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=opensearch-project) seems inaccurate since it's looking at Java code which isn't a part of this change.

FetchMigration/index_configuration_tool/monitor.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def shutdown_pipeline(endpoint: EndpointInfo):
+    shutdown_endpoint = endpoint.url + __SHUTDOWN_ENDPOINT
+    requests.post(shutdown_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl)
+
+
+def fetch_prometheus_metrics(endpoint: EndpointInfo) -> List[Metric]:
+    metrics_endpoint = endpoint.url + __PROMETHEUS_METRICS_ENDPOINT
+    response = requests.get(metrics_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl)
```
mikaylathompson, line 25: should this be doing a `response.raise_for_status` type thing here, before it assumes it can grab the content?

FetchMigration/index_configuration_tool/monitor.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+def run(args: argparse.Namespace, wait_seconds: int) -> None:
+    # TODO Remove hardcoded EndpointInfo
+    default_auth = ('admin', 'admin')
+    endpoint = EndpointInfo(args.dp_endpoint, default_auth, False)
+    prev_no_partitions_count = 0
+    terminal = False
+    target_doc_count = int(args.target_count)
+    while not terminal:
```
mikaylathompson, line 59: How high is the risk that a doc fails to be written for whatever reason, and we're stuck here forever because doc_count is never >= target? Should we also have a timeout value that if doc_count doesn't change for X seconds/minutes, we give up?

FetchMigration/index_configuration_tool/monitor.py
```diff
@@ -0,0 +1,95 @@
...
+            # No-partitions metrics should steadily tick up
+            if no_part_count is not None and no_part_count > prev_no_part_count > 0:
+                return True
+    return False
+
+
+def run(args: argparse.Namespace, wait_seconds: int) -> None:
+    # TODO Remove hardcoded EndpointInfo
+    default_auth = ('admin', 'admin')
+    endpoint = EndpointInfo(args.dp_endpoint, default_auth, False)
```
mikaylathompson, line 55: Could auth be an arg too?

FetchMigration/index_configuration_tool/monitor.py
```diff
@@ -0,0 +1,95 @@
...
+        formatter_class=argparse.RawTextHelpFormatter
+    )
+    # Required positional arguments
+    arg_parser.add_argument(
+        "dp_endpoint",
+        help="URL endpoint for the running Data Prepper process"
+    )
+    arg_parser.add_argument(
+        "target_count",
+        help="Target doc_count to reach, after which the Data Prepper pipeline will be terminated"
```
mikaylathompson, line 90: you can set `type=int` here and let argparse do the type checking for you, and eliminate the risk of a type coercion failure on line 58

mikaylathompson: (that also means that the help string will tell users that it must be an int)

FetchMigration/index_configuration_tool/monitor.py
```diff
@@ -0,0 +1,95 @@
...
+    shutdown_pipeline(endpoint)
+
+
+if __name__ == '__main__':  # pragma no cover
+    # Set up parsing for command line arguments
+    arg_parser = argparse.ArgumentParser(
+        prog="python monitor.py",
+        description="Monitoring process for a running Data Prepper pipeline.\n" +
+                    "The first input is the Data Prepper URL endpoint.\n" +
+                    "The second input is the target doc_count for termination.",
```
mikaylathompson, line 80: totally optional, but you can use triple-quotes in python to write a multiline string, and not have to manually include `\n`.

mikaylathompson (commented):
Biggest issue is around error handling -- if any of the api calls fail, even transiently, the entire thing will crash.

FetchMigration/index_configuration_tool/monitor.py
kartg: Incorporated in followup commit - the method now returns `None` if the request hits an exception. In the main monitoring loop, this code path simply skips processing for the current polling interval

FetchMigration/index_configuration_tool/monitor.py
kartg: I think that's a good callout. I'll capture this as a follow up item in JIRA - https://opensearch.atlassian.net/browse/MIGRATIONS-1303

FetchMigration/index_configuration_tool/monitor.py
kartg: it _could_ but the Data Prepper process is bundled with the Fetch Migration solution and we're not expecting the user to change/customize this

FetchMigration/index_configuration_tool/monitor.py
kartg: Also added unit tests for this

FetchMigration/index_configuration_tool/monitor.py
mikaylathompson: hmm, okay, in that case I think I understand why `default_auth` _isn't_ an argument, but why _is_ `dp_endpoint`?

mikaylathompson (approved):

PR #266, opened 2023-08-17
Add logging config so output tuples go to shared filesystem

mikaylathompson:

### Description
This creates a logger used specifically for outputting the tuples, as well as a Log4j config section that puts the logs from that logger into a file on the shared file system path `/shared-replayer-output/`. They are formatted as jsons in each line, but the overall file is not a json. The `message` in the log json is a string-formatted  The log files are rolled over every 10 MB and never deleted.

Caveats/Limitations:
- both the path for the log file and the rollover strategy are hardcoded.
- there's no accommodation for deleting files to save space, so the space used with balloon infinitely. Managing this is the user's responsibility.
- Error logs are not currently copied to the EFS/shared volume; this only handles the tuples. This should also be a pretty straightforward change, but somewhat lower priority.
- There is extra information included in the logs: timestamp, logger, thread info, etc. I played with trying to get rid of this, but it's surprisingly difficult. For the time being, I think it's okay to leave it as is. The info the user cares about is all in `message`.

Example output (as a beautified json--it's logged as one line):
```
{
    "instant": {
        "epochSecond": 1692300077,
        "nanoOfSecond": 272554000
    },
    "thread": "main",
    "level": "INFO",
    "loggerName": "OutputTupleJsonLogger",
    "message": "{\"request\":{\"Authorization\":\"Basic YWRtaW46YWRtaW4=\",\"Accept\":\"*/*\",\"User-Agent\":\"curl/7.87.0\",\"Request-URI\":\"/\",\"Host\":\"localhost:9200\",\"Method\":\"GET\",\"HTTP-Version\":\"HTTP/1.1\",\"body\":\"\"},\"primaryResponse\":{\"content-length\":\"538\",\"content-type\":\"application/json; charset=UTF-8\",\"response_time_ms\":76,\"HTTP-Version\":\"HTTP/1.1\",\"Status-Code\":\"200\",\"body\":\"ewogICJuYW1lIiA6ICIwMmE2ZmQ1YTAyNDUiLAogICJjbHVzdGVyX25hbWUiIDogImRvY2tlci1jbHVzdGVyIiwKICAiY2x1c3Rlcl91dWlkIiA6ICJCWXM0V1Q4TVEtaS03MWRNU3lkTTFRIiwKICAidmVyc2lvbiIgOiB7CiAgICAibnVtYmVyIiA6ICI3LjEwLjIiLAogICAgImJ1aWxkX2ZsYXZvciIgOiAib3NzIiwKICAgICJidWlsZF90eXBlIiA6ICJkb2NrZXIiLAogICAgImJ1aWxkX2hhc2giIDogIjc0N2UxY2M3MWRlZjA3NzI1Mzg3OGE1OTE0M2MxZjc4NWFmYTkyYjkiLAogICAgImJ1aWxkX2RhdGUiIDogIjIwMjEtMDEtMTNUMDA6NDI6MTIuNDM1MzI2WiIsCiAgICAiYnVpbGRfc25hcHNob3QiIDogZmFsc2UsCiAgICAibHVjZW5lX3ZlcnNpb24iIDogIjguNy4wIiwKICAgICJtaW5pbXVtX3dpcmVfY29tcGF0aWJpbGl0eV92ZXJzaW9uIiA6ICI2LjguMCIsCiAgICAibWluaW11bV9pbmRleF9jb21wYXRpYmlsaXR5X3ZlcnNpb24iIDogIjYuMC4wLWJldGExIgogIH0sCiAgInRhZ2xpbmUiIDogIllvdSBLbm93LCBmb3IgU2VhcmNoIgp9Cg==\",\"Reason-Phrase\":\"OK\"},\"shadowResponse\":{\"content-length\":\"566\",\"content-type\":\"application/json; charset=UTF-8\",\"response_time_ms\":617,\"HTTP-Version\":\"HTTP/1.1\",\"Status-Code\":\"200\",\"body\":\"ewogICJuYW1lIiA6ICJkYzk2MjJmYjcwMDMiLAogICJjbHVzdGVyX25hbWUiIDogImRvY2tlci1jbHVzdGVyIiwKICAiY2x1c3Rlcl91dWlkIiA6ICJIMFJ1R3p1dlNVcXpIQkNiY1U4bG1nIiwKICAidmVyc2lvbiIgOiB7CiAgICAiZGlzdHJpYnV0aW9uIiA6ICJvcGVuc2VhcmNoIiwKICAgICJudW1iZXIiIDogIjIuNy4wIiwKICAgICJidWlsZF90eXBlIiA6ICJ0YXIiLAogICAgImJ1aWxkX2hhc2giIDogImI3YTZlMDllNDkyYjFlOTY1ZDgyNzUyNWY3ODYzYjM2NmVmMGUzMDQiLAogICAgImJ1aWxkX2RhdGUiIDogIjIwMjMtMDQtMjdUMjE6NDM6MDkuNTIzMzM2NzA2WiIsCiAgICAiYnVpbGRfc25hcHNob3QiIDogZmFsc2UsCiAgICAibHVjZW5lX3ZlcnNpb24iIDogIjkuNS4wIiwKICAgICJtaW5pbXVtX3dpcmVfY29tcGF0aWJpbGl0eV92ZXJzaW9uIiA6ICI3LjEwLjAiLAogICAgIm1pbmltdW1faW5kZXhfY29tcGF0aWJpbGl0eV92ZXJzaW9uIiA6ICI3LjAuMCIKICB9LAogICJ0YWdsaW5lIiA6ICJUaGUgT3BlblNlYXJjaCBQcm9qZWN0OiBodHRwczovL29wZW5zZWFyY2gub3JnLyIKfQo=\",\"Reason-Phrase\":\"OK\"}}",
    "endOfBatch": false,
    "loggerFqcn": "org.apache.logging.log4j.spi.AbstractLogger",
    "threadId": 1,
    "threadPriority": 5
}
```

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1222

### Testing
Manual

### Check List


gregschohn (commented):
I don't like the idea of writing the output to BOTH the output stream AND the log.  I'd expect to be able to control the output with flags or redirects such that it will only go to one destination unless I explicitly specify two (not that I'd prioritize that feature).

If you extracted the TupleWriter into an interface, you should be able to setup two different ones dependent upon the command line parameters.

Secondly, since we don't have the ability to meter requests in time, limiting by size instead of time will give much more consistently sized files and will likely be drastically more manageable.

mikaylathompson:
> I don't like the idea of writing the output to BOTH the output stream AND the log. I'd expect to be able to control the output with flags or redirects such that it will only go to one destination unless I explicitly specify two (not that I'd prioritize that feature).

The reason I left the existing output stream exactly as is is that the format of the log changes--it becomes the tuple json nested within a log json. That means that directly outputting it works with the current Comparator whereas outputting it via a log will not (at least not without some comparator work or a translation layer). We had discussed the other day that we'd like it to stay fully compatible for the time being.

sumobrian:
> I don't like the idea of writing the output to BOTH the output stream AND the log. I'd expect to be able to control the output with flags or redirects such that it will only go to one destination unless I explicitly specify two (not that I'd prioritize that feature).

If we want to make this change, lets consider after cutting the release branch. We can always merge to release branch if needed but I'd like to start extracting comparator on release branch.

sumobrian (approved):

PR #258, opened 2023-08-09
IaC Minor Structure Adjustments

lewijacn:
### Description
This PR introduces the following changes:
* Rebranding the benchmarks container as the migration-console for its additional uses
* Providing an option to use a combined container of the Capture Proxy and Elasticsearch in the dockerSolution and Copilot
* Providing a volume for Replayer tuple output in the dockerSolution and Copilot (Actual usage of this with the Replayer service will be handled by another JIRA: https://opensearch.atlassian.net/browse/MIGRATIONS-1229)

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1274
https://opensearch.atlassian.net/browse/MIGRATIONS-1211
https://opensearch.atlassian.net/browse/MIGRATIONS-1212

### Testing
Manual docker and Cloud deployment testing

### Check List
- [ ] New functionality includes testing
  - [x] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #274, opened 2023-08-20
Add script to transform tuples to human readable format

mikaylathompson:

### Description
This adds a new feature as a follow up to #266 and #258 .
Those two PRs created a new EFS/shared volume attached to the replayer and migration console, and directed the replayer's output tuples to that volume.

The Replayer outputs base64 encoded message bodies in the tuples: this prevents issues where saving a gzipped message body in utf-8 in a json was corrupting them.

This PR adds a python script to the migration console that allows the user to specify a file of replayer-generated tuple logs. It loads each log line, extracts the tuple, and processes it:
- the message bodies are base64 decoded
- if the headers (Content-Encoding) indicate that the message is gzipped, it's unzipped
- if the headers (Content-Type) indicate that the message is a json, it's loaded as such
- if the uri indicates that the messages is a `_bulk` request, it's loaded as a bulk json (a list of jsons)

The processed tuples are then written to a new file. If an output file wasn't specified, it's the input file, but prefaced with `readable-`.

It takes a very tolerant approach to errors. If any errors are found (e.g. can't base64 decode, can't un-gzip, can't parse as json) it reports them (error & line number), but outputs the tuple to the final file in whatever most recent state it was able to generate (so if it can't be b64 decoded, the original tuple is written; if it can't be un-gzipped, the base64 decoded version is written). This is my best attempt to ensure that we tolerate weird data (which could be the OS cluster messing up its headers, for instance) and don't quietly hide it from a user's future analyses.

There are many ways this could be improved in the future:
- operating on an entire directory
- accepting/outputting streaming data
- more sophisticated and customizable parsing
- watching a directory and automatically running on each new file

but it is inherently a short-term tool, and I took a minimal but opinionated approach to giving the user a tool that allows them to make use of this data.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1260

### Testing
Manual.

### Check List

### DIFF
--- a/TrafficCapture/README.md
+++ b/TrafficCapture/README.md
@@ -90,8 +90,45 @@ will send requests to `capture-proxy-domain.com`, using the auth combo `admin`/`
 
 Support for Sigv4 signing and other auth options may be a future option.
 
+#### Understanding Data from the Replayer
+
+The Migration Console can be used to access and help interpret the data from the replayer.
+
+The data generated from the replayer is stored on an Elastic File System volume shared between the Replayer and Migration Console.
+It is mounted to the Migration Console at the path `/shared_replayer_output`. The Replayer generates files named `output_tuples.log`.
+These files are rolled over as they hit 10 Mb to a series of `output_tuples-%d{yyyy-MM-dd-HH:mm}.log` files.
+
+The data in these files is in the format of JSON lines, each of which is a log message containing a specific request-response-response tuple.
+The body of the messages is sometimes gzipped which makes it difficult to represent as text in a JSON. Therefore, the body field of all requests
+and responses is base64 encoded before it is logged. This makes the files stable, but not human-readable.
+
+We have provided a utility script that can parse these files and output them to a human-readable format: the bodies are
+base64 decoded, un-gzipped if applicable, and parsed as JSON if applicable. They're then saved back to JSON format on disk.
+
+To use this utility from the Migration Console,
+```sh
+$ ./humanReadableLogs.py --help
+usage: humanReadableLogs.py [-h] [--outfile OUTFILE] infile
+
+positional arguments:
+  infile             Path to input logged tuple file.
+
+options:
+  -h, --help         show this help message and exit
+  --outfile OUTFILE  Path for output human readable tuple file.
+
+# By default, the output file is the same path as the input file, but the file name is preface with `readable-`.
+$ ./humanReadableLogs.py /shared_replayer_output/tuples.log
+Input file: /shared_replayer_output/tuples.log; Output file: /shared_replayer_output/readable-tuples.log
+
+# A specific output file can also be specified.
+$ ./humanReadableLogs.py /shared_replayer_output/tuples.log --outfile local-tuples.log
+Input file: /shared_replayer_output/tuples.log; Output file: local-tuples.log
+```
+
 ### Capture Kafka Offloader
 
 The Capture Kafka Offloader will act as a Kafka Producer for offloading captured traffic logs to the configured Kafka cluster.
 
 Learn more about its functionality and setup here: [Capture Kafka Offloader](captureKafkaOffloader/README.md)
+
--- a/TrafficCapture/README.md
+++ b/TrafficCapture/README.md
@@ -96,7 +96,7 @@ The Migration Console can be used to access and help interpret the data from the
 
 The data generated from the replayer is stored on an Elastic File System volume shared between the Replayer and Migration Console.
 It is mounted to the Migration Console at the path `/shared_replayer_output`. The Replayer generates files named `output_tuples.log`.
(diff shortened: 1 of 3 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
```diff
@@ -4,10 +4,12 @@ ENV DEBIAN_FRONTEND noninteractive
 
 RUN apt-get update && \
     apt-get install -y --no-install-recommends python3.9 python3-pip python3-dev gcc libc-dev git curl && \
-    pip3 install opensearch-benchmark
```
mikaylathompson, line 7: Specifying these versions resolves a versioning conflict that was happening in some installs (I _think_ it was all new installs and my local just had a cached version that happened to work, but I'm not positive why it was mixed). The newest opensearch-benchmarks (1.1.0) has a dependency that does _not_ work with the newest version of `urllib3` (1.26.x) for this version of python, and pip wasn't detecting/resolving the issue itself. Locking to these versions fixes it.

sumobrian (commented):
The output isn't exactly JSON but it is human readable. An optimization to this would be to output valid JSON. I see two errors in the example, I've create:

1. Single quotes are used rather than double quotes for keys and values
2. Boolean values are not encoded as strings. 

This example demonstrates these issues:

```

**Comment by mikaylathompson** [comment]
To your comment about non-valid json: you're totally right, I"m just printing dicts instead of wriitng them back as JSONs. Fixing now and will rerun your test case.

**Review by sumobrian — APPROVED** [review]

File: TrafficCapture/dockerSolution/src/main/docker/migrationConsole/humanReadableLogs.py [hunk 2]
```diff
@@ -0,0 +1,146 @@
...
+    try:
+        decoded = unzipped.decode("utf-8")
+    except Exception as e:
+        logger.error(f"Body value on line {line_no} could not be decoded to utf-8: {e}. "
+                     "Skipping parsing body value.")
+        return unzipped
+    if is_json and len(decoded) > 0:
+        if is_bulk:
+            try:
+                return [json.loads(line) for line in decoded.splitlines()]
```
gregschohn, line 96: Do we expect that bulk json has one json doc on each line?  Is that how ES/OS work?  That seems pretty awful.
Aren't newlines normally considered just whitespace within JSON?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/humanReadableLogs.py
```diff
@@ -0,0 +1,146 @@
...
+            return decoded
+    return decoded
+
+
+def parse_tuple(line: str, line_no: int) -> dict:
+    item = json.loads(line)
+    message = item[LOG_JSON_TUPLE_FIELD]
+    tuple = json.loads(message)
+    try:
+        is_bulk_path = BULK_URI_PATH in get_element(URI_PATH, tuple, raise_on_error=True)
```
gregschohn, line 115: Is it possible for "_bulk" to be in another part of a path and this not be a bulk request?  Are query params included in the path?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/humanReadableLogs.py
```diff
@@ -0,0 +1,146 @@
...
+if __name__ == "__main__":
+    args = parse_args()
+    if args.outfile:
+        outfile = args.outfile
+    else:
+        outfile = args.infile.parent / f"readable-{args.infile.name}"
+    print(f"Input file: {args.infile}; Output file: {outfile}")
+
+    logging.basicConfig(level=logging.INFO)
+    with logging_redirect_tqdm():
```
gregschohn, line 142: What do these lines do?

TrafficCapture/README.md
```diff
@@ -90,8 +90,45 @@ will send requests to `capture-proxy-domain.com`, using the auth combo `admin`/`
...
+We have provided a utility script that can parse these files and output them to a human-readable format: the bodies are
+base64 decoded, un-gzipped if applicable, and parsed as JSON if applicable. They're then saved back to JSON format on disk.
+
+To use this utility from the Migration Console,
+```sh
+$ ./humanReadableLogs.py --help
+usage: humanReadableLogs.py [-h] [--outfile OUTFILE] infile
+
+positional arguments:
+  infile             Path to input logged tuple file.
```
gregschohn, line 114: Why doesn't this work on stdin/stdout?  Customers will very likely be compressing old (or all) files.  Expanding them first to files could be 100-1K times slower than piping them, not to mention much more difficult to manage the space for.

PR #276, opened 2023-08-21
MIGRATIONS-1289: Add end of segment message to capture serializer

lewijacn:
### Description
This changes adds the needed `EndOfSegmentsIndication` substream in the capture serializer for messages that needed to be segmented across multiple traffic streams. The Replayer already has logic to handle this message type, however the capture serializer had not been updated to actually send it. With this change added I no longer observe the Replayer hanging on the geonames workload and I am able to run our set of benchmarks multiple times without signs of stopping.

~~This also reworks the SecretCache to be a SecretsManager async client that retrieves the secret. This will be helpful in the future as needed with integrating with Netty, for now it has a synchronous method for constructing an auth header from a secret.~~
-> Split to https://github.com/opensearch-project/opensearch-migrations/pull/278

### Observations
As I was adding to some of the logic in the Capture Serializer, it occurred to me that we should probably adjust our model for writing to the CodedOutputStream slightly in the future. There are some cases we are writing to the CodedOuputStream a close event or an EndOfMessage event which makes no judgement on space remaining in the CodedOutputStream which could lead to error. What seems to make sense, and would probably reduce some complexity in our segmenting logic as well, is to wrap any call we make to add something to the CodedOutputStream in a form of size check to ensure it can fit and if not we flush the stream.
Task here: https://opensearch.atlassian.net/browse/MIGRATIONS-1294

While I was investigating the behavior of the Replayer in the case of a missing EndOfSegmentsIndication message, it seems it has the ability to completely halt the Replayer process without ending it which is pretty unsavory. This seems to warrant a deeper dive into our Consumers that we pass into the CapturedTrafficToHttpTransactionAccumulator to ensure that it returns consistently and does not block the accumulator.
Task here: https://opensearch.atlassian.net/browse/MIGRATIONS-1293

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1289

### Testing
Manual testing and updating unit tests

### Check List

### DIFF
--- a/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
+++ b/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
@@ -2,6 +2,7 @@ package org.opensearch.migrations.trafficcapture;
 
 import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.Timestamp;
+import org.opensearch.migrations.trafficcapture.protos.EndOfSegmentsIndication;
 import org.opensearch.migrations.trafficcapture.protos.TrafficObservation;
 import org.opensearch.migrations.trafficcapture.protos.TrafficStream;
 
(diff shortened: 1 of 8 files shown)


gregschohn (approved):
I'd like to see a test to confirm that we can write an endOfSegment and read it back in.  That would be good to add to StreamChannelConnectionCaptureSerializerTest.

gregschohn (commented):
I share your concern that other observations might not be accounting for their presence overflowing the underlying buffer.  If the max record size is 1M, that could happen in 1/1K HTTP transactions that need to spill into the next TrafficStream, which could happen pretty quickly for a busy cluster.

lewijacn:
> I'd like to see a test to confirm that we can write an endOfSegment and read it back in. That would be good to add to StreamChannelConnectionCaptureSerializerTest.

Added a couple test cases around this

PR #277, opened 2023-08-21
Handle Chunked Transfer-Encoding for humanReadableLogs

mikaylathompson:

### Description
We encountered several issues when running the humanReadableLogs on AWS:
1. It crashed on some gzipped messages. While it caught the error successfully, it output the in-progress byte string, and when that was written to JSON, JSON has no byte string type and it crashed.
2. When that was fixed, it became clear that the gzipped messages it was failing to unzip were both gzipped and had a "chunked" transfer encoding. This meant that they were split into several segments, with the length of those segments also written in, and generally were no longer gzip-compatible. By parsing through this data to extract the chunks, a continuous byte stream can be extracted, and that one is un-gzippable.

This also turned up an issue (hinted at in the last review) where the header keys used multiple capitalizations.  It would be better to have a fully case-insensitive lookup path, but for the short-term, I've made the lookup function check both the original capitalization and all lower case, which seems to cover the overwhelming majority of cases.

### Issues Resolved
No issue yet.

### Testing
Manual

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/humanReadableLogs.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/humanReadableLogs.py
@@ -166,7 +166,7 @@ def parse_tuple(line: str, line_no: int) -> dict:
                                           tuple, try_lowercase_keys=True) == TRANSFER_ENCODING_CHUNKED
         value = parse_body_value(base64value, content_encoding, content_type, is_bulk_path,
                                  is_chunked_transfer, line_no)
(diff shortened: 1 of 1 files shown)


sumobrian:
Regarding header casing, the python requests library handles differences in header casing w/ a case insensitive dictionary (CaseInsensitiveDict).

sumobrian (approved):
LGTM - We may want to swap the casing code added in this PR for a case insensitive dictionary and if this grows beyond a help script for our customers we may want to add tests.

PR #278, opened 2023-08-21
Refactor secrets cache to sdk client

lewijacn:
### Description
This reworks the SecretCache to be a SecretsManager async client that retrieves the secret. This will be helpful in the future as needed with integrating with Netty, for now it has a synchronous method for constructing an auth header from a secret.

### Issues Resolved
None

### Testing
Unit tests

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/AWSAuthService.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/AWSAuthService.java
@@ -1,30 +1,27 @@
 package org.opensearch.migrations.replay;
 
 import lombok.extern.slf4j.Slf4j;
-import software.amazon.awssdk.services.secretsmanager.SecretsManagerAsyncClient;
-import software.amazon.awssdk.services.secretsmanager.model.GetSecretValueResponse;
+import software.amazon.awssdk.services.secretsmanager.SecretsManagerClient;
 
 import java.nio.charset.Charset;
 import java.util.Base64;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.ExecutionException;
 
 @Slf4j
 public class AWSAuthService implements AutoCloseable {
 
-    private final SecretsManagerAsyncClient secretsManagerClient;
+    private final SecretsManagerClient secretsManagerClient;
 
-    public AWSAuthService(SecretsManagerAsyncClient secretsManagerClient) {
+    public AWSAuthService(SecretsManagerClient secretsManagerClient) {
         this.secretsManagerClient = secretsManagerClient;
     }
 
     public AWSAuthService() {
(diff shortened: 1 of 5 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/AWSAuthService.java
```diff
@@ -1,43 +1,47 @@
...
+     * This method synchronously returns a Basic Auth header string, with the username:password Base64 encoded
      * @param username The plaintext username
      * @param secretId The unique name of the secret or the secret ARN from AWS Secrets Manager. Its retrieved value
      *                 will fill the password part of the Basic Auth header
      * @return Basic Auth header string
      */
-    public String getBasicAuthHeaderFromSecret(String username, String secretId) {
-        String authHeaderString = username + ":" + getSecret(secretId);
+    public String getBasicAuthHeaderFromSecret(String username, String secretId) throws ExecutionException, InterruptedException {
+        String secretValue = getSecret(secretId).get().secretString();
```
gregschohn, line 38: I don't like having the synchronous method alongside the async one and then being opinionated on exceptions that can come out of the synchronous method that would only happen when dealing with something asynchronous.

You could also return a CompletableFuture by applying the getSecret Output through a finalizer that does everything else.  That let's the caller determine how to handle the exceptions and keeps these methods more consistent.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -255,7 +255,8 @@ private static String formatAuthArgFlagsAsString() {
                 .collect(Collectors.joining(", "));
     }
 
-    private static IAuthTransformerFactory buildAuthTransformerFactory(Parameters params) {
+    private static IAuthTransformerFactory buildAuthTransformerFactory(Parameters params)
+        throws ExecutionException, InterruptedException {
```
gregschohn, line 259: I can't imagine a case where I'd need to make these factories asynchronously.  I could see needing to create the AuthTransformers themselves asynchronously, or signing the requests - but this seems like it would always be part of bootstrapping at the top level.

lewijacn: I guess I am inclined to agree with you. Initially when I set this up I thought this might be used in the Netty pipeline so having async would be useful, but for our current use case I really only see synchronous and supporting the off case where someone's secrets has changed in the middle of a migration doesn't seem immediate. I've refactored to use a synchronous client now

lewijacn: I also included the fix for the failing test case since I don't plan to merge this to the release branch

PR #279, opened 2023-08-22
Sync devDeploy with recent Replayer auth changes

lewijacn:
### Description
Adjust devDeploy.sh with recent Replayer auth changes. Also remove additional "orchestration" logic from devDeploy surrounding setting the auth header, which will be taxing to keep up to date and probably doesn't belong there.

### Issues Resolved
None

### Testing
Manual deployment testing

### Check List

### DIFF
--- a/deployment/copilot/README.md
+++ b/deployment/copilot/README.md
@@ -39,16 +39,13 @@ Options:
 Deploy migration solution infrastructure composed of resources deployed by CDK and Copilot
 
 Options:
-  --skip-bootstrap          Skip one-time setup of installing npm package, bootstrapping CDK, and building Docker images.
-  --skip-copilot-init       Skip one-time Copilot initialization of app, environments, and services
-  --copilot-app-name        [string, default: migration-copilot] Specify the Copilot application name to use for deployment
-  --destroy-env             Destroy all CDK and Copilot CloudFormation stacks deployed, excluding the Copilot app level stack, for the given env/stage and return to a clean state.
-  --destroy-all-copilot     Destroy Copilot app and all Copilot CloudFormation stacks deployed for the given app across all regions.
-  --auth-header-value       [string, default: null] Prepared "authorization" header to provide the Replayer, i.e. Basic YWRtaW46QWRtaW4xMjMh. This will override a CDK configured FGAC master user auth header if setup
-  --aws-auth-header-user    [string, default: null] Plaintext username to provide the Replayer to construct an "authorization" header. Used in conjunction with --aws-auth-header-secret. This will override a CDK configured FGAC master user auth header if setup
-  --aws-auth-header-secret  [string, default: null] Secret ARN or Secret name from AWS Secrets Manager to provide the Replayer to construct an "authorization" header. Used in conjunction with --aws-auth-header-user. This will override a CDK configured FGAC master user auth header if setup
-  -r, --region              [string, default: us-east-1] Specify the AWS region to deploy the CloudFormation stacks and resources.
-  -s, --stage               [string, default: dev] Specify the stage name to associate with the deployed resources
+  --skip-bootstrap                      Skip one-time setup of installing npm package, bootstrapping CDK, and building Docker images.
+  --skip-copilot-init                   Skip one-time Copilot initialization of app, environments, and services
+  --copilot-app-name                    [string, default: migration-copilot] Specify the Copilot application name to use for deployment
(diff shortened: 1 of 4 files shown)


deployment/copilot/README.md
```diff
@@ -58,10 +55,11 @@ Requirements:
...
+1. If the user provides an explicit auth header option to the Replayer, such as providing a static value auth header or using a user and secret arn pair, this mechanism will be used for the auth header of outgoing requests. The options can be found as Parameters [here](../../TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java)
+2. If the devDeploy script is used and its Replayer command has not been altered (as in the case of 1.) and CDK deploys a target cluster with a configured FGAC user (see `fineGrainedManagerUserName` and `fineGrainedManagerUserSecretManagerKeyARN` CDK context options [here](../cdk/opensearch-service-migration/README.md)) or is running in demo mode (see `enableDemoAdmin` CDK context option), this user and secret arn pair will be provided in the Replay command for the Auth header
+3. If the user provides no auth header option and incoming captured requests have an auth header, this auth header will be reused for outgoing requests
+4. If the user provides no auth header option and incoming captured requests have no auth header, then no auth header will be used for outgoing requests
```
gregschohn, line 62: This should probably moved to the Replayer's README and linked from here.  That doesn't need to be done now though.

deployment/copilot/devDeploy.sh
```diff
@@ -65,21 +58,6 @@ while [[ $# -gt 0 ]]; do
...
-    --aws-auth-header-user)
-      REPLAY_AWS_AUTH_HEADER_USER="$2"
-      shift # past argument
-      shift # past value
-      ;;
-    --aws-auth-header-secret)
-      REPLAY_AWS_AUTH_HEADER_SECRET="$2"
-      shift # past argument
-      shift # past value
-      ;;
```
gregschohn, line 82: Could you have the dev-deploy script take in the arguments to use for the replayer?  Something like "--extraFlags" and then let the developer tack on any auth flags or anything else that might be necessary.  As it is, it will be painful for a dev to play around w/ different options.

And I would say that none of the arguments are checked - which should be fine if this is a "developer" script

deployment/copilot/devDeploy.sh
```diff
@@ -164,30 +132,15 @@ fi
 # More details on the CDK used here can be found at opensearch-migrations/deployment/cdk/opensearch-service-migration/README.md
 cdk deploy "*" --c domainName="aos-domain" --c engineVersion="OS_1.3" --c  dataNodeCount=2 --c vpcEnabled=true --c availabilityZoneCount=2 --c openAccessPolicyEnabled=true --c domainRemovalPolicy="DESTROY" --c migrationAssistanceEnabled=true --c enableDemoAdmin=true -O cdkOutput.json --require-approval never --concurrency 3
 
-# Gather CDK output which includes export commands needed by Copilot, and make them available to the environment
-found_exports=$(grep -o "export [a-zA-Z0-9_]*=[^\\;\"]*" cdkOutput.json)
-eval "$(grep -o "export [a-zA-Z0-9_]*=[^\\;\"]*" cdkOutput.json)"
+# Collect export commands from CDK output, which are needed by Copilot, wrap the values in quotes and make them available to the environment
```
gregschohn, line 135: nit - are you finding/gathering commands or values?

gregschohn (approved):
Please look at the comment on 'extra args'.  Approving this since it should unbreak my previous changes.

deployment/copilot/devDeploy.sh
lewijacn: I am a little conflicted with what we should do here. I don't want the additional tax of actually knowing what arguments the Replayer takes which makes something like "--extraFlags" seem preferrable, however this also becomes a balancing act of does my default command already have an argument that is being passed in with "--extraFlags", it seems it would be easier for the user to just update the command in the script to what they need.

What might be slightly convenient is to allow a user to specify all the arguments for the Replayer (not sure how useful this is) in the devDeploy script, or do something like "--replayerAuthFlags" which defaults to our normal setting but can be overriden with whatever other flags

Lmk your thoughts here

deployment/copilot/devDeploy.sh
lewijacn: Adjusted

PR #283, opened 2023-08-23
Use CRLF in NettyJsonToByteBufHandler

mikaylathompson:

### Description
This PR includes three (related) categories of changes:
1. It outputs the request that was replayed against the target cluster and the connection ID in the tuples that are logged as output.
2. It updates our terminology in the tuples & surrounding areas from primary/shadow to source/target, which matches what we're using in the rest of our code & documentation.
3. The NettyJsonToByteBufHandler was using line feeds (`\n`) when formatting HTTP messages, but [the spec is to use carriage return line feeds](https://www.ibm.com/docs/en/cics-ts/6.1?topic=protocol-http-requests#dfhtl21__title__3) (`\r\n`). This turned out to cause an issue in how we were parsing the HTTP message for the JSON, so rather than build in a workaround, I updated the code to use CRLFs.

These three categories are broken out quite neatly in the commits:
the first two commits are related to number 3, the next accomplishes 1, the third deals with 2, and then the final commit updates the humanReadableLogs script for 1 and 2.

### Issues Resolved
I haven't created an issue for this yet 😬 

### Testing
Manual, and updated the unit tests affected.

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/SourceTargetCaptureTuple.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/SourceTargetCaptureTuple.java
@@ -48,7 +48,6 @@ public class SourceTargetCaptureTuple {
         }
 
         private JSONObject jsonFromHttpDataUnsafe(List<byte[]> data) throws IOException {
-
             SequenceInputStream collatedStream = ReplayUtils.byteArraysToInputStream(data);
             Scanner scanner = new Scanner(collatedStream, StandardCharsets.UTF_8);
             scanner.useDelimiter("\r\n\r\n");  // The headers are seperated from the body with two newlines.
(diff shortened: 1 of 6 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/SourceTargetCaptureTuple.java
```diff
@@ -88,46 +87,58 @@ private JSONObject jsonFromHttpData(List<byte[]> data, Duration latency) throws
...
+            meta.put("sourceRequest", jsonFromHttpData(triple.sourcePair.requestData.packetBytes));
+            meta.put("targetRequest", jsonFromHttpData(triple.targetRequestData));
             //log.warn("TODO: These durations are not measuring the same values!");
-            meta.put("primaryResponse", jsonFromHttpData(triple.sourcePair.responseData.packetBytes,
+            meta.put("sourceResponse", jsonFromHttpData(triple.sourcePair.responseData.packetBytes,
                 Duration.between(triple.sourcePair.requestData.getLastPacketTimestamp(), triple.sourcePair.responseData.getLastPacketTimestamp())));
-            meta.put("shadowResponse", jsonFromHttpData(triple.shadowResponseData,
-                    triple.shadowResponseDuration));
+            meta.put("targetResponse", jsonFromHttpData(triple.targetResponseData,
+                    triple.targetResponseDuration));
```
gregschohn, line 96: nit: This should be named a tuple

gregschohn (approved):

PR #285, opened 2023-08-24
Replayer retry failed poll()

lewijacn:
### Description
When deploying our solution from a clean slate, the Replayer will continually fail at startup if the specified topic in Kafka is not available, with error: `Exception in thread "main" org.apache.kafka.common.errors.TopicAuthorizationException: Not authorized to access topics: [logging-traffic-topic]` . This is seemingly because the Replayer does not have permission to create a topic and the Capture Proxy will not create a topic until it actually receives some incoming requests which makes starting with our solution an unpleasant experience. This change will allow the Replayer to continue to retry these failures without shutting down and spinning up another Replayer container.

### Testing

### Check List


gregschohn:
Can the orchestrator/deployment process do this instead?

lewijacn:
> Can the orchestrator/deployment process do this instead?

Not easily right now, would be a decent amount of effort to get things linked up so this is possible. I do think there's merit though in eventually having the orchestrator be able to clear a Kafka topic or determine the number of messages on a topic, things of this sort, and maybe creating the topics also fits into this pattern

lewijacn:
@gregschohn Updated Replayer Kafka to continue to retry poll() exceptions

gregschohn (approved):
LGTM

PR #293, opened 2023-09-01
MIGRATIONS-1296 - Fix mend security issues

okhasawn:
### Description
This PR attempts to fix some of the currently existing mend security issues in the repo.
These are the issues it attempts to fix at the moment (more will follow soon):
-https://github.com/opensearch-project/opensearch-migrations/issues/251
-https://github.com/opensearch-project/opensearch-migrations/issues/247
-https://github.com/opensearch-project/opensearch-migrations/issues/246
-https://github.com/opensearch-project/opensearch-migrations/issues/226
-https://github.com/opensearch-project/opensearch-migrations/issues/227

### Issues Resolved
[Migrations-1296](https://opensearch.atlassian.net/browse/MIGRATIONS-1296)

### Check List


okhasawn:
This also fixes 2 of 3 vulnerabilities from this issue: https://github.com/opensearch-project/opensearch-migrations/issues/205.

The one left (jtidy) is still not fixed (at least still shows up as a CVE) in the latest version, as mentioned here: https://github.com/jtidy/jtidy/issues/63

okhasawn:
Also fixes 3 out of 4 vulns from issue #126 
The only one left is because we're using `bcel-6.5.0.jar`, I've updated that to use `bcel-6.7.0.jar`, which is the suggested fix, and what fixed the CVE locally. I ran `./gradlew trafficReplayer:dependencyCheckAnalyze` and it was no longer showing up compared to version 6.5.

TrafficCapture/trafficReplayer/build.gradle
```diff
@@ -55,19 +57,30 @@ dependencies {
...
 
     testImplementation project(':testUtilities')
     testImplementation group: 'org.apache.httpcomponents.client5', name: 'httpclient5', version: '5.2.1'
     testImplementation 'org.mockito:mockito-core:4.6.1'
     testImplementation 'org.mockito:mockito-junit-jupiter:4.6.1'
 }
 
+configurations.all {
+    resolutionStrategy.eachDependency { DependencyResolveDetails details ->
+        if (details.requested.group == 'org.apache.commons' && details.requested.name == 'commons-text') {
```
lewijacn, line 75: From mine and Omar's conversation earlier: I have a slight concern here about maintaining this in the future, say for instance the dependency which has this transitive dependency gets updated and in turn updates this transitive dependency to a later version than we have listed here (1.10.0). I don't want to actually prevent a later version from being used. If its a small fix we should add a conditional check here that the dependency is less than the version we want to insert, or otherwise don't override the version. If not a small fix let's raise an issue around this

okhasawn: Thanks Tanner, that's a very valid concern and I agree that it wouldn't be dev friendly to have to maintain that.
Fortunately, I just pushed a change where I'm specifying a minimum version and check if the dependency wants to pull a higher version of the transitive dependency. If that's the case, then the higher version would be used. In case the dependency was requesting a version lower than the one we're "targeting", then the targeted version would be used.

lewijacn: I was playing around with this a little bit and it seems like it is just doing string comparison and not actual version comparison. For instance if I set the target version to 1.9.0 and the incoming requested is 1.10.0, it will try to set the incoming requested to 1.9.0. Lmk if you see different results with this use case.

okhasawn: Oh interesting. 
What I tested before pushing that was with the `apache.bcel` dependency. I know that the requested version is 6.5, If I set the target to 6.4 then the 6.5 one is pulled.
I'll look into that and see if there's a different method of comparison for this.

okhasawn: I'm thinking of doing something with string parsing and compare the versions manually, I think it can be helpful in case we have more situations with transitive dependency versions in the future. What do you think?

okhasawn: Please let me know what you think of the proposed method I just pushed.

lewijacn: Yeah I think this in more in line with what we need to do. I would change the function you created to return a boolean and call it something like `isRequestedVersionOlder`

lewijacn: Few additional comments to capture in another JIRA:

1. You can make your new function `static`
2. Your logic for the function seems to be reversed of what I think of from the method name, I would expect if the requested part is less than the target part that it is an older version. Maybe this name was a bit misleading too, a bit wordy but we could add some clarity with `wasRequestedVersionReleasedBeforeTargetVersion`
3. With the previous change, can you make your if statement simply, without the boolean check
`if (isRequestedVersionOlder(details.requested.version, targetVersion))`

PR #294, opened 2023-09-05
[Fetch Migration] Adding orchestration script for Fetch Migration steps

kartg:
### Description
This change adds an "orchestration" Python script that runs the Fetch Migration steps in order
1. The pre-migration script is run first to compare the indices on source and target clusters. A human-readable report is printed, and an updated Data Prepper pipeline YAML file is written to the Data Prepper `pipelines` directory
2.  Next, Data Prepper is kicked off as a [sub-process](https://docs.python.org/3/library/subprocess.html) to migrate data based on the output pipeline YAML file
3. After the Data Prepper process is kicked off, the monitoring script is run to poll DP's Prometheus metrics endpoint and determine when migration is complete and the process can be shut down

The orchestrator script takes multiple command-line inputs to enable execution of these steps. The Dockerfile definition has also been updated to use the orchestrator script as the entrypoint.
* Category: New feature

### Issues Resolved
N/A

### Testing
```
fetch_orchestrator.py             24      0   100%
index_operations.py               35      0   100%
metadata_migration.py            123      1    99%
migration_monitor.py              59      0   100%
migration_monitor_params.py        5      0   100%
TOTAL                            277      1    99%

```

### Check List

### DIFF
--- a/FetchMigration/Dockerfile
+++ b/FetchMigration/Dockerfile
@@ -6,8 +6,8 @@ RUN apk update
 RUN apk add --no-cache python3 py-pip
 RUN pip install --user -r requirements.txt
 
-ENV ICT_CODE_PATH /code
-WORKDIR $ICT_CODE_PATH
+ENV FM_CODE_PATH /code
+WORKDIR $FM_CODE_PATH
 # Copy only source code
 COPY python/*.py .
 
@@ -15,4 +15,4 @@ COPY python/*.py .
 ENV PATH=/root/.local:$PATH
 
 # Include the -u flag to have stdout logged
-ENTRYPOINT python -u ./orchestrator.py $DATA_PREPPER_PATH $ICT_CODE_PATH/input.yaml https://localhost:4900
+ENTRYPOINT python -u ./fetch_orchestrator.py $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml https://localhost:4900
--- a/FetchMigration/Dockerfile
+++ b/FetchMigration/Dockerfile
@@ -14,5 +14,5 @@ COPY python/*.py .
 # update PATH
 ENV PATH=/root/.local:$PATH
 
-# make sure you include the -u flag to have our stdout logged
-ENTRYPOINT python -u ./main.py -r $ICT_CODE_PATH/input.yaml $DATA_PREPPER_PATH/pipelines/pipelines.yaml; $DATA_PREPPER_PATH/bin/data-prepper
+# Include the -u flag to have stdout logged
+ENTRYPOINT python -u ./orchestrator.py $DATA_PREPPER_PATH $ICT_CODE_PATH/input.yaml https://localhost:4900
--- a/FetchMigration/python/migration_monitor.py
+++ b/FetchMigration/python/migration_monitor.py
@@ -43,50 +43,58 @@ def get_metric_value(metric_families: List, metric_suffix: str) -> Optional[int]
     return None
 
 
-def check_if_complete(doc_count: Optional[int], in_flight: Optional[int], no_part_count: Optional[int],
-                      prev_no_part_count: int, target: int) -> bool:
+def check_if_complete(doc_count: Optional[int], in_flight: Optional[int], no_partition_count: Optional[int],
+                      prev_no_partition: int, target: int) -> bool:
     # Check for target doc_count
     # TODO Add a check for partitionsCompleted = indices
     if doc_count is not None and doc_count >= target:
         # Check for idle pipeline
         logging.info("Target doc count reached, checking for idle pipeline...")
-        logging.debug("Idle pipeline metrics: " +
-                      ",".join(map(str, [in_flight, no_part_count, prev_no_part_count])))
+        # Debug metrics
+        if logging.getLogger().isEnabledFor(logging.DEBUG):  # pragma no cover
+            debug_msg_template: str = "Idle pipeline metrics - " + \
+                "Records in flight: [{0}], " + \
+                "No-partitions counter: [{1}]" + \
+                "Previous no-partition value: [{2}]"
+            logging.debug(debug_msg_template.format(in_flight, no_partition_count, prev_no_partition))
+
         if in_flight is not None and in_flight == 0:
             # No-partitions metrics should steadily tick up
-            if no_part_count is not None and no_part_count > prev_no_part_count > 0:
+            if no_partition_count is not None and no_partition_count > prev_no_partition > 0:
                 return True
     return False
 
 
-def run(args: MigrationMonitorParams, wait_seconds: int = 30) -> None:
+def run(args: MigrationMonitorParams, poll_interval_seconds: int = 30) -> None:
     # TODO Remove hardcoded EndpointInfo
     default_auth = ('admin', 'admin')
-    endpoint = EndpointInfo(args.dp_endpoint, default_auth, False)
+    endpoint = EndpointInfo(args.data_prepper_endpoint, default_auth, False)
     target_doc_count: int = args.target_count
     prev_no_partitions_count = 0
     terminal = False
     logging.info("Starting migration monitor until target doc count: " + str(target_doc_count))
     while not terminal:
-        time.sleep(wait_seconds)
+        time.sleep(poll_interval_seconds)
         # If the API call fails, the response is empty
         metrics = fetch_prometheus_metrics(endpoint)
         if metrics is not None:
             success_docs = get_metric_value(metrics, __DOC_SUCCESS_METRIC)
             rec_in_flight = get_metric_value(metrics, __RECORDS_IN_FLIGHT_METRIC)
             no_partitions_count = get_metric_value(metrics, __NO_PARTITIONS_METRIC)
-            if success_docs is not None:
+            if success_docs is not None:  # pragma no cover
                 completion_percentage: int = math.floor((success_docs * 100) / target_doc_count)
                 progress_message: str = "Completed " + str(success_docs) + \
                                         " docs ( " + str(completion_percentage) + "% )"
                 logging.info(progress_message)
+            else:
+                logging.info("Could not fetch metrics from Data Prepper, will retry on next polling cycle...")
             terminal = check_if_complete(success_docs, rec_in_flight, no_partitions_count,
                                          prev_no_partitions_count, target_doc_count)
             if not terminal:
                 # Save no_partitions_count
                 prev_no_partitions_count = no_partitions_count
     # Loop terminated, shut down the Data Prepper pipeline
-    logging.info("Migration monitor at terminal state, shutting down...\n")
+    logging.info("Migration monitor observed successful migration and idle pipeline, shutting down...\n")
     shutdown_pipeline(endpoint)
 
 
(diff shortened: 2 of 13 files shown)


FetchMigration/Dockerfile
```diff
@@ -14,5 +14,5 @@ COPY python/*.py .
 # update PATH
 ENV PATH=/root/.local:$PATH
 
-# make sure you include the -u flag to have our stdout logged
-ENTRYPOINT python -u ./main.py -r $ICT_CODE_PATH/input.yaml $DATA_PREPPER_PATH/pipelines/pipelines.yaml; $DATA_PREPPER_PATH/bin/data-prepper
+# Include the -u flag to have stdout logged
+ENTRYPOINT python -u ./orchestrator.py $DATA_PREPPER_PATH $ICT_CODE_PATH/input.yaml https://localhost:4900
```
gregschohn, line 18: Since this will be situated within something else (which may also be situated within something else, because you never know) - you might want to disambiguate "orchestrator" with "fetch" (e.g. fetchOrchestrator.py)

What does ICT stand for?

FetchMigration/python/migration_monitor.py
```diff
@@ -47,6 +49,9 @@ def check_if_complete(doc_count: Optional[int], in_flight: Optional[int], no_par
     # TODO Add a check for partitionsCompleted = indices
     if doc_count is not None and doc_count >= target:
         # Check for idle pipeline
+        logging.info("Target doc count reached, checking for idle pipeline...")
+        logging.debug("Idle pipeline metrics: " +
+                      ",".join(map(str, [in_flight, no_part_count, prev_no_part_count])))
```
gregschohn, line 54: consider prefacing these values w/ labels.
Second question - what does it mean to be inFlight and what does "no_part" mean?  From looking through the existing code, my bet is that it's sortof like a reverse countdown - something like how many of the original partitions no longer need to be moved?  Consider a clearer name to self document this field.

FetchMigration/python/migration_monitor.py
```diff
@@ -58,24 +63,30 @@ def run(args: MigrationMonitorParams, wait_seconds: int = 30) -> None:
     # TODO Remove hardcoded EndpointInfo
     default_auth = ('admin', 'admin')
     endpoint = EndpointInfo(args.dp_endpoint, default_auth, False)
+    target_doc_count: int = args.target_count
     prev_no_partitions_count = 0
     terminal = False
+    logging.info("Starting migration monitor until target doc count: " + str(target_doc_count))
     while not terminal:
+        time.sleep(wait_seconds)
```
gregschohn, line 71: Consider renaming this to be more descriptive for the user.  e.g. poll_interval_seconds

FetchMigration/python/migration_monitor.py
```diff
@@ -58,24 +63,30 @@ def run(args: MigrationMonitorParams, wait_seconds: int = 30) -> None:
...
         # If the API call fails, the response is empty
         metrics = fetch_prometheus_metrics(endpoint)
         if metrics is not None:
             success_docs = get_metric_value(metrics, __DOC_SUCCESS_METRIC)
             rec_in_flight = get_metric_value(metrics, __RECORDS_IN_FLIGHT_METRIC)
             no_partitions_count = get_metric_value(metrics, __NO_PARTITIONS_METRIC)
+            if success_docs is not None:
```
gregschohn, line 78: Is there a reason NOT to log when there are no successful docs migrated?  It feels like there's a hole in the logging story w/out that too, especially if things get really messed up.

FetchMigration/python/migration_monitor.py
```diff
@@ -58,24 +63,30 @@ def run(args: MigrationMonitorParams, wait_seconds: int = 30) -> None:
...
-                                         prev_no_partitions_count, args.target_count)
+                                         prev_no_partitions_count, target_doc_count)
             if not terminal:
                 # Save no_partitions_count
                 prev_no_partitions_count = no_partitions_count
-
-        if not terminal:
-            time.sleep(wait_seconds)
     # Loop terminated, shut down the Data Prepper pipeline
+    logging.info("Migration monitor at terminal state, shutting down...\n")
```
gregschohn, line 89: Would be nice to state whether it was successful or not here.

FetchMigration/python/migration_monitor_params.py
```diff
@@ -4,4 +4,4 @@
 @dataclass
 class MigrationMonitorParams:
     target_count: int
-    dp_endpoint: str = "https://localhost:4900"
+    dp_endpoint: str
```
gregschohn, line 7: consider data_prepper_endpoint.  Readers might assume that you meant "data plane".

FetchMigration/python/orchestrator.py
```diff
@@ -0,0 +1,60 @@
...
+from pre_migration_params import PreMigrationParams
+
+
+__DP_EXECUTABLE_SUFFIX = "/bin/data-prepper"
+__PIPELINE_OUTPUT_FILE_SUFFIX = "/pipelines/pipeline.yaml"
+
+
+def run(dp_base_path: str, dp_config_file: str, dp_endpoint: str):
+    dp_exec_path = dp_base_path + __DP_EXECUTABLE_SUFFIX
+    output_file = dp_base_path + __PIPELINE_OUTPUT_FILE_SUFFIX
```
gregschohn, line 18: what will this output contain?  Will it be a report, logs?  It's fair for this component to leave that to DataPrepper if it's just stdout or labeled as output, but I'm still curious (so that as a reviewer, I can think about how the output needs to be shared/secured).

FetchMigration/python/orchestrator.py
```diff
@@ -0,0 +1,60 @@
...
+    pre_migration_params = PreMigrationParams(dp_config_file, output_file, report=True)
+    logging.info("Running pre-migration steps...\n")
+    pre_migration_result = pre_migration.run(pre_migration_params)
```
gregschohn, line 21: Please come up with a name that is more descriptive.  Everything is a migration, including the "pre" steps in many cases.  Say something like "index_config_migration" or "detecting_configurations", etc.

FetchMigration/python/orchestrator.py
```diff
@@ -0,0 +1,60 @@
...
+    if len(pre_migration_result.created_indices) > 0:
+        # Kick off a subprocess for Data Prepper
+        logging.info("Running Data Prepper...\n")
```
gregschohn, line 24: Do you want to print out the args that you're sending to it too?

FetchMigration/python/orchestrator.py
```diff
@@ -0,0 +1,60 @@
...
+        proc = subprocess.Popen(dp_exec_path)
+        # Data Prepper started successfully, run the migration monitor
```
gregschohn, line 26: how do you know that it was successful?  What happens when the code isn't successful?  I'm presuming that it throws, but I don't see a catch block that would explain to the user what has happened or what course of action to take next.

FetchMigration/python/orchestrator.py
```diff
@@ -0,0 +1,60 @@
...
+        migration_monitor_params = MigrationMonitorParams(pre_migration_result.target_doc_count, dp_endpoint)
+        logging.info("Starting migration monitor...\n")
+        migration_monitor.run(migration_monitor_params)
+        # Migration ended, the following is a workaround for
+        # https://github.com/opensearch-project/data-prepper/issues/3141
+        if proc.returncode is None:
+            proc.terminate()
```
gregschohn, line 33: What signal does this send?  I'm hoping that its TERM.  If it doesn't respond to TERM w/in N seconds (maybe 10 or 30), it seems like sending a KILL (-9) is appropriate - though you might want to check that it didn't find more data in the interim and has begun to migrate it, if that's a possibility.  
Generally, this is going to be a difficult concurrency problem to solve and should be solved mostly in DataPrepper (you send a signal, it cleans up and handles its own health, data and consistency)

gregschohn (commented):
Lots of naming suggestions & I have some concerns about corner cases around the process management of DataPrepper, some of which are probably outside the bounds of this PR.

FetchMigration/Dockerfile
kartg: Renamed `orchestrator.py` to `fetch_orchestrator.py`

ICT = Index Configuration Tool. This is left over from when the tool would only perform index configuration rather than the full fetch migration workflow. Changed references to `ICT` to `FM` instead.

FetchMigration/python/migration_monitor.py
kartg: * `in_flight` is a [counter published by Data Prepper](https://github.com/opensearch-project/data-prepper/blob/main/docs/monitoring.md#plugin-metrics) to describe the number of records in memory that have been pulled from the source plugin and are currently being processed by the sink plugin
* The `no_part` variables track the number of times the Data Prepper source coordination implementation attempted to fetch a  partition to move data for and found none pending. This is our indicator that data migration has finished and the DP pipeline is idle. The DP maintainers advised us to the track the _rate of change_ of this value rather than just a non-zero value, hence the `prev` counter.

Incorporated PR feedback by adding labels and renaming `no_part` variables to `no_partition` for clarity.

FetchMigration/python/migration_monitor.py
kartg: Reworded to `Migration monitor observed successful migration and idle pipeline, shutting down...`

FetchMigration/python/orchestrator.py
kartg: `output_file` is the path to which `pre_migration.py` will write a modified Data Prepper pipeline YAML file after comparing indices between the source and target clusters. This pipeline YAML file ensures that only indices that are safe to migrate are created on the target cluster (and eventually migrated). The Data Prepper executable uses this `output_file` as its pipeline configuration.

FetchMigration/python/orchestrator.py
kartg: Added an item to the backlog to improve this - https://opensearch.atlassian.net/browse/MIGRATIONS-1331

FetchMigration/python/orchestrator.py
kartg: > What signal does this send? I'm hoping that its TERM

`terminate` [sends SIGTERM](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.terminate)

> .... and should be solved mostly in DataPrepper (you send a signal, it cleans up and handles its own health, data and consistency)

Data Prepper handles this already.

> If it doesn't respond to TERM w/in N seconds (maybe 10 or 30), it seems like sending a KILL (-9) is appropriate - though you might want to check that it didn't find more data in the interim and has begun to migrate it, if that's a possibility.

Added an item to the backlog to track this improvement - https://opensearch.atlassian.net/browse/MIGRATIONS-1331
Finding more data is not a concern since the source plugin will use scroll / PIT queries against the source

FetchMigration/python/orchestrator.py
kartg: The Data Prepper process consumes its pipeline configuration from a file, which could contain credential information. I don't think we should output the contents of this file to the logs.

FetchMigration/python/orchestrator.py
kartg: Changed the notion of "pre-migration" to "metadata migration" throughout

gregschohn (approved):

PR #297, opened 2023-09-07
Add MetricsLogger to the Capture Proxy

mikaylathompson:
### Description
Adds a MetricsLogger to the Capture Proxy. For now, it creates a limited number of events, with a limited number of dimensions.

Events:
- Component of request received (logged on `LoggingHttpRequestHandler::channelRead`)
- Full request received (logged on `LoggingHttpRequestHandler::channelFinishedReadingAnHttpMessage`)
- Sending message to Kafka (logged on `KafkaCaptureFactory::createOffloader` in the handler for a completed CodedOutputStream)
- Message to Kafka failed (same as above)
- Component of a response received (logged on `LoggingHttpResponseHandler::write`)

All messages are logged with the `channelId` which corresponds to the `connectionId`/`diagnosticId` used throughout the Replayer. Some messages have additional context info, e.g. full request received also has the http method & endpoint.

The event I would like to include but wasn't able to find already existing in our code is when a full response is received. I don't think this is an essential enough event that it's worth adding a HTTP accumulator to the pipeline, but if I missed a spot where this message is accumulated already, I'd like to add it.

Events are logged in a json format (see example below). After playing with various text based log formats, this was by far the cleanest way to extract components as an AWS Metrics Filter, and should give us flexibility in the docker solution as well.
To test this out, I created a metrics filter that looked specifically for "Full request received" events and plotted them while I did two `./runTestBenchmarks.sh` runs.

Screen Shot 2023-09-06 at 11 08 10 PM

Example logs:
This shows MetricsLogger events mixed in with default logger events from the docker solution:
```
[DEBUG] 2023-09-07 04:33:27.088 [nioEventLoopGroup-3-1] FrontsideHandler - channelRead COMPLETE
{"instant":{"epochSecond":1694061092,"nanoOfSecond":224314749},"thread":"nioEventLoopGroup-3-1","level":"INFO","loggerName":"MetricsLogger","message":"Component of response received","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","contextMap":{"channelId":"0242c0fffea83007-0000000d-00000003-cb2c2632a520d48e-8e3efc89","topic":"logging-traffic-topic"},"threadId":17,"threadPriority":10}
{"instant":{"epochSecond":1694061092,"nanoOfSecond":242391443},"thread":"nioEventLoopGroup-3-1","level":"INFO","loggerName":"MetricsLogger","message":"Sending message to Kafka","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","contextMap":{"channelId":"0242c0fffea83007-0000000d-00000003-cb2c2632a520d48e-8e3efc89","diagnosticId":"0242c0fffea83007-0000000d-00000003-cb2c2632a520d48e-8e3efc89.1","topic":"logging-traffic-topic"},"threadId":17,"threadPriority":10}
{"instant":{"epochSecond":1694061207,"nanoOfSecond":85524207},"thread":"nioEventLoopGroup-3-1","level":"INFO","loggerName":"MetricsLogger","message":"Component of request received","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","contextMap":{"channelId":"0242c0fffea83007-0000000d-00000005-6aed442019229d30-21ba833a","topic":"logging-traffic-topic"},"threadId":17,"threadPriority":10}
{"instant":{"epochSecond":1694061207,"nanoOfSecond":87228901},"thread":"nioEventLoopGroup-3-1","level":"INFO","loggerName":"MetricsLogger","message":"Full request received","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","contextMap":{"channelId":"0242c0fffea83007-0000000d-00000005-6aed442019229d30-21ba833a","httpEndpoint":"/_cat/nodes","httpMethod":"GET","topic":"logging-traffic-topic"},"threadId":17,"threadPriority":10}
{"instant":{"epochSecond":1694061207,"nanoOfSecond":129397845},"thread":"nioEventLoopGroup-3-1","level":"INFO","loggerName":"MetricsLogger","message":"Component of response received","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","contextMap":{"channelId":"0242c0fffea83007-0000000d-00000005-6aed442019229d30-21ba833a","topic":"logging-traffic-topic"},"threadId":17,"threadPriority":10}
{"instant":{"epochSecond":1694061207,"nanoOfSecond":134681378},"thread":"nioEventLoopGroup-3-1","level":"INFO","loggerName":"MetricsLogger","message":"Sending message to Kafka","endOfBatch":false,"loggerFqcn":"org.apache.logging.slf4j.Log4jLogger","contextMap":{"channelId":"0242c0fffea83007-0000000d-00000005-6aed442019229d30-21ba833a","diagnosticId":"0242c0fffea83007-0000000d-00000005-6aed442019229d30-21ba833a.1","topic":"logging-traffic-topic"},"threadId":17,"threadPriority":10}
[DEBUG] 2023-09-07 04:33:27.132 [nioEventLoopGroup-3-1] FrontsideHandler - channelRead COMPLETE
```

### Issues Resolved
It won't be fully resolved, but this is the first half of https://opensearch.atlassian.net/browse/MIGRATIONS-1299.

This should be backported to the capture-and-replay-v0.1.0 branch.

### Testing
Manual testing.

### Check List


gregschohn (commented):
I'd like to see events for backside connections.  Connections created, closed & exceptions.  Having the timestamps for those will help us understand the general health of the source cluster too.

The capture proxy doesn't do any parsing of the request coming back, so we can't tell when it is done.  We could add minimal parsing of the first line w/out taking much of a performance hit at all to get the status code, but going through all of the payload stream with netty's http components would be much more expensive.  We _used_ to do that, but it isn't critical for the logged traffic and is another risk/performance hit on something that we're trying to keep as simple as possible.

As a dev, I don't think that I'll ever get comfortable having lines with different formats (JSON, vs. delimited).  I strongly prefer to see them 'blend-in' more with the rest of the context, but be clearly distinguishable (some logs use color, though I'm happy w/ just a different logger name).

MDC uses ThreadLocal.  Since netty is popping work on/off the same threads many times, ThreadLocals are often brutally difficult to get right.  It's super easy to not clear something or to recurse and smash old values.  The current codebase uses two threadlocals.  One is to handle formatting in a toString call (and the threadlocal values are pushed/popped and restore the old value) and the other is to maintain a per-thread cache of our connections since channels are bound to Threads (so we shouldn't try to move them between threads).

This doesn't have any tests associated with it.  Given the criticality of these metrics to the general health and developer experience, I'd like to see some mild integration tests run w/ every build.  Setting up a Log4j2 context specifically for your test and using a SimpleHttpServer could test that you're getting the right counts and values within about a second for even complicated cases.  This can be follow-up work, but it should still be considered a high-priority.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -8,16 +8,22 @@
 import org.opensearch.migrations.trafficcapture.IChannelConnectionCaptureSerializer;
 import org.opensearch.migrations.trafficcapture.IConnectionCaptureFactory;
 import org.opensearch.migrations.trafficcapture.StreamChannelConnectionCaptureSerializer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
```
gregschohn, line 13: See the comment at the top-level.  Since MDC is handled as a ThreadLocal, this is going to be tech-debt that is extremely difficult to track and diagnose.  Case in point, there aren't any tests in place to verify that the right values are going to be coming back out into the logs.

From a logging config management perspective, the MDC key/values can be handled fluently.  See https://chat.openai.com/share/08649025-20e3-42e6-abba-32c29722dc0c and https://github.com/opensearch-project/opensearch-migrations/pull/280

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -65,13 +74,18 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
                     // Used to essentially wrap Future returned by Producer to CompletableFuture
                     CompletableFuture cf = new CompletableFuture<>();
                     log.debug("Sending Kafka producer record: {} for topic: {}", recordId, topicNameForTraffic);
+                    MDC.put("diagnosticId", recordId);
+                    metricsLogger.info("Sending message to Kafka");
+                    MDC.remove("diagnosticId");
```
gregschohn, line 79: This is error prone from a human perspective and an exception-safe one.  It's also far less efficient than just putting it into the logging line.  "diagnosticId" is going to be used by this thread in potentially many (global) places, and you're not preserving the old value.  This is a bad pattern.  Globals are bad and that's what these are.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -55,6 +61,9 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
                 return cos;
             },
             (captureSerializerResult) -> {
+                // Structured context for MetricsLogger
+                MDC.put("channelId",connectionId);
+                MDC.put("topic", topicNameForTraffic);
```
gregschohn, line 66: Notice that you didn't clear these out or put back their old values.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -65,13 +74,18 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
...
+                    MDC.remove("diagnosticId");
                     // Async request to Kafka cluster
                     producer.send(record, handleProducerRecordSent(cf, recordId));
                     // Note that ordering is not guaranteed to be preserved here
                     // A more desirable way to cut off our tree of cf aggregation should be investigated
                     singleAggregateCfRef[0] = singleAggregateCfRef[0].isDone() ? cf : CompletableFuture.allOf(singleAggregateCfRef[0], cf);
                     return singleAggregateCfRef[0];
                 } catch (Exception e) {
+                    metricsLogger.warn("[connectionId: {0}, topic: {1}] Sending message to Kafka", connectionId, topicNameForTraffic);
+                    metricsLogger.warn("Error: ", e);
```
gregschohn, line 88: Can you codify what the log levels are going to be for?
My thought is that this should be ERROR.  The info level seems appropriate for successes.  I would wrap these decisions in a helper class though - see the comments at the top of the file.

Secondly, wouldn't this be better as one message?  I would format the message with `atError().setCause(e).setMessage(()->"...")` (personally, I find that positional printf style format statements are more error prone & prefer the string concatenation approach, but you could use String.format() or .setArgument() on the LoggingEventBuilder object returned by atError())

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -65,13 +74,18 @@ public IChannelConnectionCaptureSerializer createOffloader(String connectionId)
                     // Used to essentially wrap Future returned by Producer to CompletableFuture
                     CompletableFuture cf = new CompletableFuture<>();
                     log.debug("Sending Kafka producer record: {} for topic: {}", recordId, topicNameForTraffic);
+                    MDC.put("diagnosticId", recordId);
+                    metricsLogger.info("Sending message to Kafka");
```
gregschohn, line 78: Please send the size of the record too

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -8,16 +8,22 @@
...
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.WeakHashMap;
 import java.util.concurrent.CompletableFuture;
 
+
 @Slf4j
 public class KafkaCaptureFactory implements IConnectionCaptureFactory {
 
+    private static final Logger metricsLogger = LoggerFactory.getLogger("MetricsLogger");
```
gregschohn, line 25: Please canonicalize this within code.  A MetricsLogger class can return WrappedLogger objects that return LoggingEventBuilder objects for errorEvents, successEvents, and traceEvents (if we ever think that those are useful).  Then you can document what the log levels (trace, debug, info, warn, error) mean.  You can also make sure that the loggers are all named something consistent.  "MetricLogger.KafkaCaptureFactory" would make sense for MetricLogger.getLogger("KafkaCaptureFactory").

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -15,13 +15,18 @@
...
+import org.slf4j.LoggerFactory;
+import org.slf4j.MDC;
 
 import java.io.IOException;
 import java.time.Instant;
 import java.util.List;
 
 @Slf4j
 public class LoggingHttpRequestHandler extends ChannelInboundHandlerAdapter {
+    private static final Logger metricsLogger = LoggerFactory.getLogger("MetricsLogger");
```
gregschohn, line 28: see comment above about canonicalizing this

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -126,6 +131,12 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
     }
 
     protected void channelFinishedReadingAnHttpMessage(ChannelHandlerContext ctx, Object msg, HttpRequest httpRequest) throws Exception {
+        MDC.put("channelId",ctx.channel().id().asLongText());
+        MDC.put("httpMethod", httpRequest.method().toString());
+        MDC.put("httpEndpoint", httpRequest.uri());
+        metricsLogger.info("Full request received");
+        MDC.remove("httpMethod");
+        MDC.remove("httpEndpoint");
```
gregschohn, line 139: Did you intentionally not remove the channelId?  There's no reason to here either (& when there is, it probably won't work right or be testable - so it's likely never going to be valuable for us)

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -126,6 +131,12 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
     }
 
     protected void channelFinishedReadingAnHttpMessage(ChannelHandlerContext ctx, Object msg, HttpRequest httpRequest) throws Exception {
+        MDC.put("channelId",ctx.channel().id().asLongText());
+        MDC.put("httpMethod", httpRequest.method().toString());
+        MDC.put("httpEndpoint", httpRequest.uri());
+        metricsLogger.info("Full request received");
```
gregschohn, line 137: I'd like to see the # of bytes that came in.  I'm not sure how easy it is to recover that here (we just need to accumulate the count from the ByteBufs though).

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/BacksideHandler.java
```diff
@@ -4,6 +4,8 @@
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import lombok.extern.slf4j.Slf4j;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
gregschohn, line 8: dead lines?

TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
```diff
@@ -8,4 +8,18 @@ appender.console.layout.pattern = [%-5level] %d{yyyy-MM-dd HH:mm:ss.SSS} [%t] %c
 
 rootLogger.level = info
 rootLogger.appenderRefs = stderr
-rootLogger.appenderRef.stderr.ref = STDERR
\ No newline at end of file
+rootLogger.appenderRef.stderr.ref = STDERR
+
+appender.metrics.type = Console
+appender.metrics.name = METRICS
+appender.metrics.layout.type = JsonLayout
```
gregschohn, line 15: Please don't set this to JsonLayout by default?  I'd rather have a user turn this on to work with their metrics consumers rather than relegate humans to be second-class citizens of their logs.

gregschohn (approved):

gregschohn (approved):
I like the approach where I can see everything going into each record!

PR #300, opened 2023-09-07
Add set of minor improvements to our IaC and Replayer story

lewijacn:
### Description
This includes several minor improvements to our IaC story for deploying multiple replayers that I have come across while deploying and testing our solution.

One additional point I want to make is that setting our log4j2.properties for the tuple directory at program startup before logging gets initialized was more involved than I originally thought. I am still open to going that direction but didn't want to get hung up here. This instead uses an env variable that seems to work well for the near term

### Issues Resolved
None

### Testing
E2E Cloud Testing

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
+++ b/TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
@@ -1,5 +1,7 @@
 status = error
 
+property.tupleDir = /shared-replayer-output/${env:TUPLE_DIR_NAME}
+
 appender.console.type = Console
 appender.console.name = STDERR
 appender.console.target = SYSTEM_ERR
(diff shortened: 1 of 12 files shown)


TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
```diff
@@ -1,5 +1,7 @@
 status = error
 
+property.tupleDir = /shared-replayer-output/${env:TUPLE_DIR_NAME}
```
gregschohn, line 3: Would we be able to make this whole thing an environment variable, with a default, so that it works on developer boxes?

deployment/copilot/templates/traffic-replayer/manifest.yml
```diff
@@ -33,6 +33,10 @@ memory: 4096    # Amount of memory in MiB used by the task.
 count: 1       # Number of tasks that should be running in your service.
 exec: true     # Enable getting a shell to your container (https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-exec.html).
 
+# Pass environment variables as key value pairs.
+variables:
+  TUPLE_DIR_NAME: SERVICE_NAME_PLACEHOLDER
```
gregschohn, line 38: See the comment up above - this could just be set to include the root name too (& we can default the env variable in log4j to just ./logs/ or something like that)

gregschohn (approved):

TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
lewijacn: Great idea, I have adjusted so there is a default here if the env variable is not found

PR #308, opened 2023-09-10
Add MetricsLogger to Traffic Replayer

mikaylathompson:
### Description
Based on PR #297, which establishes a MetricsLogger convention and inserts metrics logging in various places in the capture proxy, this PR inserts metric logging in the Replayer.

In a fully successful replay of a given request, the events seen, with their contextual information are:
- Parsed traffic stream from Kafka (connectionId, sizeInBytes, topic name)
- Full captured source request was accumulated (requestId)
- Full captured source response was accumulated (requestId)
- Captured request parsed to HTTP (requestId, httpEndpoint, httpMethod)
- Request was transformed (channelId, requestId)
- Request scheduled to be sent (requestId, delayFromOriginalToScheduledStartInMs, scheduledStartTime)
- Request packet written to target (requestId, size, channelId)
- Component of response received (channelId, size)
- Full response received (channelId, httpStatus)

There are also a number of "something went wrong" events:
- Failed to write request packet (channelId, requestId)
- Request failed to be transformed (channelId, requestId)
- Failed to parse traffic stream from Kafka (topic-name)
- Failed to receive full response (channelId)
- Failed to receive component of response (channelId)

There's a _lot_ of room for improvement here, but this should give us many of the basics for being able to track the overall flow of events.

**Note about requestId:**
This is the connectionId + index of request that's used throughout the code as a label/identifier for captured requests/responses. It's also called the requestKey & diagnosticLabel. I decided to standardize on requestId so that it's easier for users to track a given piece of data through their system, and a little more descriptive than "diagnosticLabel". I'm open to switching this if there are strong opinions.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1299

### Testing
Manual. 

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
@@ -243,8 +243,8 @@ public class CapturedTrafficToHttpTransactionAccumulator {
         assert accumulation.state == Accumulation.State.NOTHING_SENT : "state == " + accumulation.state;
         var requestPacketBytes = accumulation.rrPair.requestData;
         metricsLogger.atSuccess()
-                .addKeyValue("diagnosticLabel", accumulation.getRequestId())
-                .setMessage("Full captured request was accumulated").log();
+                .addKeyValue("requestId", accumulation.getRequestId())
+                .setMessage("Full captured source request was accumulated").log();
         if (requestPacketBytes != null) {
             requestHandler.accept(accumulation.getRequestId(), requestPacketBytes);
             accumulation.state = Accumulation.State.REQUEST_SENT;
@@ -260,8 +260,8 @@ public class CapturedTrafficToHttpTransactionAccumulator {
     private void handleEndOfResponse(Accumulation accumulation) {
         assert accumulation.state == Accumulation.State.REQUEST_SENT;
         metricsLogger.atSuccess()
-                .addKeyValue("diagnosticLabel", accumulation.getRequestId())
-                .setMessage("Full captured response was accumulated").log();
+                .addKeyValue("requestId", accumulation.getRequestId())
+                .setMessage("Full captured source response was accumulated").log();
         fullDataHandler.accept(accumulation.rrPair);
         accumulation.resetForNextRequest();
     }
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
@@ -244,6 +244,7 @@ public class CapturedTrafficToHttpTransactionAccumulator {
         var requestPacketBytes = accumulation.rrPair.requestData;
         metricsLogger.atSuccess()
                 .addKeyValue("requestId", accumulation.getRequestId())
+                .addKeyValue("connectionId", accumulation.getRequestId().connectionId)
                 .setMessage("Full captured source request was accumulated").log();
         if (requestPacketBytes != null) {
             requestHandler.accept(accumulation.getRequestId(), requestPacketBytes);
@@ -261,6 +262,7 @@ public class CapturedTrafficToHttpTransactionAccumulator {
         assert accumulation.state == Accumulation.State.REQUEST_SENT;
         metricsLogger.atSuccess()
                 .addKeyValue("requestId", accumulation.getRequestId())
+                .addKeyValue("connectionId", accumulation.getRequestId().connectionId)
                 .setMessage("Full captured source response was accumulated").log();
         fullDataHandler.accept(accumulation.rrPair);
         accumulation.resetForNextRequest();
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
@@ -1,6 +1,7 @@
 package org.opensearch.migrations.replay;
 
 import lombok.extern.slf4j.Slf4j;
+import org.opensearch.migrations.coreutils.MetricsLogger;
 import org.opensearch.migrations.replay.datatypes.UniqueRequestKey;
 import org.opensearch.migrations.replay.traffic.expiration.BehavioralPolicy;
 import org.opensearch.migrations.replay.traffic.expiration.ExpiringTrafficStreamMap;
@@ -56,6 +57,8 @@ public class CapturedTrafficToHttpTransactionAccumulator {
     private final AtomicInteger exceptionConnectionCounter = new AtomicInteger();
     private final AtomicInteger connectionsExpiredCounter = new AtomicInteger();
     private final AtomicInteger requestsTerminatedUponAccumulatorCloseCounter = new AtomicInteger();
+
+    private final static MetricsLogger metricsLogger = new MetricsLogger("CapturedTrafficToHttpTransactionAccumulator");
 
     public CapturedTrafficToHttpTransactionAccumulator(Duration minTimeout,
                                                        BiConsumer<UniqueRequestKey,HttpMessageAndTimestamp> requestReceivedHandler,
@@ -239,6 +242,9 @@ public class CapturedTrafficToHttpTransactionAccumulator {
     private boolean handleEndOfRequest(Accumulation accumulation) {
         assert accumulation.state == Accumulation.State.NOTHING_SENT : "state == " + accumulation.state;
         var requestPacketBytes = accumulation.rrPair.requestData;
+        metricsLogger.atSuccess()
+                .addKeyValue("diagnosticLabel", accumulation.getRequestId())
+                .setMessage("Full captured request was accumulated").log();
         if (requestPacketBytes != null) {
             requestHandler.accept(accumulation.getRequestId(), requestPacketBytes);
             accumulation.state = Accumulation.State.REQUEST_SENT;
@@ -253,6 +259,9 @@ public class CapturedTrafficToHttpTransactionAccumulator {
 
     private void handleEndOfResponse(Accumulation accumulation) {
         assert accumulation.state == Accumulation.State.REQUEST_SENT;
+        metricsLogger.atSuccess()
+                .addKeyValue("diagnosticLabel", accumulation.getRequestId())
+                .setMessage("Full captured response was accumulated").log();
         fullDataHandler.accept(accumulation.rrPair);
         accumulation.resetForNextRequest();
     }
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
@@ -133,6 +133,7 @@ public class ReplayEngine {
         var interval = numPackets > 1 ? Duration.between(start, end).dividedBy(numPackets-1) : Duration.ZERO;
         metricsLogger.atSuccess()
                 .addKeyValue("requestId", requestKey.toString())
+                .addKeyValue("connectionId", requestKey.connectionId)
                 .addKeyValue("delayFromOriginalToScheduledStartInMs", Duration.between(originalStart, start).toMillis())
                 .addKeyValue("scheduledStartTime", start.toString())
                 .setMessage("Request scheduled to be sent").log();
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
@@ -132,8 +132,8 @@ public class ReplayEngine {
         var end = timeShifter.transformSourceTimeToRealTime(originalEnd);
         var interval = numPackets > 1 ? Duration.between(start, end).dividedBy(numPackets-1) : Duration.ZERO;
         metricsLogger.atSuccess()
-                .addKeyValue("requestKey", requestKey.toString())
-                .addKeyValue("delayFromOriginalToScheduledStart", Duration.between(originalStart, start).toString())
+                .addKeyValue("requestId", requestKey.toString())
+                .addKeyValue("delayFromOriginalToScheduledStartInMs", Duration.between(originalStart, start).toMillis())
                 .addKeyValue("scheduledStartTime", start.toString())
                 .setMessage("Request scheduled to be sent").log();
         var sendResult = networkSendOrchestrator.scheduleRequest(requestKey, start, interval, packets);
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
@@ -3,6 +3,7 @@ package org.opensearch.migrations.replay;
 import io.netty.buffer.ByteBuf;
 import io.netty.util.concurrent.ScheduledFuture;
 import lombok.extern.slf4j.Slf4j;
+import org.opensearch.migrations.coreutils.MetricsLogger;
 import org.opensearch.migrations.replay.datatypes.UniqueRequestKey;
 import org.opensearch.migrations.replay.util.DiagnosticTrackableCompletableFuture;
 
@@ -25,6 +26,7 @@ public class ReplayEngine {
     private final AtomicLong lastIdleUpdatedTimestampEpochMs;
     private final TimeShifter timeShifter;
     private final double maxSpeedupFactor;
+    private static final MetricsLogger metricsLogger = new MetricsLogger("ReplayEngine");
     /**
      * If this proves to be a contention bottleneck, we can move to a scheme with ThreadLocals
      * and on a scheduled basis, submit work to each thread to find out if they're idle or not.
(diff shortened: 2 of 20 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -239,6 +242,9 @@ private boolean  rotateAccumulationOnReadIfNecessary(String connectionId, Accumu
     private boolean handleEndOfRequest(Accumulation accumulation) {
         assert accumulation.state == Accumulation.State.NOTHING_SENT : "state == " + accumulation.state;
         var requestPacketBytes = accumulation.rrPair.requestData;
+        metricsLogger.atSuccess()
+                .addKeyValue("requestId", accumulation.getRequestId())
```
gregschohn, line 246: This is a UniqueRequestKey.  How will the composite object be emitted?  This feels like you REALLY should be doing this with a connectionId key rather than something truly unique.  I'm expecting questions like, what's the distribution of # of captured requests per connection id.  With the requestId in there too, everything would be unique.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -253,6 +259,9 @@ private boolean handleEndOfRequest(Accumulation accumulation) {
 
     private void handleEndOfResponse(Accumulation accumulation) {
         assert accumulation.state == Accumulation.State.REQUEST_SENT;
+        metricsLogger.atSuccess()
+                .addKeyValue("requestId", accumulation.getRequestId())
```
gregschohn, line 263: see above

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
```diff
@@ -129,6 +131,11 @@ public boolean isWorkOutstanding() {
         var start = timeShifter.transformSourceTimeToRealTime(originalStart);
         var end = timeShifter.transformSourceTimeToRealTime(originalEnd);
         var interval = numPackets > 1 ? Duration.between(start, end).dividedBy(numPackets-1) : Duration.ZERO;
+        metricsLogger.atSuccess()
+                .addKeyValue("requestId", requestKey.toString())
```
gregschohn, line 135: similar comment to above - I would break this into two dimensions, connectionId and requestId.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
```diff
@@ -129,6 +131,11 @@ public boolean isWorkOutstanding() {
...
+                .addKeyValue("delayFromOriginalToScheduledStartInMs", Duration.between(originalStart, start).toMillis())
```
gregschohn, line 136: This is good for now, but I'm also hoping to see a metric for drift between actual and scheduled.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
```diff
@@ -129,6 +131,11 @@ public boolean isWorkOutstanding() {
...
+                .addKeyValue("scheduledStartTime", start.toString())
```
gregschohn, line 137: I'm presuming that we can log a bunch of keys/dimensions and then choose to only turn some into proper metrics in the engine of our choosing... If that's the case, finding out the difference between the current and scheduled time could be nice.  I'm not sure if the current timestamp on the message would give us the ability to calculate that value in any/most/all metric toolkits.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -220,12 +222,21 @@ public DiagnosticTrackableCompletableFuture<String,Void> consumeBytes(ByteBuf pa
                     } else {
                         log.atInfo().setMessage(()->"Signaling previously returned CompletableFuture packet write had an exception : "
                                 + packetData + " hash=" + System.identityHashCode(packetData)).log();
+                        metricsLogger.atError(cause)
+                                .addKeyValue("channelId", channel.id().asLongText())
+                                .addKeyValue("requestId", diagnosticLabel)
```
gregschohn, line 227: Have this key off of a UniqueRequestKey value.  The diagnosticLabel could be set to anything, including null and all of the rest of the code should work just fine.  If you want to log requestIds, take that in instead of a diagnosticId.  If you want to log diagosticIds, then relabel.
Of course - I don't like either of those options as much as injecting a dependency.  The same way that we take a diagnosticLabel which is a string that's simply prepended to our log statements, you can do something similar here (& elsewhere, as I imagine that this pattern will be quite common).  You can supply a decorator Function<LoggingEventBuilder,LoggingEventBuilder> to the c'tor that adds any additional key-value-bindings that the calling context would like to use.  This is pretty similar to the MDC approach earlier, but more tightly controlled.  You could even have the Function subsumed by a metricsLogger field (instead of a static) so that none of your calls within a class would need to change/do anything.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -220,12 +222,21 @@ public DiagnosticTrackableCompletableFuture<String,Void> consumeBytes(ByteBuf pa
...
                         channel.close();
                     }
                 });
         log.atTrace().setMessage(()->"Writing packet data=" + packetData +
                 ".  Created future for writing data="+completableFuture).log();
+        metricsLogger.atSuccess()
+                .addKeyValue("channelId", channel.id().asLongText())
+                .addKeyValue("requestId", diagnosticLabel)
+                .addKeyValue("sizeInBytes", readableBytes)
+                .setMessage("Request packet written to target").log();
```
gregschohn, line 239: We don't have a guarantee that we've written a packet (the next level down the System stack could be buffering them and combining them).  Since these metrics are the most permanent things that we've done so far, it might be more appropriate to call these 'pieces' or something less network specific.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
```diff
@@ -158,6 +170,10 @@ private static Throwable unwindPossibleCompletionException(Throwable t) {
         DiagnosticTrackableCompletableFuture<String,R> finalizedFuture =
                 consumptionChainedFuture.thenCompose(v -> packetConsumer.finalizeRequest(),
                         ()->"HttpJsonTransformingConsumer.redriveWithoutTransformation.compose()");
+        metricsLogger.atSuccess()
+                .addKeyValue("requestId", diagnosticLabel)
+                .addKeyValue("channelId", channel.id().asLongText())
+                .setMessage("Request was redriven without transformation").log();
```
gregschohn, line 176: This was probably a failure.  Take a look at all of the callers, they're all exceptional cases (exceptions or missing signals)...
This is the "normal" exit (line 134).  That line should have a success metric and this existing one should have an error/warning one.
```
return StringTrackableCompletableFuture.completedFuture(v, ()->"transformedHttpMessageValue");
```

gregschohn (commented):
I still have the bottom quarter to look at.  I want to confirm that we're labeling things as success/error appropriately.  I'd like to see a trace as well.  Sending metrics for every packet could be extremely taxing/expensive.  It's really convenient for us to offload doing basic arithmetic, but it might not always be appreciated.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
mikaylathompson: Yes! I'd like to include that as well, but didn't see an obvious place to put it in. Do you have a suggestion for that?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
mikaylathompson: I like the idea of having both. I think that I would actually consider `connectionId` a dimension, and `requestId` more like a detail for an individual datapoint -- I can't imagine that we would ever use it as a dimension in cloudwatch, but it seems like a useful detail to include if a user needs to look into a specific event.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
mikaylathompson: To your first part: yes. At least using metric filters, we can grab up to 3 values from the log line and use them as dimensions, and 1 value as the actual metric.

To the second: we definitely have the information to do that in the log. Along with the timestamp as recorded by cloudwatch (which is what is used for the metric timing) Slf4J also includes an "instant" field with a very precise time that the log was generated. There's a little bit of format processing to be done, so I don't think metric filters are powerful enough to let us do it, but I assume most log processing solutions could.

```
{
    "instant":
    {
        "epochSecond": 1694369964,
        "nanoOfSecond": 184809000
    },
    "thread": "nioEventLoopGroup-2-1",
    "level": "INFO",
    "loggerName": "MetricsLogger.BacksideHttpWatcherHandler",
    "message": "Request scheduled to be sent",
    "endOfBatch": false,
    "loggerFqcn": "org.apache.logging.slf4j.Log4jEventBuilder",
    "contextMap":
    {
        "delayFromOriginalToScheduledStart": "PT0.304676469S",
        "requestKey": "0242acfffe140008-0000000e-00000005-d8a1533aa2d1dee1-3888e983.0",
        "scheduledStartTime": "2023-09-10T18:19:24.290418Z"
    },
    "threadId": 13,
    "threadPriority": 10
}
```

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -104,9 +106,17 @@ public List<TrafficStream> readNextTrafficStreamSynchronously() {
                     TrafficStream ts = TrafficStream.parseFrom(record.value());
                     // Ensure we increment trafficStreamsRead even at a higher log level
                     log.trace("Parsed traffic stream #{}: {}", trafficStreamsRead.incrementAndGet(), ts);
+                    metricsLogger.atSuccess()
+                            .addKeyValue("connectionId", ts.getConnectionId())
+                            .addKeyValue("topic-name", this.topic)
```
gregschohn, line 111: Why aren't you using camel case on this one too?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/netty/BacksideHttpWatcherHandler.java
```diff
@@ -26,6 +27,11 @@ public BacksideHttpWatcherHandler(AggregatedRawResponse.Builder aggregatedRawRes
     protected void channelRead0(ChannelHandlerContext ctx, FullHttpResponse msg) throws Exception {
         doneReadingRequest = true;
         triggerResponseCallbackAndRemoveCallback();
+//      TODO: Add requestId to this metrics log entry
+        metricsLogger.atSuccess()
+                .addKeyValue("channelId", ctx.channel().id().asLongText())
+                .addKeyValue("httpStatus", msg.status().toString())
```
gregschohn, line 33: Cool!  I forgot that we ARE parsing the HTTP responses right now (still).

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/netty/BacksideHttpWatcherHandler.java
```diff
@@ -38,6 +44,10 @@ public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
     @Override
     public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws Exception {
         aggregatedRawResponseBuilder.addErrorCause(cause);
+//      TODO: Add requestId to this metrics log entry
+        metricsLogger.atError(cause)
+                .addKeyValue("channelId", ctx.channel().id().asLongText())
+                .setMessage("Failed to receive full response").log();
```
gregschohn, line 50: "full response" and "component of response" could be hard to comprehend to an external metrics watcher.
Every handler has an exceptionCaught method (including the transformation ones).  Generally, I'd recommend putting exception counting logic at the top-level, rather than within each of the handlers.

gregschohn (commented):
How were these (& the capture proxy metrics that introduced MetricsLogger) tested?  How do we know that when there's an exception from a broken pipe that we'll get one metric value as opposed to conflicting information?  Some of the positions for these logs are currently tenuous.  As code changes, we need to make sure that the metrics stay consistent.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
mikaylathompson: UniqueRequestKeys are emitted nicely as strings, e.g. `"0242acfffe140008-0000000e-00000005-d8a1533aa2d1dee1-3888e983.0"`

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
mikaylathompson: Okay, I was struggling with how to track this one. It feels somehwere in between a success and a failure, because presumably successfully redriving is less of a failure than not doing so.  Changing this one to an error (especially since we have a throwable to use as the cause) makes sense.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
mikaylathompson: oops, no good reason. Will update.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/netty/BacksideHttpWatcherHandler.java
mikaylathompson: yeah, this was very useful!

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
gregschohn: Don't ever rely upon toString().  It's used by many different callers with different needs.  Tomorrow, it could be changed to ",0" or the longChannelId portion could be truncated.  If you need something specific, make the code state that.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
gregschohn: s/dimension/datapoint/. How you slice and dice data is up to your metrics processing system.
Knowing that a request was from a new connection vs a reused one might have a big impact on some performance numbers, at least for some systems.  Anyway - whether you make something into dimensions or just datapoints is not an implementation detail that this class (or anything in the Replayer) needs to be concerned with.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
gregschohn: RequestSenderOrchestrator currently has 4 uses of EventLoop::scheduleEvent().  That returns a ScheduledFuture.

You could add a helper function to coreUtils that takes an EventLoop, the arguments to scheduleEvent, and some metric data then wrap it all together to do the scheduling.  In the returned future, you can chain some metrics emission code to show the drift between scheduled and actual times.  This could/should be done with the RequestSenderOrchestrator having little to no knowledge that its schedules are being instrumented.  However, it should also make that call with a label or tag so that the implementation can determine what sensitivity to treat the metric emission at.

One thing to be aware of is that we schedule on a packet by packet basis, based upon the originally observed start-end times.  Those two bookend times might be things that are more critical than the intra-"packet" times.  However, having the internal datapoints will paint more complete pictures when diagnosing skew issues at a more macro level.  I just want to code to be flexible enough that it doesn't become overly expensive or inefficient past the point where it provides some users value.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
gregschohn: Are you logging that there was a redrive or that a request was successful?  It feels like you really should be emitting two different metrics: 1) Hey, I had to redrive because of some error in transformation & 2) Cool, this was a successful send, along w/ all of the other successful sends & oh, by the way, it was from a redrive.

We should be tracking the provenance/transformation path of our requests through to their responses.  It would be great to know that some request got a 403 and it underwent full-transformation or it got a 403 and it was redriven.  The TransformedOutputAndResult<TransformedTargetRequestAndResponse/AggregatedRawResponse> should contain that information (redrives should have 'ERROR' as their status)

gregschohn (approved):
I'd like to see some refactoring on how metrics context data is passed into some of the methods.  The right patterns will emerge more as we continue to track more metric points and we can attack that then.

PR #317, opened 2023-09-13
Replayer memory fixes and leak detection enhancements

gregschohn:
### Description

Fixed a leak in the replayer's PrettyPrinter (newly refactored from Utils), which didn't deal with ByteBufs in the cleanest way.  It should be tight and leak free now.

* Category Bug fix
* Why these changes are required? stability
* What is the old behavior before changes and new behavior after changes? No noticeable changes other than a slower memory consumption curve

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1313

### Testing
Unit testing

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Utils.java
```diff
@@ -70,119 +66,11 @@ public enum PacketPrintFormat {
...
-        return httpPacketBytesToString(msgType, Optional.ofNullable(packetStream).map(p->p.stream())
-                .orElse(null));
-    }
-
-    public static String httpPacketBytesToString(HttpMessageType msgType, Stream<byte[]> packetStream) {
-        return httpPacketBufsToString(msgType, packetStream.map(bArr->Unpooled.wrappedBuffer(bArr)));
-    }
-
-    public static String httpPacketBufsToString(HttpMessageType msgType, Stream<ByteBuf> packetStream) {
-        packetStream = packetStream.map(bb->bb.duplicate()); // assume that callers don't want read-indices to be spent
```
okhasawn, line 109: For my own curiosity and learning, I see that this file has been refactored to somewhere else, but this duplicate statement didn't, can you briefly explain why please?
Why don't we want that assumption anymore.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/PrettyPrinter.java
```diff
@@ -0,0 +1,172 @@
...
+                return prettyPrintNettyResponse((FullHttpResponse) httpMessage);
+            } else if (httpMessage == null) {
+                return "[NULL]";
+            } else {
+                throw new RuntimeException("Embedded channel with an HttpObjectAggregator returned an unexpected object " +
+                        "of type " + httpMessage.getClass() + ": " + httpMessage);
+            }
+        } finally {
+            holderOp.ifPresent(bbh->bbh.content().release());
+        }
```
okhasawn, line 112: I like the use of finally block here to make sure bytebufs are released.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/PrettyPrinterTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@TestWithNettyLeakDetection
+public class PrettyPrinterTest {
```
okhasawn, line 20: Not blocking, Do you think there are other test cases to account for? an empty buffer or a buffer with headers but no body for example.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/JsonEmitter.java
```diff
@@ -114,8 +111,8 @@ public ByteBuf recycleByteBufRetained() {
     private Stack<LevelContext> cursorStack;
 
     @SneakyThrows
-    public JsonEmitter() {
-        outputStream = new ImmediateByteBufOutputStream();
+    public JsonEmitter(ByteBufAllocator byteBufAllocator) {
+        outputStream = new ImmediateByteBufOutputStream(byteBufAllocator);
```
okhasawn, line 115: For my own learning, can you explain why `JsonEmitter` needs a `byteBufAllocator` as a parameter?

okhasawn (approved):
Non-blocking comments, mostly questions for my own curiosity. Thanks for this Greg!

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/JsonEmitter.java
gregschohn: Before, the JsonEmitter (by way of its inner class) just used ByteBufAllocator.DEFAULT as its ByteBufAllocator.  That may or may not be the best allocator to use for this channel, context, thread, etc.  I now have the caller pass in the ChannelHandlerContext's ByteBufAllocator.

My assumption is that the allocators are probably tied to a thread (or at least they should be) & using a thread-specific one will be much faster than one shared across many threads.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/PrettyPrinterTest.java
gregschohn: Yes, I'm not sure what the behavior should be there, so tests would be a good exercise to find out and to codify.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/PrettyPrinter.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.StringJoiner;
+import java.util.concurrent.Callable;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+@Slf4j
```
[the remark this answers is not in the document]
gregschohn: No, it isn't.  I can strike it in a future commit.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Utils.java
gregschohn: I think I made this change because there were already some cases where callers didn't care (like when byte[]s were passed in).  I generally don't like assumptions, so it's now up to the caller to call duplicate if necessary.

gregschohn (commented):
Thanks for the feedback.  I've left some minor notes, but I agree with your action items.  I'd like to get to the action items after you've completed all of the PRs for memory issues.

gregschohn:
See change [a884098](https://github.com/opensearch-project/opensearch-migrations/pull/325/commits/a8840982877b8f9b8d224cd1463c931b9bef1bbd) for cleanup items.

PR #322, opened 2023-09-13
Json emitter memory fixes

gregschohn:
### Description
(from the commit) 
I've made the JsonEmitter implement AutoCloseable so that the underlying stream can be closed, which in turn will release the compositeByteBuf that it uses to buffer contents.
The NettyJsonBodySerializeHandler calls fireChannelRead(), then immediately called release on the same buffer that it just sent downstream on the channel. That latter release() has no been removed.
I've added a new unit test to help to test the NettyJsonBodySerializeHandler and to make sure that the memory accesses are solid. I've also updated the JsonEmitter test to not leak so that we're closer to a zero-leaks target.

* Category Bug fix
* Why these changes are required? stability & memory is finite
* What is the old behavior before changes and new behavior after changes? We leaked memory and opened ourselves up to corruption and now two spots for that fun to happen should be gone.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1313

### Testing
Unit tests

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
```diff
@@ -34,16 +34,17 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
-    private void serializePayload(ChannelHandlerContext ctx, Map<String,Object> payload) throws IOException {
-        var pac = jsonEmitter.getChunkAndContinuations(payload, NUM_BYTES_TO_ACCUMULATE_BEFORE_FIRING);
-        while (true) {
-            ctx.fireChannelRead(new DefaultHttpContent(pac.partialSerializedContents));
-            pac.partialSerializedContents.release();
-            if (pac.nextSupplier == null) {
-                ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
-                break;
+    private void serializePayload(ChannelHandlerContext ctx, Map<String, Object> payload) throws IOException {
+        try (var jsonEmitter = new JsonEmitter(ctx.alloc())) {
```
okhasawn, line 38: Oh so this is why `JsonEmitter` implements `AutoCloseable`

okhasawn (approved):
Approved. Very similar to PR #317, should that one be closed?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
```diff
@@ -34,16 +34,17 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
         }
     }
 
-    private void serializePayload(ChannelHandlerContext ctx, Map<String,Object> payload) throws IOException {
-        var pac = jsonEmitter.getChunkAndContinuations(payload, NUM_BYTES_TO_ACCUMULATE_BEFORE_FIRING);
-        while (true) {
-            ctx.fireChannelRead(new DefaultHttpContent(pac.partialSerializedContents));
-            pac.partialSerializedContents.release();
```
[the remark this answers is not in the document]
gregschohn: fireChannelRead will call read, which will release.  This handler itself does not retain the buffer in its own capacity, so it shouldn't release the count either.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
gregschohn: It implements AutoCloseable because it holds onto a CompositeByteBuf.  That needs to be released at some point and AutoCloseable makes that easy and self-documenting.  And this code here is where we take advantage of the auto-close trait.

gregschohn:
See change [a884098](https://github.com/opensearch-project/opensearch-migrations/pull/325/commits/a8840982877b8f9b8d224cd1463c931b9bef1bbd) for cleanup items.

PR #324, opened 2023-09-15
Script to create Metric Filters

mikaylathompson:
### Description
This sets up about 17 metric filters in the user's AWS account that match the metrics logging lines.

There are a couple lines of caveats in the README and the beginning of the file. This could _definitely_ be improved moving forward in both simple and more complex ways.

One of the areas for big improvement in the future is linking this to the source code to ensure that the filters stay up to date with the code. This serves more as a reference implementation so is out of scope at the moment. It would also be much easier if the JSON options worked to configure the filters.

### Issues Resolved
n/a

### Testing
Manual, on several accounts.

### Check List

### DIFF
--- a/deployment/copilot/setMetricFilters.sh
+++ b/deployment/copilot/setMetricFilters.sh
@@ -1,0 +1,124 @@
+#!/bin/bash
+
+# A note on the implementation method:
+# This would be better implemented as a script using `boto` in python, or using
+# the `--cli-input-json` option. However, both of those seem to have a bug that prevents dimensions
+# and units from being provided.
+# There also seems to be a bug in this CLI interface that prevents multiple dimensions from being provided,
+# so for the time being all metrics are provided with 0 or 1 dimensions, and any additional ones need to be
+# added in the console UI.
+
+CAPUTRE_PROXY_LOG_GROUP="/copilot/migration-copilot-dev-capture-proxy-es"
+TRAFFIC_REPLAYER_LOG_GROUP="/copilot/migration-copilot-dev-traffic-replayer-default"
+# TODO: Both of the values above may change. The capture proxy could be running as a standalone and
+# therefore not have `-es` as part of the name. If multiple traffic replayers are deployed,
+# only the first would use `-default`. For now, the path to resolving this is for the user to
+# edit this file and change the values. Down the line, we should incorporate this with `devDeploy`
+# to automatically set these values. Additionally, this is currently untested with multiple
+# replayers. I believe the correct solution would be to add a dimension for which replayer is
+# being instrumented.
+
+
+# CAPTURE PROXY
+aws logs put-metric-filter \
+    --log-group-name $CAPUTRE_PROXY_LOG_GROUP \
+    --filter-name "CaptureProxy-FullRequestReceived" \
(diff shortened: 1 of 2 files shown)


deployment/copilot/setMetricFilters.sh
```diff
@@ -0,0 +1,124 @@
+#!/bin/bash
+
+# A note on the implementation method:
+# This would be better implemented as a script using `boto` in python, or using
+# the `--cli-input-json` option. However, both of those seem to have a bug that prevents dimensions
```
gregschohn, line 5: Don't say that somebody else's code has a bug unless you're citing it.

deployment/copilot/setMetricFilters.sh
```diff
@@ -0,0 +1,124 @@
...
+# and units from being provided.
+# There also seems to be a bug in this CLI interface that prevents multiple dimensions from being provided,
+# so for the time being all metrics are provided with 0 or 1 dimensions, and any additional ones need to be
+# added in the console UI.
+
+CAPUTRE_PROXY_LOG_GROUP="/copilot/migration-copilot-dev-capture-proxy-es"
```
gregschohn, line 11: CAPTURE is misspelled through the PR.

deployment/copilot/setMetricFilters.sh
```diff
@@ -0,0 +1,124 @@
...
+# replayers. I believe the correct solution would be to add a dimension for which replayer is
+# being instrumented.
+
+
+# CAPTURE PROXY
+aws logs put-metric-filter \
+    --log-group-name $CAPUTRE_PROXY_LOG_GROUP \
+    --filter-name "CaptureProxy-FullRequestReceived" \
+    --filter-pattern "{ ($.message  = \"Sent message to Kafka\") && ($.loggerName = \"MetricsLogger.*\") }" \
+    --metric-transformations metricName="FullRequestReceived",metricNamespace="capture-proxy",metricValue=1,dimensions={"Method"="$.contextMap.httpMethod"},unit="Count"
```
gregschohn, line 27: This pattern is not maintainable.  "Sent message to Kafka" could easily change.
My recommendation is to bind these tightly with the code that is emitting the logs.
There are a lot of options for how to keep the script's implementation more tightly bounded to the java code - but one option that might be the simplest, would be to
1) refactor the java code to pull your "magic strings" within the metrics logger.  That means that you'll replace things like logSuccess("...",...) with 17 different method calls, or one method call and an enum that takes 17 different scenarios.  You'd want it to be something other than strings so that developers don't corrupt the strings in multiple places.
2) you write a new utility class in the build system for copilot that loads the metrics class (or java file) and dumps the values that you want to pull in.  Either way, the MetricsLogger class should have a unit test within that codebase to test that the contract with the copilot code is being maintained.  That means that you have a unit test open the java file as text if that's the approach, or you load the enum from the loaded class.  In both cases,  make sure that there are 17 records.  If there are more/less, document in the failed test results what needs to happen in the auxiliary code to make it complete.
3) convert the single-magic strings into these lines.  If you have a JVM, I would just load the metrics logger & write a quick java program to emit this stuff.  I presume that you want to wait on calling the service OUTSIDE of build time (which seems like a good idea) - and requiring java to be present for doing this at runtime seems overkill.

gregschohn (changes requested):
See feedback

PR #346, opened 2023-10-09
Updating the Historical Capture CDK stack definition to use Fetch Migration

kartg:
### Description

This change updates the Historical Capture stack to use the Fetch Migration (i.e. Data Prepper based) Docker image rather than Logstash. Since Fetch Migration is a run-once workflow rather than a service with uptime, the Historical Capture stack now relies on manually kicking off a Fetch Migration task with the ECS cluster instead of spinning up an ECS service in the cluster.

A template Data Prepper pipeline file with safe defaults is included. The Fetch Orchestrator python module now supports reading an "inline" Data Prepper pipeline configuration via the `INLINE_PIPELINE` environment variable. If present, its value is then written out as the Data Prepper pipeline config to be used. This value is expected to be a base64-encoded string.

The commits in this PR also include several permission tweaks in the interest of least-permission security:
* Added a `FetchMigrationTaskExecutionRole` to the Migration Console to allow kicking off the Fetch Migration ECS cluster task
* Added a `fetchMigrationECRAccessSG` security group to enable pulling the Docker image from ECR, since the existing Migration Domain security group disallows all outgoing requests
* Also added CFN outputs that provide export values for use in Copilot commands, and the awscli command to actually kick off the Fetch Migration ECS task
* The CDK uploads the `INLINE_PIPELINE` value to AWS Secrets Manager so it can be imported (via ARN) into the ECS Container when the Fetch Migration task is run.

### Testing
```
(venv) $ python -m coverage run -m unittest
..........................................
Ran 42 tests in 0.067s

OK
(venv) $ python -m coverage report --omit "*/tests/*"
index_operations.py               35      0   100%
metadata_migration.py            123      1    99%
migration_monitor.py              58      0   100%
migration_monitor_params.py        5      0   100%
TOTAL                            277      1    99%

```

### Check List


deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -190,19 +190,17 @@ export class StackComposer {
...
-        // state will potentially have both stacks trying to add the same data
+        // TODO: Future work to provide orchestration between historical capture and migration assistance
         if (historicalCaptureEnabled && networkStack) {
             const historicalCaptureStack = new HistoricalCaptureStack(scope, "historicalCaptureStack", {
                 vpc: networkStack.vpc,
-                logstashConfigFilePath: logstashConfigFilePath,
+                dpPipelineTemplatePath: dpPipelineTemplatePath,
                 sourceEndpoint: sourceClusterEndpoint,
-                targetEndpoint: opensearchStack.domainEndpoint,
+                targetEndpoint: process.env.MIGRATION_DOMAIN_ENDPOINT!,
```
lewijacn, line 199: Was not expecting to get the target endpoint like this (do we ever add this to the current CDK environment?), any issue with using `opensearchStack.domainEndpoint` here

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -60,9 +60,9 @@ export class StackComposer {
         const mskARN = getContextForType('mskARN', 'string')
         const mskEnablePublicEndpoints = getContextForType('mskEnablePublicEndpoints', 'boolean')
         const mskBrokerNodeCount = getContextForType('mskBrokerNodeCount', 'number')
-        const sourceClusterEndpoint = getContextForType('sourceClusterEndpoint', 'string')
         const historicalCaptureEnabled = getContextForType('historicalCaptureEnabled', 'boolean')
-        const logstashConfigFilePath = getContextForType('logstashConfigFilePath', 'string')
+        const dpPipelineTemplatePath = getContextForType('dpPipelineTemplatePath', 'string')
```
lewijacn, line 64: Could you update the CDK readme `deployment/cdk/opensearch-service-migration/README.md` to include these new options (and remove the old ones)

FetchMigration/Dockerfile
```diff
@@ -14,5 +16,9 @@ COPY python/*.py .
 # update PATH
 ENV PATH=/root/.local:$PATH
 
+# Write custom DP config
+RUN echo "ssl: false" > $DATA_PREPPER_PATH/config/data-prepper-config.yaml
+RUN echo "metricRegistries: [Prometheus]" >> $DATA_PREPPER_PATH/config/data-prepper-config.yaml
+
 # Include the -u flag to have stdout logged
-ENTRYPOINT python -u ./fetch_orchestrator.py $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml https://localhost:4900
+ENTRYPOINT python -u ./fetch_orchestrator.py $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml http://localhost:4900
```
lewijacn, line 24: Do we have a limitation with only using http for data prepper?

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
```diff
@@ -1,62 +1,97 @@
...
+        // Create DP pipeline config from template file
+        let dpPipelineData: string = readFileSync(props.dpPipelineTemplatePath, 'utf8');
+        dpPipelineData = dpPipelineData.replace("<SOURCE_CLUSTER_HOST>", props.sourceEndpoint);
+        dpPipelineData = dpPipelineData.replace("<TARGET_CLUSTER_HOST>", props.targetEndpoint);
+        // Base64 encode
+        let encodedPipeline = Buffer.from(dpPipelineData).toString("base64");
+
+        // Create secret using Secrets Manager
+        const dpPipelineConfigSecret = new SMSecret(this, "dpPipelineConfigSecret", {
+            secretName: `${historicalCaptureFargateTask.family}-${historicalCaptureContainer.containerName}-pipelineConfig`,
```
lewijacn, line 49: I don't think we have to address this right now (as I believe there are a couple other spots in the IaC as well), but worth calling out that if a user deployed multiple instances of our migration solution in the same region (e.g. if they were migrating multiple target clusters separately) I would expect this name to cause a collision

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
```diff
@@ -1,62 +1,97 @@
...
-            desiredCount: 1
+        // The Migration Domain SG disallows all outgoing, but this prevents the ECS run task
+        // command from pulling the Docker image from ECR. Thus, create an SG to allow this.
+        // TODO - Limit outbound to only ECR endpoints somehow
+        const fetchMigrationSecurityGroup = new SecurityGroup(this, 'fetchMigrationECRAccessSG', {
+            vpc: ecsCluster.vpc
         });
 
+        // CFN outputs that will be consumed by Copilot
+        new CfnOutput(this, 'FMTaskDefArn', {
```
lewijacn, line 65: Will be glad to rid ourselves of this CfnOutput soon...

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
```diff
@@ -1,62 +1,97 @@
...
+        new CfnOutput(this, 'FMExecRoleArn', {
+            exportName: 'FetchMigrationExecutionRoleArn',
+            value: historicalCaptureFargateTask.obtainExecutionRole().roleArn,
+            description: 'ARN of the Execution Role for Fetch Migration',
+        });
+        // Documentation - https://docs.aws.amazon.com/cli/latest/reference/ecs/run-task.html#options
+        let networkConfigJson = {
+            "awsvpcConfiguration": {
+                "subnets": ecsCluster.vpc.privateSubnets.map(_ => _.subnetId),
+                "securityGroups": [process.env.MIGRATION_DOMAIN_SG_ID, fetchMigrationSecurityGroup.securityGroupId]
```
lewijacn, line 84: Same as before, surprised you can access the MIGRATION_DOMAIN_SG_ID in this way

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
```diff
@@ -1,62 +1,97 @@
...
+        historicalCaptureContainer.addSecret("INLINE_PIPELINE", 
+            ECSSecret.fromSecretsManager(dpPipelineConfigSecret)
+        );
 
-        // Create Fargate Service
-        const historicalCaptureFargateService = new FargateService(this, "historicalCaptureFargateService", {
-            cluster: ecsCluster,
-            taskDefinition: historicalCaptureFargateTask,
-            desiredCount: 1
+        // The Migration Domain SG disallows all outgoing, but this prevents the ECS run task
```
lewijacn, line 57: I wasn't aware of this as I haven't had issue with the other docker ECS services. Are you using a public ECR image versus a public docker image, which is the reason this is happening?

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
```diff
@@ -1,62 +1,97 @@
...
+        // Documentation - https://docs.aws.amazon.com/cli/latest/reference/ecs/run-task.html#options
+        let networkConfigJson = {
+            "awsvpcConfiguration": {
+                "subnets": ecsCluster.vpc.privateSubnets.map(_ => _.subnetId),
+                "securityGroups": [process.env.MIGRATION_DOMAIN_SG_ID, fetchMigrationSecurityGroup.securityGroupId]
+            }
+        }
+        let networkConfigString = JSON.stringify(networkConfigJson)
+        // Output the ECS run task command using template literals
+        let executionCommand = `aws ecs run-task --task-definition ${historicalCaptureFargateTask.taskDefinitionArn}`
```
lewijacn, line 89: Where should this command be run from? I'm trying to put together how this plays with the container you created

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
kartg: `process.env.` is used to access environment variables so this is accessing the `MIGRATION_DOMAIN_ENDPOINT` variable that is included in the networkStack's export commands. 

I believe i moved away from `opensearchStack.domainEndpoint` because this variable isn't available/populated when no changes are made to the openSearchStack.

The ideal fix here would be to use CFN Output variables. I'll tackle that in a follow up PR.

FetchMigration/Dockerfile
kartg: Not a limitation, but there's no need to use SSL and auth for the Data Prepper endpoint since it is only accessed within the Fetch Migration container for metrics information (and shutdown)

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
kartg: As noted in the [other comment](https://github.com/opensearch-project/opensearch-migrations/pull/346#discussion_r1355753633), this accesses the exported environment variable. Ideally, we would use CFN outputs for this.

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
kartg: The Fetch Migration container will host the task that this command launches. The user is expected to run this command on the Migration Console to kick off Fetch Migration.

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
kartg: The reason this is happening is because the `ecs run task` command is [executed via the Migration Console container](https://github.com/opensearch-project/opensearch-migrations/pull/346#discussion_r1355760947), which has this SG set on it. Kicking off an ECS task [requires outbound access to ECR over the public internet](https://repost.aws/knowledge-center/ecs-unable-to-pull-secrets).

The other ECS services don't run into this issue because their corresponding commands are executed on your developer host (which does not have such an outbound restriction set). Once the container is bootstrapped, this access is not needed.

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
lewijacn: Ah okay so launching this task won't spin up another container it will use the existing container in the cluster

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
lewijacn: Good to know, thanks for the info 👍

deployment/cdk/opensearch-service-migration/lib/historical-capture-stack.ts
kartg: That's correct. Note that it will launch a separate task in the historical migration cluster though - there's currently no protection from having multiple tasks performing duplicate work

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
kartg: > I'd be very surprised if opensearchStack.domainEndpoint wasn't available on every deployment (even with no changes) as I have been testing this way for a while. Happy to see this in action though and get it resolved if that is the case.

@lewijacn you're probably right. I now recall why i made this change, and it doesn't have to do with the CDK stack. Rather, using the environment variable decouples the HistoricalMigration stack from the creation of the `opensearchStack` via the CDK. This way, a user can plug in an existing stack as the target by simply setting the `MIGRATION_DOMAIN_ENDPOINT` environment variable.

If you'd prefer I revert to `opensearchStack.domainEndpoint`, let me know - i'll add a dependence in the surrounding if-condition: https://github.com/kartg/opensearch-migrations/blob/d7c2cb1fff42ad3bb19ea5e5a68fe32e6147ef5d/deployment/cdk/opensearch-service-migration/lib/stack-composer.ts#L194

lewijacn: I wouldn't worry about that too much right now, as this will get overhauled by the Parameter store changes. I was more concerned with the other comments you had, but glad to hear these are not an issue 👍

lewijacn (approved):

PR #351, opened 2023-10-15
Refactor TrafficSource packages

gregschohn:
### Description
Move all traffic source related classes into their own package.

Return TrafficStreamWithKey objects from the Traffic sources so that callers can circle back at some point in the future with a reference that can be used to decrement a refcount, advance a commit, etc.

I've begun to implement a facility within BlockingTrafficSource to restrict the total number of outstanding requests that can be outstanding at any point in time.  I hadn't written tests or completed this feature as I got distracted by other pre-reqs needed for that feature.  

**_I later realized that just about everything that I was doing in BlockingTrafficSource (& doneProcessing) wasn't the best way to limit requests._**  See https://github.com/opensearch-project/opensearch-migrations/pull/353 for how this eventually got implemented.

* Category: Refactoring
* Why these changes are required? To make the replayer more reliable and resistant to memory spikes.
* What is the old behavior before changes and new behavior after changes? Not much of a difference yet, probably more bugs (see subsequent commits)

### Issues Resolved
None - but part of https://opensearch.atlassian.net/browse/MIGRATIONS-1392 and https://opensearch.atlassian.net/browse/MIGRATIONS-1352

### Testing
Not enough for this one.

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/UniqueRequestKey.java
```diff
@@ -2,13 +2,17 @@
 
 import lombok.AllArgsConstructor;
 
-@AllArgsConstructor
 public class UniqueRequestKey {
-    public final String connectionId;
+    public final TrafficStreamKey trafficStreamKey;
     public final int requestIndex;
```
lewijacn, line 7: nit: I will be continually having to remind myself what a `requestIndex` and a `trafficStreamIndex` for TrafficStreamKey are in the context of the Replayer, would love a couple short comments

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -127,18 +131,21 @@ public List<TrafficStream> readNextTrafficStreamSynchronously() {
...
+            return trafficStream.collect(Collectors.<ITrafficStreamWithKey>toList());
         } catch (Exception e) {
             log.error("Terminating Kafka traffic stream");
             throw e;
         }
     }
 
+    @Override
+    public void commitTrafficStream(TrafficStreamKey trafficStreamKey) {
+        kafkaConsumer.commitSync(Map.of());
```
lewijacn, line 143: Since we are using an empty Map does that mean we are not committing anything here?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -604,15 +618,16 @@ private static SourceTargetCaptureTuple getSourceTargetCaptureTuple(RequestRespo
         }
     }
 
-    public void runReplay(ITrafficCaptureSource trafficChunkStream,
-                          CapturedTrafficToHttpTransactionAccumulator trafficToHttpTransactionAccumulator) {
+    public void pullReplayFromSourceToAccumulator(
```
lewijacn, line 621: nit: Stopped me a bit to understand what it meant to `pullReplay` would prefer `pullRecords` but don't know if that is much better

lewijacn (commented):
One comment around the Kafka piece was concerning other than that looks good

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/UniqueRequestKey.java
gregschohn: In subsequent edits, these become 
```
    public final TrafficStreamKeyWithRequestOffset trafficStreamKeyAndOffset;
    public final int replayerRequestIndex;
```

Does replayerRequestId help more?  I'll think about what more I can add too.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
gregschohn: Good callout - nobody calls that function yet though.  We could throw a NotImplementedException(), but I also had figured that we'd be picking up [this task](https://opensearch.atlassian.net/browse/MIGRATIONS-1379) within a couple days.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: I'm nervous about using the word Record since that's now a Kafka term.  Same with TrafficStream... However, I think that this should be pullCaptureFromSourceIntoAccumulator - or something more like that.  This isn't a replay yet, but it is a capture.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/UniqueRequestKey.java
lewijacn: I may just not have a good understanding of what that field is. Is replayerRequestIndex the number of the request for the given traffic stream or for the given connection? Or something else maybe?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
lewijacn: Gotcha as long as what we plan will cover I'm fine. I just wanted to be clear that currently we aren't doing any manual kafka commits

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
lewijacn: Agree capture is better

lewijacn (approved):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/UniqueRequestKey.java
gregschohn: replayerRequestIndex is used to sequence requests onto the netty worker threads that will transform and send the requests.  This must be indexed from 0 so that we know when we've gotten the first record.  In later commits, that's happening and there's a separate field to track what the TrafficStream's request offset was.  That allows this UniqueRequestKey to show the current offset that the process is aware of + the captured request index as it came in from the server.  That's regardless of whether we got the first TrafficStream object or a subsequent one thanks to the trafficStreamIndex.

PR #352, opened 2023-10-15
Tighter traffic stream key tracking

gregschohn:
### Description

(From the commit message.  Notice that this should be based off of [PR-351](https://github.com/opensearch-project/opensearch-migrations/pull/351)) 

Work on limiting the number of TrafficStreams/requests that can be present at any time.

Most of the lines changed are due to making TrafficStreamKeys into an interface so that I can wrap them and extend them.

Changes that were more significant include adding a couple fields of metadata to each TrafficStream object.  I haven't modified the capture code yet to fill those values in, but I do have a class in the replayer test directory that converts older versions of streams into the newer version by augmenting the TrafficStreams with metadata that was pulled in from previous TrafficStreams.

The reason that metadata is now required is because every TrafficStream object needs to stand on its own.  As the codebase was, and still is as of this commit, TrafficStreams are resequenced and ALL indices for a given connection must be present before any of the requests from the stream are accumulated and replayed.  Going forward, we want/need to commit records from Kafka on a per-record basis.  Many different connections will have TrafficStreams jumbled together, so the last TrafficStream for one connection could be surrounding many other TrafficStream sequences.  That means that we'll be issuing commits to Kafka at many partial points of connections' streams.  Notice that we won't be able to commit past a point where there could be some data that we still need to send/process.  However, we WILL run into many situations where there will be ambiguity at the first reads of a TrafficStream with index > 0.  When the prior TrafficStream isn't available, it isn't possible to determine when the first observations were reads if they were for a new request or one continued from the last TrafficStream.  The new metadata field for lastObservationWasRequestRead give the downstream processors enough information to reconstruct the requests within this TrafficStream and beyond.  requestCount lets us apply a sequential ordering to all of the requests within a TrafficStream.

BlockingTrafficStream has added the ability to track costs for each TrafficStream that has been returned by its underlying source.  Those cost values are then used as the argument to acquire for a semaphore.  The caller constructs BlockingTrafficStream with a maximum cost that is permissible and two lambda functions.  One function computes the cost, which is calculated before the TrafficStream object is returned by readNextChunk() and is persisted by the BlockingTrafficStream.  The other lambda calculates how many requests are within the TrafficStream so that the BlockingTrafficSource can countdown the number of times that each TrafficStream has a request that is marked as done.  Once that counter reaches 0, the cost is released from the semaphore and the tracking item is removed.

Work is still remaining to process TrafficStreams immediately, rather than waiting for them within the OnlineRadixSorter.  Changes need to be made to the capture kafka code to enforce relative stream ordering so that we don't need to do sorting.  Strict ordering is a critical requirement now so that we don't need to look ahead for prior TrafficStream objects.  That work will also remedy an issue that currently is causing the replayer to deadlock when TrafficStream objects are pending due to earlier TrafficStreams not being available.  However, we'll never even check for later ones or advance the cursor to expire elements because we're stuck waiting because the cost semaphore is waiting for the completion of those waiting TrafficStreams.

* Category Refactoring/Enhancement
* Why these changes are required? To improve reliability of the replayer

### Issues Resolved
None - but part of https://opensearch.atlassian.net/browse/MIGRATIONS-1392 and https://opensearch.atlassian.net/browse/MIGRATIONS-1352

### Testing

### Check List


TrafficCapture/captureProtobufs/src/main/proto/TrafficCaptureStream.proto
```diff
@@ -69,6 +69,8 @@ message TrafficObservation {
 message TrafficStream {
   string connectionId = 1; // persistent identifier for all stream objects within a single connection
   string nodeId = 5; // unique namespace such as a nodeId or computer mac address
+  optional int32 requestCount = 6;
```
lewijacn, line 72: Do we plan on these always being optional or just into we get things in place for TrafficCapture?

TrafficCapture/captureProtobufs/src/main/proto/TrafficCaptureStream.proto
```diff
@@ -69,6 +69,8 @@ message TrafficObservation {
...
+  optional bool lastObservationWasRequestRead = 7;
```
lewijacn, line 73: Do we have any concern with responses that take multiple TrafficStreams? I read this as though we are only concerned with requests that span multiple TrafficStreams

lewijacn (commented):
Couple comments around changes to Protobuf I wanted to clarify

TrafficCapture/captureProtobufs/src/main/proto/TrafficCaptureStream.proto
gregschohn: I think they'll always be optional.  In subsequent commits they're optional.  With protobuf, an optional value that isn't present in the binary will be interpreted as the default value for the type, meaning 0 and false.  That's what I want anyway and lets us remain backward compatible with existing streams.

TrafficCapture/captureProtobufs/src/main/proto/TrafficCaptureStream.proto
gregschohn: I think they'll always be optional. In subsequent commits they're optional. With protobuf, an optional value that isn't present in the binary will be interpreted as the default value for the type, meaning 0 and false. That's what I want anyway and lets us remain backward compatible with existing streams.

TrafficCapture/captureProtobufs/src/main/proto/TrafficCaptureStream.proto
lewijacn: I see and yes this was more of a replayer concern I had. That said getting the full requests would be priority and at least initially we could do a best effort approach to get the full response but there are at least restart cases where we plan to do improvement in the future

lewijacn (approved):

TrafficCapture/captureProtobufs/src/main/proto/TrafficCaptureStream.proto
gregschohn: Notice that there are 3 different ways that we'll terminate a response.  
1. The next observation received is a close
2. The next observation received is the beginning of a new request (read observations)
3. There is no next TrafficStream and the liveStreams object (an ExpiringTrafficStreamMap field in CapturedTrafficToHttpTransactionAccumulator) expires the TrafficStream.

If the proxy is expiring connections after it's configured period.  Either there's a misconfiguration that the window is too short, or we're missing data coming in from Kafka.  Both of which should involve investigation of the environment.

PR #353, opened 2023-10-15
Shift request semaphore from blocking traffic source into traffic stream limiter for traffic replayer to throttle more directly

gregschohn:
### Description

(From the commit message. Notice that this should be based off of [PR-352](https://github.com/opensearch-project/opensearch-migrations/pull/352).  You can see just the commits against that base with [this link](https://github.com/opensearch-project/opensearch-migrations/pull/353/files/d715801c77644bcb98b56d8d7aed2829f380345d..32ec5293966c619abc22a9c4ab1f072a0a6ede1c))

* Category Enhancement. Bug fix
* Why these changes are required? To limit the amount of memory consumed so that during a spike, the replayer won't exhaust memory
* What is the old behavior before changes and new behavior after changes? If there were no bugs (see PR-354), the replayer could run with any replay rate/any number of messages being received from Kafka for even a single point in time and it still wouldn't hold more requests than the number specified at the command line, limiting its total memory footprint.  In other words, this will buffer less, make requests maybe more slowly (but probably not because we were likely limited anyway), and most importantly, NOT CRASH.

### Issues Resolved
Part of https://opensearch.atlassian.net/browse/MIGRATIONS-1392 and https://opensearch.atlassian.net/browse/MIGRATIONS-1352

### Testing

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -84,15 +85,25 @@ public TrafficReplayer(URI serverUri,
                            IAuthTransformerFactory authTransformerFactory,
                            boolean allowInsecureConnections)
             throws SSLException {
-        this(serverUri, authTransformerFactory, allowInsecureConnections, 1,
+        this(serverUri, authTransformerFactory, allowInsecureConnections, 0, 1024);
```
lewijacn, line 88: Why is numSendingThreads 0 here?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -397,24 +414,31 @@ void runReplayWithIOStreams(Duration observedPacketConnectionTimeout,
...
 
+            liveTrafficStreamLimiter.addWork(1);
             var requestPushFuture = transformAndSendRequest(replayEngine, request, requestKey);
             requestFutureMap.put(requestKey, requestPushFuture);
-            requestPushFuture.map(f->f.whenComplete((v,t)-> log.atTrace()
-                    .setMessage(()->"Summary response value for " + requestKey + " returned=" + v).log()),
+            liveRequests.put(requestKey, true);
+            requestPushFuture.map(f->f.whenComplete((v,t)->{
+                        liveRequests.remove(requestKey);
+                liveTrafficStreamLimiter.doneProcessing(1);
```
lewijacn, line 430: nit: Indentation looks off here

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -397,24 +414,31 @@ void runReplayWithIOStreams(Duration observedPacketConnectionTimeout,
...
     }
 
+    ConcurrentHashMap<UniqueRequestKey, Boolean> liveRequests = new ConcurrentHashMap<>();
+
     private BiConsumer<UniqueRequestKey, HttpMessageAndTimestamp>
     getRecordedRequestReconstructCompleteHandler(ReplayEngine replayEngine) {
         return (requestKey, request) -> {
             replayEngine.setFirstTimestamp(request.getFirstPacketTimestamp());
 
+            liveTrafficStreamLimiter.addWork(1);
```
lewijacn, line 424: If we have allotted a max of 1024 concurrent requests which are all full, will we block right here until space becomes available?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: 0 tells netty to decide.  It will decide based upon the number of CPU cores that are on the host.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: I'm likely going to wait on this until I get checkstyle going on the whole repo.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
lewijacn: Okay and we have other logic to say we are at max concurrent requests with requests being blocked don't take in anymore Kafka records?

lewijacn (approved):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: addWork (from this callback) is run from within pullCaptureFromSourceToAccumulator, so blocking within the callback means that the entire accumulation thread is blocked.  pullCaptureFromSourceToAccumulator pulls from Kafka or any other producer.  The Kafka traffic source (& others) can only read an bounded number of requests with each call to readNextTrafficStreamChunk().  By blocking here, that puts necessary constraints on when we'll be bringing more captures/requests into the process.

PR #354, opened 2023-10-15
Update capture to use new protobuf fields to improve reliability and consistency between stream faults for the replayer

gregschohn:
### Description
Builds upon https://github.com/opensearch-project/opensearch-migrations/pull/353, so only the last commit has new changes.

* Category Enhancement, New features
* Why these changes are required? To provide more data within each TrafficStream object so that the replayer can work more reliably after picking up TrafficStreams from an index > 0.
* What is the old behavior before changes and new behavior after changes?  The capture proxy now _correctly_ writes some of the critical metadata that the replayer needs so that it won't drop messages.

### Issues Resolved
None

### Testing

### Check List


TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -332,6 +342,9 @@ private void addSubstreamMessage(int captureFieldNumber, int dataFieldNumber, in
         }
         // Write data field
         writeByteBufferToCurrentStream(dataFieldNumber, byteBuffer);
+        if (captureFieldNumber == TrafficObservation.READ_FIELD_NUMBER) {
```
lewijacn, line 345: One concern about the placement here. This should work fine for non-segmented messages as far as I can tell. For segmented messages it may be obvious for us from the Replayer that this is a continuation but at least here we are not setting this `readObservationsAreWaitingForEom = true`. Is this something we are checking in the Replayer?

lewijacn (commented):
Looks good! One additional thing I was thinking about (doesn't have to be done now) is does this change the role of the ConditionallyReliableLoggingHttpRequestHandler class now? Since we are ensuring order now is this still relevant?

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
lewijacn: This can be ignored, even segmented messages enter this method

lewijacn: Separately from our discussion, we will probably need this if condition to have `TrafficObservation.READSEGMENT_FIELD_NUMBER` for segmented requests that enter here

gregschohn: Ahhh - here’s what I was thinking, though I think that I was wrong (I’ll write some tests).
If we get an observation that starts with a segment, we’re here because the message wasn’t big enough to fit within the available buffer.  I was thinking that that meant the PREVIOUS buffer, so that any traffic stream that started with a segment must mean that there were read bytes on the previous trafficStream.  However, we could have an observation bigger than our max TrafficStream size (10K message, 1K max size).  In that case, we’d need to segment the first message too - so we’d end up back in an ambiguous position.

lewijacn: Assuming we will cover this in a update somewhere else

lewijacn (approved):

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
gregschohn: I'm working on this within https://opensearch.atlassian.net/browse/MIGRATIONS-1379.

PR #362, opened 2023-10-20
Add negative authorization test to E2E script

okhasawn:
### Description
This PR adds a test to the E2E script that sends unauthorized credentials to the clusters.
This ensures the security of the authentication process and validates that unauthorized access is prevented.

### Issues Resolved
[MIGRATIONS-1153](https://opensearch.atlassian.net/browse/MIGRATIONS-1153)

### Check List

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -164,8 +164,9 @@ class E2ETests(unittest.TestCase):
         self.assertEqual(response.status_code, HTTPStatus.OK)
 
     def test_004_negativeAuth(self):
+        # This sends negative credentials to the clusters to validate that unauthorized access is prevented.
         alphabet = string.ascii_letters + string.digits
-        for _ in range(10):  # Adjust the range as needed
+        for _ in range(10):
             username = ''.join(secrets.choice(alphabet) for _ in range(8))
             password = ''.join(secrets.choice(alphabet) for _ in range(8))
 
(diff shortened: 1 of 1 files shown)


test/tests.py
```diff
@@ -160,3 +162,20 @@ def test_003_jupyterAwake(self):
         # Making sure that the Jupyter notebook is up and can be reached.
         response = requests.get(self.jupyter_endpoint)
         self.assertEqual(response.status_code, HTTPStatus.OK)
+
+    def test_004_negativeAuth(self):
```
sumobrian, line 166: I would add what the test is doing to the function name. For example,
test_004_negativeAuth_InvalidCreds.

sumobrian: Also I would add at least one more 0 to your test case numbers.

sumobrian (changes requested):
I mentioned that you should break up test cases into separate review but I think there are few more related test cases you should add.

1. Missing Authentication: Attempt to access resources without any authentication details.

2. Invalid Authentication: Use expired, malformed, or otherwise invalid tokens, session IDs, or credentials.

I would normally add authorization tests too (where you are testing insufficient privileges) but in this case, save these for another PR.

okhasawn:
> I mentioned that you should break up test cases into separate review but I think there are few more related test cases you should add.
> 
>     1. Missing Authentication: Attempt to access resources without any authentication details.

Just addressed in a different test
 
>     2. Invalid Authentication: Use expired, malformed, or otherwise invalid tokens, session IDs, or credentials.
> 
> 

The cluster we currently deploy doesn't use token-based authentication, so at this point, this is not possible. BUT, just took a quick look and found it might be possible to use both basic and token-based authentication.

sumobrian (approved):

PR #368, opened 2023-10-24
Reliability improvements for recovering mid channel traffic streams

gregschohn:
### Description

Fix bugs within the transferral of traffic streams from the capture proxy through the CapturedTrafficToHttpTransactionAccumulator.  I’ve rewritten the state machine in the latter as I’ve also added some new states.  Along with that are a couple tests that attempt (& I believe do) test through all possible permutations of captures that are relevant to the accumulator.

To run the more E2E testing between the capture and accumulation code, I wanted to make sure that I could take advantage of the new ordering behavior in the (Kafka)Offloader but I had realized that the ordering changes ONLY went in for the KafkaCaptureFactory.  To consolidate things better, I refactored the StreamChannelConnectionCaptureSerializer to take an object in its constructor instead of a couple callback functions.  I also eliminated the hash map lookups for each of the factories so that that point of contention and suspicion is now gone (even though it was never implicated, it was another spot to check).  The Kafka offloaded as well as others now extend an OrderedStreamLifeceycleManager that chains the CompletableFutures so that the order of messages committed to durable storage is preserved.  That OrderedStreamLifeceycleManager itself extends from a base class that doesn’t provide ordering so other offloaders that don’t require ordered writes can just extend that.

I’ve also found and fixed a couple bugs/bad assumptions within the capture code.  One of the fixes was done by removing the count field from ReadSegment and WriteSegment.  We were miscalculating the size that we were budgeting for in one place.  Since no code was using it, I removed those fields.

Lastly, I’m beginning to track the TrafficStreams that are being “held” by a given request.  These are the stream objects that can be “released” (committed, forgotten, etc) by callers once the request is fully handled.  I’m using that feature in the ExhaustiveCapturedTrafficToHttpTransactionAccumulatorTest, though there are still bugs.  Partial transactions are cobbled together when they are expired and sent through with their held TrafficStreams set within the RequestResponsePairs.  That eliminates a lot of the value, but I'll figure out how to deal with that in subsequent commits.

* Category: Reliability... Enhancement, New features, Bug fix, Refactoring, Maintenance
* Why these changes are required? So that when the replayer is terminated, gracefully or not, when it picks back up, it is guaranteed to still send all requests at least once.
* What is the old behavior before changes and new behavior after changes?  New behavior should be slightly better stability.  Recovery is still a WIP.

### Issues Resolved
This work doesn't resolve any issues, but it is work toward resolving https://opensearch.atlassian.net/browse/MIGRATIONS-1392

### Testing
See ExhaustiveCapturedTrafficToHttpTransactionAccumulatorTest

### Check List


TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/OrderedStreamLifecyleManager.java
```diff
@@ -0,0 +1,17 @@
...
+    CompletableFuture<Object> futureForLastClose = CompletableFuture.completedFuture(null);
+
+    protected abstract CodedOutputStreamHolder createStream();
+
+    protected CompletableFuture<Object> closeStream(CodedOutputStreamHolder outputStreamHolder, int index) {
+        futureForLastClose = futureForLastClose.thenCompose(v -> kickoffCloseStream(outputStreamHolder, index));
+        return futureForLastClose;
+    }
+
+    protected abstract CompletableFuture<Object> kickoffCloseStream(CodedOutputStreamHolder outputStreamHolder,
```
lewijacn, line 15: Reading this at face value I don't know when I should closeStream or kickoffCloseStream. I initially thought I might have to call kickoffCloseStream before I called closeStream. Is it safe to say this is `closeAllStreamsAsync` and `closeStreamAsync`? Not tied to that name but feel like we need something to make this clearer

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamLifecycleManager.java
```diff
@@ -0,0 +1,9 @@
+package org.opensearch.migrations.trafficcapture;
+
+import java.util.concurrent.CompletableFuture;
+
+public abstract class StreamLifecycleManager {
```
lewijacn, line 5: nit: feels like an interface

lewijacn (commented):
Reviewed for capture-proxy will look at the replayer side tomorrow

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/OrderedStreamLifecyleManager.java
gregschohn: I was not sure about the nomenclature that I chose at the time & I'm less sure of it now.
closeStream is inherited from StreamLifecycleManager.  That's what callers will be using.  The OrderedStreamLifecycleManager implements a version of closeStream() that delegates to the subclass to create a CompletableFuture and kick if off.  Nobody other than the superclass should likely call kickoffCloseStream since that will just return one CompletableFuture, not a CompletableFuture with all of the work chained together.

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamLifecycleManager.java
gregschohn: Thanks - will change.

gregschohn: I think that it was like that before I pulled the ordering out of it.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -175,103 +171,139 @@ private static enum CONNECTION_STATUS {
...
+                                                                      String connectionId,
+                                                                      Instant timestamp) {
+        if (accum.state != Accumulation.State.ACCUMULATING_READS) {
+            return Optional.empty();
+        }
+
         if (observation.hasRead()) {
-            rotateAccumulationOnReadIfNecessary(connectionId, accum);
-            assert accum.state == Accumulation.State.NOTHING_SENT;
+            assert accum.state == Accumulation.State.ACCUMULATING_READS;
```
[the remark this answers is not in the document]
gregschohn: It was more of a product of refactoring.  I'll remove it in part 2.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -175,103 +171,139 @@ private static enum CONNECTION_STATUS {
...
+            accum.rrPair.requestData.finalizeRequestSegments(timestamp);
+        }
+        return Optional.of(CONNECTION_STATUS.ALIVE);
+    }
+
+    private Optional<CONNECTION_STATUS> handleObservationForWriteState(Accumulation accum,
+                                                                       TrafficObservation observation,
+                                                                       String connectionId,
+                                                                       Instant timestamp) {
+        if (accum.state != Accumulation.State.ACCUMULATING_WRITES) {
```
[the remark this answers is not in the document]
gregschohn: This is defensive coding.  handleObservationForReadState() needs to do it since we fall through (actually, to this state).  There is logging at the top level if we come out of the chain without handling the observation (see [line 185](https://github.com/opensearch-project/opensearch-migrations/pull/368/files/9a7139fae08caa69b2e6322a09e671e41bc3ba5e#diff-c11d4821bd721ac56c8c6d0a1f58eec12894b9c17e7c0a45a94465c328d01156R186)

PR #371, opened 2023-10-26
Refactoring to make testing the TrafficReplayer easier.

gregschohn:
### Description

Make the JSONWriter implement a Consumer and have the replayer invoke an arbitrary Consumer so that it's easier to do different kinds of checks.  That also required me to weaken the private restriction on fields in the top-level 'tuple' class so that other writers can access the data. Inline occurrences of TrafficStreamKeyWithRequestOffset.  It was only really used by UniqueRequestKey and it makes more sense and is less error-prone to keep the source and replayer indices side-by-side.  Plus, there's no penalty of having an extra object (locality, allocation).

* Category: Refactoring
* Why these changes are required? See the FullTrafficReplayerTest to see where I'm trying to go with the changes.
* What is the old behavior before changes and new behavior after changes? Should be no difference

### Issues Resolved
None

### Testing
gradlew build

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/Accumulation.java
```diff
@@ -43,8 +41,8 @@ public RequestResponsePacketPair getOrCreateTransactionPair(ITrafficStreamKey tr
         if (rrPair != null) {
             return rrPair;
         }
-        var tskWithOffset = new TrafficStreamKeyWithRequestOffset(trafficStreamKey, startingSourceRequestIndex);
-        return rrPair = new RequestResponsePacketPair(new UniqueRequestKey(tskWithOffset, getIndexOfCurrentRequest()));
+        var urk = new UniqueRequestKey(trafficStreamKey, startingSourceRequestIndex, getIndexOfCurrentRequest());
```
lewijacn, line 44: Back to UniqueRequestKey 😄

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/TrafficStreamGenerator.java
```diff
@@ -0,0 +1,258 @@
...
+import java.util.Optional;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+@Slf4j
+public class TrafficStreamGenerator {
```
lewijacn, line 23: It would be good to know how this operates at a high level. It seems like it could be very helpful, does this create traffic streams with any sequence of TrafficObservations or is there sanity checks to make more legitimate streams?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/UniqueRequestKey.java
```diff
@@ -23,12 +21,12 @@ public int getReplayerRequestIndex() {
 
     @Override
     public String toString() {
-        return trafficStreamKeyAndOffset.getTrafficStreamKey() + "." + getSourceRequestIndex() +
-                (trafficStreamKeyAndOffset.getRequestIndexOffsetAtFirstObservation() == 0 ? "" :
-                        "(offset: "+trafficStreamKeyAndOffset.getRequestIndexOffsetAtFirstObservation()+")");
+        return trafficStreamKey + "." + getSourceRequestIndex() +
+                (sourceRequestIndexOffsetAtFirstObservation == 0 ? "" :
+                        "(offset: "+sourceRequestIndexOffsetAtFirstObservation+")");
```
lewijacn, line 26: nit: offset is a little confusing to me here I would have expected initial although I may be missing the value here. I originally though this was the number of indices from the `sourceRequestInddexOffsetAtFirstObservation` so I guess I thought this was (getSourceRequestIndex() - sourceRequestInddexOffsetAtFirstObservation)

lewijacn (approved):

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/TrafficStreamGenerator.java
gregschohn: Totally agree.  I'll put at least a TODO on it.  The truth is that this thing is just a very specific noise machine at the moment, but it's closer to being a great utility than I had originally realized.

PR #376, opened 2023-11-01
METRICS -- the kickoff PR

mikaylathompson:
### Description
There's a fair amount going on here; mostly all pretty straightforward.
#### Infrastructure side:
- Adds an opensearch analytics cluster (single node) & dashboard to the docker-compose
- Adds an OpenTelemetry Collector box to the docker compose
	- ⚠️ This depends on a pre-built otelcol binary--I don't want to check a huge binary into git, let's discuss how to best handle this in our build system. Right now, this won't build on anyone else's system without some external steps. I think this is an unmergeable state to be in.
	- The box has an `otel-config.yaml` file that specifies that all incoming logs (via otlp) are slightly adjusted (moving some attribute fields) and then pushed to the opensearch analytics cluster, as well as logged on the box itself. The logging is basically a debug thing, and can be stripped out, but is fairly helpful for now.
- Updates the commands for capture-proxy and replayer to include the otel-col hostname/port

#### Code side:
- The MetricsLogger code is updated to use enumerated Events and Attributes. This ensures that we have consistency across the code base, and we can build on this to e.g. automatically generate index mappings down the road.
- Transforms our `log4j2.properties` files to `log4j2.xml` and adds the OpenTelemetry appender
	- This is annoying, to say the least, but I could not get the OpenTelemetry appender to work with a `properties` file, so I ended up with this solution. If there is very strong objection to it, let's discuss.
- Calls an OpenTelemetry initialization function upon startup in the replayer (always) and the capture proxy (if an otel collector endpoint is provided)
- MetricsLogger invocations throughout the codebase are updated to use the new events and attributes.

#### Out of Scope
- Accuracy & completeness of the metrics we're logging. I have simply transformed what we had before to the new API, generally without modifying much of anything about what we're actually tracking. There's a lot of room for improvement here, but it's going to be done in future PRs. If I have made the current state _worse_ anywhere, that's within scope, but otherwise, feel free to note improvements, but I will not consider them blocking changes.
- CDK & AWS deployment -- for the time being, I have focused exclusively on the docker solution. This code should not fail to deploy to AWS, but it also won't include the otel collector box/etc. so there should be no change in behavior. This is clearly a fast-follow from this PR.

### Issues Resolved
[MIGRATIONS-1366](https://opensearch.atlassian.net/browse/MIGRATIONS-1366) and [MIGRATIONS-1367](https://opensearch.atlassian.net/browse/MIGRATIONS-1367)

### Testing
A lot of manual testing. Due to the otelcol binary issue mentioned above, I'm expecting this to fail any integration tests right now, which is something that needs to be fixed before merging.

### Check List


TrafficCapture/dockerSolution/src/main/docker/otelcol/otel-config.yml
```diff
@@ -0,0 +1,66 @@
...
+    scrapers:
+      cpu:
+      disk:
+      filesystem:
+      load:
+      memory:
+      network:
+      process:
+      processes:
+      paging:
```
gregschohn, line 15: Are these being used?  Which machines are we scraping from?

TrafficCapture/dockerSolution/src/main/docker/otelcol/otel-config.yml
```diff
@@ -0,0 +1,66 @@
...
+
+processors:
+  memory_limiter:
+    limit_mib: 300
```
gregschohn, line 19: what does this mean?

TrafficCapture/dockerSolution/src/main/docker/otelcol/otel-config.yml
```diff
@@ -0,0 +1,66 @@
...
+        action: insert
+      - key: log4j.context_data.channel_id
+        action: delete
+
+extensions:
+  basicauth/client:
+    client_auth: 
+      username: "admin"
+      password: "admin"
+  pprof:
```
gregschohn, line 39: what is pprof?

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpRequestHandler.java
```diff
@@ -126,11 +128,10 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
 
     protected void channelFinishedReadingAnHttpMessage(ChannelHandlerContext ctx, Object msg, HttpRequest httpRequest) throws Exception {
         super.channelRead(ctx, msg);
-        metricsLogger.atSuccess()
-                .addKeyValue("channelId", ctx.channel().id().asLongText())
-                .addKeyValue("httpMethod", httpRequest.method().toString())
-                .addKeyValue("httpEndpoint", httpRequest.uri())
-                .setMessage("Full request received").log();
+        metricsLogger.atSuccess(MetricsEvent.RECEIVED_FULL_HTTP_REQUEST)
+                .setAttribute(MetricsAttributeKey.CHANNEL_ID, ctx.channel().id().asLongText())
```
gregschohn, line 132: At some point in the future, it will be nice to have typesafe helpers to do these transformations in one spot.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/coreutils/MetricsLogBuilder.java
```diff
@@ -0,0 +1,45 @@
...
+
+import lombok.extern.slf4j.Slf4j;
+import org.slf4j.event.Level;
+import org.slf4j.Logger;
+import org.slf4j.spi.LoggingEventBuilder;
+
+
+@Slf4j
+public
+class MetricsLogBuilder {
```
gregschohn, line 11: I'm not sure how to do this, but it would be great to get a semantic model (like SonarQube or IntelliJ) to know that there should be a warning if emit has not been called on these built-up values.  It might already do that by virtue of builders being returned and then discarding the last value.  So - maybe it's a future action item, maybe not

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -121,6 +123,12 @@ static class Parameters {
                         "how long after connection should the be recycled " +
                         "(closed with a new connection taking its place)")
         String destinationConnectionPoolTimeout = "PT30S";
+        @Parameter(required = false,
+        names = {"--otelCollectorEndpoint"},
+        arity = 1,
+        description = "Endpoint (host:port) for the OpenTelemetry Collector to which metrics logs should be forwarded." +
```
gregschohn, line 129: does http/https/grpc matter?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -289,6 +297,8 @@ public static void main(String[] args) throws IOException, InterruptedException,
             return;
         }
 
+        initializeOpenTelemetry("traffic-replayer", params.otelCollectorEndpoint);
```
gregschohn, line 300: Why do we initialize this even if the endpoint was null?  What's the behavior when this is null (is there a slight impact, still about the same?)

gregschohn (commented):
I have a number of questions for my own knowledge & to make sure that I understand what I'm looking at.
I have a strong preference to keeping the properties files in the same format.  If we think that the XML format is better, we can switch, but let's get this version through to see what you had to do for otel collection first (as it is, it's a big challenge to see specifically what has changed).

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/coreutils/MetricsEvent.java
```diff
@@ -0,0 +1,30 @@
+package org.opensearch.migrations.coreutils;
+
+public enum MetricsEvent {
```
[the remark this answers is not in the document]
mikaylathompson: I was thinking about this, because grouping by application makes a lot of sense. I won't have this in the about-to-be-pushed update, but I think it makes sense, and I'm looking into the link you included--thanks.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/coreutils/MetricsLogBuilder.java
mikaylathompson: Completely agreed--it would be super nice to have.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/coreutils/MetricsLogger.java
```diff
@@ -22,15 +28,40 @@ public MetricsLogger(String source) {
...
+                                                                        .setEndpoint(collectorEndpoint)
+                                                                        .build())
+                                                        .build())
+                                        .build())
+                        .build();
+        GlobalOpenTelemetry.set(sdk);
+
+        // Add hook to close SDK, which flushes logs
+        Runtime.getRuntime().addShutdownHook(new Thread(sdk::close));
+        OpenTelemetryAppender.install(GlobalOpenTelemetry.get());
```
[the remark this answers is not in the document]
mikaylathompson: I think there may be a missing word(s) in your comment -- are you asking if/when this block will return? Agreed that this needs more investigation & documentation.

TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml
```diff
@@ -13,7 +13,7 @@ services:
     environment:
       - http.port=19200
     # Run processes for elasticsearch and capture proxy, and exit if either one ends
-    command: /bin/sh -c '/usr/local/bin/docker-entrypoint.sh eswrapper & /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.proxyserver.CaptureProxy  --kafkaConnection kafka:9092 --destinationUri  https://localhost:19200  --insecureDestination --listenPort 9200 --sslConfigFile /usr/share/elasticsearch/config/proxy_tls.yml & wait -n 1'
+    command: /bin/sh -c '/usr/local/bin/docker-entrypoint.sh eswrapper & /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.proxyserver.CaptureProxy  --kafkaConnection kafka:9092 --destinationUri  https://localhost:19200  --insecureDestination --listenPort 9200 --sslConfigFile /usr/share/elasticsearch/config/proxy_tls.yml --otelCollectorEndpoint http://otel-collector:4317 & wait -n 1'
```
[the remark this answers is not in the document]
mikaylathompson: Nope, I have no idea. Will check with @lewijacn

TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml
```diff
@@ -91,32 +89,67 @@ services:
...
+  #     - migrations
+  #   ports:
+  #     - "8888:8888"
+  #   volumes:
+  #     - ${TRAFFIC_COMPARATOR_DIRECTORY}:/trafficComparator
+  #     - sharedComparatorSqlResults:/shared
+  #   environment:
+  #     # this needs to match the output db that traffic_comparator writes to
+  #     - COMPARISONS_DB_LOCATION=/shared/comparisons.db
+  #   command: /bin/sh -c 'cd trafficComparator && pip3 install --editable ".[data]" && jupyter notebook --ip=0.0.0.0 --port=8888 --no-browser --allow-root'
```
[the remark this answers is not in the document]
mikaylathompson: Yeah, good point. I'm going to uncomment them for now, because I think they deserve their own commit (or ideally a way to conditionally enable them) but they're somewhat annoying for testing something unrelated at this point. My computer can only handle so many containers at once.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
mikaylathompson: This behavior is subject to discussion, but we have slightly different modes in the capture proxy vs the replayer (based on the fact that we have way more control/consistency for the replayer).

In the capture proxy:
- if an `--otelCollectorEndpoint` is provided, opentel is intialized with the provided value
- if an `--otelCollectorEndpoint` is NOT provided, opentel is never initialized

in the replayer:
- if an `--otelCollectorEndpoint` is provided, opentel is intialized with the provided value
- if an `--otelCollectorEndpoint` is NOT provided, opentel IS initialized with the default value (`http://otel-collector:4137` from line 268)

TrafficCapture/dockerSolution/src/main/docker/opensearchDashboards/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM opensearchproject/opensearch-dashboards:2.7.0
```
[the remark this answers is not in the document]
mikaylathompson: Good point -- the reason this was initially used is that it allows us to load a certificate and use ssl verification. Right now we're not actually taking advantage of that (it was causing some issues), but it may also come in handy to load index mappings and preconfigure a dashboard. If you don't feel strongly, I'd lean towards keeping it this way for now, but it's not actually giving us any benefits in the short term.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: These should stay the same.  We should be able to run the replayer without otel.

TrafficCapture/dockerSolution/src/main/docker/otelcol/otel-config.yml
```diff
@@ -2,23 +2,16 @@ receivers:
...
 
 processors:
-  memory_limiter:
-    limit_mib: 300
   batch:
   attributes:
+    # This processor is currently renaming two attributes
+    # that are prefixed with `log4j.context_data.` to the base attribute name
+    # to make queries within OpenSearch clearer. Both the `insert from_attribute`
+    # and the `delete` actions will fail silently if the attribute is not present,
```
gregschohn, line 12: What does a failure mean?  The event is dropped on the floor but no other impact?

gregschohn (commented):
Can you answer the remaining questions?  I'm eager to know how this new dependency and transient outages to it will impact the reliability of the proxy and replayer.

gregschohn (approved):
I'm approving this to expedite its merge even though I have what would otherwise be blocking comments that are still outstanding.

PR #377, opened 2023-11-01
[Fetch Migration] Handle idle pipeline when target doc count is never reached

kartg:
### Description

This commit introduces changes to track various Data Prepper metrics (as well as API failure counters) in order to detect an idle pipeline or non-responsive Data Prepper subprocess. A new `ProgressMetrics` class has been added to track these metrics and encapsulate detection logic. Much of the migration-success logic from the monitoring module has been moved to this class now. Unit test updates and improvements are also included.

This PR also includes a second commit which refactors/merges the `run` and `monitor_local` functions together (since most of their code/logic is identical) for improved unit test coverage.

* Category
  * Enhancement, bug fix, Refactoring
  * Without these logic changes, the monitoring module would _only_ shutdown the Data Prepper pipeline when the target doc count was reached. If this failed to occur for any reason, or if the Data Prepper API was unresponsive, the overall Fetch Migration workflow would never conclude.

### Testing
```
index_operations.py               37      0   100%
metadata_migration.py            123      0   100%
TOTAL                            405      4    99%
```

### Check List

### DIFF
--- a/FetchMigration/python/migration_monitor.py
+++ b/FetchMigration/python/migration_monitor.py
@@ -91,11 +91,11 @@ def check_and_log_progress(endpoint_info: EndpointInfo, progress: ProgressMetric
     return progress
 
 
-def __should_continue_monitoring(metrics: ProgressMetrics, proc: Optional[Popen] = None) -> bool:
-    return not metrics.is_in_terminal_state() and (proc is None or is_process_alive(proc))
+def __should_continue_monitoring(progress: ProgressMetrics, proc: Optional[Popen] = None) -> bool:
+    return not progress.is_in_terminal_state() and (proc is None or is_process_alive(proc))
 
 
-# Last parameter is optional, and signifies a local Data Prepper process
+# The "dp_process" parameter is optional, and signifies a local Data Prepper process
 def run(args: MigrationMonitorParams, dp_process: Optional[Popen] = None, poll_interval_seconds: int = 30) -> int:
     endpoint_info = EndpointInfo(args.data_prepper_endpoint)
     progress_metrics = ProgressMetrics(args.target_count, __IDLE_THRESHOLD)
--- a/FetchMigration/python/migration_monitor.py
+++ b/FetchMigration/python/migration_monitor.py
@@ -1,6 +1,5 @@
 import argparse
 import logging
-import math
 import subprocess
 import time
 from subprocess import Popen
@@ -12,14 +11,20 @@ from prometheus_client.parser import text_string_to_metric_families
 
 from endpoint_info import EndpointInfo
 from migration_monitor_params import MigrationMonitorParams
+from progress_metrics import ProgressMetrics
 
 # Path to the Data Prepper Prometheus metrics API endpoint
 # Used to monitor the progress of the migration
-__METRICS_API_PATH = "/metrics/prometheus"
-__SHUTDOWN_API_PATH = "/shutdown"
-__DOC_SUCCESS_METRIC = "_opensearch_documentsSuccess"
-__RECORDS_IN_FLIGHT_METRIC = "_BlockingBuffer_recordsInFlight"
-__NO_PARTITIONS_METRIC = "_noPartitionsAcquired"
+__METRICS_API_PATH: str = "/metrics/prometheus"
+__SHUTDOWN_API_PATH: str = "/shutdown"
+__DOC_SUCCESS_METRIC: str = "_opensearch_documentsSuccess"
+__RECORDS_IN_FLIGHT_METRIC: str = "_BlockingBuffer_recordsInFlight"
+__NO_PARTITIONS_METRIC: str = "_noPartitionsAcquired"
+__IDLE_THRESHOLD: int = 5
+
+
+def is_process_alive(proc: Popen) -> bool:
+    return proc.returncode is None
 
 
 # Gracefully shutdown a subprocess
@@ -29,7 +34,7 @@ def shutdown_process(proc: Popen) -> Optional[int]:
     try:
         proc.wait(timeout=60)
     except subprocess.TimeoutExpired:
-        if proc.returncode is None:
+        if is_process_alive(proc):
             # Failed to terminate, send SIGKILL
             proc.kill()
     return proc.returncode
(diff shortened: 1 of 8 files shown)


FetchMigration/python/progress_metrics.py
```diff
@@ -0,0 +1,138 @@
...
+    _METRIC_API_FAIL_KEY: str = "metric_api_fail"
+    _SUCCESS_DOCS_KEY = "success_docs"
+    _REC_IN_FLIGHT_KEY = "records_in_flight"
+    _NO_PART_KEY = "no_partitions"
+
+    target_doc_count: int
+    idle_threshold: int
+    current_values_map: dict[str, Optional[int]]
+    prev_values_map: dict[str, Optional[int]]
+    counter_map: dict[str, int]
```
mikaylathompson, line 22: Minor, but I'm not sure I get why we're using a single dict with key prefixes to track idle values and failures--does it make more sense to keep them separately?

mikaylathompson: This also isn't a blocker for me, but goign forward, I think it's little cleaner to use something like named tuples or dataclasses for these dictionaries. There's (I believe) a fixed set of possible values for each of these, so it seems clearer to a reader + less error prone to use defined lists of possible values.

FetchMigration/python/migration_monitor.py
```diff
@@ -12,14 +11,20 @@
...
-__SHUTDOWN_API_PATH = "/shutdown"
-__DOC_SUCCESS_METRIC = "_opensearch_documentsSuccess"
-__RECORDS_IN_FLIGHT_METRIC = "_BlockingBuffer_recordsInFlight"
-__NO_PARTITIONS_METRIC = "_noPartitionsAcquired"
+__METRICS_API_PATH: str = "/metrics/prometheus"
+__SHUTDOWN_API_PATH: str = "/shutdown"
+__DOC_SUCCESS_METRIC: str = "_opensearch_documentsSuccess"
+__RECORDS_IN_FLIGHT_METRIC: str = "_BlockingBuffer_recordsInFlight"
+__NO_PARTITIONS_METRIC: str = "_noPartitionsAcquired"
+__IDLE_THRESHOLD: int = 5
```
mikaylathompson, line 23: I see how this is used in ProgressMetrics, but I don't have much understanding on where this value comes from and why 5 is the right number. Can you elaborate?

FetchMigration/python/migration_monitor.py
```diff
@@ -60,106 +65,79 @@ def get_metric_value(metric_families: List, metric_suffix: str) -> Optional[int]
...
+            completion_percentage = progress.update_success_doc_count(success_docs)
             progress_message: str = "Completed " + str(success_docs) + \
                                     " docs ( " + str(completion_percentage) + "% )"
             logging.info(progress_message)
+            if progress.all_docs_migrated():
+                logging.info("All documents migrated...")
         else:
+            progress.record_success_doc_value_failure()
             logging.warning("Could not fetch progress stats from Data Prepper response, " +
                             "will retry on next polling cycle...")
```
mikaylathompson, line 87: I think there's still more logic here than I would expect, given ProgressMetrics. Does it make sense for `ProgressMetrics` to have an `update` (or whatever) function that accepts a `metrics` object, so this function can be stripped down to basically:
```py
metrics = metrics = fetch_prometheus_metrics(endpoint_info)
progress.update(metrics)
if progress.all_docs_migrated():
    # do whatever
else:
	# do other stuff.
```

FetchMigration/python/migration_monitor.py
```diff
@@ -60,106 +65,79 @@ def get_metric_value(metric_families: List, metric_suffix: str) -> Optional[int]
...
     else:
+        progress.record_metric_api_failure()
         logging.warning("Data Prepper metrics API call failed, will retry on next polling cycle...")
-    # TODO - Handle idle non-terminal pipeline
-    return terminal, prev_no_partitions_count
+    return progress
+
+
+def __should_continue_monitoring(metrics: ProgressMetrics, proc: Optional[Popen] = None) -> bool:
+    return not metrics.is_in_terminal_state() and (proc is None or is_process_alive(proc))
```
mikaylathompson, line 95: Oof, it is highly confusing that in the previous function, the ProgressMetrics object is called `progress` and there's a different type of object called `metrics`. Can these be standardized?

FetchMigration/python/migration_monitor.py
```diff
@@ -60,106 +65,79 @@ def get_metric_value(metric_families: List, metric_suffix: str) -> Optional[int]
...
+        if dp_process is not None:
+            # Wait on local process
+            try:
+                dp_process.wait(timeout=poll_interval_seconds)
+            except subprocess.TimeoutExpired:
+                pass
+        else:
+            # Thread sleep
+            time.sleep(poll_interval_seconds)
+        if dp_process is None or is_process_alive(dp_process):
```
mikaylathompson, line 113: Why might `dp_process` be None? It seems surprising that woudl be handled in the same case as `is_process_alive == True`

kartg: `dp_process` being `None` means the Data Prepper process is not a subprocess - this allows the migration monitor module to be used against remote Data Prepper processes as well. While the Fetch Migration solution today only uses a local subprocess, this flexibility allows the migration monitor to be used in other scenarios.

Here, `dp_process is None` functions as short-circuit logic for the `is_process_alive` check - which shouldn't be run unless the Data Prepper process is local.

The comment preceding the `run` method was incorrect, so I updated this

FetchMigration/python/progress_metrics.py
kartg: > Minor, but I'm not sure I get why we're using a single dict with key prefixes to track idle values and failures--does it make more sense to keep them separately?

Yes, these could be stored in separate structures In my mind, they're both counter values - one for the number of times a metric value is idle, and the other for the number of failures - but i can see how this conflation makes the code harder to read.

I'll look into incorporating this, as well as your feedback on stronger control of possible values, in a follow-up PR.

FetchMigration/python/progress_metrics.py
```diff
@@ -0,0 +1,138 @@
...
+            self.prev_values_map[key] = self.current_values_map[key]
+            # Store new value
+            self.current_values_map[key] = val
+            # Track idle value metrics
+            idle_value_key = self.__get_idle_value_key_name(key)
+            if self.prev_values_map[key] == val:
+                self.__increment_counter(idle_value_key)
+            else:
+                self.__reset_counter(idle_value_key)
+        self.current_values_map[key] = val
```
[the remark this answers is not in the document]
kartg: Good spot, removed the duplicate assignment

FetchMigration/python/progress_metrics.py
```diff
@@ -0,0 +1,138 @@
...
+    def is_in_terminal_state(self) -> bool:
+        return self.is_migration_complete_success() or self.is_migration_idle() or self.is_too_may_api_failures()
+
+    def log_idle_pipeline_debug_metrics(self):  # pragma no cover
+        if logging.getLogger().isEnabledFor(logging.DEBUG):
+            debug_msg_template: str = "Idle pipeline metrics - " + \
+                                      "Records in flight: [{0}], " + \
+                                      "No-partitions counter: [{1}]" + \
+                                      "Previous no-partition value: [{2}]"
+            logging.debug(debug_msg_template.format(self.__get_current_value(self._REC_IN_FLIGHT_KEY),
```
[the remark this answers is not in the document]
kartg: D'oh! I forgot about those after using them elsewhere in the Fetch codebase. Updated.

mikaylathompson (approved):

PR #380, opened 2023-11-02
Linting round3

gregschohn:
### Description
More improvements to fix issues caught in running a lint checker.

### Testing
./gradlew allTests

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
@@ -11,6 +11,7 @@ import lombok.AllArgsConstructor;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.opensearch.migrations.replay.datahandlers.IPacketFinalizingConsumer;
+import org.opensearch.migrations.replay.datahandlers.http.IHttpMessage;
 import org.opensearch.migrations.replay.datatypes.HttpRequestTransformationStatus;
 import org.opensearch.migrations.replay.datatypes.ITrafficStreamKey;
 import org.opensearch.migrations.replay.datatypes.TransformedPackets;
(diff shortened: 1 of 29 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -308,20 +310,19 @@ public static void main(String[] args) throws IOException, InterruptedException,
...
-                    var timeShifter = new TimeShifter(params.speedupFactor);
-                    tr.runReplayWithIOStreams(Duration.ofSeconds(params.observedPacketConnectionTimeout),
-                            blockingTrafficStream, bufferedOutputStream, timeShifter, tupleWriter);
-                    log.info("Done processing TrafficStreams");
-                }
-            }
+                new FileOutputStream(params.outputFilename, true);
+             var bufferedOutputStream = new BufferedOutputStream(outputStream);
+             var blockingTrafficStream = TrafficCaptureSourceFactory.createTrafficCaptureSource(params,
+                     Duration.ofSeconds(params.lookaheadTimeSeconds));
```
lewijacn, line 316: Can we add another line here or something to distinguish that the try parenthesis has now ended

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -418,6 +428,8 @@ void runReplayWithIOStreams(Duration observedPacketConnectionTimeout,
                         new TrafficReplayerAccumulationCallbacks(replayEngine, resultTupleConsumer));
         try {
             pullCaptureFromSourceToAccumulator(trafficChunkStream, trafficToHttpTransactionAccumulator);
+        } catch (InterruptedException ex) {
+            throw ex;
```
lewijacn, line 432: Do we need to add a `Thread.currentThread().interrupt();` here?

lewijacn (approved):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: No.  The general rule is if you're suppressing that this was an interruption, you should reset the interrupt state.  Since we're propagating (so that we don't obscure it), the final catch-receiver can handle re-interrupting.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: You & I think alike ;).
I got fed up with too many resources and not knowing where the distinction was & did that in the last delinting PR.

PR #385, opened 2023-11-03
Fixing Multi-Proxy build

okhasawn:
### Description
This PR restores the ability to create a setup that has 2 coordinator nodes with a Capture Proxy on each node.
This is needed after a change to the Dockerfile that creates the ES image was made, along with renaming the Capture Proxy.

### Check List


TrafficCapture/dockerSolution/src/main/docker/docker-compose-multi.yml
```diff
@@ -14,8 +14,9 @@ services:
       - cluster.name=es-cluster
       - discovery.seed_hosts=capture-proxy-es2
       - cluster.initial_master_nodes=capture-proxy-es1,capture-proxy-es2
+      - discovery.type=zen
```
lewijacn, line 17: We should probably add a comment, that different versions of elasticsearch might not use this discovery type and use "multi-node"

lewijacn (approved):

PR #387, opened 2023-11-03
[Fetch Migration] CDK bugfix and other minor updates

kartg:
### Description
* Category: Bug fix
> Why these changes are required?

The Fetch Migration stack now uses `StringParameter.valueFromLookup` to resolve the target cluster endpoint from Parameter Store. This is required instead of the `StringParameter.valueForStringParameter` that was previously used because _valueForStringParameter_ returns a `Token` that is only resolved at deployment time. 

This PR also includes two other minor updates:
* Bumping the Fetch Migration task definition to 1 CPU and 4 GB of memory
* Additional documentation in the pipeline template file to aid readability

> What is the old behavior before changes and new behavior after changes?

Since the value of the target cluster endpoint is substituted into the pipeline template file prior to deployment time, a string representation of the Token object (eg. `${Token[TOKEN.123]}`) was being written instead of the actual endpoint URI. Using _valueFromLookup_ looks up the value immediately, so the substitution occurs correctly with the bugfix.

### Issues Resolved
N/A

### Testing
Bug identified while using CDK. Bugfix verified by re-deploying CDK and verifying that the correct value is now used.

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts
@@ -26,7 +26,7 @@ export class FetchMigrationStack extends Stack {
         super(scope, id, props);
 
         // Import required values
-        const targetClusterEndpoint = StringParameter.valueForStringParameter(this, `/migration/${props.stage}/${props.defaultDeployId}/osClusterEndpoint`)
+        const targetClusterEndpoint = StringParameter.valueFromLookup(this, `/migration/${props.stage}/${props.defaultDeployId}/osClusterEndpoint`)
         const domainAccessGroupId = StringParameter.valueForStringParameter(this, `/migration/${props.stage}/${props.defaultDeployId}/osAccessSecurityGroupId`)
         // This SG allows outbound access for ECR access as well as communication with other services in the cluster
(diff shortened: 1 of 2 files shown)


deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts
```diff
@@ -26,7 +26,7 @@ export class FetchMigrationStack extends Stack {
         super(scope, id, props);
 
         // Import required values
-        const targetClusterEndpoint = StringParameter.valueForStringParameter(this, `/migration/${props.stage}/${props.defaultDeployId}/osClusterEndpoint`)
+        const targetClusterEndpoint = StringParameter.valueFromLookup(this, `/migration/${props.stage}/${props.defaultDeployId}/osClusterEndpoint`)
```
lewijacn, line 29: I don't think this makes sense. Value from lookup requires that the value be there at synthesis time.

kartg:
Moving to draft PR based on [this comment](https://github.com/opensearch-project/opensearch-migrations/pull/387#discussion_r1382141596)

kartg:
Closing in favor of https://github.com/opensearch-project/opensearch-migrations/pull/398

PR #390, opened 2023-11-06
[Fetch Migration] Enable migration for identical, empty target cluster index

kartg:
### Description
* Category: Enhancement
>  Why these changes are required?

This change enables Fetch Migration to migrate data for indices that are identical between the source and target cluster, _and_ the document count on the target cluster is zero. Such behavior allow users to execute Fetch Migration in an idempotent manner without having to manually remove empty indices on the target cluster (that were created by a previous run of Fetch Migration that did not complete successfully)

> What is the old behavior before changes and new behavior after changes?

Prior to this change, all indices that were identical between the source and target clusters were ineligible for data migration. Now, we check the doc_count for identical indices on the target cluster and only exclude identical indices with a non-zero document count.

### Testing
```
index_diff.py                     20      0   100%
index_doc_count.py                 5      0   100%
index_operations.py               49      0   100%
metadata_migration.py            118      0   100%
TOTAL                            437      4    99%

```

Also tested end-to-end via the CDK. Log output from a test run:

```
INFO:root:Running pre-migration steps...

Identical indices in the target cluster: [logs-241998, sonested, logs-181998, logs-221998, reindexed-logs, nyc_taxis, geonames, logs-191998, logs-201998, logs-231998, logs-211998]

Identical empty indices in the target cluster (data will be migrated): [sonested, reindexed-logs, nyc_taxis, geonames]

Indices present in both clusters with conflicting settings/mappings (data will not be migrated): []

Indices to be created in the target cluster (data will be migrated): []

Total number of documents to be moved: 3000

...
```

### Check List

### DIFF
--- a/FetchMigration/python/index_diff.py
+++ b/FetchMigration/python/index_diff.py
@@ -1,0 +1,31 @@
+import utils
+from index_operations import SETTINGS_KEY, MAPPINGS_KEY
+
+
+# Computes and captures differences in indices between a "source" cluster
+# and a "target" cluster. Indices that exist on the source cluster but not
+# on the target cluster are considered "to-create". "Conflicting" indices
+# are present on both source and target clusters, but differ in their index
+# settings or mappings.
+class IndexDiff:
+    indices_to_create: set
+    identical_indices: set
+    identical_empty_indices: set
+    conflicting_indices: set
+
+    def __init__(self, source: dict, target: dict):
+        self.identical_empty_indices = set()
+        self.conflicting_indices = set()
+        indices_in_target = set(source.keys()) & set(target.keys())
+        for index in indices_in_target:
+            # Check settings
+            if utils.has_differences(SETTINGS_KEY, source[index], target[index]):
+                self.conflicting_indices.add(index)
+            # Check mappings
+            if utils.has_differences(MAPPINGS_KEY, source[index], target[index]):
+                self.conflicting_indices.add(index)
+        self.identical_indices = set(indices_in_target) - set(self.conflicting_indices)
+        self.indices_to_create = set(source.keys()) - set(indices_in_target)
+
+    def set_identical_empty_indices(self, indices: set):
+        self.identical_empty_indices = indices
--- a/FetchMigration/python/index_diff.py
+++ b/FetchMigration/python/index_diff.py
@@ -16,16 +16,20 @@ class IndexDiff:
     def __init__(self, source: dict, target: dict):
         self.identical_empty_indices = set()
         self.conflicting_indices = set()
-        indices_in_target = set(source.keys()) & set(target.keys())
-        for index in indices_in_target:
+        # Compute index names that are present in both the source and target
+        indices_intersection = set(source.keys()) & set(target.keys())
+        # Check if these "common" indices are identical or have metadata conflicts
+        for index in indices_intersection:
             # Check settings
             if utils.has_differences(SETTINGS_KEY, source[index], target[index]):
                 self.conflicting_indices.add(index)
             # Check mappings
             if utils.has_differences(MAPPINGS_KEY, source[index], target[index]):
                 self.conflicting_indices.add(index)
-        self.identical_indices = set(indices_in_target) - set(self.conflicting_indices)
-        self.indices_to_create = set(source.keys()) - set(indices_in_target)
+        # Identical indices are the subset that do not have metadata conflicts
+        self.identical_indices = set(indices_intersection) - set(self.conflicting_indices)
+        # Indices that are not already on the target need to be created
+        self.indices_to_create = set(source.keys()) - set(indices_intersection)
 
     def set_identical_empty_indices(self, indices: set):
         self.identical_empty_indices = indices
--- a/FetchMigration/python/index_operations.py
+++ b/FetchMigration/python/index_operations.py
@@ -1,14 +1,22 @@
+import jsonpath_ng
 import requests
 
 from endpoint_info import EndpointInfo
 
 # Constants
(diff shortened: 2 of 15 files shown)


deployment/cdk/opensearch-service-migration/dp_pipeline_template.yaml
```diff
@@ -1,18 +1,36 @@
+# Name of the Data Prepper pipeline
 historical-data-migration:
+  # Source cluster configuration
   source:
     opensearch:
+      # DO NOT CHANGE the hosts value - this will be updated by the CDK
```
sumobrian, line 6: I would add a note that they would need to change this only if running locally (in Docker)

deployment/cdk/opensearch-service-migration/dp_pipeline_template.yaml
```diff
@@ -1,18 +1,36 @@
...
       disable_authentication: true
       indices:
+        # Indices to exclude - exclude system indices by default
         exclude:
         - index_name_regex: \.*
+  # Target cluster configuration
   sink:
   - opensearch:
-      bulk_size: 10
+      # DO NOT CHANGE the hosts value - this will be updated by the CDK
```
sumobrian, line 20: Same as comment above: I would add a note that they would need to change these only if running locally (in Docker)

sumobrian (changes requested):

FetchMigration/python/index_operations.py
```diff
@@ -42,8 +50,14 @@ def create_indices(indices_data: dict, endpoint: EndpointInfo):
             raise RuntimeError(f"Failed to create index [{index}] - {e!s}")
 
 
-def doc_count(indices: set, endpoint: EndpointInfo) -> int:
-    actual_endpoint = endpoint.url + ','.join(indices) + __COUNT_ENDPOINT
-    resp = requests.get(actual_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl)
+def doc_count(indices: set, endpoint: EndpointInfo) -> IndexDocCount:
+    actual_endpoint = endpoint.url + ','.join(indices) + __SEARCH_COUNT_PATH
+    resp = requests.get(actual_endpoint, auth=endpoint.auth, verify=endpoint.verify_ssl, json=__SEARCH_COUNT_PAYLOAD)
```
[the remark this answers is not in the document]
kartg: I've got an [open issue](https://opensearch.atlassian.net/browse/MIGRATIONS-1292) tracking such fixes across the implementation

FetchMigration/python/metadata_migration.py
```diff
@@ -99,50 +100,30 @@ def validate_pipeline_config(config: dict):
...
-# (indices to create), (identical indices), (indices with conflicts)
-def print_report(index_differences: tuple[set, set, set], count: int):  # pragma no cover
-    print("Identical indices in the target cluster (no changes will be made): " +
-          utils.string_from_set(index_differences[1]))
-    print("Indices in target cluster with conflicting settings/mappings: " +
-          utils.string_from_set(index_differences[2]))
-    print("Indices to create: " + utils.string_from_set(index_differences[0]))
-    print("Total documents to be moved: " + str(count))
+def print_report(diff: IndexDiff, total_doc_count: int):  # pragma no cover
+    print("Identical indices in the target cluster: " +
```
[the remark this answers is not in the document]
kartg: Updated all `print` statements to use `logging`

deployment/cdk/opensearch-service-migration/dp_pipeline_template.yaml
kartg: Updated comment to:

```
# CDK code will replace this value, so DO NOT CHANGE this
# unless the file is being used outside of the CDK
```

deployment/cdk/opensearch-service-migration/dp_pipeline_template.yaml
kartg: After the [most recent updates to the Fetch Migration CDK](https://github.com/opensearch-project/opensearch-migrations/commit/353ab1168cee47338d4b51f5bf65acc8e2f5213f), the `hosts` value in the `sink` will be replaced by an [inline target host value](https://github.com/opensearch-project/opensearch-migrations/blob/44978733f3bc0cb3a61981daf4c078ebdf33b4e2/deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts#L84) through the CDK, so changing this value will not affect that logic nor reflect in the final setup. I've updated the comment here to:

```
# Note - CDK code will replace this value with the target cluster endpoint
```

This is in contrast to the `hosts` value in `source` where the string value matters because the CDK [performs a string substitution](https://github.com/opensearch-project/opensearch-migrations/blob/44978733f3bc0cb3a61981daf4c078ebdf33b4e2/deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts#L71) to inject the source cluster endpoint.

FetchMigration/python/index_diff.py
```diff
@@ -0,0 +1,31 @@
...
+class IndexDiff:
+    indices_to_create: set
+    identical_indices: set
+    identical_empty_indices: set
+    conflicting_indices: set
+
+    def __init__(self, source: dict, target: dict):
+        self.identical_empty_indices = set()
+        self.conflicting_indices = set()
+        indices_in_target = set(source.keys()) & set(target.keys())
```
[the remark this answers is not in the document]
kartg: Renamed to `indices_intersection` and added comments to explain the logic here

sumobrian (approved):

PR #372, opened 2023-10-26
[Fetch Migration] Improvements to subprocess handling

kartg:
### Description

The `migration_monitor.py` module now includes a separate function (`monitor_local`) to monitor a Data Prepper subprocess. 

Previously, the migration monitor only interacted with the Data Prepper process via API calls. This has the limitation that the monitor cannot differentiate between the subprocess intermittently not responding (or not yet started) versus a non-running subprocess. Thus, the API calls simply retry indefinitely if no response is received - which resulted in a indefinite polling loop if the subprocess had crashed or ended (say due to a bad configuration, OOM error, etc.)

The new function adds logic to poll the running subprocess and its return code (via Python's `Popen` object) - this is in addition to the monitoring via API calls. Unit tests for these use-cases have been added/updated.

This PR also includes an additional commit which adds a fix to `index_operations.py` to filter out system indices (so they are excluded from metadata migration). Short-circuit logic for empty source clusters has been added, along with related unit tests.

### Testing
```
$ python -m coverage report --omit "*/tests/*"
Name                           Stmts   Miss  Cover
--------------------------------------------------
endpoint_info.py                   6      0   100%
fetch_orchestrator.py             25      0   100%
index_operations.py               37      0   100%
metadata_migration.py            125      1    99%
metadata_migration_params.py       7      0   100%
metadata_migration_result.py       5      0   100%
migration_monitor.py              96      3    97%
migration_monitor_params.py        5      0   100%
utils.py                          13      0   100%
--------------------------------------------------
TOTAL                            319      4    99%

```

### Check List
- [x] New functionality includes testing
  - [x] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #398, opened 2023-11-09
[Fetch Migration] Support for dynamic target host substitution via environment variable, and fetch_orchestrator updates

kartg:
### Description

This PR includes a few changes:

1 - Fetch Migration now supports supplying the target cluster endpoint via environment variable (`INLINE_TARGET_HOST`) similar to `INLINE_PIPELINE` behavior. If present, the value of this env variable is written as the hostname for the Data Prepper sink in the Data Prepper pipeline configuration file _before_ kicking off any steps in the Fetch Migration workflow.

2 - The CDK has been changed to utilize the above environment variable. This mitigates the [bug described here](https://github.com/opensearch-project/opensearch-migrations/pull/387)

3 - `fetch_orchestrator` has been updated to assume a local Data Prepper process, so it no longer accepts a Data Prepper "endpoint" URL but constructs this based on input parameters (port value and SSL flag). This incorporates [PR feedback from a previous change](https://github.com/opensearch-project/opensearch-migrations/pull/372#discussion_r1375003829)

3 - `fetch_orchestrator` now accepts additional flags that let it only execute metadata migration (and skip data migration). This is useful to run index setup prior to performing a Capture & Replay migration.

### Testing
CDK synth snippet:

```
...
      "Name": "fetch-migration",
      "Secrets": [
       {
        "Name": "INLINE_PIPELINE",
        "ValueFrom": {
         "Ref": "<pipeline-config-secret>"
        }
       },
       {
        "Name": "INLINE_TARGET_HOST",
        "ValueFrom": "<os-cluster-endpoint-arn>"
       }
      ]
     }
    ],
...
```

Code coverage:

```
fetch_orchestrator.py             71      0   100%
fetch_orchestrator_params.py      22      0   100%
index_operations.py               37      0   100%
metadata_migration.py            123      0   100%
TOTAL                            473      4    99%
```

### Check List
  - [] All tests pass, including unit test, integration test and doctest

### DIFF
--- a/FetchMigration/Dockerfile
+++ b/FetchMigration/Dockerfile
@@ -18,5 +18,6 @@ ENV PATH=/root/.local:$PATH
 RUN echo "ssl: false" > $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 RUN echo "metricRegistries: [Prometheus]" >> $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 
-# Include the -u flag to have stdout logged
-ENTRYPOINT python3 -u ./fetch_orchestrator.py --insecure $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml
+# Include the -u flag to have unbuffered stdout (for detached mode)
+# "$0" "$@" allows passing extra args/flags to the entrypoint when the image is run
+ENTRYPOINT python3 -u ./fetch_orchestrator.py --insecure $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml "$0" "$@"
--- a/FetchMigration/Dockerfile
+++ b/FetchMigration/Dockerfile
@@ -19,5 +19,5 @@ RUN echo "ssl: false" > $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 RUN echo "metricRegistries: [Prometheus]" >> $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 
 # Include the -u flag to have unbuffered stdout (for detached mode)
-# "$0" "$@" allows passing extra args/flags to the entrypoint when the image is run
-ENTRYPOINT python3 -u ./fetch_orchestrator.py --insecure $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml "$0" "$@"
+# "$@" allows passing extra args/flags to the entrypoint when the image is run
+ENTRYPOINT python3 -u ./fetch_orchestrator.py --insecure $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml "$@"
--- a/FetchMigration/Dockerfile
+++ b/FetchMigration/Dockerfile
@@ -19,4 +19,4 @@ RUN echo "ssl: false" > $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 RUN echo "metricRegistries: [Prometheus]" >> $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 
 # Include the -u flag to have stdout logged
-ENTRYPOINT python3 -u ./fetch_orchestrator.py $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml http://localhost:4900
+ENTRYPOINT python3 -u ./fetch_orchestrator.py --insecure $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml
--- a/FetchMigration/python/fetch_orchestrator.py
+++ b/FetchMigration/python/fetch_orchestrator.py
@@ -2,20 +2,79 @@ import argparse
 import base64
 import logging
 import os
+import re
 import subprocess
 import sys
 from typing import Optional
+
+import yaml
 
 import metadata_migration
 import migration_monitor
 from metadata_migration_params import MetadataMigrationParams
 from migration_monitor_params import MigrationMonitorParams
 
+__PROTOCOL_PREFIX_PATTERN = re.compile(r"^https?://")
+__HTTPS_PREFIX = "https://"
 __DP_EXECUTABLE_SUFFIX = "/bin/data-prepper"
 __PIPELINE_OUTPUT_FILE_SUFFIX = "/pipelines/pipeline.yaml"
 
 
+def __get_env_string(name: str) -> Optional[str]:
+    val: str = os.environ.get(name, "")
+    if len(val) > 0:
+        return val
+    else:
+        return None
+
+
+def update_target_host(dp_config: dict, target_host: str):
+    # Inline target host only supports HTTPS, so force it
+    target_with_protocol = target_host
+    match = __PROTOCOL_PREFIX_PATTERN.match(target_with_protocol)
+    if match:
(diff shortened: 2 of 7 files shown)


FetchMigration/Dockerfile
```diff
@@ -18,5 +18,6 @@ ENV PATH=/root/.local:$PATH
 RUN echo "ssl: false" > $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 RUN echo "metricRegistries: [Prometheus]" >> $DATA_PREPPER_PATH/config/data-prepper-config.yaml
 
-# Include the -u flag to have stdout logged
-ENTRYPOINT python3 -u ./fetch_orchestrator.py $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml http://localhost:4900
+# Include the -u flag to have unbuffered stdout (for detached mode)
+# "$0" "$@" allows passing extra args/flags to the entrypoint when the image is run
+ENTRYPOINT python3 -u ./fetch_orchestrator.py --insecure $DATA_PREPPER_PATH $FM_CODE_PATH/input.yaml "$0" "$@"
```
lewijacn, line 23: Are we currently adding any extra args/flags to the entrypoint?

deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts
```diff
@@ -67,8 +67,8 @@ export class FetchMigrationStack extends Stack {
 
         // Create DP pipeline config from template file
         let dpPipelineData: string = readFileSync(props.dpPipelineTemplatePath, 'utf8');
+        // Replace only source cluster host - target host will be overridden by inline env var
         dpPipelineData = dpPipelineData.replace("<SOURCE_CLUSTER_HOST>", props.sourceEndpoint);
-        dpPipelineData = dpPipelineData.replace("<TARGET_CLUSTER_HOST>", targetClusterEndpoint);
```
lewijacn, line 71: This part is a bit confusing for me. We will no longer replace <TARGET_CLUSTER_HOST> in the yaml file but it still exists. Do we replace it in the Python somewhere with the environment variable (I looked but may have missed this)?

lewijacn: On a different note, we should add a small note in the README (or if we can do a comment in the pipeline yaml even better) for setting up FetchMigration that a user does not need to fill in the `SOURCE_CLUSTER_HOST` or `TARGET_CLUSTER_HOST` values in the yaml but that they will be autofilled by us.

FetchMigration/python/fetch_orchestrator.py
```diff
@@ -2,33 +2,100 @@
...
-    logging.info("Running pre-migration steps...\n")
+def __get_env_string(name: str) -> Optional[str]:
+    val: str = os.environ.get(name, "")
+    if len(val) > 0:
+        return val
+    else:
+        return None
+
+
+def update_target_host(dp_config: dict, target_host: str):
```
lewijacn, line 32: Just to confirm this support urls like: `https:vpc-123:443` and `vpc-123`. I have an upcoming change that plans to store a full url in SSM versus the shortened form

FetchMigration/python/fetch_orchestrator.py
```diff
@@ -2,33 +2,100 @@
...
+        # Note - this does not preserve comments
+        yaml.safe_dump(pipeline_yaml, pipeline_file)
+
+
+def run(params: FetchOrchestratorParams) -> Optional[int]:
+    # This is expected to be a base64 encoded string
+    inline_pipeline = __get_env_string("INLINE_PIPELINE")
+    inline_target_host = __get_env_string("INLINE_TARGET_HOST")
+    if inline_pipeline is not None:
+        write_inline_pipeline(params.pipeline_file_path)
```
lewijacn, line 76: nit: seems like you could have passed the two env variables as parameters so this function didn't have to retrieve them

deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts
kartg: > This part is a bit confusing for me. We will no longer replace <TARGET_CLUSTER_HOST> in the yaml file but it still exists. Do we replace it in the Python somewhere with the environment variable (I looked but may have missed this)?

Yes, that's what the [write_inline_target_host function](https://github.com/kartg/opensearch-migrations/blob/1333db5de17876e2767667c9e5f7034b18adb336/FetchMigration/python/fetch_orchestrator.py#L63) does. `<TARGET_CLUSTER_HOST>` no longer needs to be specified because we're not performing a simple string substitution anymore - rather, the Python function parses the Data Prepper pipeline definition structurally and replaces the `host` value of the appropriate `sink` plugin

kartg: > On a different note, we should add a small note in the README (or if we can do a comment in the pipeline yaml even better) for setting up FetchMigration that a user does not need to fill in the SOURCE_CLUSTER_HOST or TARGET_CLUSTER_HOST values in the yaml but that they will be autofilled by us.

Included in a parallel PR 😄

https://github.com/opensearch-project/opensearch-migrations/pull/390/files#diff-4e4fd6e3cb3114763fe0616b8fb3d9af7dacae3d58faee516575d30054ac599d

FetchMigration/Dockerfile
kartg: Not currently, but given that metadata migration is essential for Traffic Capture & Replay to replicate data accurately, IMO its useful to enable the Docker image to be run with the `--createonly` flag as a pre-requisite step.

FetchMigration/python/fetch_orchestrator.py
kartg: `https://vpc-123:443` would be left as-is (since it's already using HTTPS) but `vpc-123` would be changed to `https://vpc-123` to force HTTPS. Would that be an issue for your change?

lewijacn: The first option was supposed to be `https://vpc-123:443`

deployment/cdk/opensearch-service-migration/lib/fetch-migration-stack.ts
lewijacn: Great thanks!

FetchMigration/python/fetch_orchestrator.py
lewijacn: Sounds good wanted to ensure we could support both. We use the second currently but will be moving to the first.

lewijacn (approved):

FetchMigration/python/fetch_orchestrator.py
kartg: Incorporated

kartg:
CodeCov failures are in [Java code](https://app.codecov.io/gh/opensearch-project/opensearch-migrations/pull/398/indirect-changes), which this PR does not touch. I'll go ahead and merge this.

PR #400, opened 2023-11-09
BYOC - Bring Your Own Cluster

lewijacn:
### Description
A lesser known acronym for allowing an existing cluster to be used with the Migration solution. This change supports providing a endpoint url of an existing target cluster, and then using that endpoint throughout the Migration solution. This results in no Domain stack being created and an SSM parameter for the endpoint being created as it normally would.

A couple caveats that are worth pointing out. We currently do not support bringing in existing security groups for imported target clusters. This is a current limitation around supporting multiple replay scenarios all from a single migration console, with each additional replay scenario needing to potentially update the security groups of that console container being one issue. With imported target clusters, we still create a default security group that is linked to each of our applicable Migration services (replayer, console, fetch). It will not make any changes to the target cluster, and as such would be the responsibility of the user to add this security group to their target cluster if needed (this will be included in documentation).

Additionally IAM permissions have been added for accessing OpenSearch and OpenSearch Serverless AWS services with SigV4 to applicable Migration services (replayer, console, fetch).  Since this policy is given to the service containers and not placed on the actual cluster resource, it is more generic in the resources it allows, with this currently being any cluster in the given region and account. For these IAM permissions to be useful for SigV4 requests to the target cluster, access policies on the target cluster must allow the IAM task roles of these Migration services (this is included in documentation).

There is also a minor change here for storing the full domain URL in SSM and not just the host name for 

~~Deployment testing still in progress as well as updating the documentation for users~~

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1361

### Testing
Deployment testing and jest testing

### Check List


gregschohn:
For clarification concerning "We currently do not support bringing in existing security groups for imported target clusters."
Do we allow users to specify an existing VPC & security group when deploying the fetch/console/replayer initially?

deployment/cdk/opensearch-service-migration/README.md
```diff
@@ -126,17 +126,36 @@ After the benchmark has been run, the indices and documents of the source and ta
...
 
+See [Configuring SigV4 Replayer Requests](#configuring-sigv4-replayer-requests) for details on enabling SigV4 requests from the Traffic Replayer to the target cluster after following the previous setup
+
+## Configuring SigV4 Replayer Requests
+With any [required setup](#importing-target-clusters) on the target cluster having been completed, a user can then use the `trafficReplayerExtraArgs` option to specify the Traffic Replayer argument for enabling SigV4 auth, which the below sections show. **Note**: The `trafficReplayerEnableClusterFGACAuth` option should not be used if enabling SigV4 auth for the Traffic Replayer as only one auth mechanism should be specified. See [here](#how-is-an-authorization-header-set-for-requests-from-the-replayer-to-the-target-cluster) for more details on how the Traffic Replayer sets its auth header.
+#### OpenSearch Service
+```shell
+# e.g. --sigv4-auth-header-service-region es,us-east-1
+"trafficReplayerExtraArgs": "--sigv4-auth-header-service-region es,<region>"
```
gregschohn, line 151: Put the region in all caps, it will be more noticeable so that a user realizes that they need to replace it.

deployment/cdk/opensearch-service-migration/README.md
```diff
@@ -126,17 +126,36 @@ After the benchmark has been run, the indices and documents of the source and ta
...
 
+See [Configuring SigV4 Replayer Requests](#configuring-sigv4-replayer-requests) for details on enabling SigV4 requests from the Traffic Replayer to the target cluster after following the previous setup
+
+## Configuring SigV4 Replayer Requests
+With any [required setup](#importing-target-clusters) on the target cluster having been completed, a user can then use the `trafficReplayerExtraArgs` option to specify the Traffic Replayer argument for enabling SigV4 auth, which the below sections show. **Note**: The `trafficReplayerEnableClusterFGACAuth` option should not be used if enabling SigV4 auth for the Traffic Replayer as only one auth mechanism should be specified. See [here](#how-is-an-authorization-header-set-for-requests-from-the-replayer-to-the-target-cluster) for more details on how the Traffic Replayer sets its auth header.
```
gregschohn, line 147: This is confusing.  I think that dropping the second word, "any" for "the" helps simplify it.
The trafficReplayerExtraArgs are args to what?  
"as only one auth mechanism should be specified"... Rotating/splitting this sentence might help.... Only one auth (authN or authZ?) mechanism, FGAC/SigV4..., can be used.  
Is SigV4 referred to as SigV4 or AWS Auth in the service documentation?

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
```diff
@@ -9,13 +9,38 @@ import {StackPropsExt} from "./stack-composer";
...
 
 export class NetworkStack extends Stack {
     public readonly vpc: IVpc;
 
+    // Validate a proper url string is provided and return an url string which contains a protocol, host name, and port.
+    // If a port is not provided, the default protocol port (e.g. 443, 80) will be explicitly added
+    static validateAndReturnFormattedHttpURL(urlString: string) {
+        // URL will throw error if the urlString is invalid
+        let url = new URL(urlString);
+        if (url.protocol !== "http:" && url.protocol !== "https:") {
```
gregschohn, line 25: the colon is included in the protocol field?

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
```diff
@@ -9,13 +9,38 @@ import {StackPropsExt} from "./stack-composer";
...
+            throw new Error(`Invalid url protocol for target endpoint: ${urlString} was expecting 'http' or 'https'`)
+        }
+        // URLs that contain the default protocol port (e.g. 443, 80) will not show in the URL toString()
+        let formattedUrlString = url.toString()
+        if (formattedUrlString.endsWith("/")) {
```
gregschohn, line 30: If a users specifies http://host/directory/ you'll end up keeping 'directory'.
I'd parse the URL as you do, then check that the path/query string is empty (throw an error if it isn't) and then reconstruct the URL by concatenating the protocol, host, port (with a default if it wasn't included).  Formatting a string and parsing it is tedious and error prone.

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
```diff
@@ -78,5 +103,15 @@ export class NetworkStack extends Stack {
             });
         }
 
+        if (props.targetClusterEndpoint) {
+            const formattedClusterEndpoint = NetworkStack.validateAndReturnFormattedHttpURL(props.targetClusterEndpoint)
+            const deployId = props.addOnMigrationDeployId ? props.addOnMigrationDeployId : props.defaultDeployId
```
gregschohn, line 108: What is an addOnMigrationDeployId?  The scalar (rather than collection) type is confusing me.

deployment/cdk/opensearch-service-migration/lib/service-stacks/traffic-replayer-stack.ts
```diff
@@ -89,10 +89,11 @@ export class TrafficReplayerStack extends MigrationServiceCore {
                 "secretsmanager:DescribeSecret"
             ]
         })
+        const openSearchPolicy = createOpenSearchIAMAccessPolicy(<string>props.env?.region, <string>props.env?.account)
+        const openSearchServerlessPolicy = createOpenSearchServerlessIAMAccessPolicy(<string>props.env?.region, <string>props.env?.account)
 
         const deployId = props.addOnMigrationDeployId ? props.addOnMigrationDeployId : props.defaultDeployId
-        const cdkDomainEndpoint = StringParameter.valueForStringParameter(this, `/migration/${props.stage}/${deployId}/osClusterEndpoint`)
-        const osClusterEndpoint = props.customTargetEndpoint ? props.customTargetEndpoint : `https://${cdkDomainEndpoint}:443`
+        const osClusterEndpoint = StringParameter.valueForStringParameter(this, `/migration/${props.stage}/${deployId}/osClusterEndpoint`)
```
gregschohn, line 96: This can be ignored, but given that some clusters might be OpenSearch and others might be Elasticsearch, even if both of these are being managed by the OpenSearch managed service, would it be better to drop the 'os' from the prefix?  I concede that even if it is, now might not be the right time to make the change.

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -228,49 +229,54 @@ export class StackComposer {
...
+                stage: stage,
+                defaultDeployId: defaultDeployId,
+                addOnMigrationDeployId: addOnMigrationDeployId,
+                ...props,
+            });
+
+            if (networkStack) {
+                openSearchStack.addDependency(networkStack)
+            }
+            this.stacks.push(openSearchStack)
```
gregschohn, line 278: what is the value of openSearchStack when targetClusterEndpoing was specified, will there be a side effect on this.stacks?

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -340,12 +348,11 @@ export class StackComposer {
         }
 
         let trafficReplayerStack
-        if ((trafficReplayerServiceEnabled && networkStack && openSearchStack && migrationStack && mskUtilityStack) || (addOnMigrationDeployId && networkStack)) {
+        if ((trafficReplayerServiceEnabled && networkStack && migrationStack && mskUtilityStack) || (addOnMigrationDeployId && networkStack)) {
             trafficReplayerStack = new TrafficReplayerStack(scope, `traffic-replayer-${deployId}`, {
                 vpc: networkStack.vpc,
                 enableClusterFGACAuth: trafficReplayerEnableClusterFGACAuth,
                 addOnMigrationDeployId: addOnMigrationDeployId,
-                customTargetEndpoint: trafficReplayerTargetEndpoint,
```
gregschohn, line 348: what was this for if we didn't support bringing your own cluster?

deployment/cdk/opensearch-service-migration/options.md
```diff
@@ -63,6 +62,12 @@ These tables list all CDK context configuration values a user can specify for th
...
 
+### Imported Target Cluster Options
+| Name                             | Type   | Example                                                                                           | Description                                                                                                                                                                       |
+|----------------------------------|--------|---------------------------------------------------------------------------------------------------|:----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| targetClusterEndpoint            | string | `"https://vpc-demo-opensearch-cluster-cv6hggdb66ybpk4kxssqt6zdhu.us-west-2.es.amazonaws.com:443"` | The target cluster endpoint which will be used to replay captured traffic against. This endpoint is ideally in the same VPC as the Migration services to allow for easier linking |
```
gregschohn, line 68: "ideally" or "must be"?

deployment/cdk/opensearch-service-migration/README.md
```diff
@@ -126,17 +126,36 @@ After the benchmark has been run, the indices and documents of the source and ta
...
 
 #### OpenSearch Service
-If an open access policy is in place for the Domain, either configured manually or with the `openAccessPolicyEnabled` [option](./options.md) for Domains created with this CDK, this should be sufficient and no changes should be needed on this front. If a custom access policy is in place it must allow the IAM task roles for the applicable Migration services (Traffic Replayer and Migration Console) to access.
+For a Domain, there are typically two items that need to be configured to allow proper functioning of this solution
+1. The Domain should have a security group that allows communication from the applicable Migration services (Traffic Replayer, Migration Console, Fetch Migration). This CDK will automatically create an `osClusterAccessSG` security group, which has already been applied to the Migration services, that a user should then add to their existing Domain to allow this access.
+2. The access policy on the Domain should be an open access policy that allows all access or an access policy that at least allows the IAM task roles for the applicable Migration services (Traffic Replayer, Migration Console, Fetch Migration)
```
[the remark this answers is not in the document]
lewijacn: Yeah maybe something like that would be preferable and hadn't really thought of it. Have created this task around trying it out: https://opensearch.atlassian.net/browse/MIGRATIONS-1431

deployment/cdk/opensearch-service-migration/README.md
lewijacn: Args for the Traffic Replayer, if this is not clear let me know what can help.  

Have rearranged the structure. I've updated to be more precise here but it seems like a combo of wanting to use SigV4 authentication for our authorization header. Let me know if you have better wording.

I've normally seen SigV4: https://docs.aws.amazon.com/AmazonS3/latest/API/sig-v4-authenticating-requests.html

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
lewijacn: For this typescript library yes, i was a little taken as well

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
lewijacn: Have adjusted to throw error for paths that are provided. Its worth pointing out here that by default the URL class here will always have "/" as the path unless an actual one is provided (which we now throw an error for) so we remove it for our structure

lewijacn:
> For clarification concerning "We currently do not support bringing in existing security groups for imported target clusters." Do we allow users to specify an existing VPC & security group when deploying the fetch/console/replayer initially?

We always allow a user to specify an existing vpcId which will be applied to all services. There is also support for providing a security group for a Domain that we specifically create. For the specific services no we don't allow providing custom security groups

gregschohn (approved):

PR #404, opened 2023-11-11
Reliability to get at-least-once request processing behavior in the Replayer with Kafka

gregschohn:
### Description

See [PR-392](https://github.com/opensearch-project/opensearch-migrations/pull/392) and [PR-399](https://github.com/opensearch-project/opensearch-migrations/pull/399) for previous partial PRs.

* Category  Reliability Improvements
* Why these changes are required?  So that when a replayer dies, for whatever reason, a new replayer will pickup where the last replayer had confirmed.  It may replay messages again, but it should not drop any from the Kafka stream due to handoff.
* What is the old behavior before changes and new behavior after changes? There would have been holes, possibly large (depending on buffering) around recoveries.  Now there will be a much smaller overlap where messages might have been sent to the target multiple times (by the old/dying replayer and the new one that is picking up the work)

### Issues Resolved

### Testing
Unit testing, deep log analysis to fix lots of issues, and running OSB with the dockerSolution

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -20,25 +24,108 @@
...
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+/**
+ * Adapt a Kafka stream into a TrafficCaptureSource.
+ *
+ * Notice that there's a critical gap between how Kafka accepts commits and how the
```
lewijacn, line 41: Thanks for this info!

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -20,25 +24,108 @@
...
+ * BlockingTrafficSource throttles calls to Kafka.  The BlockingTrafficSource may
+ * block calls to readNextTrafficStreamChunk() until some time window elapses.  This
+ * could be a very large window in cases where there were long gaps between recorded
+ * requests from the capturing proxy.  For example, if a TrafficStream is read and it
```
lewijacn, line 45: It feels like I'm missing a chunk of text that got deleted at the end of this line

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -20,25 +24,108 @@
...
+ * I believe that this can be mitigated, hopefully fully, by adding a keepAlive/do nothing
+ * call that the BlockingTrafficSource can use.  That can be implemented in a source
+ * like this with Kafka by polling, then resetting the position on the stream if we
+ * aren't supposed to be reading new data.
+ */
 @Slf4j
 public class KafkaProtobufConsumer implements ISimpleTrafficCaptureSource {
 
+    @ToString(callSuper = true)
+    private static class TrafficStreamKeyWithKafkaRecordId extends PojoTrafficStreamKey {
```
lewijacn, line 72: nit: We should probably pull these inner classes out at some point

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
```diff
@@ -20,25 +24,108 @@
...
+ *
+ * See
+ * <a href="https://kafka.apache.org/21/javadoc/org/apache/kafka/clients/consumer/KafkaConsumer.html#failuredetection">...</a>
+ *
+ * "Basically if you don't call poll at least as frequently as the configured max interval,
+ * then the client will proactively leave the group so that another consumer can take
+ * over its partitions. When this happens, you may see an offset commit failure (as
+ * indicated by a CommitFailedException thrown from a call to commitSync())."
+ *
+ * I believe that this can be mitigated, hopefully fully, by adding a keepAlive/do nothing
```
lewijacn, line 63: What happens with these recent changes when we have not polled in a while and the consumer leaves the group? We will hit a commit failure and try to process again?

gregschohn: We'll fail and forget. Though as other messages are processed, we have the potential in the future to not fail. 
However, notice that there are some patterns where we could NEVER successfully commit.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
gregschohn: As I wrote it, I figured out a fix. I plan on patching this shortly and amending the comments. 
See https://opensearch.atlassian.net/browse/MIGRATIONS-1428

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaProtobufConsumer.java
gregschohn: Why?  I see them as implementation details that nobody else outside of the parent class needs to be concerned with.

lewijacn: From a personal perspective it clutters the class, but to each their own

lewijacn (approved):

PR #409, opened 2023-11-12
Remove Comparator and Jupyter since it will be replaced by OTEL and Analytics Engine

sumobrian:
### Description

Removing references to Comparator and Jupyter in the project. Future builds should exclude these containers and stacks. This pull request can serve as a resource for future projects involving Comparator. We will apply our insights to improve comparison processes using OpenTelemetry and an analytics engine. This approach aims to generate visualizations and tools for evaluating updates, migrations, or comparisons more effectively.

### Testing
Verified no unit tests were broken.
Verified Docker Solution end-to-end test using OpenSearch benchmarks ran as expected.

### Check List


TrafficCapture/dockerSolution/build.gradle
```diff
@@ -101,7 +47,7 @@ javaContainerServices.each { projectName, dockerImageName ->
     CommonUtils.createDockerfile(project, projectName, baseImageProjectOverrides, dockerFilesForExternalServices)
 }
 
-(javaContainerServices + trafficComparatorServices).forEach { projectName, dockerImageName ->
+(javaContainerServices).forEach { projectName, dockerImageName ->
```
mikaylathompson, line 50: very minor, but the parens here are now unnecessary

TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml
```diff
@@ -43,9 +43,7 @@ services:
         condition: service_started
       opensearchtarget:
         condition: service_started
-      trafficcomparator:
-        condition: service_healthy
-    command: /bin/sh -c "/runJavaWithClasspath.sh org.opensearch.migrations.replay.TrafficReplayer https://opensearchtarget:9200 --auth-header-value Basic\\ YWRtaW46YWRtaW4= --insecure --kafka-traffic-brokers kafka:9092 --kafka-traffic-topic logging-traffic-topic --kafka-traffic-group-id default-logging-group | nc trafficcomparator 9220"
+    command: /bin/sh -c "/runJavaWithClasspath.sh org.opensearch.migrations.replay.TrafficReplayer https://opensearchtarget:9200 --auth-header-value Basic\\ YWRtaW46YWRtaW4= --insecure --kafka-traffic-brokers kafka:9092 --kafka-traffic-topic logging-traffic-topic --kafka-traffic-group-id default-logging-group"
```
mikaylathompson, line 46: We still need to pipe the output here _somewhere_ or all the tuples will end up in our logs. So something like `| /dev/null` (and probably create a ticket to stop writing the tuples to stdout

gregschohn: This is what we do within the TrafficReplayer today to write tuples both to a logfile and to output, either specified by a command line argument OR stdout...
```
        @Override
        @SneakyThrows
        public void accept(SourceTargetCaptureTuple triple) {
            JSONObject jsonObject = toJSONObject(triple);

            tupleLogger.info(()->jsonObject.toString());
            outputStream.write((jsonObject.toString()+"\n").getBytes(StandardCharsets.UTF_8));
            outputStream.flush();
        }
```

I don't think that we'll EVER want to use stdout to pipe data to another process.  tupleLogger will write the data as json to rotating logfiles.  I'd like to see outputStream get killed here & all the way up through the TrafficReplayer to remove the argument.  Notice that we write the string twice.  It may not make sense to buffer it, given that some of these are 10MB, we may not get great caching effects.  Removing a write that we don't care about will be a major win though.

As far as value in tuples, I've generally disabled these from my console logs as they easily dominate the output and drown out everything else.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
```diff
@@ -58,7 +58,7 @@ public class HttpJsonTransformingConsumer<R> implements IPacketFinalizingConsume
     private final List<List<Integer>> chunkSizes;
     // This is here for recovery, in case anything goes wrong with a transformation & we want to
     // just dump it directly.  Notice that we're already storing all of the bytes until the response
-    // comes back so that we can format the output that goes to the comparator.  These should be
+    // comes back so that we can format the output.  These should be
```
gregschohn, line 61: reformat to not make it look like you deleted something :)

gregschohn (commented):
In addition to the change on stdout, can you please cut a branch with the trafficComparator & Jupyter.  We can merge mainline to it if/when we like.  That let's us remember where it is, keep it relatively warm, yet not have it cluttering the mainline

gregschohn (approved):

PR #421, opened 2023-11-14
Transformer configuration UX changes

gregschohn:
### Description
Transformation improvements.
I've added all of the transformer jars to the runtime closure of the replayer since it's going to be a challenge for users using a cloud-deployment shrink-wrap solution to add addition jarfiles.
Change the syntax/rules for Transformer configs.
* If no argument was specified for the transformer config, no transformer used, regardless of what's in the classpath.
* To use multiple transformers, you’ll need to specify the transformer-config as a json array (of json transformer configs)
* To use one transformer with default settings, you can JUST put the name of the transformer in the arg. e.g. --transformer-config JsonTransformerForOpenSearch23PlusTargetTransformerProvider.

I've also updated the key name to only use the short, simple name of a class rather than the fully qualified one with the namespace.
Add a NoopTransformer, which makes unit and sanity testing a lot simpler.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1429

### Testing
gradle build

### Check List


TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/TransformationLoaderTest.java
```diff
@@ -28,4 +28,28 @@ public void testTransformationLoader() throws Exception {
         Assertions.assertNotEquals(origDocStr, docWithNewHostnameStr);
     }
 
+    @Test
+    public void testThatSimpleNoopTransformerLoads() throws Exception {
+        var noopTransformer = new TransformationLoader()
+                .getTransformerFactoryLoader("localhost", "NoopTransformerProvider");
```
lewijacn, line 34: Do we currently have any other existing transformation providers besides "NoopTransformerProvider" ?

lewijacn (approved):

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/TransformationLoaderTest.java
gregschohn: In the base package, no.  In TrafficReplayer, there's a HostTransformer & soon a UserAgentTransformer.  Transformers exist for Jolt and JMESPath in separate packages.

Issue #367, opened 2023-10-24
PUT index requests with mappings objects always blindly excise


sanathbhat:
### What is the bug?
PUT index requests with a `mappings` object are always indiscriminately subject to excision of the top level object within the `mappings` object.

### How can one reproduce the bug?
With traffic capture and replayer modules running, make a put index request to source cluster via capture proxy server with the following sample body:
```
{
  "settings": {
    "number_of_shards": 1,
    "number_of_replicas": 1
  },
  "mappings": {
    "properties": {
      "field1": {
        "type": "text"
      },
      "field2": {
        "type": "keyword"
      }
    }
  }
}
```
This would create the index on the source cluster as expected but the target cluster returns the following error:

> {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [field1 : {type=text}] [field2 : {type=keyword}]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: Root mapping definition has unsupported parameters:  [field1 : {type=text}] [field2 : {type=keyword}]","caused_by":{"type":"mapper_parsing_exception","reason":"Root mapping definition has unsupported parameters:  [field1 : {type=text}] [field2 : {type=keyword}]"}},"status":400}

On checking the logs, it can be seen that the target cluster received the request body shown in `Target request:` section below, instead of the original, where the `properties` object has been excised and its contents directly pushed into its parent `mappings`. The likely cause of this can be traced to these lines of code: https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/transform/JsonTypeMappingTransformer.java#L77-L87

### What is the expected behavior?
The above behavior was probably targeted at some open source version compatibility issue as a proof-of-concept solution but causes the replayer module to indiscriminately modify all put index request bodies making some of them erroneous and thereby failing to create indices on the target cluster. The expected behavior is that the index should be correctly created in both the source and target clusters wherein the replayer module carefully performs any request body transformations in an intelligent way only if required.

### What is your host/environment?
Source/target cluster platforms: OCI
Source cluster version: OS 2.3
Target cluster version: OS 2.8
Oracle Linux host running capture proxy and replayer modules

### Do you have any screenshots?
[...]

PR #428, opened 2023-11-14
Add option to read Transformations from a file

okhasawn:
### Description
This PR adds the option (`--transformer-config-file`) for the Traffic Replayer to read the transformations from a file.

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -346,8 +354,23 @@ public static void main(String[] args)
...
+            if (params.transformerConfigFile != null && !params.transformerConfigFile.isBlank()) {
+                try {
+                    configContent = Files.readString(Paths.get(params.transformerConfigFile), StandardCharsets.UTF_8);
+                } catch (IOException e) {
+                    System.err.println("Error reading transformer configuration file: " + e.getMessage());
+                    System.exit(5);
+                }
+            } else if (params.transformerConfig != null && !params.transformerConfig.isBlank()) {
+                configContent = params.transformerConfig;
+            }
```
gregschohn, line 371: Please put these lines into a helper function to return the config (or null).  The helper function could throw or call exit() directly.  We'll do a rewrite of all of this parsing logic that's getting spread out in the near future.

gregschohn (commented):
Once you refactor that, I'm fine with this change.  Thanks!

gregschohn (commented):
Please also log the transformerConfig to info so that users will know what transformations were used for a given run.

okhasawn:
Thanks for the comments @gregschohn!
All addressed in new commits.

okhasawn:
I tested attempting #367 with the following `TransformerConfig.json` file:
```
[
  {"JsonJMESPathTransformerProvider": { "script": "{ \"settings\": settings, \"mappings\": payload.mappings.*.properties | [merge(@)] }"}}]
```

Results: can confirm that the above transformation does fix the bug. I was able to reproduce the bug without the transformation.

gregschohn (approved):
I'd like to revisit this whole file and refactor it.  I'm not sure exactly what this would look like after refactoring (specifically around error handling & System.exit()).  This is a much needed feature in the interim though.

PR #430, opened 2023-11-15
[MIGRATIONS-1326] Add status code comparison metric

mikaylathompson:
### Description
This uses the metrics logging code (from #376 ) to add a new metric that contains the source and target response status codes and a "are they the same" value to make using this simpler.

It also fixes an issue discovered by sonarqube regarding outputting the entire stack trace in our metrics.

### Issues Resolved
[MIGRATIONS-1326](https://opensearch.atlassian.net/browse/MIGRATIONS-1326)

### Testing
Manually tested

### Check List


TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/coreutils/MetricsLogger.java
```diff
@@ -76,7 +76,6 @@ public MetricsLogBuilder atError(MetricsEvent event, Throwable cause) {
         }
         return new MetricsLogBuilder(logger).atError(event)
                 .setAttribute(MetricsAttributeKey.EXCEPTION_MESSAGE, cause.getMessage())
-                .setAttribute(MetricsAttributeKey.EXCEPTION_STACKTRACE, cause.getStackTrace().toString())
```
gregschohn, line 79: For future reference, this change is NOT required to add a new metric.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/SourceTargetCaptureTuple.java
```diff
@@ -119,6 +125,26 @@ private JSONObject toJSONObject(SourceTargetCaptureTuple tuple) {
                     .ifPresent(d-> meta.put("targetResponse", jsonFromHttpData(d, tuple.targetResponseDuration)));
             meta.put("connectionId", tuple.uniqueRequestKey);
             Optional.ofNullable(tuple.errorCause).ifPresent(e->meta.put("error", e));
+
+            if (meta.has("sourceResponse") && meta.has("targetResponse")) {
+                String sourceResponseStatus = (String) meta.getJSONObject("sourceResponse").get("Status-Code");
+                String targetResponseStatus = (String) meta.getJSONObject("targetResponse").get("Status-Code");
```
gregschohn, line 131: The sourceResponse/targetResponse are filled in with code that runs through this function...
```
        private JSONObject jsonFromHttpData(@NonNull List<byte[]> data) {
            try {
                return jsonFromHttpDataUnsafe(data);
            } catch (Exception e) {
                log.warn("Putting what may be a bogus value in the output because transforming it " +
                        "into json threw an exception");
                return new JSONObject(Map.of("Exception", e.toString()));
            }
        }
```

That means that we might have the "Status-Code" values of null.  That will be an issue down at line 137 where we'll throw an NPE.

gregschohn (changes requested):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/SourceTargetCaptureTuple.java
```diff
@@ -119,6 +125,26 @@ private JSONObject toJSONObject(SourceTargetCaptureTuple tuple) {
                     .ifPresent(d-> meta.put("targetResponse", jsonFromHttpData(d, tuple.targetResponseDuration)));
             meta.put("connectionId", tuple.uniqueRequestKey);
             Optional.ofNullable(tuple.errorCause).ifPresent(e->meta.put("error", e));
+
+            if (meta.has("sourceResponse") && meta.has("targetResponse")) {
```
gregschohn, line 129: If there was no response for the target (or even the source), we probably want to still emit the other status code, and especially the part that it doesn't match, right?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/SourceTargetCaptureTuple.java
```diff
@@ -27,6 +31,8 @@
 @Slf4j
 public class SourceTargetCaptureTuple implements AutoCloseable {
     public static final String OUTPUT_TUPLE_JSON_LOGGER = "OutputTupleJsonLogger";
+    private static final MetricsLogger metricsLogger = new MetricsLogger("SourceTargetCaptureTuple");
```
gregschohn, line 34: It would be a lot cleaner if this were decomposed into separate responsibilities.  Here's what we do with the raw requests/responses.
Some of those handlers may convert these to higher order messages.
When we have higher order messages, we should be able to attach consumers/listeners on top of those.
Right now, it's exceptionally easy to see how metrics would be completely suppressed if we decided NOT to log tuple outputs.

gregschohn (changes requested):
Still requesting changes.

gregschohn:
This has been superceded by https://github.com/opensearch-project/opensearch-migrations/pull/433

PR #433, opened 2023-11-15
Status code metrics

gregschohn:
### Description
Carry over from [PR-430](https://github.com/opensearch-project/opensearch-migrations/pull/430).  Along the way, so that I could test the code, and prepare for adding [one line of output per transaction](https://opensearch.atlassian.net/browse/MIGRATIONS-1432), I've refactored the final response ("tuple") handling code.  

org.json is now no longer parsing our HTTP responses, so that dependency for the replayer goes away.
We also print out a line (as configured by default) to stdout with a quick status of the status codes and latencies between the responses for each request.
 
* Category Enhancement, Metrics, Refactoring
* Why these changes are required? To make it easier to test and to add metrics.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1406
https://opensearch.atlassian.net/browse/MIGRATIONS-1326
https://opensearch.atlassian.net/browse/MIGRATIONS-1432

### Testing
gradle slowTests.

### Check List
- [x] New functionality includes testing
  - [x] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #437, opened 2023-11-16
Add transformer to remove types if they existed under mappings

### LINKED ISSUE #367: PUT index requests with mappings objects always blindly excise

okhasawn:
### Description
This PR adds a transformer that would remove "types" from a request if it existed under "mappings", otherwise it would keep the index's "properties", which used to be mistakenly removed.

### Issues Resolved
#367 

### Check List


TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/build.gradle
```diff
@@ -4,6 +4,9 @@ plugins {
 
 dependencies {
     implementation project(':replayerPlugins:jsonMessageTransformers:jsonMessageTransformerInterface')
+    implementation project(':replayerPlugins:jsonMessageTransformers:jsonJMESPathMessageTransformer')
+    implementation project(':replayerPlugins:jsonMessageTransformers:jsonJMESPathMessageTransformerProvider')
+    implementation group: 'io.burt', name: 'jmespath-core', version: '0.6.0'
```
gregschohn, line 9: You shouldn't need these two.  Providers are basically just external entrypoints.  You're providing your own.  The jmespath library (or whatever implementation that the jsonJMESPathMessageTransformer package uses) will be picked automatically.
(see the next comment for how to shake the compile error that you would get)

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/main/java/org/opensearch/migrations/transform/Es2OsPost23.java
```diff
@@ -0,0 +1,9 @@
+package org.opensearch.migrations.transform;
+
+import io.burt.jmespath.jcf.JcfRuntime;
+public class Es2OsPost23 extends JsonJMESPathTransformerProvider {
+    public JsonJMESPathTransformer createTransformer(Object jsonConfig) {
+        String script = "{ \"settings\": settings, \"mappings\": payload.mappings.*.properties | [merge(@)] }";
+        return new JsonJMESPathTransformer(new JcfRuntime(), script);
```
gregschohn, line 7: OK - now I see why you had to add the dependency... I setup the c'tor to take the dependency because we could be using JSON nodes from jackson or something else - and those themself will require additional dependencies.

Anyway, please add an override constructor that defaults to the JcfRuntime & use that instead...
```
    public JsonJMESPathTransformer(String script) {
        this(new JcfRuntime(), script);
    }
```

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/main/resources/META-INF/services/org.opensearch.migrations.transform.IJsonTransformerProvider
```diff
@@ -0,0 +1 @@
+org.opensearch.migrations.transform.Es2OsPost23
```
gregschohn, line 1: I don't like the casing or the two different numbers.  Since the 23 is used to represent a sequence number, we should probably not ALSO load a number as an abbreviation.  That leaves something like EsToOsPost23.  Then we have the casing.  I didn't want to use underscores before, but they seem like the least bad option.... What do you think of ES_OS_Post23?  We'll likely have lots of occurrences of ES/OS (& other things too), so coming up w/ abbreviations and establishing underscores could allow this to scale.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/test/java/org/opensearch/migrations/transform/Es2OsPost23Test.java
```diff
@@ -0,0 +1,5 @@
+package org.opensearch.migrations.transform;
+
+public class Es2OsPost23Test {
+
+}
```
gregschohn, line 5: It looks like the excision test is still in place & I presume that it is passing.  Could you either delete this file or move the excision test into here?

Also please write a test that shows that your new transform fixes the bug that was reported.  You can take a look at the `normalizeJson` functions in [this PR](https://github.com/opensearch-project/opensearch-migrations/pull/433/files) and do that here with hardcoded input/output.  (IntelliJ will escape as necessary if you paste INTO a string)

okhasawn:
> Should there be another test case? Looking at #367, the reporter is migrating TO OS 2.8. Do we need to alter the original transformation or did we just need to disable it since we were running it on requests that didn't have type mappings? Should we be checking that we're removing the type mappings value and not just _any_ value? Before approving, I just want to make sure that I understand how we're fixing the bug. Then you can add comments to it and mark it as fixed.
> 
> I'm still not sure about the name of the transform, but we're limited by using the classname since that has strong restrictions on valid characters. We'll also need to include "Provider" on them, OR choose to have less descriptively named providers.
> 
> I think there's another task to clean that up.

I'll add another test case, I also think that the name of the transform should be changed, since what it essentially does is it removes type mappings, if and only if they existed, but keeps properties, therefore it doesn't remove _any_ value, and would handle both cases.

I would still say that disabling it on requests that don't have type mappings can still be considered a solution to the bug.
If that can be detected and run transformations based on certain formats we want, then that can reduce the need for complicated transformations.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/test/java/org/opensearch/migrations/transform/Es2OsPost23Test.java
```diff
@@ -65,4 +65,4 @@ private static void transformAndVerifyResult(Map<String,Object> json, String exp
     static IJsonTransformer getJsonTransformer() {
         return new JsonTypeMappingTransformer();
```
gregschohn, line 66: You should be testing the new transformer & you should be deleting this one, right?

okhasawn: Added a new one that uses JMES transformers.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/main/java/org/opensearch/migrations/transform/Es2OsPost23.java
```diff
@@ -1,9 +1,30 @@
 package org.opensearch.migrations.transform;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+
 public class Es2OsPost23 implements IJsonTransformerProvider {
```
gregschohn, line 8: Please affix this with TransformerProvider and override getName() to get the value that you want.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/main/java/org/opensearch/migrations/transform/Es2OsPost23.java
```diff
@@ -1,9 +1,30 @@
...
 
+    @Override
     public JsonJMESPathTransformer createTransformer(Object jsonConfig) {
-        String script = "{ \"settings\": settings, \"mappings\": payload.mappings.*.properties | [merge(@)] }";
+        ObjectMapper objectMapper = new ObjectMapper();
+        Map<String, Object> jsonMap = objectMapper.convertValue(jsonConfig, Map.class);
```
gregschohn, line 13: Take a look at the provider documentation for `createTransformer`...
```
     * @param jsonConfig is a List, Map, String, or null that should be used to configure the
     *                   IJsonTransformer that is being created
```

What are you trying to do with convert?  If you have a map being passed in, this will work.  Anything else will cause it to throw.  You'd be better off to simply cast this to Map<String,Obiect>.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/main/java/org/opensearch/migrations/transform/Es2OsPost23.java
```diff
@@ -1,9 +1,30 @@
...
+        String script;
+
+        if (jsonMap.containsKey("mappings") && jsonMap.get("mappings") instanceof Map) {
```
gregschohn, line 16: the config value coming in is meant to configure a transformer (like how you're passing the script to the JMESPath Transformer's c'tor).  TransformationLoader only creates ONE transformer from a provider AND it passes the configuration, NOT a document, to createTransformer().

You'll want to make sure that you add another test that uses TransformationLoader to configure this transformer and then give it a couple documents.  Those tests would have failed & you could have spotted the inconsistencies.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/main/java/org/opensearch/migrations/transform/Es2OsPost23.java
```diff
@@ -1,9 +1,30 @@
...
+            Map<String, Object> mappings = (Map<String, Object>) jsonMap.get("mappings");
+            if (mappings.containsKey("properties")) {
+                // No type - no transformation needed for "mappings" since "properties" is directly under it.
+                script = "{\"settings\": settings, \"mappings\": mappings}";
```
gregschohn, line 20: Once you cleanup the issues above, you can still have a Transformer that contains a couple JMESPath transformers within it as fields and switch which one you'd like to use within transformJson.

TrafficCapture/replayerPlugins/jsonMessageTransformers/Es2OsPost23/src/test/java/org/opensearch/migrations/transform/Es2OsPost23Test.java
```diff
@@ -53,6 +65,15 @@ private static Map<String,Object> parseJsonFromResourceName(String resourceName)
         }
     }
 
+    private static void transformJMESAndVerifyResult(Map<String,Object> json, String expectedValueSource) throws Exception{
+        var jmesTransformer = getJMESTransformer(json);
```
gregschohn, line 69: These tests much be updated - they should send a config value (which should be null, since the user shouldn't have anything to configure) - NOT the document.

gregschohn (changes requested):
We need to figure out a naming pattern,  after some deliberation, I like `SOURCE-TARGET` where both source and target are structured as `STORE_TYPE[/[..]VERSION[..]]` where .. on the left means up to (& including).  The trailing .. means after.  Leaving the version off altogether implies all versions.

Examples would be
```
ES/..6.8-OS/2.3..
ES/..6.8-OS/..2.3
ES/..6.8-OS/2.4..
Solr-OS
```

PR #456, opened 2023-11-18
Blocking traffic source with keep alives

gregschohn:
### Description

Commits would not always be applied (which was often) because the consumer would be kicked out of the group.  Now there's logic in the BlockingTrafficSource and the Kafka source implementation to stay within the group.  There's also now a lot more attention to rebalances to maintain state, specifically commit state, and to purge it appropriately when the status/consumer-broker relationship changes.

* Category Bug fix
* Why these changes are required? It was easy for the replayer to get hung up and to not preserve at least once AND IN-ORDER guarantees.
* What is the old behavior before changes and new behavior after changes?  Flakiness and some subsequent messages not being replayed when earlier messages were being replayed.

### Issues Resolved
No ticket, but this is follow up work/bugs from the [Kafka Reliability task](https://opensearch.atlassian.net/browse/MIGRATIONS-1379).

Is this a backport? If so, please add backport PR # and/or commits #. Should be a backport.

### Testing
build; unit tests; docker OSB 4 test dataset doc count confirmation

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/traffic/source/BlockingTrafficSource.java
```diff
@@ -126,6 +113,42 @@ public Duration getBufferTimeWindow() {
...
+                        .addArgument(stopReadingAtRef.get())
+                        .log();
+                var nextTouchOp = underlyingSource.getNextRequiredTouch();
+                if (nextTouchOp.isEmpty()) {
+                    readGate.acquire();
+                } else {
+                    var nextInstant = nextTouchOp.get();
+                    final var nowTime = Instant.now();
+                    var waitIntervalMs = Duration.between(nowTime, nextInstant).toMillis();
+                    log.error("Next touch at " + nextInstant + " ... in " + waitIntervalMs + "ms (now="+nowTime+")");
```
lewijacn, line 134: nit: Wrong log level

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/TrackingKafkaConsumer.java
```diff
@@ -0,0 +1,240 @@
...
+            }
+        } catch (IllegalStateException e) {
+            throw e;
+        } catch (RuntimeException e) {
+            log.atWarn().setCause(e).setMessage("Unable to poll the topic: " + topic + " with our Kafka consumer. " +
+                    "Swallowing and awaiting next metadata refresh to try again.").log();
+        } finally {
+            resume();
+        }
+        safeCommit();
```
lewijacn, line 122: I wasn't expecting anything to be committed with touch(), what is happening here?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -0,0 +1,221 @@
...
+ * simply run a background job to keep the client warm.  We need the caller to touch
+ * this object periodically to keep the connection alive.
+ */
+@Slf4j
+public class KafkaTrafficCaptureSource implements ISimpleTrafficCaptureSource {
+
+    public static final String MAX_POLL_INTERVAL_KEY = "max.poll.interval.ms";
+    // see https://stackoverflow.com/questions/39730126/difference-between-session-timeout-ms-and-max-poll-interval-ms-for-kafka-0-10
+    public static final String DEFAULT_POLL_INTERVAL_MS = "60000";
+    private static final MetricsLogger metricsLogger = new MetricsLogger("KafkaProtobufConsumer");
```
lewijacn, line 67: nit: We should probably update the naming to match at some point

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    @Override
+    public void touch() {
+        trackingKafkaConsumer.touch();
+    }
+
+    /**
+     * If messages are outstanding, we need to keep the connection alive, otherwise, there's no
+     * reason to.  It's OK to fall out of the group and rejoin once ready.
```
lewijacn, line 154: Is it? This probably needs to be discussed another time in the context of scalability, but for our single Replayer setup currently is this better than just staying in the group?

lewijacn (approved):
Approved with some nitpicks and understanding questions

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/traffic/source/ITrafficCaptureSource.java
```diff
@@ -14,4 +17,14 @@ public interface ITrafficCaptureSource extends Closeable {
     default void commitTrafficStream(ITrafficStreamKey trafficStreamKey) throws IOException {}
 
     default void close() throws IOException {}
+
+    /**
+     * Keep-alive call to be used by the BlockingTrafficSource to keep this connection alive if
+     * this is required.
+     * @return The time that the next call to touch() must be completed for this source to stay
```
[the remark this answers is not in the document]
gregschohn: Moved that to the next method below

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/TrackingKafkaConsumer.java
gregschohn: We would like to see commits by synced to Kafka as soon as possible.  However, Kafka doesn't let us do that asynchronously, so we have to do it on the polling thread & we need to make sure that polls have recently been called.  Doing this in touch lets us do it more frequently (when commits are coming in) & let's us be sure that we should still be in the group because there was a recent poll().

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
gregschohn: Is it OK to keep the connection alive when they're outstanding OR to fall out of the group when they are not?
I think both of those points are the best policies (given my current knowledge).  
I'm assuming that "holding" messages for relatively long periods of time is a design trait for the system and that we should allow that to proceed.  That also implies that we'll need other metrics to find out when we're holding messages because of overwork - in which cases we should offload some partitions to other replayers.
The 'leaving the group' part of the statement makes sense to me if the caller REALLY knows that they aren't going to do anything in the interim.  If they change their mind, it's a pretty quick reconnect anyway (albeit with potentially different partition assignments).

PR #460, opened 2023-11-30
Otel metrics and traces

gregschohn:
### Description

This change will break the flow of messages emitted by the existing MetricsLogger class into OpenSearch.  The biggest reason for that is that the change is that I'm now using the otel/opentelemetry-collector:latest image rather than a bespoke one.  

Aside from the breakage, there's a lot being added in the form of Metrics and Traces.  We're using OpenTelemetry to send both, currently to a collector (just as we were doing with logs).  The collector sends data to two new trace collection containers (Jaeger & Zipkin) and has a Prometheus container pulling metrics from it.

The Java application code itself eschews some of the typical OpenTelemetry techniques for instrumentation.  Instead of using ThreadLocals to pass maybe-present values around within contexts, which each instrumentation point needs to determine how to use them, custom context classes for Connections, Requests, KafkaRecords, etc are constructed and explicitly passed into functions and into callbacks.  Those classes implement IWithAttributes and the fillAttributes() function to select which fields should be included within the instruments that are being emitted.  

The contexts themselves are tightly related to Spans.  Usually a new context will have a new span, a new span will always require a new context.  The context classes themselves have the ability to chain back to a parent scope.  When the context is converted into an Attributes object for the instruments, the attributes from parent contexts will also be included (with key-values from subclasses overwriting their parent's values in cases of conflict).

There are some judicious uses of generic wildcard constraints to make it quicker and more foolproof to create spans so that they're appropriately associated as children with the containing context's span.  There's also support to make it easier to store start timestamps to emit duration metrics.

There's a LOT left to do here, but there's a lot that's done and I'd like to get feedback on the patterns that are emerging.  Some of the top remaining items.

* More metrics, more traces
* Getting separate namespaces for capture and replayer working for metrics
* Figuring out our cloud story.  Should we deploy some more ECS containers for the AWS CDK or should we use AWS cloud native stuff, AWS hosted stuff?
* Tests, Tests, Tests - Otel has some easy to use facilities to simplify checking within tests.
* Hardening the interfaces more.  There are lots of inline strings.  These should be managed from more centralized places and we should have tests to do double-entry book-keeping so that they don't change without warnings.
* Moving the existing MetricsLogger code out.

* Category - Enhancement
* Why these changes are required? Visibility into what our services are doing.

### Issues Resolved
Part of [Improve Metrics Explanations](https://opensearch.atlassian.net/browse/MIGRATIONS-1438)

### Testing
Lots of manual testing for now

### Check List


TrafficCapture/trafficReplayer/src/test/resources/log4j2.properties
```diff
@@ -18,3 +18,8 @@ logger.OutputTupleJsonLogger.level = OFF
 logger.KPC.name = org.opensearch.migrations.replay.kafka.KafkaProtobufConsumer
 logger.KPC.level = DEBUG
 logger.KPC.appenderRef.stdout.ref = Console
+
+logger.RSO.name = org.opensearch.migrations.replay.RequestSenderOrchestrator
+logger.RSO.level = TRACE
```
lewijacn, line 23: Should this be at trace as a default?

lewijacn (commented):
Added some comments, but had a general question:

Wasn't sure what the status of MetricsLogger is? Are we in a transition state with it currently, but plan to remove it in future?

PR #461, opened 2023-11-30
Add endpoints and authentication details (including SigV4) as parameters to E2E test script

okhasawn:
### Description
This PR adds the ability to provide the endpoints and authentication details as parameters to the test script.
The end goal is to be able to provide either local or cloud based endpoints as parameters and the authentication details as parameters and have the E2E test script run successfully.
This PR also enables to the E2E test script to run on endpoints that use SigV4.

### Issues Resolved
[Migrations-1442](https://opensearch.atlassian.net/browse/MIGRATIONS-1442)

### Check List


test/tests.py
```diff
@@ -187,36 +224,38 @@ def test_0004_negativeAuth_invalidCreds(self):
...
 
         # This test will send an incorrect URI
         incorrectUri = "/_cluster/incorrectUri"
-        response = requests.get(f'{self.proxy_endpoint}{incorrectUri}', auth=self.auth, verify=False)
+        response = requests.get(f'{self.proxy_endpoint}{incorrectUri}', auth=self.source_auth,
+                                verify=self.source_verify_ssl)
         self.assertEqual(response.status_code, HTTPStatus.METHOD_NOT_ALLOWED)
 
-    def test_0007_OSB(self):
+    def test_0006_OSB(self):
```
lewijacn, line 258: This seems like it will need to be tweaked to work for AWS as well.  For AWS we will already be on the migration console box so we would just need to make the `runTestBenchmarks.sh` call from the right directory

lewijacn (commented):
As far as this structure for local runs I only have one comment that I added. If we want this PR to be just for local I can go ahead and approve.

As far as running this on AWS it seems we still need a couple things, but correct me if I'm wrong

1. Support for AWS SigV4 to verify clusters which have SigV4 enabled
2. A mechanism to only run certains tests depending on some input parameter e.g. I only want to run sigv4 tests on AWS.
3. Make sure existing test cases work for both local and AWS, I believe only `test_0006_OSB` needs to be tweaked but didn't check closely

.github/workflows/e2eTest.yml
```diff
@@ -44,4 +44,4 @@ jobs:
         run: |
           cd test
           chmod +x ./tests.py
-          pytest -n 3 tests.py
+          pytest tests.py --proxy_endpoint="https://localhost:9200" --source_endpoint="https://localhost:19200" --target_endpoint="https://localhost:29200" --source_auth_type="basic" --source_username="admin" --source_password="admin" --target_auth_type="basic" --target_username="admin" --target_password="admin" --source_verify_ssl=False --target_verify_ssl=False
```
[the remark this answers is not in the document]
okhasawn: You're right, I don't need to specify those in the workflow anymore. I had updated the defaults value after.
I'll update this to remove the values and add the verbose flag. Thank you!

test/tests.py
okhasawn: You're right, this specific test won't work on AWS as of this moment. I imagined that this script will end up existing on the migration console and running from there. I can update this test in a future PR to support that.
As for the other tests; per my testing on our deployed AWS solution (w/o SigV4 whatsoever), they will pass.

okhasawn:
Thank you for reviewing!
As for your comments:

- SigV4 support will be added in a future PR.
- Certain tests can be ran using `pytest`'s method of running tests which is for example: `pytest tests.py::E2ETests::test_0001_index tests.py::E2ETests::test_0002_document` will get this to only run those two tests specified.
- You're right, existing tests will work except for the one you mentioned. I'll add a flag to check if it's running on AWS in this PR to exclude this test for now, and update this test to work on AWS in a future PR as well.

okhasawn:
Added option to run on SigV4 enabled endpoints.

lewijacn (commented):
Just to confirm we still have the one test that won't work for both AWS and local, seems like we need some pattern in place so that only proper tests will run when this test script is executed

test/tests.py
```diff
@@ -264,16 +265,22 @@ def test_0005_invalidIncorrectUri(self):
...
 
     def test_0006_OSB(self):
-        cmd = ['docker', 'ps', '--format="{{.ID}}"', '--filter', 'name=migration']
-        container_id = subprocess.run(cmd, stdout=subprocess.PIPE, text=True).stdout.strip().replace('"', '')
-
-        if container_id:
-            cmd_exec = f"docker exec {container_id} ./runTestBenchmarks.sh"
-            logger.warning(f"Running command: {cmd_exec}")
+        if self.deployment_type == "cloud":
+            cmd_exec = f"./runTestBenchmarks --unique-id {self.index}"
```
lewijacn, line 269: The path here for runTestBenchmarks will be in a different directory. It's probably easiest to use `/root/runTestBenchmarks.sh`

test/tests.py
```diff
@@ -103,7 +104,7 @@ def setup_authentication(self, auth_type, username, password):
         return None
 
     def set_common_values(self):
-        self.index = f"my_index_{uuid.uuid4()}"
+        self.index = self.unique_id
```
lewijacn, line 107: It seems like it would be nice to keep the previous format like: `f"test_index_{self.unique_id}"` to be able to clearly tell what our test index is

lewijacn: I imagine our unique id for the integ tests may just be the name of the test and the pipeline run number like `full_run_21`

test/tests.py
lewijacn: Could we also use the self.unique_id instead of self.index here? It seems like it may be clearer

test/operations.py
```diff
@@ -1,52 +1,49 @@
 import requests
 import json
-from typing import Optional, Tuple
 
 
-def create_index(endpoint: str, index_name: str, auth: Optional[Tuple[str, str]] = None, verify_ssl: bool = False):
+def create_index(endpoint: str, index_name: str, auth, verify_ssl: bool = False, protocol: str = 'https'):
```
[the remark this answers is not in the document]
okhasawn: I'll remove this. Mistakenly added from something else I was testing.

test/operations.py
```diff
@@ -1,52 +1,49 @@
...
     response = requests.put(f'{endpoint}/{index_name}', auth=auth, verify=verify_ssl)
```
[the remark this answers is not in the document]
okhasawn: That's correct!

test/tests.py
okhasawn: Good idea, I'll update in the next commit.

lewijacn (approved):

PR #471, opened 2023-12-08
Fix more Kafka committing errors

gregschohn:
Bugfixes include:
* Run `touch()` and `readNextTrafficStreamChunk()` activities for the Kafka client (except for `close` as per [this issue](https://opensearch.atlassian.net/browse/MIGRATIONS-1445)) on a dedicated thread, rather than using CompletableFuture.supplyAsync(), which could change the actual thread being used.
* Putting synchronized blocks around the OffsetLifecycleTracker's add, remove, toString() calls since those will be called from the Kafka worker threads and the BlockingTrafficSource's (TrafficReplayer's main) thread.
* TrackingKafkaConsumer.resume now calls resume instead of pause() (major issue that resulted in the consumer staying paused until restart)
* touch() calculations were off, resulting in us falling out of the consumer group when records were in flight.  Now, the test to determine if touch() should be called is based upon if a record has been returned by getNextBatchOfRecords has yet to be fully committed to Kafka.
* Consumer generations are now bound on a partition by partition basis so that we can more selectively punt and preserve commits for no-longer-tracked vs still-tracked partitions.
* ParsedHttpMessagesAsDicts threw an exception when trying to call array() on a ByteBuf that was a CompositeByteBuf.  Now that code fills the byte array with a slower getBytes() call.
* Commits were being called multiple times if there were multiple requests within a TrafficStream.  This PR includes a test for that, but a merged PR includes the fix (to use trafficStreamKeysBeingHeld ONLY for holds and not for stream identification too)

Refactorings include
* Splitting the gzip file input traffic stream away from the upgrades that were done for legacy streams.  That makes the new super class very useful to use to reconstruct error situations that may have been found in the wild.

### Description

* Category Bug fixes
* Why these changes are required? Every time touch() was called on the kafka traffic source, the client would stay indefinitely paused.  Other issues were present that would have caused issues with guaranteeing continuous 'at-least-once' delivery.
* What is the old behavior before changes and new behavior after changes?  Replayer would hang until it was manually restarted.

### Issues Resolved
Continuation of https://opensearch.atlassian.net/browse/MIGRATIONS-1379

### Testing
Only gradle.

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -166,27 +172,30 @@ public CompletableFuture<List<ITrafficStreamWithKey>> readNextTrafficStreamChunk
...
-            var records = trackingKafkaConsumer.getNextBatchOfRecords();
-            Stream<ITrafficStreamWithKey> trafficStream = StreamSupport.stream(records.spliterator(), false)
-                    .map(kafkaRecord -> {
+            return trackingKafkaConsumer.getNextBatchOfRecords((offsetData,kafkaRecord) -> {
                         try {
                             TrafficStream ts = TrafficStream.parseFrom(kafkaRecord.value());
+                            metricsLogger.atSuccess(MetricsEvent.PARSED_TRAFFIC_STREAM_FROM_KAFKA)
+                                    .setAttribute(MetricsAttributeKey.CONNECTION_ID, ts.getConnectionId())
+                                    .setAttribute(MetricsAttributeKey.TOPIC_NAME, trackingKafkaConsumer.topic)
+                                    .setAttribute(MetricsAttributeKey.SIZE_IN_BYTES, ts.getSerializedSize()).emit();
```
lewijacn, line 187: Don't we already do this is in the next four lines?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -68,6 +71,7 @@ public class KafkaTrafficCaptureSource implements ISimpleTrafficCaptureSource {
 
 
     final TrackingKafkaConsumer trackingKafkaConsumer;
+    private final ExecutorService kafkaExecutor;
```
lewijacn, line 74: Remind me again what prompted the need for a dedicated kafkaExecutor?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
gregschohn: Shoot - yes we do.  Bad merge.  I've pulled one of these away

PR #472, opened 2023-12-08
Add ability to use ECS Kafka and OpenSearch containers for testing use cases in CDK

lewijacn:
### Description
This change works in some existing CDK to allow Kafka as an ECS service as well as a new service stack to spin up an OpenSearch docker image as an ECS service. Both of these cases use a single instance approach currently and are labelled as testing accordingly. These services allow for faster turn around testing, as well as lay some ground work for allowing different streaming provides in our solution and help avoid getting tied to a particular approach.

~~**Pending**: Final wiring for Kafka and testing~~

### Issues Resolved
None currently

### Testing
Manual and deployment testing

### Check List


deployment/cdk/opensearch-service-migration/lib/service-stacks/opensearch-container-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+        const ulimits: Ulimit[] = [
+            {
+                name: UlimitName.MEMLOCK,
+                softLimit: -1,
+                hardLimit: -1
+            },
+            {
+                name: UlimitName.NOFILE,
+                softLimit: 65536,
+                hardLimit: 65536
```
gregschohn, line 45: will these limits be supported on the container that OS is running on?  What's the behavior if the process breaches these bounds what happens?

deployment/cdk/opensearch-service-migration/lib/service-stacks/opensearch-container-stack.ts
```diff
@@ -0,0 +1,69 @@
...
+            }
+        ]
+
+        this.createService({
+            serviceName: "opensearch",
+            dockerImageRegistryName: "opensearchproject/opensearch:2",
+            securityGroups: securityGroups,
+            environment: {
+                "cluster.name": "os-docker-cluster",
+                "node.name": "opensearch-node1",
```
gregschohn, line 55: should you say 'test' or target' or target-test?

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -94,6 +110,8 @@ export class StackComposer {
         if (!contextJSON) {
             throw new Error(`No CDK context block found for contextId '${contextId}'`)
         }
+        console.log('Received following context block for deployment: ')
+        console.log(contextJSON)
```
gregschohn, line 114: thanks! I'd recommend putting a trailer after though so that it's clear where it ends.

deployment/cdk/opensearch-service-migration/lib/service-stacks/opensearch-container-stack.ts
lewijacn: Fargate supports overriding default ulimits: https://stackoverflow.com/questions/65289458/ulimits-on-aws-ecs-fargate, which is what I'm utilizing here with defaults that opensearch uses on their website. I'm not sure though on what happens if we were to try and say breach the NOFILE limit

deployment/cdk/opensearch-service-migration/lib/service-stacks/opensearch-container-stack.ts
lewijacn: Just to be clear is this in reference to the `node.name`? In which case I don't have a huge preference would like to have the suffix `-node` on it

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
lewijacn: Added 👍

deployment/cdk/opensearch-service-migration/README.md
```diff
@@ -205,7 +205,7 @@ Although this CDK does not set up the Capture Proxy on source cluster nodes (exc
 
 #### Capture Proxy on OpenSearch/Elasticsearch nodes
 Before [setting up Capture Proxy instances](../../../TrafficCapture/trafficCaptureProxyServer/README.md#how-to-attach-a-capture-proxy-on-a-coordinator-node) on the source cluster, the IAM policies and Security Groups for the nodes should allow access to the Migration tooling:
-1. The coordinator nodes should add the `migrationMSKSecurityGroup` security group to allow access to Kafka
+1. The coordinator nodes should add the `migrationStreamingSourceSG` security group to allow access to Kafka
```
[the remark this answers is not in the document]
lewijacn: Agreed good call will update

deployment/cdk/opensearch-service-migration/lib/service-stacks/opensearch-container-stack.ts
gregschohn: I guess the worst case scenario is that the process dies because it exhausts one of those limits, which it would have probably done when it bumped into the limit anyway.

gregschohn (approved):

PR #473, opened 2023-12-12
Suppress capturing responses

gregschohn:
### Description
This is the successor to [dropping captures for requests](https://github.com/opensearch-project/opensearch-migrations/pull/468).  It builds upon some of the [refactoring that was done for the instrumentation work](https://github.com/opensearch-project/opensearch-migrations/pull/460).
 
* Category: Bug fix, Enhancement
* Why these changes are required? To suppress capturing response packets when the requests weren't captured.
* What is the old behavior before changes and new behavior after changes? Spurious response packets would have followed the dropped observation, or if none of the request was present, they could have followed the last request/response, which would have resulted in invalid looking responses from the source cluster (to the point where it could have been impossible to determine what the responses actually were).

### Testing

### Check List


TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/LoggingHttpHandler.java
```diff
@@ -60,46 +89,50 @@ public HttpMessage createMessage(String[] initialLine) throws Exception {
                     , new PassThruHttpHeaders(headersToPreserve)
             );
         }
-    }
 
+        @Override
+        public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+            if (captureState.captureIgnoreState == CaptureIgnoreState.IGNORE_RESPONSE) {
+                captureState.captureIgnoreState = CaptureIgnoreState.CAPTURE;
+            }
```
lewijacn, line 97: Trying to understand this... If the state is to ignore response and we have a read we will set to capture. Is that because reads in this case will only be for request packets and we can reset back to capture or is something else happening here?

gregschohn: The default assumption is that all requests and their responses will be captured until a request is marked to IGNORE_REQUEST.  This is a reset since a new request has begun... and yes, reads are only for requests and writes are only for responses and that's why we now have two states for ignore.

PR #475, opened 2023-12-15
Replayer instrumentation

gregschohn:
### Description

See the [README](https://github.com/opensearch-project/opensearch-migrations/pull/475/files#diff-831bc42adf6aab6b6de0bb277f083f4d06c96e3e7c23d54725100099b151a22f) included with this PR.

TLDR - lots of proxy and replayer code now passes contexts throughout the call graph.  Those Instumented Contexts are invoked to signal different pieces of progress, which are then turned into metrics and traces that are sent to an otel-collector.  The otel collector is similar to what it was before this PR, but the configurations and its deployment, both in docker and for AWS, have been slightly adjusted.

For AWS Cloud deployments, we now configure CloudWatch and X-Ray to received metrics and traces (via OTEL's Collector).  For the docker solution, we use Prometheus and Jaeger ("all-in-one") to store metric and trace data.  A Grafana container is also started alongside the other containers to allow for discovery and (eventually) dashboards.

* Category Enhancement
* Why these changes are required? Observability

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1438

### Testing
Unit testing + E2E testing

### Check List
- [partial] New functionality includes testing


TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IScopedInstrumentationAttributes.java
```diff
@@ -0,0 +1,44 @@
...
+        meterIncrementEvent(getEndOfScopeCountMetric());
+        meterHistogramMillis(getEndOfScopeDurationMetric());
+    }
+
+    default void close() {
+        endSpan();
+        sendMeterEventsForEnd();
+    }
+
+    default void addException(Throwable e) {
```
AndreKurait, line 36: Should this add a call to setObservedExceptionToIncludeInMetrics or can setObservedExceptionToIncludeInMetrics and associated functions be removed in favor of the exceptionCounter

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/ISpanGenerator.java
```diff
@@ -0,0 +1,8 @@
+package org.opensearch.migrations.tracing;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.Span;
+
+import java.util.function.Function;
+
+public interface ISpanGenerator extends Function<Attributes, Span> { }
```
AndreKurait, line 8: Remove, no usages

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/ISpanWithParentGenerator.java
```diff
@@ -0,0 +1,9 @@
+package org.opensearch.migrations.tracing;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.Span;
+
+import java.util.function.BiFunction;
+
+public interface ISpanWithParentGenerator extends BiFunction<Attributes, Span, Span> {
```
AndreKurait, line 8: Remove, no usages

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IScopedInstrumentationAttributes.java
AndreKurait: I had an older version pulled locally. IInstrumentationAttributes has been changed to remove setObservedExceptionToIncludeInMetrics from IInstrumentationAttributes

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+        return builder; // nothing more to do
+    }
+
+    private static SpanBuilder addLinkedToBuilder(Stream<Span> linkedSpanContexts, SpanBuilder spanBuilder) {
+        return Optional.ofNullable(linkedSpanContexts)
+                .map(ss->ss.collect(Utils.foldLeft(spanBuilder, (b,s)->b.addLink(s.getSpanContext()))))
+                .orElse(spanBuilder);
+    }
+
+    private static Span buildSpanWithParent(SpanBuilder builder, Attributes attrs, Span parentSpan,
```
AndreKurait, line 123: This was pretty hard to understand and read the order of calls on the builder. 

I'd prefer this, note as well a rename of "linkedSpanContexts" to "linkedSpans" to align with the type

```java
    private static SpanBuilder addLinkedSpansToBuilder(Stream<Span> linkedSpans, SpanBuilder spanBuilder) {
        if (linkedSpans != null) {
            linkedSpans.map(Span::getSpanContext).forEach(spanBuilder::addLink);
        }
        return spanBuilder;
    }

    private static SpanBuilder addParentToBuilder(Span parentSpan, SpanBuilder spanBuilder) {
        return parentSpan != null
            ? spanBuilder.setParent(Context.current().with(parentSpan))
            : spanBuilder.setNoParent();
    }

    private static Span buildSpanWithParent(SpanBuilder builder, Attributes attrs, Span parentSpan,
                                            Stream<Span> linkedSpans) {
        addParentToBuilder(parentSpan, builder);
        addLinkedSpansToBuilder(linkedSpans, builder);
        return builder.startSpan().setAllAttributes(attrs);
    }
```

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+                        .setEndpoint(collectorEndpoint)
+                        .setTimeout(2, TimeUnit.SECONDS)
+                        .build())
+                .setScheduleDelay(100, TimeUnit.MILLISECONDS)
+                .build();
+        final var metricReader = PeriodicMetricReader.builder(OtlpGrpcMetricExporter.builder()
+                        .setEndpoint(collectorEndpoint)
+                        // see https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/prometheus/
+                        // "A Prometheus Exporter MUST only support Cumulative Temporality."
+                        //.setAggregationTemporalitySelector(AggregationTemporalitySelector.deltaPreferred())
```
AndreKurait, line 55: Instead of having this code commented which would break the Prometheus exporter, should we instead call

```java
.setAggregationTemporalitySelector(AggregationTemporalitySelector.alwaysCumulative())
```

even thought it is the default in the builder to make it explicit as a requirement in our usage.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+                                                                    @NonNull String serviceName) {
+        var serviceResource = Resource.getDefault().toBuilder()
+                .put(ResourceAttributes.SERVICE_NAME, serviceName)
+                .build();
+
+        final var spanProcessor = BatchSpanProcessor.builder(OtlpGrpcSpanExporter.builder()
+                        .setEndpoint(collectorEndpoint)
+                        .setTimeout(2, TimeUnit.SECONDS)
+                        .build())
+                .setScheduleDelay(100, TimeUnit.MILLISECONDS)
```
AndreKurait, line 49: Is there a reason we may want the span's processed at a 10x faster interval than the metrics?

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+
+        // Add hook to close SDK, which flushes logs
+        Runtime.getRuntime().addShutdownHook(new Thread(openTelemetrySdk::close));
+        return openTelemetrySdk;
+    }
+
+    public static OpenTelemetry initializeOpenTelemetry(String collectorEndpoint, String serviceName) {
+        return Optional.ofNullable(collectorEndpoint)
+                .map(endpoint -> initializeOpenTelemetryForCollector(endpoint, serviceName))
+                .orElse(OpenTelemetrySdk.builder().build());
```
AndreKurait, line 81: Do we intend to ignore serviceName when a collectorEndpoint is not passed

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.stream.Stream;
+
+public class RootOtelContext implements IRootOtelContext {
+    private final OpenTelemetry openTelemetryImpl;
+    private final String scopeName;
+    @Getter
+    @Setter
+    Exception observedExceptionToIncludeInMetrics;
+
+    public static OpenTelemetry initializeOpenTelemetryForCollector(@NonNull String collectorEndpoint,
```
AndreKurait, line 39: nit: I prefer limiting public methods to reduce integration choices later (i.e. have other parts of code call null-safe method)

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+import java.time.Duration;
+import java.util.Optional;
+import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
+
+public class RootOtelContext implements IRootOtelContext {
+    private final OpenTelemetry openTelemetryImpl;
+    private final String scopeName;
+    @Getter
+    @Setter
```
AndreKurait, line 36: If we expect extending classes to set this value with the setter, let's add it to the IRootOtelContext

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    @Override
+    public MeterProvider getMeterProvider() {
+        return getOpenTelemetry().getMeterProvider();
+    }
+
+    @Override
+    public AttributesBuilder fillAttributes(AttributesBuilder builder) {
+        return builder; // nothing more to do
```
AndreKurait, line 114: For consistency, should we call the fillAttributes method from IRootOtelContext/IInstrumentationAttributes? 

Also, do we need to set the hadException attribute?

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/ISpanGenerator.java
gregschohn: Done

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/ISpanWithParentGenerator.java
gregschohn: Done

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
```diff
@@ -0,0 +1,109 @@
...
+        return getPopulatedSpanAttributes(Attributes.builder());
+    }
+
+    default Attributes getPopulatedSpanAttributes(AttributesBuilder builder) {
+        return getPopulatedSpanAttributesBuilder(builder).build();
+    }
+
+    default AttributesBuilder getPopulatedSpanAttributesBuilder(AttributesBuilder builder) {
+        var currentObj = this;
+        var stack = new ArrayList<IInstrumentationAttributes>();
```
AndreKurait, line 44: Nit: an ArrayDeque may make this a bit easier to follow

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
```diff
@@ -0,0 +1,109 @@
...
+
+    default Attributes getPopulatedSpanAttributes() {
+        return getPopulatedSpanAttributes(Attributes.builder());
+    }
+
+    default Attributes getPopulatedSpanAttributes(AttributesBuilder builder) {
+        return getPopulatedSpanAttributesBuilder(builder).build();
+    }
+
+    default AttributesBuilder getPopulatedSpanAttributesBuilder(AttributesBuilder builder) {
```
AndreKurait, line 42: It seems a bit strange here that the passed in AttributesBuilder's values will be overriden if set in any enclosingScope, initially I was thinking it would have priority over any enclosing values since presumably it's set by the caller in the current scope.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: The code above will only work if the builder is a mutable builder.  The way that the code is written now, the builder could be mutable or immutable since the newly returned builder is returned as input into the next call.  I could use the javadocs to see that 'this' is returned, but that requires the programmer to remember that convention and makes it harder to do audits over the code.  Looking to rebind the "current builder" as  the newly returned builder value is what makes it difficult to call two functions in sequence (rather than compositionally).

The middle ternary could replace the monadic ofNullable, but I prefer the ofNullable because it creates a newer & tighter scope for the nonNullable value that we're operating on and fences values my from leaving their scope during refactorings.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
```diff
@@ -0,0 +1,109 @@
...
+        var currentObj = this;
+        var stack = new ArrayList<IInstrumentationAttributes>();
+        while (currentObj != null) {
+            stack.add(currentObj);
+            currentObj = currentObj.getEnclosingScope();
```
AndreKurait, line 47: It seems that a span is guaranteed to have all the attributes of it's parent from buildSpan definition, so is it the IInstrumentationAttributes job to loop through all the enclosingScopes to re-build these attributes?

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Do you think it would make sense to do both?  I'd like to be able to find the calls when I'm searching and looking at our configuration code.  These APIs are fast evolving so maybe it changes and, more likely, there's a good chance that somebody not very familiar with otel will try to make the same mistake.  The goal here isn't to make sure that we do what prometheus needs (we may not always use prometheus) but to give the reason that we can't do more.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Good point.
I'm not sure why I clocked that so tightly.  The default is 5s.  I'll kick it back to that in case there's a way eventually load that via an environment variable, etc.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -0,0 +1,139 @@
...
+import io.opentelemetry.exporter.otlp.trace.OtlpGrpcSpanExporter;
+import io.opentelemetry.sdk.OpenTelemetrySdk;
+import io.opentelemetry.sdk.logs.SdkLoggerProvider;
+import io.opentelemetry.sdk.logs.export.BatchLogRecordProcessor;
+import io.opentelemetry.sdk.metrics.SdkMeterProvider;
+import io.opentelemetry.sdk.metrics.export.PeriodicMetricReader;
+import io.opentelemetry.sdk.resources.Resource;
+import io.opentelemetry.sdk.trace.SdkTracerProvider;
+import io.opentelemetry.sdk.trace.export.BatchSpanProcessor;
+import io.opentelemetry.semconv.resource.attributes.ResourceAttributes;
```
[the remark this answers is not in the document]
gregschohn: I've swapped the semconv jarfile and now refer to [ResourceAttributes](https://github.com/open-telemetry/semantic-conventions-java/blob/2be178a7fd62d1073fa9b4f0f0520772a6496e0b/src/main/java/io/opentelemetry/semconv/ResourceAttributes.java).

jarfile switch
```
-    implementation group: 'io.opentelemetry', name: 'opentelemetry-semconv', version: '1.30.0-alpha'
+    implementation group: 'io.opentelemetry.semconv', name: 'opentelemetry-semconv', version: '1.23.1-alpha'
```

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Done

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: I've added some documentation, a warning, and have changed the name to 'initializeOpenTelemetryWithCollectorOrAsNoop'.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Why would you want to reduce integration choices later?  The initializeOpenTelemetryWithCollectorOrAsNoop has a looser contract.  If a caller wants to make sure that they create a collector-backed OpenTelemetry object, they would need to call the loose one, then do their own checks.  It seems like it's a noop at the moment for us to maintain both?
I actually see the outer one as a convenience function.  I could see a method to return the empty factory which takes no arguments, which I've also added (meaning that there are now 3 public functions).
I could see keeping these as protected methods, but it's too cumbersome to overload all of the different constructors for each subclass, hence the reason that I'm putting it directly into the caller's hands.  I've also already committed to making Otel classes part of the API, so the decisions fit together, in my head... What do you think?

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Thanks - the setter shouldn't be here.  The root context should never be using this, the value should always be null.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Hmmmm... we could call fillAttributes, but this class is a very special case of an attributes object.  It serves as the base case so that it can be the root object of a hierarchy of contexts to give everybody in the tree access to its functionality.  However, if the interface's fillAttributes function is modified (even slightly), it will be to support a general case, not a one-off case.  
My take is that this is a tie.  Do you think that there's an argument that we should be calling the interface's default impl?

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
gregschohn: Good idea.  I'm not sure that the new code is easier to read, but I like it.
```
        var stack = new ArrayDeque<IInstrumentationAttributes>();
        while (currentObj != null) {
            stack.add(currentObj);
            currentObj = currentObj.getEnclosingScope();
        }
        // reverse the order so that the lowest attribute scopes will overwrite the upper ones if there were conflicts
        return StreamSupport.stream(
                Spliterators.spliteratorUnknownSize(stack.descendingIterator(), Spliterator.ORDERED), false)
                .collect(Utils.foldLeft(builder, (b, iia)->iia.fillAttributes(b)));
```

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
gregschohn: Hmmmm - thanks.  That's a good point & not really what I should be intending to do.  This is a throwback from when different types of attributeBuilders were passed in here.

gregschohn: I've deleted the AttributesBuilder as an argument.  Now it would be up to the caller to set the values on the AttributeBuilder AFTER filling them (from all of the enclosing scopes)

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
gregschohn: Yes.  It sounds like you're thinking that we could pull the enclosing scope's span value and not have to do the rebuilding.  That's probably correct for most of the code paths today, but one fuzzy spot would be that if an attribute was added to a span outside of fillAttributes (ie. getCurrentSpan().addAttribute() (e.g. adding an event or an exception).  Should we pick those up in downstream scopes?  Probably not.
However, should we be picking up parent scopes' attributes at all?  I really want to say "no", but early on in testing, I've noticed that if parent spans aren't closed out, only the children end up getting rendered by jaeger (and maybe exported by otel).  That leaves us in a bind to make children spans useful since they don't have any attributes.
I'd be interested in talking this through some.  It feels like a nuanced decision, though maybe one that doesn't matter too much.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
AndreKurait: I was thinking a more standard `for each`.

Note, the `add` method of a deque is on the front so by the time it's filled it's in the order you want to iterate.

```java
// Fill Attributes in LIFO order to prioritize nested scope's attributes
var attributeAccumulator = Attributes.builder();
for (IInstrumentationAttributes obj : stack) {
  attributeAccumulator = obj.fillAttributes(attributeAccumulator);
}
return attributeAccumulator;
```

If you must use Streams, you can use Streams::reduce or your foldLeft
```java
// Stream directly over the stack in LIFO order to prioritize nested scope's attributes
return stack.stream().collect(
  Utils.foldLeft(Attributes.builder(),
    (accumBuilder, attributes) -> attributes.fillAttributes(accumBuilder))
  );
```

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/tracing/WireCaptureContexts.java
```diff
@@ -0,0 +1,284 @@
...
+        @Override
+        public void onBytesRead(int size) {
+            meterIncrementEvent(getMetrics().bytesReadCounter, size);
+        }
+    }
+
+    public static class BlockingContext
+            extends HttpMessageContext
+            implements IWireCaptureContexts.IBlockingContext {
+        public static final String ACTIVITY_NAME = "blocked";
```
AndreKurait, line 182: Should we align this to a ActivityNames class like the replayer code?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/tracing/IReplayContexts.java
```diff
@@ -0,0 +1,347 @@
...
+import org.opensearch.migrations.replay.datatypes.ITrafficStreamKey;
+import org.opensearch.migrations.replay.datatypes.UniqueReplayerRequestKey;
+import org.opensearch.migrations.tracing.IScopedInstrumentationAttributes;
+import org.opensearch.migrations.tracing.IWithTypedEnclosingScope;
+
+import java.time.Instant;
+
+public abstract class IReplayContexts {
+
+    public static class ActivityNames {
```
AndreKurait, line 15: Would this be a lot more efficient as a protobuf defined enum? Reading the [OTEL Spec](https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding), it seems they have made optimizations to require integer serialization for enums which is stricter than the standard protobuf spec.

AndreKurait: Let me know if you think that is just referring to the otel structure itself as opposed to the span attributes

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IWithStartTimeAndAttributes.java
```diff
@@ -0,0 +1,23 @@
...
+import io.opentelemetry.api.metrics.DoubleHistogram;
+
+import java.time.Duration;
+import java.time.Instant;
+
+public interface IWithStartTimeAndAttributes extends IInstrumentationAttributes {
+    Instant getStartTime();
+
+    default Duration getSpanDuration() {
+        return Duration.between(getStartTime(), Instant.now());
```
AndreKurait, line 13: For duration metrics, I prefer using `System.nanoTime()` to better handle wall clock changes like during synchronization. Thoughts on storing a wall clock start time with instant as well as the System.nanoTime() start to use on durations?

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/tracing/KafkaRecordContext.java
```diff
@@ -0,0 +1,62 @@
...
+    @Override
+    public @NonNull MetricInstruments getMetrics() {
+        return getRootInstrumentationScope().getKafkaOffloadingInstruments();
+    }
+
+    @Override
+    public String getActivityName() { return "stream_flush_called"; }
+
+    @Override
+    public AttributesBuilder fillAttributes(AttributesBuilder builder) {
```
AndreKurait, line 57: Discussed offline, my preference is to chain our extending interface's fillAttributes so that our metrics are additive. May require adding an override to in between interfaces to satisfy java rules.

Also, call the attributeBuilder puts after the super call.

AndreKurait:
Note, performed some benchmarks with and without this code. Regarding CPU usage, the ratio between the Capture Proxy CPU and ElasticSearch CPU was 33:67, without the PR, it was 27:73

Notably, the memory usage of the Capture Proxy increased from 273MB to 475MB.

I find these both to be acceptable at the moment and will improve over time

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
gregschohn: There should be inheritable values (like ids) and non-inheritable ones (like exceptions, sizes, and non-identifying runtime attributes).  I'll work on making that change.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/kafka/KafkaTracingTest.java
```diff
@@ -0,0 +1,10 @@
+package org.opensearch.migrations.replay.kafka;
```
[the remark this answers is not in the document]
gregschohn: I'm not sure what it was for.  I'm deleting it.

TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/tracing/WireCaptureContexts.java
gregschohn: Done - thanks

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/tracing/IReplayContexts.java
gregschohn: I'm not sure that it is relevant for us, but knowing how protobuf works, they'd need to be creating on the fly schema definitions anyway & I don't think that it would matter if these were strings (that could be mapped) or ints (which could reach into the billions).

There might be a lot of value in doing this as an enum.  I plan to pull MetricNames and ActivityNames at some point into tests to confirm that they're all mapping to observed traces and metrics within tests (effectively more coverage stats).

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IWithStartTimeAndAttributes.java
gregschohn: Thanks - good point

gregschohn: Here's a nice history of nanoTime on parallel systems: https://stackoverflow.com/questions/510462/is-system-nanotime-completely-useless/54566928#54566928

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/tracing/KafkaRecordContext.java
gregschohn: This is cumbersome to do with Java, but the compiler does provide some protections.  For a default method defined within a hierarchy of interfaces, I can specify the name of any interface _after_ the deepest default implementation and those will compile without error.  **However**, if refer to say `IInstrumentationAttributes.super.fillAttributes()` but then implement fillAttributes() in IWithStartTimeAndAttributes, the compiler will complain with `java: not an enclosing class: org.opensearch.migrations.tracing.IWithStartTimeAndAttributes`

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
```diff
@@ -0,0 +1,112 @@
...
+        var builder = Attributes.builder();
+        var currentObj = this;
+        var stack = new ArrayDeque<IInstrumentationAttributes>();
+        while (currentObj != null) {
+            stack.add(currentObj);
+            currentObj = currentObj.getEnclosingScope();
+        }
+        // reverse the order so that the lowest attribute scopes will overwrite the upper ones if there were conflicts
+        return StreamSupport.stream(
+                Spliterators.spliteratorUnknownSize(stack.descendingIterator(), Spliterator.ORDERED), false)
```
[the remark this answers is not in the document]
gregschohn: done & now there's a test to verify that it is correct

TrafficCapture/dockerSolution/README.md
```diff
@@ -8,13 +8,61 @@ down again.
...
+file into the container to provide the collector credentials that it needs to push metrics and traces to CloudWatch
+and X-Ray.  In addition to the [default configuration](src/main/docker/composeExtensions/otel-prometheus-jaeger-opensearch.yml)
+to use local containers, a [third option](src/main/docker/composeExtensions/otel-everything.yml) will use _BOTH_ local
+containers AND the AWS services.
+
+## Maintaining configurations
+
+### Otel-Collector configurations
+
+The migrations components use OpenTelemetry for instrumentation so that different systems can be utilized acrpss
```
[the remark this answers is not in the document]
gregschohn: Thx - corrcted

TrafficCapture/dockerSolution/otelConfigs/configSnippets/base.yaml
```diff
@@ -0,0 +1,14 @@
+receivers:
+  otlp:
+    protocols:
+      grpc:
+
+processors:
```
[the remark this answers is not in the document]
gregschohn: I don't believe that that works when processors and exporters aren't present.  If memory serves, when I included opensearch as an exporter and started up a collector that hadn't built that exporter, I got errors or warnings.  Secondly, in some cases, configurations are contradictory - sure, I could figure out if we could setup 2 configurations and choose one.  However, putting an insecure setup into an image that we're shipping for production is even less appealing - then I'm back to figuring out how to mask/add between the different projected scenarios.

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/IInstrumentationAttributes.java
```diff
@@ -43,17 +33,24 @@ default Attributes getPopulatedSpanAttributes() {
...
+        var builder = stack.stream()
+                .collect(Utils.foldLeft(Attributes.builder(), (b, iia)->iia.fillAttributesForSpansBelow(b)));
+        return fillExtraAttributesForThisSpan(builder);
+    }
+
+    default AttributesBuilder fillAttributesForSpansBelow(AttributesBuilder builder) {
+        return builder;
+    }
+
+    default AttributesBuilder fillExtraAttributesForThisSpan(AttributesBuilder builder) {
```
AndreKurait, line 52: Do we not need to add hadException here?

AndreKurait (commented):
Barring an updated PR description and merge commit, i'm happy with these changes

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/CommonScopedMetricInstruments.java
```diff
@@ -0,0 +1,61 @@
...
+    public CommonScopedMetricInstruments(Meter meter, String activityName,
+                                         double firstBucketSize, double lastBucketCeiling) {
+        this(meter, activityName, getBuckets(activityName, firstBucketSize, lastBucketCeiling));
+    }
+
+    private static List<Double> getBuckets(String activityName,
+                                           double firstBucketSize, double lastBucketCeiling) {
+        var buckets = getExponentialBucketsBetween(firstBucketSize, lastBucketCeiling, 2.0);
+        log.atInfo().setMessage(() -> "Setting buckets for " + activityName + " to " +
+                buckets.stream().map(x -> "" + x).collect(Collectors.joining(",", "[", "]"))).log();
```
AndreKurait, line 31: What is this actually printing? Are we accurately showing what's a < end vs <=?

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/CommonScopedMetricInstruments.java
```diff
@@ -27,22 +26,30 @@ public CommonScopedMetricInstruments(Meter meter, String activityName,
...
+        return buckets;
+    }
+
+    private static List<Double> getExponentialBucketsBetween(double firstBucketSize, double lastBucketCeiling) {
+        return getExponentialBucketsBetween(firstBucketSize, lastBucketCeiling, 2.0);
+    }
+
+    private static List<Double> getExponentialBucketsBetween(double firstBucketSize, double lastBucketCeiling,
```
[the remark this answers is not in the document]
AndreKurait: Thanks!

AndreKurait (approved):

AndreKurait (approved):

AndreKurait (changes requested):
Please address CI/CD build failure

PR #478, opened 2023-12-20
[Fetch Migration] Suppress exit code for graceful termination (SIGTERM)

kartg:
### Description
The orchestrator now suppresses exit code `143` (graceful termination following a `SIGTERM`) if it is returned by the migration monitor, since this is a result of a workaround for https://github.com/opensearch-project/data-prepper/issues/3141

Unit tests have been added to cover this scenario. This change also ensures that a return_code is always returned by the function, and updates some stale comments.
* Category: Bug fix

### Testing
Unit test coverage:
```
endpoint_info.py                  24      0   100%
endpoint_utils.py                106      1    99%
fetch_orchestrator.py             76      0   100%
fetch_orchestrator_params.py      22      0   100%
index_diff.py                     20      0   100%
index_doc_count.py                 5      0   100%
index_operations.py               76      0   100%
metadata_migration.py             65      0   100%
migration_monitor.py              90      0   100%
progress_metrics.py               91      0   100%
TOTAL                            606      1    99%

```

### Check List

### DIFF
--- a/FetchMigration/python/fetch_orchestrator.py
+++ b/FetchMigration/python/fetch_orchestrator.py
@@ -75,7 +75,7 @@ def write_inline_target_host(pipeline_file_path: str, inline_target_host: str):
         yaml.safe_dump(pipeline_yaml, pipeline_file)
 
 
-def run(params: FetchOrchestratorParams) -> Optional[int]:
+def run(params: FetchOrchestratorParams) -> int:
     # This is expected to be a base64 encoded string
     inline_pipeline = __get_env_string("INLINE_PIPELINE")
     inline_target_host = __get_env_string("INLINE_TARGET_HOST")
@@ -93,6 +93,7 @@ def run(params: FetchOrchestratorParams) -> Optional[int]:
                                                         report=True, dryrun=params.is_dry_run)
     logging.info("Running metadata migration...\n")
     metadata_migration_result = metadata_migration.run(metadata_migration_params)
+    return_code: int = 0
     if metadata_migration_result.target_doc_count == 0:
         logging.warning("Target document count is zero, skipping data migration...")
     elif len(metadata_migration_result.migration_indices) > 0 and not params.is_only_metadata_migration():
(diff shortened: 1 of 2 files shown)


FetchMigration/python/fetch_orchestrator.py
```diff
@@ -103,8 +104,12 @@ def run(params: FetchOrchestratorParams) -> Optional[int]:
         migration_monitor_params = MigrationMonitorParams(metadata_migration_result.target_doc_count,
                                                           params.get_local_endpoint())
         logging.info("Starting migration monitor...\n")
-        return migration_monitor.run(migration_monitor_params, proc)
+        return_code = migration_monitor.run(migration_monitor_params, proc)
+        # Suppress non-zero return code for graceful termination (SIGTERM)
+        if return_code == 143:
```
gregschohn, line 109: A citation to the DataPrepper code (github) showing where/why it returns 143 would be great.  143 seems like it could be a fragile number to pin on.  This should also be a constant to document it a bit more.
I guess DataPrepper can also just run in "one-off" mode where it exits after a task is done (and then exits with code 0)?

gregschohn (commented):
Is there an integ test in place that runs DataPrepper and confirms that the error code and this codebase works as expected?  If so, LGTM.  If not, it would be really nice to have that to ensure that we'll know when DataPrepper changes and we need to make an update.

FetchMigration/python/fetch_orchestrator.py
kartg: `143` is the [standard exit code](https://itsfoss.com/linux-exit-codes/) for a `SIGTERM` signal to indicate a graceful shutdown of the process. This behavior isn't managed by Data Prepper but rather [by the JVM](http://journal.thobe.org/2013/02/jvms-and-kill-signals.html).

I wasn't able to find any system libraries in Python that have this declared as a constant. IMO a test case isn't needed for this behavior since it is the expected behavior for any Linux process. Let me know if you disagree?

gregschohn (approved):

PR #488, opened 2024-01-18
FEATURE: Add ability to export Kafka records

lewijacn:
### Description
Introduces a `kafkaExport.sh` script onto the migration console for exporting all initially detected records for a given topic and n number of partitions to a gzip archive file. This uses the packaged Kafka console scripts to detect all partitions for a topic and the given number of records in each partition. Then feeds this info to the KafkaPrinter which has been updated to allow limits for partitions, and will complete when all limits have been reached.

image

Also included was the ability to export this archive to S3, mainly for the AWS use case but applicable to a docker case with the proper AWS credentials. To accommodate this, an S3 bucket for `migration-artifacts` is now created in the CDK deployment and linked to the migration console so that needed permissions are available for exports. This S3 bucket should also be useful for any Replayer log/tuple exports or other use cases in the future.

image

#### Minor additional changes:
* Added ability to specify a different docker build context besides the directory of the Dockerfile for CDK ECS services. This is not currently in use by any of the ECS services, but was in use with testing having a build context at the root of the project.
* Increased OSB to 1.2 which alleviated a conflict between awscli and osb previously encountered on the migration console.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1458
https://opensearch.atlassian.net/browse/MIGRATIONS-1459

### Testing
Local/Cloud testing and unit testing for KafkaPrinter

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/KafkaPrinter.java
```diff
@@ -126,26 +192,61 @@ public static void main(String[] args) {
...
+            }
         }
     }
 
-    static void processNextChunkOfKafkaEvents(Consumer<String, byte[]> kafkaConsumer, java.util.function.Consumer<Stream<byte[]>> binaryReceiver) {
+    static void processNextChunkOfKafkaEvents(Consumer<String, byte[]> kafkaConsumer, java.util.function.Consumer<Stream<ConsumerRecord<String, byte[]>>> binaryReceiver) {
         var records = kafkaConsumer.poll(CONSUMER_POLL_TIMEOUT);
-        binaryReceiver.accept(StreamSupport.stream(records.spliterator(), false)
-                .map(ConsumerRecord::value));
+        binaryReceiver.accept(StreamSupport.stream(records.spliterator(), false));
```
gregschohn, line 230: do you want to flatten records from across different partitions into just one stream?  See my comments about preserving partition boundaries.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/KafkaPrinter.java
```diff
@@ -34,9 +40,42 @@
  * 3. Clearing records for a topic with the same group id
  */
 public class KafkaPrinter {
+
     private static final Logger log = LoggerFactory.getLogger(KafkaPrinter.class);
     public static final Duration CONSUMER_POLL_TIMEOUT = Duration.ofSeconds(1);
 
+    static class Partition {
```
gregschohn, line 47: There's already a tuple class for topics and partitions in org.apache.kafka.common (TopicPartition)

gregschohn (approved):
There are a number of changes that I'd like to see to this PR, but the risk and tech-debt being introduced in the grand scheme of the codebase is minor, so I'll defer to your judgment about what should or needs to be addressed now and what can wait.

Since we don't know what the long-term needs are for using kafka topics exported in the format supported here, I'm fine with a more incremental approach.  Do we have any tests on the replayer to confirm that it works (well) with files of this format?  We DO have tests that integrate kafka and the replayer, but I cannot recall any for delimited stdio streams.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/KafkaPrinter.java
lewijacn: Have added default to preserve partition boundaries and create separate files

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/KafkaPrinter.java
```diff
@@ -164,24 +162,49 @@ public static void main(String[] args) {
...
 
+        String baseOutputPath = params.outputDirectoryPath == null ? "./" : params.outputDirectoryPath;
+        baseOutputPath = !baseOutputPath.endsWith("/") ? baseOutputPath + "/" : baseOutputPath;
+        String uuid = UUID.randomUUID().toString();
+        boolean separatePartitionOutputs = false;
+        Map<Integer, CodedOutputStream> partitionOutputStreams = new HashMap<>();
+        // Grab all partition records
+        if (capturedRecords.isEmpty()) {
+            OutputStream os = params.outputDirectoryPath == null ? System.out : new FileOutputStream(String.format("%s%s_%s_%s.proto", baseOutputPath, params.kafkaTrafficTopic, "all", uuid));
+            partitionOutputStreams.put(0, CodedOutputStream.newInstance(os));
```
gregschohn, line 188: This block can roll into the next block - excise and make the first if statement include `captureRecords.isEmpty() ||...`

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/kafkaExport.sh
```diff
@@ -55,25 +59,29 @@ while [[ $# -gt 0 ]]; do
...
+echo "Group name: $group"
 
 set -o xtrace
-./runJavaWithClasspath.sh org.opensearch.migrations.replay.KafkaPrinter --kafka-traffic-brokers "$broker_endpoints" --kafka-traffic-topic "$topic" --kafka-traffic-group-id "$group" $(echo "$msk_auth_settings") --timeout-seconds "$timeout_seconds" --partition-limit "$comma_sep_partition_offsets" >> "$file_name"
+./runJavaWithClasspath.sh org.opensearch.migrations.replay.KafkaPrinter --kafka-traffic-brokers "$broker_endpoints" --kafka-traffic-topic "$topic" --kafka-traffic-group-id "$group" $(echo "$msk_auth_settings") --timeout-seconds "$timeout_seconds" --partition-limits "$comma_sep_partition_offsets" --output-directory "./$dir_name"
 set +o xtrace
 
-tar -czvf "$archive_name" "$file_name" && rm "$file_name"
+cd $dir_name
+tar -czvf "$archive_name" *.proto && rm *.proto
```
gregschohn, line 79: I'm fine doing this later, but this whole script would be much faster and easier to manage if you wrote compressed streams first, then tarred them together.  As it is, you'll have a huge number of bytes going to disk (which might be remote) and back.  Since the java program uses multiple files, you'd have to manage it within the java program (with a GZipInputStream).

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/KafkaPrinter.java
```diff
@@ -230,31 +253,47 @@ static void processNextChunkOfKafkaEvents(Consumer<String, byte[]> kafkaConsumer
         binaryReceiver.accept(StreamSupport.stream(records.spliterator(), false));
     }
 
-    static java.util.function.Consumer<Stream<ConsumerRecord<String, byte[]>>> getDelimitedProtoBufOutputter(OutputStream outputStream, Map<Partition, PartitionTracker> capturedRecords) {
-        CodedOutputStream codedOutputStream = CodedOutputStream.newInstance(outputStream);
+    static java.util.function.Consumer<Stream<ConsumerRecord<String, byte[]>>> getDelimitedProtoBufOutputter(Map<TopicPartition, PartitionTracker> capturedRecords,
+        Map<Integer, CodedOutputStream> partitionOutputStreams, boolean separatePartitionOutputs) {
```
gregschohn, line 257: I should have asked before, but why are you using a CodedOutputStream (full disclosure, I wrote it with a CodedOutputStream) here instead of [writeDelimitedTo](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/MessageLite.html#writeDelimitedTo-java.io.OutputStream-)?  That would let you pass in a compressed output stream and you wouldn't need to worry about the uncompressed buffering problem.

gregschohn (approved):
There are some things that would be nice to change, but they aren't critical; especially given the support nature of this code.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/KafkaPrinter.java
lewijacn: I like the idea of using this `writeDelimitedTo` function but it seems to be on the Message interface, which for us it seems we would have to take the byte[] and have our TrafficStream protobuf object parse it, then we could use `writeDelimitedTo`. I wasn't sure after these requirements if there would be much benefit

gregschohn: Ahhh - good point.  The current code is probably more efficient, so until we have a reason to, there's probably not a good reason to make the change now.

PR #510, opened 2024-02-16
Flaky tcp connection trace test fix

gregschohn:
### Description
tcpConnections weren't being properly tracked.  Now they are because they're tied to the netty channel itself.  I've also improved some logging and fixed a multi-threading bug in shutdown.

* Category Bug/test fix
* Why these changes are required? Metrics and traces were wrong.
* What is the old behavior before changes and new behavior after changes? Flaky test and now that test should work all the time

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1502

### Testing
./gradlew build slowTest

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -227,7 +239,11 @@ private void deactivateChannel() {
                     log.atDebug().setMessage(()->"Ignoring an exception that the "+handlerName+" wasn't present").log();
                 }
             }
-            while (!(pipeline.last() instanceof SslHandler) && (pipeline.last() != null)) {
+            while (true) {
```
AndreKurait, line 242: Did we intend to remove the `&& (pipeline.last() != null)` condition?

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/FullReplayerWithTracingChecksTest.java
```diff
@@ -122,42 +125,70 @@ public void testStreamWithRequestsWithCloseIsCommittedOnce(int numRequests) thro
         log.info("done");
     }
 
+    private static class TraceProcessor {
+        Map<String, List<SpanData>> byName;
+        public TraceProcessor(InMemorySpanExporter testSpanExporter) {
+            byName = testSpanExporter.getFinishedSpanItems().stream().collect(Collectors.groupingBy(SpanData::getName));
+        }
+
+        private int getCount(String k) {
```
AndreKurait, line 134: Can we rename this to imply we're removing the Span?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
gregschohn: Yes, the ConnectionClosedListenerHandler handler will ALWAYS be present at the head of the pipeline now.  Before, we could have either SSL or no handler, but the ConnectionClosedListenerHandler is always applicable and should be carried forward until the connection is closed (at which point, the pipeline will be torn down)

PR #517, opened 2024-02-21
Fix some flaky tests that were checking if metrics were being emitted.

gregschohn:
Fix some flaky tests that were checking if metrics were being emitted.

They had a resource lock around them because they were using some static constructs provided by the Otel SDK. Now the resource locks on tests are gone for the ConditionallyReliableLoggingHttpHandlerTest and each test constructs its own non-static otel sdk. I've also tried to make tests that confirm otel instrumentation a bit safer.  They properly shut down metric exporter threads now.  They also wait a couple cycles before returning the in-memory exported metrics. I really wish that there were an exporter that didn't have to wait, but at least for now, this seems like the only option. At some point, it might be worth implementing a different MeterProvider which would in turn build out a different implementation hierarchy to return instruments that have getters.  Using mock classes to spy on calls could work too.

### Description

* Category: Test fixes
* Why these changes are required? Make builds more reliable
* What is the old behavior before changes and new behavior after changes? Tests should be more stable and use less resources (threads)

### Issues Resolved
No Jira

### Testing

### Check List


AndreKurait:
> I really wish that there were an exporter that didn't have to wait.

Why can't we use InMemoryMetricReader/InMemoryMetricExporter?

AndreKurait (approved):

PR #518, opened 2024-02-21
Enable Gradle Build Scans

AndreKurait:
### Description
Enable Gradle Build Scans with option to auto agree to tos and add scan for all runs. Approved by AWS Security (Internal Ticket V1251862415)
* Category -  New Feature
* Why these changes are required? Currently it is very difficult to debug failues in git workflows for test/build execution. This enables gradle scan which uploads and provides a UI for this. 
* What is the old behavior before changes and new behavior after changes? Before, no scan published, now Scan published.

### Issues Resolved
N/A

N/A

### Testing
View Actions Results in this PR. Ran with --scan, --no-scan and no parameter with and without the env variable set to confirm behavior.

### Check List


.github/workflows/e2eTest.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           cache-read-only: false
 
       - name: Start Docker Solution
-        run: ./gradlew dockerSolution:ComposeUp -x test
+        run: ./gradlew dockerSolution:ComposeUp -x test --scan --info --stacktrace
```
gregschohn, line 33: Is --info bumping the level just for scan or for all output?  If it's for all output, do you know how much bigger the logfiles will grow?

TrafficCapture/settings.gradle
```diff
@@ -40,3 +40,13 @@ addSubProjects('', new File(rootProject.projectDir,'replayerPlugins'))
 include 'replayerPlugins:jsonMessageTransformers:jsonJoltMessageTransformerProvider:untitled'
 findProject(':replayerPlugins:jsonMessageTransformers:jsonJoltMessageTransformerProvider:untitled')?.name = 'untitled'
 
+// Enable Gradle Scan Publishing for CI/CD
+if (System.getenv().containsKey("CI")) {
```
gregschohn, line 44: How can I stop getting prompted when I run --scan on my own command line?  Should this CI check be something more specific to our repo?  If a user runs on a shell with CI as a variable, they might not have agreed to the terms.

gregschohn (changes requested):

TrafficCapture/settings.gradle
AndreKurait: I don't have any objections to make it over the whole repo

gregschohn: Can we key this off of an environment variable like "I_AGREE_TO_TERMS_AND_SERVICE_FOR_GRADLE_SCANS"?  That variable wouldn't be accidentally set and is clear.  I'm concerned that if you set it to something generic, or nothing, that somebody will fork the repo, add something that they shouldn't send, and not even realize that they had done so.

.github/workflows/e2eTest.yml
AndreKurait: It's only for the Gradle specific logs (like reason for task cache ignored), minimally impacts the size

TrafficCapture/gradlew
```diff
@@ -241,4 +241,11 @@ eval "set -- $(
         tr '\n' ' '
     )" '"$@"'
 
+# Check if OS_MIGRATIONS_GRADLE_SCAN_TOS_AGREE_AND_ENABLED exists, and if --scan is not in the arguments and --no-scan is not provided
+if [ -n "${OS_MIGRATIONS_GRADLE_SCAN_TOS_AGREE_AND_ENABLED+set}" ] && ! printf '%s\n' "$@" | grep -q -- '--scan' && ! printf '%s\n' "$@" | grep -q -- '--no-scan'; then
+    echo "Enabling Gradle Scan based on environment variable OS_MIGRATIONS_GRADLE_SCAN_TOS_AGREE_AND_ENABLED"
+    set -- "$@" '--scan'  # Append --scan to the arguments
+fi
```
gregschohn, line 248: This file is autogenerated.  Whenever we upgrade gradle and/or rerun the wrapper task, this will be recreated, removing these lines.  If you could do this within the build.gradle files, the change would be more durable and easier to maintain.

TrafficCapture/settings.gradle
```diff
@@ -40,3 +40,11 @@ addSubProjects('', new File(rootProject.projectDir,'replayerPlugins'))
 include 'replayerPlugins:jsonMessageTransformers:jsonJoltMessageTransformerProvider:untitled'
 findProject(':replayerPlugins:jsonMessageTransformers:jsonJoltMessageTransformerProvider:untitled')?.name = 'untitled'
 
+if (System.getenv().containsKey("OS_MIGRATIONS_GRADLE_SCAN_TOS_AGREE_AND_ENABLED") && hasProperty('gradleEnterprise')) {
```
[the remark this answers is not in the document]
AndreKurait: If someone has the env variable set, but then passes `--no-scan` to disable it, we will have an error if we try to set variables to the gradleEnterprise configuration

TrafficCapture/gradlew
AndreKurait: I agree, but i had not found a way to do this inside build.gradle hence it's inclusion here

gregschohn (approved):
LGTM, provided that there's nothing alarming in the checks

PR #520, opened 2024-02-23
MIGRATIONS-1519: Add support for providing Fargate CPU arch in CDK

lewijacn:
### Description
Add CDK option to provide default CPU arch `defaultFargateCpuArch` across Fargate containers for `X86_64` and `ARM64`

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1519

### Testing
CDK jest tests and manual AWS deployment

### Check List


deployment/cdk/opensearch-service-migration/default-values.json
```diff
@@ -6,6 +6,7 @@
   "nodeToNodeEncryptionEnabled": true,
   "encryptionAtRestEnabled": true,
   "vpcEnabled": true,
+  "defaultFargateCpuArch": "X86_64",
```
AndreKurait, line 9: This doesn't seem like a setting that should be able to be modified as it depends on the image that's built (unless we change our build script to build within a container and be able to build for whichever target irregardless of host system).


I'd prefer this get auto-resolved, possibly through an `auto` setting here then using `const hostArchitecture = process.arch;`

See example for output on M1:

```bash
➜  TrafficCapture git:(main) ✗ node
Welcome to Node.js v18.18.2.
Type ".help" for more information.
> process.arch
'arm64'
```

and for X86 EC2:
```
[akurait@ip-10-0-3-66 ~]$ node
Welcome to Node.js v18.18.2.
Type ".help" for more information.
> process.arch
'x64'
```

lewijacn: Good idea, have updated so as a default we will try to use `process.arch`

deployment/cdk/opensearch-service-migration/test/common-utilities.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    const detectedArch2 = validateFargateCpuArch(cpuArch2)
+    expect(detectedArch2).toEqual(CpuArchitecture.ARM64)
+
+    const cpuArch3 = "x86_64"
+    const detectedArch3 = validateFargateCpuArch(cpuArch3)
+    expect(detectedArch3).toEqual(CpuArchitecture.X86_64)
+
+    const cpuArch4 = "X86_64"
+    const detectedArch4 = validateFargateCpuArch(cpuArch4)
+    expect(detectedArch4).toEqual(CpuArchitecture.X86_64)
```
AndreKurait, line 19: Can we have one test that passes nothing and expects it to be something valid? I think that will currently fail on X86 as my test with node showed the response to process.arch is `x64`

lewijacn: Added test case 👍

AndreKurait (approved):

PR #526, opened 2024-02-25
Fix traffic limiting

gregschohn:
### Description

Divide traffic source accumulation into a producer and consumer.  

Previously, a single thread acquired TrafficStreams from the TrafficSource and then sent them through the CapturedTrafficToHttpTransactionAccumulator.  That accumulator itself calls a callback object when requests and responses have been rematerialized.  The callbacks that the TrafficReplayer supplied, however, also blocked when the number of requests was high and they needed to be throttled.  

Now, the TrafficStreamLimiter takes a callback task to execute instead of just providing blocking functions.  The new queueWork() function will never block.  Instead, it will move work onto a transfer queue.  That transferQueue is read by a new worker thread that is also owned by the TrafficStreamLimiter that will consume or throttle consumption in the order that items were received.  

We still throttle the number of requests, but we allow them to accumulate in memory without processing them or sending them to the target server, which is what the original design has always intended to do.  Now we're able to do it without impacting the traffic source objects, which can now expect to be running on a dedicated thread where only the traffic source itself is responsible for any blocking.

* Category - Bug fix
* Why these changes are required?  When there are more requests outstanding than --max-concurrent-requests (default of 1024) that causes the TrafficStreamLimiter to kick in and throttle AND some of those throttled messages took more than the kafka poll timeout to be released, the Kafka consumer will fall out of the group.
* What is the old behavior before changes and new behavior after changes?  Old behavior was that the replayer would fall out of the group, causing it to do and drop a significant amount of work.  New behavior will still buffer a lot of memory but should keep the kafka thread active to repeatedly call poll() to keep the consumer in the client group.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1515

### Testing
A new test was added specifically to mimic the up-front burst pattern. Some other tests failed because of subtle design changes that were then reverted. A number of other tests had failed due to concurrency and general flakiness. Those were fixed/improved, but given the novelty of concurrent test runs (from an earlier PR), there are likely other ones that still need to be cleaned up. 

### Check List


TrafficCapture/build.gradle
```diff
@@ -55,11 +55,11 @@ allprojects {
         if (project.hasProperty('excludeTests')) {
             exclude project.property('excludeTests')
         }
-        useJUnitPlatform {
-            systemProperty 'junit.jupiter.execution.parallel.enabled', 'true'
-            systemProperty 'junit.jupiter.execution.parallel.mode.default', "concurrent"
-            systemProperty 'junit.jupiter.execution.parallel.mode.classes.default', 'concurrent'
-        }
+//        useJUnitPlatform {
```
AndreKurait, line 58: instead of just commenting, can you set `junit.jupiter.execution.parallel.enabled = false` for individual projects with a comment

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/AccumulationCallbacks.java
```diff
@@ -18,11 +18,10 @@ void onTrafficStreamsExpired(RequestResponsePacketPair.ReconstructionStatus stat
                                  @NonNull IReplayContexts.IChannelKeyContext ctx,
                                  @NonNull List<ITrafficStreamKey> trafficStreamKeysBeingHeld);
 
-    void onConnectionClose(int channelInteractionNumber,
-                           @NonNull IReplayContexts.IChannelKeyContext ctx,
+    void onConnectionClose(int channelInteractionNum,
+                           IReplayContexts.@NonNull IChannelKeyContext ctx, int channelSessionNumber,
```
AndreKurait, line 22: Did you intend to keep this `ReplayContexts.@NonNull IChannelKeyContext ctx`?

gregschohn: That's a very strange way to annotate.  I've changed it to `@NonNull ReplayContexts.IChannelKeyContext`.

TrafficCapture/build.gradle
gregschohn: I've uncommented these. We'll see how they do for the next couple builds.

TrafficCapture/buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
```diff
@@ -64,7 +64,7 @@ class CommonUtils {
             // constructed in the configuration phase and the classpath won't be realized until the
             // execution phase.  Therefore, we need to have docker run the command to resolve the classpath
             // and it's simplest to pack that up into a helper script.
-            runCommand("printf \"#!/bin/sh\\njava -cp `echo /jars/*.jar | tr \\   :` \\\"\\\$@\\\" \" > /runJavaWithClasspath.sh");
+            runCommand("printf \"#!/bin/sh\\njava -XX:MaxRAMPercentage=80.0 -cp `echo /jars/*.jar | tr \\   :` \\\"\\\$@\\\" \" > /runJavaWithClasspath.sh");
```
AndreKurait, line 67: Thanks for this, should we document this somewhere?
Do we need to define ram limits in the docker solution to limit each container?

AndreKurait (approved):
I read through all the code in this PR. I don't understand it all completely, though I didn't see anything concerning, each file seems to be an improvement over the previous version.

We have an open question regarding if connection close behavior is impacted by this CR. Overall, it seems this CR is an improvement to replayer scaling, but i'd like to make sure we don't hamper the existing lower throughput use case with it. At a minimum, i'd like to see that as a fast follow.

I'm good if you force the TrafficReplayer slowTests to run in series to get them to pass for this PR

PR #550, opened 2024-04-04
Make RFS runnable as Container

lewijacn:
### Description
This changes introduces a Dockerfile for RFS that is linked to the gradle build process such that a user can make changes to RFS code or dependencies and simply execute `./gradlew buildDockerImages` from the RFS directory and create a Docker image with the latest changes. A user could alternatively execute `./gradlew composeUp` and create a docker environment with source/target/RFS containers using latest changes. **Note**: The gradle link is in place for docker compose but more work is needed here to limit manual steps a user needs to take to test RFS in this environment.

Along with this a few command line arguments were incorporated into RFS

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1601

### Testing
Local docker testing

### Check List


RFS/docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+#FROM amazoncorretto:11-alpine-jdk AS builder
+#COPY /gradle /rfs-build/gradle
+#COPY build.gradle /rfs-build
+#COPY gradlew /rfs-build
+#COPY settings.gradle /rfs-build
+#COPY gradle.properties /rfs-build
+#COPY /src /rfs-build/src
+#COPY /buildSrc /rfs-build/buildSrc
+#WORKDIR /rfs-build
+#RUN ./gradlew copyRuntimeJars
```
chelma, line 11: Did you intend to leave these lines in here?

RFS/docker/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+# Using same base image as other Java containers in this repo
+FROM openjdk:11-jre
+#COPY --from=builder /rfs-build/build/runtimeJars /rfs-app/jars
```
chelma, line 15: Same - leftover from something?

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
```diff
@@ -59,18 +61,27 @@ public static class Args {
         @Parameter(names = {"--movement-type"}, description = "What you want to move - everything, metadata, or data.  Default: 'everything'", required = false, converter = MovementType.ArgsConverter.class)
         public MovementType movementType = MovementType.EVERYTHING;
 
+        @Parameter(names = {"--index-template-whitelist"}, description = "List of index template names to migrate (e.g. 'posts_index_template1, posts_index_template2')", required = false)
+        public List<String> indexTemplateWhitelist;
+
+        @Parameter(names = {"--component-template-whitelist"}, description = "List of component template names to migrate (e.g. 'posts_template1, posts_template2')", required = false)
+        public List<String> componentTemplateWhitelist;
```
chelma, line 68: Hmmm, this is actually more potentially confusing than I thought originally.  Per our convo, before ES 7.8, there was only `templates`, so if you use the script to migrate a 6.8 cluster you wouldn't see an appropriate option.

Would you mind changing `--index-template-whitelist` to `--templates-whitelist` and updating the description to indicate it refers to legacy templates for ES 6.8 and index templates for 7.10?  Not a perfect fix long term, but probably enough for the moment.

TBH, this argument list is getting pretty long an gnarly.  Not sure what to do about it right now, but we should think about ways to make it easier to navigate the many different use-cases it's trying to address.

chelma (changes requested):
Can you please add a section to the README explaining how to use the new Docker functionality, hit the source/target clusters, and launch RFS?

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
chelma: Ugh, we'll also want a check so that `--component-templates` aren't supplied to a ES 6.8 migration.

I'm wondering if we want different wrapper scripts for different source versions, or something...

chelma: Created a follow-up JIRA for this: https://opensearch.atlassian.net/browse/MIGRATIONS-1646

RFS/docker/Dockerfile
lewijacn: Taking them out, were more notes for me

RFS/docker/Dockerfile
lewijacn: Removed

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
lewijacn: Updated as per comment for now. Need to think about this more but having a pattern for handling different arguments for different versions could be nice

lewijacn:
> Can you please add a section to the README explaining how to use the new Docker functionality, hit the source/target clusters, and launch RFS?

Have added, will be revisiting the docker compose part of this in my other task so it will be more solid after I do some proper testing there

RFS/README.md
```diff
@@ -1,5 +1,20 @@
...
+        - [Using a source cluster](#using-a-source-cluster)
+        - [Using Docker](#using-docker)
+        - [Handling auth](#handling-auth)
+    - [How to set up an ES 6.8 Source Cluster w/ an attached debugger](#how-to-set-up-an-es-68-source-cluster-w-an-attached-debugger)
+    - [How to set up an ES 7.10 Source Cluster running in Docker](#how-to-set-up-an-es-710-source-cluster-running-in-docker)
+        - [Providing AWS permissions for S3 snapshot creation](#providing-aws-permissions-for-s3-snapshot-creation)
+        - [Setting up the Cluster w/ some sample docs](#setting-up-the-cluster-w-some-sample-docs)
+    - [How to set up an OS 2.11 Target Cluster](#how-to-set-up-an-os-211-target-cluster)
+
+
```
chelma, line 17: Thank you!

chelma:
> will be revisiting the docker compose part of this in my other task so it will be more solid after I do some proper testing there

Got it, makes sense.  I'm hoping we'll be able to combine the special sauce I made for testing S3 snapshots into the docker compose setup, otherwise it probably won't be too useful 😛

chelma (approved):
LGTM, but we'll need to combine our Docker stuff soon.

PR #551, opened 2024-04-04
RFS can now take the source snapshot

chelma:
### Description
* Updated README with better instructions on how to set up and run, test
* Added the ability for the RFS script to take the source snapshot from an existing cluster
* Added a Dockerfile to use as a source cluster in local testing

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1621

### Testing
* Tested manually snapshotting w/ the new Dockerfile as a source against an AOS 2.11 Domain target in own AWS Account
* Also manually confirmed that using a local directory or existing snapshot in S3 still works

### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,31 @@ __pycache__
 *.egg-info*
 .python-version
 logs
-test/opensearch-cluster-cdk/
+.vscode/
+
+# Build files
+plugins/opensearch/loggable-transport-netty4/.gradle/
+
 RFS/.gradle/
+RFS/bin/
 RFS/build/
+
+TrafficCapture/captureKafkaOffloader/bin/
+TrafficCapture/captureOffloader/bin/
+TrafficCapture/captureProtobufs/bin/
+TrafficCapture/coreUtilities/bin/
+TrafficCapture/nettyWireLogging/bin/
+TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJMESPathMessageTransformer/bin/
+TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJMESPathMessageTransformerProvider/bin/
+TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJoltMessageTransformer/bin/
+TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJoltMessageTransformerProvider/bin/
+TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonMessageTransformerInterface/bin/
+TrafficCapture/replayerPlugins/jsonMessageTransformers/openSearch23PlusTargetTransformerProvider/bin/
+TrafficCapture/testUtilities/bin/
+TrafficCapture/trafficCaptureProxyServer/bin/
+TrafficCapture/trafficCaptureProxyServerTest/bin/
+TrafficCapture/trafficReplayer/bin/
+
+# CDK files from end-to-end testing
 opensearch-cluster-cdk/
+test/opensearch-cluster-cdk/
--- a/.gitignore
+++ b/.gitignore
@@ -16,21 +16,7 @@ RFS/.gradle/
 RFS/bin/
 RFS/build/
 
-TrafficCapture/captureKafkaOffloader/bin/
-TrafficCapture/captureOffloader/bin/
-TrafficCapture/captureProtobufs/bin/
-TrafficCapture/coreUtilities/bin/
-TrafficCapture/nettyWireLogging/bin/
-TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJMESPathMessageTransformer/bin/
-TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJMESPathMessageTransformerProvider/bin/
-TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJoltMessageTransformer/bin/
-TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonJoltMessageTransformerProvider/bin/
-TrafficCapture/replayerPlugins/jsonMessageTransformers/jsonMessageTransformerInterface/bin/
-TrafficCapture/replayerPlugins/jsonMessageTransformers/openSearch23PlusTargetTransformerProvider/bin/
-TrafficCapture/testUtilities/bin/
-TrafficCapture/trafficCaptureProxyServer/bin/
-TrafficCapture/trafficCaptureProxyServerTest/bin/
-TrafficCapture/trafficReplayer/bin/
+TrafficCapture/**/bin/
 
 # CDK files from end-to-end testing
 opensearch-cluster-cdk/
--- a/RFS/docker/TestSource_ES_7_10/Dockerfile
+++ b/RFS/docker/TestSource_ES_7_10/Dockerfile
@@ -1,0 +1,20 @@
+FROM docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2
+
+# Install the S3 Repo Plugin
+RUN echo y | /usr/share/elasticsearch/bin/elasticsearch-plugin install repository-s3
+
+# Install the AWS CLI for testing purposes
+RUN curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+    unzip awscliv2.zip && \
+    ./aws/install
+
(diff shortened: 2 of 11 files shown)


RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
```diff
@@ -103,19 +116,50 @@ public static void main(String[] args) {
             throw new IllegalArgumentException("Unsupported target version: " + sourceVersion);
         }
 
+        /*
+         * You have three options for providing the snapshot data
+         * 1. A local snapshot directory
+         * 2. A source host we'll take the snapshot from
```
lewijacn, line 122: Another use case to potentially add in future, seems like would be to allow taking a local snapshot and using

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
```diff
@@ -0,0 +1,152 @@
...
+
+        ObjectNode body = mapper.createObjectNode();
+        body.put("type", "s3");
+        body.set("settings", settings);
+
+        // Register the repo; it's fine if it already exists
+        RestClient client = new RestClient(connectionDetails);
+        String bodyString = body.toString();
+        RestClient.Response response = client.put(targetName, bodyString, false);
+        if (response.code == HttpURLConnection.HTTP_OK) {
```
lewijacn, line 78: Is this always OK and never CREATED?

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
```diff
@@ -0,0 +1,152 @@
...
+        if (rawBasePath.endsWith("/")) {
+            return rawBasePath.substring(0, rawBasePath.length() - 1);
+        } else {
+            return rawBasePath;
+        }
+    }
+
+    public void registerRepo() throws Exception {
+        // Assemble the REST path
+        String targetName = "_snapshot/s3_repo";
```
lewijacn, line 62: A more descriptive name would be nice, something like `s3_migration_snapshot_repo` and have it as a static field of the class

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
```diff
@@ -0,0 +1,152 @@
...
+        // Assemble the REST path
+        String targetName = "_snapshot/s3_repo/" + getSnapshotName();
+
+        // Assemble the request body
+        ObjectNode body = mapper.createObjectNode();
+        body.put("indices", "_all");
+        body.put("ignore_unavailable", true);
+        body.put("include_global_state", true);
+
+        // Register the repo; it's fine if it already exists
```
lewijacn, line 96: This comment seems like it should be removed

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
```diff
@@ -0,0 +1,152 @@
...
+        JsonNode firstSnapshot = responseJson.path("snapshots").get(0);
+        JsonNode stateNode = firstSnapshot.path("state");
+        String state = stateNode.asText();
+
+        if (state.equals("SUCCESS")) {
+            return true;
+        } else if (state.equals("IN_PROGRESS")) {
+            return false;
+        } else {
+            logger.error("Snapshot " + getSnapshotName() + " has failed");
```
lewijacn, line 129: We should log the failed state as well

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
```diff
@@ -0,0 +1,152 @@
...
+     */
+    public String getBucketName() {
+        return s3Uri.split("/")[2];
+    }
+
+    /*
+     * Extracts the base path from the S3 URI; could be nested arbitrarily deep
+     * s3://my-bucket-name/my-folder/my-nested-folder => my-folder/my-nested-folder
+     */
+    public String getBasePath() {
```
lewijacn, line 44: Nice to have some validation around the URI at some point, assuming if someone accidentally passed an http s3 link for the `s3Uri` this might produce a weird error

.gitignore
```diff
@@ -7,7 +7,31 @@ __pycache__
...
+.vscode/
+
+# Build files
+plugins/opensearch/loggable-transport-netty4/.gradle/
+
 RFS/.gradle/
+RFS/bin/
 RFS/build/
+
+TrafficCapture/captureKafkaOffloader/bin/
```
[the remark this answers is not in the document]
chelma: Good call, done.

RFS/docker/TestSource_ES_7_10/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2
```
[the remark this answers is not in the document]
chelma: Agreed!  I also want to poke around w/ the improved AWS Cred handling approach that Greg found.

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
chelma: Potentially; I thought about that as a option and wasn't able to think of a situation where I'd need it for testing but if you have something in mind I'm all ears.

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
chelma: +1.  There's a lot of stuff like that we'll need to tackle before handing the code out to users "for real".  I figured that would naturally follow from adding unit tests, but the code's shape has been changing so rapidly that I haven't thought writing those had made sense so far.  We should keep a close eye on when we think the shape is stabilizing.

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
chelma: Done!

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
chelma: I haven't encountered a 201, but I'll add that here too just in case.

chelma: Done

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
chelma: I thought it provided value, but I guess it wasn't clear enough.  Will re-write.

chelma: Done

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
chelma: Ah, good catch.  I was thinking the REST Client would log that, but that might not be the case in this situation.

chelma: Done!

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
lewijacn: My mind was around setting up a docker compose for local testing that had a low barrier for entry using a local snapshot. I'm worried having to setup an S3 bucket might turn some users away from initial trying the tool.

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
lewijacn: It seemed like it was saying this method would register the repo, but was pretty sure that happened in another method

lewijacn (approved):

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
chelma: Got it - yep, I could see that.

RFS/src/main/java/com/rfs/common/S3SnapshotCreator.java
chelma: Oh, whoops, sorry - you're completely correct.  I'll change that locally and it will flow out the next time I commit some code.

PR #552, opened 2024-04-04
Add info about interacting with the docker solution & configure grafana data sources automatically

mikaylathompson:
### Description
* Category:
	* Documentation & tiny feature
	* 	I'm walking through the setup process as a very rusty user. Following the link on the main README to the `dockerSolution` README, I find documentation on how to set up the docker solution, but nothing about what to do then. I had to puzzle out how to access the clusters, migration console, metrics, etc., but if this is a default user entrypoint to our tools, we should explain how to access them.
	* Additionally, while a grafana container was set up, it required the user to configure everythign from scratch. In the future, we'd like many things to be pre-configured so the user can just pull up grafana and see a dashboard immediately. The first step to get there is automatically configuring the datasources, which are very predictable within our docker setup, allowing the user to go straight to exploring metrics.

### Issues Resolved
n/a (as far as I know)

### Testing
n/a

### Check List


TrafficCapture/dockerSolution/README.md
```diff
@@ -14,6 +14,32 @@ While in the TrafficCapture directory, run the following command:
...
+A sample source cluster (running Elasticsearch 7.10) is set up with the Capture Proxy running in front of it. To route
+calls through the Capture Proxy, send requests to `localhost:9200` with the default credentials (admin, admin), e.g.:
+```sh
+curl --insecure https://localhost:9200/_cat/indices -u 'admin:admin'
+```
+
+You can send the same calls to the source cluster while bypassing the Capture Proxy (calls will not be relayed to the
+target cluster) via `localhost:19200`, and to the target cluster directly at `localhost:29200`.
+
+For sample data that exercises various endpoints with a range of datatypes, you can `ssh` into the Migration Console
```
gregschohn, line 28: ssh threw me here.  There's no ssh involved.  You could say 'exec', but that might not be clear to somebody new to Docker.  Maybe s/ssh into/execute a shell within/ would be better.
ssh kept making me think that you were trying to hit some kind of a deployed cloud resource.

TrafficCapture/dockerSolution/README.md
```diff
@@ -14,6 +14,32 @@ While in the TrafficCapture directory, run the following command:
...
+(`docker exec -i -t $(docker ps -aqf "ancestor=migrations/migration_console:latest") bash` or via the Docker console)
+and run `./runTestBenchmarks.sh`. By default, this runs four workloads from
+[Opensearch-Benchmark](https://github.com/opensearch-project/opensearch-benchmark) against the Capture Proxy endpoint
+(9200). The Migration Console contains other utility functions (`./catIndices.sh`, `kafka-tools`, etc.) to interact
+with the various containers of the solution.
+
+You can also access the metrics generated by the solution in Grafana. While the solution is running, go to
```
gregschohn, line 35: I would prefix this with "with the default docker-compose" configuration launched with :dockerSolution:composeUp, instrumentation containers will be started (see below for other options).

TrafficCapture/dockerSolution/README.md
```diff
@@ -14,6 +14,32 @@ While in the TrafficCapture directory, run the following command:
...
+[http://localhost:3000](http://localhost:3000/) and enter the default credentials (admin, admin). Connections to
+Jaeger and Prometheus are automatically provisioned (see them under `Connections->Data sources`), so you can go
+directly to `Explore` and define a query using the supplied data from either data source.
+
+Traces for the capture proxy and replayer are available via Jaeger at [http://localhost:16686](http://localhost:16686).
```
gregschohn, line 40: With the prefix above, I'd pull this sentence into the last paragraph to bind it to the clause.

TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml
```diff
@@ -35,6 +35,7 @@ services:
       - "3000:3000"
     volumes:
       - ./grafana_data:/var/lib/grafana
+      - ./grafana_datasources.yaml:/usr/share/grafana/conf/provisioning/datasources/datasources.yaml
```
gregschohn, line 38: thank you for figuring this out!
Should this be a mounted file or a configuration file that's within the image?  I can't think of how I would want to change this, and it seems like it might be useful for a quick-and-dirty deployment elsewhere w/out compose.

mikaylathompson: Yep, I think you're right. This was helpful for testing, but a user is quite unlikely to need that.

PR #553, opened 2024-04-05
Fix chunked headers

AndreKurait:
### Description
Currently when the replayer accumulates a traffic stream with the HTTP Request Header spanning multiple observations, the chunking logic causes all headers to be lost. This is occurring in the writeHeadersAsChunks method of NettyJsonToByteBufHandler because the netty CompositeByteBuf does not behave as the logic expected it to regarding writing to added component buffers.

This change moves to using a standard netty byte buff with the output stream for headers and slices afterwards to achieve the desired distribution.

This change also includes enhancements to HttpJsonTransformingConsumerTest to test more cases and permutations regarding chunking and transformations.

This change also includes a fix to header equivalency checks when header names are duplicated.

* Category: Bug Fix
* Why these changes are required? Handle larger headers that span multiple observations
* What is the old behavior before changes and new behavior after changes? Headers would be dropped yeilding Bad Request when parsing, now headers process normally and request is replayed as expected.

### Issues Resolved
[MIGRATIONS-1645](https://opensearch.atlassian.net/browse/MIGRATIONS-1645)

### Testing
Unit Testing

### Check List


TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumerTest.java
```diff
@@ -118,7 +132,7 @@ private void walkMaps(Object o) {
         try (var sampleStream = HttpJsonTransformingConsumer.class.getResourceAsStream(
                 "/requests/raw/post_formUrlEncoded_withFixedLength.txt")) {
             var allBytes = sampleStream.readAllBytes();
-            Arrays.fill(allBytes, allBytes.length-10, allBytes.length, (byte)' ');
+            Arrays.fill(allBytes, allBytes.length-30, allBytes.length, (byte)' ');
```
AndreKurait, line 135: This change was needed otherwise netty converted the request instead of parsing in the HttpServerCodec/HttpObjectDecoder. Created a separate item to evaluate behavior and correct if needed https://opensearch.atlassian.net/browse/MIGRATIONS-1647

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
```diff
@@ -29,6 +29,7 @@
 @Slf4j
 public class HttpByteBufFormatter {
 
+    private static final String LINE_DELIMITER = "\r\n";
```
gregschohn, line 32: I think that this should be parameterized to not give those reading logs the false sense that CRLFs may have been included on all requests and responses passed through this class.
I DO think that it's valuable to output these as CRLF, especially when that was the original convention and maybe for some downstream processing.  I'd prefer to see this a param that's by default set to use CRLF everywhere except for when we're just using this class this for log output.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestPreliminaryConvertHandler.java
```diff
@@ -136,9 +143,10 @@ private void handlePayloadNeutralTransformationOrThrow(ChannelHandlerContext ctx
     private boolean headerFieldsAreIdentical(HttpRequest request, HttpJsonMessageWithFaultingPayload httpJsonMessage) {
         if (!request.uri().equals(httpJsonMessage.path()) ||
                 !request.method().toString().equals(httpJsonMessage.method()) ||
-                request.headers().size() != httpJsonMessage.headers().strictHeadersMap.size()) {
+                request.headers().names().size() != httpJsonMessage.headers().strictHeadersMap.size()) {
             return false;
         }
+        // Depends on header size check above for correctness
```
gregschohn, line 149: thanks for the comment.  That's why the size check was important and not just an optimization :)

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonToByteBufHandler.java
```diff
@@ -149,28 +143,31 @@ private void writeHeadersIntoByteBufs(ChannelHandlerContext ctx,
...
                 writeHeadersIntoStream(httpJson, bbos);
             }
-            for (var bb : bufs) {
-                ctx.fireChannelRead(bb);
+
+            int index = 0;
+            var chunkSizeIterator = headerChunkSizes.iterator();
+            while (index < buf.writerIndex()) {
+                if (!chunkSizeIterator.hasNext()) {
+                    throw Lombok.sneakyThrow(new RuntimeException("Ran out of input chunks for mapping"));
```
gregschohn, line 160: If this should only happen when there was a programmer error, make the error message a bit more clear.  Use an IllegalStateException too - that will help indicate that it wasn't just a transient runtime problem.

gregschohn (commented):
Please change the runtime exception.  That's a linter violation.  
I'd also like to dig in more and get more context about the ByteBuf release

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestPreliminaryConvertHandler.java
```diff
@@ -67,8 +68,14 @@ public void channelRead(@NonNull ChannelHandlerContext ctx, @NonNull Object msg)
         } else {
             // ByteBufs shouldn't come through, but in case there's a regression in
             // RequestPipelineOrchestrator.removeThisAndPreviousHandlers to remove the handlers
-            // in order rather in reverse order
-            super.channelRead(ctx, msg);
+            // in order rather in reverse order this will release the byte buffs and complete the channel
+            if (msg instanceof ByteBuf) {
+                ((ByteBuf) msg).release();
+                super.channelReadComplete(ctx);
```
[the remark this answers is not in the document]
AndreKurait: Updated this logic to release pending messages when we tear down the embedded channel

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
AndreKurait: Updated to be a passed in parameter

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonToByteBufHandler.java
AndreKurait: Updated to be IllegalStateException

gregschohn (approved):
Thanks!  Looks good.

PR #557, opened 2024-04-06
Create new context trackers but don't observe them yet

gregschohn:
### Description
Add the notion of context tracking when a context is created to the time that it is closed as a first-class feature of the coreUtilities tracing library.
A number of changes were made to existing interfaces to reduce the scope of 'naked' Spans so that it's easier to audit if they're being used directly to close spans. The getCurrentSpan() needs to be public because it's in an interface and used internally, but it should really be private and require friend access only within its package. However, I'm not sure how to do better guardrails in Java.

* Category Diagnostic Enhancement
* Why these changes are required? To eventually help us find longstanding/still active (stuck) work in our systems
* What is the old behavior before changes and new behavior after changes? There will be a performance/memory impact from tracking traces until they're ended.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1618

### Testing
All tests pass

### Check List


TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/ActiveContextTracker.java
```diff
@@ -0,0 +1,34 @@
...
+        orderedScopes.add(scopedContext);
+    }
+
+    @Override
+    public void onContextClosed(IScopedInstrumentationAttributes scopedContext) {
+        orderedScopes.remove(scopedContext);
+    }
+
+    public Stream<IScopedInstrumentationAttributes> getOldestActiveScopes() {
+        return orderedScopes.stream();
```
AndreKurait, line 32: I get that the oldest here is the ordering, however based on the method name i was expecting some sort of filtering on the active scopes

AndreKurait (approved):
LGTM

TrafficCapture/coreUtilities/src/main/java/org/opensearch/migrations/tracing/ActiveContextTracker.java
gregschohn: Everything in this object will be active since the contexts are getting cleared out.  I don't mind the redundancy, but I'm also OK with 60 character names.  Let me know if you think it's clearer if it is shorter.

PR #575, opened 2024-04-10
Add RFS to CDK

lewijacn:
### Description
**Note**: This builds off changes being reviewed here: https://github.com/opensearch-project/opensearch-migrations/pull/566

This changes adds RFS as an ECS service that can be enabled in the migration CDK.

Also included are changes to make RFS usable in the E2E test script.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1604
https://opensearch.atlassian.net/browse/MIGRATIONS-1637

### Testing
CDK and E2E script deployment testing

### Check List


peternied:
> Given that this is an open-source monorepo for a growing number of projects, I have a strong preference to limit the tracked artifacts to just code/configs. This PR includes test data and that's a very slippery slope that we've so far avoided.
> 
> In this case, I think that its easy to workaround the issue with docker layers. Longer term, we should discuss how we'd like to share larger datasets or those that cannot be easily regenerated from code.

I share these concerns over time, I'd suggest looking at how benchmarks does this for its test workloads [1].  I think there is a sweet spot in the project cycle to locally optimize, but we might be exiting that phase with ever increasing sized test datasets.  

- [1] https://github.com/opensearch-project/opensearch-benchmark-workloads

gregschohn:
I haven't looked into how OSB handles workloads yet, but here's a Dockerfile that uses build phases to do a once, when used, dataset build.  It also self-documents how we've constructed the tarball since anybody can run it.  Note that I built on ARM, so I had to install development tools to build a dependency for OSB... all of that will be independent of our final ES image layer - it's only a one-time charge when building the image the first time.

```
FROM docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2 as base

ENV ELASTIC_SEARCH_CONFIG_FILE=/usr/share/elasticsearch/config/elasticsearch.yml

# without this line, elasticsearch will complain that there aren't enough nodes
RUN echo "discovery.type: single-node" >> $ELASTIC_SEARCH_CONFIG_FILE


FROM base as makeDataSet1

RUN cd /etc/yum.repos.d/ && \
    sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-* && \
    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
RUN yum install -y python3.9
RUN yum install -y gcc python39-devel
RUN yum install -y vim git less

RUN pip3 install opensearch-benchmark

RUN /usr/local/bin/docker-entrypoint.sh eswrapper & echo $! > /tmp/esWrapperProcess.pid && sleep 10 &&\
     opensearch-benchmark execute-test --distribution-version=1.0.0 --target-host=http://localhost:9200 \
    --workload=geonames --pipeline=benchmark-only --test-mode --kill-running-processes  \
    --workload-params "target_throughput:0.5,bulk_size:10,bulk_indexing_clients:1,search_clients:1" &&  \
    tar -cvzf esDataFiles.tgz /usr/share/elasticsearch/data


FROM base

RUN echo y | /usr/share/elasticsearch/bin/elasticsearch-plugin install https://maven.search-guard.com/search-guard-suite-release/com/floragunn/search-guard-suite-plugin/7.10.2-53.5.0/search-guard-suite-plugin-7.10.2-53.5.0.zip
# add search-guard, which provides TLS, as well as other features that we don't need to consider
RUN pushd /usr/share/elasticsearch/plugins/search-guard-7/tools ; chmod ugo+x ./install_demo_configuration.sh ; yes | ./install_demo_configuration.sh ; popd

ENV PROXY_TLS_CONFIG_FILE=/usr/share/elasticsearch/config/proxy_tls.yml
COPY disableTlsConfig.sh enableTlsConfig.sh /root/
RUN chmod ugo+x /root/disableTlsConfig.sh /root/enableTlsConfig.sh
ENV PATH=${PATH}:/usr/share/elasticsearch/jdk/bin/
RUN sed 's/searchguard/plugins.security/g' $ELASTIC_SEARCH_CONFIG_FILE |  \
    grep ssl.http > $PROXY_TLS_CONFIG_FILE

# The following two commands are more convenient for development purposes,
# but maybe not for a demo to show individual steps
RUN /root/enableTlsConfig.sh $ELASTIC_SEARCH_CONFIG_FILE
# Alter this config line to either enable(searchguard.disabled: false) or disable(searchguard.disabled: true) HTTP auth
RUN echo "searchguard.disabled: false" >> $ELASTIC_SEARCH_CONFIG_FILE

COPY --from=makeDataSet1 /usr/share/elasticsearch/esDataFiles.tgz /root/esDataFiles.tgz

CMD /usr/local/bin/docker-entrypoint.sh eswrapper
```

RFS/docker/TestSource_ES_7_10/test-resources/inventory.md
```diff
@@ -0,0 +1,45 @@
+## Test Resources
```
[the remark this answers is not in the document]
lewijacn: Thanks for the sample and discussion. Wasn't sure this would be possible within the stage of a Dockerfile but do prefer the reproducibility of this and eliminating this as a pattern that others might follow to store larger datasets. Have updated to use a multi-stage Docker approach instead now

RFS/README.md
```diff
@@ -78,9 +78,9 @@ RFS has support for packaging its java application as a Docker image by using th
 ```
 Also built into this Docker/Gradle support is the ability to spin up a testing RFS environment using Docker compose. This compose file can be seen [here](./docker/docker-compose.yml) and includes the RFS container, a source cluster container, and a target cluster container.
 
-This environment can be spun up with the Gradle command, and use the optional `-Pdataset` flag to preload a defined dataset from `docker/TestSource_ES_7_10/test-resources`
+This environment can be spun up with the Gradle command, and use the optional `-PshouldGenerateData` flag to preload a dataset from the `generateDatasetStage` in the multi-stage Docker [here](docker/TestSource_ES_7_10/Dockerfile). This stage will take a few minutes to run on its first attempt if it is generating data, as it will be making requests with OSB. This will be cached for future runs as long as the image remains the same.
 ```shell
-./gradlew composeUp -Pdataset='small-benchmark-single-node.tar.gz'
+./gradlew composeUp -PshouldGenerateData=true
```
gregschohn, line 83: generate data isn't quite right since the user may never see generation happen (if they're using a cached layer).
From a docker image perspective, one has OSB datasets from an OSB run and the other doesn't.  I would call this something like dataset=osb_4testWorkloads or something like that.

RFS/docker/TestSource_ES_7_10/generateDataset.sh
```diff
@@ -0,0 +1,27 @@
...
+generate_data_requests() {
+  endpoint="http://localhost:9200"
+  # If auth or SSL is used, the correlating OSB options should be provided in this array
+  options=()
+  client_options=$(IFS=,; echo "${options[*]}")
+  set -o xtrace
+
+  echo "Running opensearch-benchmark workloads against ${endpoint}"
+  echo "Running opensearch-benchmark w/ 'geonames' workload..." &&
+  opensearch-benchmark execute-test --distribution-version=1.0.0 --target-host=$endpoint --workload=geonames --pipeline=benchmark-only --test-mode --kill-running-processes --workload-params "target_throughput:0.5,bulk_size:10,bulk_indexing_clients:1,search_clients:1" --client-options=$client_options &&
```
gregschohn, line 12: I checked, but didn't find a way to JUST LOAD the data.  The opensearch-benchmark runs load data and then they do tests on it.  That latter part takes up most of the time & is work that we're not interested in for RFS.  We might really, really want to load the data, do an RFS migration, then run the rest of the test so that we could test CDC on a historically migrated cluster.

It might be a good idea to open an issue or submit a PR with a new option for OSB.

RFS/README.md
lewijacn: Have changed this to be similar to `dataset=osb_4testWorkloads`

RFS/docker/TestSource_ES_7_10/Dockerfile
```diff
@@ -1,10 +1,29 @@
...
+
+
+FROM base AS generateDatasetStage
+
+ARG SHOULD_GENERATE_DATA=false
+COPY generateDataset.sh /root
+RUN chmod ug+x /root/generateDataset.sh
+
+RUN /root/generateDataset.sh ${SHOULD_GENERATE_DATA} &&  \
+    cd /usr/share/elasticsearch/data && tar -cvzf esDataFiles.tgz nodes
```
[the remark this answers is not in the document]
lewijacn: I have removed the tarring step for now. Part of me would like to build more support here to make generating different datasets in parallel a bit easier, but I'm leaning more toward seeing the usefulness of this current piece and then reiterating if we find its something we'd really like. Testing multiple node clusters or even multiple RFS instances will look pretty different than what we have currently so I am also cautious that might shift direction.

RFS/docker/TestSource_ES_7_10/generateDataset.sh
lewijacn: We should probably raise an issue as I don't have a good understanding of how intertwined these two things are from looking at the actual workloads: https://github.com/opensearch-project/opensearch-benchmark-workloads/tree/main/geonames

RFS/docker/TestSource_ES_7_10/Dockerfile
```diff
@@ -1,4 +1,28 @@
...
+ENV PATH=${PATH}:/usr/share/elasticsearch/jdk/bin/
+
+RUN cd /etc/yum.repos.d/ && \
+    sed -i 's/mirrorlist/#mirrorlist/g' /etc/yum.repos.d/CentOS-* && \
+    sed -i 's|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g' /etc/yum.repos.d/CentOS-*
+RUN yum install -y gcc python3.9 python39-devel vim git less
+RUN pip3 install opensearch-benchmark
+
+
+FROM base AS generateDatasetStage
```
gregschohn, line 16: You need to rotate the `FROM base as generateDatasetStage` upward.  As it is, you'll install all of the yum packages for the final version, rather than ONLY propagate the ES data that you generated.  As this is now, there's no reason to do a multistage build because you'll pretty much have the same amount of stuff in your final image - and there's going to be a lot that you don't need!

gregschohn (commented):
Please tweak the beginning of your generateDataStage in the Dockerfile definition, it will create a tighter final image that has less baggage and surprises.

RFS/docker/TestSource_ES_7_10/Dockerfile
lewijacn: Since this was only a test ES source Docker image that we don't vend, its been helpful to have these packages in the final version for any adhoc testing when running. As I'm thinking about this more it may make sense to have the migration console be a part of the docker compose and let it handle things like this

gregschohn (approved):
I'd like to see a comment explaining why the base starts as late as it does.
Please add a line item to https://opensearch.atlassian.net/browse/MIGRATIONS-1628 to update this to 'un-migration-consolify' this image

lewijacn:
> I'd like to see a comment explaining why the base starts as late as it does. Please add a line item to https://opensearch.atlassian.net/browse/MIGRATIONS-1628 to update this to 'un-migration-consolify' this image

Sure have added a TODO on the Dockerfile and updated the Jira task

PR #580, opened 2024-04-15
Fixing some SonarLint findings.

gregschohn:
### Description
Fixing some SonarLint findings.

* Category Code Cleanup
* What is the old behavior before changes and new behavior after changes? There should be no behavioral changes

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1635

### Testing
Unit tests

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/RawPackets.java
```diff
@@ -21,4 +20,13 @@ public boolean equals(Object o) {
         }
         return true;
     }
+
+    @Override
+    public int hashCode() {
+        int result = 29;
```
AndreKurait, line 26: I don't love these magic numbers, would prefer some constants here

AndreKurait: Discussed offline, i'll leave it up to Greg on whether or not to use `Arrays.hashCode(toArray());`

AndreKurait (approved):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datatypes/RawPackets.java
gregschohn: I'm much more comfortable with 2 primes for a hash function than doing an copy.  
Honestly, I'd rather see this as a one-liner map-reduce operation like this, but most hashCode operations, especially those generated by IDEs and code generators (we have a lot of protobuf generated ones in out captureProtobufs package).
```return stream().mapToInt(Arrays::hashCode).reduce(29, (a,b) -> 31*a+b);```

PR #584, opened 2024-04-16
Timeouts for target responses

gregschohn:
### Description
There's now a command line option (default 30s) to timeout the response from the target server after read inactivity.  This should prevent the replayer from freezing up when a connection just goes completely dormant.  The AggregatedRawResponse should indicate that there's was a read timeout and the unit tests assert that.

This change also includes some refactoring to lessen the burden (only slightly) of needing to change some constructors.  I am still struggling to find some of the convenience features that I would have gotten in C++ (optional arguments, lightweight cast operators, etc).

* Category Enhancement, New feature, Bug fix, Refactoring
* Why these changes are required? The replayer could send requests that didn't get responses and that would eventually cause a full-stop since the original requests' records would never be committed.
* What is the old behavior before changes and new behavior after changes? Now non-responsive connections will error out and we'll move on.  Notice that we'll eventually want to put retries in even after a timeout, but for now, it will be a lost request (as other errant requests would be).

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1617

### Testing
See unit test changes (one test was already present from earlier commits)

### Check List
- [x] New functionality has been documented (through unit tests and command line args)


TrafficCapture/trafficReplayer/src/testFixtures/java/org/opensearch/migrations/replay/RootReplayerConstructorExtensions.java
```diff
@@ -0,0 +1,69 @@
...
+import org.opensearch.migrations.replay.tracing.IRootReplayerContext;
+import org.opensearch.migrations.replay.traffic.source.TrafficStreamLimiter;
+import org.opensearch.migrations.transform.IAuthTransformerFactory;
+import org.opensearch.migrations.transform.IJsonTransformer;
+
+import javax.net.ssl.SSLException;
+import java.net.URI;
+import java.time.Duration;
+
+public class RootReplayerConstructorExtensions extends TrafficReplayerTopLevel {
```
AndreKurait, line 12: Would the builder pattern help here simplify the constructor behavior with optional and default parameters?

AndreKurait (approved):

TrafficCapture/trafficReplayer/src/testFixtures/java/org/opensearch/migrations/replay/RootReplayerConstructorExtensions.java
gregschohn: I've tried the builder pattern a couple times.  One big issue is that it's a lot of code to maintain (even more than what's here).   Lombok won't work since I want to set defaults.  For now, I would prefer to have to think about and see all of the arguments rather than forget about one.  If this code gets to the point where every combination is in a constructor, then the builder pattern would obviously be better.  For now though, I don't like how with the builder pattern, I could, for example, forget to specify the URI.  Sure, that could be a runtime check, but this way can do compile time checks.

Of course, I could so something like a hybrid builder where the builder has a c'tor and then setters (or composed parts).  I don't know if anybody would like that and it seems burdensome.  I was comforted a bit when I had realized that this is only required because of testing.

PR #589, opened 2024-04-16
Fix ByteBuf Release in HttpByteBufFormatter when Exception

AndreKurait:
### Description
Previously, if the stream lambda threw an exception, it would be short circuited and ByteBufs may not get released. This fixes that by saving an exception until the end

* Category: Bug fix

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
@@ -11,6 +11,7 @@ import io.netty.handler.codec.http.HttpContentDecompressor;
 import io.netty.handler.codec.http.HttpMessage;
 import io.netty.handler.codec.http.HttpObjectAggregator;
 import io.netty.handler.codec.http.HttpServerCodec;
+import lombok.Lombok;
 import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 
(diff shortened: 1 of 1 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
```diff
@@ -161,9 +162,21 @@ public static HttpMessage parseHttpMessageFromBufs(HttpMessageType msgType, Stre
                 new HttpObjectAggregator(Utils.MAX_PAYLOAD_SIZE_TO_PRINT)  // Set max content length if needed
         );
 
+        final Throwable[] throwable = {null};
```
gregschohn, line 165: As a matter of style, I'd prefer to see an AtomicReference.  They're about equal, but the codebase typically uses AtomicReferences for the matter.  There might be a difference of nanoseconds, but I find the API for a reference is clearer for these purposes than an array (that could have multiple elements).

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
```diff
@@ -172,6 +185,9 @@ public static HttpMessage parseHttpMessageFromBufs(HttpMessageType msgType, Stre
         });
 
         try {
+            if (throwable[0] != null) {
+                throw Lombok.sneakyThrow(throwable[0]);
+            }
             return channel.readInbound();
```
gregschohn, line 191: I'm not convinced that the above is going to fix a leak.  Please connect with me off-line so that we can discuss.
I'm suspicious that the finally block for channel.finishAndReleaseAll() should be covering everything post the channel declaration.  I suspect that we have some stuff built within the channel, then throw before we go to check it out.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
```diff
@@ -161,9 +162,21 @@ public static HttpMessage parseHttpMessageFromBufs(HttpMessageType msgType, Stre
...
             try {
-                channel.writeInbound(b.retainedDuplicate());
+                if(throwable[0] == null) {
+                    channel.writeInbound(b.retainedDuplicate());
+                }
+            } catch (Throwable t) {
+                if (!releaseByteBufs) {
+                    throw t;
+                }
+                // Delay throw if releaseByteBufs
```
gregschohn, line 175: Is the stream processing retains in chunks or strictly on-demand?  I've always presumed that it was the latter.
Maybe converting to a Supplier could be safer - or taking the retain away from the stream generation.

PR #597, opened 2024-04-22
Updated RFS Docker Compose to handle AWS Creds; fix S3Repo bug

chelma:
### Description
* Updated the DockerCompose file for RFS to accept AWS Creds for the containers
* Updated RFS README to explain how to leverage the new behavior
* Fixed a bug in the S3Repo class' handling of S3 URI's (it was failing to handle passing in the bucket root path as an acceptable repo location)
* Unit tested the S3Repo class

### Issues Resolved
Done while working on https://opensearch.atlassian.net/browse/MIGRATIONS-1600

### Testing
* Ran the new unit tests (`./gradlew tests`)
* Ran the RFS process from snapshot through reindexing using the docker compose setup

### Check List

### DIFF
--- a/RFS/docker/docker-compose.yml
+++ b/RFS/docker/docker-compose.yml
@@ -8,6 +8,9 @@ services:
     environment:
       - discovery.type=single-node
       - path.repo=/snapshots
+      - AWS_ACCESS_KEY_ID=${access_key}
+      - AWS_SECRET_ACCESS_KEY=${secret_key}
+      - AWS_SESSION_TOKEN=${session_token}
     ports:
       - '19200:9200'
     volumes:
--- a/RFS/docker/docker-compose.yml
+++ b/RFS/docker/docker-compose.yml
@@ -26,6 +26,10 @@ services:
         condition: service_started
     networks:
       - migrations
+    environment:
+      - AWS_ACCESS_KEY_ID=${access_key}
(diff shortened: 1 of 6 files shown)


RFS/docker/docker-compose.yml
```diff
@@ -8,6 +8,9 @@ services:
     environment:
       - discovery.type=single-node
       - path.repo=/snapshots
+      - AWS_ACCESS_KEY_ID=${access_key}
```
mikaylathompson, line 11: is there a reason you're not using the typical env var naming conventions locally? (e.g. not `- AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}` and the corresponding change in the documentation). Not a big deal, but I'm not sure I understand the reason.

mikaylathompson (approved):

RFS/docker/docker-compose.yml
chelma: It's so that you can use a different set of credentials on the host machine if you want to.  I was having trouble with my host machine (mac laptop) picking up and using the creds intended for the RFS Containers, which was pretty annoying because that meant that when I ran the AWS CLI to get renewed creds for the Containers using STS, the CLI would use the expired credential set I was trying to renew! 🙃

PR #595, opened 2024-04-18
Add Documentation for proxy failure modes

AndreKurait:
### Description
Add Documentation for proxy failure modes
* Category: Documentation
* Why these changes are required? Give a starting point for specifying failure cases in the proxy to educate users around behaviors and configuration
* What is the old behavior before changes and new behavior after changes? No functional change

### Issues Resolved
[MIGRATIONS-1487](https://opensearch.atlassian.net/browse/MIGRATIONS-1487)

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
Viewed README in markdown viewer

### Check List
- [ x] New functionality includes testing
  - [ x] All tests pass, including unit test, integration test and doctest
- [x ] New functionality has been documented
- [x ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #602, opened 2024-04-23
Enhance logging with Async and high performance appenders

AndreKurait:
### Description

Move log4j2 appenders to high performance async loggers:
   * Existing logging was performed synchronous and blocking based on appender behavior. Moves this logic to be async with higher performance appender configuration.
   * Note: This will still eventually block on flushing out the appenders, but there will be a buffer between logging and io streams.
   * Existing Data Flow (All Blocking Calls):
            Application → Log4j2 Appender → Standard Out → ECSLogDriver → Cloudwatch Logs
   * New Data Flow (Blocking unless specified):
            Application → [AsyncLogger](https://github.com/apache/logging-log4j2/blob/f0bd87818d43efa8337f56526ced2020c308d9bf/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLogger.java#L66) (async)→ Log4j2 Appender → ECSLogDriver → Cloudwatch Logs

The AsyncLogger internally buffers so the rest can be "blocking" and still not block the application flow until it's internal queue is full

Enhance logging with microsecond clock and enhanced performance options
    * Added precision on logging and moved console appender to direct reference instead of through standard output stream.

Add multilinePattern to log driver

Fixed some logging unreliable unit tests with specifying a synchronous logger.

Documentation on Proxy Behavior added in [PR#595](https://github.com/opensearch-project/opensearch-migrations/pull/595)

* Category: Enhancement/Bug Fix
* What is the old behavior before changes and new behavior after changes? In AWS, logs were one line per line, now combined (e.g. stack trace together). Now microsecond instead of millisecond timing. Higher throughput replayer when debug/trace logging is enabled.

### Issues Resolved
[MIGRATIONS-1663](https://opensearch.atlassian.net/browse/MIGRATIONS-1663)

### Testing
Deployed to AWS and tested in high performance scenarios with high logging (Debug/Trace). Verified multiline behavior in AWS.

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/build.gradle
+++ b/TrafficCapture/trafficCaptureProxyServer/build.gradle
@@ -29,10 +29,12 @@ dependencies {
     implementation project(':coreUtilities')
 
     implementation group: 'io.netty', name: 'netty-all', version: '4.1.100.Final'
-    implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.20.0'
-    implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.20.0'
-    implementation group: 'org.apache.logging.log4j', name: 'log4j-slf4j2-impl', version: '2.20.0'
-    implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.7'
+
+    implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.23.1'
+    implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.23.1'
+    implementation group: 'org.apache.logging.log4j', name: 'log4j-slf4j2-impl', version: '2.23.1'
+    implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.13'
+    implementation group: 'com.lmax', name: 'disruptor', version: '4.0.0'
 
     implementation group: 'com.beust', name: 'jcommander', version: '1.82'
     implementation 'com.google.protobuf:protobuf-java:3.22.2'
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
@@ -8,6 +8,7 @@ appender.console.target = SYSTEM_ERR
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = [%-5level] %d{DEFAULT_MICROS}{UTC} [%t] %c{1} - %msg%equals{ ctx=%mdc}{ ctx=\{\}}{}%n
 appender.console.direct = true
+appender.console.includeLocation = true
 
 rootLogger.level = info
 rootLogger.appenderRefs = stderr
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
@@ -8,7 +8,6 @@ appender.console.target = SYSTEM_ERR
 appender.console.layout.type = PatternLayout
 appender.console.layout.pattern = [%-5level] %d{DEFAULT_MICROS}{UTC} [%t] %c{1} - %msg%equals{ ctx=%mdc}{ ctx=\{\}}{}%n
 appender.console.direct = true
-appender.console.includeLocation = true
 
 rootLogger.level = info
 rootLogger.appenderRefs = stderr
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
@@ -6,7 +6,8 @@ appender.console.type = Console
 appender.console.name = STDERR
 appender.console.target = SYSTEM_ERR
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = [%-5level] %d{yyyy-MM-dd HH:mm:ss,SSS}{UTC} [%t] %c{1} - %msg%equals{ ctx=%mdc}{ ctx=\{\}}{}%n
+appender.console.layout.pattern = [%-5level] %d{DEFAULT_MICROS}{UTC} [%t] %c{1} - %msg%equals{ ctx=%mdc}{ ctx=\{\}}{}%n
+appender.console.direct = true
 
 rootLogger.level = info
 rootLogger.appenderRefs = stderr
--- a/TrafficCapture/trafficReplayer/src/main/resources/log4j2.component.properties
+++ b/TrafficCapture/trafficReplayer/src/main/resources/log4j2.component.properties
@@ -1,0 +1,2 @@
+Log4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
+AsyncLogger.WaitStrategy=Yield
--- a/deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
@@ -136,7 +136,7 @@ export class MigrationServiceCore extends Stack {
                 // and  "[ERROR] 2024-12-31 23:59:59..."
                 // and  "2024-12-31 23:59:59..."
                 multilinePattern: "^(\\[[A-Z ]{1,5}\\] )?\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}",
-                // Log appenders are non-blocking so defer async behavior to them
+                // Defer buffering behavior to log4j2 for greater flexibility
                 mode: AwsLogDriverMode.BLOCKING,
             }),
             ulimits: props.ulimits
--- a/deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
@@ -12,7 +12,8 @@ import {
     ServiceConnectService,
     Ulimit,
     OperatingSystemFamily,
-    Volume
+    Volume,
+    AwsLogDriverMode
 } from "aws-cdk-lib/aws-ecs";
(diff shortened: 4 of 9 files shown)


deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -130,7 +131,13 @@ export class MigrationServiceCore extends Stack {
             portMappings: props.portMappings,
             logging: LogDrivers.awsLogs({
                 streamPrefix: `${props.serviceName}-logs`,
-                logGroup: serviceLogGroup
+                logGroup: serviceLogGroup,
+                // E.g. "[INFO ] 2024-12-31 23:59:59,99999999"
+                // and  "[ERROR] 2024-12-31 23:59:59,999"
+                // and  "2024-12-31 23:59:59"
+                multilinePattern: "^(\\[[A-Z ]{1,5}\\] )?\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}(,\\d{2,8})?",
+                // Log appenders are non-blocking so defer async behavior to them
```
peternied, line 139: This comments describes the opposite of what is on the AWS docs - https://docs.aws.amazon.com/AmazonECS/latest/developerguide/using_awslogs.html

Maybe remove the comment?

I guess the question is - is it OK to have the container be unable to continue executing while logs are committed to the cloudwatch log system?  It seems like the answer is YES - the logs are critical to the operation of the container and we'd rather no more work take place while the logs are committed to storage.

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -169,7 +176,9 @@ export class MigrationServiceCore extends Stack {
                     logGroup: serviceLogGroup
                 })
             },
-            cloudMapOptions: cloudMapOptions
+            cloudMapOptions: cloudMapOptions,
+            minHealthyPercent: 0,
```
peternied, line 180: 0 means that none of the containers are required to be up... that doesn't seem correct, shouldn't there be a min threadhold, like 50?  If not could you put a quick inline comment to indicate what the intention of this setting is?

TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
```diff
@@ -6,7 +6,8 @@ appender.console.type = Console
 appender.console.name = STDERR
 appender.console.target = SYSTEM_ERR
 appender.console.layout.type = PatternLayout
-appender.console.layout.pattern = [%-5level] %d{yyyy-MM-dd HH:mm:ss,SSS}{UTC} [%t] %c{1} - %msg%equals{ ctx=%mdc}{ ctx=\{\}}{}%n
+appender.console.layout.pattern = [%-5level] %d{DEFAULT_MICROS}{UTC} [%t] %c{1} - %msg%equals{ ctx=%mdc}{ ctx=\{\}}{}%n
+appender.console.direct = true
```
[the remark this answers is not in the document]
peternied: Not sure this topology is fully aligned, but the way the `serviceContainer` is setup is below to blocking which would cause tiny write to be committed immediately and could show up in strange slow down.

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -130,7 +131,13 @@ export class MigrationServiceCore extends Stack {
             portMappings: props.portMappings,
             logging: LogDrivers.awsLogs({
                 streamPrefix: `${props.serviceName}-logs`,
-                logGroup: serviceLogGroup
+                logGroup: serviceLogGroup,
+                // E.g. "[INFO ] 2024-12-31 23:59:59,99999999"
+                // and  "[ERROR] 2024-12-31 23:59:59,999"
+                // and  "2024-12-31 23:59:59"
+                multilinePattern: "^(\\[[A-Z ]{1,5}\\] )?\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}(,\\d{2,8})?",
```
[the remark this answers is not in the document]
AndreKurait: The intention here is to provide a flexible regex for how AWS identifies the start of a new log line based on the starting characters in a line, while trying to decrease the likelihood that some intermediate print statement would match and thus separate from it's associated line. The log4j2.properties defines the actual log pattern.

I'll remove the mili/micro second pattern from this as it's not helping filter any potential patterns since it's an optional match.

TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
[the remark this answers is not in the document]
AndreKurait: Existing Data Flow  (All Blocking Calls):
     Application → Log4j2 → Standard Out → ECSLogDriver → Cloudwatch Logs


New Data Flow (Blocking unless specified):
     Application → [AsyncLogger](https://github.com/apache/logging-log4j2/blob/f0bd87818d43efa8337f56526ced2020c308d9bf/log4j-core/src/main/java/org/apache/logging/log4j/core/async/AsyncLogger.java#L66) (async)→ Log4j2 → ECSLogDriver → Cloudwatch Logs

The AsyncLogger internally buffers so the rest can be "blocking" and still not block the application flow

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: see https://github.com/opensearch-project/opensearch-migrations/pull/602#discussion_r1576770648

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: Removed in new revision.

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: My mental model here is that spikes in logging latency should not result in spikes in application latency (definitely for proxy, maybe debatable for replay).

That contributed to the change for an async base logger. How we behave when that buffer fills up and isn't drained can be a different story

TrafficCapture/trafficReplayer/src/main/resources/log4j2.component.properties
```diff
@@ -0,0 +1,2 @@
+Log4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
+AsyncLogger.WaitStrategy=Yield
```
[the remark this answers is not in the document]
AndreKurait: See https://logging.apache.org/log4j/2.x/manual/configuration.html#asyncQueueFullPolicy

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
peternied: I still think the comment is incorrect, what do you think about removing it?

AndreKurait: What about "Defer buffering behavior to log4j2 for greater flexibility".

Blocking is the default behavior so I could remove that definition all together, but I was wanting to be explicit if we would rely on buffering/dropping from another part of the system

peternied (approved):

TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
```diff
@@ -6,7 +6,9 @@ appender.console.type = Console
...
+appender.console.includeLocation = true
```
mikaylathompson, line 11: Can you elaborate on including this/why it's important?

https://logging.apache.org/log4j/2.x/manual/async.html#location-location-location
> The [performance impact](https://logging.apache.org/log4j/2.x/performance.html#asyncLoggingWithLocation) of taking a stack trace snapshot is even higher for asynchronous loggers: logging with location is 30-100 times slower than without location.

mikaylathompson (approved):

TrafficCapture/trafficCaptureProxyServer/build.gradle
```diff
@@ -29,10 +29,12 @@ dependencies {
...
-    implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.20.0'
-    implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.20.0'
-    implementation group: 'org.apache.logging.log4j', name: 'log4j-slf4j2-impl', version: '2.20.0'
-    implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.7'
+
+    implementation group: 'org.apache.logging.log4j', name: 'log4j-api', version: '2.23.1'
+    implementation group: 'org.apache.logging.log4j', name: 'log4j-core', version: '2.23.1'
+    implementation group: 'org.apache.logging.log4j', name: 'log4j-slf4j2-impl', version: '2.23.1'
+    implementation group: 'org.slf4j', name: 'slf4j-api', version: '2.0.13'
+    implementation group: 'com.lmax', name: 'disruptor', version: '4.0.0'
```
mikaylathompson, line 37: Mostly a side note -- it's very weird that Mend decided it didn't need to run on this push ([see here](https://github.com/opensearch-project/opensearch-migrations/pull/602/checks?check_run_id=24174154631)). You're adding and updating dependencies!

TrafficCapture/trafficCaptureProxyServer/src/main/resources/log4j2.properties
AndreKurait: Removed, had originally included it due to the `%c{1}` in our layout pattern, though reading now that `%c` means logger name whereas `%C` means class which would have necessitated location

TrafficCapture/trafficCaptureProxyServer/build.gradle
mikaylathompson: Yeah, usually I can see the results when I modify `build.gradle` files before merging. But it's currently doing this same thing to me on the werkzeug PR. 

But definitely not a blocker for this!

PR #609, opened 2024-04-24
Change gradle copy to sync to clear old jars

mikaylathompson:
### Description
There is a step in our gradle build that copies jar files from the project (e.g. Replayer) build artifacts directory into a `dockerSolution` build directory, which is where the generated Dockerfile expects them to be to copy them into the image.

However, this `dockerSolution` build directory is only "emptied" on a `gradle clean`, which means that as versions of our dependencies are updated, that build directory is effectively append-only and continues collecting old versions of jars, which are then copied into our images.

This causes issues. In this specific case, conflicting versions of log4j2 were causing the replayer to crash on startup, when the image was being built in a non-clean environment.
We're also probably making our images larger than they need to be.

The same pattern was being used in TrafficCapture and RFS gradle build scripts, so both have been updated.

### Issues Resolved
n/a

### Testing
Local testing.
Along with functional testing, I did:
```shell
# after running with Copy
-> ls ~/code/opensearch-migrations/TrafficCapture/dockerSolution/build/docker/trafficReplayer/jars > copied_jars.txt
-> cat copied_jars.txt | grep 'log4j-'
log4j-api-2.21.1.jar
log4j-api-2.23.1.jar # note the conflicting versions here
log4j-core-2.21.1.jar
log4j-core-2.23.1.jar
log4j-slf4j2-impl-2.21.1.jar
log4j-slf4j2-impl-2.23.1.jar
# after running with Sync
-> ls ~/code/opensearch-migrations/TrafficCapture/dockerSolution/build/docker/trafficReplayer/jars > synced_jars.txt
# compare (output is truncated to the relevant changes in this case)
-> diff copied_jars.txt synced_jars.txt
74d59
< log4j-api-2.21.1.jar
76d60
< log4j-core-2.21.1.jar
78d61
< log4j-slf4j2-impl-2.21.1.jar
81d63
```

### Check List

### DIFF
--- a/TrafficCapture/buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
+++ b/TrafficCapture/buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
@@ -1,6 +1,6 @@
 package org.opensearch.migrations.common
 
-import org.gradle.api.tasks.Copy
+import org.gradle.api.tasks.Sync
 import org.gradle.api.Project
 import com.bmuschko.gradle.docker.tasks.image.Dockerfile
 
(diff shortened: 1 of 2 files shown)


AndreKurait (approved):

TrafficCapture/buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
```diff
@@ -27,7 +27,9 @@ class CommonUtils {
         copyArtifact(project, projectName, projectName, destDir)
     }
     static def copyArtifact(Project project, String artifactProjectName, String destProjectName, String destDir) {
```
gregschohn, line 29: Is it posible to rename this to syncArtifact?  It would be great to modernize that too!

gregschohn (approved):
This is awesome.  I had no idea that gradle could be smart like this!  I don't recall ever seeing a project with sync, but it's so much better than copy.  This should really reduce the need to do a clean!
Thanks!

PR #615, opened 2024-04-26
Favor ByteBuf Duplicates, NettyJsonContentAuthSigner and StreamChannelConnectionCaptureSerializer Refactoring

AndreKurait:
### Description
This change seeks to simplify the complexity ByteBuf reading and duplicating indexes through use of `.duplicate` over `.slice` and being more intentional about side effects to the buffer after reading.

This change also seeks to harden NettyJsonContentAuthSigner by flushing the buffer in error cases and removing unnecessary release/retains.

Furthermore, moved to the Netty Base64 Encoder in ParsedHttpMessagesAsDicts to directly decode from ByteBufs.

Also, fixed up some trace logging that previously wasn't using a supplier to reduce unnecessary cycles.

This change also refactors StreamChannelConnectionCaptureSerializer to clean up the interfaces, improve performance, and remove bugs around chunking and serializing data.

* Category: Refactoring
* Why these changes are required? Simplify logic and make it easier to understand
* What is the old behavior before changes and new behavior after changes? NettyJsonContentAuthSigner was less defensible around exceptions or handler being removed or channel closed scenarios and now will flush out buffer in those cases. ByteBufs are less likely to have their reader index modified unintentionally due to use of slicing.

### Issues Resolved
[MIGRATIONS-1696](https://opensearch.atlassian.net/browse/MIGRATIONS-1696)

### Testing
Unit testing

### Check List

### DIFF
--- a/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
+++ b/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
@@ -39,13 +39,33 @@ public class CodedOutputStreamSizeUtil {
         int tsTagAndContentSize = CodedOutputStream.computeInt32Size(TrafficObservation.TS_FIELD_NUMBER, tsContentSize) + tsContentSize;
 
         // Capture required bytes
-        int dataSize = CodedOutputStream.computeByteBufferSize(dataFieldNumber, buffer);
+        int dataSize = computeByteBufferRemainingSize(dataFieldNumber, buffer);
         int captureTagAndContentSize = CodedOutputStream.computeInt32Size(observationFieldNumber, dataSize) + dataSize;
 
         // Observation and closing index required bytes
         return bytesNeededForObservationAndClosingIndex(tsTagAndContentSize + captureTagAndContentSize,
                 Integer.MAX_VALUE);
     }
+
+    /**
+     * This function determines the number of bytes needed to write the remaining bytes in a byteBuffer and its tag.
+     * Use this over CodeOutputStream.computeByteBufferSize(int fieldNumber, ByteBuffer buffer) due to the latter
+     * relying on the ByteBuffer capacity instead of limit in size calculation.
+     */
+    public static int computeByteBufferRemainingSize(int fieldNumber, ByteBuffer buffer) {
+        return CodedOutputStream.computeTagSize(fieldNumber) + computeByteBufferRemainingSizeNoTag(buffer);
+    }
+
+    /**
+     * This function determines the number of bytes needed to write the remaining bytes in a byteBuffer. Use this over
+     * CodeOutputStream.computeByteBufferSizeNoTag(ByteBuffer buffer) due to the latter relying on the
+     * ByteBuffer capacity instead of limit in size calculation.
+     */
+    public static int computeByteBufferRemainingSizeNoTag(ByteBuffer buffer) {
+        int bufferSize = buffer.remaining();
+        return CodedOutputStream.computeUInt32SizeNoTag(bufferSize) + bufferSize;
+    }
+
 
     /**
      * This function determines the number of bytes needed to store a TrafficObservation and a closing index for a
--- a/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
+++ b/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
@@ -50,16 +50,16 @@ public class CodedOutputStreamSizeUtil {
     /**
      * This function determines the number of bytes needed to write the readable bytes in a byteBuf and its tag.
      */
-    public static int computeByteBufRemainingSize(int fieldNumber, ByteBuf buffer) {
-        return CodedOutputStream.computeTagSize(fieldNumber) + computeByteBufRemainingSizeNoTag(buffer);
+    public static int computeByteBufRemainingSize(int fieldNumber, ByteBuf buf) {
+        return CodedOutputStream.computeTagSize(fieldNumber) + computeByteBufRemainingSizeNoTag(buf);
     }
 
     /**
      * This function determines the number of bytes needed to write the readable bytes in a byteBuf.
      */
-    public static int computeByteBufRemainingSizeNoTag(ByteBuf buffer) {
-        int bufferSize = buffer.readableBytes();
-        return CodedOutputStream.computeUInt32SizeNoTag(bufferSize) + bufferSize;
+    public static int computeByteBufRemainingSizeNoTag(ByteBuf buf) {
+        int bufSize = buf.readableBytes();
+        return CodedOutputStream.computeUInt32SizeNoTag(bufSize) + bufSize;
     }
 
 
--- a/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
+++ b/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
@@ -2,10 +2,10 @@ package org.opensearch.migrations.trafficcapture;
 
 import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.Timestamp;
+import io.netty.buffer.ByteBuf;
 import org.opensearch.migrations.trafficcapture.protos.TrafficObservation;
 import org.opensearch.migrations.trafficcapture.protos.TrafficStream;
 
-import java.nio.ByteBuffer;
 import java.time.Instant;
 
 /**
@@ -28,18 +28,18 @@ public class CodedOutputStreamSizeUtil {
 
     /**
      * This function calculates the maximum bytes that would be needed to store a [Read/Write]SegmentObservation, if constructed
-     * from the given ByteBuffer and associated segment field numbers and values passed in. This estimate is essentially
-     * the max size needed in the CodedOutputStream to store the provided ByteBuffer data and its associated TrafficStream
+     * from the given ByteBuf and associated segment field numbers and values passed in. This estimate is essentially
+     * the max size needed in the CodedOutputStream to store the provided ByteBuf data and its associated TrafficStream
      * overhead. The actual required bytes could be marginally smaller.
      */
     public static int maxBytesNeededForASegmentedObservation(Instant timestamp, int observationFieldNumber,
-                                                             int dataFieldNumber, ByteBuffer buffer) {
+                                                             int dataFieldNumber, ByteBuf buf) {
         // Timestamp required bytes
         int tsContentSize = getSizeOfTimestamp(timestamp);
         int tsTagAndContentSize = CodedOutputStream.computeInt32Size(TrafficObservation.TS_FIELD_NUMBER, tsContentSize) + tsContentSize;
 
         // Capture required bytes
-        int dataSize = computeByteBufferRemainingSize(dataFieldNumber, buffer);
+        int dataSize = computeByteBufRemainingSize(dataFieldNumber, buf);
         int captureTagAndContentSize = CodedOutputStream.computeInt32Size(observationFieldNumber, dataSize) + dataSize;
 
         // Observation and closing index required bytes
@@ -48,21 +48,17 @@ public class CodedOutputStreamSizeUtil {
     }
 
     /**
-     * This function determines the number of bytes needed to write the remaining bytes in a byteBuffer and its tag.
-     * Use this over CodeOutputStream.computeByteBufferSize(int fieldNumber, ByteBuffer buffer) due to the latter
-     * relying on the ByteBuffer capacity instead of limit in size calculation.
+     * This function determines the number of bytes needed to write the readable bytes in a byteBuf and its tag.
      */
-    public static int computeByteBufferRemainingSize(int fieldNumber, ByteBuffer buffer) {
-        return CodedOutputStream.computeTagSize(fieldNumber) + computeByteBufferRemainingSizeNoTag(buffer);
+    public static int computeByteBufRemainingSize(int fieldNumber, ByteBuf buffer) {
+        return CodedOutputStream.computeTagSize(fieldNumber) + computeByteBufRemainingSizeNoTag(buffer);
     }
 
     /**
-     * This function determines the number of bytes needed to write the remaining bytes in a byteBuffer. Use this over
-     * CodeOutputStream.computeByteBufferSizeNoTag(ByteBuffer buffer) due to the latter relying on the
-     * ByteBuffer capacity instead of limit in size calculation.
+     * This function determines the number of bytes needed to write the readable bytes in a byteBuf.
      */
-    public static int computeByteBufferRemainingSizeNoTag(ByteBuffer buffer) {
-        int bufferSize = buffer.remaining();
+    public static int computeByteBufRemainingSizeNoTag(ByteBuf buffer) {
+        int bufferSize = buffer.readableBytes();
         return CodedOutputStream.computeUInt32SizeNoTag(bufferSize) + bufferSize;
     }
 
--- a/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
+++ b/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
@@ -3,6 +3,7 @@ package org.opensearch.migrations.trafficcapture;
 import com.google.protobuf.CodedOutputStream;
 import com.google.protobuf.Descriptors;
 import com.google.protobuf.Timestamp;
+import com.google.protobuf.WireFormat;
 import io.netty.buffer.ByteBuf;
 
 import java.util.function.IntSupplier;
@@ -182,8 +183,12 @@ public class StreamChannelConnectionCaptureSerializer<T> implements IChannelConn
     }
 
     private void writeByteBufferToCurrentStream(int fieldNum, ByteBuffer byteBuffer) throws IOException {
-        if (byteBuffer.remaining() > 0) {
-            getOrCreateCodedOutputStream().writeByteBuffer(fieldNum, byteBuffer);
+        if (byteBuffer.hasRemaining()) {
+            // CodedOutputStream.writeByteBuffer writes based on capacity and ignores limits so prefer write
+            getOrCreateCodedOutputStream().writeTag(fieldNum, WireFormat.WIRETYPE_LENGTH_DELIMITED);
+            getOrCreateCodedOutputStream().writeUInt32NoTag(byteBuffer.remaining());
+            getOrCreateCodedOutputStream().write(byteBuffer.duplicate());
+            assert byteBuffer.hasRemaining() : "byteBuffer position should not be modified when writing.";
         } else {
             getOrCreateCodedOutputStream().writeUInt32NoTag(0);
         }
@@ -279,8 +284,7 @@ public class StreamChannelConnectionCaptureSerializer<T> implements IChannelConn
         addDataMessage(captureFieldNumber, dataFieldNumber, timestamp, buf.nioBuffer());
     }
 
-    void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuffer nioBuffer) throws IOException {
-        var readOnlyDataBuffer = nioBuffer.asReadOnlyBuffer();
+    void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuffer buffer) throws IOException {
         int segmentFieldNumber;
         int segmentDataFieldNumber;
         if (captureFieldNumber == TrafficObservation.READ_FIELD_NUMBER) {
@@ -296,40 +300,44 @@ public class StreamChannelConnectionCaptureSerializer<T> implements IChannelConn
         // the potentially required bytes for simplicity. This could leave ~5 bytes of unused space in the CodedOutputStream
         // when considering the case of a message that does not need segments or for the case of a smaller segment created
         // from a much larger message
-        int messageAndOverheadBytesLeft = CodedOutputStreamSizeUtil.maxBytesNeededForASegmentedObservation(timestamp,
-            segmentFieldNumber, segmentDataFieldNumber, readOnlyDataBuffer);
-        int trafficStreamOverhead = messageAndOverheadBytesLeft - readOnlyDataBuffer.capacity();
+        final int messageAndOverheadBytesLeft = CodedOutputStreamSizeUtil.maxBytesNeededForASegmentedObservation(timestamp,
+            segmentFieldNumber, segmentDataFieldNumber, buffer);
+        final int dataSize = CodedOutputStreamSizeUtil.computeByteBufferRemainingSizeNoTag(buffer);
+        final int trafficStreamOverhead = messageAndOverheadBytesLeft - dataSize;
 
-        // Ensure that space for at least one data byte and overhead exists, otherwise a flush is necessary.
-        flushIfNeeded(() -> (trafficStreamOverhead + 1));
+        // Ensure that space for at least one data byte, one length byte, and overhead exists, otherwise a flush is necessary.
+        flushIfNeeded(() -> (trafficStreamOverhead + 2)).join();
+        assert getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft() == -1 ||
+               getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft() > trafficStreamOverhead
+                : "COS does not have space for data";
 
         // If our message is empty or can fit in the current CodedOutputStream no chunking is needed, and we can continue
         var spaceLeft = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
-        if (readOnlyDataBuffer.limit() == 0 || spaceLeft == -1 || messageAndOverheadBytesLeft <= spaceLeft) {
+        if (!buffer.hasRemaining() || spaceLeft == -1 || messageAndOverheadBytesLeft <= spaceLeft) {
             int minExpectedSpaceAfterObservation = spaceLeft - messageAndOverheadBytesLeft;
-            addSubstreamMessage(captureFieldNumber, dataFieldNumber, timestamp, readOnlyDataBuffer);
+            addSubstreamMessage(captureFieldNumber, dataFieldNumber, timestamp, buffer);
             observationSizeSanityCheck(minExpectedSpaceAfterObservation, captureFieldNumber);
             return;
         }
 
-        while(readOnlyDataBuffer.position() < readOnlyDataBuffer.limit()) {
+        var readBuffer = buffer.duplicate();
+        while(readBuffer.hasRemaining()) {
+            flushIfNeeded(() -> (trafficStreamOverhead + 2)).join();
             // COS checked for unbounded limit above
-            int availableCOSSpace = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
-            int chunkBytes = messageAndOverheadBytesLeft > availableCOSSpace ? availableCOSSpace - trafficStreamOverhead : readOnlyDataBuffer.limit() - readOnlyDataBuffer.position();
-            ByteBuffer bb = readOnlyDataBuffer.slice();
-            bb.limit(chunkBytes);
-            bb = bb.slice();
-            readOnlyDataBuffer.position(readOnlyDataBuffer.position() + chunkBytes);
-            addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, timestamp, bb);
-            int minExpectedSpaceAfterObservation = availableCOSSpace - chunkBytes - trafficStreamOverhead;
+            final int availableCOSSpace = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
+            final int maxLengthSpace = CodedOutputStream.computeUInt32SizeNoTag(readBuffer.remaining());
+            final int maxBytesSpace = availableCOSSpace - trafficStreamOverhead - maxLengthSpace;
+            final int nextChunkBytes = Math.min(maxBytesSpace, readBuffer.remaining());
+
+            var dataBytes = new byte[nextChunkBytes];
+            readBuffer.get(dataBytes, 0, nextChunkBytes);
+            addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, timestamp, ByteBuffer.wrap(dataBytes));
+
+            final int minExpectedSpaceAfterObservation = maxBytesSpace - nextChunkBytes;
             observationSizeSanityCheck(minExpectedSpaceAfterObservation, segmentFieldNumber);
-            // 1 to N-1 chunked messages
-            if (readOnlyDataBuffer.position() < readOnlyDataBuffer.limit()) {
-                flushCommitAndResetStream(false);
-                messageAndOverheadBytesLeft = messageAndOverheadBytesLeft - chunkBytes;
-            }
         }
         writeEndOfSegmentMessage(timestamp);
+
     }
 
     void addSubstreamMessage(int captureFieldNumber, int dataFieldNumber, int dataCountFieldNumber, int dataCount,
@@ -342,7 +350,7 @@ public class StreamChannelConnectionCaptureSerializer<T> implements IChannelConn
             segmentCountSize = CodedOutputStream.computeInt32Size(dataCountFieldNumber, dataCount);
         }
         if (byteBuffer.remaining() > 0) {
-            dataSize = CodedOutputStream.computeByteBufferSize(dataFieldNumber, byteBuffer);
+            dataSize = CodedOutputStreamSizeUtil.computeByteBufferRemainingSize(dataFieldNumber, byteBuffer);
             captureClosureLength = CodedOutputStream.computeInt32SizeNoTag(dataSize + segmentCountSize);
         }
         beginSubstreamObservation(timestamp, captureFieldNumber, captureClosureLength + dataSize + segmentCountSize);
--- a/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
+++ b/TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
@@ -6,7 +6,6 @@ import com.google.protobuf.Timestamp;
 import io.netty.buffer.ByteBuf;
 
 import java.util.function.IntSupplier;
-import java.util.function.Supplier;
 import lombok.NonNull;
 import lombok.extern.slf4j.Slf4j;
 import org.opensearch.migrations.trafficcapture.protos.CloseObservation;
@@ -276,8 +275,12 @@ public class StreamChannelConnectionCaptureSerializer<T> implements IChannelConn
         writeByteStringToCurrentStream(dataFieldNumber, str);
     }
 
-    private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuf buffer) throws IOException {
-        var byteBuffer = buffer.nioBuffer();
+    void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuf buf) throws IOException {
+        addDataMessage(captureFieldNumber, dataFieldNumber, timestamp, buf.nioBuffer());
+    }
+
+    void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuffer nioBuffer) throws IOException {
+        var readOnlyDataBuffer = nioBuffer.asReadOnlyBuffer();
         int segmentFieldNumber;
         int segmentDataFieldNumber;
         if (captureFieldNumber == TrafficObservation.READ_FIELD_NUMBER) {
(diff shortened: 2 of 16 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
```diff
@@ -100,18 +102,16 @@ public static void fillStatusCodeMetrics(@NonNull IReplayContexts.ITupleHandling
...
-        return bytes;
-    }
-
     private static Map<String, Object> fillMap(LinkedHashMap<String, Object> map,
                                                HttpHeaders headers, ByteBuf content) {
-        String base64body = Base64.getEncoder().encodeToString(getBytesFromByteBuf(content));
-        map.put("body", base64body);
-        headers.entries().stream().forEach(kvp -> map.put(kvp.getKey(), kvp.getValue()));
-        return map;
+        try (var bufHolder = RefSafeHolder.create(Base64.encode(content, false, Base64Dialect.STANDARD))) {
```
AndreKurait, line 107: This is equivalent to the default [java.util.Base64.getEncoder()](https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html) since Java8 which is the standard charset without line separators

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
```diff
@@ -90,14 +90,10 @@ private HttpRequestDecoder getHttpRequestDecoderHandler() {
...
+        chunks.add(nextRequestPacket.retainedSlice());
         chunkSizes.get(chunkSizes.size() - 1).add(nextRequestPacket.readableBytes());
-        if (log.isTraceEnabled()) {
-            byte[] copy = new byte[nextRequestPacket.readableBytes()];
-            nextRequestPacket.duplicate().readBytes(copy);
-            log.trace("HttpJsonTransformingConsumer[" + this + "]: writing into embedded channel: "
-                    + new String(copy, StandardCharsets.UTF_8));
-        }
+        log.atTrace().log(() -> "HttpJsonTransformingConsumer[" + this + "]: writing into embedded channel: "
+                   + nextRequestPacket.toString(StandardCharsets.UTF_8));
```
gregschohn, line 96: Ughh - I wish I would have known about this one beforehand.  I'm not sure why I never found it :(.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
```diff
@@ -12,35 +12,49 @@
 public class NettyJsonContentAuthSigner extends ChannelInboundHandlerAdapter {
     IAuthTransformer.StreamingFullMessageTransformer signer;
     HttpJsonMessageWithFaultingPayload httpMessage;
-    List<HttpContent> receivedHttpContents;
+    List<HttpContent> httpContentsBuffer;
```
gregschohn, line 15: why was this renamed to 'Buffer' if it still stores HttpContents?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
```diff
@@ -12,35 +12,49 @@
...
-        this.receivedHttpContents = new ArrayList<>();
+        this.httpContentsBuffer = new ArrayList<>();
     }
 
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         if (msg instanceof HttpJsonMessageWithFaultingPayload) {
             httpMessage = (HttpJsonMessageWithFaultingPayload) msg;
         } else if (msg instanceof HttpContent) {
-            receivedHttpContents.add(((HttpContent) msg).retainedDuplicate());
```
gregschohn, line 27: how can you guarantee that you can shed this retain?  My rule is that if a class is holding some refCnted resource for an asynchronous activity, it should do a retain.  It looks like the class didn't adequately release in error cases, but I'm not sure that shedding the retain is completely safe.

I guess the argument is that if we're not propagating it, then we're not releasing it, so as long as we hold back on firing it to the next handler, we're fine.  Please confirm.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
AndreKurait: This is what helped me understand what is happening in this handler.

HttpContents are read into this handler and buffered from flowing to the other handlers until they are all in and the headers are signed, then we send the updated headers and all the contents.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
AndreKurait: >if we're not propagating it, then we're not releasing it (yet)

Yes, this seemed to best align to me with the implicit contracts with handlers.

AndreKurait: I'm not counting this as an asynchronous activity since it happens in context with the embedded channel processing.

I didn't add this, but it may be helpful to include a `msg.touch()` here

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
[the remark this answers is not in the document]
AndreKurait: good catch, i'll name this encodedBuf

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
```diff
@@ -90,14 +90,10 @@ private HttpRequestDecoder getHttpRequestDecoderHandler() {
 
     @Override
     public DiagnosticTrackableCompletableFuture<String, Void> consumeBytes(ByteBuf nextRequestPacket) {
-        chunks.add(nextRequestPacket.duplicate().readerIndex(0).retain());
+        chunks.add(nextRequestPacket.retainedDuplicate());
```
AndreKurait, line 93: @gregschohn, please verify if this behavior change is correct. I'm not sure why we were resetting the readerIndex before

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -313,6 +313,7 @@ private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant
             // COS checked for unbounded limit above
             int availableCOSSpace = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
             int chunkBytes = messageAndOverheadBytesLeft > availableCOSSpace ? availableCOSSpace - trafficStreamOverhead : byteBuffer.limit() - byteBuffer.position();
+            // Need to use slice over duplicate as we depend on limits
```
AndreKurait, line 316: I couldn't change these to duplicate as it seems the ByteBuffer::limit() logic relies on marks which are not duplicated with duplicate.

gregschohn: I don’t fully understand this comment
I would rather the code read `slice().limit(chunkBytes).slice()`.
However, that feels like it could be accomplished other ways too - why do we need 2 slices?  Is this a spot where slice(rIdx, rIdx+chunkBytes) could make more sense?

AndreKurait: unfortunately, these are ByteBufers and not netty ByteBufs and so we don't have slice(rIdx..). I've since refactored this function to be more explicit about behavior and removing unnecessary slicing.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonTransformingConsumer.java
AndreKurait: That's the impression I got too, will keep this as retainedDuplicate()

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -276,8 +280,11 @@ private void addStringMessage(int captureFieldNumber, int dataFieldNumber,
         writeByteStringToCurrentStream(dataFieldNumber, str);
     }
 
-    private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuf buffer) throws IOException {
-        var byteBuffer = buffer.nioBuffer();
+    void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant timestamp, ByteBuf buf) throws IOException {
```
AndreKurait, line 283: Moved these to package-private to allow for unit testing with nio ByteBuffers to verify some underlying behavior that was easier to do than through Netty ByteBufs

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
gregschohn: How about something like deferredHttpContents or receivedHttpContentsToSendAfterComputation?  Yeah, those are long, but they could really clear up confusion.  Buffer has 2 alternate meanings here.  One is for a byte buffer and the other is something that you're holding only to batch some operations for efficiency.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
```diff
@@ -12,35 +12,48 @@
...
+            ctx.fireChannelRead(httpMessage);
+            httpMessage = null;
+        }
+        httpContentsBuffer.forEach(ctx::fireChannelRead);
+        httpContentsBuffer.clear();
+    }
+
+    @Override
+    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+        flushDownstream(ctx);
```
gregschohn, line 50: We log a warning message if we aren't able to supply a signature.  I realize that we'd need to do a bit more work to keep track of that, but it would be helpful to find out when it does happen.

gregschohn (commented):
I'd like to see the ByteBuffer conversion in the serializer be reversed.  I think that it will be a big performance hit and won't always work as you are hoping that it would (making it less maintainable).  The other comments should be minor/easy to address

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -293,44 +300,47 @@ private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant
...
-        int trafficStreamOverhead = messageAndOverheadBytesLeft - byteBuffer.capacity();
+        final int messageAndOverheadBytesLeft = CodedOutputStreamSizeUtil.maxBytesNeededForASegmentedObservation(timestamp,
+            segmentFieldNumber, segmentDataFieldNumber, buffer);
+        final int dataSize = CodedOutputStreamSizeUtil.computeByteBufferRemainingSizeNoTag(buffer);
+        final int trafficStreamOverhead = messageAndOverheadBytesLeft - dataSize;
 
-        // Ensure that space for at least one data byte and overhead exists, otherwise a flush is necessary.
-        flushIfNeeded(() -> (trafficStreamOverhead + 1));
+        // Ensure that space for at least one data byte, one length byte, and overhead exists, otherwise a flush is necessary.
+        flushIfNeeded(() -> (trafficStreamOverhead + 2)).join();
```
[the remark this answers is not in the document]
AndreKurait: Overhead is `maxBytesNeededForASegmentedObservation` minus `computeByteBufferRemainingSizeNoTag`. We want to make sure we have at least space for one packet data byte, which also means we need one length byte for that packet byte hence the + 2

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -293,44 +300,47 @@ private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant
...
-            ByteBuffer bb = byteBuffer.slice();
-            bb.limit(chunkBytes);
-            bb = bb.slice();
-            byteBuffer.position(byteBuffer.position() + chunkBytes);
-            addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, timestamp, bb);
-            int minExpectedSpaceAfterObservation = availableCOSSpace - chunkBytes - trafficStreamOverhead;
+            final int availableCOSSpace = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
+            final int maxLengthSpace = CodedOutputStream.computeUInt32SizeNoTag(readBuffer.remaining());
+            final int maxBytesSpace = availableCOSSpace - trafficStreamOverhead - maxLengthSpace;
+            final int nextChunkBytes = Math.min(maxBytesSpace, readBuffer.remaining());
```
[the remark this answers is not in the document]
AndreKurait: Updated

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
```diff
@@ -100,18 +102,16 @@ public static void fillStatusCodeMetrics(@NonNull IReplayContexts.ITupleHandling
...
-        return bytes;
-    }
-
     private static Map<String, Object> fillMap(LinkedHashMap<String, Object> map,
                                                HttpHeaders headers, ByteBuf content) {
-        String base64body = Base64.getEncoder().encodeToString(getBytesFromByteBuf(content));
-        map.put("body", base64body);
-        headers.entries().stream().forEach(kvp -> map.put(kvp.getKey(), kvp.getValue()));
-        return map;
+        try (var encodedBufHolder = RefSafeHolder.create(Base64.encode(content, false, Base64Dialect.STANDARD))) {
```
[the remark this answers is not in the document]
AndreKurait: Encode does not release the content, the caller of fillMap releases the FullHttpMessage it has

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
```diff
@@ -100,18 +102,16 @@ public static void fillStatusCodeMetrics(@NonNull IReplayContexts.ITupleHandling
...
+            var encodedBuf = encodedBufHolder.get();
+            assert encodedBuf != null : "Base64.encode should not return null";
+            String base64body = encodedBuf.toString(StandardCharsets.UTF_8);
+            map.put("body", base64body);
+            headers.entries().forEach(kvp -> map.put(kvp.getKey(), kvp.getValue()));
```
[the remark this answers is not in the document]
AndreKurait: Addressed

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -183,8 +183,12 @@ private void writeTimestampForNowToCurrentStream(Instant timestamp) throws IOExc
     }
 
     private void writeByteBufferToCurrentStream(int fieldNum, ByteBuffer byteBuffer) throws IOException {
-        if (byteBuffer.remaining() > 0) {
-            getOrCreateCodedOutputStream().writeByteBuffer(fieldNum, byteBuffer);
+        if (byteBuffer.hasRemaining()) {
+            // CodedOutputStream.writeByteBuffer writes based on capacity and ignores limits so prefer write
+            getOrCreateCodedOutputStream().writeTag(fieldNum, WireFormat.WIRETYPE_LENGTH_DELIMITED);
```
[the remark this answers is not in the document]
AndreKurait: The comment above
> // CodedOutputStream.writeByteBuffer writes based on capacity and ignores limits so prefer write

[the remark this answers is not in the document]
AndreKurait: The alternative is to use CodedOutputStream.writeByteBuffer(fieldNum, byteBuffer.slice()), though i was trying to avoid using slice to have greater clarity.

I'm also going to change this method to modify the position of the passed in byteBuffer so to achieve that we would have had to do

```
var slicedByteBuf = byteBuffer.slice();
getOrCreateCodedOutputStream().writeByteBuffer(fieldNum, slicedByteBuf);
byteBuffer.position(slicedByteBuf.position());
```

let me know if you would prefer that to
```
var codedOutputSream = getOrCreateCodedOutputStream();
codedOutputSream.writeTag(fieldNum, WireFormat.WIRETYPE_LENGTH_DELIMITED);
codedOutputSream.writeUInt32NoTag(byteBuffer.remaining());
codedOutputSream.write(byteBuffer);
```

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
AndreKurait: What i wanted to test specifically was behavior if ByteBuf.nioBuffer returns a buffer with Capacity != Limit.

I had difficulty testing this, because it only happens in very specific cases.. i had only theorized it until now where i've been able to produce the case:

```
        var nettyBuf = Unpooled.wrappedBuffer(byteBuffer.asReadOnlyBuffer());
        nettyBuf.writerIndex(nettyBuf.writerIndex() - 1);
        var nioBuffer = nettyBuf.nioBuffer();
        Assertions.assertTrue(nioBuffer.limit() < nioBuffer.capacity());
```

Would you prefer this?

AndreKurait: Fixed this with a further refactor that cleaned up and consolidated the interface between Netty ByteBufs and nio ByteBuffers

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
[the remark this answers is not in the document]
AndreKurait: Fixed this with a further refactor that cleaned up and consolidated the interface between Netty ByteBufs and nio ByteBuffers

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -293,44 +300,47 @@ private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant
...
-            addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, timestamp, bb);
-            int minExpectedSpaceAfterObservation = availableCOSSpace - chunkBytes - trafficStreamOverhead;
+            final int availableCOSSpace = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
+            final int maxLengthSpace = CodedOutputStream.computeUInt32SizeNoTag(readBuffer.remaining());
+            final int maxBytesSpace = availableCOSSpace - trafficStreamOverhead - maxLengthSpace;
+            final int nextChunkBytes = Math.min(maxBytesSpace, readBuffer.remaining());
+
+            var dataBytes = new byte[nextChunkBytes];
+            readBuffer.get(dataBytes, 0, nextChunkBytes);
+            addSubstreamMessage(segmentFieldNumber, segmentDataFieldNumber, timestamp, ByteBuffer.wrap(dataBytes));
```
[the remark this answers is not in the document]
AndreKurait: Fixed this with a further refactor that cleaned up and consolidated the interface between Netty ByteBufs and nio ByteBuffers

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -183,8 +183,12 @@ private void writeTimestampForNowToCurrentStream(Instant timestamp) throws IOExc
...
 
     private void writeByteBufferToCurrentStream(int fieldNum, ByteBuffer byteBuffer) throws IOException {
-        if (byteBuffer.remaining() > 0) {
-            getOrCreateCodedOutputStream().writeByteBuffer(fieldNum, byteBuffer);
+        if (byteBuffer.hasRemaining()) {
+            // CodedOutputStream.writeByteBuffer writes based on capacity and ignores limits so prefer write
+            getOrCreateCodedOutputStream().writeTag(fieldNum, WireFormat.WIRETYPE_LENGTH_DELIMITED);
+            getOrCreateCodedOutputStream().writeUInt32NoTag(byteBuffer.remaining());
+            getOrCreateCodedOutputStream().write(byteBuffer.duplicate());
+            assert byteBuffer.hasRemaining() : "byteBuffer position should not be modified when writing.";
```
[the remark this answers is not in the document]
AndreKurait: I refactored this and changed contracts around reporting bytes read, please check the latest commit

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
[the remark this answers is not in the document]
AndreKurait: I renamed to make this more intuitive

AndreKurait:
> I'd like to see the ByteBuffer conversion in the serializer be reversed. I think that it will be a big performance hit and won't always work as you are hoping that it would (making it less maintainable). The other comments should be minor/easy to address

I've updated the serializer to be more deterministic around edge cases and require less data copying. Please take a look at the latest revision

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -182,11 +191,44 @@ private void writeTimestampForNowToCurrentStream(Instant timestamp) throws IOExc
...
+        final int lengthFieldSpaceForInitialEstimatedAndOneMoreByte = CodedOutputStream.computeUInt32SizeNoTag(initialEstimatedMaxWriteSpace + 1);
+
+        int maxWriteBytesSpace = totalAvailableSpace - lengthFieldSpaceForInitialEstimatedAndOneMoreByte;
+
+        return Math.min(maxWriteBytesSpace, requestedWriteableSpace);
+    }
+
+    // Similar to calculateMaxWritableSpace but perform pessimistic calculation with fewer operations. In some cases returns
+    // up to 1 byte fewer than what could be written out of the available space.
+    public static int pessimisticallyCalculateMaxWritableSpace(int totalAvailableSpace, int requestedWriteableSpace) {
```
AndreKurait, line 211: @gregschohn, added, depended on, and tested pessimisticallyCalculateMaxWritableSpace instead of calculateMaxWritableSpace

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -293,54 +334,53 @@ private void addDataMessage(int captureFieldNumber, int dataFieldNumber, Instant
...
 
+        flushIfNeeded(maxBytesNeededForOneSegmentWithOneDataByteWithLengthByte).join();
+        var spaceLeft = currentOutputStreamWriteableSpaceLeft();
         // If our message is empty or can fit in the current CodedOutputStream no chunking is needed, and we can continue
-        var spaceLeft = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
-        if (byteBuffer.limit() == 0 || spaceLeft == -1 || messageAndOverheadBytesLeft <= spaceLeft) {
+        if (buf.readableBytes() == 0 || spaceLeft == -1 || messageAndOverheadBytesLeft <= spaceLeft) {
             int minExpectedSpaceAfterObservation = spaceLeft - messageAndOverheadBytesLeft;
-            addSubstreamMessage(captureFieldNumber, dataFieldNumber, timestamp, byteBuffer);
+            addSubstreamMessage(captureFieldNumber, dataFieldNumber, timestamp, buf.duplicate());
```
gregschohn, line 350: nit, you could duplicate before the branch since both branches need it

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -467,11 +507,49 @@ private void writeEndOfSegmentMessage(Instant timestamp) throws IOException {
...
+        }
+
+        @Override
+        public boolean isOpen() {
+            throw new UnsupportedOperationException();
+        }
+
+        @Override
+        public void close() {
+            throw new UnsupportedOperationException();
```
gregschohn, line 552: why would you penalize somebody for calling close()?  It's more future-proof to just leave this as a noop

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -467,11 +507,49 @@ private void writeEndOfSegmentMessage(Instant timestamp) throws IOException {
...
+            var bytesToWrite = src.remaining();
+            byteOutput.write(src);
+            return bytesToWrite - src.remaining();
+        }
+
+        @Override
+        public long write(ByteBuffer[] srcs, int offset, int length) throws IOException {
+            for (int i = offset; i < offset + length; i++) {
+                write(srcs[i]);
+            }
```
gregschohn, line 536: nit - this could be a one-liner with the help of Arrays

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -467,11 +507,49 @@ private void writeEndOfSegmentMessage(Instant timestamp) throws IOException {
...
+            return length;
```
gregschohn, line 537: This value is incorrect - you need to accumulate the sizes and return the bytes written...
```
Returns:
The number of bytes written, possibly zero
```

TrafficCapture/captureOffloader/src/test/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializerTest.java
```diff
@@ -370,18 +590,18 @@ protected CompletableFuture<Void> kickoffCloseStream(CodedOutputStreamHolder out
                     "Unknown outputStreamHolder sent back to StreamManager: " + outputStreamHolder);
             }
             var osh = (CodedOutputStreamAndByteBufferWrapper) outputStreamHolder;
-            log.trace("Getting ready to flush for " + osh);
-            log.trace("Bytes written so far... " + StandardCharsets.UTF_8.decode(osh.getByteBuffer().duplicate()));
+            log.atTrace().log(() -> "Getting ready to flush for " + osh);
```
gregschohn, line 593: oh darn - all of those setMessage() calls that I've been doing should probably be like this

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
```diff
@@ -4,43 +4,67 @@
...
+        httpContentsBuffer.forEach(ctx::fireChannelRead);
+        httpContentsBuffer.clear();
+        return messageFlushed;
+    }
+
+    @Override
+    public void handlerRemoved(ChannelHandlerContext ctx) throws Exception {
+        boolean messageFlushed = flushDownstream(ctx);
+        if (messageFlushed) {
+            log.atWarn().setMessage(() -> "Failed to sign message due to handler removed").log();
```
gregschohn, line 57: see the next comment to be a bit more clear about what is known at this line & be more clear that you're showing an opinion, since the code around this line will change eventually

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
```diff
@@ -4,43 +4,67 @@
...
+        }
+        super.handlerRemoved(ctx);
+    }
+
+    @Override
+    public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
+        boolean messageFlushed = flushDownstream(ctx);
+        if (messageFlushed) {
+            log.atWarn().setMessage(() -> "Failed to sign message due to channel unregistered").log();
```
gregschohn, line 66: nit - append "before the end of the http contents were received" or - "previously read messages were still in a working state when the channel was unregistered.  This message may not have been properly signed".  It probably wasn't, but assume that the rest of the code could change.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/ResultsToLogsConsumerTest.java
```diff
@@ -32,6 +33,10 @@
 @Slf4j
 @WrapWithNettyLeakDetection(repetitions = 4)
 class ResultsToLogsConsumerTest extends InstrumentationTest {
+    static {
+        // Synchronize logging to for assertions
+        LogManager.setFactory(new Log4jContextFactory(new ClassLoaderContextSelector()));
```
gregschohn, line 38: how does the code achieve the comment?

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/ResultsToLogsConsumerTest.java
```diff
@@ -153,11 +158,11 @@ public void testOutputterForGet() throws IOException {
                         "        \"Request-URI\": \"/test\",\r\n" +
                         "        \"Method\": \"GET\",\r\n" +
                         "        \"HTTP-Version\": \"HTTP/1.1\",\r\n" +
-                        "        \"body\": \"\",\r\n" +
                         "        \"Host\": \"foo.example\",\r\n" +
                         "        \"auTHorization\": \"Basic YWRtaW46YWRtaW4=\",\r\n" +
                         "        \"Content-Type\": \"application/json\",\r\n" +
-                        "        \"content-length\": \"0\"\r\n" +
+                        "        \"content-length\": \"0\",\r\n" +
+                        "        \"body\": \"\"\r\n" +
```
gregschohn, line 165: Wow - this could be so much more helpful!

TrafficCapture/trafficReplayer/src/testFixtures/java/org/opensearch/migrations/replay/TestUtils.java
```diff
@@ -118,7 +118,7 @@ protected void channelRead0(ChannelHandlerContext ctx, FullHttpRequest msg) thro
     private static String getStringFromContent(FullHttpRequest fullRequest) throws IOException {
         try (var baos = new ByteArrayOutputStream()) {
             var bb = fullRequest.content();
-            bb.readBytes(baos, bb.readableBytes());
+            bb.duplicate().readBytes(baos, bb.readableBytes());
             return baos.toString(StandardCharsets.UTF_8);
```
gregschohn, line 122: why not just call .toString(StandardCharsets.UTF_8)?

gregschohn (commented):
This is still a really hard PR to go through because the CodedOutputStream stuff has always been so challenging.  I think that 80% of this is great.  We just need to cleanup some names to get the rest of it to the point where it can be maintained.
Thanks for persevering.

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -163,7 +172,7 @@ private void beginSubstreamObservation(Instant timestamp, int captureTagFieldNum
         final var captureTagNoLengthSize = CodedOutputStream.computeTagSize(captureTagFieldNumber);
         final var observationContentSize = tsTagSize + tsContentSize + captureTagNoLengthSize + captureTagLengthAndContentSize;
         // Ensure space is available before starting an observation
-        flushIfNeeded(() -> CodedOutputStreamSizeUtil.bytesNeededForObservationAndClosingIndex(observationContentSize, numFlushesSoFar + 1));
+        flushIfNeeded(CodedOutputStreamSizeUtil.bytesNeededForObservationAndClosingIndex(observationContentSize, numFlushesSoFar + 1)).join();
```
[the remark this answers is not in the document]
AndreKurait: Thanks! That's the behavior i was missing to understand. I don't think we should chain them explicitly, though i'm not exactly sure how to make more explicit the behavior each of these functions performs synchronously vs asynchronously, perhaps a javadoc comment would go a long way here

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
AndreKurait: Thanks!

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/ResultsToLogsConsumerTest.java
AndreKurait: The ClassLoaderContextSelector forces a unified synchronous logger

TrafficCapture/trafficReplayer/src/testFixtures/java/org/opensearch/migrations/replay/TestUtils.java
AndreKurait: Good catch. Thanks!

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
AndreKurait: I'm thinking `computeMaxLengthDelimitedFieldSizeForSpace` which is analogous to (internal method) `computeLengthDelimitedFieldSize` in CodedOutputStream

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/CodedOutputStreamSizeUtil.java
```diff
@@ -28,25 +28,41 @@ public static int getSizeOfTimestamp(Instant t) {
...
 
         // Observation and closing index required bytes
         return bytesNeededForObservationAndClosingIndex(tsTagAndContentSize + captureTagAndContentSize,
                 Integer.MAX_VALUE);
     }
 
+    /**
+     * This function determines the number of bytes needed to write the readable bytes in a byteBuf and its tag.
+     */
+    public static int computeByteBufRemainingSize(int fieldNumber, ByteBuf buffer) {
```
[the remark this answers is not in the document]
AndreKurait: done

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -133,9 +134,17 @@ private CodedOutputStreamHolder getOrCreateCodedOutputStreamHolder() throws IOEx
         }
     }
 
-    public CompletableFuture<T> flushIfNeeded(IntSupplier requiredSize) throws IOException {
-        var spaceLeft = getOrCreateCodedOutputStreamHolder().getOutputStreamSpaceLeft();
-        if (spaceLeft != -1 && spaceLeft < requiredSize.getAsInt()) {
+    public int currentOutputStreamWriteableSpaceLeft() throws IOException {
+        // Writeable bytes is the space left minus the space needed to complete the next flush
+        var maxFieldNum = Math.max(TrafficStream.NUMBEROFTHISLASTCHUNK_FIELD_NUMBER, TrafficStream.NUMBER_FIELD_NUMBER);
+        var spaceNeededForNextFlush = CodedOutputStream.computeInt32Size(maxFieldNum, numFlushesSoFar + 1);
```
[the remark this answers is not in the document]
AndreKurait: Updated

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -182,11 +191,44 @@ private void writeTimestampForNowToCurrentStream(Instant timestamp) throws IOExc
         }
     }
 
-    private void writeByteBufferToCurrentStream(int fieldNum, ByteBuffer byteBuffer) throws IOException {
-        if (byteBuffer.remaining() > 0) {
-            getOrCreateCodedOutputStream().writeByteBuffer(fieldNum, byteBuffer);
+    // Optimistically calculates maxWriteableSpace taking into account lengthSpace. In some cases, this may
```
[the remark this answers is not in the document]
AndreKurait: Updated, kept this definition in the tests

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
AndreKurait: Updated

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonContentAuthSigner.java
AndreKurait: Updated

TrafficCapture/captureOffloader/src/main/java/org/opensearch/migrations/trafficcapture/StreamChannelConnectionCaptureSerializer.java
```diff
@@ -182,11 +191,44 @@ private void writeTimestampForNowToCurrentStream(Instant timestamp) throws IOExc
...
+    // up to 1 byte fewer than what could be written out of the available space.
+    public static int pessimisticallyCalculateMaxWritableSpace(int totalAvailableSpace, int requestedWriteableSpace) {
+        final int pessimisticLengthFieldSpace = CodedOutputStream.computeUInt32SizeNoTag(totalAvailableSpace);
+        int maxWriteBytesSpace = totalAvailableSpace - pessimisticLengthFieldSpace;
+        return Math.min(maxWriteBytesSpace, requestedWriteableSpace);
+    }
+
+    private void readByteBufIntoCurrentStream(int fieldNum, ByteBuf buf) throws IOException {
+        var codedOutputStream = getOrCreateCodedOutputStream();
+        if (buf.readableBytes() > 0) {
```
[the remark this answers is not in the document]
AndreKurait: Updated

gregschohn (approved):

PR #616, opened 2024-04-26
Fix missing gatheringRequest trace for non-blocking requests

mikaylathompson:
### Description

The `gatheringRequest` span was missing from `GET` requests, but present for mutating requests. There was an unnecessary overridden method specifically for non-blocking requests, which wasn't closing that request.

The override is now removed, and tests are added to demonstrate the fixed behavior. The testing in this section was refactored a bit to prepare for additional trace-related testing. 

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1658

### Testing
Unit tests & manual testing

### Check List

### DIFF
--- a/TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
+++ b/TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
@@ -1,0 +1,103 @@
+package org.opensearch.migrations.trafficcapture.netty;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.embedded.EmbeddedChannel;
+import io.opentelemetry.sdk.trace.data.SpanData;
+import lombok.Getter;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.opensearch.migrations.testutils.TestUtilities;
+import org.opensearch.migrations.tracing.IContextTracker;
+import org.opensearch.migrations.tracing.InMemoryInstrumentationBundle;
+import org.opensearch.migrations.trafficcapture.StreamChannelConnectionCaptureSerializer;
+import org.opensearch.migrations.trafficcapture.netty.tracing.RootWireLoggingContext;
+import org.opensearch.migrations.trafficcapture.protos.TrafficObservation;
+import org.opensearch.migrations.trafficcapture.protos.TrafficStream;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.SequenceInputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.Collections;
+import java.util.List;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+import static org.opensearch.migrations.trafficcapture.netty.TestStreamManager.consumeIntoArray;
+
+@Slf4j
+public class RootWireLoggingContextTest {
+
+  private static void writeMessageAndVerifyTraces(byte[] fullTrafficBytes, Consumer<EmbeddedChannel> channelWriter, boolean shouldBlock,
+                                                  List<String> expectedTraces) throws IOException {
+    try (var rootContext = new TestRootContext(true, true)) {
+      var streamManager = new TestStreamManager();
+      var offloader = new StreamChannelConnectionCaptureSerializer("Test", "c", streamManager);
+
+      EmbeddedChannel channel = new EmbeddedChannel(
+          new ConditionallyReliableLoggingHttpHandler(rootContext,
+              "n", "c", ctx -> offloader,
+              new RequestCapturePredicate(), x -> shouldBlock));
+      channelWriter.accept(channel);
+
+      // we wrote the correct data to the downstream handler/channel
+      var outputData = new SequenceInputStream(Collections.enumeration(channel.inboundMessages().stream()
+          .map(m -> new ByteArrayInputStream(consumeIntoArray((ByteBuf) m)))
+          .collect(Collectors.toList())))
+          .readAllBytes();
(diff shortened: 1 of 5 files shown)


TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
```diff
@@ -0,0 +1,103 @@
...
+      Assertions.assertEquals(1, streamManager.flushCount.get());
+
+      List<SpanData> finishedSpans = rootContext.instrumentationBundle.getFinishedSpans();
+      Assertions.assertTrue(!finishedSpans.isEmpty());
+      Assertions.assertTrue(!rootContext.instrumentationBundle.getFinishedMetrics().isEmpty());
+
+      List<String> finishedSpanNames = finishedSpans.stream().map(x -> x.getName()).collect(Collectors.toList());
+      expectedTraces.forEach(trace ->
+          Assertions.assertTrue(finishedSpanNames.contains(trace), "finishedSpans does not contain " + trace)
+      );
```
gregschohn, line 77: You could collect finishedSpanNames into a set too (& 'var' instead of 'List<String>' would let you swap back and forth.
An easier explanation of when things go wrong for this test could be to make a big string with the expected span names sorted & to sort the finishedSpans in the same way.  Something like
```
testSpanExporter.getFinishedSpanItems().stream().map(SpanData::getName).sorted().collect(Collectors.joining("\n"))
```
Now when those two lists don't match up, the user will understand how extensive the mismatch was.  Also, the cognitive load to assert that the test is complete is lower.

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
```diff
@@ -0,0 +1,103 @@
...
+      if (shouldBlock) {
+        Assertions.assertTrue(finishedSpanNames.contains("blocked"), "finishedSpans does not contain 'blocked' on a blocking request");
+      }
```
gregschohn, line 80: Is this block redundant?  It feels like it should be, even if expectedTraces didn't include this name, it could/should be added.

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+  }
+
+
+  @Test
+  public void testThatAGetProducesGatheringRequestTrace()
+      throws IOException {
```
gregschohn, line 88: style nit: these could be on the same line

gregschohn (commented):
Please remove the static import.  The other improvements would be nice, but otherwise, this LGTM.  Thanks!

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+      EmbeddedChannel channel = new EmbeddedChannel(
+          new ConditionallyReliableLoggingHttpHandler(rootContext,
+              "n", "c", ctx -> offloader,
+              new RequestCapturePredicate(), x -> shouldBlock));
+      channelWriter.accept(channel);
+
+      // we wrote the correct data to the downstream handler/channel
+      var outputData = new SequenceInputStream(Collections.enumeration(channel.inboundMessages().stream()
+          .map(m -> new ByteArrayInputStream(consumeIntoArray((ByteBuf) m)))
```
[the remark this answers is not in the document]
mikaylathompson: Oh, nice, this is simpler. I'm going to replace it in `ConditionallyReliableLoggingHttpHandlerTest` too, which means we get to get rid of `consumeIntoArray`

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
mikaylathompson: ah, okay, I like this. I was currently considering `expectedTraces` to pass if it was a subset of the actual traces, which is what led to the `blocked` note below. BUt it's probably clearer to just make the tests specify the entire list of expected traces.

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/RootWireLoggingContextTest.java
```diff
@@ -0,0 +1,103 @@
...
+  private static void writeMessageAndVerifyTraces(byte[] fullTrafficBytes, Consumer<EmbeddedChannel> channelWriter, boolean shouldBlock,
+                                                  List<String> expectedTraces) throws IOException {
+    try (var rootContext = new TestRootContext(true, true)) {
+      var streamManager = new TestStreamManager();
+      var offloader = new StreamChannelConnectionCaptureSerializer("Test", "c", streamManager);
+
+      EmbeddedChannel channel = new EmbeddedChannel(
+          new ConditionallyReliableLoggingHttpHandler(rootContext,
+              "n", "c", ctx -> offloader,
+              new RequestCapturePredicate(), x -> shouldBlock));
```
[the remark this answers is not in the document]
mikaylathompson: oof, I like this idea and went down this path, but it requires adding a dependency on CaptureProxy to this nettyWireLogging library, which is the opposite direction we want to go. I think it potentially makes sense to move that method somewhere more accessible, but is it alright if I push dealing with it out of this PR?

gregschohn (approved):
These look good.  Thanks!

PR #617, opened 2024-04-26
Use public.ecr.aws images

AndreKurait:
### Description
Currently we are getting throttling from docker and ecr within github actions. To resolve this, we should 

### TODO: before this change is merged, we should include adding credentials for authenticated ecr pulling to extend our limits.

* Category: Bug Fix
* Why these changes are required? Fix throttling for github actions
* What is the old behavior before changes and new behavior after changes? Some test runs failed and need to be retried.

### Issues Resolved

### Testing
Tested locally

### Check List
  - [x ] All tests pass, including unit test, integration test and doctest

### DIFF
--- a/FetchMigration/Dockerfile
+++ b/FetchMigration/Dockerfile
@@ -1,4 +1,4 @@
-FROM opensearchproject/data-prepper:2.5.0
+FROM public.ecr.aws/opensearchproject/data-prepper:2.5.0
 COPY python/requirements.txt .
 
 # Install dependencies to local user directory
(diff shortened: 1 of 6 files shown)


peternied:
@AndreKurait Is this still needed? Seems like the recent change [1] by @mikaylathompson will addresses the same root problem

- [1] #698

sumobrian:
Closing. We can reopen if needed.

gregschohn:
If the anonymous pull rates are lower for ECR than w/ docker, this solves the problem for us, but not any other users that are trying to build these images in a continuous environment.  I'd love to be able to keep using the same identifiers for the images that we want (like grafana) and no be stuck with baking the image repo that works best for any given user into the code.  If there was a [mirror](https://docs.docker.com/docker-hub/mirror/) that we could use instead, that might be the simplest solution.

PR #620, opened 2024-04-29
Add trafficReplayerMaxUptime parameter and functionality

AndreKurait:
### Description
This change adds a new deployment parameter for the trafficReplayer `trafficReplayerMaxUptime` along with functionality and documentation. This allows a user to set a max uptime after which AWS will startup a new container and kill the old one.

This achieves the functionality by adding a sidecar "MaxUptimeContainer" which succeeds a healthcheck after a given startup time and fails after a given maxUptime. This allows ECS to provision a new task and wait for the startup time before sending SIGTERM to the old task.

Shutdown Behavior: Note, shutdown logs for the existing replayer were not observed and replayer. Created task to follow up on SIGTERM behavior in [MIGRATIONS-1698](https://opensearch.atlassian.net/browse/MIGRATIONS-1698)

* Category: New Feature
* Why these changes are required? This allows customers to have more controlled behavior of the traffic replayer by limiting edge case scenarios that can arise from a long running task like out of memory errors due to leaks and the time between startup and teardown increasing
* What is the old behavior before changes and new behavior after changes? New parameter `trafficReplayerMaxUptime` added, no change to default behavior when parameter is not specified

### Issues Resolved
[MIGRATIONS-1670](https://opensearch.atlassian.net/browse/MIGRATIONS-1670)

### Testing
Deployed to AWS and verified restarting behavior. Note, an observation on shutdown behavior was made resulting in followup [MIGRATIONS-1698](https://opensearch.atlassian.net/browse/MIGRATIONS-1698)

### Check List


deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -146,6 +148,38 @@ export class MigrationServiceCore extends Stack {
...
+                        "UPTIME=$(awk '{print int($1)}' /proc/uptime); " +
+                        `test $UPTIME -gt ${startupPeriodSeconds} && ` +
+                        `test $UPTIME -lt ${maxUptimeSeconds}`
+                    ],
+                    timeout: Duration.seconds(2),
+                    retries: 1,
+                    startPeriod: Duration.seconds(startupPeriodSeconds * 2)
+                }
+            });
+            maxUptimeContainer.addContainerDependencies({
```
gregschohn, line 176: can you add a comment to say which way the dependency is running?  Who depends upon whom.
What does "START" mean two lines below?

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -146,6 +148,38 @@ export class MigrationServiceCore extends Stack {
...
+                    "/bin/sh",
+                    "-c",
+                    "sleep infinity"
+                ],
+                essential: true,
+                healthCheck: {
+                    command: [
+                        "CMD-SHELL",
+                        "UPTIME=$(awk '{print int($1)}' /proc/uptime); " +
+                        `test $UPTIME -gt ${startupPeriodSeconds} && ` +
```
gregschohn, line 168: why is this line needed?

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -146,6 +148,38 @@ export class MigrationServiceCore extends Stack {
...
+                image: ContainerImage.fromRegistry("public.ecr.aws/amazonlinux/amazonlinux:2023-minimal"), 
+                memoryLimitMiB: 64, 
+                entryPoint: [
+                    "/bin/sh",
+                    "-c",
+                    "sleep infinity"
+                ],
+                essential: true,
+                healthCheck: {
+                    command: [
```
gregschohn, line 165: A one-line, high-level comment about using uptime of the host every time the healthcheck is called would be helpful for others that aren't familiar with ECS/

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
```diff
@@ -146,6 +148,38 @@ export class MigrationServiceCore extends Stack {
...
+                        "CMD-SHELL",
+                        "UPTIME=$(awk '{print int($1)}' /proc/uptime); " +
+                        `test $UPTIME -gt ${startupPeriodSeconds} && ` +
+                        `test $UPTIME -lt ${maxUptimeSeconds}`
+                    ],
+                    timeout: Duration.seconds(2),
+                    retries: 1,
```
gregschohn, line 172: what will these mean if there's a failure?  Just bail and consider it dead?

deployment/cdk/opensearch-service-migration/options.md
```diff
@@ -24,6 +24,7 @@ These tables list all CDK context configuration values a user can specify for th
...
 | trafficReplayerExtraArgs             | string  | "--sigv4-auth-header-service-region es,us-east-1 --speedup-factor 5"   | Extra arguments to provide to the Replayer command. This includes auth header options and other parameters supported by the [Traffic Replayer](../../../TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java).                                     |
+| trafficReplayerMaxUptime          | string  | "PT2H"                                                                 | The maximum uptime for the Traffic Replayer service, specified in ISO 8601 duration format. This controls how long the Traffic Replayer will run before automatically shutting down. Example values: "PT30M" (30 minutes), "PT2H" (2 hours). When this duration is reached, ECS will initiate the startup of a new Traffic Replayer task to ensure continuous operation. This mechanism ensures that the Traffic Replayer service can manage its resources effectively and prevent issues associated with long running processes. Set to the greater of the given value 5 minutes. |
```
gregschohn, line 27: 2H seems a little arbitrary.  I would say in terms of most easy to understand, hourly, daily (if it's a big problem to recycle too often) or even every 6 hours.  Doing this often enough that an operator WILL see it happen could be a plus.

Where is the default actually defined?

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: This sets the minimum time that a task should be up before considering it heatlhy. This is needed because we don't have healthchecks configured for our replayer. 

We can reduce it lower than 30 seconds, and possibly remove it, but then we increase the risk of having a few second delay between the partitions being revoked on the existing task vs assigned to the new one

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: Say /proc/uptime is unavailable or something like that, ecs, will retry the call once more, but then consider it dead. Given this container is essential, the task will be killed and restarted

deployment/cdk/opensearch-service-migration/options.md
AndreKurait: There is no default currently so the deployment will not deploy the artifacts (maxUptimeContainer) needed to monitor this. 

I'll change this to discuss hourly and daily as examples

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: Added

gregschohn (approved):
Approved with a rec to add one more comment.  Thanks!

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-service-core.ts
AndreKurait: added

PR #622, opened 2024-04-30
Increase default MAX_PAYLOAD_BYTES_TO_PRINT to 100MiB

AndreKurait:
### Description
Increases MAX_PAYLOAD_BYTES_TO_PRINT to 100 MiB to keep in line with [AWS Opensearch Support](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/limits.html#network-limits)
* Category: Bug Fix
* Why these changes are required? The default configuration should work with aws opensearch 
* What is the old behavior before changes and new behavior after changes? Large requests over 1MB would not be included in the tuple output. Now by default all requests that are supported within AWS Managed Opensearch will work.

### Issues Resolved
[MIGRATIONS-1715](https://opensearch.atlassian.net/browse/MIGRATIONS-1715)

### Testing
Unit Testing

### Check List


gregschohn (approved):
This is sloppy, but I'm fine doing this since we're already tracking the bytes anyway.  I'm also OK since we'll likely soon be discarding this code to stream it w/out the aggregator anyway.

PR #623, opened 2024-05-01
[RFS] Refactored DocumentReindexer, HTTP client usage, added a OpenSearchClient

chelma:
### Description
* Checkpoint commit.  Refactored the RFS code:
    * Cleaned up DocumentReindexer error handling and logging
    * Made reactor-netty the sole HTTP client provider
    * Created an OpenSearchClient layer above the RestClient

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1691

### Testing
* Updated existing unit tests
* Ran locally

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
+++ b/RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
@@ -358,12 +358,17 @@ public class ReindexFromSnapshot {
 
                         Flux<Document> documents = new LuceneDocumentsReader().readDocuments(luceneDirPath, indexMetadata.getName(), shardId);
                         String targetIndex = indexMetadata.getName() + indexSuffix;
-                        DocumentReindexer.reindex(targetIndex, documents, targetConnection);
 
-                        logger.info("Shard reindexing completed");
+                        int targetShardId = shardId; // Define in local context for the lambda
+                        DocumentReindexer.reindex(targetIndex, documents, targetConnection)
+                            .doOnError(error -> logger.error("Error during reindexing: " + error))
+                            .doOnSuccess(done -> logger.info("Reindexing completed for index " + targetIndex + ", shard " + targetShardId))
+                            // Wait for the shard reindexing to complete before proceeding; fine in this demo script, but
+                            // shouldn't be done quite this way in the real RFS Worker.
+                            .block();
                     }
                 }
-                logger.info("Refreshing newly added documents");
+                logger.info("Refreshing target cluster to reflect newly added documents");
                 DocumentReindexer.refreshAllDocuments(targetConnection);
                 logger.info("Refresh complete");
             }
--- a/RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
+++ b/RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
@@ -182,9 +182,10 @@ public class ReindexFromSnapshot {
                 // ==========================================================================================================            
                 logger.info("==================================================================");
                 logger.info("Attempting to create the snapshot...");
+                OpenSearchClient sourceClient = new OpenSearchClient(sourceConnection);
                 SnapshotCreator snapshotCreator = repo instanceof S3Repo
-                    ? new S3SnapshotCreator(snapshotName, sourceConnection, s3RepoUri, s3Region)
-                    : new FileSystemSnapshotCreator(snapshotName, sourceConnection, snapshotLocalRepoDirPath.toString());
+                    ? new S3SnapshotCreator(snapshotName, sourceClient, s3RepoUri, s3Region)
+                    : new FileSystemSnapshotCreator(snapshotName, sourceClient, snapshotLocalRepoDirPath.toString());
                 snapshotCreator.registerRepo();
                 snapshotCreator.createSnapshot();
                 while (!snapshotCreator.isSnapshotFinished()) {
@@ -266,14 +267,15 @@ public class ReindexFromSnapshot {
                 logger.info("==================================================================");
                 logger.info("Attempting to recreate the Global Metadata...");
 
+                OpenSearchClient targetClient = new OpenSearchClient(targetConnection);
                 if (sourceVersion == ClusterVersion.ES_6_8) {
                     ObjectNode root = globalMetadata.toObjectNode();
                     ObjectNode transformedRoot = transformer.transformGlobalMetadata(root);                    
-                    GlobalMetadataCreator_OS_2_11.create(transformedRoot, targetConnection, Collections.emptyList(), templateWhitelist);
+                    GlobalMetadataCreator_OS_2_11.create(transformedRoot, targetClient, Collections.emptyList(), templateWhitelist);
                 } else if (sourceVersion == ClusterVersion.ES_7_10) {
                     ObjectNode root = globalMetadata.toObjectNode();
                     ObjectNode transformedRoot = transformer.transformGlobalMetadata(root);                    
-                    GlobalMetadataCreator_OS_2_11.create(transformedRoot, targetConnection, componentTemplateWhitelist, templateWhitelist);
+                    GlobalMetadataCreator_OS_2_11.create(transformedRoot, targetClient, componentTemplateWhitelist, templateWhitelist);
                 }
             }
 
@@ -301,6 +303,7 @@ public class ReindexFromSnapshot {
                 // ==========================================================================================================
                 logger.info("==================================================================");
                 logger.info("Attempting to recreate the indices...");
+                OpenSearchClient targetClient = new OpenSearchClient(targetConnection);
                 for (IndexMetadata.Data indexMetadata : indexMetadatas) {
                     String reindexName = indexMetadata.getName() + indexSuffix;
                     logger.info("Recreating index " + indexMetadata.getName() + " as " + reindexName + " on target...");
@@ -308,7 +311,7 @@ public class ReindexFromSnapshot {
                     ObjectNode root = indexMetadata.toObjectNode();
                     ObjectNode transformedRoot = transformer.transformIndexMetadata(root);
                     IndexMetadataData_OS_2_11 indexMetadataOS211 = new IndexMetadataData_OS_2_11(transformedRoot, indexMetadata.getId(), reindexName);
-                    IndexCreator_OS_2_11.create(reindexName, indexMetadataOS211, targetConnection);
+                    IndexCreator_OS_2_11.create(reindexName, indexMetadataOS211, targetClient);
                 }
             }
 
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -16,16 +16,14 @@ public class DocumentReindexer {
     private static final int MAX_BATCH_SIZE = 1000; // Arbitrarily chosen
 
     public static Mono<Void> reindex(String indexName, Flux<Document> documentStream, ConnectionDetails targetConnection) throws Exception {   
-        RestClient client = new RestClient(targetConnection);
-
-        String targetPath = indexName + "/_bulk";
+        OpenSearchClient client = new OpenSearchClient(targetConnection);
 
         return documentStream
             .map(DocumentReindexer::convertDocumentToBulkSection)  // Convert each Document to part of a bulk operation
             .buffer(MAX_BATCH_SIZE) // Collect until you hit the batch size
             .doOnNext(bulk -> logger.info(bulk.size() + " documents in current bulk request"))
             .map(DocumentReindexer::convertToBulkRequestBody)  // Assemble the bulk request body from the parts
-            .flatMap(bulkJson -> client.postBulkAsync(targetPath, bulkJson) // Send the request
+            .flatMap(bulkJson -> client.sendBulkRequest(indexName, bulkJson) // Send the request
                 .doOnSuccess(unused -> logger.debug("Batch succeeded"))
                 .doOnError(error -> logger.error("Batch failed", error))
                 .onErrorResume(e -> Mono.empty()) // Prevent the error from stopping the entire stream
@@ -53,7 +51,7 @@ public class DocumentReindexer {
 
     public static void refreshAllDocuments(ConnectionDetails targetConnection) throws Exception {
         // Send the request
-        RestClient client = new RestClient(targetConnection);
-        client.get("_refresh", false);
+        OpenSearchClient client = new OpenSearchClient(targetConnection);
+        client.refresh();
     }
 }
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -18,7 +18,7 @@ public class DocumentReindexer {
     private static final Logger logger = LogManager.getLogger(DocumentReindexer.class);
     private static final int MAX_BATCH_SIZE = 1000; // Arbitrarily chosen
 
-    public static void reindex(String indexName, Flux<Document> documentStream, ConnectionDetails targetConnection) throws Exception {
+    public static Mono<Void> reindex(String indexName, Flux<Document> documentStream, ConnectionDetails targetConnection) throws Exception {
         String targetUrl = "/" + indexName + "/_bulk";
         HttpClient client = HttpClient.create()
             .host(targetConnection.hostName)
@@ -32,16 +32,19 @@ public class DocumentReindexer {
                 }
             });
 
-        documentStream
+        return documentStream
             .map(DocumentReindexer::convertDocumentToBulkSection)  // Convert each Document to part of a bulk operation
             .buffer(MAX_BATCH_SIZE) // Collect until you hit the batch size
+            .doOnNext(bulk -> logger.info(bulk.size() + " documents in current bulk request"))
             .map(DocumentReindexer::convertToBulkRequestBody)  // Assemble the bulk request body from the parts
-            .flatMap(bulkJson -> sendBulkRequest(client, targetUrl, bulkJson)) // Send the request
+            .flatMap(bulkJson -> sendBulkRequest(client, targetUrl, bulkJson) // Send the request
+                .doOnSuccess(unused -> logger.debug("Batch succeeded"))
+                .doOnError(error -> logger.error("Batch failed", error))
+                .onErrorResume(e -> Mono.empty()) // Prevent the error from stopping the entire stream
+            )
             .retryWhen(Retry.backoff(3, Duration.ofSeconds(1)).maxBackoff(Duration.ofSeconds(5)))
-            .subscribe(
-                response -> logger.info("Batch uploaded successfully"),
-                error -> logger.error("Failed to upload batch", error)
-            );
+            .doOnComplete(() -> logger.debug("All batches processed"))
+            .then();
     }
 
     private static String convertDocumentToBulkSection(Document document) {
@@ -53,7 +56,6 @@ public class DocumentReindexer {
     }
 
     private static String convertToBulkRequestBody(List<String> bulkSections) {
-        logger.info(bulkSections.size() + " documents in current bulk request");
         StringBuilder builder = new StringBuilder();
         for (String section : bulkSections) {
             builder.append(section).append("\n");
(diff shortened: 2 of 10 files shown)


RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
```diff
@@ -358,12 +361,17 @@ public static void main(String[] args) throws InterruptedException {
 
                         Flux<Document> documents = new LuceneDocumentsReader().readDocuments(luceneDirPath, indexMetadata.getName(), shardId);
                         String targetIndex = indexMetadata.getName() + indexSuffix;
-                        DocumentReindexer.reindex(targetIndex, documents, targetConnection);
 
-                        logger.info("Shard reindexing completed");
+                        int targetShardId = shardId; // Define in local context for the lambda
```
gregschohn, line 365: super minor style thing - you can make this explicitly final and keep the same name other than some convention (like prepending the word 'final') to show that you're only doing it for the lambda.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -1,47 +1,36 @@
...
-            .port(targetConnection.port)
-            .headers(h -> {
-                h.set("Content-Type", "application/json");
-                if (targetConnection.authType == ConnectionDetails.AuthType.BASIC) {
-                    String credentials = targetConnection.username + ":" + targetConnection.password;
-                    String encodedCredentials = Base64.getEncoder().encodeToString(credentials.getBytes());
-                    h.set("Authorization", "Basic " + encodedCredentials);
-                }
-            });
+    public static Mono<Void> reindex(String indexName, Flux<Document> documentStream, ConnectionDetails targetConnection) throws Exception {
```
gregschohn, line 18: Returning something for every future operation makes a ton of sense.  Take a look at my [PR to overhaul our async promises](https://github.com/opensearch-project/opensearch-migrations/pull/614).  Returning these so that they can be chained, even if just for monitoring is a good move.  
I'm not familiar with the reactor client, but there may be a longer term question about whether you want/need to embrace reactor primitives vs more universal ones (like CompletableFuture), which we have significant diagnostic support for at the moment.  Longer term (PR-614 switched out the netty future stuff for CompletableFutures, so it's a straightforward if borderline arduous task).

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -1,47 +1,36 @@
...
-        documentStream
+        return documentStream
             .map(DocumentReindexer::convertDocumentToBulkSection)  // Convert each Document to part of a bulk operation
             .buffer(MAX_BATCH_SIZE) // Collect until you hit the batch size
+            .doOnNext(bulk -> logger.info(bulk.size() + " documents in current bulk request"))
             .map(DocumentReindexer::convertToBulkRequestBody)  // Assemble the bulk request body from the parts
-            .flatMap(bulkJson -> sendBulkRequest(client, targetUrl, bulkJson)) // Send the request
+            .flatMap(bulkJson -> client.sendBulkRequest(indexName, bulkJson) // Send the request
+                .doOnSuccess(unused -> logger.debug("Batch succeeded"))
+                .doOnError(error -> logger.error("Batch failed", error))
```
gregschohn, line 28: You'll want to include some identifier so that you can determine the cause and impact of the failure.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -1,47 +1,36 @@
...
-                    String credentials = targetConnection.username + ":" + targetConnection.password;
-                    String encodedCredentials = Base64.getEncoder().encodeToString(credentials.getBytes());
-                    h.set("Authorization", "Basic " + encodedCredentials);
-                }
-            });
+    public static Mono<Void> reindex(String indexName, Flux<Document> documentStream, ConnectionDetails targetConnection) throws Exception {   
+        OpenSearchClient client = new OpenSearchClient(targetConnection);
 
-        documentStream
+        return documentStream
```
gregschohn, line 21: reactor does look like a pretty nice and powerful library.  I'm wondering if we should think about using it for the replayer, especially as we get more serious about retries.  I'm wondering if it supports scheduling.

RFS/src/main/java/com/rfs/common/FileSystemSnapshotCreator.java
```diff
@@ -9,14 +9,14 @@ public class FileSystemSnapshotCreator extends SnapshotCreator {
     private static final Logger logger = LogManager.getLogger(FileSystemSnapshotCreator.class);
     private static final ObjectMapper mapper = new ObjectMapper();
 
-    private final ConnectionDetails connectionDetails;
+    private final OpenSearchClient client;
```
gregschohn, line 12: nit: what kind of client is this for - it's to upload source cluster data, source cluster metadata, or RFS metadata?

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -35,7 +35,7 @@ public Flux<Document> readDocuments(Path luceneFilesBasePath, String indexName,
                 try {
                     reader.close();
                 } catch (IOException e) {
-                    logger.error("Failed to close IndexReader", e);
+                    throw new RuntimeException("Failed to close IndexReader", e);
```
gregschohn, line 38: nit: you'll get snagged on a linter here.  If you use Lombok.sneakyThrows(e), you're fine.  It makes backtraces easier to understand and handle too.  Typed exception handling is a language construct, not a VM/runtime one - so you basically just say, "hey compiler, let me throw this thing & ignore your checks" - making this the same as a RuntimeException to callers.  This becomes especially critical once you have to unpack exceptions from async activities that could be wrapped numerous times otherwise (`completedException.cause().cause().instanceof(...) || completedException.cause().cause().cause().instanceof(...)` - remember that from the Divine Comedy ;) )

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -28,108 +25,61 @@ public Response(int responseCode, String responseBody, String responseMessage) {
...
-            logger.error("Response Code: " + responseCode + ", Response Message: " + conn.getResponseMessage() + ", Response Body: " + responseBody);
-        }
-
-        conn.disconnect();
 
-        return new Response(responseCode, responseBody, conn.getResponseMessage());
+        this.client = HttpClient.create()
+            .baseUrl(connectionDetails.url)
+            .headers(h -> {
+                h.add("Content-Type", "application/json");
```
gregschohn, line 36: Is this always true for ES/OS?  I feel like I've seen some whacky content types, including PNGs and fonts (no idea why) go in as documents.  Would it make sense to try to run those as batch puts?  I wouldn't worry about it much - just put a check somewhere on the doctypes and throw if it's something truly whacky.

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -28,108 +25,61 @@ public Response(int responseCode, String responseBody, String responseMessage) {
...
+        return client.get()
+            .uri("/" + path)
+            .responseSingle((response, bytes) -> bytes.asString()
+                .map(body -> new Response(response.status().code(), body, response.status().reasonPhrase()))
+                .doOnNext(resp -> logResponse(resp, quietLogging)));
+    }
 
-        // Report the results
-        int responseCode = conn.getResponseCode();
+    public Response get(String path, boolean quietLogging) {
```
gregschohn, line 53: where do you use the blocking calls?  Given that it looks like everything is async, why not go all the way :)

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -28,108 +25,61 @@ public Response(int responseCode, String responseBody, String responseMessage) {
...
+        return putAsync(path, body, quietLogging).block();
     }
 
-    private void setAuthHeader(HttpURLConnection conn) {
-        if (connectionDetails.authType == ConnectionDetails.AuthType.BASIC) {
-            String auth = connectionDetails.username + ":" + connectionDetails.password;
-            byte[] encodedAuth = Base64.getEncoder().encode(auth.getBytes(StandardCharsets.UTF_8));
-            String authHeaderValue = "Basic " + new String(encodedAuth);
-            conn.setRequestProperty("Authorization", authHeaderValue);
+    private void logResponse(Response response, boolean quietLogging) {
```
gregschohn, line 78: Should this be protected so that others can leverage the class with different policies on logging?

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -0,0 +1,127 @@
...
+        public BulkResponse(int responseCode, String responseBody, String responseMessage) {
+            super(responseCode, responseBody, responseMessage);
+        }
+
+        public boolean hasBadStatusCode() {
+            return !(code == HttpURLConnection.HTTP_OK || code == HttpURLConnection.HTTP_CREATED);
+        }
+
+        public boolean hasFailedOperations() {
+            return body.contains("\"errors\":true");
```
gregschohn, line 25: Leave a TODO to strengthen this so that you're looking in the right place or a comment as to why it works?  You might already be immune to checking within a document if the documents are already json encoded/escaped.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -0,0 +1,127 @@
...
+        return createObjectIdempotent(targetPath, settings);
+    }
+
+    private boolean createObjectIdempotent(String objectPath, ObjectNode settings){
+        RestClient.Response response = client.get(objectPath, true);
+        if (response.code == HttpURLConnection.HTTP_NOT_FOUND) {
+            client.put(objectPath, settings.toString(), false);
+            return true;
+        } else if (response.code == HttpURLConnection.HTTP_OK) {
+            logger.warn(objectPath + " already exists. Skipping creation.");
```
gregschohn, line 86: if this was an idempotent operation, should this be a warning log message?

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -0,0 +1,127 @@
...
+        String targetPath = indexName;
+        return createObjectIdempotent(targetPath, settings);
+    }
+
+    private boolean createObjectIdempotent(String objectPath, ObjectNode settings){
+        RestClient.Response response = client.get(objectPath, true);
+        if (response.code == HttpURLConnection.HTTP_NOT_FOUND) {
+            client.put(objectPath, settings.toString(), false);
+            return true;
+        } else if (response.code == HttpURLConnection.HTTP_OK) {
```
gregschohn, line 85: Is this the right status code?  It doesn't seem like it from the log line below

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
chelma: Thanks - done.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
chelma: I'm not sure myself, yet.  I think there is a limit to how far up the call stack this stuff will propagate; I've hit that limit in some local changes I have that aren't in this PR.  Definitely something to keep an eye on, agree!

Took a look at the referenced PR; don't think the RFS behavior is complex enough to need that level of abstraction.  I could be proven wrong, but it feels so far like we can get by with just using the Reactor primitives?

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
chelma: Yeah, I think it does support scheduling.  It's a pub-sub model, and you can write your own subscriber model and pull from the publisher however you want.  At least, that's my understanding from the docs.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
chelma: In this case, the error message includes the response body of the failed POST.  I think that should be sufficient?

RFS/src/main/java/com/rfs/common/FileSystemSnapshotCreator.java
chelma: Quite literally, a client to interact w/ OpenSearch as a layer above the RestClient.  So, yes.  We may have another client layer abstraction above it (such as the OpenSearchCmsClient class I have in local changes that uses an OpenSearchClient to interact w/ the Coordinating Metadata Store).

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
chelma: Done!

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
chelma: Left a comment; hope it's satisfactory.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
chelma: It should be the correct status code, as it's for the GET not the PUT.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
chelma: Yeah, you're right, probably shouldn't be a WARN.  Will update.

RFS/src/main/java/com/rfs/common/RestClient.java
chelma: I honestly don't even know how to check for that at this point; I'll make a JIRA to help capture the miss here.

chelma: Done - https://opensearch.atlassian.net/browse/MIGRATIONS-1718

RFS/src/main/java/com/rfs/common/RestClient.java
chelma: We use the blocking calls where performance isn't important but ordering is - like whether to do a PUT based on the response from a GET.

RFS/src/main/java/com/rfs/common/RestClient.java
chelma: I ripped it out in my local changes, consider this function gone.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -0,0 +1,130 @@
...
+
+    public OpenSearchClient(ConnectionDetails connectionDetails) {
+        this.connectionDetails = connectionDetails;
+        this.client = new RestClient(connectionDetails);
+    }
+
+    /*
+     * Create a legacy template if it does not already exist; return true if created, false otherwise.
+     */
+    public boolean createLegacyTemplateIdempotent(String templateName, ObjectNode settings){
```
peternied, line 54: Design Nit-pick: Separate use of idempotent and non-idempotent functionality with intermediate classes so you call looks like `client.idempotent().createLegacyTemplate(...)` vs `client.stateful().createSnapshot(...)`

Along these lines, expose the `client.idempotent()` interface at certain levels of the project to be really sure you aren't doing the wrong thing.

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -28,108 +25,61 @@ public Response(int responseCode, String responseBody, String responseMessage) {
...
-        }
-
-        conn.disconnect();
 
-        return new Response(responseCode, responseBody, conn.getResponseMessage());
+        this.client = HttpClient.create()
+            .baseUrl(connectionDetails.url)
+            .headers(h -> {
+                h.add("Content-Type", "application/json");
+                if (connectionDetails.authType == ConnectionDetails.AuthType.BASIC) {
```
peternied, line 37: Design Nitpick: We are going to need a better way to handle authentication in the future such as sigv4 and other methods.  IMO this is the case for using some kind of preconstructed client for interacting with the cluster.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
gregschohn: Having a chain of 2-3 dependencies may be complex enough if you want to make sure that things aren't wedging forever or unintentionally holding up other calls.  I'm not sure where you're at, but if you have a web-client, you'll have more than that.  The reason to use something like TrackedFutures is that if you have > 1K activities per second, you're not really going to want to log enough details that if one goes wrong, you can find out what the issue is.  That would be slow and expensive (in CPU/storage + user time).  The approach that the replayer now takes is to keep those graphs around, then render only the ones that are becoming a problem.

It would be good to check the impact of TrackedFutures before putting it into more parts of the solution (it hasn't gone into the proxy yet).

This thread of comment came up because of the return type.  If you're worried about how high your reactor client is going, you could convert a mono object to a CompletableFuture (like what is done in NettyFutureBinders).

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
gregschohn: Oh, I missed the error, sorry.  If error is an exception, you could also do `logger.atError(()->"Label").setCause(error).log`

RFS/src/main/java/com/rfs/common/FileSystemSnapshotCreator.java
gregschohn: That tells me what the class is.  I was asking about what this particular instance was being used for within this class.  I was trying to understand how this class worked.

RFS/src/main/java/com/rfs/common/RestClient.java
gregschohn: Shouldn't you be chaining those calls?  I'm not sure how the threading model works with reactor, but with a netty client, all IO for a given channel runs on one thread.  Blocking on that thread can have profound, negative consequences (deadlock being one of them)

RFS/src/main/java/com/rfs/common/RestClient.java
gregschohn: It would be good for users to try to keep as much unity in our command line options as possible, but I haven't thought of a good way to do that.  I'm wary to try to mix and match different params into one object for a single Beust class.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -36,6 +39,7 @@ public Flux<Document> readDocuments(Path luceneFilesBasePath, String indexName,
                     reader.close();
                 } catch (IOException e) {
                     logger.error("Failed to close IndexReader", e);
```
gregschohn, line 41: you can also write this as logger.atError().setCause(e).setMessage(()->"...").log().
From my read, when I've looked at this in the past, the multi-arg error function is a pretty loose convention that just happens to work with log4j2, but it isn't really an slf4j standard.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
chelma: I'll take another pass at all the logging stuff a bit later, if you don't mind; want to get the core functionality in place right now.

RFS/src/main/java/com/rfs/common/FileSystemSnapshotCreator.java
chelma: Well, it's used to take a snapshot from source cluster to the local filesytem.  Should be self-evident from the class/method names?  If not, open to suggestions for ways to make it more obvious.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
chelma: I'll do logging pass a bit later, want to get the core functionality in place first.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
chelma: Good suggestion; we'll see how the code develops.  This isn't the only thing in the class I'm not feeling confident in so I expect there to be continuing changes to it.

RFS/src/main/java/com/rfs/common/RestClient.java
chelma: Probably; I think @gregschohn has some tasks created to look into auth.  Not required for the current iteration so I didn't invest the time in it.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
chelma: I've actually found how high up the Rector types go at this point; it's not as high as I feared and actually makes sense (to me, at least).  You'll get a good look when I put out the next PR soon w/ the RfsWorker coded up to perform the Snapshot Phase.

peternied (approved):
Looks like a good starting point, happy to see this merged

sumobrian (approved):
Approving on @peternied's behalf since he was not on codeowners/maintainers list at the time he approved.

PR #626, opened 2024-05-01
Add RFS gradle tasks for jacoco test report and copying of jars

lewijacn:
### Description
As part of our Solutions scanning we need to be able to generate a jacoco test report and provide list of jars for SonarQube to scan.

### Issues Resolved
N/A

### Testing
Local gradle builds

### Check List

### DIFF
--- a/RFS/build.gradle
+++ b/RFS/build.gradle
@@ -72,7 +72,7 @@ clean.doFirst {
 }
 
 // Utility task to allow copying required libraries into a 'dependencies' folder for security scanning
-tasks.register('copyDependencies', Copy) {
+tasks.register('copyDependencies', Sync) {
     duplicatesStrategy = DuplicatesStrategy.EXCLUDE
 
     from configurations.runtimeClasspath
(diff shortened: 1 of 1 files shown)


RFS/build.gradle
```diff
@@ -70,6 +71,23 @@ clean.doFirst {
     delete project.file("./docker/build")
 }
 
+// Utility task to allow copying required libraries into a 'dependencies' folder for security scanning
+tasks.register('copyDependencies', Copy) {
```
mikaylathompson, line 75: Is this another good use case for Sync? It seems like we could run into the same issue with leaving old dependencies in here.

mikaylathompson:
Also, is this only relevant for RFS?

lewijacn:
> Also, is this only relevant for RFS?

This logic already exists in C&R

RFS/build.gradle
lewijacn: It probably doesn't hurt to add just in case, but normally I'd expect a full clean each time by the Solutions pipeline that uses this.

mikaylathompson (approved):

PR #636, opened 2024-05-06
Add E2E large request test

AndreKurait:
### Description
Add E2E large request test
Fix documentation for E2E target_password

This task builds on top of #637

* Category: Enhancement
* Why these changes are required? Current E2E script misses tests on large requests which can have differing behavior
* What is the old behavior before changes and new behavior after changes? New Test which checks behavior for 20MB requests

### Issues Resolved

### Testing
Ran tests

### Check List

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -5,6 +5,7 @@ import pytest
 import requests
 import secrets
 import string
+from operations import generate_large_doc
 import subprocess
 import time
 import unittest
@@ -110,7 +111,7 @@ class E2ETests(unittest.TestCase):
                 return True
         return False
 
-    def assert_source_target_doc_match(self, index_name, doc_id):
+    def assert_source_target_doc_match(self, index_name, doc_id, doc_body : dict = None):
         source_response = get_document(self.source_endpoint, index_name, doc_id, self.source_auth,
                                        self.source_verify_ssl)
         self.assertEqual(source_response.status_code, HTTPStatus.OK)
@@ -127,6 +128,8 @@ class E2ETests(unittest.TestCase):
         target_document = target_response.json()
         target_content = target_document['_source']
         self.assertEqual(source_content, target_content)
+        if doc_body is not None:
+            self.assertEqual(source_content, doc_body)
 
     def set_common_values(self):
         self.index_prefix_ignore_list = ["test_", ".", "searchguard", "sg7", "security-auditlog"]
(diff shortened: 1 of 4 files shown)


test/tests.py
```diff
@@ -305,3 +311,59 @@ def test_0006_OSB(self):
...
+        finally:
+            proxy_single_connection_session.close()
+
+    def test_0008_largeRequest(self):
+        index_name = f"test_0008_{self.unique_id}"
+        doc_id = "1"
+
+        # Create large document, 20MB which is less than the default max of
+        # 100MB in http.max_content_length
+        large_doc = generate_large_doc(size_mb=20)
```
peternied, line 350: Why not go big with a 100mb doc?

test/tests.py
```diff
@@ -305,3 +311,59 @@ def test_0006_OSB(self):
...
+        self.assertEqual(proxy_response.status_code, HTTPStatus.CREATED)
+
+        # Wait for the measured duration
+        logger.debug(f"Waiting {round(duration, 3)} seconds for expected time "
+                     f"to replay large doc creation")
+
+        time.sleep(duration)
+
+        # Verify document created on source and target
+        self.assert_source_target_doc_match(index_name, doc_id)
```
peternied, line 369: What do you think about should validating a checksum of the doc to make sure it arrived as expected?

test/tests.py
peternied: Is it the doc itself that is important to be large vs a 100mb bulk request of 100kb documents, or is that a good example of another test scenario?

AndreKurait: I don't believe the replayer would differentiate the two at all, though it could be a good test.

I'll update to 99MB

test/tests.py
AndreKurait: Added a doc check

test/tests.py
```diff
@@ -305,3 +313,62 @@ def test_0006_OSB(self):
...
+                time.sleep(seconds_between_requests)
+
+        try:
+            for doc_id_int in range(number_of_docs):
+                doc_id = str(doc_id_int)
+                self.assert_source_target_doc_match(index_name, doc_id)
+        finally:
+            proxy_single_connection_session.close()
+
+    def test_0008_largeRequest(self):
```
peternied, line 346: This test looks good to me, thanks for the updates @AndreKurait

AndreKurait: Updated with only this test, please take a look

peternied (approved):

PR #637, opened 2024-05-06
Add test that verifies replayer function when waiting a minute between requests

AndreKurait:
### Description
Add test that verifies replayer function when waiting a minute between requests
Adds functionality within E2E test suite to explicitly reuse connection for requests

This was created in part to trigger potential edge cases that could have been associated with [ReadTimeoutHandler bug](https://opensearch.atlassian.net/browse/MIGRATIONS-1723)

* Category: Enhancement
* Why these changes are required? Verify no impact for bug
* What is the old behavior before changes and new behavior after changes? New test case

### Issues Resolved
[MIGRATIONS-1722](https://opensearch.atlassian.net/browse/MIGRATIONS-1722)

### Testing
E2E testing, verified bug is triggered during test case

### Check List
  - [x ] All tests pass, including unit test, integration test and doctest

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -316,11 +316,9 @@ class E2ETests(unittest.TestCase):
         # requests on the same connection on the proxy that has a minute gap
         seconds_between_requests = 60  # 1 minute
 
-        # Set a consistent session to be used for this test
-        session = Session()
+        proxy_single_connection_session = Session()
         adapter = HTTPAdapter(pool_connections=1, pool_maxsize=1, max_retries=1)
-        session.mount('http://', adapter)
-        session.mount('https://', adapter)
+        proxy_single_connection_session.mount(self.proxy_endpoint, adapter)
 
         index_name = f"test_0007_{self.unique_id}"
 
@@ -329,7 +327,7 @@ class E2ETests(unittest.TestCase):
         for doc_id_int in range(number_of_docs):
             doc_id = str(doc_id_int)
             proxy_response = create_document(self.proxy_endpoint, index_name, doc_id, self.source_auth,
-                                             self.source_verify_ssl, session)
+                                             self.source_verify_ssl, proxy_single_connection_session)
             self.assertEqual(proxy_response.status_code, HTTPStatus.CREATED)
 
             if doc_id_int + 1 < number_of_docs:
@@ -340,4 +338,4 @@ class E2ETests(unittest.TestCase):
                 doc_id = str(doc_id_int)
                 self.assert_source_target_doc_match(index_name, doc_id)
         finally:
-            session.close()
+            proxy_single_connection_session.close()
--- a/test/tests.py
+++ b/test/tests.py
@@ -1,21 +1,22 @@
+import boto3
 import json
+import logging
+import pytest
+import requests
+import secrets
+import string
 import subprocess
+import time
+import unittest
+from http import HTTPStatus
+from requests import Session
+from requests.adapters import HTTPAdapter
+from requests.exceptions import ConnectionError, SSLError
+from requests_aws4auth import AWS4Auth
+from typing import Tuple, Callable
 
 from operations import create_index, check_index, create_document, \
     delete_document, delete_index, get_document
-from http import HTTPStatus
-from typing import Tuple, Callable
-import unittest
-import logging
-import time
-import requests
-import string
-import secrets
-import pytest
-import boto3
-from requests_aws4auth import AWS4Auth
-
-from requests.exceptions import ConnectionError, SSLError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


test/tests.py
```diff
@@ -305,3 +310,34 @@ def test_0006_OSB(self):
...
+    def test_0007_timeBetweenRequestsOnSameConnection(self):
+        # This test will verify that the replayer functions correctly when
+        # requests on the same connection on the proxy that has a minute gap
+        seconds_between_requests = 60  # 1 minute
+
+        # Set a consistent session to be used for this test
+        session = Session()
+        adapter = HTTPAdapter(pool_connections=1, pool_maxsize=1, max_retries=1)
+        session.mount('http://', adapter)
+        session.mount('https://', adapter)
```
gregschohn, line 323: why are you using both of these?

test/tests.py
```diff
@@ -305,3 +310,34 @@ def test_0006_OSB(self):
...
+
+        index_name = f"test_0007_{self.unique_id}"
+
+        number_of_docs = 3
+
+        for doc_id_int in range(number_of_docs):
```
gregschohn, line 329: You could do this as a simpler integral for loop (or while loop)

test/tests.py
```diff
@@ -305,3 +310,34 @@ def test_0006_OSB(self):
...
+            doc_id = str(doc_id_int)
+            proxy_response = create_document(self.proxy_endpoint, index_name, doc_id, self.source_auth,
+                                             self.source_verify_ssl, session)
+            self.assertEqual(proxy_response.status_code, HTTPStatus.CREATED)
+
+            if doc_id_int + 1 < number_of_docs:
+                time.sleep(seconds_between_requests)
+
+        try:
+            for doc_id_int in range(number_of_docs):
```
gregschohn, line 339: Do you need a sleep before this, or some other way to poll on the target cluster in case the document hasn't been replayed yet?

test/tests.py
AndreKurait: to work in both ssl and non-ssl cases

test/tests.py
AndreKurait: I've seen `for elem in range()` much more frequently used in python than for or while loops

test/tests.py
AndreKurait: the assert_source_target_doc_match does a poll on the target. No sleep is needed

test/tests.py
gregschohn: What does mount do?

test/tests.py
gregschohn: where is the poll?  I didn't see it.

AndreKurait: >        target_response = retry_request(get_document, args=(self.target_endpoint, index_name, doc_id,
>                                                           self.target_auth, self.target_verify_ssl),
>                                        expected_status_code=HTTPStatus.OK)
Its the retry_request. That's how the existing tests check, it performs a backoff and retry until it get's the desired status code or preconfigured max number or retries occur

test/tests.py
AndreKurait: mount applies the adapter to a given uri prefix. I've since modified the code to directly use the proxy endpoint for clarity here

test/tests.py
```diff
@@ -305,3 +310,32 @@ def test_0006_OSB(self):
             if source_count != target_count:
                 self.assertEqual(source_count, target_count, f'{index}: doc counts do not match - '
                                                              f'Source = {source_count}, Target = {target_count}')
+
+    def test_0007_timeBetweenRequestsOnSameConnection(self):
+        # This test will verify that the replayer functions correctly when
+        # requests on the same connection on the proxy that has a minute gap
+        seconds_between_requests = 60  # 1 minute
```
peternied, line 317: This test will take at least 180 seconds to run?  Is there anyway the system can be setup with shorter limits such as 10 second buffers that would still validate the scenario, but only take ~30 to run?

AndreKurait: This test specifically is waiting a minute between requests as a somewhat arbitrary value. For the purposes of the included bug, i was interested in the delay being > than target connection timeout, which is set at the default of 30 seconds.

As is, i could reduce this delay to 35 seconds and achieve that goal, or we could set up infrastructure to reduce that and make a constrained test.

My thoughts is that the shorter limits tests would be better as a unit test possibly with docker containers, whereas this is intended to validate broad scenarios which may cover different bugs

peternied: How do we know 60 second delay is correct to evaluate vs a 90 second delay would actually show an issue?

AndreKurait: We don't it's arbitrary

peternied: So why are we writing this test, did we have an issue in the past this would have caught?  Could we revert the fix and confirm the test would validate that scenario?

gregschohn (approved):

test/tests.py
AndreKurait: We did not have a specific behavior bug that this would have caught, but there was a component which was broken and this test triggered the error case showing a "ReadTimeoutException" in the log. That I did tests (as i created this test case prior to pushing out that other fix)

peternied: I guess my concern would be - if that scenario happened again do we have enough information to diagnose what took place, but at this point it feels like I am being pedantic, I'd rather have the test running to prompt investigation and improvement to data collection troubleshooting.

peternied (approved):

PR #640, opened 2024-05-08
Fix ReadTimeoutException between requests

AndreKurait:
### Description
Currently with the traffic replayer, in the NettyPacketToHttpConsumer, the ReadTimeoutHandler is coupled with the  ConnectionClosedListenerHandler.

This means that we count the time in between request towards the ReadTimeout. This causes the channel to throw an exception while it is waiting to be re-used and may cause race conditions if the channel is reused exactly at the timeout interval.

This change places the ReadTimeoutHandler behind the handlers which remain between requests.

This change also fixes an issue where the BacksideHttpWatcherHandler was propagating exceptions through even though  it shouldn't since it handles them through the callback finalization.

* Category: Bug Fix
* Why these changes are required? Properly handle read timeouts
* What is the old behavior before changes and new behavior after changes? ReadTimeoutException would occur between requests on the same connection/channel, now the exception only occurs if there is a read timeout during the request and is not sent to the bottom of the pipeline.

### Issues Resolved
[MIGRATIONS-1723](https://opensearch.atlassian.net/browse/MIGRATIONS-1723)

### Testing
Ran existing and added new unit tests. Also ran E2E tests from #637 and verified no ReadTimeoutException in logs (they were occurring before this change)

### Check List
- [ x] New functionality has been documented

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
@@ -22,7 +22,6 @@ import org.opensearch.migrations.replay.util.TextTrackedFuture;
 import org.opensearch.migrations.replay.util.TrackedFuture;
 
 import java.net.URI;
-import java.time.Duration;
 import java.util.concurrent.CompletableFuture;
 
 @Slf4j
@@ -32,7 +31,6 @@ public class ClientConnectionPool {
     private final SslContext sslContext;
     public final NioEventLoopGroup eventLoopGroup;
     private final LoadingCache<Key, ConnectionReplaySession> connectionId2ChannelCache;
-    private final Duration timeout;
 
     @EqualsAndHashCode
     @AllArgsConstructor
@@ -48,11 +46,9 @@ public class ClientConnectionPool {
     public ClientConnectionPool(@NonNull URI serverUri,
                                 SslContext sslContext,
                                 @NonNull String targetConnectionPoolName,
-                                int numThreads,
-                                @NonNull Duration timeout) {
+                                int numThreads) {
         this.serverUri = serverUri;
         this.sslContext = sslContext;
-        this.timeout = timeout;
         this.eventLoopGroup =
                 new NioEventLoopGroup(numThreads, new DefaultThreadFactory(targetConnectionPoolName));
 
@@ -80,7 +76,7 @@ public class ClientConnectionPool {
         return new AdaptiveRateLimiter<String, ChannelFuture>()
                 .get(() ->
                         NettyPacketToHttpConsumer.createClientConnection(eventLoop,
-                                        sslContext, serverUri, connectionContext, timeout)
+                                        sslContext, serverUri, connectionContext)
                                 .whenComplete((v,t)-> {
                                     if (t == null) {
                                         log.atDebug().setMessage(() -> "New network connection result for " +
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
@@ -5,6 +5,7 @@ import io.netty.buffer.ByteBuf;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
 import io.netty.channel.ChannelInitializer;
 import io.netty.channel.ChannelOption;
 import io.netty.channel.ChannelPipeline;
@@ -53,6 +54,7 @@ public class NettyPacketToHttpConsumer implements IPacketFinalizingConsumer<Aggr
     public static final String BACKSIDE_HTTP_WATCHER_HANDLER_NAME = "BACKSIDE_HTTP_WATCHER_HANDLER";
     public static final String CONNECTION_CLOSE_HANDLER_NAME = "CONNECTION_CLOSE_HANDLER";
     public static final String SSL_HANDLER_NAME = "ssl";
+    public static final String READ_TIMEOUT_HANDLER_NAME = "readTimeoutHandler";
     public static final String WRITE_COUNT_WATCHER_HANDLER_NAME = "writeCountWatcher";
     public static final String READ_COUNT_WATCHER_HANDLER_NAME = "readCountWatcher";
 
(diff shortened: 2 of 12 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
```diff
@@ -48,11 +46,9 @@ private Key getKey(String connectionId, int sessionNumber) {
     public ClientConnectionPool(@NonNull URI serverUri,
                                 SslContext sslContext,
                                 @NonNull String targetConnectionPoolName,
-                                int numThreads,
-                                @NonNull Duration timeout) {
+                                int numThreads) {
```
gregschohn, line 49: hooray.  I never liked passing the request timeout to the connection pool

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -66,29 +68,39 @@ public class NettyPacketToHttpConsumer implements IPacketFinalizingConsumer<Aggr
...
         }
         @Override
         public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
             socketContext.close();
             super.channelUnregistered(ctx);
         }
+
+        @Override
+        public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) {
+            socketContext.close();
```
gregschohn, line 86: You should add a line beforehand for `socketContext.addTraceException(cause, true)`

gregschohn: Was the socketContext NOT being closed when there was an exception?  Does an exception skip unregistration?
I think it's safe to call close() twice, but can you check?  If it isn't, we should have a check in this class (like nulling out the socketContext)

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumerTest.java
```diff
@@ -315,6 +316,90 @@ public void testMetricCountsFor_testThatConnectionsAreKeptAliveAndShared(boolean
         Assertions.assertEquals(2, connectionsClosedCount);
     }
 
+    @ParameterizedTest
+    @CsvSource({"false", "true"})
+    @Tag("longTest")
+    public void testResponseTakesLongerThanTimeout(boolean useTls) throws Exception {
+        var testDurationBuffer = Duration.ofSeconds(10);
```
gregschohn, line 323: nit: rename this to something like "max time for test to wait for read timeout or response"?
You could also put it closer to the get() call to bind the application and value to make the intent clearer.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumerTest.java
```diff
@@ -315,6 +316,90 @@ public void testMetricCountsFor_testThatConnectionsAreKeptAliveAndShared(boolean
         Assertions.assertEquals(2, connectionsClosedCount);
     }
 
+    @ParameterizedTest
+    @CsvSource({"false", "true"})
+    @Tag("longTest")
```
gregschohn, line 321: How long will this actually take to run?  It seems like it should be < 1s.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumerTest.java
```diff
@@ -315,6 +316,90 @@ public void testMetricCountsFor_testThatConnectionsAreKeptAliveAndShared(boolean
...
+                () -> Stream.of(EXPECTED_REQUEST_STRING.getBytes(StandardCharsets.UTF_8)));
+            // Make sure finalize is triggered shortly after the response timeout
+            var aggregatedResponse = requestFinishFuture.get(responseTimeout.plus(testDurationBuffer));
+            Assertions.assertInstanceOf(ReadTimeoutException.class, aggregatedResponse.getError());
+        }
+    }
+
+    @ParameterizedTest
+    @CsvSource({"false", "true"})
+    public void testTimeBetweenRequestsLongerThanResponseTimeout(boolean useTls) throws Exception {
```
gregschohn, line 359: logging would be nice in this function to see the timestamps

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumerTest.java
```diff
@@ -315,6 +316,90 @@ public void testMetricCountsFor_testThatConnectionsAreKeptAliveAndShared(boolean
...
+                "targetPool for testTimeBetweenRequestsLongerThanResponseTimeout", 1);
+
+            var timeShifter = new TimeShifter();
+            timeShifter.setFirstTimestamp(Instant.now());
+            var sendingFactory = new ReplayEngine(
+                new RequestSenderOrchestrator(clientConnectionPool,
+                    (replaySession, ctx1) -> new NettyPacketToHttpConsumer(replaySession, ctx1, responseTimeout)),
+                new TestFlowController(), timeShifter);
+            for (int i = 0; i < 2; ++i) {
+                if (i != 0) {
```
gregschohn, line 380: nit, you could put this at the end of a while(true) loop after your i<2 test to break.

gregschohn (approved):
Thanks.  I have some minor improvements that would be nice additions for minimal cost.  Thanks for spotting this and fixing it.  I'm not sure why I doubled up on the connection close handler.

There is a side note that we might want to put a timeout on how stale a connection could be before we recycle it.  We're using a guava CacheBuilder so adding a timeout isn't too challenging.  We SHOULD already get some of that because of the timeouts that we do on source connections... _however_, if we're playing back at a slower or variable speed,  or the target has some different parameters, that might be an issue.

I'd rather invest time into managing retries well and find out how prevalent that issue is before worrying about it.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
AndreKurait: You're right. Socket context would have been closed twice with this in cases an exception reaches ConnectionClosedListenerHandler. It doesn't occur in the timeout case since that handler is after this one.

It doesn't look like .close should be called multiple times, it seems that would double count metrics. I'll do a couple changes to fix this.
1. remove the  socketContext.close from exception caught
2. change channelUnregistered -> channelInactive. Technically, unregistered can be called multiple times, but inactive is only called once on channel close

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumerTest.java
AndreKurait: The repettitions were making it take longer. I'll change that to 1 reps and it's 1 second for both tls and non-tls

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
AndreKurait: In our use, channel is only unregistered when it's closed, but there can be implementations where channel is unregistered from the event loop and re-registered again so inactive follows the contracts better

AndreKurait (commented):
Addressed all PR feedback. Thanks @gregschohn

PR #650, opened 2024-05-13
[RFS] Added backoff to RFS Metadata work leases

chelma:
### Description
* Added simple backoff to the RFS Metadata Migration work lease

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1733

### Testing
* Updated existing unit tests
* Added new unit tests

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/cms/CmsEntry.java
+++ b/RFS/src/main/java/com/rfs/cms/CmsEntry.java
@@ -1,5 +1,6 @@
 package com.rfs.cms;
 
+import com.rfs.common.RfsException;
 
 public class CmsEntry {
     public static enum SnapshotStatus {
@@ -29,6 +30,21 @@ public class CmsEntry {
         public static final int METADATA_LEASE_MS = 1 * 60 * 1000; // 1 minute, arbitrarily chosen
         public static final int MAX_ATTEMPTS = 3; // arbitrarily chosen
 
+        public static int getLeaseDurationMs(int numAttempts) {
+            if (numAttempts > MAX_ATTEMPTS) {
+                throw new CouldNotFindNextLeaseDuration("numAttempts=" + numAttempts + " is greater than MAX_ATTEMPTS=" + MAX_ATTEMPTS);
+            } else if (numAttempts < 1) {
+                throw new CouldNotFindNextLeaseDuration("numAttempts=" + numAttempts + " is less than 1");
+            }
+            return METADATA_LEASE_MS * numAttempts; // Arbitratily chosen algorithm
+        }
+
+        // TODO: We should be ideally setting the lease expiry using the server's clock, but it's unclear on the best
+        // way to do this.  For now, we'll just use the client's clock.
+        public static String getLeaseExpiry(long currentTime, int numAttempts) {
+            return Long.toString(currentTime + getLeaseDurationMs(numAttempts));
+        }
+
         public final MetadataStatus status;
         public final String leaseExpiry;
         public final Integer numAttempts;
@@ -37,6 +53,12 @@ public class CmsEntry {
             this.status = status;
             this.leaseExpiry = leaseExpiry;
             this.numAttempts = numAttempts;
+        }
+    }
+
+    public static class CouldNotFindNextLeaseDuration extends RfsException {
+        public CouldNotFindNextLeaseDuration(String message) {
+            super("Could not find next lease duration.  Reason: " + message);
         }
     }
 }
--- a/RFS/src/main/java/com/rfs/cms/OpenSearchCmsEntry.java
+++ b/RFS/src/main/java/com/rfs/cms/OpenSearchCmsEntry.java
@@ -1,6 +1,6 @@
 package com.rfs.cms;
 
-import java.util.Date;
+import java.time.Instant;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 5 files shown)


RFS/src/main/java/com/rfs/cms/CmsEntry.java
```diff
@@ -29,6 +30,21 @@ public static class Metadata {
         public static final int METADATA_LEASE_MS = 1 * 60 * 1000; // 1 minute, arbitrarily chosen
         public static final int MAX_ATTEMPTS = 3; // arbitrarily chosen
 
+        public static int getLeaseDurationMs(int numAttempts) {
```
peternied, line 33: As written it isn't clear where numAttempts is tracked, looks like individual callers - so if working 1, 2, and 3 check on the lease currently associated with worker 4 they will get different durations back depending on which attempt they have to pick up the work?

Shouldn't the number of attempts be a field on the metadata object?

chelma: It is a field on the `CmsEntry.Metadata` class; scroll down a bit.  Though maybe your underlying concern is why this is a class method vs. an instance method?  In my mind, an instance of the class probably shouldn't be calculating the lease duration for a new instance of the class, so it felt more natural to move this to be a class method.  There's different ways this could have been done, such as a static factory method that just returns a new `CmsEntry.Metadata` with an updated lease, but the approach I took didn't seem to unreasonable.  

Thoughts?

peternied: @chelma I don't understand the relationship between the number of attempts and the duration of the lease.

Shouldn't the duration of the lease always be the same and the lease is refreshed after a failed attempt?

For example, let's say the worker has a lot of work to do. It seems to make sense that it would periodically update the timestamp of the lease. It doesn't seem to matter the nature of the work is, only that the worker node indicates it is responsible and no other worker should pick up the work.

Is this the model you were thinking for how these leases work ?

chelma: Hmm, not quite.  There's two pieces of the puzzle here:
* Is this work item achievable?
* Is someone still making progress on it?

Having workers continually renew the leases doesn't help solve the first piece, so you can end up in situations where the work will never get finished and that causes higher level system design issues.  As a result, the approach we chose to start with is as follows:
* When an RFS Worker picks up an item, it has a limited window to accomplish the entire task based on the initial lease duration
* If the RFS Worker fails to accomplish the task in that timeframe, it could indicate a problem with the Worker **or** the task.
* If an RFS Worker detects its lease has expired, it immediately kills itself, meaning that if the problem was with the worker, we'll get a new one.
* The next RFS Worker to pick up the task adds a backoff to the lease duration in case the problem was with the task, because it doesn't know why a second attempt is needed.
* We're guaranteed that all tasks will end, either successfully or unsuccessfully, but we should never end up with "forever-tasks"

chelma: In the future, I want the RFS Scaler to ensure that the RFS Workers associated w/ expired work items have been reaped, as well.

peternied: You mentioned a number of examples, I'm not clear on when a lease becomes invalid, can you tie together which examples relate to the duration of the lease being increased based on attempt number?

It sounds like you are saying an attempt number is only used in computation of the active worker holding the lease.  This is fine, but this seems like it shouldn't be part of the CMS, but as part of an active workers state machine flow.  Maybe you can take the following example of the state and confirm or modify how this would work?


### Metadata state over time
I'm arbitrarily moving forward in time and capturing the state of the store 

#### T=0
```
Task1={metadata={status=in_progress, worker=1, lease={t=5}}, ...} // Worker 1 has started, at t+5 the lease will expire  
Task2={metadata={status=in_progress, worker=2, lease={t=5}}, ...} // Worker 2 has started, at t+5 the lease will expire  
Task3={metadata={status=not_started, ...} // Not been picked up
```

...
#### T=4
```
Task1={metadata={status=in_progress, worker=1, lease={t=5}}, ...} // Worker 1 is still working on this task
Task2={metadata={status=completed, worker=2}, ...} // Worker 2 completed this item
Task3={metadata={status=in_progress, worker=2, lease={t=9}}, ...} Worker 2 has started, at t+5 the lease will expire  
```

#### T=5
```
Task1={metadata={status=in_progress, worker=1, lease={t=???}, attempt=???}, ...} // ??? 
Task2={metadata={status=completed, worker=2}, ...} // Worker 2 completed this item
Task3={metadata={status=in_progress, worker=2, lease={t=9}}, ...} Worker 2 is still working on the task
```

Outstanding questions?
- If `worker=1` needs another attempt on `Task1`, what is updated in the metadata store?
- When does `worker=2` evict the lease grabbed on `Task1`, how does it take it over?

peternied: Circling back on this topic, the way the time is computed is up to the worker, and the lease doesn't get updated.  If the lease has expired, its up to the worker that picks up the work to make the appropriate updates.

Where this circles back to the code definition is that the code that determines that lease time itself is not well positioned to have the information need to calculate how long things will take - this will likely need to be refactored in the future as this functionality is expanded upon.

peternied (approved):

PR #651, opened 2024-05-13
Extend E2E script to allow CDK customization

lewijacn:
### Description
This change includes the following changes to allow more customization with our awsE2ESolutionSetup.sh script as well as minor bug fixes:
* Allow option to migration CDK to provide a file path for loading CDK context and add tests cases around this
* Update e2e script to load CDK context for both source and target via a provided context file
* Add `clean-up-all` option to e2e script for cleaning up all deployed resources after running
* Modify spots in e2e script to use absolute path from single variable instead of on-the-fly relative paths
* Bug fix around Custom Resource for getting MSK broker endpoints which still tried to retrieve these endpoints on 'Delete' lambda events

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1690

### Testing
CDK tests and manual AWS testing

### Check List

### DIFF
--- a/test/awsE2ESolutionSetup.sh
+++ b/test/awsE2ESolutionSetup.sh
@@ -3,6 +3,12 @@
 # Note: As this script will deploy an E2E solution in AWS, it assumes all the dependencies of the migration solution (e.g. aws cli, cdk cli,
 # configured aws credentials, git, java, docker) as well as 'jq'
 
+script_abs_path=$(readlink -f "$0")
+TEST_DIR_PATH=$(dirname "$script_abs_path")
+ROOT_REPO_PATH=$(dirname "$TEST_DIR_PATH")
+TMP_DIR_PATH="$TEST_DIR_PATH/tmp"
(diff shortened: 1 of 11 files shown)


test/awsE2ESolutionSetup.sh
```diff
@@ -3,14 +3,20 @@
 # Note: As this script will deploy an E2E solution in AWS, it assumes all the dependencies of the migration solution (e.g. aws cli, cdk cli,
```
chelma, line 3: No action necessarily required, just a question for you - are you having trouble with maintaining/extending/debugging this script?  Have we considered migrating its functionality to something like Python, etc?

chelma (approved):
Overall, makes sense.  I have one questions but it's not a blocker.

test/awsE2ESolutionSetup.sh
lewijacn: I'm 1000% onboard with migrating this script to python, so much so that I let AI take a swing at it the other day to see how well it could do with it, and with some parameters it did a decent job and might be worth looking into as a reference

PR #656, opened 2024-05-15
Top level directory refactoring

gregschohn:
### Description

TODO: I will update jenkins settings on how to invoke gradle since gradlew has moved after/as this is merged.

Make gradle work at the top-level of the repo.  RFS and TrafficCapture settings have been unified into gradle settings at the root.  coreUtilities has also been migrated to the root so that it's easily visible and in a logical spot.  gradlew has also been moved.  I recommend installing the function below into .zshrc files so that 'agradle' can be invoked from any directory.

```
function agradle {                                     
  local dir=$(pwd)

  while [[ "$dir" != "/" ]]; do
    if [[ -x "$dir/gradlew" ]]; then
      "$dir/gradlew" "$@"
      return
    fi
    dir=$(dirname "$dir")
  done

  echo "No gradlew script found in the ancestor directories"
}
```

* Category Refactoring
* Why these changes are required? To make our projects more consistent and easier to develop.  More immediately, to easily pull in the coreUtilities instrumentation classes.
* What is the old behavior before changes and new behavior after changes?  RFS and TrafficCapture were disjoint gradle projects with separate settings.  Now one gradle project can understand both of them at the same time.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1738

### Testing
Manual tests w/ C&R w/ OSB.  All unit tests pass.

### Check List


settings.gradle
```diff
@@ -0,0 +1,65 @@
...
+    include projectName
+
+    project(projectName).projectDir = dir
+
+    for (File subdir : dir.listFiles()) {
+        addSubProjects(projectName, subdir)
+    }
+}
+
+rootProject.name = 'MigrationsAssistant'
```
peternied, line 40: Seems like this should be called 'OpenSearch Migration' unless we are thinking of changing the repo name?

settings.gradle
```diff
@@ -0,0 +1,65 @@
...
+include 'TrafficCapture:captureKafkaOffloader'
```
peternied, line 41: Can we programmatically include these projects rather than list them?

settings.gradle
```diff
@@ -0,0 +1,65 @@
...
+include 'TrafficCapture:captureOffloader'
+include 'TrafficCapture:captureProtobufs'
+include 'TrafficCapture:nettyWireLogging'
+include 'TrafficCapture:trafficCaptureDependencies'
```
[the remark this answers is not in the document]
gregschohn: They should and soon, but I wanted to leave the people working on RFS currently to make that call.  
I'm hoping that this commit/merge can open the door to changes like that but also not require them to be made immediately.

settings.gradle
gregschohn: How are you thinking that they could be programmatically included?

peternied: Let me rephrase - do we have to enumerate each project in this way? Seems like we have a method to find all projects with `addSubProjects()` that could be used.

settings.gradle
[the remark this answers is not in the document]
peternied: What do you think about moving the dependencies project to the top level now to minimize future changes on the projects under TrafficCapture.  Then how/when RFS moves it does on its own timeline?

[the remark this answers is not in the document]
gregschohn: I'll move it up, but I'll need a new name other than trafficCaptureDependencies.  Does "commonDependencyVersionConstraints" work?  commonDependencies sounds like the project should be logging, coreUtilities, etc.  But I don't like the idea that we need any dependencies at all in any of our projects, hence the reason that I've added the extra words.

QQ: If these tools were in separate repos, would you still want to see common dependencies locked across repos/projects?  Would you have a replayer project, capture project, & RFS project all take a dependency on a common dependency project?  I think that it would still make sense, but it does (literally & intentionally) bind constraints on separate projects.

chelma:
@gregschohn  LGTM, but - what does the gradle command look like to run the various components?  Is there namespacing or something?  Sorry for the noob question...

.github/workflows/e2eTest.yml
```diff
@@ -30,7 +30,7 @@ jobs:
           cache-read-only: false
 
       - name: Start Docker Solution
-        run: ./gradlew dockerSolution:ComposeUp -x test --scan --info --stacktrace
+        run: ../gradlew dockerSolution:ComposeUp -x test --scan --info --stacktrace
         working-directory: TrafficCapture
```
peternied, line 34: I'd rather delete this line than use `..` above

peternied: duplicated in a couple of other places as well

RFS/README.md
```diff
@@ -74,18 +74,18 @@ gradle run --args='-n global_state_snapshot --source-host $SOURCE_HOST --source-
 ### Using Docker
 RFS has support for packaging its java application as a Docker image by using the Dockerfile located in the `RFS/docker` directory. This support is directly built into Gradle as well so that a user can perform the below action, and generate a fresh Docker image (`migrations/reindex_from_snapshot:latest`) with the latest local code changes available.
 ```shell
-./gradlew buildDockerImages
+../gradlew buildDockerImages
```
peternied, line 77: What about indicating moving to the project root, or not including the path directive at all?
```suggestion
cd {PROJECT_ROOT}
./gradlew buildDockerImages
```

RFS/build-preloaded-source-image.gradle
```diff
@@ -0,0 +1,115 @@
...
+    dependsOn createInitialElasticsearchContainer
+    targetContainerId createInitialElasticsearchContainer.getContainerId()
+}
+
+task createClientContainer(type: DockerCreateContainer) {
+    dependsOn startInitialElasticsearchContainer, buildDockerImage_trafficGenerator
+    targetImageId 'migrations/osb_traffic_generator:latest'
+    containerName = "traffic-generator-container-${uniqueId}"
+    hostConfig.network = myNetworkName
+    //cmd = ['skip_dataset']
```
peternied, line 45: Remove?

RFS/build-preloaded-source-image.gradle
```diff
@@ -0,0 +1,115 @@
...
+
+// Task to commit the source container to create a new image that will include the loaded data
+def sourceContainerCommitTask = task commitSourceContainer() {
+    dependsOn waitClientContainer
+
+    doLast {
+        def client = createDockerClient()
+        def containerId = createInitialElasticsearchContainer.getContainerId().get()
+
+        println "Committing container ${containerId} to image my-new-primary-image:latest"
```
peternied, line 67: parameterize this so it doesn't get out of date?  Note might need slightly more refactoring
```suggestion
        println "Committing container ${containerId} to image ${image}:${version}"
```

TrafficCapture/README.md
```diff
@@ -60,8 +60,8 @@ Partial example output of the OpenSearch Benchmark tool:
 Running opensearch-benchmark w/ 'geonames' workload...
 
    ____                  _____                      __       ____                  __                         __
-/ __ \____  ___  ____ / ___/___  ____ ___________/ /_     / __ )___  ____  _____/ /_  ____ ___  ____ ______/ /__
-/ / / / __ \/ _ \/ __ \\__ \/ _ \/ __ `/ ___/ ___/ __ \   / __  / _ \/ __ \/ ___/ __ \/ __ `__ \/ __ `/ ___/ //_/
+  / __ \____  ___  ____ / ___/___  ____ ___________/ /_     / __ )___  ____  _____/ /_  ____ ___  ____ ______/ /__
+ / / / / __ \/ _ \/ __ \\__ \/ _ \/ __ `/ ___/ ___/ __ \   / __  / _ \/ __ \/ ___/ __ \/ __ `__ \/ __ `/ ___/ //_/
```
peternied, line 64: Revert these lines?

gregschohn: This corrects the output from OSB. It uses an italicized font and requires those spaces to make the big characters legible

.github/workflows/e2eTest.yml
chelma: ...now I'm curious what the difference is!

chelma: Looked into it, not sure there is a practical difference here

RFS/README.md
gregschohn: what if I added this to each of the three READMEs or the top README?  It's long, but it works all through the tree.

`agradle() { local dir=$(pwd); while [[ "$dir" != "/" ]]; do [[ -x "$dir/gradlew" ]] && "$dir/gradlew" "$@" && return; dir=$(dirname "$dir"); done; echo "No gradlew script found in the ancestor directories"; }`

RFS/build-preloaded-source-image.gradle
gregschohn: removed the print altogether

RFS/README.md
peternied: This seems like it creates more complexity than ease of use.  IMO its much simpler to force everything to be done from the root of the project, at least when you can get away with a single purpose oriented repo

settings.gradle
gregschohn: This might be a good idea, but I have enough changes in this PR.  We can revisit this behavior change later.

RFS/README.md
gregschohn: I agree that the agradle thing is scary and off-putting.  I appreciate that '..' is unambiguous and won't affect the user's current environment as `cd` would.

settings.gradle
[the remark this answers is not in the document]
peternied: What do you think about a shorter `dependencyConstraints`? IMO when/if we get to another repo that might be a good time to reconsider the naming convention. 

I'm not too concerned when by looking at the project gradle its clear there is no actual source code and its consistently a single thing - a big list of dependency versions

gregschohn:
In response to @chelma [here](https://github.com/opensearch-project/opensearch-migrations/pull/656#issuecomment-2120570104)... if you’re in the same directory as before, nothing changes other than the path/command to run gradlew - for which I use agradle.  However, now you can also build from anywhere in the tree by suffixing with the fully-qualified path.  i.e. :RFS:build from w/in TrafficCapture.

chelma (approved):
Thanks - LGTM!

.github/workflows/e2eTest.yml
peternied: Note, I'm happy to revisit this post merge myself if I've got other ideas

RFS/README.md
peternied: Resolving this thread, I'm happy to revisit this post merge myself if I've got other ideas

peternied (approved):
Looks good to merge, but it seems like we've got a failure with the link checker finding.  Can you exclude/fix this issue before merging?

```
### Errors in deployment/cdk/opensearch-service-migration/package-lock.json

* [ERR] [https://paypal.me/naturalintelligence](https://paypal.me/naturalintelligence) |
    Failed: Network error: error sending request for url (https://www.paypal.me/naturalintelligence)
```

PR #660, opened 2024-05-15
Add code hygiene for python

peternied:
### Description
Add code hygiene checks for python, broken apart by the different projects within the repo.

### Testing
New GitHub action is triggered on any changes via a pull request or push to the repository.

### Check List

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -31,7 +31,7 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           files: cluster_migration_core/coverage.xml
-  setup-python:
+  code-hygiene:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


.github/workflows/python-tests.yml
```diff
@@ -35,3 +31,42 @@ jobs:
         uses: codecov/codecov-action@v4
         with:
           files: cluster_migration_core/coverage.xml
+  code-hygiene:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        location:
```
peternied, line 39: Looks like we might have some dead code, should we delete some of these?

.github/workflows/python-tests.yml
```diff
@@ -35,3 +31,42 @@ jobs:
...
+        working-directory: ${{ matrix.location }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+      - run: python3 -m pip install -r dev-requirements.txt
+        continue-on-error: true # Not all projects have dev-requirements.txt, safe to ignore
+      - run: python3 -m pip install -r requirements.txt
+        continue-on-error: true # Not all projects have requirements.txt, safe to ignore
```
peternied, line 56: Maybe replace these with a pipenv file?

peternied (commented):
What do folks think?

.github/workflows/python-tests.yml
```diff
@@ -35,3 +31,42 @@ jobs:
...
+      - run: python3 -m pip install isort flake8 mypy
+      - id: isort
+        continue-on-error: true # Success/failure checked in last job step
+        run: python3 -m isort --check .
+      - id: flake8
+        continue-on-error: true # Success/failure checked in last job step
+        run: python3 -m flake8 .
+      - id: mypy
+        continue-on-error: true # Success/failure checked in last job step
+        run: python3 -m mypy .
```
chelma, line 66: I looked into mypy a bit and think I'm in favor of trying it out.  I'd love to take a gradual approach here if possible.  Maybe you have thoughts on the topic, but what about the following high level approach
* Add mypy to the repo's Python packages
* Make it easy to run mypy and other style checks locally before submitting your PR
* Update READMEs as appropriate to document how to run the checks locally
* Configure mypy to:
    * Error if a type annotation exists, but is wrong
    * Warn if a type annotation is missing, but not count it as a failure
    * Ignore missing type annotations from third-party, imported Python modules

peternied: Assuming all of the projects that I'm running this job over are code we care about - I'll fix all the errors to get us into a passing state before merging.  How we handle different errors, such as missing type annotations can be done on a case-by-case basis.

> Update READMEs as appropriate to document how to run the checks locally

For which README(s) should be updated, do we do this at the project level or individual project level?

peternied:
Since its been a while since I've put attention on this PR I'm going to close it out; if/when I get time I'll reopen this.

PR #678, opened 2024-05-23
Rfs instrumentation

gregschohn:
### Description
Instruments RestClient within RFS so that every call has a context.  Instrument work coordination so that every low-level and high-level interaction has a scope and metrics are emitted appropriately.

* Category New feature/Refactoring/OE
* Why these changes are required? To get better observability into RFS
* What is the old behavior before changes and new behavior after changes? Now HTTP calls in RFS emit spans and metrics.

*Still TODO* - verify that the metrics for the work coordinator are as expected.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1738

### Testing
Limited manual testing.  RFS doesn't have much test infrastructure yet

### Check List


RFS/src/main/java/com/rfs/tracing/RootRfsContext.java
```diff
@@ -0,0 +1,51 @@
...
+import io.opentelemetry.api.OpenTelemetry;
+import org.opensearch.migrations.tracing.IContextTracker;
+import org.opensearch.migrations.tracing.RootOtelContext;
+
+public class RootRfsContext extends RootOtelContext {
+    public static final String SCOPE_NAME = "rfs";
+    public final RfsContexts.CreateSnapshotContext.MetricInstruments snapshotInstruments;
+    public final RfsContexts.ClusterMetadataContext.MetricInstruments metadataMetrics;
+    public final RfsContexts.MigrateIndexTemplateContext.MetricInstruments indexTemplateInstruments;
+    public final RfsContexts.MigrateIndexTemplateContext.MetricInstruments genericRequestInstruments;
```
chelma, line 12: Why two of this type?

chelma (commented):
Spent some time looking things over and left some comments, but probably better to discuss verbally.  Had trouble piecing together the bigger picture here.

RFS/src/main/java/com/rfs/tracing/RootRfsContext.java
gregschohn: fixed in a later commit

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
```diff
@@ -0,0 +1,411 @@
...
+        }
+
+        @Override
+        public IRfsContexts.ICheckedIdempotentPutRequestContext createMigrateIndexTemplateContext() {
+            return new CheckedIdempotentPutRequestContext(rootInstrumentationScope, this,
+                    "createGetSnapshotContext");
+        }
+    }
+
+    public static class MigrateIndexTemplateContext
```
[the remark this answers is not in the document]
gregschohn: It's what you asked me to instrument in task MIGRATIONS-1738

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
```diff
@@ -0,0 +1,411 @@
...
+            return new MetricInstruments(meter, ACTIVITY_NAME);
+        }
+
+        @Override
+        public MetricInstruments getMetrics() {
+            return getRootInstrumentationScope().metadataMetrics;
+        }
+
+        @Override
+        public IRfsContexts.IIndexTemplateContext createMigrateLegacyIndexTemplateContext() {
```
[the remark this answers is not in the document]
gregschohn: I did a `s/IndexTemplateContext/TemplateContext/` in a later commit

RFS/src/main/java/com/rfs/tracing/IRfsContexts.java
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    public interface IClusterMetadataContext extends IScopedInstrumentationAttributes {
+        String ACTIVITY_NAME = ActivityNames.MIGRATE_METADATA;
+
+        IIndexTemplateContext createMigrateLegacyIndexTemplateContext();
+
+        ICheckedIdempotentPutRequestContext createComponentTemplateContext();
+
+        ICheckedIdempotentPutRequestContext createMigrateIndexTemplateContext();
```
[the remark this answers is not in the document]
gregschohn: Different how?  Those calls do a get, then a put, so there are potentially multiple requests, the IChecked...Context can create request scopes as needed for those two kinds of calls.

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
```diff
@@ -0,0 +1,411 @@
...
+            return new CheckedIdempotentPutRequestContext(rootInstrumentationScope, this,
+                    "createComponentTemplateContext");
+        }
+
+        @Override
+        public IRfsContexts.ICheckedIdempotentPutRequestContext createMigrateIndexTemplateContext() {
+            return new CheckedIdempotentPutRequestContext(rootInstrumentationScope, this,
+                    "createGetSnapshotContext");
+        }
+    }
```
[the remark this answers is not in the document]
gregschohn: I'm not sure what your enumeration is... I'll admit that having methods to create a context that is in reference to an HTTP request that is creating something is confusing.  Given that this code is already changing underfoot and that the task was to establish the instrumentation pattern, this is something that can be picked up later, right?

RFS/src/main/java/com/rfs/RunRfsWorker.java
```diff
@@ -91,6 +97,14 @@ public static class Args {
 
         @Parameter(names = {"--log-level"}, description = "What log level you want.  Default: 'info'", required = false, converter = Logging.ArgsConverter.class)
         public Level logLevel = Level.INFO;
+
+
+        @Parameter(required = false,
```
peternied, line 102: Blocking; I know how to visualize the output of these instrumentation, can you update the readme or update the code to allow for a very basic log things to files?  I am OK if the dump isn't a file but a running Jager instance as long as its easy for me to spin up and connect.

RFS/src/test/java/com/rfs/tracing/TestContext.java
```diff
@@ -0,0 +1,31 @@
+package com.rfs.tracing;
+
+import lombok.Getter;
+import org.opensearch.migrations.tracing.BacktracingContextTracker;
+import org.opensearch.migrations.tracing.IContextTracker;
+import org.opensearch.migrations.tracing.InMemoryInstrumentationBundle;
+
+public class TestContext extends RootRfsContext {  @Getter
```
peternied, line 8: There are a couple of unused methods here - lets either remove them or lets refactor to converge on a single TestContext factory.

RFS/src/test/java/com/rfs/integration/SnapshotStateTest.java
```diff
@@ -60,6 +61,7 @@ public void tearDown() throws Exception {
     @Test
     public void SingleSnapshot_SingleDocument() throws Exception {
         // Setup
+        final var testContext = TestContext.noOtelTracking();
```
peternied, line 64: Nit; this should be refactored into the SimpleRestoreFromSnapshot_ES_7_10.  I'll do this later - thanks for merging this in.

RFS/src/test/java/com/rfs/common/RestClientTest.java
```diff
@@ -0,0 +1,96 @@
...
+
+@Slf4j
+class RestClientTest {
+    @Test
+    public void testGetEmitsInstrumentation() throws Exception{
+        var rootContext = TestContext.withAllTracking();
+        try (var testServer = SimpleNettyHttpServer.makeServer(false, null,
+                this::makeResponseContext)) {
+            var restClient = new RestClient(new ConnectionDetails("http://localhost:" + testServer.port, null, null));
+            try (var topScope = rootContext.createSnapshotCreateContext()) {
```
peternied, line 31: Creating the scope in this case doesn't seem as useful since you only operating inside a single scope.  If we wanted to 'prove' that the scopes were propigated, lets add a second scope where postAsync is used that could be differentiate on, what do you think?

RFS/src/main/java/com/rfs/RunRfsWorker.java
```diff
@@ -91,6 +97,14 @@ public static class Args {
...
+                names = {"--otelCollectorEndpoint"},
```
peternied, line 103: Looks like we are using kabob casing

```suggestion
                names = {"--otel-collector-endpoint"},
```

RFS/src/main/java/com/rfs/version_os_2_11/GlobalMetadataCreator_OS_2_11.java
```diff
@@ -91,7 +99,8 @@ protected void createComponentTemplates(GlobalMetadataData_OS_2_11 globalMetadat
 
                 logger.info("Setting Component Template: " + templateName);
                 ObjectNode settings = (ObjectNode) templates.get(templateName);
-                client.createComponentTemplate(templateName, settings);
+                client.createComponentTemplate(templateName, settings,
+                        context.createComponentTemplateContext()); // purposefully make a new one each loop iteration
```
peternied, line 103: What do you think of rewriting with try with resource pattern rather than a commment? E.g:

```java
for (String templateName : templateWhitelist) {
    // ...
    try (final var context = context.createComponentTemplateContext()) {
        client.createComponentTemplate(templateName, settings, context);
    } 
}
```

peternied: After revisiting I would recommend an alternative to make the method `createComponentTemplate(...)`  responsiblity to create the context

RFS/src/main/java/com/rfs/tracing/RootRfsContext.java
```diff
@@ -0,0 +1,51 @@
...
+public class RootRfsContext extends RootOtelContext {
+    public static final String SCOPE_NAME = "rfs";
+    public final RfsContexts.CreateSnapshotContext.MetricInstruments snapshotInstruments;
+    public final RfsContexts.ClusterMetadataContext.MetricInstruments metadataMetrics;
+    public final RfsContexts.MigrateTemplateContext.MetricInstruments indexTemplateInstruments;
+    public final RfsContexts.GenericRequestContext.MetricInstruments genericRequestInstruments;
+    public final RfsContexts.CheckedIdempotentPutRequestContext.MetricInstruments getTwoStepIdempotentRequestInstruments;
+    public final RfsContexts.CreateIndexContext.MetricInstruments createIndexInstruments;
+    public final RfsContexts.DocumentReindexContext.MetricInstruments documentReindexInstruments;
+    public final RfsContexts.WorkingStateContext.MetricInstruments workingStateUpdateInstruments;
```
peternied, line 16: 1) Why are all of these made and accessible outside the root context, I'm not sure what the case for access is?
2) Shouldn't they be created and passed into `create***Context(...)`?  Note if this is about singular creation, lets use a lazy supplier

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
```diff
@@ -0,0 +1,453 @@
...
+            return new GenericRequestContext(rootInstrumentationScope, this,
+                    "DocumentReindexContext.createRefreshContext");
+        }
+
+    }
+
+    public static class WorkingStateContext
+            implements IRfsContexts.IWorkingStateContext {
+        @Getter
+        @NonNull
```
peternied, line 385: As a project go all in or out on these annotations.  Want to raise an issue/ticket on this?

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
```diff
@@ -0,0 +1,453 @@
...
+
+    private RfsContexts() {}
+
+    public static final String COUNT_UNITS = "count";
+
+    public static class GenericRequestContext
+            extends BaseSpanContext<RootRfsContext>
+            implements IRfsContexts.IRequestContext {
+
+        public static final AttributeKey<String> CALL_TYPE_ATTR = AttributeKey.stringKey("callType");
```
peternied, line 30: What is a call type, what does it represent?

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
```diff
@@ -0,0 +1,453 @@
...
+
+        @Override
+        public IScopedInstrumentationAttributes getEnclosingScope() {
+            return null;
+        }
+
+        @Override
+        public IRfsContexts.IRequestContext createRegisterRequest() {
+            return new GenericRequestContext(rootInstrumentationScope, this, "createRegisterRequest");
+        }
```
peternied, line 171: Maybe this is for flexiblity, but the method that calls `createRegisterRequest()` is registerSnapshotRepo, seems like the name should be from the calling method annotated onto the GenericRequestContext.

What I would consider a better alternative is to have the caller provide the label, e.g.
```
createRegisterRequest(...) {
     ...
     
     try (final var requestContext = context.createRequestContext("createRegisterRequest")) {
        // Use it
     }
     ...
}
```

It would be even better if this was automatically using the callstack

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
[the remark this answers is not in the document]
peternied: @gregschohn this is something I called out in another comment.  If the the code changes, should the context name change with it, or do we need to be intentional about the changes to telemetry?

Note; I think the answer is it depends.
- If there is a specific contract to uphold, consider if there is an external system that expects `RFS.DocumentProcessor.DocsRead.Count` we should not break that system even if we rename DocumentProcessor to LuceneDocumentsProcessor
- If there is no contract we should limit the amount of 'legacy' that can build up in our systems.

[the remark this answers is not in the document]
peternied: For the current part of the lifecycle I would rather we have less flexiblity and less to cleanup as we quickly iterate on this first productized version, we could always come back later.

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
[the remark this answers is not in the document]
peternied: @chelma @gregschohn do we need to revisit how/what these context mean?

peternied (changes requested):

RFS/src/main/java/com/rfs/RunRfsWorker.java
peternied: Resolved by this PR https://github.com/gregschohn/opensearch-migrations/pull/4

peternied: Since I was able to create that PR I do not consider this to be a blocking issue anymore - resolving this thread.

peternied (approved):
@gregschohn Thanks for the conversation yesterday about the outstanding comments, I'll leave it to you if you'd prefer to merge or address any of the comment threads.

I created the following PRs against this branch on your fork, feel free to merge or close - but I'd value your thoughts

- https://github.com/gregschohn/opensearch-migrations/pull/3
- https://github.com/gregschohn/opensearch-migrations/pull/4
- https://github.com/gregschohn/opensearch-migrations/pull/5

RFS/src/main/java/com/rfs/tracing/BaseRootRfsContext.java
```diff
@@ -0,0 +1,19 @@
+package com.rfs.tracing;
+
+import io.opentelemetry.api.OpenTelemetry;
+import org.opensearch.migrations.tracing.IContextTracker;
+import org.opensearch.migrations.tracing.RootOtelContext;
+
+public class BaseRootRfsContext extends RootOtelContext {
```
peternied, line 7: We've got many stand alone apps that are all under the migration space, what about MigrationAppContext? If we are following the model that everything should be shared, shouldn't this migrate out into the core utils space?

MetadataMigration/src/main/java/com/rfs/MetadataMigration.java
```diff
@@ -78,6 +84,13 @@ public static class Args {
             + " This can be useful for migrating to targets which use zonal deployments and require additional replicas to meet zone requirements.  Default: 0")
             , required = false)
         public int minNumberOfReplicas = 0;
+
+        @Parameter(required = false,
+                names = {"--otelCollectorEndpoint"},
+                arity = 1,
+                description = "Endpoint (host:port) for the OpenTelemetry Collector to which metrics logs should be" +
```
AndreKurait, line 91: do you mean scheme:host:port?

gregschohn: We pass the argument onto [this call](https://javadoc.io/doc/io.opentelemetry/opentelemetry-exporter-otlp/0.12.0/io/opentelemetry/exporter/otlp/OtlpGrpcSpanExporter.Builder.html#setEndpoint(java.lang.String)) in Otel.  It simply says "Sets the OTLP endpoint to connect to. Optional, defaults to "localhost:4317".

Since the endpoint runs OTLP, specifying http or https is probably incorrect.  We should remove them from our docs and our deployment code (& double check).

CreateSnapshot/src/main/java/com/rfs/CreateSnapshot.java
```diff
@@ -65,6 +71,13 @@ public static class Args {
                 required = false,
                 description = "The maximum snapshot rate in megabytes per second per node")
         public Integer maxSnapshotRateMBPerNode;
+
+        @Parameter(required = false,
```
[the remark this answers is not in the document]
peternied: Also lets use a standard format for arguments, in this case `--otel-collector-endpoint`

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -140,12 +153,21 @@ public static void main(String[] args) throws Exception {
...
         validateArgs(arguments);
 
+
+        var rootWorkCoordinationContext = new RootWorkCoordinationContext(
+                RootOtelContext.initializeOpenTelemetryWithCollectorOrAsNoop(arguments.otelCollectorEndpoint, "docMigrationCoordination"),
+                new CompositeContextTracker(new ActiveContextTracker(), new ActiveContextTrackerByActivityType()));
+        var rootDocumentContext = new RootDocumentMigrationContext(
+                RootOtelContext.initializeOpenTelemetryWithCollectorOrAsNoop(arguments.otelCollectorEndpoint, "docMigration"),
+                new CompositeContextTracker(new ActiveContextTracker(), new ActiveContextTrackerByActivityType()),
+                rootWorkCoordinationContext);
```
peternied, line 163: There are multiple `root` contexts at the same time?  Lets drop the word root unless it means something, `WorkCoordinationContext` and `DocumentMigrationContext`

RFS/src/main/java/org/opensearch/migrations/snapshot/creation/tracing/CreateSnapshotContext.java
```diff
@@ -0,0 +1,61 @@
+package org.opensearch.migrations.snapshot.creation.tracing;
```
peternied, line 1: This extra layer `snapshot.creation` doesn't make much sense to me.  If this package names are weird, its because we are working backward from functionality in an executable in a shard library space rather treating this as a shared library with a clear sense of what context is required at each level.

I'd rather have generic IContext use in the library and concrete implementation in the executable project so we can keep this clearer.  What do you think?

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -192,11 +214,12 @@ public static DocumentsRunner.CompletionStatus run(Function<Path,LuceneDocuments
...
                                                        SnapshotShardUnpacker.Factory unpackerFactory,
-                                                       long maxShardSizeBytes)
+                                                       long maxShardSizeBytes,
+                                                       RootDocumentMigrationContext rootDocumentContext)
             throws IOException, InterruptedException, NoWorkLeftException {
         var scopedWorkCoordinator = new ScopedWorkCoordinator(workCoordinator, leaseExpireTrigger);
-        confirmShardPrepIsComplete(indexMetadataFactory, snapshotName, indexAllowlist, scopedWorkCoordinator);
-        if (!workCoordinator.workItemsArePending()) {
+        confirmShardPrepIsComplete(indexMetadataFactory, snapshotName, indexAllowlist, scopedWorkCoordinator, rootDocumentContext);
+        if (!workCoordinator.workItemsArePending(rootDocumentContext.getWorkCoordinationContext()::createItemsPendingContext)) {
```
peternied, line 222: This is hacky, we've got a instance that should accept a context and internally it can create the work item pending context.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -208,13 +231,15 @@ public static DocumentsRunner.CompletionStatus run(Function<Path,LuceneDocuments
                     }
                     return shardMetadata;
                 },
-                unpackerFactory, readerFactory, reindexer).migrateNextShard();
+                unpackerFactory, readerFactory, reindexer)
+                .migrateNextShard(rootDocumentContext::createReindexContext);
```
peternied, line 235: The documents runner to have a context passed into it, no?

Also the way `DocumentsRunner` is written is hard to understand, lets reduce the inlining

peternied (commented):
Quick hit of comments, still reviewing

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
gregschohn: Why is it hacky?  The lifecycle of the context should be controlled by the work coordinator.  It seems more hacky to make every call to the work coordinator be responsible for creating and tearing down contexts around singly-scoped calls.

RFS/src/main/java/org/opensearch/migrations/snapshot/creation/tracing/CreateSnapshotContext.java
gregschohn: I plan on moving everything in this directory to the executable project.  I'm doing the class refactorings now, but holding off on the file moves because additional files will need to move around rather than just the ones that I've created here.  I can do that as a fast-follow, but I'd like to do it in such a way to minimize conflicts.

As per the package name - "snapshot.creation" didn't sit completely at ease with me.  I wasn't sure what this package should be called longer term (it definitely shouldn't be 'com.rfs', nor should it likely go under the top-level "org.opensearch.migrations" - I'm open to other suggestions.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
gregschohn: Root does have meaning - the root context is a constructor and the root of the scope-trees for a given component.  Every scope needs a root, directly or indirectly.  From within the work coordination code (which should be considered its own independent component), there needs to be a root, hence the root coordination context.  

However, the doc migration code has its own containing context, hence the reason to splice the root and scopes from work coordination into the reindexing contexts when appropriate.

CreateSnapshot/src/main/java/com/rfs/CreateSnapshot.java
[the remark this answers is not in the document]
gregschohn: Ugh - I've changed this for now, but we have to switch everything over to be camel case.  This will be a bug waiting to happen.  See https://opensearch.atlassian.net/browse/MIGRATIONS-1837

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
peternied: Its hacky because it requires the caller RfsMigrateDocuments needs to know how to generate the correct kind of contexts for the work coordinater, it shouldn't need to know that.  Instead the work coordinator should be context aware - as in passed via construction, then `workItemsArePending` is called creates the appropriate context.

If we were passing a concrete context in function parameters, then I wouldn't have as much of an issue - but in this case a factory is being passed in.  While this is clever so a factory doesn't need to be provided into instance creation it makes this code very hard to interpret.

Ideally we would have code that follows a _single_ pattern of access to the context such as the following.  
```java
var docIndexer = new DocumentIndexer(context);
docIndexer.doIndexing();

// OR but less preferred

var docIndexer = new DocumentIndexer();
docIndexer.doIndexing(context);
```

RFS/src/main/java/com/rfs/cms/OpenSearchCmsClient.java
peternied, line 1: Remove this empty file?

commonDependencyVersionConstraints/build.gradle
```diff
@@ -101,6 +101,7 @@ dependencies {
 
     // make sure that this is compatible with the testcontainers version
     api group: 'org.opensearch', name: 'opensearch-testcontainers', version: '2.0.1'
+    api group: 'org.testcontainers', name: 'toxiproxy', version: '1.19.8'
```
peternied, line 104: This should be clustered with the other `org.testcontainers` resources and reusing the variable
```suggestion
    api group: 'org.testcontainers', name: 'toxiproxy', version: testcontainers
```

RFS/src/main/java/com/rfs/tracing/RfsContexts.java
peternied: Lets revisit this later, we've got many usages no need to stick on this instance here.

peternied:
OK - did another full pass, there are still a number of unresolved comments, I think some of them are outdated, but you might want to double check.

I posted a number of IMO non-controversial minor changes, please take a quick pass at those.

For the comments that between me and approval are:
- Relationship between contexts: https://github.com/opensearch-project/opensearch-migrations/pull/678#discussion_r1672432432
- Responsibility of context creation: https://github.com/opensearch-project/opensearch-migrations/pull/678#discussion_r1672454102


image

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
peternied: [Offline Recap] - For this use case the WorkCoordinator has a slightly different use case than other parts of the system so lets run with this style.  Maybe we need to revisit in the future, but there will always be exceptions.  Marking this resolved.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
peternied: [Offline Recap] - So yes this is generally true.  There is some slightly awkward terminology in the system, but there isn't any fundamental disagreement with how this works.  I'm fine moving forward with out any changes, resolving this thread.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/ProcessLifecycleTest.java
```diff
@@ -0,0 +1,182 @@
...
+            // This test is dependent on the toxiproxy being disabled before Migrate Documents is called.
+            // The Document Migration process will throw an exception immediately, which will cause an exit.
+            "AT_STARTUP, 1",
+            // This test is dependent upon the max lease duration that is passed to the command line.  It's set
+            // to such a short value (1s), that no document migration will exit in that amount of time.  For good
+            // measure though, the toxiproxy also adds latency to the requests to make it impossible for the
+            // migration to complete w/in that 1s.
+            "WITH_DELAYS, 2"})
+    public void testProcessExitsAsExpected(String failAfterString, int expectedExitCode) throws Exception {
+        final var failHow = FailHow.valueOf(failAfterString);
```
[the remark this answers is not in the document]
gregschohn: Not with CsvSource.  I could have used methodSource, but that makes code less apparent.  I think that it's clearer to the reader to have the test cases just before the method.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
gregschohn: The underlying difference is that the WorkCoordination code is its own independent subsystem from the rest of document migration.  Work coordination could be used from a number of different calling contexts, so the contexts that are relevant for each method of coordination may have different context ancestries.  That's the reason that a context factory instead is sent to each method rather than to the constructor.  The methods themselves could take a context instance as an arg, but then it would be up to the caller to close/autoclose the resource.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
gregschohn: This is mostly true.  This part isn't quite...
> CoordinationContext is a child of the DocumentMigrationContext where it pulls configuration.

Both of those root contexts are 'root' to their callers.  Document migration contexts will be created from the root object or children that were returned by that root.  Wook coordination contexts will be created by the work coordination root context.   Those coordination contexts may have ancestor spans that are document migration spans.

For both 'root' contexts, they derive their configuration from the otel sdk, which, now in the latest code, is the same object - meaning that there's just one "service" in the otel-space for this process, which probably makes a lot more sense now.

peternied (approved):

PR #689, opened 2024-05-29
[Documentation] Add details on focused Historical Migration deployment for E2E script

lewijacn:
### Description
See title

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1717

### Testing
N/A

### Check List


test/README.md
```diff
@@ -1,5 +1,63 @@
 ## End-to-End Testing
 
+### AWS E2E Testing
+Featured in this directory is a `./awsE2ESolutionSetup` script whose goal is to encapsulate deploying an E2E environment in AWS. This script will perform the following key actions:
+1. Deploy an Elasticsearch source cluster to EC2 using the following [CDK](https://github.com/lewijacn/opensearch-cluster-cdk/tree/migration-es)
```
chelma, line 5: Looks like you liked your personal fork; was this intentional?

test/README.md
```diff
@@ -1,5 +1,63 @@
...
+   * The CDK context options for this CDK can be used to customize this deployment and can be provided with a custom context file using the `--context-file` option
+   * A sample CDK context file can be seen [here](defaultCDKContext.json)
+2. Build the Docker Images
```
chelma, line 8: For newbies, would be helpful to explain which Docker Images you're referring to here.  Mind add those details?

test/README.md
```diff
@@ -1,5 +1,63 @@
...
+3. Deploy the Migration Assistant CDK (containing the Migration tooling and potentially an OpenSearch Domain) located in this repo [here](../deployment/cdk/opensearch-service-migration)
+   * The CDK context options for this CDK can be used to customize this deployment and can be provided with a custom context file using the `--context-file` option
+   * A sample CDK context file can be seen [here](defaultCDKContext.json)
```
chelma, line 11: I suspect folks will find it confusing to use the same CDK Context file for both the source cluster setup and the Migration Assistant.  Well, at least, I'd find it confusing.  What do you think about having a separate file for each?  Should be a pretty fast change.

test/README.md
```diff
@@ -1,5 +1,63 @@
...
+4. Add the traffic stream source security group created by the Migration Assistant CDK to the source cluster load balancer
```
chelma, line 12: Another sentence or two here explaining what this means would be appreciated by someone doing this for the first time, I think.  Mind adding that?

test/README.md
```diff
@@ -1,5 +1,63 @@
...
+    "restrictServerAccessTo": "0.0.0.0/0"
+  }
+}
+```
+Note: That this deployment will still deploy an unused Kafka broker ECS service, as a traffic stream source is still required for Migration Assistant deployments, but this linkage should be untangled with: https://opensearch.atlassian.net/browse/MIGRATIONS-1532
+
+In the same vein, the `--skip-capture-proxy` option can be provided to skip any capture proxy setup on source cluster nodes
+```
+./awsE2ESolutionSetup.sh --context-file <CONTEXT_FILE> --migration-context-id rfs-backfill -skip-capture-proxy
+```
```
chelma, line 58: This is probable exceeding the scope of the task you're currently working on, but - I'm curious why we have a command line parameter to skip the capture proxy setup instead of it just being an option via CDK Context?

Also - how do we specify the source cluster context we want to use in that file?

test/README.md
```diff
@@ -1,5 +1,63 @@
...
+    "engineVersion": "OS_2.11",
+    "domainName": "os-cluster-<STAGE>",
+    "dataNodeCount": 2,
+    "openAccessPolicyEnabled": true,
+    "domainRemovalPolicy": "DESTROY",
+    "artifactBucketRemovalPolicy": "DESTROY",
+    "kafkaBrokerServiceEnabled": true,
+    "trafficReplayerServiceEnabled": false,
+    "reindexFromSnapshotServiceEnabled": true,
+    "sourceClusterEndpoint": "<SOURCE_CLUSTER_ENDPOINT>"
```
chelma, line 32: Can we get a dictionary or breakdown of all the magic "phrases" that we can embed in these context files, and what they do?  We should probably explain how we're auto-replacing them with concrete values, too.

chelma (changes requested):

test/README.md
lewijacn: Yes this is where we have the ES EC2 CDK currently

test/README.md
lewijacn: Since each top-level object in this context file has an id that must be referenced if not using the default objects, I'm hoping that this won't be too confusing. I like being able to have multiple different configurations in one place and being able to select a given one to use.

test/README.md
lewijacn: Sure have added some more details now

test/README.md
lewijacn: Good mention, have added and updated some incorrect wording here

test/README.md
lewijacn: The capture proxy setup we do currently is outside the scope of the CDK and is done via AWS CLI commands, which this script contains

There is another command argument to specify the id that I didn't include, but is probably clearer to include so have updated now

test/README.md
lewijacn: Yes meant to add this, have included now

test/README.md
chelma: I'll be curious to hear why this is the case, but given this is a documentation task it's not a blocker for me.

test/README.md
chelma: I think what feels "off" to me isn't that there's multiple, different config "objects" in the files, it's that there's multiple types of config "objects" in the file.  It seems like we're already exposing users to the concept that there's two different CDK entities they're using by having them specify keys into this config JSON file.  Having a separate config file seems like a natural extension, and will enable us to do things like validate all the entries in each file using a common schema.

How much work do you think it would be to just split them out?

test/README.md
```diff
@@ -1,5 +1,80 @@
...
+
+The following CDK context values will be replaced by the `awsE2ESolutionSetup.sh` script if specified
+
+Migration Context substitutable values
+* `<STAGE>` replaces with default stage or stage provided with `--stage` argument
+* `<VPC_ID>` replaces with VPC ID created or used by source cluster
+* `<SOURCE_CLUSTER_ENDPOINT>` replaces with source cluster load balancer endpoint from source deployment
+
+Source Context substitutable values
+* `<STAGE>` replaces with default stage or stage provided with `--stage` argument
```
chelma, line 72: Yeah, we're already exposing customers to the fact there's two different types of context here.  Let's split out the file so all context objects in each file are the same type.  That way, users will be less likely to do things like using the wrong sub variable for the wrong context type since there will be a clearer distinction between the two context domains.

chelma (changes requested):

test/README.md
lewijacn: Personally I feel its an extra tax on a user to now have to create two files to use this script, and would expect any validation of context options to be done by the relevant CDK.

That being said, its not something high on my list of concerns and the change itself should be small and then require updating our Jenkins pipelines. Will see about adding this into the PR

lewijacn: Added now 👍

chelma (approved):
LGTM

PR #691, opened 2024-05-31
Annotation based telemetry

peternied:
### Description
This is an alternative approach to adding telemetry by using annotations.  I've only validated the metrics side of the system, still need to 

#### Todo to leave draft
- [ ] Validate the span processor
- [ ] Refactor test configuration
- [ ] Refactor WriteMeteringHandler/ReadMeteringHandler or defer
- [ ] Compare and contrast approach with https://github.com/opensearch-project/opensearch-migrations/pull/678

### Issues Resolved
- Resolves https://opensearch.atlassian.net/browse/MIGRATIONS-1738

### Check List

### DIFF
--- a/RFS/build.gradle
+++ b/RFS/build.gradle
@@ -5,6 +5,7 @@ plugins {
     id "com.avast.gradle.docker-compose" version "0.17.4"
     id 'com.bmuschko.docker-remote-api'
     id 'io.freefair.lombok' version '8.6'
+    id 'com.ryandens.javaagent-application' version "0.5.1"
 }
 
 import com.bmuschko.gradle.docker.tasks.image.DockerBuildImage
@@ -64,6 +65,7 @@ dependencies {
     testImplementation group: 'org.mockito', name: 'mockito-junit-jupiter'
     testRuntimeOnly group: 'org.junit.jupiter', name: 'junit-jupiter-engine'
 
+    javaagent("io.opentelemetry.javaagent:opentelemetry-javaagent:1.11.1")
 
     // Integration tests
     testImplementation group: 'org.testcontainers', name: 'testcontainers'
@@ -74,6 +76,17 @@ dependencies {
     implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind'
     implementation group: 'com.fasterxml.jackson.core', name: 'jackson-core'
     implementation group: 'com.fasterxml.jackson.core', name: 'jackson-annotations'
+
+    implementation platform("io.opentelemetry:opentelemetry-bom:1.34.1")
+
+    implementation('io.opentelemetry:opentelemetry-api')
+    implementation('io.opentelemetry:opentelemetry-sdk')
+    implementation('io.opentelemetry:opentelemetry-sdk-metrics')
+    
+    implementation("io.opentelemetry:opentelemetry-exporter-logging")
+    implementation('io.opentelemetry:opentelemetry-sdk-testing')
+    implementation('io.opentelemetry:opentelemetry-sdk-extension-autoconfigure')
+    implementation('io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.4.0')
 }
 
 application {
--- a/RFS/build.gradle
+++ b/RFS/build.gradle
@@ -87,6 +87,9 @@ dependencies {
     implementation('io.opentelemetry:opentelemetry-sdk-testing')
     implementation('io.opentelemetry:opentelemetry-sdk-extension-autoconfigure')
     implementation('io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.4.0')
+
+    testImplementation testFixtures(project(path: ':coreUtilities'))
+    testImplementation testFixtures(project(path: ':testHelperFixtures'))
 }
 
 application {
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -32,7 +32,6 @@ public class DocumentReindexer {
             .build();
     }
 
-
     private static final int MAX_BATCH_SIZE = 1000; // Arbitrarily chosen
 
     @WithSpan
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -6,6 +6,11 @@ import java.util.List;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.lucene.document.Document;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.metrics.LongCounter;
+import io.opentelemetry.instrumentation.annotations.SpanAttribute;
+import io.opentelemetry.instrumentation.annotations.WithSpan;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.util.retry.Retry;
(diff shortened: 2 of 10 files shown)


RFS/build.gradle
```diff
@@ -74,6 +76,20 @@ dependencies {
...
+    implementation('io.opentelemetry:opentelemetry-sdk')
+    implementation('io.opentelemetry:opentelemetry-sdk-metrics')
+    
+    implementation("io.opentelemetry:opentelemetry-exporter-logging")
+    implementation('io.opentelemetry:opentelemetry-sdk-testing')
+    implementation('io.opentelemetry:opentelemetry-sdk-extension-autoconfigure')
+    implementation('io.opentelemetry.instrumentation:opentelemetry-instrumentation-annotations:2.4.0')
+
+    testImplementation testFixtures(project(path: ':coreUtilities'))
+    testImplementation testFixtures(project(path: ':testHelperFixtures'))
```
peternied, line 92: TODO: pare this down to only whats needed

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -6,21 +6,42 @@
...
 public class DocumentReindexer {
     private static final Logger logger = LogManager.getLogger(DocumentReindexer.class);
+    private static final LongCounter documentsProcessedCounter;
+    private static final LongCounter batchesProcessedCounter;
+    static {
+        final var meter = GlobalOpenTelemetry.getMeter("DocumentReindexer");
+        documentsProcessedCounter = meter.counterBuilder("documentsProcessed")
+            .setDescription("Counts the number of documents processed")
+            .setUnit("1")
+            .build();
```
peternied, line 28: Curious what we think about this as a pattern - have the metrics defined in the class where they are used vs a centralized store.

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -38,34 +69,52 @@ public RestClient(ConnectionDetails connectionDetails) {
             });
     }
 
+    @WithSpan
```
peternied, line 72: Need to double check if these work with Mono<?> - whereas they do work with completable futures.

RFS/src/test/java/com/rfs/common/RestClientTest.java
```diff
@@ -0,0 +1,107 @@
...
+
+            long bytesRead = allMetricData.stream().filter(md -> md.getName().startsWith("readBytes"))
+                    .reduce((a, b) -> b).get().getLongSumData().getPoints()
+                    .stream()
+                    .filter(pd -> pd.getAttributes().asMap().values().stream().noneMatch(attrVal -> attrVal.equals(kvp.getKey())))
+                    .mapToLong(pd -> pd.getValue())
+                    .sum();
+            Assertions.assertEquals(kvp.getValue()[1], bytesRead);
+        }
+        // var finishedSpans = rootContext.instrumentationBundle.getFinishedSpans().stream()
```
peternied, line 72: Doesn't seem correct to validate spans in this scenario, as there really isn't an interesting 'context' where things are occurring since the client is declared in the test case.

I think the scenario still needs coverage, need to drill in more.  Not sure how important _detailed_ span traces are for RFS, curious for your thoughts @gregschohn

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -6,21 +6,42 @@
...
-    public static Mono<Void> reindex(String indexName, Flux<Document> documentStream, OpenSearchClient client) throws Exception {   
+    @WithSpan
+    public static Mono<Void> reindex(@SpanAttribute("indexName") String indexName, Flux<Document> documentStream, OpenSearchClient client) throws Exception {   
 
         return documentStream
             .map(DocumentReindexer::convertDocumentToBulkSection)  // Convert each Document to part of a bulk operation
+            .map(d -> { documentsProcessedCounter.add(1); return d; })
             .buffer(MAX_BATCH_SIZE) // Collect until you hit the batch size
-            .doOnNext(bulk -> logger.info(bulk.size() + " documents in current bulk request"))
+            .doOnNext(bulk -> { logger.info(bulk.size() + " documents in current bulk request"); batchesProcessedCounter.add(1); })
```
gregschohn, line 44: two lines for the different statements please.

RFS/build.gradle
```diff
@@ -5,6 +5,7 @@ plugins {
     id "com.avast.gradle.docker-compose" version "0.17.4"
     id 'com.bmuschko.docker-remote-api'
     id 'io.freefair.lombok' version '8.6'
+    id 'com.ryandens.javaagent-application' version "0.5.1"
```
gregschohn, line 8: A couple of deficiencies that concern me (and concerned me originally when I looked into alternate otel approaches).
* I don't think that there's an easy way to hook all of your spans into the ActiveContextTracker.  As it is, 3 lines would do that for RFS.  The ActiveContextTracker is a lifesaver into visibility when work items never terminate.
* I don't know how to chain/nest spans.  Attribute propagation looks like overkill in the coreUtils packages, but is invaluable when a top-level span doesn't finish.
* How would span linking work when parents are no longer in the backtrace of a given method call?
* It seems like there are about 20 lines of code added to each file that emits a metric.  The coreUtils solution takes that down to about 3 (import, context param, pass context param as is or by creating a new child).  If a developer spends most of their day perfecting the class that they're working in, why require them to also pollute the file with lots of instrumentation details.  20 lines of code doesn't take into account projecting a metric with different sets of attributes so that we can do analysis over more scenarios (like we do with status codes for the replayer).
* The clarity of the overall instrumentation story seems much better for the coreUtils packages.  With this code, I don't know how spans are related and it's hard to know where, if at all, a metric is emitted.  That sounds like it should be obvious, but we could easily add the # of bytes across the wire in 3 places and the contributors for each one may have no idea that other codepaths did the same (with the same or different metric names).
* Within instrumented code, spans and metrics are disjoint, so the implementor needs to know to keep these things in sync/manually make changes.

So, what are the benefits of this approach over direct utilization of the tools provided by org.opensearch.migrations.tracing?  As per the README for the coreUtils package...  In which ways is this alternative approach better or what was missing from that list?

>This library adapts the parts of OpenTelemetry to make it more natural and more foolproof throughout the rest of the TrafficCapture packages. This package introduces the concept of "Contexts" to build manage all tracing and metering instrumentation.
>
>Just as the otel metering and tracing can be efficiently disabled by not configuring them, this library provides some future-proofing by defining interfaces to track attributes, activities, exceptions, etc - but through descriptive interfaces where callers describe which actions they're performing, preventing the code from becoming overly complex.
>
>The goals of the instrumentation package are to
>
>1. make instrumentation classes easy to use.
>2. make it easy to create new safe and easy to use instrumentation classes.
>3. be efficient enough to use in most cases and flexible enough to tune in cases where the cost is too high.
>
>The third point is still a work in progress as the exact performance penalty isn't understood yet. However, work for point #2 dovetails into #3.  As context creations are chained together, a no-op uber-context can be created with zero memory footprint and minimal CPU penalty.  The first couple points are accomplished by putting contextual information alongside other data as first class parameters and fields. For example, where a method might require an identifier, a context might be passed instead so that the function can retrieve identifying information via the context AND have the ability to instrument activity within the appropriate context.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
peternied: @gregschohn Thanks for your thoughts, I see two tracks:
- **Where/how metrics are constructed**
  - It doesn't matter to me if they are statically defined or initialized through a DI pattern (lazy providers?)  I'm not married to use of `GlobalOpenTelemetry`.  
- **Where are metrics declared and used**
  - Generally: I typically see two models, single giant metrics file vs just-in-time-declaration.  As someone that has often had to splunk through other codebases the more distant the metric declaration is from its use makes it harder to be sure it was declared correctly unit errors like MB vs MBit are my bane.
  - For this case, the document and batch processing are features of this code.  Its a little messy that batches directly relate to a batch index request and I could see moving some of this processing onto the OpenSearchClient - I am OK with promoting/demoting where metrics are captured as long as we have a notation idea who is responsible.  While these metrics are easy to push downsteam, I think how often the Flux queue is blocked would need to be tracked here.  
  
Lets revisit this on your PR & future iteratinos rather than this revision.

RFS/src/main/java/com/rfs/common/RestClient.java
peternied: @gregschohn I've had a hard time getting started 'monitoring' the spans and so I'm going to call it quits on iterating here

peternied: The largest unknown to me is if the addition of the extra telemetry instrumented could cause us issues around deployment or performance impacts

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -289,7 +291,8 @@ public Optional<ObjectNode> updateDocument(String indexName, String documentId,
         }
     }
 
-    public Mono<BulkResponse> sendBulkRequest(String indexName, String body) {
+    @WithSpan
```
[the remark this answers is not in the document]
peternied: _Deferring since we aren't going to move forward with automatic instrumentation_

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -1,18 +1,49 @@
...
+    }
+
     public static class Response {
         public final int code;
         public final String body;
         public final String message;
 
-        public Response(int responseCode, String responseBody, String responseMessage) {
+        @WithSpan
+        public Response(@SpanAttribute("responseCode") int responseCode, String responseBody, String responseMessage) {
```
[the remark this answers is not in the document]
peternied: Ooh good callout - I missed that this was a constructor.  I'm guessing it would only track until the constructor returned the instance.

_Deferring since we aren't going to move forward with automatic instrumentation_

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -38,34 +69,52 @@ public RestClient(ConnectionDetails connectionDetails) {
...
     public Mono<Response> getAsync(String path) {
-        return client.get()
+        return client
+            .doOnRequest(sizeMetricsHandler("GET", path))
+            .get()
             .uri("/" + path)
             .responseSingle((response, bytes) -> bytes.asString()
-                .map(body -> new Response(response.status().code(), body, response.status().reasonPhrase())));
+            .map(body -> new Response(response.status().code(), body, response.status().reasonPhrase())))
+            .doOnError(Span.current()::recordException);
```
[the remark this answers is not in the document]
peternied: I never actually spun up monitoring on this so this was a debug todo

_Deferring since we aren't going to move forward with automatic instrumentation_

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
peternied: ACK - this was a quick hack that should never get past draft PRs

peternied:
@gregschohn Thanks for looking into this and providing some thoughts on what has already been done in the Traffic Capture space - I am going to close out this PR.  It (mostly) served its purpose in getting me familiar with one of the options and helps clarify the advantages to the time investment in with the manual instrumentation and span definition.

PR #695, opened 2024-06-04
[Migration Console] Providing a services.yaml file on ECS

mikaylathompson:
### Description
This change creates a service.yaml file as resource are created by CDK, and loads it into ParameterStore as a parameter. The migration console docker file is updated to include a `loadServicesFromParameter.sh` script that checks whether a particular env var is set, and if so, pulls the value of the parameter and writes it to a file on startup.

Currently, the file is populated with `metrics_source` (cloudwatch), `target_cluster` and `source_cluster`. There's more work to do to support pulling cluster credentials from SecretsManager, but I'm going to capture that in another task because it's not directly related to this one.

### Issues Resolved
[MIGRATIONS-1744](https://opensearch.atlassian.net/browse/MIGRATIONS-1744)

### Testing
Unit tests added, plus manual testing.

### Check List


peternied:
@mikaylathompson Latest failure was from the following a rate limiting issue - I know Andre had looked into this by using an alternative hosting provider, but maybe we could update our build job to use the gradle cache [1] for docker images which would bypass the download if we are using the same base image.  Want to take that on or create an issue?

```
Execution failed for task ':TrafficCapture:dockerSolution:buildDockerImage_otelCollector'.
> Could not build image: toomanyrequests: Rate exceeded

* Try:
> Run with --debug option to get more log output.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':TrafficCapture:dockerSolution:buildDockerImage_otelCollector'.
```

- [1] https://github.com/peternied/contribution-rate/blob/main/.github/workflows/reports.yml#L28-L43

deployment/cdk/opensearch-service-migration/lib/opensearch-domain-stack.ts
```diff
@@ -117,6 +119,18 @@ export class OpenSearchDomainStack extends Stack {
...
   }
 
+  generateTargetClusterYaml(domain: Domain, adminUserName: string | undefined) {
+    let targetCluster = new ClusterYaml()
+    targetCluster.endpoint = `https://${domain.domainEndpoint}:443`;
+    if (adminUserName && domain.masterUserPassword) {
+      // TODO: need to actually deal with the password as a secret (pass the arn)
+      targetCluster.basic_auth = { user: adminUserName, passsword: domain.masterUserPassword }
+    } else {
+      targetCluster.no_auth = ''
```
gregschohn, line 129: why is no_auth a string and not a boolean?  Or is it really just null vs non-null?

gregschohn (approved):
I left one comment to change the name for the parameter-store ts file.  Other than that, this lgtm.

deployment/cdk/opensearch-service-migration/lib/opensearch-domain-stack.ts
mikaylathompson: Yeah, it's just testing what's present -- so it'll be:
```yaml
target_cluster:
  endpoint: xyz
  no_auth:
```
vs
```
target_cluster:
  endpoint: xyz
  basic_auth:
    user: ...
```

PR #698, opened 2024-06-06
Cache docker images for E2E testing workflow in Github Actions

mikaylathompson:
### Description
We’re frequently encountering issues where builds fail due to rate limiting in the github actions workflows for pulling docker images.

For example, see https://github.com/opensearch-project/opensearch-migrations/pull/695#issuecomment-2148465012.
While we’ve considered multiple ways to resolve this, a quick one (suggested in the comment above) is to use the github actions caching mechanism to cache our docker images.

One issue I hit along the way was that I was defining the action as:
```
      - name: Cache Docker Images
        uses: ScribeMD/docker-cache@0.5.0
        with:
          key: docker-${{ runner.os }}-${{ hashFiles('**/docker-compose.yml', '**/Dockerfile') }}
```
which does the same thing as now--a hash of the file contents of all docker-compose and Dockerfiles in the repo. However, it turns out that this key is evaluated at runtime--both during the original action (loading from cache) and the post-action (saving to cache). However, our gradle build process includes generating new dockerfiles, so the value of this key was changing from pre to post--meaning that there would almost by definition never be a cache hit.  Calculating the hash in a separate step and storing it resolves this issue.

### Issues Resolved
[MIGRATIONS-1761](https://opensearch.atlassian.net/browse/MIGRATIONS-1761)

### Testing
A lot of manual testing -- I left my former branch open as a PR in my fork to preserve the testing history: https://github.com/mikaylathompson/opensearch-migrations/pull/1

### Check List

### DIFF
--- a/.github/workflows/e2eTest.yml
+++ b/.github/workflows/e2eTest.yml
@@ -22,6 +22,19 @@ jobs:
           java-version: '11'
           distribution: 'corretto'
 
+      - name: Generate Cache Key from Dockerfiles
+        id: generate_cache_key
+        run: |
+          files=$(find . -type f \( -name 'docker-compose.yml' -o -name 'Dockerfile' \))
+          file_contents=$(cat $files)
+          key=$(echo "${file_contents}" | sha1sum | awk '{print $1}')
+          echo "key=${key}" >> "$GITHUB_OUTPUT"
+
+      - name: Cache Docker Images
+        uses: ScribeMD/docker-cache@0.5.0
+        with:
+          key: docker-${{ runner.os }}-${{ steps.generate_cache_key.outputs.key }}
+
       - name: Setup Gradle
         uses: gradle/actions/setup-gradle@v3
         with:


gregschohn (approved):
If this works, please put this in the gradle-build-and-test.yml file too

.github/workflows/e2eTest.yml
```diff
@@ -22,6 +22,19 @@ jobs:
           java-version: '11'
           distribution: 'corretto'
 
+      - name: Generate Cache Key from Dockerfiles
```
peternied, line 25: > However, it turns out that this key is evaluated at runtime

Good find and workaround, since this is a bug in the action, could you create an issue so they can fix it in a future release?

peternied (approved):

.github/workflows/e2eTest.yml
mikaylathompson: Done: https://github.com/ScribeMD/docker-cache/issues/823

mikaylathompson: (I have most of a PR for them, but the tests are pretty brittle and I'm not sure how much time I want to put into resolving this before I get an indication about whether they'd be open to it anyways)

mikaylathompson:
I want to give this a few days of bake time so we can spot any issues, but issue for adding it to the other pipeline created here: https://opensearch.atlassian.net/browse/MIGRATIONS-1766

PR #705, opened 2024-06-06
Proposal for transformations design

peternied:
### Description
I've written out a design proposal for how we might want to pivot our efforts on building up transformations.  This is meant to solicit feedback and be iterative.

I'll publish this early if folks are curious for early feedback, but when its really ready for review I'll message more broadly.


peternied:
Closing in favor of #783

PR #711, opened 2024-06-10
[Testing] Add metric verification to e2e tests

mikaylathompson:
### Description
Adds a decorator to our end to end tests that asserts that specific metrics are present in the metrics store after a test is run.

The biggest limitation (noted in the code) is that we don't currently have any way to attribute which test run a given metric or data-point is associated with. For now, I'm just using a 2 minute lookback to at least verify that the metrics came from this run of the e2e tests and perhaps limit it to the latest test or two.

However, I have verified that if I shut off either prometheus or the otel collector, this causes the test to fail. I think that adds significant value over where we are now, and provides a jumping off point to continue to use the console library from the e2e tests, as well as to expand our metric verification (do the values look correct?) in the future.

### Issues Resolved
[MIGRATIONS-1699](https://opensearch.atlassian.net/browse/MIGRATIONS-1699) & [MIGRATIONS-1700](https://opensearch.atlassian.net/browse/MIGRATIONS-1700)

I also encountered and resolved a number of bugs while testing this:
- We were mounting the console library directory to the migration console, which was very helpful for local development, but breaking use of the library in some circumstances (including github action workflows). This is still available in the docker-compose, but commented out by default.
- Some fallout from #718 that wasn't updated in the CDK 
- I was unlucky enough to be testing last night at the very moment that numpy 2.0 was released. That caused a conflict with another dependency (h5py) so I've constrained the numpy version for the time being.
- we weren't actually attaching the get metrics policy to the migration console in cdk
- Our migration console was running out of CPU/memory when running `runBenchmarks`, which wasn't caught because we weren't checking the exit code of that command.
- A non-standalone capture proxy (e.g. running on an elasticsearch node on EC2) doesn't have an otel collector since the move to the sidecar model and therefore doesn't emit metrics. For the time being, I've disabled checking for captureProxy metrics in a cloud deployment.

### Testing
This is testing.

### Check List


test/operations.py
```diff
@@ -67,3 +72,31 @@ def get_document(endpoint: str, index_name: str, doc_id: str, auth,
...
+            logger.warning(f"Running command: {cmd_exec} on container {container_id}")
+            process = subprocess.run(cmd_exec, shell=True, capture_output=True, text=True)
+            return process.returncode, process.stdout, process.stderr
+        else:
+            raise ContainerNotFoundError(filter_criteria)
+
+    else:
+        # In a cloud deployment case, we run the e2e tests directly on the migration console, so it's just a local call
+        logger.warning(f"Running command: {command} locally")
+        process = subprocess.run(cmd_exec, shell=True, capture_output=True)
```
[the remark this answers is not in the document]
mikaylathompson: You're completely right. This should be `command`.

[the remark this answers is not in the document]
mikaylathompson: And I also ran this on the Jenkins secondary pipeline to confirm that this path was exercised and works: http://opense-jenki-yfnf7ayv4o49-506509445.us-east-1.elb.amazonaws.com/job/OS%20Migrations%20Default%20E2E%20Test/1478/console

lewijacn (approved):

test/tests.py
```diff
@@ -188,6 +188,11 @@ def assert_metrics(self, expected_metrics: Dict[str, List[str]], lookback_minute
         logger.debug(f"Waiting {wait_before_check_seconds} before checking for metrics.")
         time.sleep(wait_before_check_seconds)
         for component, expected_comp_metrics in expected_metrics.items():
+            if component == "captureProxy" and self.deployment_type == "cloud":
+                # We currently do not emit captureProxy metrics from a non-standalone proxy, which is the scenario
```
lewijacn, line 192: Thanks for comment here, a log statement that metric assertions are being skipped could also bring clarity to say users looking at this test through Jenkins logs

deployment/cdk/opensearch-service-migration/test/migration-console-stack.test.ts
```diff
@@ -81,7 +81,7 @@ test('Test that services yaml parameter is created', () => {
     const yamlFileContents = value['Fn::Join'][1].join('')
     expect(yamlFileContents).toContain('source_cluster')
     expect(yamlFileContents).toContain('target_cluster')
-    expect(yamlFileContents).toContain('metrics_source:\n  type: cloudwatch')
```
lewijacn, line 84: Is cloudwatch null here because we haven't set it up in our services.yaml with CDK yet?

mikaylathompson: No, it's totally set up and working (this is basically what the cloud deployment option tests).

It's actually because all we care about is the presence of the key. So the YAML _should_ look like:
```
metrics:
  cloudwatch:
```
or, if an optional value is provided,
```
metrics:
  cloudwatch:
    region: us-east-1
```
but an empty value = null, and the YAML library dumps it with the explicit type. But I agree it's confusing. 

This comment prompted me to go down a little mini rabbit hole, and I think it's actually possible to change the dump preferences, so i'm going to do a quick check with that and see if that makes it much clearer.

test/tests.py
mikaylathompson: I like that, also a good reminder to change this in the future.

PR #715, opened 2024-06-10
New RFS work coordination

gregschohn:
### Description

### Issues Resolved

### Testing

### Check List


commonDependencyVersionConstraints/build.gradle
```diff
@@ -90,6 +90,9 @@ dependencies {
     api group: 'org.testcontainers', name: 'testcontainers', version: testcontainers
     api group: 'org.testcontainers', name: 'toxiproxy', version: testcontainers
 
+    // make sure that this is compatible with the testcontainers version
```
peternied, line 93: Does this comment needs to be followed up on?  Otherwise lets delete, it works right?

RFS/src/test/java/com/rfs/cms/TransactionalOpenSearchDataStoreTest.java
```diff
@@ -0,0 +1,239 @@
...
+ * in all non-contentious cases, clients only need to make one call per work item.  Multiple calls are only
+ * required when a lease has expired and a new one is being granted since the worker/client needs to make the
+ * GET call to find out the new expiration value.
+ */
+@Slf4j
+public class TransactionalOpenSearchDataStoreTest {
+    private static final String INDEX_NAME = ".migrations_working_state";
+
+    final static OpensearchContainer<?> container =
+            new OpensearchContainer<>("opensearchproject/opensearch:2.11.0");
```
peternied, line 41: Might want to make sure this works on 1.3.x instead of 2.X.

RFS/src/test/java/com/rfs/cms/TransactionalOpenSearchDataStoreTest.java
```diff
@@ -0,0 +1,239 @@
...
+            "      \"expirationWindow\": {EXPIRATION_WINDOW},\n" +
+            "      \"workerId\": \"{WORKER_ID}\"\n" +
+            "    },\n" +
+            "    \"source\": \"" +
+            "      if (ctx._source.scriptVersion != \\\"{SCRIPT_VERSION}\\\") {" +
+            "        throw new IllegalArgumentException(\\\"scriptVersion mismatch.  Not all participants are using the same script: sourceVersion=\\\" + ctx.source.scriptVersion);" +
+            "      } " +
+            "      long serverTimeSeconds = System.currentTimeMillis() / 1000;" +
+            "      if (Math.abs(params.clientTimestamp - serverTimeSeconds) > {CLOCK_DEVIATION_SECONDS_THRESHOLD}) {" +
+            "        throw new IllegalArgumentException(\\\"The current times indicated between the client and server are too different.\\\");" +
```
peternied, line 90: Why do they need to be the same at all?  The server could be different from the client in a way that would be considered 'wrong' but as long as its _consistent_ it satisfies our needs, no?

peternied (commented):
Nice work here - thanks for building this out as a test too!

commonDependencyVersionConstraints/build.gradle
gregschohn: Thanks Peter!  Scariest 5 minutes of my life between when I read this and verified that it does indeed work!.
I had meant to do that much earlier, but then got caught up.

RFS/src/test/java/com/rfs/cms/TransactionalOpenSearchDataStoreTest.java
gregschohn: Yes, but... right now, we're not doing any kind of adjustments or calibrations to use the server side clock when calculating when we should be killing the process.  This is sortof a poor substitution that hopefully adds some sanity.  I would hope that some day we'll get better time coordination, but the truth is that we're probably at a much coarser granularity than what even the worst offenders of clock skew would exhibit for this to be a problem.

peternied:
@gregschohn isnt this PR replaced by #739 ?

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
peternied, line 1: For a fast follow - lets get tests around all of the different scripts in this class that deploy them and check assumptions about how they operate, I've got no idea if we've got pitfalls in the code that is strings.

peternied (approved):

PR #717, opened 2024-06-11
[RFS] Collection of Minor Changes

chelma:
### Description
* Updated the RFS Docker Compose setup to run Elasticsearch 7.17 by default as a source
* Updated RFS to accept a maximum shard size it will attempt to migrate.  The default chosen is 50 GB.  If it encounters a shard larger than the max, it will increment the attempt count without downloading/migrating, resulting in the shard being eventually marked as FAILED.
* Updated RFS to delete the Snapshot Blobfiles from S3 after unpacking them into their Lucene index.  We do not delete the metadata files as they are not likely to cause problems with disk space (~5 kB/shard), we can add that behavior later with the abstractions I've introduced, and I'm currently on a tight schedule at the moment.
* Updated RFS to delete the shards' unpacked Lucene Indices from disk after we're done with it (whether we successfully migrated it or not).  Does not delete the directory that the Lucene Indices were housed in.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1749

### Testing
* Added/updated unit tests
* Ran a migration using the Docker Compose setup locally.  Confirmed that shards larger than a specified max were rejected without downloading any files, confirmed that the s3 downloaded blobfiles were cleaned up, and confirmed that the Lucene Index directories were cleaned out.

### Check List


RFS/src/main/java/com/rfs/common/DeletingSourceRepoAccessor.java
```diff
@@ -0,0 +1,52 @@
...
+        public void close() throws IOException {
+            try {
+                super.close();
+            } finally {
+                Files.deleteIfExists(filePath);
+            }
+        }
+    }
+
+    public static class CouldNotLoadRepoFile extends RuntimeException {
```
peternied, line 47: Move this out into an Exceptions package, we we don't need class specific versions of the same named exception between this class an DefaultSourceRepoAccessor

RFS/src/main/java/com/rfs/common/SourceRepoAccessor.java
```diff
@@ -0,0 +1,47 @@
+package com.rfs.common;
```
peternied, line 1: Please run formatter of your choice of this file any for any other changes to make whitespace and indenting consistent

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
```diff
@@ -58,6 +60,35 @@ public void unpack(ShardMetadata.Data shardMetadata) {
         }
     }
 
+    public void cleanUp(ShardMetadata.Data shardMetadata) {
```
peternied, line 63: Rather than have this method, lets have reuse the java convention to implement AutoCloseable so its clear that when you are done with an unpacker it should clean up itself.

This allows code like:

```java
try (var unpacker = getShardUnpacker()) {
   unpack.things();
}
// unpacker close method was automatically called here
```

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
```diff
@@ -18,10 +18,10 @@ public class DocumentsRunner implements Runner {
 
     private final DocumentsStep.SharedMembers members;
 
-    public DocumentsRunner(GlobalState globalState, CmsClient cmsClient, String snapshotName, IndexMetadata.Factory metadataFactory,
-                ShardMetadata.Factory shardMetadataFactory, SnapshotShardUnpacker unpacker, LuceneDocumentsReader reader,
-                DocumentReindexer reindexer) {
-        this.members = new DocumentsStep.SharedMembers(globalState, cmsClient, snapshotName, metadataFactory, shardMetadataFactory, unpacker, reader, reindexer);
+    public DocumentsRunner(GlobalState globalState, CmsClient cmsClient, String snapshotName, long maxShardSizeBytes,
```
peternied, line 21: This is nigh impossible to read - lets move to using @requireargsconstructor as a quick stop gap, but I think we need to have better data modeling so we don't have to pass every configuration option around, clustering by functionality or having a settings object that can be peaked into would handle this

RFS/src/main/java/com/rfs/common/SourceRepoAccessor.java
chelma: Reformatted, thanks.

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
chelma: I don't find it hard to read, but feel free to update it how you'd like.

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
```diff
@@ -58,6 +72,36 @@ public void unpack(ShardMetadata.Data shardMetadata) {
         }
     }
 
+    @Override
+    public void close() {
+        try {
+            Path luceneIndexDir = Paths.get(luceneFilesBasePath + "/" + shardMetadata.getIndexName() + "/" + shardMetadata.getShardId());
```
gregschohn, line 78: It looks like you're deleting a session directory here.  Why did you need to delete files above when tied to the stream?

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
```diff
@@ -58,6 +72,36 @@ public void unpack(ShardMetadata.Data shardMetadata) {
...
+            if (Files.exists(luceneIndexDir)) {
+                deleteRecursively(luceneIndexDir);
+            }
+            
+        } catch (Exception e) {
+            throw new CouldNotCleanUpShard("Could not clean up shard: Index " + shardMetadata.getIndexId() + ", Shard " + shardMetadata.getShardId(), e);
```
gregschohn, line 84: Should/does this kill the process?  Seems like it probably should since it would be safer to recycle the process, especially since you were at the end of working (if you have other pending work, maybe it should not take more on and flush out the current work items).

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
gregschohn: In general, I go back and forth on this one in other areas of Java code.  Doing the 'right' thing with builders is cumbersome and it only amounts to a runtime check.
Please try to put these on separate lines and order them to tell the best story possible.  This is basically dependency injection and everything is required, right?

gregschohn (commented):
I'm a bit confused at what the deletion model looks like.  Please help to clarify it so that I can understand and approve this.

RFS/docker/TestSource_ES_7_17/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM docker.elastic.co/elasticsearch/elasticsearch:7.17.21 AS base
```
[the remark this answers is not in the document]
chelma: Probably; we'll want something like that for testing purposes once we have a bunch of different source/target versions to support.  I'm unclear on how generalizable the Dockerfiles will be across major versions though.

RFS/src/main/java/com/rfs/DemoPrintOutSnapshot.java
```diff
@@ -27,6 +27,7 @@
 import com.rfs.common.SnapshotRepo;
 import com.rfs.common.SnapshotShardUnpacker;
 import com.rfs.common.ClusterVersion;
+import com.rfs.common.DefaultSourceRepoAccessor;
```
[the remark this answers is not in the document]
chelma: This one, definitely not; been keeping it around "just in case" but now that you've pointed it out I can't think of a case where I'd want it.  Will delete

RFS/src/main/java/com/rfs/RunRfsWorker.java
```diff
@@ -132,6 +137,7 @@ public static void main(String[] args) throws Exception {
         String targetPass = arguments.targetPass;
         List<String> indexTemplateAllowlist = arguments.indexTemplateAllowlist;
         List<String> componentTemplateAllowlist = arguments.componentTemplateAllowlist;
+        long maxShardSizeBytes = arguments.maxShardSizeBytes;
```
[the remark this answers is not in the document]
chelma: I guess they should be `finals`, huh?  Will start doing that.  I make the intermediate copies to shorten the name when referring to them and also to make a conceptual split between the argument passed in and the thing we're using in the program.  Makes more sense when we're doing validation or other processing first, but I just tend to apply the pattern generally.

Will make them final per your suggestion, though.

RFS/src/main/java/com/rfs/RunRfsWorker.java
```diff
@@ -165,10 +171,11 @@ public static void main(String[] args) throws Exception {
             indexWorker.run();
 
             ShardMetadata.Factory shardMetadataFactory = new ShardMetadataFactory_ES_7_10(repoDataProvider);
-            SnapshotShardUnpacker unpacker = new SnapshotShardUnpacker(sourceRepo, luceneDirPath, ElasticsearchConstants_ES_7_10.BUFFER_SIZE_IN_BYTES);
+            DeletingSourceRepoAccessor repoAccessor = new DeletingSourceRepoAccessor(sourceRepo);
```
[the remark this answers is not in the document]
chelma: Happy to rename.  `EphemeralSourceRepoAccessor` tickles my brain a bit better, so I'll go with that.

RFS/src/main/java/com/rfs/common/DeletingSourceRepoAccessor.java
```diff
@@ -0,0 +1,46 @@
...
+    @Override
+    protected InputStream load(Path path) {
+        try {
+            return new DeletingFileInputStream(path);
+        } catch (Exception e) {
+            throw new CouldNotLoadRepoFile("Could not load file: " + path, e);
+        }
+    }
+
+    public static class DeletingFileInputStream extends FileInputStream {
```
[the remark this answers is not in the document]
chelma: Probably easier to talk this one out.

RFS/src/main/java/com/rfs/common/PartSliceStream.java
```diff
@@ -12,24 +12,23 @@
  */
 
 public class PartSliceStream extends InputStream {
```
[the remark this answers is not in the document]
chelma: The special sauce is that it provides a single stream-like object that seamlessly reads through multiple Snapshot blob files that have split into multiple parts, per the Elasticsearch/OpenSearch convention of not having any individual file bigger than ~1 GB.

We don't have to do things this way, but it's what the ES/OS code did and I haven't had a reason to change it.

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
chelma: Not sure I understand the question.  Can you re-phrase?

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
chelma: Makes sense.  With the way things are currently implemented, I believe it will kill the process.

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
chelma: @gregschohn Yeah, this is just dependency injection.  I can put them on new lines if that makes it a bit clearer, but the bigger issue is that there's something that feels off with how the SharedMembers object is created in an intermediate scope between the main() and the Step classes that need it.  It feels like the SharedMembers definition might not actually belong in the Step class?

RFS/src/main/java/com/rfs/ReindexFromSnapshot.java
```diff
@@ -329,7 +329,8 @@ public static void main(String[] args) throws InterruptedException {
                 } else {
                     bufferSize = ElasticsearchConstants_ES_7_10.BUFFER_SIZE_IN_BYTES;
                 }
-                SnapshotShardUnpacker unpacker = new SnapshotShardUnpacker(repo, luceneDirPath, bufferSize);
+                DefaultSourceRepoAccessor repoAccessor = new DefaultSourceRepoAccessor(repo);
```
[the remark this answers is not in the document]
chelma: We can once we integrate the new version of this stuff into the Migration Assistant, which currently has a dependency on this file.

RFS/src/main/java/com/rfs/RunRfsWorker.java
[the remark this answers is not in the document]
gregschohn: I guess my question was more - why setup the aliases in the first place?

RFS/src/main/java/com/rfs/common/PartSliceStream.java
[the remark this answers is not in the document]
gregschohn: Is it like the [SequenceInputStream](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/io/SequenceInputStream.html), or maybe like that with a bit more support through it?

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
gregschohn: Is this redundant to what you are doing with the DeletingFileInputStream above?  This place feels like where all of your cached files should be deleted - or are there other files that are created outside of this class?

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
gregschohn: Do you have a test to confirm that?  I ask because making sure that processes die, especially through refactorings, can be pretty tricky to maintain.

gregschohn (commented):
Responded - trying to figure out where all of the ephemeral files are written and who all has responsibilities to delete them.

RFS/src/main/java/com/rfs/common/PartSliceStream.java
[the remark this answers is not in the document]
chelma: Looks like "kinda"?  Basically, you just have a base blobfile name (foo) and if it's larger than 1 GB it's split into multiple files (foo.part0, foo.part1, foo.part2).  The PartSliceStream creates a stream that mimics as if they were all the same file you're reading from.

RFS/src/main/java/com/rfs/common/DeletingSourceRepoAccessor.java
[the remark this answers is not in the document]
chelma: After discussion, we decided to keep this around in case we need it in the future but use the Default (non-deleting) version in RFS for now.  If the disk fills up from our s3 downloads, it should kill the process and naturally free up disk space by getting a new container.

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
chelma: Per discussion - there are two things that we need to clean up: the raw snapshot files we download from s3, and the files we convert those into which Lucene actually cares about.  This is for cleaning up the later.

gregschohn: As per the [jira](https://opensearch.atlassian.net/browse/MIGRATIONS-1786) in the log message added above, we should eventually get this down to no extra file buffers, which reduces the number of resources that we're managing and should make the code more efficient and easier to maintain.

gregschohn (approved):

PR #728, opened 2024-06-18
[Console] Backfill -- Add interface and YAML specification

mikaylathompson:
### Description
This sets the groundwork for implementing backfill backend adapters for RFS by establishing the base class/interface and logic layer.

The OSI Backfill implementation already exists, but this PR does some refactoring around it. It creates--but does not implement--the RFS Backfill backends (one for docker and one for ECS).

I also laid out a starting point for the backfill section of the services.yaml (this is very much not stable and is highly subject to change as we figure out what we need) and enforce it with Cerberus schemas.

### Issues Resolved
[MIGRATIONS-1776](https://opensearch.atlassian.net/browse/MIGRATIONS-1776) (fully completed) & [MIGRATIONS-1777](https://opensearch.atlassian.net/browse/MIGRATIONS-1777) (partially completed -- the remaining component is to populate the backfill info from CDK for a cloud-deployed migration console)

### Testing
Unit tests added, manually tested in docker setup.

### Check List


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cli.py
```diff
@@ -0,0 +1,97 @@
...
+def test_cli_cat_indices_e2e(runner, env):
+    with requests_mock.Mocker() as rm:
+        rm.get(f"{env.source_cluster.endpoint}/_cat/indices",
+               status_code=200,
+               text=source_cat_indices)
+        rm.get(f"{env.target_cluster.endpoint}/_cat/indices",
+               status_code=200,
+               text=target_cat_indices)
+        result = runner.invoke(cli, ['--config-file', str(VALID_SERVICES_YAML), 'clusters', 'cat-indices'],
+                               catch_exceptions=True)
```
chelma, line 91: Learned something new here - thanks!

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
```diff
@@ -0,0 +1,126 @@
...
+            raise ValueError("target_cluster must be provided for RFS backfill")
+
+        if 'docker' in config[BackfillType.reindex_from_snapshot.name]:
+            logger.debug("Creating Docker RFS backfill instance")
+            return DockerRFSBackfill(config=config,
+                                     target_cluster=target_cluster)
+        elif 'ecs' in config[BackfillType.reindex_from_snapshot.name]:
+            logger.debug("Creating ECS RFS backfill instance")
+            return ECSRFSBackfill(config=config,
+                                  target_cluster=target_cluster)
```
chelma, line 45: Now that I see this, I'm even more inclined to use ABCs to enforce the interface contract we're currently implying, but not guaranteeing, exists.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
```diff
@@ -0,0 +1,126 @@
...
+    if len(config.keys()) > 1:
+        raise UnsupportedBackfillTypeError(', '.join(config.keys()))
+    raise UnsupportedBackfillTypeError(next(iter(config.keys())))
+
+
+def describe(backfill: Backfill, as_json=False) -> str:
+    response = backfill.describe()
+    if as_json:
+        return json.dumps(response)
+    return yaml.safe_dump(response)
```
chelma, line 57: Not a blocker, just curious - why do we need both, and when should I use either?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
```diff
@@ -0,0 +1,126 @@
...
+            return ECSRFSBackfill(config=config,
+                                  target_cluster=target_cluster)
+
+    logger.error(f"An unsupported metrics source type was provided: {config.keys()}")
+    if len(config.keys()) > 1:
+        raise UnsupportedBackfillTypeError(', '.join(config.keys()))
+    raise UnsupportedBackfillTypeError(next(iter(config.keys())))
+
+
+def describe(backfill: Backfill, as_json=False) -> str:
```
chelma, line 53: I'm curious about this paradigm - why wrap the Backfill in this way?  Having the additional layer here could be surprising to folks and I'm interested to hear what, if anything, would go wrong if someone came in and used the object directly?  

As an alternative, why not have the return-type modularity exist in something like an OutputType enum or an Outputter object that the Backfill object contains?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
```diff
@@ -0,0 +1,126 @@
...
+    response = backfill.describe()
+    if as_json:
+        return json.dumps(response)
+    return yaml.safe_dump(response)
+
+
+def create(backfill: Backfill, *args, **kwargs) -> str:
+    if isinstance(backfill, RFSBackfill):
```
chelma, line 61: Now we're doing runtime type checking, which is sometimes an indication that encapsulation is wonky.  Why not have this behavior owned by the RFSBackfill object?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_rfs.py
```diff
@@ -0,0 +1,63 @@
...
+        v = Validator(RFS_BACKFILL_SCHEMA)
+        if not v.validate(self.config):
+            raise ValueError("Invalid config file for RFS backfill", v.errors)
+
+
+class DockerRFSBackfill(RFSBackfill):
+    def __init__(self, config: Dict, target_cluster: Cluster) -> None:
+        super().__init__(config)
+        self.target_cluster = target_cluster
+        self.docker_config = self.config["reindex_from_snapshot"]["docker"]
```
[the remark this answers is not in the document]
chelma: I see now we're actually doing that elsewhere.

chelma (changes requested):

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
mikaylathompson: This is in preparation for the web API, which will likely want jsons for almost everything. So the idea is to eventually have a json version and optionally a non-json version for every call.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
mikaylathompson: Basically, why the logic layer?

The idea is two-fold:
1. The CLI and API are intended to be very thin clients, potentially generated automatically from a spec. This logic/middleware layer will be the layer that those clients interact with, so all they have to know is what functions are available from the library. Potentially, these files can even be used as the spec to generate the clients.
2. The CLI and API should never have to know exactly what underlying backend adapters they're dealing with. But in some cases (e.g. CLI params that only apply to one type) _someone_ in between the CLI and the class itself needs to know what's happening under the hood. That's what can happen in this layer.

Someone who's building on top of the library directly is welcome to use the underlying classes, but then they're taking on responsibility for that.

Very open to feedback on a better way to do it.

For the return type specifically, I like the idea of standardizing it more--I'm not sure I quite understand what you mean by "an Outputter object that the Backfill object contains"

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/logic/backfill.py
mikaylathompson: Yeah, i'm not crazy about this piece. I think you're right that this is a good case to have the class itself handle it.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_base.py
```diff
@@ -0,0 +1,45 @@
...
+        "schema": {
+            "opensearch_ingestion": {"type": "dict"},
+            "reindex_from_snapshot": {"type": "dict"},
+        },
+        "check_with": contains_one_of({"opensearch_ingestion", "reindex_from_snapshot"})
+    }
+}
+
+
+class Backfill:
```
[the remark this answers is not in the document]
mikaylathompson: Yeah, I think you're right. No good reason, mostly just hadn't gotten to it yet.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_base.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    Interface for backfilling data from a source to target cluster.
+    """
+    def __init__(self, config: Dict) -> None:
+        v = Validator(SCHEMA)
+        self.config = config
+        if not v.validate({"backfill": self.config}):
+            raise ValueError("Invalid config file for backfill", v.errors)
+
+    def create(self):
```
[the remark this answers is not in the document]
mikaylathompson: Oh, good point. Right now these descriptions pretty much just exist in my head. I'll start adding them as I go through.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/README.md
```diff
@@ -45,14 +45,16 @@ backfill:
             - "migration_deployment=1.0.6"
 ```
 
-### Services.yaml spec
+## Services.yaml spec
```
peternied, line 48: Can we generate a table of contents for this file?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/README.md
```diff
@@ -61,58 +63,119 @@ Exactly one of the following blocks must be present:
...
 - `prometheus`:
     - `endpoint`: required
 
 - `cloudwatch`: may be empty if region is not specified
     - `aws_region`:  optional. if not provided, the usual rules are followed for determining aws region. (`AWS_DEFAULT_REGION`, `~/.aws/config`, etc.)
 
+### Backfill
+
+Backfill can be performed via several mechansims. The primary two supported by the console library are Reindex From Snapshot (RFS) and
+OpenSearch Ingestion Pipeline (OSI).
```
peternied, line 80: Do we want to provide any context on why we offer both or what the difference is?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
```diff
@@ -93,42 +98,75 @@ def start_replayer_cmd(ctx):
...
+
 
+@backfill_group.command(name="describe")
+@click.pass_obj
+def describe_backfill_cmd(ctx):
+    click.echo(logic_backfill.describe(ctx.env.backfill, as_json=ctx.json))
 
-@backfill_group.command(name="create-migration")
+
+@backfill_group.command(name="create")
```
peternied, line 110: I'm not sure I understand as a user what a 'create' does in this context - is this the right name vs 'prepare' or 'configure'? How is this different from 'start'?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/environment.py
```diff
@@ -0,0 +1,63 @@
...
+    "backfill": {"type": "dict", "required": False},
+    "metrics_source": {"type": "dict", "required": False},
+}
+
+
+class Environment:
+    source_cluster: Optional[Cluster] = None
+    target_cluster: Optional[Cluster] = None
+    backfill: Optional[Backfill] = None
+    metrics_source: Optional[MetricsSource] = None
```
peternied, line 27: If all of these are `None` is that an error, should this class have logic to check this?

peternied (commented):
Looking good, handful of outstanding comments

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/README.md
mikaylathompson: Yes, but I don't think this is the place to do it. By the time you're here, you're really deep in the weeds.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/environment.py
mikaylathompson: There's a requirement in the schema that target_cluster is populated, so from that direction we can be confident that `target_cluster` ends up populated and that might be worth checking at the end of the constructor.

But otherwise, I think it's possible that a user ends up here with nothing else provisioned. That's certainly an unusual case, so maybe we should throw them a warning if they don't have `backfill` OR `replay` OR `metadata_migration` defined, but in general a lot of `None`s is a very possible state.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
mikaylathompson: Yes, I think it could be equivalent to 'prepare'.  I'm not sure which is better.

It's different than `start` because the end state of this command should _not_ be "data is flowing", but just "ready to start (i.e. make data flow)". Specifically, an OSI pipeline needs to be created. This will be a no-op for RFS at this point (though in the future, we could use this to take the config file params and create a task definition, for instance).

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/README.md
mikaylathompson: Yup, adding it.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
```diff
@@ -93,42 +98,75 @@ def start_replayer_cmd(ctx):
...
+@backfill_group.command(name="stop")
 @click.option('--pipeline-name', default=None, help='Optionally specify a pipeline name')
 @click.pass_obj
-def stop_migration_backfill_cmd(ctx, pipeline_name):
-    """Stop migration action"""
-    ctx.env.backfill.stop(pipeline_name=pipeline_name)
+def stop_backfill_cmd(ctx, pipeline_name):
+    exitcode, message = logic_backfill.stop(ctx.env.backfill, pipeline_name=pipeline_name)
+    if exitcode != 0:
+        raise click.ClickException(message)
```
chelma, line 140: Not a blocker, just curious - why raise ClickException rather than anything else?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_base.py
```diff
@@ -0,0 +1,62 @@
...
+    Interface for backfilling data from a source to target cluster.
+    """
+    def __init__(self, config: Dict) -> None:
+        v = Validator(SCHEMA)
+        self.config = config
+        if not v.validate({"backfill": self.config}):
+            raise ValueError("Invalid config file for backfill", v.errors)
+
+    @abstractmethod
+    def create(self, *args, **kwargs) -> CommandResult:
```
chelma, line 36: Not gonna block on it, but if we have a create, we should have a teardown too.

chelma (approved):
Have some non-blocking comments, but overall looks good.  Thanks Mikayla!

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
peternied: I suppose getting this in front of users will generate more feedback and we can iterate on this if need be.

peternied (approved):

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
mikaylathompson: It exits with a non-zero exit code and also displays the error "nicely" --not with a full traceback, etc. I think subclassing that exception to create custom ones could be a good way to go in the future to get more specific errors while keeping the convenience features.

Docs are here: https://click.palletsprojects.com/en/8.1.x/exceptions/#exception-handling

PR #731, opened 2024-06-18
Update readme around end to end testing

peternied:
### Description
Update readme around end to end testing

### Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,16 +4,26 @@ OpenSearch upgrade, migration, and comparison tooling facilitates OpenSearch mig
 
 ## Table of Contents
 
-- [ Supported cluster versions and platforms](#supported-cluster-versions-and-platforms)
+- [OpenSearch upgrades, migrations, and comparison tooling](#opensearch-upgrades-migrations-and-comparison-tooling)
+- [Table of Contents](#table-of-contents)
+- [Supported cluster versions and platforms](#supported-cluster-versions-and-platforms)
+  - [Supported Source and Target Versions](#supported-source-and-target-versions)
+  - [Supported Source and Target Platforms](#supported-source-and-target-platforms)
 - [Build and deploy](#build-and-deploy)
-    - [Local deployment](#local-deployment)
-    - [AWS deployment](#aws-deployment)
+  - [Local deployment](#local-deployment)
+  - [AWS deployment](#aws-deployment)
 - [Developer contributions](#developer-contributions)
-    - [Pre-commit hooks](#pre-commit-hooks)
-    - [Traffic Capture Proxy and Replayer](#traffic-capture-proxy-and-replayer)
-    - [Running Tests](#running-tests)
+  - [Pre-commit hooks](#pre-commit-hooks)
+  - [Traffic Capture Proxy and Replayer](#traffic-capture-proxy-and-replayer)
+  - [Fetch Migration](#fetch-migration)
+  - [Testing](#testing)
+    - [Component and Unit testing](#component-and-unit-testing)
+    - [Cloud end to end testing](#cloud-end-to-end-testing)
+    - [Local end to end testing](#local-end-to-end-testing)
 - [Security](#security)
 - [License](#license)
+- [Releasing](#releasing)
+- [Publishing](#publishing)
 
 ## Supported cluster versions and platforms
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,12 +69,71 @@ The FetchMigration directory hosts tools that simplify the process of backfillin
...
+        MC --> SC
+        MC --> TC
+    end
+```
+
+#### Local end to end testing
+
+Locally performing end to end test cases is done slightly differently, by subtituting cloud services for the local file system, and TestContainers to provider source and target migration clusters its much faster to validate different configurations and versions.  By running inside of a JUnit JVM this also allows deep access into the running historical worker for monitoring or fault injection.  When test cases themselves do not require full source or target clusters using mock endpoints or data can greatly accelerate test speed.
+
+These tests are all be run with the command `./gradlew test` from the root directory.
```
peternied, line 113: Hmm, what about some of the python code we have, seems like we are using the python test runner - does that apply here or is do we think that is basically covered in the component/unit test kind of area.

README.md
```diff
@@ -59,12 +69,71 @@ The FetchMigration directory hosts tools that simplify the process of backfillin
...
+
+        MC --> ECS
+        MC --> SC
+        MC --> TC
+    end
+```
+
+#### Local end to end testing
+
+Locally performing end to end test cases is done slightly differently, by subtituting cloud services for the local file system, and TestContainers to provider source and target migration clusters its much faster to validate different configurations and versions.  By running inside of a JUnit JVM this also allows deep access into the running historical worker for monitoring or fault injection.  When test cases themselves do not require full source or target clusters using mock endpoints or data can greatly accelerate test speed.
```
mikaylathompson, line 111: nit: subtituting -> substituting

mikaylathompson: and "**using** Test Containers to **provide** source and target migration clusters**, it's**"

README.md
mikaylathompson, line 1: Do you see a role for docker-level local end to end tests? In particular, we have generally prioritized making the docker/opensource tooling an equal citizen with the AWS-specific tooling. In this document, it looks like the docker solution is not explicitly tested (my read of this is that the local testing is effectively testing, e.g. RFS, using containers in a supporting role, not thinking of it as testing the docker solution as an e2e system)

README.md
mikaylathompson: Currently, we use python tests both for unit testing (e.g. console lib or fetch), as well as to run our existing end to end tests. It might be possible to run the latter from within gradle (though I'm not sure we gain much from that), but my understanding from this diagram (and the emphasis on running the tests in the JVM) is that you're suggesting rewriting/porting/scrapping those tests. Is that accurate? If so, running python tests might not be that much of a concern.

Is that what you're asking here, or are you wondering if there's a role that we're using python for that java tests couldn't fill?

README.md
peternied: Thank you! Btw I found out I wasn't using a spell checker in my IDE - fixed!

README.md
peternied: Good distinction, python is well suited for the defining the console link CLI and API.  This code is hard to test beyond the unit level with anything other than the real control plane systems (be it a cloud service, kuberneties, or local docker) and java-based migration tools.  What do you think of this?

I guess I'm saying that testing the python code in an end to end way involves `Cloud end to end testing` kind of structure rather than another category - I am curious we think there is a way to test the python codebase that is missing here. (Fake java migration tools that spit out errors?)

README.md
peternied: @mikaylathompson  There is a huge amount of utility for local testing to be up in docker, my concern is about how much time we'd spend getting Docker to work instead of Kubernetes which handles abstractions a cleaner level.  Its an interesting problem because today we are using the docker based end-to-end testing as an entry before the 'cloud' testing in jenkins ci stack.

Maybe you've got an idea for a section that describes what/how docker fits in, want to write something up and submit a PR to my repo or just make a comment here?

mikaylathompson: I'm grouping Docker and Kubernetes together as the open source solution. I'm fine with postponing the actual testing until we've stabilized that solution, but I don't see any mention of how either of those fits into this e2e testing framework.

README.md
```diff
@@ -59,12 +69,71 @@ The FetchMigration directory hosts tools that simplify the process of backfillin
...
 
-More documentation on this test script can be found here:
-[End-to-End Testing](test/README.md)
+#### Component and Unit testing
+
+While end to end tests can be very useful for testing integration boundaries, for simplier components and functions it is advised to write junit tests that focus on edge cases.  When adding functionality its advised to start at these lower level test and then see what end to end tests might need to be included or modified to check for coverage if needed.
+
+#### Cloud end to end testing
+
+These tests are setup once with a deployment into a cloud ecosystem, because full infrastructure is inheriantly expensive monitarily and time wise - there will be only a small number of tests that make sure the migrations tools are working as expected.  In this example diagram, tests to validate historical migration from a snapshot use the full cloud resources.  Control of the test scenarios is largely through the deployed migration console.
```
mikaylathompson, line 82: nits: inheriantly -> inherently; monitarily -> monetarily;

README.md
mikaylathompson: It seems to fit in between the cloud and local e2e solutions as you've laid them out. It _could_ be done locally, but doesn't really (I think) make sense to do it as part of the JVM. If you're thinking of it as a cloud thing, that's fine with me, but I think it deserves to be called out somwehere.

README.md
mikaylathompson: Yep, I'm fine with that. I think we definitely could (should?) do some testing of the CLI and API that involves fake/toy containers running locally , but I don't think it's a particularly high priority to make that a local thing instead of a cloud-level thing.

peternied:
@mikaylathompson I've been thinking about where we go with this PR and what it means to merge it.  While it is useful to call out these kinds of ideals, as you mention the gap in 'local' vs 'cloud' testing where we might use docker / kubernetes is pretty big.

I'm going to close out this PR as it doesn't really change anything about our existing systems.  As we look to do the 'grand' refactor to make this project more straight forward we can revisit this topic.

PR #734, opened 2024-06-18
[Console] Populate services.yaml in CDK for ECS RFS backfill 

mikaylathompson:
## Note
This is based off of #728, so the diff is much smaller than it appears here -- it should just be the final 3 commits. **It should not be merged until that one is.** At that point, I'll merge from main if necessary and ensure that the diff shrinks to just the incremental commits.

### Description
Backfill is added a full concept in #728, which also defines the structure for the backfill entry in the `services.yaml` file.

In this PR, that structure is filled out for the case of RFS running on ECS and deployed via CDK.

I also realized that `cluster_name` needs to be a required field because it will be necessary for the `ecs update-service` calls made by the RFS-on-ECS backend adapter.

### Issues Resolved
[MIGRATIONS-1777](https://opensearch.atlassian.net/browse/MIGRATIONS-1777)

### Testing
Commits add unit tests. Also manually tested via `./awsE2ESolutionSetup.sh`.

### Check List


deployment/cdk/opensearch-service-migration/lib/migration-services-yaml.ts
```diff
@@ -10,16 +10,44 @@ export class MetricsSourceYaml {
...
+    backfill: RFSBackfillYaml | OSIBackfillYaml;
 
     stringify(): string {
         return yaml.stringify({
             source_cluster: this.source_cluster,
             target_cluster: this.target_cluster,
-            metrics_source: this.metrics_source
+            metrics_source: this.metrics_source,
+            backfill: this.backfill
+            ? (this.backfill instanceof RFSBackfillYaml
```
peternied, line 47: Its strange that this logic is in `ServicesYaml`.  By using polymorphism the different types can self-describe how they are formatted as json internally

peternied (approved):

PR #736, opened 2024-06-19
Fixed a few bugs in the gradle setup for rfs

chelma:
### Description
* Fixed a few RFS gradle bugs
* Switched RFS Docker Compose back to ES 7.10
* ES 7.17 test source container is not standing up correctly

### Issues Resolved
N/A

### Testing
manual testing

### Check List

### DIFF
--- a/RFS/build-preloaded-source-image.gradle
+++ b/RFS/build-preloaded-source-image.gradle
@@ -24,10 +24,9 @@ def createNetworkTask = task createNetwork(type: Exec) {
 }
 task createInitialElasticsearchContainer(type: DockerCreateContainer) {
     dependsOn createNetwork, buildDockerImage_emptyElasticsearchSource_7_17
-    targetImageId 'migrations/emptyElasticsearchSource_7_17:latest'
+    targetImageId 'migrations/empty_elasticsearch_source_7_17:latest'
     containerName = "elasticsearch-${uniqueId}"
     hostConfig.network = myNetworkName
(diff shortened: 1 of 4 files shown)


AndreKurait (approved):
Thanks for taking a look. Let's quickly follow up on if we can get 7_17 working for the demo

RFS/build-preloaded-source-image.gradle
```diff
@@ -64,14 +63,14 @@ def sourceContainerCommitTask = task commitSourceContainer() {
...
-                .withRepository("elasticsearch_rfs_source")
+                .withRepository("migrations/elasticsearch_rfs_source")
                 .withTag("latest")
                 .exec()
     }
 }
 
 task removeClientContainer(type: DockerRemoveContainer) {
-    dependsOn waitClientContainer
+    dependsOn commitSourceContainer
```
gregschohn, line 73: why would you need to wait for the source to be committed?  Shouldn't the client finishing be independent of when the sync happens?

gregschohn (approved):

RFS/build-preloaded-source-image.gradle
chelma: Good question.  We probably don't need to do this, but it made the sequence of operations clearer in my mind when I was trying to figure out why the `commitSourceContainer` task didn't appear to be firing.  In reality the commit task was firing, it just didn't report anything to stdout.

PR #739, opened 2024-06-19
Build reindex from snapshot via new document project

gregschohn:
### Description

### Issues Resolved

### Testing

### Check List


RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -0,0 +1,401 @@
...
+                "      \"status\": {\n" +
+                "        \"type\": \"keyword\",\n" +
+                "        \"norms\": false\n" +
+                "      }\n" +
+                "    }\n" +
+                "  }\n" +
+                "}\n";
+
+        var response = httpClient.makeJsonRequest(PUT_METHOD, INDEX_NAME, null, body);
+        if ((response.getStatusCode() / 100) != 2) {
```
peternied, line 97: This is cute - but hard to understand on first glance.  If we are going through the trouble of making our own http client abstraction it could at the very least have a helper method for functionality like this

peternied: E.g. `response.wasNotSuccessful()`

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -0,0 +1,401 @@
...
+    public WorkAcquisitionOutcome createOrUpdateLeaseForWorkItem(String workItemId, Duration leaseDuration)
+            throws IOException {
+        var startTime = Instant.now();
+        var updateResponse = createOrUpdateLeaseForDocument(workItemId, leaseDuration.toSeconds());
+        var resultFromUpdate = getResult(updateResponse);
+
+        if (resultFromUpdate == DocumentModificationResult.CREATED) {
+            return new WorkItemAndDuration(workItemId, startTime.plus(leaseDuration));
+        } else {
+            final var httpResponse = httpClient.makeJsonRequest(GET_METHOD, INDEX_NAME + "/_doc/" + workItemId,
```
peternied, line 205: Why do we need a second request? I feel like I'm missing something

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -0,0 +1,401 @@
...
+    @NonNull
+    public WorkAcquisitionOutcome createOrUpdateLeaseForWorkItem(String workItemId, Duration leaseDuration)
+            throws IOException {
+        var startTime = Instant.now();
+        var updateResponse = createOrUpdateLeaseForDocument(workItemId, leaseDuration.toSeconds());
+        var resultFromUpdate = getResult(updateResponse);
+
+        if (resultFromUpdate == DocumentModificationResult.CREATED) {
+            return new WorkItemAndDuration(workItemId, startTime.plus(leaseDuration));
+        } else {
```
peternied, line 204: Since you are return in the if statement it would be easier to read if you got rid of the else block so all DocumentModificationResult checks were on the same vertical indent

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -0,0 +1,401 @@
...
+        if (DocumentModificationResult.UPDATED != DocumentModificationResult.parse(resultStr)) {
+            throw new IllegalStateException("Unexpected response for workItemId: " + workItemId + ".  Response: " +
+                response.toDiagnosticString());
+        }
+    }
+
+    enum UpdateResult {
+        SUCCESSFUL_ACQUISITION,
+        VERSION_CONFLICT,
+        NOTHING_TO_AQUIRE
```
peternied, line 259:
```suggestion
        NOTHING_TO_ACQUIRE
```

RFS/src/main/java/com/rfs/cms/ProcessManager.java
```diff
@@ -0,0 +1,51 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
+
+/**
+ * This class takes an expiration time and guarantees that the process will NOT run past that expiration
+ * time unless the work is marked as complete before that expiration.  This class may, but does not need to,
+ * synchronize its clock with an external source of truth for better accuracy.
+ */
+public class ProcessManager {
```
peternied, line 20: In the language of this code base this is a WorkManager - process isn't a noun, right?

RFS/src/main/java/com/rfs/cms/IWorkCoordinator.java
```diff
@@ -0,0 +1,80 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.ToString;
+
+import java.io.IOException;
+import java.time.Duration;
+import java.time.Instant;
+
+public interface IWorkCoordinator extends AutoCloseable {
```
peternied, line 12: I do not yet grok this interface, what do you think about adding readme /w diagram?

RFS/src/main/java/com/rfs/cms/ReactorHttpClient.java
```diff
@@ -0,0 +1,97 @@
...
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+import java.util.function.BiFunction;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+@Slf4j
+public class ReactorHttpClient implements AbstractedHttpClient {
```
peternied, line 24: Why use apache and reactor, why can't we just use one?

RFS/src/main/java/com/rfs/cms/ScopedWorkCoordinatorHelper.java
```diff
@@ -0,0 +1,43 @@
+package com.rfs.cms;
+
+import lombok.NonNull;
+import lombok.extern.slf4j.Slf4j;
+
+import java.io.IOException;
+
+@Slf4j
+public class ScopedWorkCoordinatorHelper {
```
peternied, line 9: 'Helper' doesn't mean anything, lets nix it, as has been done in usages of this class
```suggestion
public class ScopedWorkCoordinator {
```

RFS/src/main/java/com/rfs/cms/ScopedWorkCoordinatorHelper.java
```diff
@@ -0,0 +1,43 @@
...
+    public final IWorkCoordinator workCoordinator;
+    final ProcessManager processManager;
+
+    public ScopedWorkCoordinatorHelper(IWorkCoordinator workCoordinator, ProcessManager processManager) {
+        this.workCoordinator = workCoordinator;
+        this.processManager = processManager;
+    }
+
+    public interface WorkItemGetter {
+        @NonNull IWorkCoordinator.WorkAcquisitionOutcome apply(IWorkCoordinator wc);
```
peternied, line 20:
```suggestion
        @NonNull IWorkCoordinator.WorkAcquisitionOutcome get(IWorkCoordinator wc);
```

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -16,37 +20,64 @@
...
+        Document value;
+    }
+    public class Done extends DocumentOrDone {}
 
-                return Flux.range(0, reader.maxDoc()) // Extract all the Documents in the IndexReader
-                .handle((i, sink) -> {
-                    Document doc = getDocument(reader, i);
+    public BlockingQueue<DocumentOrDone> packageDocumentsIntoQueue() {
+        var docQueue = new ArrayBlockingQueue<DocumentOrDone>(NUM_DOCUMENTS_BUFFERED, true);
+        new Thread(() -> {
```
peternied, line 38: Lets source the thread from specific threadpool

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -16,37 +20,64 @@
...
-                } catch (IOException e) {
-                    logger.error("Failed to close IndexReader", e);
-                    throw Lombok.sneakyThrow(e);
+                    var item = queue.take(); // TODO - this is still a blocking call!
+                    if (item instanceof Done) {
+                        sink.complete();
+                        break;
+                    }
+                    sink.next(((DocumentHolder)item).value);
+                } catch (InterruptedException e) {
```
peternied, line 66: Lets also log this when it occurs

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -16,37 +20,64 @@
...
+                if (queue.isEmpty() && Thread.currentThread().isInterrupted()) {
+
+                    break;
                 }
             }
-        );
+        });
     }
 
+    @SneakyThrows
```
peternied, line 80: Why sneaky throws, this already throws IOException?

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
```diff
@@ -1,74 +1,81 @@
...
-        return logger;
-    }
+        try (var unpacker = unpackerFactory.create(shardMetadata)) {
+            var reader = readerFactory.apply(unpacker.unpack());
+            Flux<Document> documents = reader.readDocuments();
 
-    public static class DocumentsMigrationPhaseFailed extends Runner.PhaseFailed {
-        public DocumentsMigrationPhaseFailed(GlobalState.Phase phase, WorkerStep nextStep, Optional<CmsEntry.Base> cmsEntry, Exception e) {
-            super("Documents Migration Phase failed", phase, nextStep, cmsEntry, e);
+            final int finalShardId = shardMetadata.getShardId(); // Define in local context for the lambda
```
peternied, line 71: I don't think this comment is accurate, no?

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
```diff
@@ -1,74 +1,81 @@
...
+            reindexer.reindex(shardMetadata.getIndexName(), documents)
+                    .doOnError(error -> log.error("Error during reindexing: " + error))
+                    .doOnSuccess(done -> log.info("Reindexing completed for Index " + shardMetadata.getIndexName() +
+                                    ", Shard " + finalShardId))
+                    // Wait for the reindexing to complete before proceeding
+                    .block();
+            log.info("Docs migrated");
```
peternied, line 78: Is this needed with the doOnSuccess on 74? If yes, it feels like it should include some additional context in the log output.

RFS/src/main/java/com/rfs/worker/IndexAndShard.java
```diff
@@ -0,0 +1,23 @@
...
+import lombok.ToString;
+
+@AllArgsConstructor
+@Getter
+@ToString
+public class IndexAndShard {
+    String indexName;
+    int shard;
+
+    public static String formatAsWorkItemString(String name, int shardId) {
```
peternied, line 14: Throw something up if name contains "__"

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
```diff
@@ -0,0 +1,179 @@
...
+                Assertions.assertNull(nextWorkItem);
+            }
+
+            Thread.sleep(expiration.multipliedBy(2).toMillis());
+        }
+    }
+
+    static AtomicInteger nonce = new AtomicInteger();
+    @SneakyThrows
+    private static Void getWorkItemAndVerity(String workerSuffix, ConcurrentHashMap<String, String> seenWorkerItems,
```
peternied, line 162:
```suggestion
    private static Void getWorkItemAndVerify(String workerSuffix, ConcurrentHashMap<String, String> seenWorkerItems,
```

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
```diff
@@ -0,0 +1,179 @@
...
+            }
+        }
+
+        for (int runs=0; runs<2; ++runs) {
+            final var seenWorkerItems = new ConcurrentHashMap<String, String>();
+            var allFutures = new ArrayList<CompletableFuture<Void>>();
+            final var expiration = Duration.ofSeconds(5);
+            for (int i = 0; i < NUM_DOCS; ++i) {
+                int finalI = i;
+                int finalRuns = runs;
```
peternied, line 141: Not actually final ...?

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
```diff
@@ -0,0 +1,179 @@
...
+                3600, "firstPass_"+ workerSuffix)) {
+            var doneId = DUMMY_FINISHED_DOC_ID + "_" + nonce.incrementAndGet();
+            workCoordinator.createOrUpdateLeaseForDocument(doneId, 1);
+            workCoordinator.completeWorkItem(doneId);
+
+            var nextWorkItem = workCoordinator.acquireNextWorkItem(expirationWindow);
+            log.error("Next work item picked=" + nextWorkItem);
+            Assertions.assertNotNull(nextWorkItem);
+            Assertions.assertNotNull(nextWorkItem.workItemId);
+            Assertions.assertTrue(nextWorkItem.leaseExpirationTime.isAfter(Instant.now()));
```
peternied, line 174: Doesn't seem to make sense to compare 'local' time against the clusters timestamp, that seems like a source of random failure in the future

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -0,0 +1,168 @@
...
+                new OpenSearchClient(esSourceContainer.getUrl(), null));
+        var tempDir = Files.createTempDirectory("opensearchMigrationReindexFromSnapshot_test_snapshot");
+        try {
+            esSourceContainer.copySnapshotData(tempDir.toString());
+
+            var targetClient = new OpenSearchClient(osTargetContainer.getHttpHostAddress(), null);
+            var sourceRepo = new FileSystemRepo(tempDir);
+            migrateMetadata(sourceRepo, targetClient, SNAPSHOT_NAME);
+
+            migrateDocumentsWithOneWorker(sourceRepo, SNAPSHOT_NAME);
```
peternied, line 80: Lets add some simple source vs target checks, like `GET /_count` are equal

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -0,0 +1,168 @@
...
+    final static long TOLERABLE_CLIENT_SERVER_CLOCK_DIFFERENCE_SECONDS = 3600;
+    final static ElasticsearchContainer esSourceContainer =
+            new ElasticsearchContainer(new ElasticsearchContainer.Version("elasticsearch_rfs_source",
+                    "preloaded-ES_7_10"));
+    final static OpensearchContainer<?> osTargetContainer =
+            new OpensearchContainer<>("opensearchproject/opensearch:2.13.0");
+
+    @BeforeAll
+    static void setupSourceContainer() {
+        esSourceContainer.start();
```
peternied, line 62: Think you need to put some things in the source container to expect to be migrated

peternied (commented):
Whew! Done with my initial pass, need to sleep on this change and revisit tomorrow, lots of changes here - great work

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
gregschohn: The source container is the elasticsearch_rfs_source container, which is being built automatically by gradle (every time still :( ).  For now, that includes our 4 OpenSearch Benchmark test runs of 1000 docs / index.  I'd like to improve our cacheable image generation and to include different variants (7.17, 6.8; many more shards; fewer shards), etc.

That will probably be in future PRs though.

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
gregschohn: Right now, the clients (Document migration workers) handle leases by the 'honor system'.  If the clock on the client is way off, we'll have multiple parties stepping on each others toes.  This is a simple sanity check to build some protection against things from completely falling apart.

I'd like to maintain the client time for the kill-switch based off of the cluster time, as it's returned in the http headers of each target request.  The responses may be for different nodes and there could be skew there, so that could be problematic too.

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
gregschohn: whoops - that was an intellij auto-fix for putting a mutating value into a lambda.  The latest code cons-es an effectively final string outside of the lambda to workaround the underlying issue.

RFS/src/main/java/com/rfs/cms/ReactorHttpClient.java
gregschohn: I haven't gotten the ReactorHttpClient to work yet.  I'll add a comment & a jira task.

RFS/src/main/java/com/rfs/cms/ScopedWorkCoordinatorHelper.java
gregschohn: Thanks.  Changed.  (to me), helper means that this isn't taking ownership of anything... however, the class morphed and is no longer just a helper that contains static methods.

RFS/src/main/java/com/rfs/cms/ProcessManager.java
gregschohn: Renamed to LeaseExpireTrigger since that's what it's really doing

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
gregschohn: Fixed

RFS/src/main/java/com/rfs/cms/AbstractedHttpClient.java
```diff
@@ -0,0 +1,55 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.LinkedHashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public interface AbstractedHttpClient extends AutoCloseable {
```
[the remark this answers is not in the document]
peternied: [From offline] We can nix this class when we have a better handle on what http client we are going to use.

RFS/src/main/java/com/rfs/cms/IWorkCoordinator.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    /**
+     *
+     */
+    class NoAvailableWorkToBeDone implements WorkAcquisitionOutcome {
+        @Override
+        public <T> T visit(WorkAcquisitionOutcomeVisitor<T> v) throws IOException {
+            return null;
```
peternied, line 44: This looks like a bug, shouldn't this be `v.onNoAvailableWorkToBeDone`?

RFS/src/main/java/com/rfs/cms/IWorkCoordinator.java
peternied: So there is the problem of allocating work, then there is the problem of defining work to the allocated, having a single interface in one place makes this harder to understand, seems like this would be better suited to be in two interfaces - in the case where there is a dual consumer and producer, it could just implement both interfaces.  What do you think?

RFS/src/main/java/com/rfs/cms/ScopedWorkCoordinatorHelper.java
peternied: Or maybe `tryAcquire(...)` is a better fit?

RFS/src/main/java/com/rfs/cms/IWorkCoordinator.java
gregschohn: I've added a lot more documentation & pulled the core members to the top to improve readability.  I'll try to get Chris's original documentation on this algorithm into the tree.

RFS/src/main/java/com/rfs/cms/ScopedWorkCoordinatorHelper.java
gregschohn: Good point.  Changed.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
gregschohn: I haven't gotten the pub-sub code working yet.  I'll revisit this later.  It isn't actually critical for this PR since I did what I needed to do with Fluxes after all.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
gregschohn: See the above comment - not relevant any more.

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
peternied: There is the general problem of a single source of truth for time I think we don't need to verify that scenario in this case of these tests.  What would we do about a failure at this line - it seems like unactionable failure to me?

I'd rather remove the validation of this concept from these tests and bake them into separate more focused test cases that assure the time management/coordination is working as expected - this seems like a good candidate for a follow up (and I don't think is a code requirement for the moment).  What do you think?

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
gregschohn: Hmmm - yeah, that will work for now because the first arg to Flux.using is a callable, so exceptions are fine :).

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
gregschohn: You're right & I've shed the whole line & inlined it into a logging lambda anyway now.

peternied (approved):
I know we are pushing towards a deadline - and I think we've made huge strides with this PR.  What I would advise to keep moving fast would be to merge as is and then rather than create follow up items - just keep working on the outstanding comments on this PR until they are all resolved.  What do you think?

I've just done a pass of them and there are far fewer that we need to put this set of changes to 'rest'.  Making github issues or jira tasks seems like low-value overhead when we've got the context in this PR.

RFS/src/main/java/com/rfs/worker/IndexAndShard.java
gregschohn: Thanks!  Done

RFS/src/main/java/com/rfs/worker/DocumentsRunner.java
gregschohn: Probably not, but that's what was in the previous incarnations of the code, so I'll leave it for now to make sure that nobody else is expecting that line to be present.

RFS/src/test/java/com/rfs/cms/WorkCoordinatorTest.java
gregschohn: 👍

DocumentsFromSnapshotMigration/build.gradle
```diff
@@ -143,15 +144,30 @@ tasks.named("buildDockerImage_elasticsearchRFSSource") {
...
+        excludeTags 'longTest'
+    }
+    jacoco {
+        enabled = false
+    }
+}
+
+task slowTest(type: Test) {
+    useJUnitPlatform()
+    dependsOn buildDockerImage_elasticsearchRFSSource
```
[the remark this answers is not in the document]
gregschohn: Right now, I think that the time is roughly the same.  It takes about 2 minutes to load the test data that I'm using.  The hope with the prebuilt image is that if I'm running repeatedly, I don't need to keep regenerating my source data.  I'll work on making the prebuilt stuff cacheable (it isn't now, which admittedly sucks), but the additional tax is just a couple extra docker operations (and an extra startup cycle of ES).

RFS/src/main/java/com/rfs/worker/SnapshotRunner.java
```diff
@@ -10,8 +10,8 @@ private SnapshotRunner() {}
 
     protected static void waitForSnapshotToFinish(SnapshotCreator snapshotCreator) throws InterruptedException {
         while (!snapshotCreator.isSnapshotFinished()) {
-            log.info("Snapshot not finished yet; sleeping for 5 seconds...");
-            Thread.sleep(5000);
+            log.info("Snapshot not finished yet; sleeping for 1 seconds...");
+            Thread.sleep(1000);
```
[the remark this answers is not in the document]
gregschohn: whoops - I had copied that code in.  Values are fixed now

RFS/src/main/java/com/rfs/cms/IWorkCoordinator.java
gregschohn: 👍

PR #747, opened 2024-06-19
Reenable large request test

AndreKurait:
### Description
Reenable large request test
* Category: Test fix
* Why these changes are required? This is a required feature so should be tested. 
* What is the old behavior before changes and new behavior after changes? Test disabled, now test enabled and working

### Issues Resolved
[MIGRATIONS-1798](https://opensearch.atlassian.net/browse/MIGRATIONS-1798)

### Testing
Github Actions E2E Workflow

### Check List

### DIFF
--- a/test/tests.py
+++ b/test/tests.py
@@ -405,21 +405,20 @@ class E2ETests(unittest.TestCase):
         finally:
             proxy_single_connection_session.close()
 
-    def test_0008_largeRequest(self):
-        for i in range(10):
-            index_name = f"test_0008_{self.unique_id}_{i}"
-            doc_id = "1"
+    def test_0008_largeRequest_replayedToTarget(self):
+        index_name = f"test_0008_{self.unique_id}"
+        doc_id = "1"
 
-            # Create large document, 99MiB
-            # Default max 100MiB in ES/OS settings (http.max_content_length)
-            large_doc = generate_large_doc(size_mib=99)
+        # Create large document, 99MiB
+        # Default max 100MiB in ES/OS settings (http.max_content_length)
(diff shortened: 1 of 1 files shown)


test/tests.py
```diff
@@ -404,7 +404,6 @@ def test_0007_timeBetweenRequestsOnSameConnection(self):
         finally:
             proxy_single_connection_session.close()
 
-    @unittest.skip
     def test_0008_largeRequest(self):
```
peternied, line 407: Since this test was flaky before, while in draft can you make the test repeat itself 10x and if we see they all pass, then we could remove the repeat logic and re-enable.  What do you think?

peternied (changes requested):

test/tests.py
peternied: Thanks, lets get it back in!

AndreKurait: https://github.com/opensearch-project/opensearch-migrations/actions/runs/9602550138/job/26483635785?pr=747

AndreKurait: Wasn't able to get it working with 99MIB, but was with 50 so moving to that
https://github.com/opensearch-project/opensearch-migrations/actions/runs/9602018665

peternied: If we've got a change that runs with 10x iterations I'm ok merging that, with these recent iterations don't build as much confidence.

peternied:
@AndreKurait Looks like this has been outstanding for a bit, care to clean this test up so we can use it again?

peternied:
@AndreKurait I'm going to close this out to keep our PR queue smaller, feel free to reopen if you pick this back up.

PR #754, opened 2024-06-21
Support local snapshots from CLI fro SnapshotCreate & MetadataMigration

mikaylathompson:
### Description
This allows testing snapshots and metadata migration locally (and long-term supporting it in our docker option)

### Issues Resolved
My development speed

### Testing
Manual -- we're flying pretty blind here.
Steps for manual testing:
- disable password verification on source and target cluster
- create a shared volume in the docker-compose file
- for source cluster in docker-compose:
  - mount shared volume (in this case `/snapshot`)
  - set in environment: path.repo=/snapshot
- for migration console in docker-compose
  - mount shared volume
  - run `chmod -R a+rwX /snapshot/` (there is absolutely a better way to do this, which is in part why I'm not checking in the pieces of this test setup, but it works for this)
  - .`/createSnapshot/bin/CreateSnapshot --snapshot-name 2024-06-21-test-0001 --file-system-repo-path /snapshot/test --source-host http://source:9200`
  - `./metadataMigration/bin/MetadataMigration --snapshot-name 2024-06-21-test-0001 --file-system-repo-path /snapshot/test/ --target-host http://opensearchtarget:9200`

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/FileSystemRepo.java
+++ b/RFS/src/main/java/com/rfs/common/FileSystemRepo.java
@@ -1,9 +1,5 @@
 package com.rfs.common;
 
-import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
-import software.amazon.awssdk.regions.Region;
-import software.amazon.awssdk.services.s3.S3AsyncClient;
-
 import java.io.IOException;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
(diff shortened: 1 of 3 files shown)


RFS/src/main/java/com/rfs/common/FileSystemRepo.java
```diff
@@ -1,5 +1,9 @@
 package com.rfs.common;
 
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
```
gregschohn, line 5: Not needed, right?

gregschohn (approved):

RFS/src/main/java/com/rfs/common/FileSystemRepo.java
mikaylathompson: huh, yeah, weird. Removing them.

PR #761, opened 2024-06-24
Cat command always shows all indices

peternied:
### Description
Between different versions of ES and OS they have different system indices definitions, this can lead to the cat outputs looking different that is confusing

### Testing
Tested the cat indices command on a ES7.10 cluster and OS2.14 cluster, no updates to existing test infrastructure

### Check List


mikaylathompson:
we have the cat indices endpoint being mocked here: https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cli.py#L156-L162, this should just be updated to match

mikaylathompson (approved):

PR #767, opened 2024-06-25
Shutdown executors created by RFS LeaseExpireTrigger and KafkaTrafficCaptureSource

gregschohn:
Both of those classes used the DefaultThreadFactory, which uses daemon threads, which will keep a process alive even after main() has finished.

### Description

* Category Bug fix
* Why these changes are required? Even due to top-level exceptions and main returning, processes could hang indefinitely as the process would wait for daemon threads to finish (which would never happen).
* What is the old behavior before changes and new behavior after changes? RfsMigrateDocuments would hang when run from gradle or a container.  Nowe it won't

### Issues Resolved
https://opensearch.atlassian.net/jira/software/c/projects/MIGRATIONS/boards/1?selectedIssue=MIGRATIONS-1809

### Testing
manual testing and existing tests from the build.  New tests are necessary, but that will be follow-up work (see [Jira task](https://opensearch.atlassian.net/jira/software/c/projects/MIGRATIONS/boards/1?selectedIssue=MIGRATIONS-1809))

### Check List


chelma:
I'm currently working on the testing for this change.

chelma (approved):
Thanks, Greg!

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -95,6 +98,7 @@ public KafkaTrafficCaptureSource(@NonNull RootReplayerContext globalContext,
         this.behavioralPolicy = behavioralPolicy;
         kafkaConsumer.subscribe(Collections.singleton(topic), trackingKafkaConsumer);
         kafkaExecutor = Executors.newSingleThreadExecutor(new DefaultThreadFactory("kafkaConsumerThread"));
+        isClosed = new AtomicBoolean(false);
```
peternied, line 101: Nit; join declare and assignment into a single line

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -33,6 +33,8 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.RejectedExecutionException;
```
peternied, line 36: Nit: Looks like an unused import

peternied (approved):
Since this is mostly declarative doesn't seem like a test is needed.

Was the underlying problem that the process won't exit because there were outstanding thread pools - seems like that doesn't need a full setup/teardown to verify, but thanks for the follow up task.

AndreKurait (approved):

gregschohn:
> Since this is mostly declarative doesn't seem like a test is needed.
> 
> Was the underlying problem that the process won't exit because there were outstanding thread pools - seems like that doesn't need a full setup/teardown to verify, but thanks for the follow up task.

I agreed with that concept when I first wrote FullTest.  Making sure that processes die is a critical component to this system's design (and same for the capture proxy and replayer).  Since we got into trouble from an unknown, it seems prudent to put a functional test in place now.  Sure, we could verify that there aren't any more daemon threads, etc - but there could be other gotchas, maybe even in native code from a dependency, so running a full-scale, "does the process terminate task", while pedantic, should give us more peace of mind & will help us to find what might otherwise be tricky issues to track down.

PR #769, opened 2024-06-26
Bugfixes in RFS work coordination update queries to claim leases

gregschohn:
The easy one which probably wasn't ever causing issues is to convert 409 conflict responses immediately to a VERSION_CONFLICT result rather than waiting to parse the response body (& possibly be surprised). The other was to set the operation to noop for the update_by_query result when the conditional checks (like whether the new expiration is > the old expiration).  That removes a big (& only?) pressure where the subsequent call to get the assigned document would return 0 results - which made sense since no document was actually updated, even though the update_by_query reported that one was.

### Description

* Category Bug fix
* Why these changes are required?  To make sure that we're making forward progress on work assignments and not creating unnecessary noise.
* What is the old behavior before changes and new behavior after changes? We'd get lots of assertions that allegedly updated documents couldn't be found.  Now we don't check because we won't falsely mark them as being updated.

### Issues Resolved
https://opensearch.atlassian.net/jira/software/c/projects/MIGRATIONS/boards/1?assignee=63d1918f69c7ae3958d1f57f&selectedIssue=MIGRATIONS-1810

### Testing
Ran through the FullTest, observing none of the "The query for the assigned work document returned 0" errors.  The test still doesn't assert that that shouldn't happen though.

### Check List


RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -400,6 +402,9 @@ UpdateResult assignOneWorkItem(long expirationWindowSeconds) throws IOException
 
         var response = httpClient.makeJsonRequest(POST_METHOD,  INDEX_NAME + "/_update_by_query?refresh=true&max_docs=1",
                 null, body);
+        if (response.getStatusCode() == 409) {
+            return UpdateResult.VERSION_CONFLICT;
```
AndreKurait, line 406: I think we need to check for resultTree['noop'] > 0 on the below elseIf

gregschohn: Thanks - addressed in the latest commit (if I understood the concern)

AndreKurait (approved):
Thanks for this! I deployed this within my AWS Environment and seeing it resolved my issues with concurrency

peternied:
> Failing after 1s — 4.00% of diff hit (target 68.51%)

Nothing is calling that code that is changed, can we add some validation since there are some clear off-by-one errors possible?

PR #773, opened 2024-06-26
Test multiple source versions through Document Migration's 'FullTest'

gregschohn:
### Description

The big change here is how we build preloaded images for testing, which requires coordination between a source cluster and a client container.  We had done this with gradle, but working out the dependency graph such that we only ran the docker commands when images were out of date was challenging.  It also was significant gradle logic, putting test logic within the build system.

Now the images are orchestrated and built within java (test) code w/ the help of testcontainers and github's docker library.  There were challenges to get the newly created image to NOT be deleted by the testcontainer framework (unset a label).  Another big hurdle has to do with how elasticsearch, especially 7.17, is shut down.  It usually leaves lock files through the data directory tree, which prevents the process from starting up.  

The workaround there is to launch the initial process separate from the container's main task (so that its death doesn't kill the container and prevent us from committing it), do the client work to fill it, tear the server down, cleaning up leftover files along the way, then run the commit before tearing the container down completely.

* Category Test Enhancements
* Why these changes are required? To test more versions of clusters and to do it faster
* What is the old behavior before changes and new behavior after changes? We'd always rebuild images, now we cache them.  Before, we weren't able to cache a preloaded 7.17 image.  Now we can.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1811

### Testing

### Check List


RFS/src/testFixtures/java/com/rfs/framework/ElasticsearchContainer.java
```diff
@@ -2,39 +2,77 @@
 
 import java.io.File;
 import java.time.Duration;
+import java.util.Map;
 
+import lombok.EqualsAndHashCode;
 import lombok.extern.slf4j.Slf4j;
 import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
```
peternied, line 11: Lets import real thing rather that pull in this alternative version.

RFS/src/testFixtures/java/com/rfs/framework/ElasticsearchContainer.java
```diff
@@ -2,39 +2,77 @@
...
     public static final String CLUSTER_SNAPSHOT_DIR = "/usr/share/elasticsearch/snapshots";
-    public static final Version V7_10_2 =
-            new Version("docker.elastic.co/elasticsearch/elasticsearch:7.10.2", "7.10.2");
+    public static final Version ES_V7_10_2 =
+            new ElasticsearchVersion("docker.elastic.co/elasticsearch/elasticsearch-oss:7.10.2", "7.10.2");
+    public static final Version ES_V7_17 =
+            new ElasticsearchVersion("docker.elastic.co/elasticsearch/elasticsearch:7.17.22", "7.17.22");
+
+    public static final Version OS_V1_3_15 =
+            new OpenSearchVersion("opensearchproject/opensearch:1.3.16", "1.3.16");
```
peternied, line 26: This versions numbers aren't aligned 😆

RFS/src/testFixtures/java/com/rfs/framework/ElasticsearchContainer.java
```diff
@@ -2,39 +2,77 @@
...
 import org.testcontainers.containers.wait.strategy.Wait;
+import org.testcontainers.shaded.com.google.common.collect.ImmutableMap;
 import org.testcontainers.utility.DockerImageName;
 
 /**
  * Containerized version of Elasticsearch cluster
  */
 @Slf4j
-public class ElasticsearchContainer implements AutoCloseable {
+public class ElasticsearchContainer extends GenericContainer<ElasticsearchContainer> {
```
peternied, line 18: Want to rename this to something like `SearchClusterContainer` and then delete the OpenSearch variant?

commonDependencyVersionConstraints/build.gradle
```diff
@@ -16,6 +16,11 @@ dependencies {
 
     api group: 'org.slf4j', name: 'slf4j-api', version: '2.0.13'
 
+    api group: 'com.github.docker-java', name: 'docker-java', version: '3.3.6'
```
peternied, line 19: Lets add a shared version number for all of these

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
```diff
@@ -0,0 +1,285 @@
...
+        }
+        if (pullIfUnavailable) {
+            log.warn("Image not found. Pulling image: " + fullName);
+            dockerClient.pullImageCmd(imageName)
+                    .withTag(tag)
+                    .exec(new PullImageResultCallback())
+                    .awaitCompletion(PULL_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+            return getExistingImage(dockerClient, imageName, tag, false);
+        }
+        return null;
```
peternied, line 149: Seems like we should throw here rather than make the caller responsible for a null check

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
```diff
@@ -0,0 +1,285 @@
...
+    private void makeNewImage(DockerClient dockerClient, String imageName, String tag)
+            throws IOException, InterruptedException
+    {
+        var imageResponse = dockerClient.inspectImageCmd(baseSourceVersion.imageName).exec();
+        var originalEntrypoint = imageResponse.getConfig().getEntrypoint();
+        var originalCmd = imageResponse.getConfig().getCmd();
+        final var replacementCommand = Streams.concat(
+                        Stream.of("/bin/sh", "-c"),
+                        Stream.of(Streams.concat(
+                                Stream.of("printenv && "),
```
peternied, line 166: I'm not sure I understand what this command is doing, if my guess is right.  We need to be very careful with environment variable output - will this be in the context of the container or the jenkins/github worker?  If the later, it can include security tokens and

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
```diff
@@ -0,0 +1,285 @@
...
+                                        Stream.of("&", "tail", "-f", "/dev/null"))
+                                .collect(Collectors.joining(" "))))
+                .toArray(String[]::new);
+
+        try (var network = Network.newNetwork();
+             var serverContainer = new ElasticsearchContainer(baseSourceVersion)
+                     .withNetwork(network)
+                     .withNetworkAliases(serverNameAlias)
+                     .withCreateContainerCmdModifier(cmd -> cmd.withEntrypoint(replacementCommand))
+                     .withStartupTimeout(Duration.ofDays(1))) {
```
peternied, line 178: That is too long, maybe 1-5 minute?

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
```diff
@@ -0,0 +1,285 @@
...
+        return image.getId();
+    }
+
+    private int getHashCodeOfImagesAndArgs(DockerClient dockerClient, boolean pullIfUnavailable)
+        throws InterruptedException
+    {
+        return new HashHelper(
+                getImageId(dockerClient, baseSourceVersion.imageName, pullIfUnavailable),
+                getImageId(dockerClient, dataLoaderImageName, pullIfUnavailable),
+                generatorContainerArgs).hashCode();
```
peternied, line 111: Pretty sure you can remove HashHelper entirely
```suggestion
        return Objects.hash(
                getImageId(dockerClient, baseSourceVersion.imageName, pullIfUnavailable),
                getImageId(dockerClient, dataLoaderImageName, pullIfUnavailable),
                generatorContainerArgs);
```

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
gregschohn: whoops!  Thanks.  1 day was REALLY useful so that I could keep the container alive to run experiments for an hour, etc.

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
gregschohn: Thanks - leftover debugging.  Removed

RFS/src/testFixtures/java/com/rfs/framework/PreloadedDataContainerOrchestrator.java
gregschohn: Null is pretty reasonable - see this code
```
if (getExistingImage(dockerClient, imageName, tag, false) == null) {
   makeNewImage...
```

Since I'm acting upon it in a pretty expected path, I'd rather not make it look like an exception.

peternied (commented):
Need a little more time to circle back on this, really cool idea - using docker as the 'replay' point is very useful with complex setups that are reused

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -74,26 +75,43 @@
...
+    public static final String SOURCE_SERVER_ALIAS = "source";
+    public static final int MAX_SHARD_SIZE_BYTES = 64 * 1024 * 1024;
 
     public static Stream<Arguments> makeArgs() {
-        var sourceImageNames = List.of("migrations/elasticsearch_rfs_source");
-        var targetImageNames = List.of("opensearchproject/opensearch:2.13.0", "opensearchproject/opensearch:1.3.0");
+        var sourceImageNames = List.of(
+                makeParamsForBase(SearchClusterContainer.ES_V7_17),
+                makeParamsForBase(SearchClusterContainer.ES_V7_10_2));
+        var targetImageNames = List.of("opensearchproject/opensearch:2.13.0");//, "opensearchproject/opensearch:1.3.0");
```
peternied, line 89: Can we use SearchClusterContainer.OpenSearch... for these references too?

DocumentsFromSnapshotMigration/build.gradle
```diff
@@ -135,12 +133,11 @@ task buildDockerImages {
     for (dockerService in dockerServices) {
         dependsOn "buildDockerImage_${dockerService.projectName}"
     }
-    dependsOn buildDockerImage_elasticsearchRFSSource
 }
 
-tasks.named("buildDockerImage_elasticsearchRFSSource") {
-    dependsOn(':TrafficCapture:dockerSolution:buildDockerImage_elasticsearchTestConsole')
-}
+//tasks.named("buildDockerImage_elasticsearchRFSSource") {
```
peternied, line 138: Could we delete this commented out code?

peternied (approved):
Looks like a great step forward in these tests, 11min runtime on my cloud desktop.

Lets merge this - I think we need review the output itself, I'm seeing what looks like issues in the log output:

#### Unparsable docs?
```
15:08:58.993 ForkJoinPool.commonPool-worker-3 ERROR LuceneDocumentsReader:68 - Unable to parse Document id from Document.  The Document's Fields: 
... ^ repeated hundreds of times...
```

#### First exception for run isn't catching 'unexpected' exceptions
```
15:11:57.018 Test worker INFO  FullTest:158 - First exception for run
com.rfs.RfsMigrateDocuments$NoWorkLeftException: No work items are pending/all work items have been processed.  Returning.
...
```

#### Cleanup the tostring on the test properties

```
[1] com.rfs.framework.SearchClusterContainer$ElasticsearchVersion@e791c8e, migrations/elasticsearch_client_test_console:latest, [/root/runTestBenchmarks.sh, --endpoint, http://source:9200/], opensearchproject/opensearch:1.3.16, 1
```

PR #777, opened 2024-06-26
Two fixes to work coordination

gregschohn:
### Description

Make the check for the setup index more reliable and only scan for work that is currently expired.

The HEAD check has now been rolled into the doUntil block for running setup() so that even 400s from the PUT command can be ignored and retried again in a sane manner.  The HEAD call should be the authoritative call when it returns 200. The second fix is for searching for items to update.  I had been searching for items that weren't even updated yet, causing a lot more pressure to create false positives and inefficiencies all around.

* Category Bug fix
* Why these changes are required? Too much contention/no guarantee on convergence for RFS Work coordination.
* What is the old behavior before changes and new behavior after changes? Lots of thrashing to setup an initial index and a large number of PotentialClockDriftDetectedExceptions bleeding upward.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1810

### Testing
Manual/gradle build

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
+++ b/RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
@@ -71,13 +71,6 @@ public class OpenSearchWorkCoordinator implements IWorkCoordinator {
     }
 
     public void setup() throws IOException, InterruptedException {
-        var indexCheckResponse = httpClient.makeJsonRequest(HEAD_METHOD, INDEX_NAME, null, null);
-        if (indexCheckResponse.getStatusCode() == 200) {
-            log.info("Not creating " + INDEX_NAME + " because it already exists");
-            return;
-        }
-        log.atInfo().setMessage("Creating " + INDEX_NAME + " because it's HEAD check returned " +
-                indexCheckResponse.getStatusCode()).log();
         var body = "{\n" +
                 "  \"settings\": {\n" +
                 "   \"index\": {" +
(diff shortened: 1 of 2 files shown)


RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -123,7 +123,7 @@ public String toString() {
                             return "[ statusCode: " + r.getStatusCode() + ", payload: " + payloadStr + "]";
                         }
                     },
-                    (response, ignored) -> (response.getStatusCode() / 100) != 2);
+                    (response, ignored) -> (response.getStatusCode() / 100) == 2);
```
peternied, line 126: This bug suggests there is a missing test case, can we add a unit test?

peternied (changes requested):

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
gregschohn: agree - I hope to eventually pickup details like this once telemetry is in place.
Assert that the number of retries for this test run was >0, this one was 0, etc.

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
```diff
@@ -123,7 +123,7 @@ public String toString() {
                             return "[ statusCode: " + r.getStatusCode() + ", payload: " + payloadStr + "]";
                         }
                     },
-                    (response, ignored) -> (response.getStatusCode() / 100) != 2);
```
chelma, line 126: Just curious - what was the original logic you were trying to achieve here?  I didn't quite follow your explanation in the PR overview.

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
gregschohn: Functionally, the code and tests converge as it was and is.
It doesn't do it efficiently.  I'd like to get a lot of efficiency numbers in place to confirm that we're working with an expected amount of contention, etc.  

I've opened https://opensearch.atlassian.net/jira/software/c/projects/MIGRATIONS/boards/1?selectedIssue=MIGRATIONS-1812 specifically to cover this case

RFS/src/main/java/com/rfs/cms/OpenSearchWorkCoordinator.java
gregschohn: The test determines when we're done needing to retry.  When this is 'true', the work is deemed to be done.  In other words, we would keep looping until we got something other than a 200.  Notice that the beginning of the method was a HEAD that didn't have the same affliction - so in the worst case scenario, we'd burn a process and be resolved the next time that the program ran

AndreKurait (approved):

peternied (approved):

PR #782, opened 2024-06-27
Enable gradle caching over the whole project

AndreKurait:
### Description
Now that we have multiple projects on the base, moving the TrafficCapture gradle properties to the top level will enable caching over the whole project so code changes in RFS don't trigger a rebuild of traffic capture
 
* Category: Enhancement
* Why these changes are required? Faster build and test times
* What is the old behavior before changes and new behavior after changes? GHA took ~30 min. We will see how long they take after this change is merged

### Issues Resolved
N/A - Teammate request

### Testing
GHA, Local Building

### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -63,6 +63,7 @@ jobs:
       - uses: gradle/actions/setup-gradle@v3
         with:
           gradle-version: ${{ env.gradle-version }}
+          gradle-home-cache-cleanup: true
       - name: Run Gradle Build
         run: ./gradlew build -x test -x TrafficCapture:dockerSolution:build --scan --stacktrace
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/CI.yml
```diff
@@ -63,6 +63,8 @@ jobs:
       - uses: gradle/actions/setup-gradle@v3
         with:
           gradle-version: ${{ env.gradle-version }}
+          gradle-home-cache-cleanup: true
+          cache-read-only: false
```
peternied, line 67: I don't think we want this setting enabled, that would allow non-main branches to push artifacts into the cache.  IMO I think only main should push artifacts
https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#caching-build-state-between-jobs

peternied (commented):
I'm not sure that we need to make any changes to adopt gradle caching, looks like we are already getting cache entries: https://github.com/opensearch-project/opensearch-migrations/actions/caches

Looks like there might be some tests in RFS that don't work when run in parallel, let me know if you need a hand debugging

.github/workflows/CI.yml
AndreKurait: Removed `cache-read-only: false` and parallelism

peternied (approved):

PR #783, opened 2024-06-28
Basic TransformationRule: IndexMappingTypeRemoval

peternied:
### Description
Adding in a formal transformation rule into the codebase, these should be applicable with in RFS, TrafficCapture, or other systems by supporting the `Entity` interface on types that are supported.  Created and wired up IndexMappingTypeRemoval into RFS.

_Note: This is a minimal building block for future metadata / transformation work; replayer support, refactoring existing transformations, version management, rules grouping, audit entries are out of scope for this pull request.  There can be considerable investment in this space built off of this change._

### Testing
Example test cases for the transformer and end to end test to support.

### Check List


RFS/src/testFixtures/java/com/rfs/framework/SearchClusterContainer.java
```diff
@@ -17,60 +18,35 @@
  */
 @Slf4j
 public class SearchClusterContainer extends GenericContainer<SearchClusterContainer> {
-    public static final String CLUSTER_SNAPSHOT_DIR = "/usr/share/elasticsearch/snapshots";
+    public static final String CLUSTER_SNAPSHOT_DIR = "/tmp/snapshots";
```
peternied, line 21: When running on my linux machine the original path caused OS to fail to start because it couldn't access the path.

gregschohn: I thought that this path was for the local container, not the host.

RFS/src/testFixtures/java/com/rfs/framework/SearchClusterContainer.java
```diff
@@ -107,20 +83,34 @@ public String getUrl() {
 
     @Override
     public void close() {
-        log.info("Stopping ElasticsearchContainer version:" + version.prettyName);
+        log.info("Stopping container version:" + version.prettyName);
         log.debug("Instance logs:\n" + this.getLogs());
         this.stop();
     }
 
     public enum INITIALIZATION_FLAVOR {
```
gregschohn, line 91: I'm not convinced that enums storing more than integral values is a good pattern.  In some cases, this seems natural - maybe this is one of them.  Please move this toward the top of the file since it has some static constant policies baked into it.

coreUtilities/src/main/java/org/opensearch/migrations/transformation/CanApplyResult.java
```diff
@@ -0,0 +1,27 @@
...
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+/**
+ * The result after checking if a transformer can be applied to an entity
+ */
+public abstract class CanApplyResult {
+    public final static CanApplyResult YES = new Yes();
+    public final static CanApplyResult NO = new No();
```
gregschohn, line 11: why not use an enum for this?
CanApply or CanApplyEnum might be quicker to grok.  I had thought that you were indicating if a result could be applied (I admit that in retrospect, that's dumb, but I may not be the only one).

coreUtilities/src/main/java/org/opensearch/migrations/transformation/entity/Entity.java
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Common interface for entities that are used to perform transformations
+ *
+ * Notes on performance: ObjectNode requires a fully marshalled json, for
+ * less than 1mb json bodies this will likely be fine; however, if dealing with larger
+ * objects or perf issues are encountered should consider adding an alternative
+ * path for JsonParser.  This might be needed for processing document bodies
+ * specifically.
+ */
+public interface Entity {
```
gregschohn, line 14: Entity is _really_ generic.  I know that you're namespaced, but how likely will this bleed into other areas (like Http, or anything else)?

coreUtilities/src/main/java/org/opensearch/migrations/transformation/entity/Entity.java
```diff
@@ -0,0 +1,19 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Common interface for entities that are used to perform transformations
+ *
+ * Notes on performance: ObjectNode requires a fully marshalled json, for
+ * less than 1mb json bodies this will likely be fine; however, if dealing with larger
+ * objects or perf issues are encountered should consider adding an alternative
+ * path for JsonParser.  This might be needed for processing document bodies
+ * specifically.
```
gregschohn, line 12: This is an interesting comment.  Do you think that we might setup some transformations that are single-pass, where we wouldn't realize the entire DOM-tree in memory?  As it is, the replayer currently uses the JsonParser to at least stream the results in and the JsonEmitter to stream them out, to avoid truly egregious buffering.  I don't know of a streaming transformation specification for json & fear that forcing users/devs to figure that out could be really challenging (I have experience here w/ SAX models for XML & it would be challenging for me).

coreUtilities/build.gradle
```diff
@@ -51,6 +51,12 @@ dependencies {
     implementation group: 'io.opentelemetry', name: 'opentelemetry-exporter-otlp'
     implementation group: 'io.opentelemetry.semconv', name: 'opentelemetry-semconv'
 
+    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-databind'
+    implementation group: 'com.fasterxml.jackson.core', name: 'jackson-core'
```
gregschohn, line 55: These should not be here.  coreUtilities is pulled in by every single package.  For example, there's no reason for netty wire loggers or the proxy to be pulling jackson in.  Please break the transformers into their own common library.

coreUtilities/src/main/java/org/opensearch/migrations/transformation/entity/Index.java
```diff
@@ -0,0 +1,7 @@
+package org.opensearch.migrations.transformation.entity;
+
+/**
+ * Represents an Index object for transformation
+ */
+public interface Index extends Entity {
```
gregschohn, line 6: I know that we're under an opensearch namespace right now, but very little is opensearch specific and there could be other users of a traffic-replay tool and by extension these common coreUtilities.  Longer term, I suspect that  but I suspect you'll want to have a coreTransforms package and some specific ones for ES/OS and then maybe even for specific versions.

coreUtilities/src/main/java/org/opensearch/migrations/transformation/rules/IndexMappingTypeRemoval.java
```diff
@@ -0,0 +1,54 @@
...
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Supports transformation of the Index Mapping types that were changed from mutliple types to a single type between ES 6 to 7
+ */
+public class IndexMappingTypeRemoval implements TransformationRule<Index> {
+
+    @Override
+    public CanApplyResult canApply(final Index index) {
+        final var mappingNode = index.raw().get("mappings");
```
gregschohn, line 19: Won't you need the path (or maybe even headers) to make some determinations here in the general case too?  How will this fit in with the IHttpMessage defined in the existing transformerPlugins?

This seems like it will work great for the Metadata Migration case, but having to create a specific Index type for replays could be an assumption that makes the code more brittle and less performant.

coreUtilities/src/main/java/org/opensearch/migrations/transformation/rules/IndexMappingTypeRemoval.java
```diff
@@ -0,0 +1,54 @@
...
+
+        if (mappingNode == null || mappingNode.isObject()) {
+            return CanApplyResult.NO;
+        }
+
+        // Detect multiple type mappings, eg:
+        // { mappings: [{ foo: {...}}, { bar: {...} }] } } or
+        // { mappings: [{ foo: {...}, bar: {...}] } }
+        if (mappingNode.size() > 1 || mappingNode.get(0).size() > 1) {
```
gregschohn, line 28: path() would likely make this cleaner.  You can string things together, then check if the result exists (or how it has evaluated)

coreUtilities/src/main/java/org/opensearch/migrations/transformation/rules/IndexMappingTypeRemoval.java
```diff
@@ -0,0 +1,54 @@
...
+            return new Unsupported("Multiple mapping types are not supported");
+        }
+
+        // There is a type under mappings, e.g. { mappings: [{ foobar: {...} }] }
+        return CanApplyResult.YES;
```
gregschohn, line 33: Why isn't this just a boolean? (ignore the earlier comments on the class if the answer is that it should be)

RFS/src/test/java/com/rfs/integration/EndToEndTest.java
```diff
@@ -1,67 +1,117 @@
...
+import com.rfs.worker.SnapshotRunner;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
 
 /**
  * Tests focused on setting up whole source clusters, performing a migration, and validation on the target cluster
  */
-@Slf4j
 public class EndToEndTest {
```
chelma, line 36: How would you feel about adding these to the "slow tests"?  E.g. use the `@Tag("longTest")` annotation here.

coreUtilities/src/main/java/org/opensearch/migrations/transformation/TransformationRule.java
```diff
@@ -0,0 +1,22 @@
...
+     * Given an entity can a transformation be run on it
+     *
+     * MUST ALWAYS BE READ ONLY
+     */
+    CanApplyResult canApply(T entity);
+
+    /**
+     * Apply a transformation on the entity
+     * @param entity The entity to be transformed in place
+     * @return true if the entity was updated, or false if no changse were made
```
chelma, line 19: Typo - `changse` -> `changes`

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -1,38 +1,47 @@
...
         // Transform the original "templates", but put them into the legacy "templates" bucket on the target
-        if (root.get("templates") != null) {
-            ObjectNode templatesRoot = (ObjectNode) root.get("templates").deepCopy();
-            templatesRoot.fieldNames().forEachRemaining(templateName -> {
-                ObjectNode template = (ObjectNode) templatesRoot.get(templateName);
-                logger.info("Transforming template: " + templateName);
-                logger.debug("Original template: " + template.toString());
-                TransformFunctions.removeIntermediateMappingsLevels(template);
-                TransformFunctions.removeIntermediateIndexSettingsLevel(template); // run before fixNumberOfReplicas
-                TransformFunctions.fixReplicasForDimensionality(template, awarenessAttributeDimensionality);
```
chelma, line 31: I think I'm missing something, as I don't see where we're calling these now...?

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -47,21 +56,26 @@ public ObjectNode transformGlobalMetadata(ObjectNode root) {
...
+        var copy = index.deepCopy();
+        tranformIndex(copy);
+        return copy;
+    }
+
+    private void tranformIndex(Index index) {
+        logger.debug("Original Object: " + index.raw().toString());
+        var newRoot = index.raw();
 
-        TransformFunctions.removeIntermediateMappingsLevels(newRoot);
```
chelma, line 57: Was this just not necessary?  Or are we calling it elsewhere?

RFS/src/test/java/com/rfs/integration/EndToEndTest.java
```diff
@@ -1,67 +1,117 @@
...
+            assertThat("Shouldn't exist yet, body:\n" + res.getRight(), res.getLeft(), equalTo(404));
+
+            // Action
+            // Migrate indices
+            var indexMetadataFactory = new IndexMetadataFactory_ES_6_8(repoDataProvider);
+            var indexCreator = new IndexCreator_OS_2_11(targetClient);
+            new IndexRunner(snapshotName, indexMetadataFactory, indexCreator, transformer, List.of()).migrateIndices();
+            
+            res = targetClusterOperations.get("/barstool");
+            assertThat(res.getRight(), res.getLeft(), equalTo(200));
```
chelma, line 98: The unit tests for transformations make a ton of sense to cover all the scenarios with a given transformation, but it still feels like we should try this stuff against a real source/target and see what happens.  It feels like we could construct a elaborate library of rarely-used transformations that succeed in our unit testing but fail when run "for real" otherwise.  The big thing here IMO is just exercising the transformation in a single case rather than every case (that's what the unit tests are for).

What do you think?

coreUtilities/src/main/java/org/opensearch/migrations/transformation/rules/IndexMappingTypeRemoval.java
```diff
@@ -0,0 +1,54 @@
...
+    public CanApplyResult canApply(final Index index) {
+        final var mappingNode = index.raw().get("mappings");
+
+        if (mappingNode == null || mappingNode.isObject()) {
+            return CanApplyResult.NO;
+        }
+
+        // Detect multiple type mappings, eg:
+        // { mappings: [{ foo: {...}}, { bar: {...} }] } } or
+        // { mappings: [{ foo: {...}, bar: {...}] } }
```
chelma, line 27: The JSON in your comment is wonky (curly braces don't seem to match up?), I think, and it's making it hard to infer what the intended behavior of your check here is.  Unless I'm just reading it wrong?

coreUtilities/src/main/java/org/opensearch/migrations/transformation/rules/IndexMappingTypeRemoval.java
```diff
@@ -0,0 +1,54 @@
...
+
+import org.opensearch.migrations.transformation.CanApplyResult;
+import org.opensearch.migrations.transformation.CanApplyResult.Unsupported;
+import org.opensearch.migrations.transformation.TransformationRule;
+import org.opensearch.migrations.transformation.entity.Index;
+
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+/**
+ * Supports transformation of the Index Mapping types that were changed from mutliple types to a single type between ES 6 to 7
```
chelma, line 13: Can we get a full example transformation here?  I'm having a hard time understanding the intent even after reading through the code and unit tests a couple times.

chelma (changes requested):
Overall, I like where things are going.  Had some comments/questions though.  Thanks!

RFS/src/testFixtures/java/com/rfs/framework/SearchClusterContainer.java
peternied: This is the path on the container, but it looks like it didn't work on OS (OpenSearch) as I saw startup failures.

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
chelma: This is in that method AFAIK.  Unless you're calling it via the `indexTransformations.forEach()` call, which I don't think you are?

chelma: Oh, I understand now.  It is being invoked in your list of transformations; I just didn't understand what the transform in there was doing.  The new comment in that transformation class clarified things.  Thanks!

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -1,38 +1,47 @@
...
-                TransformFunctions.fixReplicasForDimensionality(template, awarenessAttributeDimensionality);
-                logger.debug("Transformed template: " + template.toString());
-                templatesRoot.set(templateName, template);
+        var originalTemplates = root.get("templates");
+        if (originalTemplates != null) {
+            var templates = mapper.createObjectNode();
+            originalTemplates.fieldNames().forEachRemaining(templateName -> {
+                var templateCopy = (ObjectNode) originalTemplates.get(templateName).deepCopy();
+                var indexTemplate = (Index) () -> templateCopy;
+                transformIndex(indexTemplate);
```
chelma, line 41: Oh, I think I finally see now what's going on.  We're applying a list of transformations for index JSON to all the of the template JSON as well?  This is surprising to me.  Why not have separate transformation lists, which may share similar sets of transforms?

RFS/src/main/java/com/rfs/version_es_6_8/IndexMetadataData_ES_6_8.java
```diff
@@ -68,7 +69,12 @@ public ObjectNode getSettings() {
     }
 
     @Override
-    public ObjectNode toObjectNode() {
+    public ObjectNode raw() {
```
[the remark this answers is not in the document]
peternied: Going with `rawJson`, since the java type doesn't really help too much.

coreUtilities/src/main/java/org/opensearch/migrations/transformation/rules/IndexMappingTypeRemoval.java
peternied: Lets keep this as is for now, we can build the individual transformers up over time.

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
peternied: I've created a separate list and changed how the private function works so it passes if it is a concrete or abstract index.

coreUtilities/build.gradle
peternied: Done!

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -1,38 +1,50 @@
...
-                TransformFunctions.fixReplicasForDimensionality(template, awarenessAttributeDimensionality);
-                logger.debug("Transformed template: " + template.toString());
-                templatesRoot.set(templateName, template);
+        var originalTemplates = root.get("templates");
+        if (originalTemplates != null) {
+            var templates = mapper.createObjectNode();
+            originalTemplates.fieldNames().forEachRemaining(templateName -> {
+                var templateCopy = (ObjectNode) originalTemplates.get(templateName).deepCopy();
+                var indexTemplate = (Index) () -> templateCopy;
+                transformIndex(indexTemplate, IndexType.Concrete);
```
chelma, line 44: Pretty sure this is a bug - shouldn't this be `IndexType.Template`?

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -47,21 +59,38 @@ public ObjectNode transformGlobalMetadata(ObjectNode root) {
...
-        logger.debug("Original Object: " + root.toString());
         logger.debug("Transformed Object: " + newRoot.toString());
-        return newRoot;
-    }    
+    }
+
+    private enum IndexType {
+        Concrete,
+        Template;
+    }
```
chelma, line 95: Not going to block on this, but - is this the correct home for this enumerated type?  I'm guessing you'll be moving this to a more centralized location as soon as you need to do a different transformer.

RFS/src/main/java/com/rfs/version_os_2_11/IndexCreator_OS_2_11.java
```diff
@@ -28,7 +29,8 @@ public Optional<ObjectNode> create(ObjectNode root, String indexName, String ind
         // Assemble the request body
         ObjectNode body = mapper.createObjectNode();
         body.set("aliases", indexMetadata.getAliases());
-        body.set("mappings", indexMetadata.getMappings());
+        // Working around for missing OS_1_3 definition
```
chelma, line 32: Thanks for the comment, but I think I'm even more confused with it.  Why do we care about OS 1.3 in a class specific to OS 2.11?  This feels important because it seems like its pointing to a bigger concept in the code under review I just don't understand.

peternied: Since we aren't using a factory to produce the creators, in the `FullTest.java` we aren't actually using a OS 1.3 set of creators.  The reason the test failed is that in 1.3 the mapping definition is not an object, but an array.  I'm happy to disable the tests if you'd prefer.

RFS/src/main/java/com/rfs/transformers/Transformer_ES_6_8_to_OS_2_11.java
peternied: Yea - this is a good candidate to refactor over time

RFS/src/main/java/com/rfs/version_os_2_11/IndexCreator_OS_2_11.java
chelma: Per conversation - the issue here is that the snapshot format is different for 1.3 than 2.11, and we're trying to use the 2.11 constructs in our tests to support 1.3 without actually writing the code required to formally support 1.3 (e.g. `IndexMetadataData_OS_1_3`, `IndexCreator_OS_1_3`, etc).  A simple fix is to just revert the change in the 2.11 index creator so it works for 2.11 snapshots and disable 1.3 test support in the tests given we don't currently support 1.3 in reality.  We can then circle back to support 1.3 properly later.  Alternatively we could add the formal support now but that seems like overkill.  Happy with either approach, but breaking our abstraction to informally support tests in this way doesn't seem like the the right balance for me.

peternied: I'll disable the test case, using `RFS Dynamically select transformer based on target cluster type` for follow up [1].

-[1] https://opensearch.atlassian.net/browse/MIGRATIONS-1827

gregschohn (approved):

RFS/src/main/java/com/rfs/version_es_6_8/IndexMetadataData_ES_6_8.java
```diff
@@ -30,7 +32,7 @@ public String getId() {
     }
 
     @Override
-    public ObjectNode getMappings() {
+    public JsonNode getMappings() {
```
chelma, line 35: Just curious - do you think it makes sense to change all our ObjectNodes to JsonNodes at some point?  JsonNode is more flexible, as exhibited here, but the type casting it will require could get annoying.  I'm really not sure what the right answer is.  Maybe it's just exactly what you did - use ObjectNode wherever possible and fall back to JsonNode when we must.

RFS/src/main/java/com/rfs/models/IndexMetadata.java
```diff
@@ -11,8 +12,24 @@
...
+public interface IndexMetadata extends Index {
+
+    /*
+    * Defines the behavior expected of an object that will surface the metadata of an index stored in a snapshot
+    * See: https://github.com/elastic/elasticsearch/blob/v7.10.2/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetadata.java#L1475
+    * See: https://github.com/elastic/elasticsearch/blob/v6.8.23/server/src/main/java/org/elasticsearch/cluster/metadata/IndexMetaData.java#L1284
+    */
+    public JsonNode getAliases();
+    public String getId();
+    public JsonNode getMappings();
```
chelma, line 28: Not a blocker, just trying to understand - it looks like we only updated the 6.8 implementation of this method to return a `JsonNode`, while the 7.10 version still returns an ObjectNode.  Is this intentional or an oversight?

chelma (approved):
Thanks for all the work on this one!

RFS/src/main/java/com/rfs/version_es_6_8/IndexMetadataData_ES_6_8.java
peternied: I think JsonNode is likely a better candidate - not sure if/when it will make sense to 'cut everything over', good thing to watch for oppertunities

RFS/src/main/java/com/rfs/models/IndexMetadata.java
peternied: Intentional / limiting the number of changes.

PR #785, opened 2024-07-01
Updated RFS Doc Migration main() to use local snapshot

chelma:
### Description
* Updated the `RfsMigrateDocuments.java` entrypoint to accept a local snapshot
* Added a new, local test that runs the full `RfsMigrateDocuments.java` against local target/source in Docker containers and a local snapshot.  The test asserts that the process exits with the expected status code.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1807
* https://opensearch.atlassian.net/browse/MIGRATIONS-1809

### Testing
Ran the new tests using `gradle DocumentsFromSnapshotMigration:slowTest`

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/build.gradle
+++ b/DocumentsFromSnapshotMigration/build.gradle
@@ -148,10 +148,6 @@ test {
 }
 
 task slowTest(type: Test) {
-    doFirst {
-        println "Running slow tests..."
-    }
-
     useJUnitPlatform()
     dependsOn(':TrafficCapture:dockerSolution:buildDockerImage_elasticsearchTestConsole')
     jacoco {
@@ -163,7 +159,7 @@ task slowTest(type: Test) {
         showExceptions true
         showCauses true
         showStackTraces true
-        // showStandardStreams = true
+        // showStandardStreams true
     }
     reports {
         html.required = true
@@ -174,8 +170,6 @@ task slowTest(type: Test) {
 jacocoTestReport {
     dependsOn slowTest
     reports {
-        xml.required = true
-        xml.destination file("${buildDir}/reports/jacoco/test/jacocoTestReport.xml")
         html.required = true
         html.destination file("${buildDir}/reports/jacoco/test/html")
     }
--- a/DocumentsFromSnapshotMigration/build.gradle
+++ b/DocumentsFromSnapshotMigration/build.gradle
@@ -148,10 +148,26 @@ test {
 }
 
 task slowTest(type: Test) {
+    doFirst {
+        println "Running slow tests..."
+    }
+
     useJUnitPlatform()
     dependsOn(':TrafficCapture:dockerSolution:buildDockerImage_elasticsearchTestConsole')
     jacoco {
         enabled = true
+    }
+    testLogging {
+        events "passed", "skipped", "failed"
+        exceptionFormat "full"
+        showExceptions true
+        showCauses true
+        showStackTraces true
+        // showStandardStreams = true
+    }
+    reports {
+        html.required = true
+        html.destination file("${buildDir}/reports/tests/slowTest")
     }
 }
 
--- a/DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
+++ b/DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
@@ -2,6 +2,7 @@ package com.rfs;
 
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
+import com.beust.jcommander.ParameterException;
 
 import java.io.IOException;
 import java.net.URI;
@@ -22,6 +23,7 @@ import com.rfs.worker.ShardWorkPreparer;
 
 import com.rfs.common.DefaultSourceRepoAccessor;
 import com.rfs.common.DocumentReindexer;
+import com.rfs.common.FileSystemRepo;
 import com.rfs.common.IndexMetadata;
 import com.rfs.common.LuceneDocumentsReader;
 import com.rfs.common.OpenSearchClient;
@@ -49,20 +51,33 @@ public class RfsMigrateDocuments {
                 description = "The name of the snapshot to migrate")
         public String snapshotName;
 
+        @Parameter(names = {"--snapshot-local-dir"},
+                required = false,
+                description = ("The absolute path to the directory on local disk where the snapshot exists.  Use this parameter"
+                    + " if have a copy of the snapshot disk.  Mutually exclusive with --s3-local-dir, --s3-repo-uri, and --s3-region."
+                ))
+        public String snapshotLocalDirPath = null;
+
         @Parameter(names = {"--s3-local-dir"},
-                required = true,
-                description = "The absolute path to the directory on local disk to download S3 files to")
-        public String s3LocalDirPath;
+                required = false,
+                description = ("The absolute path to the directory on local disk to download S3 files to.  If you supply this, you must"
+                    + " also supply --s3-repo-uri and --s3-region.  Mutually exclusive with --snapshot-local-dir."
+                ))
+        public String s3LocalDirPath = null;
 
         @Parameter(names = {"--s3-repo-uri"},
-                required = true,
-                description = "The S3 URI of the snapshot repo, like: s3://my-bucket/dir1/dir2")
-        public String s3RepoUri;
+                required = false,
+                description = ("The S3 URI of the snapshot repo, like: s3://my-bucket/dir1/dir2.  If you supply this, you must"
+                    + " also supply --s3-local-dir and --s3-region.  Mutually exclusive with --snapshot-local-dir."
+                ))
+        public String s3RepoUri = null;
 
         @Parameter(names = {"--s3-region"},
-                required = true,
-                description = "The AWS Region the S3 bucket is in, like: us-east-2")
-        public String s3Region;
+                required = false,
+                description = ("The AWS Region the S3 bucket is in, like: us-east-2.  If you supply this, you must"
+                    + " also supply --s3-local-dir and --s3-repo-uri.  Mutually exclusive with --snapshot-local-dir."
+                ))
+        public String s3Region = null;
 
         @Parameter(names = {"--lucene-dir"},
                 required = true,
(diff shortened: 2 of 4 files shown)


gregschohn:
In the PR description, say "main", or even "program" rather than entrypoint.  I was wondering what you were doing with docker when I read the description.

DocumentsFromSnapshotMigration/build.gradle
```diff
@@ -148,11 +148,27 @@ test {
 }
 
 task slowTest(type: Test) {
+    doFirst {
+        println "Running slow tests..."
+    }
+
```
gregschohn, line 154: Remove

DocumentsFromSnapshotMigration/build.gradle
```diff
@@ -148,11 +148,27 @@ test {
...
+        exceptionFormat "full"
+        showExceptions true
+        showCauses true
+        showStackTraces true
+        // showStandardStreams = true
+    }
+    reports {
+        html.required = true
+        html.destination file("${buildDir}/reports/tests/slowTest")
+    }
```
gregschohn, line 171: For checkin, make sure that this is consistent with our other projects.  I'm not sure currently what it should be, but it's evolved over time.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -120,8 +155,16 @@ public static void main(String[] args) throws Exception {
...
-                        new S3Uri(arguments.s3RepoUri), arguments.s3Region);
+                SourceRepo sourceRepo;
+                if (arguments.snapshotLocalDirPath == null) {
+                    sourceRepo = S3Repo.create(
+                        Paths.get(arguments.s3LocalDirPath),
+                        new S3Uri(arguments.s3RepoUri),
+                        arguments.s3Region
+                    );
+                } else {
+                    sourceRepo = new FileSystemRepo(Paths.get(arguments.snapshotLocalDirPath));
```
gregschohn, line 166: lift the args check up and add a default else to throw to make the code more defensible for the future.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +314,90 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+    public static Stream<Arguments> makeProcessExitArgs() {
+        var sourceImageArgs = SOURCE_IMAGES.stream().map(name -> makeParamsForBase(name)).collect(Collectors.toList());
+        var targetImageNames = TARGET_IMAGES.stream().map(SearchClusterContainer.Version::getImageName).collect(Collectors.toList());
+
+        return sourceImageArgs.stream()
+            .flatMap(a->
+            targetImageNames.stream().map(b->Arguments.of(a[0], a[1], a[2], b)));
+    }
+
+    @ParameterizedTest
```
gregschohn, line 326: Why do you think this test needs to be parameterized on source/target endpoints?  I would rather see a parameter on something like shouldThrow vs shouldSucceed.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +314,90 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+    @MethodSource("makeProcessExitArgs")
+    public void testProcessExitsAsExpected(SearchClusterContainer.Version baseSourceImageVersion,
+            String generatorImage, String[] generatorArgs,
+            String targetImageName) throws Exception {
+
+        try (var esSourceContainer = new PreloadedSearchClusterContainer(baseSourceImageVersion,
+                    SOURCE_SERVER_ALIAS, generatorImage, generatorArgs);
+                OpensearchContainer<?> osTargetContainer =
```
gregschohn, line 334: Setting up a toxiproxy in between your future process and your target will let you inject some severe timeouts that should cause timeouts to propagate.  Those should eventually cause a lease expiration, which would be really useful to check.

Take a look at KafkaConfigurationCaptureProxyTest for an example of where we use the toxiproxy test container.  You might also be able to route a test to a dummy endpoint that returns 404 (see SimpleNettyHttpServer).  That should cause retries to go on & on.  For testing purposes, you might need to parameterize (at the command line) the initial lease time for either of these.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +314,90 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+                processBuilder.command().addAll(Arrays.asList(args));
+
+                Process process = processBuilder.start();
+                log.atInfo().setMessage("Process started with ID: " + Long.toString(process.toHandle().pid())).log();
+                
+                // Kill the process and fail if we have to wait too long
+                boolean finished = process.waitFor(2, TimeUnit.MINUTES);
+                if (!finished) {
+                    log.atError().setMessage("Process timed out, attempting to kill it...").log();
+                    process.destroy(); // Try to be nice about things first...
```
gregschohn, line 380: Is there some kind of dump that should be done before terminating the process?  lsof, thread inspection, CPU stats, etc?  A 5 minute investment is all that we need now.  We hopefully can reproduce this easily locally if the test fails and a breakpoint or long-sleep would work fine in the local case.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +314,90 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+                    if (!process.waitFor(10, TimeUnit.SECONDS)) {
+                        log.atError().setMessage("Process still running, attempting to force kill it...").log();
+                        process.destroyForcibly(); // ..then avada kedavra
+                    }
+                    Assertions.fail("The process did not finish within the timeout period.");
```
gregschohn, line 385: print out the time period

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +314,90 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+                }
+
+                int exitCode = process.exitValue();
+
+                // Check if the exit code is as expected
+                int expectedExitCode = 0;
```
gregschohn, line 391: this would need to be parameterized for the error case.

DocumentsFromSnapshotMigration/build.gradle
chelma: Yeah, probably unnecessary now with testLogging section below.

DocumentsFromSnapshotMigration/build.gradle
chelma: This was pulled from the RFS sub-project; will check what the other sub-projects use though.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: I don't; I was just following the existing paradigm until I knew how to insert exceptions, which is what I really wanted to parameterize on.  I had to have something in the draft PR though.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: Will investigate; thanks.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: I'll take another look, but what I really wanted was the stdout/stderr streams.  However, attaching something to read from the process prevents it from timing out the wait correctly, and I wasn't able to find a way to get it to work the way I wanted.  General process stats may be easier.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: Yep!

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: Will do.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: I did something a bit simpler this time around, but I'd be interested in following up later for more sophisticated testing of the work coordination.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
chelma: I found a way to get stderr/stdout; it just comes down to when and how you set up the stream to read it.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +316,112 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+
+                var targetClient = new OpenSearchClient(targetAddress, null);
+                var sourceRepo = new FileSystemRepo(tempDirSnapshot);
+                migrateMetadata(sourceRepo, targetClient, SNAPSHOT_NAME, INDEX_ALLOWLIST);
+
+                // Stop the target container if we don't want it to be available.  We've already cached the address it was
+                // using, so we can have reasonable confidence that nothing else will be using it and bork our test.
+                if (!targetAvailable) {
+                    osTargetContainer.stop();
+                }
```
gregschohn, line 371: This is wasteful.  You could bind to a random port (see PortFinder) or setup a local http server if you wanted to keep the connection open.  

Can you document what behavior you're looking the client (migration process) to exhibit?  e.g. "conn_refused means that this part fails, then that part fails, then the process exits"

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +316,112 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+                Process process = processBuilder.start();
+                log.atInfo().setMessage("Process started with ID: " + Long.toString(process.toHandle().pid())).log();
+                
+                // Kill the process and fail if we have to wait too long
+                int timeoutSeconds = 90;
+                boolean finished = process.waitFor(timeoutSeconds, TimeUnit.SECONDS);
+                if (!finished) {
+                    // Print the process output
+                    StringBuilder output = new StringBuilder();
+                    try (BufferedReader reader = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
```
gregschohn, line 394: Each of the 'new' objects should be in its own line to do proper cleanup for the try-with-resources.
```
try (
  var processStream = process.getInputStream();
  var isr = new InputStreamReader(processStream);
  var reader = new BufferedReader(isr)
) ...
```

DocumentsFromSnapshotMigration/src/test/java/com/rfs/FullTest.java
```diff
@@ -308,6 +316,112 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(SourceRep
...
+                        String line;
+                        while ((line = reader.readLine()) != null) {
+                            output.append(line).append(System.lineSeparator());
+                        }
+                    }
```
gregschohn, line 399: something like `var output = reader.lines().join(Collectors.joining("\n");` could replace doing it yourself.
Since you're in a test and not worrying about persisting this output, I'd make the test run the same on all platforms, with the same line ending for all

gregschohn (approved):
Minor comments.  Please file a new task to add a test to make sure that process death for expiration works as expected.

chelma:
> Minor comments. Please file a new task to add a test to make sure that process death for expiration works as expected.

Done - https://opensearch.atlassian.net/browse/MIGRATIONS-1823

PR #791, opened 2024-07-02
Allow specifying the allowed hosts for Console API

lewijacn:
### Description
We have made assumptions about what allowed hosts are able to use Django API, and this provides a way for a user to specify specific hosts to allow

### Issues Resolved
N/A

### Testing
Local testing

### Check List


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/console_api/console_api/settings.py
```diff
@@ -28,8 +28,23 @@
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = True
 
-DEPLOYED_STAGE = os.environ.get('MIGRATION_STAGE')
-ALLOWED_HOSTS = ['migration-console', f'migration-console.migration.{DEPLOYED_STAGE}.local', 'localhost']
+
+def get_allowed_hosts():
```
mikaylathompson, line 32: minor, but a type annotation would help make it a bit more clear what's going on here.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/console_api/console_api/settings.py
```diff
@@ -28,8 +28,23 @@
...
+        return ['localhost']
+    # Remove extra spacing and/or quotes
+    hosts = hosts.strip()
+    hosts = hosts.replace('"', '')
+    hosts = hosts.replace('\'', '')
+
+    host_list = []
+    for host in hosts.split(','):
+        host_list.append(host.strip())
+    return host_list
```
mikaylathompson, line 44: This is minor, but we're doing some unnecessary stuff here--the initial strip doesn't matter because we're stripping the individual items (and that's the level we want to do it at)--and we can do this more directly/concisely. On the other hand, I know that I love python one-liners more than the average person, so feel free to disregard this if it makes it less intuitive. (I did make it a two-liner instead of a one-liner though)

```suggestion
    hosts = hosts.replace('"', '').replace('\'', '')
    return [host.strip() for host in hosts.split(',')]
```

mikaylathompson (approved):
I made nitpicky python comments. You can take or leave them.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/console_api/console_api/settings.py
lewijacn: Added

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/console_api/console_api/settings.py
lewijacn: Agree this is cleaner, have updated now

PR #793, opened 2024-07-03
[Console] Refactor the logic layer

mikaylathompson:
### Description
There's a moderately high amount of stuff going on here.
- Logic -> Middleware
- all the factory methods (get_replayer, get_metrics_source, get_backfill, etc.) are now grouped together and out of the individual logic classes
- there's a decorator to handle `as_json` in commands. Right now, this is only being used by middleware functions for `describe`, but to support an api, we'll need to make this universal down the line. When we get to that point, we should be returning `CommandResult[Dict]` for all but the most trivial of responses so that it can be meaningfully converted to a json (and then have a method to convert it to string for the non-json path).
- Create an error handling decorator that handles our usual cases for error checking and how to output ExitCode tuples. 

### Issues Resolved

### Testing

### Check List


.whitesource
```diff
@@ -1,23 +1,17 @@
 {
```
AndreKurait, line 1: This diff should be removed with a rebase

AndreKurait (approved):
Thanks for this refactor. Seems like a step in the right direction. I see some improvements to be made for the decorators, particularly with better ability to check typing statically, but will leave that to future changes as it's a 2 way door

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/middleware/backfill.py
```diff
@@ -0,0 +1,56 @@
...
+def status(backfill: Backfill, deep_check: bool, *args, **kwargs) -> CommandResult[Tuple[BackfillStatus, str]]:
+    logger.info(f"Getting backfill status with {deep_check=}")
+    return backfill.get_status(deep_check, *args, **kwargs)
+
+
+@handle_errors("backfill",
+               on_success=lambda status: (ExitCode.SUCCESS, status))
+def scale(backfill: Backfill, units: int, *args, **kwargs) -> CommandResult[str]:
+    logger.info(f"Scaling backfill to {units} units")
+    return backfill.scale(units, *args, **kwargs)
```
AndreKurait, line 56: Can you help me understand what *args and **kwargs are intended for if we're passing all the variables directly like units here

mikaylathompson: That's a good question. The way I've been thinking about is that it's for optional params that might apply to one type of service but not another, and this ensures they get passed all the way through. So far, we've almost entirely avoided actually having that kind of parameter, so putting these in place may have been premature.

mikaylathompson: I think we definitely want to explicitly specify the params that are core to the functionality (like `units`), but I don't have a clear sense of whether we'll actually need the rest. So far, we've done well at pushing everything we can into the services.yaml.

PR #803, opened 2024-07-09
Small collection of SonarQube violation fixes

chelma:
### Description
A quick set of fixes to SonarQube violations

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1826

### Testing
Ran `gradle RFS:test` and `gradle DocumentsFromSnapshotMigration:slowTest`

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -22,7 +22,7 @@ public class DocumentReindexer {
         return documentStream
             .map(this::convertDocumentToBulkSection)  // Convert each Document to part of a bulk operation
             .buffer(MAX_BATCH_SIZE) // Collect until you hit the batch size
-            .doOnNext(bulk -> logger.info(bulk.size() + " documents in current bulk request"))
+            .doOnNext(bulk -> logger.info("{} documents in current bulk request", bulk.size()))
             .map(this::convertToBulkRequestBody)  // Assemble the bulk request body from the parts
             .flatMap(bulkJson -> client.sendBulkRequest(indexName, bulkJson) // Send the request
                 .doOnSuccess(unused -> logger.debug("Batch succeeded"))
@@ -50,9 +50,9 @@ public class DocumentReindexer {
         return builder.toString();
     }
 
-    public void refreshAllDocuments(ConnectionDetails targetConnection) throws Exception {
+    public void refreshAllDocuments(ConnectionDetails targetConnection) {
         // Send the request
-        OpenSearchClient client = new OpenSearchClient(targetConnection);
-        client.refresh();
+        OpenSearchClient refreshClient = new OpenSearchClient(targetConnection);
+        refreshClient.refresh();
     }
 }
--- a/RFS/src/main/java/com/rfs/common/GlobalMetadata.java
+++ b/RFS/src/main/java/com/rfs/common/GlobalMetadata.java
@@ -14,6 +14,8 @@ import com.fasterxml.jackson.dataformat.smile.SmileFactory;
 
 
 public class GlobalMetadata {
+    private GlobalMetadata() {}
+
     /**
     * Defines the behavior required to read a snapshot's global metadata as JSON and convert it into a Data object
     */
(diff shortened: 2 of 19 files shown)


RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -50,9 +50,9 @@ private String convertToBulkRequestBody(List<String> bulkSections) {
         return builder.toString();
     }
 
-    public void refreshAllDocuments(ConnectionDetails targetConnection) throws Exception {
+    public void refreshAllDocuments(ConnectionDetails targetConnection) {
         // Send the request
-        OpenSearchClient client = new OpenSearchClient(targetConnection);
-        client.refresh();
+        OpenSearchClient refreshClient = new OpenSearchClient(targetConnection);
+        refreshClient.refresh();
```
peternied, line 56: Was there a rule about the naming? Client seems like a perfectly cromulent name

RFS/src/main/java/com/rfs/common/GlobalMetadata.java
```diff
@@ -14,6 +14,8 @@
 
 
 public class GlobalMetadata {
+    private GlobalMetadata() {}
```
peternied, line 17: Note; in my outstanding change, these classes are promoted to 'functional', I think the rules are triggering here because there are no instance references so its seen as a utility class.

Recommendations like this are more about a code smell why does this class exist if it doesn't have any properties/fields - for example this could have been a package path

RFS/src/main/java/com/rfs/common/S3Repo.java
```diff
@@ -187,7 +187,7 @@ public void prepBlobFiles(ShardMetadata.Data shardMetadata) {
 
         String blobFilesS3Prefix = s3RepoUri.key + "indices/" + shardMetadata.getIndexId() + "/" + shardMetadata.getShardId() + "/";
 
-        logger.info("Downloading blob files from S3: s3://" + s3RepoUri.bucketName + "/" + blobFilesS3Prefix + " to " + shardDirPath);
+        logger.info("Downloading blob files from S3: s3://%s/%s to %s", s3RepoUri.bucketName, blobFilesS3Prefix, shardDirPath);
```
peternied, line 190: I think we broadly are using `logger.atLevel(Info).setMessage(...).addArgument(() -> ...)....log();` @AndreKurait @gregschohn Did you have more context, is that a convention we should be strongly following?

RFS/src/main/java/com/rfs/common/SnapshotShardUnpacker.java
```diff
@@ -45,25 +45,25 @@ public Path unpack() {
...
-                    if (fileMetadata.getName().startsWith("v__")) {
-                        final BytesRef hash = fileMetadata.getMetaHash();
-                        indexOutput.writeBytes(hash.bytes, hash.offset, hash.length);
-                    } else {
-                        try (InputStream stream = new PartSliceStream(repoAccessor, fileMetadata, shardMetadata.getIndexId(), shardMetadata.getShardId())) {
-                            final byte[] buffer = new byte[Math.toIntExact(Math.min(bufferSize, fileMetadata.getLength()))];
-                            int length;
-                            while ((length = stream.read(buffer)) > 0) {
-                                indexOutput.writeBytes(buffer, 0, length);
+            try (FSDirectory primaryDirectory = FSDirectory.open(luceneIndexDir, lockFactory)) {
```
peternied, line 48: 🤞 that a test tells us if this creates an issue or not

peternied (approved):

RFS/src/main/java/com/rfs/common/S3Repo.java
mikaylathompson: At some point Greg did a pass that changed all of the log messages to the fluent style `atLevel().setMessage()` and it meant we could run 15x faster or something absurd.

mikaylathompson: ah, well, not quite 15x faster. but it did free up 50% of our CPU utilization.

https://github.com/opensearch-project/opensearch-migrations/pull/280

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
chelma: It was overlapping with the class member of the same name.  It feels like this method probably shouldn't exist as it does, but I decided to leave it so that I can focus on the SonarQube findings more specifically.

RFS/src/main/java/com/rfs/common/S3Repo.java
chelma: I'm happy to have that be our convention going forward.  Will start swapping to that when I see the old style.

PR #818, opened 2024-07-12
Separate spotless checks from GitHub Action test workflows

peternied:
### Description
Separate spotless checks from GitHub Action test workflows

### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,6 +11,20 @@ env:
   node-version: '18.x'
 
 jobs:
+  style-check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          java-version: ${{ env.java-version }}
+          distribution: "corretto"
+      - uses: gradle/actions/setup-gradle@v3
+        with:
+          gradle-version: ${{ env.gradle-version }}
+      - name: Run Gradle Build
+        run: ./gradlew spotlessCheck
+
   python-lint:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


gregschohn:
I have a question for the migrations developers - should spotless violations block any/all merges?  Vote with emoticons.
The last commit in the PR requires spotless for merges too.  I believe that's how builds currently work too since spotless is tied up into the build command previously, but it's negotiable.

.github/workflows/CI.yml
```diff
@@ -181,6 +188,7 @@ jobs:
 
   all-ci-checks-pass:
     needs:
+      - style-check
```
gregschohn, line 191: I would ask that you continue to leave this out for the time being until we find if it's critical or not.

gregschohn (changes requested):
As per my other comments, the general PR looks good, but I would prefer us to NOT block merges on the spotless task succeeding.

.github/workflows/CI.yml
peternied: You have a good notion, we should acknowledge that it can be off putting making a pull request and getting a failing check on style.

However, I think we should keep it for these reasons:
1) We only have a single class of checks right now, _run on everything_.  If we make different checks required at different times it can make it hard to have confidence in the state of the repository.  As much as I don't like spending time running CI for the end to end tests when I update a readme.
2) If someone comes around later and updates the style, it hides the change from the original author making it hard to know what changes are actually associated with which authors.  I put a lot of stock in `git blame` and I'd hate to loss out on that tool because some authors follow spotless and some don't.

gregschohn: do we like the standard?  Will it make the code easier to read or harder?  I wrote a lot of the reformatted code, so I'm biased, but I'd love to get some neutral eyes on it.
How many people have done a significant sampling of the codebase before/after?  There were two approvals on the PR to make the formatting change, but one has voted to NOT put this into effect and has backtracked.  

The mandatory-ness of this check also broke our jenkins builds for 5 days and maybe a dozen commits.  Will we have separate jenkins workflows and checks there too?  If we want a mandatory check here, we should have an analogous one in jenkins that breaks the builds in the same way as github actions.

chelma: > As per my other comments, the general PR looks good, but I would prefer us to NOT block merges on the spotless task succeeding.

I'm actually in favor of going further - removing Spotless as a blocker for any of our developer actions.  I'm finding our current rules making it harder to read and write code, and I'm struggling to figure out which specific rules are impacting my code in ways I don't think make sense.  If/when we have a less burdensome rule set I'm open to re-integrating it as a blocker.

In the meantime, I'm fine with letting folks optionally update their code to conform to the Spotless standard.  If there's wacky formatting in a PR, that's something that we can address iteratively in that PR.  The existing wacky formatting in the code base can also be address organically as we encounter it and have occasion to change it.

chelma (changes requested):
Per my comments on @gregschohn's thread, I strongly prefer we don't block any of our developer processes on Spotless in its current form.

.github/workflows/CI.yml
peternied: > check also broke our jenkins builds for 5 days and maybe a dozen commits

This is a separate issue and AFAIK they have been resolved - isn't this expected when we don't have good visibility into Jenkins?  I know we have more refactors coming in, and I expect they will also destabilize our different build platforms.  This seems like a tax we are willing to pay.

> do we like the standard

It is OpenSearch Core's standard and I like the standard.  And I can see how we have code that pushes against the standard - specifically managing json data.  There are places where I believe there are code smells that caused the formatting to get strange - such passing large number of arguments including lambdas.

While I can understand having judgements about how the code looks compared to hand styled that is a tradeoff for that no matter who authors code in this codebase it will be consistent.

---

This pull request does not address any either of these issues.

@gregschohn To keep moving forward, what do you think about making a change for what state you would like - removing many of the formatters rules, or removing it entirely?

peternied (commented):
- #827 has been merged, @gregschohn @chelma do you mind re-reviewing this change?

gregschohn (approved):

PR #821, opened 2024-07-16
[RFS] Added 7.10 and 7.17 sources to RFS EndToEndTest

chelma:
### Description
* Updated the RFS EndToEnd test to run 7.10 and 7.17 sources
* Fixed a bug in the `OpenSearchClient` where it would consider any response in the final `put` of  `createObjectIdempotent` successful
* Fixed a bug in `Transformer_ES_7_10_OS_2_11` where we were returning the wrong object from the `transformIndexMetadata` method
* Did NOT update the end-to-end tests to perform document migration yet in order to reduce the scope of this particular PR.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1857

### Testing
* Ran the various local tests

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/OpenSearchClient.java
+++ b/RFS/src/main/java/com/rfs/common/OpenSearchClient.java
@@ -12,6 +12,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
+import com.rfs.common.RestClient.Response;
 import com.rfs.tracing.IRfsContexts;
 import lombok.NonNull;
 import reactor.core.publisher.Mono;
@@ -98,7 +99,7 @@ public class OpenSearchClient {
         ObjectNode settings,
         IRfsContexts.ICheckedIdempotentPutRequestContext context
     ) {
-        RestClient.Response response = client.getAsync(objectPath, context.createCheckRequestContext())
+        RestClient.Response getResponse = client.getAsync(objectPath, context.createCheckRequestContext())
             .flatMap(resp -> {
                 if (resp.code == HttpURLConnection.HTTP_NOT_FOUND || resp.code == HttpURLConnection.HTTP_OK) {
                     return Mono.just(resp);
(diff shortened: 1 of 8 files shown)


RFS/src/main/java/com/rfs/transformers/Transformer_ES_7_10_OS_2_11.java
```diff
@@ -97,6 +97,6 @@ public IndexMetadata transformIndexMetadata(IndexMetadata indexData) {
         TransformFunctions.fixReplicasForDimensionality(newRoot, awarenessAttributeDimensionality);
 
         logger.debug("Transformed Object: " + newRoot.toString());
-        return indexData;
+        return copy;
```
peternied, line 100: 👍

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -118,8 +118,26 @@ private Optional<ObjectNode> createObjectIdempotent(
             .retryWhen(Retry.backoff(3, Duration.ofSeconds(1)).maxBackoff(Duration.ofSeconds(10)))
             .block();
 
-        if (response.code == HttpURLConnection.HTTP_NOT_FOUND) {
-            client.put(objectPath, settings.toString(), context.createCheckRequestContext());
+        if (getResponse.code == HttpURLConnection.HTTP_NOT_FOUND) {
```
peternied, line 121: This line made me do a double take, what do you think about making a bool local called 'objectDoesNotExist` that gets the value of `getResponse.code == HttpURLConnection.HTTP_NOT_FOUND` so its clear what the body of this if block is doing?

peternied: Also, inverting the condition to return early if the object exists would help make this clearer yet.

RFS/src/test/java/com/rfs/framework/ClusterOperations.java
```diff
@@ -145,4 +145,71 @@ public void createES6LegacyTemplate(final String templateName, final String patt
...
+        final var indexTemplateJson = "{"
+            + "\"index_patterns\": [\""
+            + indexPattern
+            + "\"],"
+            + "\"composed_of\": [\""
+            + componentTemplateName
+            + "\"],"
+            + "\"priority\": 1,"
+            + "\"version\": 1"
+            + "}";
```
peternied, line 200: Nit: I think this might be cleaner with string.format, but this is fine. 
image

peternied (approved):
Nice, thanks for adding these

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
chelma: Will queue the change for the next PR.

RFS/src/test/java/com/rfs/framework/ClusterOperations.java
chelma: Definitely not my first choice, but it makes Spotless happy...

gregschohn: > Nit: I think this might be cleaner with string.format, but this is fine. image

That would also be cleaner to grok w/ a map and jackson to output it.  Can the values ever contain any characters that could cause json an issue?

PR #829, opened 2024-07-17
Split up Auth transformations from replayer

AndreKurait:
### Description
Split up Auth transformations from replayer by creating a separate project awsCoreUtilities for the sigv4 transformation  and splitting up the other http utilities 

Improves unit testing for the sigv4 library

* Category: Refactoring
* Why these changes are required? Enable sharing of this code between replayer and RFS
* What is the old behavior before changes and new behavior after changes? Code moved to sub projects

### Issues Resolved
Part of [MIGRATIONS-1842](https://opensearch.atlassian.net/browse/MIGRATIONS-1842)

### Testing
Unit testing

### Check List


awsCoreUtilities/build.gradle
peternied, line 1: For the new package, lets drop `core`, and rename to `awsShared` or `awsUtilities`.

coreUtilities/src/main/java/org/opensearch/migrations/transform/ListKeyAdaptingCaseInsensitiveHeadersMap.java
```diff
@@ -18,7 +18,7 @@
  */
 @EqualsAndHashCode(callSuper = false)
 public class ListKeyAdaptingCaseInsensitiveHeadersMap extends AbstractMap<String, Object> {
-    protected final StrictCaseInsensitiveHttpHeadersMap strictHeadersMap;
+    public final StrictCaseInsensitiveHttpHeadersMap strictHeadersMap;
```
peternied, line 21: Why did you have to open up access?

RFS/build.gradle
```diff
@@ -47,6 +47,8 @@ dependencies {
     implementation group: 'software.amazon.awssdk', name: 's3'
     implementation group: 'software.amazon.awssdk', name: 's3-transfer-manager'
     implementation group: 'software.amazon.awssdk.crt', name: 'aws-crt'
+    implementation group: 'software.amazon.awssdk', name: 'auth'
+    implementation group: 'software.amazon.awssdk', name: 'sdk-core'
```
peternied, line 51: Shouldn't this be picked up by the new library?

gregschohn: Yes!  Please don't require downstream consumers to know what other jars they'll need to depend upon.  You can also expose those details as an api dependency if there's some AWS context, etc that needs to interface with the rest of the exposed implementation.

AndreKurait: Removed

awsCoreUtilities/build.gradle
AndreKurait: renamed to `awsUtilities`

TrafficCapture/trafficReplayer/README.md
```diff
@@ -39,9 +39,9 @@ over the constructed JSON document.
 Since payloads can be arbitrarily complex (compression, chunking), and may not be subject to any changes via the
 transformation rules, the HttpJsonTransformer creates the channel pipeline *only* to parse the HTTP headers. The
 transformation is run on this partial, in-construction
-[HttpJsonMessageWithFaultablePayload](./src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonMessageWithFaultingPayload.java)
+[HttpJsonMessageWithFaultablePayload](../../coreUtilities/src/main/java/org/opensearch/migrations/transform/HttpJsonMessageWithFaultingPayload.java)
```
gregschohn, line 42: This file is super specific to the replayer.  Even where other components will need to do transformations, they probably won't need to optimize away reading payload streams.  We do that in the replayer because it might mean that we can leave bytebuffs go right through the netty pipeline directly and never need to aggregate or transform them.  That adds up for the 10MB+ requests that aren't uncommon for opensearch.

For use cases outside of the replayer, I'd suspect that somebody already has the components of an IHttpMessage and just writes a quick bridge.

Other files that I wouldn't move out.
* JsonKeysForHttpMessage
* HttpJsonMessageWithFaultingPayload.java
* PayloadNotLoadedException.java
* PayloadAccessFaultingMap
* RemovingAuthTransformerFactory
* ListKeyAdapting...Headers
* StrictCaseInsensitive...Headers

I don't think that the remaining transforms classes belong in coreUtilities.  I also don't think that you'll want to make the awsUtilities package take a dependency on coreUtilities.  Instead of relying upon an IHttpMessage, you could inline IHttpMessage to be verb, uri, protocol, & headers.  Then you don't need to worry about having too many adapter shims laying around and the need to couple libraries together.

TrafficCapture/trafficReplayer/README.md
```diff
@@ -39,9 +39,9 @@ over the constructed JSON document.
...
 message. When the transformation (or any other code), attempts to access the payload contents, it will throw a
-[PayloadNotLoadedException](./src/main/java/org/opensearch/migrations/replay/datahandlers/PayloadNotLoadedException.java)
+[PayloadNotLoadedException](../../coreUtilities/src/main/java/org/opensearch/migrations/transform/PayloadNotLoadedException.java)
```
gregschohn, line 44: Same - don't move this one

awsUtilities/src/test/java/org/opensearch/migrations/aws/SigV4SigningTransformationTest.java
```diff
@@ -0,0 +1,131 @@
...
+            () -> Clock.fixed(Instant.EPOCH, ZoneOffset.UTC)
+        );
+
+        // Create a mock HTTP message
+        var httpMessage = new HttpJsonMessageWithFaultingPayload();
+        httpMessage.setMethod(method);
+        httpMessage.setPath(path);
+
+        StrictCaseInsensitiveHttpHeadersMap strictHeaders = new StrictCaseInsensitiveHttpHeadersMap();
+        ListKeyAdaptingCaseInsensitiveHeadersMap headers = new ListKeyAdaptingCaseInsensitiveHeadersMap(strictHeaders);
```
gregschohn, line 72: This class is overkill.  Assume that not all users will be using ListKeyAdaptingCaseInsensitiveHeadersMap.  Just send a regular map and make sure that it works.  It's useful to do sig4v signer tests in the replayer with these sorts of classes, but if I looked at this test by itself, I wouldn't use the sigv4 signer because it looks too complicated.

coreUtilities/src/main/java/org/opensearch/migrations/transform/ListKeyAdaptingCaseInsensitiveHeadersMap.java
gregschohn: and it should stay in the replayer.  It only exists as a more efficient implementation of headers maps for our own purposes.

gregschohn (changes requested):

AndreKurait:
Closed in favor of https://github.com/opensearch-project/opensearch-migrations/pull/834 based on feedback

PR #834, opened 2024-07-22
Split SigV4 signing into separate package

AndreKurait:
### Description
Split up Auth transformations from replayer by creating a separate project awsUtilities for the sigv4 transformation  and splitting up the other http utilities 

Improves unit testing for the sigv4 library

* Category: Refactoring
* Why these changes are required? Enable sharing of this code between replayer and RFS
* What is the old behavior before changes and new behavior after changes? Code moved to sub projects

### Issues Resolved
Part of [MIGRATIONS-1842](https://opensearch.atlassian.net/browse/MIGRATIONS-1842)

### Testing
Unit testing

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -424,19 +424,13 @@ private static IAuthTransformerFactory buildAuthTransformerFactory(Parameters pa
...
-                public IAuthTransformer getAuthTransformer(IHttpMessage httpMessage) {
-                    return new SigV4Signer(defaultCredentialsProvider, serviceName, region, "https", null);
-                }
-
-                @Override
-                public void close() {
-                    defaultCredentialsProvider.close();
-                }
-            };
+            return new SigV4AuthTransformerFactory(
```
gregschohn, line 427: Note to self - what does this factory look like, why IAuthTransformer?  Where does that interface live.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/PayloadAccessFaultingMap.java
```diff
@@ -7,8 +7,8 @@
 import java.util.Optional;
 import java.util.Set;
 
+import org.opensearch.migrations.IHttpMessage;
```
gregschohn, line 10: Will probably be cleaner if you drop this here and adapt it later on, the same way that you'll likely adapt the reactor or java objects elsewhere.  Right now, this creates a hard dependency on the awsUtils library, which isn't/shouldn't be necessary for the rest of this package.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonMessageWithFaultingPayload.java
```diff
@@ -48,11 +49,24 @@ public void setProtocol(String value) {
     }
 
     @Override
-    public Map<String, Object> headersMap() {
-        return Collections.unmodifiableMap(headers());
+    public Map<String, List<String>> headers() {
```
gregschohn, line 52: This function is a great example of why you want to move this to a separate adapter class.  It looks just like a plain old getter, but it does a significant amount of construction.  Imagine somebody calling this from the inside of a loop.  Given that you only need _this_ interface for the adapter, you should move it out so that this method isn't confused to be an intrinsically, first-class, 'native' function on the class.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonMessageWithFaultingPayload.java
```diff
@@ -48,11 +49,24 @@ public void setProtocol(String value) {
...
+            } else if (entry.getValue() != null) {
+                convertedHeaders.put(entry.getKey(), Collections.singletonList(entry.getValue().toString()));
+            }
+        }
+
+        return Collections.unmodifiableMap(convertedHeaders);
     }
 
-    public ListKeyAdaptingCaseInsensitiveHeadersMap headers() {
+    public ListKeyAdaptingCaseInsensitiveHeadersMap headersInternal() {
```
gregschohn, line 69: Following through from the last comment - I'm not sure why everybody in the replayer dealing with this object deals with headers through an 'internal' interface.  This should be the public one.  For adapting to other package boundaries, do that separate from the class.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/transform/IAuthTransformerFactory.java
```diff
@@ -2,6 +2,8 @@
 
 import java.io.IOException;
 
+import org.opensearch.migrations.IHttpMessage;
+
 public interface IAuthTransformerFactory extends AutoCloseable {
     IAuthTransformer getAuthTransformer(IHttpMessage httpMessage);
```
gregschohn, line 8: I wouldn't use IHttpMessage here. This package only needs to deal with HttpJsonMessageWithFaultingPayload objects (which is also what you've spec'ed the function out for).

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/transform/SigV4AuthTransformerFactory.java
```diff
@@ -0,0 +1,49 @@
...
+            Supplier<Clock> timestampSupplier) {
+        this.credentialsProvider = credentialsProvider;
+        this.service = service;
+        this.region = region;
+        this.protocol = protocol;
+        this.timestampSupplier = timestampSupplier;
+    }
+
+    @Override
+    public IAuthTransformer getAuthTransformer(IHttpMessage httpMessage) {
```
gregschohn, line 34: This can take HttpJsonMessageWithFaultingPayload and the body should adapt it to an IHttpMessage object.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/PayloadAccessFaultingMap.java
AndreKurait: Created an adapter

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonMessageWithFaultingPayload.java
AndreKurait: Addressed, thanks for the comment

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/transform/SigV4AuthTransformerFactory.java
AndreKurait: addressed

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/HttpJsonMessageWithFaultingPayload.java
AndreKurait: addressed

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/helpers/IHttpMessageAdapter.java
```diff
@@ -0,0 +1,50 @@
+package org.opensearch.migrations.replay.datahandlers.http.helpers;
```
gregschohn, line 1: This class would be easier to immediately comprehend and understand its purpose if it were an inner class of SigV4AuthTransformer

AndreKurait: Moved

PR #835, opened 2024-07-22
Rfs Unified Client with SigV4

AndreKurait:
### Description
Rewrite RFS to use a unified http client with sigv4 support.
~Refactors the SigV4 client use the non-deprecated `AwsV4HttpSigner`~

Note: Changes in migration console/cdk to call RFS with these parameters has not been added.

* Category Refactoring, Maintenance
* Why these changes are required? Reduce surface area for bugs 
* What is the old behavior before changes and new behavior after changes? No SigV4 Support, now SigV4 is supported with new parameters to RFS.

### Issues Resolved
[MIGRATIONS-1842](https://opensearch.atlassian.net/browse/MIGRATIONS-1842)

### Testing
Unit test suite

Tested some new flows with SigV4 which showed correct verification on the server in positive and negative cases
```
Positive

(.venv) bash-5.2# ./createSnapshot/bin/CreateSnapshot --snapshot-name sigv4clustersnapshot --s3-repo-uri s3://migration-artifacts-767398060394-aws-integ-us-east-2/folder --s3-region us-east-2 --source-host https://vpc-instance-store-backup-4a5wsdhlqoemni5sr2ptv5ynsy.us-east-2.es.amazonaws.com:443 --source-aws-region us-east-2 --source-aws-service-signing-name es --s3-role-arn arn:aws:iam::767398060394:role/OS-Snapshot
2024-08-02 14:14:30,357 WARN o.o.m.t.RootOtelContext [main] Collector endpoint=null, so serviceName parameter 'rfs' is being ignored since a no-op OpenTelemetry object is being created
2024-08-02 14:14:30,512 INFO c.r.CreateSnapshot [main] Running CreateSnapshot with --snapshot-name sigv4clustersnapshot --s3-repo-uri s3://migration-artifacts-767398060394-aws-integ-us-east-2/folder --s3-region us-east-2 --source-host https://vpc-instance-store-backup-4a5wsdhlqoemni5sr2ptv5ynsy.us-east-2.es.amazonaws.com:443 --source-aws-region us-east-2 --source-aws-service-signing-name es --s3-role-arn arn:aws:iam::767398060394:role/OS-Snapshot
2024-08-02 14:14:31,718 INFO c.r.w.SnapshotRunner [main] Attempting to initiate the snapshot...
2024-08-02 14:14:34,523 INFO c.r.c.SnapshotCreator [main] Snapshot repo registration successful
2024-08-02 14:14:34,699 INFO c.r.c.SnapshotCreator [main] Snapshot sigv4clustersnapshot creation initiated
2024-08-02 14:14:34,700 INFO c.r.w.SnapshotRunner [main] Snapshot in progress...
2024-08-02 14:14:35,200 INFO c.r.w.SnapshotRunner [main] Snapshot not finished yet; sleeping for 1000ms...
2024-08-02 14:14:36,231 INFO c.r.w.SnapshotRunner [main] Snapshot not finished yet; sleeping for 1000ms...
(.venv) bash-5.2# 


Negative (wrong service signing name)

(.venv) bash-5.2# ./createSnapshot/bin/CreateSnapshot --snapshot-name sigv4clustersnapshot --s3-repo-uri s3://migration-artifacts-767398060394-aws-integ-us-east-2/folder --s3-region us-east-2 --source-host https://vpc-instance-store-backup-4a5wsdhlqoemni5sr2ptv5ynsy.us-east-2.es.amazonaws.com:443 --source-aws-region us-east-2 --source-aws-service-signing-name opensearchservice --s3-role-arn arn:aws:iam::767398060394:role/OS-Snapshot
2024-08-02 14:15:02,217 WARN o.o.m.t.RootOtelContext [main] Collector endpoint=null, so serviceName parameter 'rfs' is being ignored since a no-op OpenTelemetry object is being created
2024-08-02 14:15:02,295 INFO c.r.CreateSnapshot [main] Running CreateSnapshot with --snapshot-name sigv4clustersnapshot --s3-repo-uri s3://migration-artifacts-767398060394-aws-integ-us-east-2/folder --s3-region us-east-2 --source-host https://vpc-instance-store-backup-4a5wsdhlqoemni5sr2ptv5ynsy.us-east-2.es.amazonaws.com:443 --source-aws-region us-east-2 --source-aws-service-signing-name opensearchservice --s3-role-arn arn:aws:iam::767398060394:role/OS-Snapshot
2024-08-02 14:15:03,417 INFO c.r.w.SnapshotRunner [main] Attempting to initiate the snapshot...
2024-08-02 14:15:05,835 ERROR c.r.c.OpenSearchClient [reactor-http-epoll-2] Could not register snapshot repo: _snapshot/migration_assistant_repo. Response Code: 403, Response Message: Forbidden, Response Body: {"message":"Credential should be scoped to correct service: 'es'. "}
2024-08-02 14:15:07,229 ERROR c.r.c.OpenSearchClient [reactor-http-epoll-2] Could not register snapshot repo: _snapshot/migration_assistant_repo. Response Code: 403, Response Message: Forbidden, Response Body: {"message":"Credential should be scoped to correct service: 'es'. "}
2024-08-02 14:15:10,046 ERROR c.r.c.OpenSearchClient [reactor-http-epoll-2] Could not register snapshot repo: _snapshot/migration_assistant_repo. Response Code: 403, Response Message: Forbidden, Response Body: {"message":"Credential should be scoped to correct service: 'es'. "}
2024-08-02 14:15:15,386 ERROR c.r.c.OpenSearchClient [reactor-http-epoll-2] Could not register snapshot repo: _snapshot/migration_assistant_repo. Response Code: 403, Response Message: Forbidden, Response Body: {"message":"Credential should be scoped to correct service: 'es'. "}
2024-08-02 14:15:15,389 ERROR c.r.c.SnapshotCreator [main] Snapshot repo registration failed
reactor.core.Exceptions$RetryExhaustedException: Retries exhausted```

### Check List

### CODE REVIEW

File: RFS/src/main/java/com/rfs/common/http/ConnectionContext.java [hunk 1]
```diff
@@ -24,39 +22,32 @@ public enum Protocol {
...
-    public final String password;
-    public final AuthType authType;
     public final boolean insecure;
-
-    public ConnectionDetails(Params params) {
-        this(params.getHost(), params.getUsername(), params.getPassword(), params.isInsecure());
+    @Getter
+    private final AuthTransformer authTransformer;
+
+    public ConnectionContext(String url, String username, String password, boolean insecure) {
```
peternied, line 29: Can we pair the constructors down so there is only one that accepts the Params object?  IMO any case  where we are assembling this context manually seems like a bug (except maybe tests)

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
```diff
@@ -85,9 +92,24 @@ public static interface Params {
...
+    @Builder
+    @Getter
+    public static class ConstParams implements Params {
+        private String host;
+        private String username;
+        private String password;
+        private String sigv4Region;
+        private String sigv4ServiceName;
+        private boolean insecure;
+    }
```
peternied, line 111: Pretty sure this is unneeded if we require the context be created from Params

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
```diff
@@ -102,6 +124,14 @@ public static class TargetArgs implements Params {
             "--target-password" }, description = "Optional.  The target password; if not provided, will assume no auth on target", required = false)
         public String password = null;
 
+        @Parameter(names = {
+            "--target-sigv4-region" }, description = "The source sigv4 region, if not provided will assume no sigv4 on source", required = false)
+        public String sigv4Region = null;
+
+        @Parameter(names = {
+            "--target-sigv4-service-name" }, description = "The source sigv4 service name, if not provided will assume es if sigv4 region provided", required = false)
+        public String sigv4ServiceName = "es";
```
peternied, line 133: Lets remove this, the service name shouldn't be an exposed parameter - as its tailored to the endpoint that is called, ES for the managed service, maybe something different for the serverless endpoints , etc.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
```diff
@@ -102,6 +124,14 @@ public static class TargetArgs implements Params {
             "--target-password" }, description = "Optional.  The target password; if not provided, will assume no auth on target", required = false)
         public String password = null;
 
+        @Parameter(names = {
+            "--target-sigv4-region" }, description = "The source sigv4 region, if not provided will assume no sigv4 on source", required = false)
```
peternied, line 128: Rename to Target AWS Region or something similar, while this is used for sigv4 signing, its about the region the endpoint is inside.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
```diff
@@ -85,9 +92,24 @@ public static interface Params {
...
+        private String host;
+        private String username;
+        private String password;
+        private String sigv4Region;
+        private String sigv4ServiceName;
+        private boolean insecure;
+    }
+
     @Getter
     public static class TargetArgs implements Params {
```
peternied, line 114: Where is the role assumed for the request passed in?

RFS/src/main/java/com/rfs/common/http/RestClient.java
```diff
@@ -0,0 +1,185 @@
...
+            this.message = responseMessage;
+            this.headers = responseHeaders;
+        }
+    }
+
+    public final ConnectionContext connectionContext;
+    private final HttpClient client;
+    private final AuthTransformer authTransformer;
+
+    @SneakyThrows
```
peternied, line 49: Lets not use `@SneakyThrows` on any library systems since it hides a failure state.  If you want to catch and wrap with an exception derived from RuntimeException thats OK with me rather than a checked exception.  What do you think?

RFS/src/main/java/com/rfs/common/http/RestClient.java
```diff
@@ -0,0 +1,185 @@
...
+            .collect(java.util.stream.Collectors.toMap(
+                Map.Entry::getKey,
+                Map.Entry::getValue,
+                (v1, v2) -> v1 + "," + v2
+            ));
+    }
+
+    public Response get(String path, IRfsContexts.IRequestContext context) {
+        return getAsync(path, context).block();
+    }
```
peternied, line 154: Design nitpick: I'd rather we only expose have async methods with no alternatives, e.g. getAsync(...) -> get(...).  `.block()` is trivial to do inline and I'd prefer that we default to async behavior

RFS/src/main/java/com/rfs/common/http/RestClient.java
```diff
@@ -0,0 +1,185 @@
...
+
+
+    public Mono<Response> asyncRequest(HttpMethod method, String path, String body, Map<String, List<String>> additionalHeaders,
+                                       @Nullable IRfsContexts.IRequestContext context) {
+        Map<String, List<String>> headers = new HashMap<>(Map.of(
+            "User-Agent", List.of("RfsWorker-1.0"),
+            "Host", List.of(getHostHeader(connectionContext.url))
+        ));
+        if (body != null) {
+            headers.put("Content-Type", List.of("application/json"));
```
peternied, line 106: Move those into constants

RFS/src/main/java/com/rfs/common/http/RestClient.java
```diff
@@ -0,0 +1,185 @@
...
+        var convertedHeaders = additionalHeaders.entrySet().stream().collect(Collectors
+            .toMap(Map.Entry::getKey, e -> List.of(e.getValue())));
+        return asyncRequest(method, path, body, convertedHeaders, context);
+    }
+
+
+    public Mono<Response> asyncRequest(HttpMethod method, String path, String body, Map<String, List<String>> additionalHeaders,
+                                       @Nullable IRfsContexts.IRequestContext context) {
+        Map<String, List<String>> headers = new HashMap<>(Map.of(
+            "User-Agent", List.of("RfsWorker-1.0"),
```
peternied, line 102: This agent string seems like it should come from the irequest context

RFS/src/main/java/com/rfs/common/http/SigV4AuthTransformer.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class SigV4AuthTransformer implements AuthTransformer {
+    private final SigV4Signer signer;
+
+    public SigV4AuthTransformer(AwsCredentialsProvider credentialsProvider, String service, String region, String protocol, Supplier<Clock> timestampSupplier) {
+        this.signer = new SigV4Signer(credentialsProvider, service, region, protocol, timestampSupplier);
+    }
+
+    @Override
+    public Mono<TransformedRequest> transform(String method, String path, Map<String, List<String>> headers, Mono<String> body) {
```
peternied, line 23: Is there a reason that body is a string? I think we should be able to sign any body type such as binary / compressed data.

RFS/src/main/java/com/rfs/cms/AbstractedHttpClient.java
peternied, line 1: Nit: Abstract**ed**HttpClient.java -> AbstractHttpClient.java

RFS/src/main/java/com/rfs/common/http/AuthTransformer.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import reactor.core.publisher.Mono;
+
+public interface AuthTransformer {
+    Mono<TransformedRequest> transform(String method, String path, Map<String, List<String>> headers, Mono<String> body);
```
peternied, line 12:
```suggestion
    Mono<TransformedRequest> applyAuth(String method, String path, Map<String, List<String>> headers, Mono<String> body);
```

RFS/src/main/java/com/rfs/common/http/AuthTransformer.java
```diff
@@ -0,0 +1,39 @@
...
+            List<String> authHeader = new ArrayList<>();
+            authHeader.add("Basic " + encodedCredentials);
+            newHeaders.put("Authorization", authHeader);
+            return Mono.just(new TransformedRequest(newHeaders, body));
+        };
+    }
+
+    static AuthTransformer noAuth() {
+        return (method, path, headers, body) -> Mono.just(new TransformedRequest(new HashMap<>(headers), body));
+    }
```
peternied, line 28: Nit: Move these out into their own class files + tests

RFS/src/main/java/com/rfs/common/http/AuthTransformer.java
```diff
@@ -0,0 +1,39 @@
...
+
+    class TransformedRequest {
+        public final Map<String, List<String>> headers;
+        public final Mono<String> body;
+
+        public TransformedRequest(Map<String, List<String>> headers, Mono<String> body) {
+            this.headers = headers;
+            this.body = body;
+        }
+    }
```
peternied, line 38: Lombok this for less code
```suggestion
    @AllArgsConstructor
    @Getter
    class TransformedRequest {
        final Map<String, List<String>> headers;
        final Mono<String> body;
    }
```

RFS/src/main/java/com/rfs/common/http/RestClient.java
```diff
@@ -0,0 +1,185 @@
...
+import lombok.SneakyThrows;
+import reactor.core.publisher.Mono;
+import reactor.netty.Connection;
+import reactor.netty.http.client.HttpClient;
+import reactor.netty.http.client.HttpClientRequest;
+import reactor.netty.tcp.SslProvider;
+import reactor.util.annotation.Nullable;
+
+public class RestClient {
+    public static class Response {
```
peternied, line 31: Move this out into its own file, also you can lombok away the constructor

RFS/src/main/java/com/rfs/common/http/RestClient.java
```diff
@@ -0,0 +1,185 @@
...
+
+        public Response(int responseCode, String responseBody, String responseMessage, Map<String, String> responseHeaders) {
+            this.code = responseCode;
+            this.body = responseBody;
+            this.message = responseMessage;
+            this.headers = responseHeaders;
+        }
+    }
+
+    public final ConnectionContext connectionContext;
```
peternied, line 45: Please do not use public fields, only getters

RFS/src/main/java/com/rfs/common/http/SigV4AuthTransformer.java
```diff
@@ -0,0 +1,52 @@
...
+
+                @Override
+                public String path() {
+                    return path;
+                }
+
+                @Override
+                public String protocol() {
+                    return "HTTP/1.1";
+                }
```
peternied, line 40: This is kind of strange to include, can we avoid referencing the protocol here?

peternied (commented):
Good work here, I know some of this is still subject to chat, but getting these thoughts out there as you iterate.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: This is different for serverless, "aoss".

Should we instead have a parameter for cluster type? Standalone, Amazon OpenSearch Service, Amazon OpenSearch Serverless Service?

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: Do we want to require this parameter only if Sigv4 is enabled?

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: This is used in unit tests. Alternatively i can move this to a TestFixtures

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: When creating a ConnectionContext, we use `DefaultCredentialsProvider.create()` which is consistent with the Replayer and RFS S3 Access https://github.com/opensearch-project/opensearch-migrations/blob/7ab6fc46c983b71bf190f43f9948ee3693034ca0/RFS/src/main/java/com/rfs/common/S3Repo.java#L91

If we think we need the feature to use credentials to assume a different role, let's capture that in a JIRA to update it throughout the solution.

RFS/src/main/java/com/rfs/common/http/RestClient.java
AndreKurait: Github is making this look like it's all new code whereas i just moved the directory here. I'll update the commit history to reflect it's just a file move

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: Done

RFS/src/main/java/com/rfs/common/http/RestClient.java
AndreKurait: While this @SneakyThrows was pre-existing in the RestClient, i've removed it.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
peternied: I'd rather not have to ask - if its sigv4, it has to be an Amazon managed system.  🤞 we can determine AOS from AOSS without the user specifying anything.

RFS/src/main/java/com/rfs/common/http/RestClient.java
peternied: Oh thanks for the context, fine with me either way.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: The only way i see we can accurately determine AOS from AOSS (consider custom domain names/dns redirection/proxy layers) is by sending a failing request to the server and parsing the reject body. Let me know if you like this approach/have a better idea.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: https://opensearch.atlassian.net/browse/MIGRATIONS-1895

RFS/src/main/java/com/rfs/common/http/RestClient.java
AndreKurait: This existed in the existing RestClient implementation. Made the requested change as well

peternied: Thanks!

RFS/src/main/java/com/rfs/common/http/SigV4AuthTransformer.java
AndreKurait: This is the existing IHttpMessage spec. I've made this implementation now throw UnsupportedOperationException.

RFS/src/main/java/com/rfs/common/http/RestClient.java
AndreKurait: done

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
AndreKurait: Discussed offline. We'll keep a parameter `aws-service-signing-name` to specify this

RFS/src/main/java/com/rfs/common/http/AuthTransformer.java
AndreKurait: complete

RFS/src/main/java/com/rfs/common/http/RestClient.java
AndreKurait: Request context is not always present. Discussed offline and will keep it as it is for now

RFS/src/main/java/com/rfs/common/http/SigV4AuthTransformer.java
AndreKurait: Body was a string for existing rest client compatibility. I've since moved it to a bytebuffer

RFS/src/main/java/com/rfs/common/http/AuthTransformer.java
AndreKurait: Renamed AuthTransformer to RequestTransformer instead

RFS/src/test/java/com/rfs/common/ConnectionContextTest.java
```diff
@@ -0,0 +1,151 @@
+package com.rfs.common;
```
AndreKurait, line 1: This file stemmed from ConnectionDetailsTest.java

RFS/src/main/java/com/rfs/cms/AbstractedHttpClient.java
AndreKurait: I'm not sure we're going to keep this class long term anyways. I'll leave as is

AndreKurait:
@gregschohn, Please take a look at the changes to the awsUtilities SigV4Signer

AndreKurait:
Adding @peternied as assigner as he reviewed the draft pr

peternied (approved):
Great work - just minor comment, overall looks good to me

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/transform/SigV4AuthTransformerFactoryTest.java
```diff
@@ -44,18 +44,19 @@ public void testSignatureProperlyApplied() throws Exception {
         var mockCredentialsProvider = new MockCredentialsProvider();
         DefaultHttpHeaders expectedRequestHeaders = new DefaultHttpHeaders();
         // netty's decompressor and aggregator remove some header values (& add others)
-        expectedRequestHeaders.add("host", "localhost");
+        expectedRequestHeaders.add("Host", "localhost");
```
gregschohn, line 47: Can you point to a test (for the replayer) that shows that when host (& maybe other header values) have irregular casing, that the signature still works just fine?  I noticed a number of Map() projections by copying which could have destroyed the case insensitive semantics of StrictCaseInsensitiveHttpHeadersMap and ListKeyAdaptingCaseInsensitiveHeadersMap.

awsUtilities/src/main/java/org/opensearch/migrations/aws/SigV4Signer.java
```diff
@@ -61,76 +48,67 @@ public SigV4Signer(
...
+            readBuffer.get(temp);
+            outputStream.write(temp, 0, temp.length);
         }
-        if (messageDigest == null) {
-            try {
-                this.messageDigest = MessageDigest.getInstance("SHA-256");
-            } catch (NoSuchAlgorithmException e) {
-                throw Lombok.sneakyThrow(e);
-            }
+    }
```
gregschohn, line 67: This code is far inferior to the old code. We may have multiple 100MB requests going through the replayer at a time.  All that you need here is a checksum, so there's no reason to buffer hundreds of MBs of data and put the whole process at risk of crashing or thrashing.  Please change this.  There's a reason that the API and implementation were constructed as they were before.

gregschohn (dismissed):
#1 concern - please bring back the streaming approach to payload processing for sigv4 rather than buffering.  I don't believe that the latter is tenable for the loads that I've seen go through some clusters.  There were a number of other copies that should be replaced with adapters.  In some cases, the copies may break existing behaviors.

CreateSnapshot/src/main/java/com/rfs/CreateSnapshot.java
```diff
@@ -52,6 +52,10 @@ public static class Args {
             "--max-snapshot-rate-mb-per-node" }, required = false, description = "The maximum snapshot rate in megabytes per second per node")
         public Integer maxSnapshotRateMBPerNode;
 
+        @Parameter(names = {
+            "--s3-role-arn" }, required = false, description = "The role arn to use for Amazon Managed Service s3 snapshots")
```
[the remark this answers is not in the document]
AndreKurait: Will replace with `The role ARN the cluster will assume to write a snapshot to S3`

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
```diff
@@ -0,0 +1,161 @@
...
+
+        boolean isInsecure();
+
+        default ConnectionContext toConnectionContext() {
+            return new ConnectionContext(this);
+        }
+    }
+
+    @Getter
+    public static class TargetArgs implements IParams {
```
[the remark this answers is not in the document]
peternied: Dynamic parameters gives us property bags - much easier to introduce bugs if something isn't found in the bag.  I'd rather we stick to fully fleshed out types even if we have some largely duplicate classes.  Note; I hope to improve a little this as we iterate on the metadata cli experience.

gregschohn (commented):
Thanks for changing the Signer class back.  We'll address the deprecation in a future commit.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/transform/SigV4AuthTransformerFactoryTest.java
AndreKurait: Added to [SigV4AuthTransformerFactoryTest.java](https://github.com/opensearch-project/opensearch-migrations/pull/835/files#diff-5d7d4229d1ede018e9c0028fc658fc3b06c246d3da24520d82ce040bd5afd836)

AndreKurait:
> #1 concern - please bring back the streaming approach to payload processing for sigv4 rather than buffering. I don't believe that the latter is tenable for the loads that I've seen go through some clusters. There were a number of other copies that should be replaced with adapters. In some cases, the copies may break existing behaviors.

I addressed this concern and many of the nit's that you left

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -1,132 +1,182 @@
...
-                .trustManager(InsecureTrustManagerFactory.INSTANCE)
-                .build();
-
-            sslProvider = SslProvider.builder().sslContext(sslContext).handlerConfigurator(sslHandler -> {
-                SSLEngine engine = sslHandler.engine();
-                SSLParameters sslParameters = engine.getSSLParameters();
-                sslParameters.setEndpointIdentificationAlgorithm(null);
-                engine.setSSLParameters(sslParameters);
-            }).build();
+        if (connectionContext.insecure) {
```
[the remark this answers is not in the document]
AndreKurait: Right now this is a parameter exposed through migration console. If we're renaming existing parameters i'd like to do it all together

RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -1,132 +1,182 @@
...
+                            response.status().reasonPhrase(),
+                            extractHeaders(response.responseHeaders())
+                        ))
+                ))
+            .doOnError(t -> {
+                if (context != null) {
+                    context.addTraceException(t, true);
+                }
+            })
+            .doOnTerminate(() -> contextCleanupRef.get().run());
```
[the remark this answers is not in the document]
AndreKurait: Updated.

RFS/src/main/java/com/rfs/common/http/ConnectionContext.java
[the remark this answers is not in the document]
gregschohn: What does "not found in the bag" mean?  I'd expect with modularity and commonality, the cost/benefit of testing would improve

PR #838, opened 2024-07-24
Include services.yaml hash in migration console 

AndreKurait:
### Description
Currently, some changes to the cdk.context.json will result in only the yaml getting updated in SSM and not trigger a re-deployment of the migration console which will result in stale parameters being used. This change alleviates this by updating generating a hash to put in the task definition so that a new task definition is created every time an update to the yaml is made.

* Category : Bug Fix
* Why these changes are required? Get rid of stale serivices.yaml
* What is the old behavior before changes and new behavior after changes? Sometimes, console would not be updated and remain with old values, now is fixed.

### Issues Resolved
[MIGRATIONS-1888](https://opensearch.atlassian.net/browse/MIGRATIONS-1888)

### Testing
Deployed to AWS and verified successful startup and env variables include hash
image

### Check List


peternied (changes requested):
Thanks for integrating this, lets add test case to confirm that after the configuration is updated the migration console is redeployed.  I think this can be done by checking this by comparing the logical id of the container from `Template.fromStack(...)`.

AndreKurait:
> Thanks for integrating this, lets add test case to confirm that after the configuration is updated the migration console is redeployed. I think this can be done by checking this by comparing the logical id of the container from `Template.fromStack(...)`.

Added this @peternied

peternied:
@AndreKurait Looks like the CDK tests are failing?

mikaylathompson (approved):
Good solution for this, thanks!

mikaylathompson:
You might need something like this: https://github.com/opensearch-project/opensearch-migrations/blob/main/deployment/cdk/opensearch-service-migration/test/migration-services-yaml.test.ts#L10-L11

PR #841, opened 2024-07-25
Add `--no-daemon` to jenkins gradle command

mikaylathompson:
### Description
Jenkins is encountering a pattern where the gradle build fails every other run with:
```
* What went wrong:
Gradle build daemon has been stopped: since the JVM garbage collector is thrashing
```

In these cases, the gradle daemon expires and the next build succeeds. The following build fails again, indicating that a gradle daemon can "handle" only one run before running out of memory. We don't encounter this on github actions because we run with a new daemon each time.

This flag will run gradle without a daemon on Jenkins, thus avoiding the long-living-daemon-runs-out-of-memory issue.

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/vars/defaultIntegPipeline.groovy
+++ b/vars/defaultIntegPipeline.groovy
@@ -50,7 +50,7 @@ def call(Map config = [:]) {
             stage('Build') {
                 steps {
                     timeout(time: 1, unit: 'HOURS') {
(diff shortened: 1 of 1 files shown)


gregschohn (approved):
There's likely an underlying bug that we need to solve, but for Jenkins, we should be using a fresh daemon every time anyway!

PR #848, opened 2024-07-26
Replayer retries

gregschohn:
### Description

Add support for RequestSenderOrchestrator.scheduleRequest() to cause retries when connections couldn't be made (retrying indefinitely, until a connection CAN be made) or when the target response is worse than the source response (limited to 4 retries).  This precipitated other changes throughout the system.  

Below are those other changes, which are mostly relevant for developers.  However, some of these changes will have minor user-impact.

* Output tuple responses are now lists of responses rather than a single response.  The humanReadableLogs script has been updated to decode each of the list elements.
* HttpByteBufFormatter relied upon HttpObjectAggregator to parse headers (and the status code) but would throw an exception when the body was larger than the pre-configured size.  Multiple areas of code (often for outputting/diagnostics) have needed to clip the contents of HTTP messages, but we hadn't tested beyond capacities.  Now an internal 'TruncatingAggregator' class accepts the headers and then accumulates initial bytes until the limit is satisfied, consuming/dropping the remaining ones.  Other helper methods have been added to HttpByteBufFormatter to facilitate easily setting up an EmbeddedChannel/HTTP pipeline that additional custom handlers can be applied to.  One such custom handler is the BulkErrorFindingHandler, which uses a json streaming parser to find if bulk responses had any errors (used to determine if a retry is necessary).
* HttpRequestTransformationStatus has been converted from an enum to a class so that causal Exceptions can be tied to the result and propagated through the system more easily.  Those causal exceptions were easy to drop/ignore before.
* To support retrying connection failures, how/when/where connections are created had been updated.  ConnectionReplaySession has less responsibility to create connections.  The supplier previously passed to it is now a BiFunction that takes an EventLoop and an ITargetRequestContext.  getChannelFutureInActiveState() now delegates to the BiFunction unless the cached channel is still active.  It's up to the caller to deal with any exception that may have been propagated by the BiFunction.  The BiFunction is provided by NettyPacketToHttpConsumer, which reduces the coupling between the ConnectionReplaySession and the HTTP subsystem.  That allows the ConnectionReplaySession to act much more as a plain old datatype/cached value.
* Since connections are reused between requests, but created due to a request, there's a new context (IRequestConnectingContext) under the target request.  This won't be present when a connection has been reused, but otherwise will be when a new connection is required.
* Remove AdaptiveRateLimiter.  It was meant for some of the retry stuff that this PR pertains to.  Since it wasn't used even here, it seemed better to kill the dependency altogether.
* I've got yet another toxiproxy wrapper.  Notice that there is also one for CaptureProxy tests (ToxiproxyContainerTestBase).  This is just test code, but merging their behaviors/implementations together would be a great future improvement.
* Be more careful about how connections are terminated when the system is being shutdown (in NettyFutureBinders).  This probably never impacted the replayer, but we shut the replayer down a lot in unit tests.  Some of the changes in connection handling exacerbated deadlocks.

* Category: Enhancement, New feature
* Why these changes are required? To make the traffic to a target more reliable.
* What is the old behavior before changes and new behavior after changes?  No retries for any kind of transient error before.  Now we'll try harder :)

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1214
https://opensearch.atlassian.net/browse/MIGRATIONS-1311

### Testing
Gradle unit testing, manual docker-compose OSB tests + CICD tests

### Check List


AndreKurait (commented):
25% done with this review, here are my notes, nothing blocking

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/RequestSenderOrchestrator.java
```diff
@@ -300,27 +366,74 @@ private Duration getDelayFromNowMs(Instant to) {
...
+                    }
+                },
+                () -> "checking response to determine if the request should be retried")
+            .getDeferredFutureThroughHandle((dtr,t) -> {
+                if (t != null) {
+                    return TextTrackedFuture.failedFuture(t, () -> "failed future");
+                }
+                if (dtr.directive == RetryDirective.RETRY) {
+                    var newStartTime = referenceStartTime.plus(nextRetryDelay);
+                    log.atInfo().setMessage(() -> "Making request scheduled at " + newStartTime).log();
```
AndreKurait, line 400: nit: info -> debug

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/RequestSenderOrchestrator.java
```diff
@@ -300,27 +366,74 @@ private Duration getDelayFromNowMs(Instant to) {
...
+                           Duration nextRetryDelay,
+                           Duration interval,
+                           RetryVisitor<T> visitor)
+    {
+        if (eventLoop.isShuttingDown()) {
+            return TextTrackedFuture.failedFuture(new IllegalStateException("EventLoop is shutting down"),
+                () -> "sendRequestWithRetries is failing due to the pending shutdown of the EventLoop");
+        }
+        return sendPackets(senderSupplier.get(), eventLoop,
+            byteBufList.streamRetained().iterator(), referenceStartTime, interval, new AtomicInteger())
```
AndreKurait, line 387: I'd prefer this to use refSafeMap to retain the stream and ensure a stream close at the end of the future

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TransformedTargetRequestAndResponseList.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    @Override
+    public String toString() {
+        final StringBuilder sb = new StringBuilder("TransformedTargetRequestAndResponse{");
+        sb.append(", transformStatus=").append(getTransformationStatus());
+        sb.append(responseList.stream()
+            .map(AggregatedRawResponse::toString)
+            .collect(Collectors.joining("\n", "[", "]")));
+        return sb.toString();
```
AndreKurait, line 57: Did you intend to close this with `}`, also, i don't see needing the `, ` in transformStatus on line 53

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/DefaultRetry.java
```diff
@@ -0,0 +1,54 @@
...
+import org.opensearch.migrations.replay.AggregatedRawResponse;
+import org.opensearch.migrations.replay.IRequestResponsePacketPair;
+import org.opensearch.migrations.replay.RequestSenderOrchestrator;
+import org.opensearch.migrations.replay.util.TextTrackedFuture;
+import org.opensearch.migrations.replay.util.TrackedFuture;
+
+import io.netty.buffer.ByteBuf;
+
+public class DefaultRetry implements RequestRetryEvaluator {
+    static final int MAX_RETRIES = 4;
```
AndreKurait, line 14: Did we talk about this being infinite by default?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
```diff
@@ -152,20 +168,87 @@ private static String prettyPrintNettyMessage(StringJoiner sj, boolean sorted, H
...
+                HttpContent httpContent = (HttpContent) msg;
+                ByteBuf content = httpContent.content();
+                var contentSize = content.readableBytes();
+                if (contentSize > bytesLeftToRead) {
+                    var bytesToTruncate = contentSize - bytesLeftToRead;
+                    bytesDropped += bytesToTruncate;
+                    content.writerIndex(content.writerIndex() - bytesToTruncate);
+                }
+                bytesLeftToRead -= content.readableBytes();
+                aggregatedContents.addComponent(true, content);
```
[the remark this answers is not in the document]
gregschohn: Thanks.  I've adjusted this to not add the buffer if it had 0 bytes.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpMessageAndTimestamp.java
```diff
@@ -55,6 +57,17 @@ public HttpMessageAndTimestamp(Instant firstPacketTimestamp) {
         this.packetBytes = new RawPackets();
     }
 
+    public ByteBuf asByteBuf() {
+        var compositeBuf = Unpooled.compositeBuffer();
+        packetBytes.stream()
+            .map(Unpooled::wrappedBuffer)
+            .forEach(buffer -> {
+                compositeBuf.addComponent(buffer);
```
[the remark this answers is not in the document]
gregschohn: thanks.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TransformedTargetRequestAndResponseList.java
gregschohn: Removed the leading comma.  I left the prefix suffix as `[`/`]` since the contents are of a list.  Why were you asking about `}`?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -386,11 +437,17 @@ public TrackedFuture<String, AggregatedRawResponse> finalizeRequest() {
...
-        }, () -> "Waiting for previous consumes to set the future").map(f -> f.whenComplete((v, t) -> {
-            if (channel == null) {
-                log.atInfo().setMessage(() -> "").log();
-            } else {
-                deactivateChannel();
+        }, () -> "Waiting for previous consumes to set the future")
+        .map(f -> f.whenComplete((v, t) -> {
+            try {
+                if (channel == null) {
+                    log.atInfo().setMessage(() -> "").log();
```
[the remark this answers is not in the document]
gregschohn: agree.  I added some text and made it trace.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/DefaultRetry.java
gregschohn: You & I talked about infinite as the default for _connection_ errors and it still is.  This is for errors like 404 or 5xx that are retryable.  If there's a single request that causes issues, we'll want to give it some fair tries, but then move on.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/DefaultRetry.java
```diff
@@ -0,0 +1,54 @@
...
+
+    public static boolean retryIsUnnecessaryGivenStatusCode(int statusCode) {
+        switch (statusCode) {
+            case 200:
+            case 201:
+            case 401:
+            case 403:
```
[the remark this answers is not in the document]
gregschohn: We will still want forward progress to be made.  There are a lot of users (logs users for one) that may not do a historical transfer first.  We don't want a 404 to block a replay - which is what will eventually happen (once we synchronize time).  Even just today, retrying indefinitely will block any other requests from the same connection from going through.

That said, this list is for target response codes that require no further inspection - upon seeing these codes in the target response, we're done.  If we got a 200, we don't retry under any circumstances, even if the source response wasn't 200.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TransformedTargetRequestAndResponseList.java
AndreKurait: >Why were you asking about }?

The start of this string is `TransformedTargetRequestAndResponse{` and that `{` is never closed

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/DefaultRetry.java
AndreKurait: Sounds good

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/DefaultRetry.java
```diff
@@ -0,0 +1,54 @@
...
+                return true;
+            default:
+                return statusCode >= 300 && statusCode < 400;
+        }
+    }
+
+    public TextTrackedFuture<RequestSenderOrchestrator.RetryDirective>
+    getRetryDirectiveUnlessExceededMaxRetries(List<AggregatedRawResponse> previousResponses) {
+        var d = previousResponses.size() > MAX_RETRIES ?
```
AndreKurait, line 30: nit: is current response in previous retries? If not, this should be a `>=`

 (3 previousResponses + current response == 4 calls == 1 original call and 3 retries)

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/OpenSearchDefaultRetry.java
AndreKurait, line 1: Discussed offline. A few notes that i'd like to see.

1. There's no retry limit for failed bulk requests
2. There's no comparison between source and target status code for non-bulk requests
3. (Unit tests for the above) 
4. (Existing Comment) - there's an off by one on the retry count

AndreKurait (changes requested):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/DefaultRetry.java
```diff
@@ -40,15 +44,48 @@ public static boolean retryIsUnnecessaryGivenStatusCode(int statusCode) {
...
+                    () -> "checking the accumulated source response value");
+            }
+        } else {
+            return TextTrackedFuture.completedFuture(RequestSenderOrchestrator.RetryDirective.RETRY, () -> "retrying");
         }
-        return getRetryDirectiveUnlessExceededMaxRetries(previousResponses);
+    }
+
+    private RequestSenderOrchestrator.RetryDirective shouldRetry(int sourceStatus, int targetStatus) {
+        return targetStatus >= 300 && sourceStatus < 300 ?
```
AndreKurait, line 87: A case that's interesting to me here is if the source is 404 and target is 500, we won't retry. This is probably fine from a side effect perspective, but we should at least document this from a test case and written documentation

AndreKurait (approved):

AndreKurait:
Please update the PR overview before merging

PR #850, opened 2024-07-29
Always run all checks job

peternied:
### Description
When the "all-ci-checks-pass" was marked as `Skipped` it wouldn't prevent merging.  I've change the job to always run, even if a downstream job hasn't passed, and it validates the results to be sure that all jobs completed successful.

Note; created this PR as a follow up after contacting GitHub support over the `Required` check not working as expected.

### Testing
- Failure case verified with [#1](https://github.com/peternied/opensearch-migrations/actions/runs/10152498555)
image
- Passing case should be confirmed with this PR without any failures.

### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -215,6 +215,7 @@ jobs:
       - python-lint
       - python-tests
       - fetch-migration-docker-build
+    if: always()
     runs-on: ubuntu-latest
     steps:
       - run: |
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -218,5 +218,8 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     steps:
+      - if: ${{ contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'failure') || contains(needs.*.result, 'skipped') }}
+        run: |
+          echo "One or more job cancelled, failed, or skipped" && exit 1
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -215,7 +215,11 @@ jobs:
       - python-lint
       - python-tests
       - fetch-migration-docker-build
+    if: always()
     runs-on: ubuntu-latest
     steps:
+      - if: ${{ contains(needs.*.result, 'cancelled') || contains(needs.*.result, 'failure') || contains(needs.*.result, 'skipped') }}
```
AndreKurait, line 221: When is a job "skipped"?

Say, for example, we have our CDK tests run only if a file in the CDK directory was changed, would that register as skipped for PRs that don't include that?

peternied: We don't have any jobs that are skipped.  If we did want to support conditional execution of some of our jobs we would need to change this code.  Note; I couldn't check for `every(needs.*.result, 'success')` which would have been preferred

AndreKurait: This clearly works with the way our current code is setup. What I was trying to get at would be if excluding `|| contains(needs.*.result, 'skipped')` would be more defensive assuming users can't intentionally skip tasks or error cases aren't coded as skipped

AndreKurait (approved):

PR #852, opened 2024-07-30
Fix connection metering for reactor clients.

gregschohn:
As it was, there was potential for an unbounded amount of repetitive counting because each request had a new pair of metering handlers attached to the connection.  Since reactor doesn't have a clear way to cleanup after each and every single request (consider retries), I've taken the view that every request that we make, first we try to remove any handlers that might have been present for the last request + removing them entirely in the doFinally() block.  Notice that we want to track the values on a request by request basis, hence the reason that we don't want to even reuse the handlers on different requests. I had to update the test to trigger this because the reactor HttpClient will use multiple connections by default, hence the reason that we weren't previously seeing this within the test, though it had become flaky, which is how this was caught.

### Description
Read/Write metering for reactor clients will be more (hopefully, completely) accurate.
* Category: Bug fix/Test fix
* Why these changes are required? We were misrepresenting how many bytes went over the wire and we were also severely bloating memory utilization, to the point that heavily reused clients could have created significant performance penalties.
* What is the old behavior before changes and new behavior after changes? See above

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1890

### Testing
Unit testing

### Check List

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,1 +1,2 @@
 org.gradle.caching=true
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
--- a/RFS/src/main/java/com/rfs/common/RestClient.java
+++ b/RFS/src/main/java/com/rfs/common/RestClient.java
@@ -1,17 +1,21 @@
 package com.rfs.common;
 
 import java.util.Base64;
-import java.util.function.BiConsumer;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.BiFunction;
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLParameters;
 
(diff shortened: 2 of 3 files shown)


gradle.properties
```diff
@@ -1 +1,2 @@
 org.gradle.caching=true
+org.gradle.jvmargs=-Xmx2g -XX:MaxMetaspaceSize=512m -XX:+HeapDumpOnOutOfMemoryError -Dfile.encoding=UTF-8
```
gregschohn, line 2: I needed to do this to get gradle to build reliably

peternied (approved):
Thanks for walking through this with me @gregschohn - the relationship between the metrics / span names being verified so thoroughly doesn't warrant additional testing.

AndreKurait (approved):
LGTM. In the future i'd like to see this get cleaned up with perhaps a more reactor-way to get these metrics instead of going as low level, but for now this works

PR #856, opened 2024-08-01
Create command output models for metadata migration tool

peternied:
### Description
This creates a layer between the command line interface and the operations being done - and the additional operations that will be supported `configuration` and `evaluate`.  Also includes an end to end test case - similar to the all inclusive version in RFS but streamlined with the new structure.

Functionality the metadata migration tool has the same behavior with these modifications.

### Issues Resolved
- https://opensearch.atlassian.net/browse/MIGRATIONS-1864

### Check List


MetadataMigration/src/main/java/org/opensearch/migrations/commands/Evaluate.java
```diff
@@ -0,0 +1,12 @@
+package org.opensearch.migrations.commands;
+
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class Evaluate {
+
+    public EvaluateResult execute() {
+        log.atError().setMessage("configure is not supported").log();
```
mikaylathompson, line 9: minor configure->evaluate

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
```diff
@@ -0,0 +1,49 @@
...
+
+@Slf4j
+public class MetadataMigration {
+
+    public static void main(String[] args) throws Exception {
+        MetadataArgs arguments = new MetadataArgs();
+        JCommander.newBuilder().addObject(arguments).build().parse(args);
+
+        var context = new RootMetadataMigrationContext(
+            RootOtelContext.initializeOpenTelemetryWithCollectorOrAsNoop(arguments.otelCollectorEndpoint, "rfs"),
```
mikaylathompson, line 23: I think this was possibly addressed in a recent PR, but we should set the name here to "metadata" or something similar to ensure these traces are distinguishable from the document migration ones

MetadataMigration/src/test/java/org/opensearch/migrations/EndToEndTest.java
```diff
@@ -0,0 +1,114 @@
...
+    private File localDirectory;
+
+    @Test
+    void metadataMigrateFrom_ES_v7_10_to_OS_v2_14() throws Exception {
+        final var targetVersion = SearchClusterContainer.OS_V2_14_0;
+        try (
+            final var sourceCluster = new SearchClusterContainer(SearchClusterContainer.ES_V7_10_2);
+            final var targetCluster = new SearchClusterContainer(targetVersion)
+        ) {
+            migrateFrom_ES_v7_X(sourceCluster, targetCluster);
```
mikaylathompson, line 42: the expect/assert seems kind of hidden here -- we're expecting it to throw an Exception?

That's because we don't support 2.14? Can you add a little more clarification here on what/why. I assume long-term this won't throw an exception, and I want to make sure it's clear when we should change the expected behavior for this test.

MetadataMigration/src/test/java/org/opensearch/migrations/commands/MigrateTest.java
```diff
@@ -0,0 +1,37 @@
...
+import org.opensearch.migrations.metadata.tracing.RootMetadataMigrationContext;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.mockito.Mockito.mock;
+
+class MigrateTest {
+
+    @Test
+    void migrate_failsInvalidInvalidParameters() {
```
mikaylathompson, line 16: Why `InvalidInvalid`? Maybe something about missing/empty args?

MetadataMigration/src/test/java/org/opensearch/migrations/commands/MigrateTest.java
```diff
@@ -0,0 +1,37 @@
...
+    @Test
+    void migrate_failsUnexpectedException() {
+        var args = new MetadataArgs();
+        args.fileSystemRepoPath = "";
+        var context = mock(RootMetadataMigrationContext.class);
+        var meta = new MetadataMigration(args);
+
+        var configureSource = meta.migrate().execute(context);
+
+        assertThat(configureSource.getExitCode(), equalTo(888));
```
mikaylathompson, line 35: the 999 and 888 look fairly magical/arbitrary from here, maybe worth comparing against the constant that defines it?

mikaylathompson: Ah, I just realized that they're defined privately in Migrate.

Is there a reason the exit codes aren't defined in a independent class so they can be used by all the function classes without being redefined? If not, I'd lean in that direction.

MetadataMigration/src/test/java/org/opensearch/migrations/commands/MigrateTest.java
peternied: Copypasta, fixed

mikaylathompson (approved):
LGTM

PR #857, opened 2024-08-01
[Console] Use credentials from secrets manager (python & CDK)

mikaylathompson:
## Before Merging
Migration console side is verified in a real deployment; currently workign on CDK.

### Description
Adds support for a secrets manager secret to be used as the password value.

### Issues Resolved
[MIGRATIONS-1760](https://opensearch.atlassian.net/browse/MIGRATIONS-1760) & [MIGRATIONS-1900](https://opensearch.atlassian.net/browse/MIGRATIONS-1900)

### Testing
Good unit test coverage. Need to manually test.

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
@@ -1,11 +1,14 @@
 from typing import Any, Dict, Optional
 from enum import Enum
-import requests
-from requests.auth import HTTPBasicAuth
-from cerberus import Validator
 import logging
 import subprocess
 from console_link.models.schema_tools import contains_one_of
+
+import boto3
+from cerberus import Validator
+import requests
+import requests.auth
+from requests.auth import HTTPBasicAuth
 
 requests.packages.urllib3.disable_warnings()  # ignore: type
 
@@ -24,19 +27,18 @@ BASIC_AUTH_SCHEMA = {
     "schema": {
         "username": {
             "type": "string",
-            "required": False,
+            "required": True,
         },
         "password": {
             "type": "string",
             "required": False,
-            "excludes": ["aws_secret_arn"]
         },
-        "aws_secret_arn": {
+        "password_from_secret_arn": {
             "type": "string",
             "required": False,
-            "excludes": ["password"]
         }
     },
+    "check_with": contains_one_of({"password", "password_from_secret_arn"})
 }
 
 SIGV4_SCHEMA = {
@@ -86,6 +88,29 @@ class Cluster:
         elif 'sigv4' in config:
             self.auth_type = AuthMethod.SIGV4
 
+    def _get_basic_auth_password(self) -> str:
+        assert self.auth_type == AuthMethod.BASIC_AUTH
+        assert self.auth_details is not None  # for mypy's sake
+        if "password" in self.auth_details:
+            return self.auth_details["password"]
+        # Pull password from AWS Secrets Manager
+        assert "password_from_secret_arn" in self.auth_details
+        client = boto3.client('secretsmanager')
+        password = client.get_secret_value(self.auth_details["password_from_secret_arn"])
+        return password["SecretString"]
+
+    def _generate_auth_object(self) -> requests.auth.AuthBase | None:
+        if self.auth_type == AuthMethod.BASIC_AUTH:
+            assert self.auth_details is not None  # for mypy's sake
+            password = self._get_basic_auth_password()
+            return HTTPBasicAuth(
+                self.auth_details.get("username", None),
+                password
+            )
+        elif self.auth_type is AuthMethod.NO_AUTH:
+            return None
+        raise NotImplementedError(f"Auth type {self.auth_type} not implemented")
+
     def call_api(self, path, method: HttpMethod = HttpMethod.GET, data=None, headers=None,
                  timeout=None, session=None, raise_error=True, **kwargs) -> requests.Response:
         """
@@ -93,16 +118,8 @@ class Cluster:
         """
         if session is None:
             session = requests.Session()
-        if self.auth_type == AuthMethod.BASIC_AUTH:
-            assert self.auth_details is not None  # for mypy's sake
-            auth = HTTPBasicAuth(
-                self.auth_details.get("username", None),
-                self.auth_details.get("password", None)
-            )
-        elif self.auth_type is AuthMethod.NO_AUTH:
-            auth = None
-        else:
-            raise NotImplementedError(f"Auth type {self.auth_type} not implemented")
+        
+        auth = self._generate_auth_object()
 
         # Extract query parameters from kwargs
         params = kwargs.get('params', {})
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
@@ -94,7 +94,7 @@ class Cluster:
         if "password" in self.auth_details:
             return self.auth_details["password"]
         # Pull password from AWS Secrets Manager
-        assert "password_from_secret_arn" in self.auth_details
+        assert "password_from_secret_arn" in self.auth_details  # for mypy's sake
         client = boto3.client('secretsmanager')
         password = client.get_secret_value(SecretId=self.auth_details["password_from_secret_arn"])
         return password["SecretString"]
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
@@ -96,7 +96,7 @@ class Cluster:
         # Pull password from AWS Secrets Manager
         assert "password_from_secret_arn" in self.auth_details
         client = boto3.client('secretsmanager')
-        password = client.get_secret_value(self.auth_details["password_from_secret_arn"])
+        password = client.get_secret_value(SecretId=self.auth_details["password_from_secret_arn"])
         return password["SecretString"]
 
     def _generate_auth_object(self) -> requests.auth.AuthBase | None:
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cluster.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cluster.py
@@ -1,15 +1,13 @@
+from base64 import b64encode
+import os
+
 import pytest
+from moto import mock_aws
+import boto3
 
 import console_link.middleware.clusters as clusters_
 from console_link.models.cluster import AuthMethod, Cluster
 from tests.utils import create_valid_cluster
-
-# Define a valid cluster configuration
-valid_cluster_config = {
-    "endpoint": "https://opensearchtarget:9200",
-    "allow_insecure": True,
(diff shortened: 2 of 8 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
```diff
@@ -24,19 +27,18 @@
     "schema": {
         "username": {
             "type": "string",
-            "required": False,
+            "required": True,
```
chelma, line 30: Interesting - why change this?  I'm guessing it was a bug to be optional before?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
```diff
@@ -24,19 +27,18 @@
...
-            "excludes": ["aws_secret_arn"]
         },
-        "aws_secret_arn": {
+        "password_from_secret_arn": {
             "type": "string",
             "required": False,
-            "excludes": ["password"]
         }
     },
+    "check_with": contains_one_of({"password", "password_from_secret_arn"})
```
peternied, line 41: Nice!

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
```diff
@@ -86,23 +88,38 @@ def __init__(self, config: Dict) -> None:
...
+    def _get_basic_auth_password(self) -> str:
+        assert self.auth_type == AuthMethod.BASIC_AUTH
+        assert self.auth_details is not None  # for mypy's sake
+        if "password" in self.auth_details:
+            return self.auth_details["password"]
+        # Pull password from AWS Secrets Manager
+        assert "password_from_secret_arn" in self.auth_details
+        client = boto3.client('secretsmanager')
+        password = client.get_secret_value(SecretId=self.auth_details["password_from_secret_arn"])
+        return password["SecretString"]
```
chelma, line 100: I don't think this is a blocker for this PR since it's such a big possible change, but I'm curious to get your thoughts here.  It feels odd to mix input validation and actual work the way this module does; typically you validation everything before doing work as separate layers.  Is there a way to split out the validation so it's all accomplished before any work is done?  It feels like that could make the code easier to read and test.

Maybe we end up with two classes - a ClusterConfig type that takes in the schema and validates its contents and a Cluster type that takes in a reference to a ClusterConfig in order to do things like the `call_api` method.  Or something like that?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cluster.py
```diff
@@ -71,6 +69,53 @@ def test_missing_endpoint_refused():
...
+            "username": "XXXXX",
+            "password": "XXXXXXXXXXXXXX",
+            "password_from_secret_arn": "arn:aws:secretsmanager:us-east-1:12345678912:secret:master-user-os-pass"
+        },
+    }
+    with pytest.raises(ValueError) as excinfo:
+        Cluster(two_passwords)
+    assert "Invalid config file for cluster" in excinfo.value.args[0]
+    assert excinfo.value.args[1]["cluster"][0]['basic_auth'] == [
+        "More than one value is present: ['password', 'password_from_secret_arn']"
```
chelma, line 99: I'm assuming that this specific messaging is generated by `cerberus`.  I'm concerned this places us at risk of having our unit tests fail if they decide to slightly change their formatting, as we're tightly coupled to an implementation detail we don't control.

I'm curious what value-add you see in checking for the specific string on top of a `ValueError` being thrown?  It feels like we should either trust our library to do the right thing or, if we don't trust it (or we want to make sure it's failing for the reasons we expect) we should be catching the exception thrown by `cerberus` at a lower level and re-throwing an exception we control.

I imagine we do this in all of our unit tests, so changing it would be a substantial task.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
mikaylathompson: Yeah, I think this was from when I had a misunderstanding about what was included in the secret before--I originally thought it was username & password.

mikaylathompson: It's much simpler now that it's username and one of [password, password_from_secret_arn]

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
mikaylathompson: First of all, is this comment on these specific lines of code because of the `assert` line? I should add a comment to it, but the config has already been validated, etc. so there's no risk of not having this value, but mypy (type hint checking) is not equally convinced of that unless I add this line.

mikaylathompson: So I don't think we're actually mixing in any input validation. It is doing all the validation before any work is done.

Validation is the schema + lines ~74-78 above.

mikaylathompson: I'll call out why I'm doing the assert though.

chelma: Sorry, this is just were I ended up sticking the comment.  It's really about the entire structure of the responsibilities of the class and, I suspect, this layer of the code in general.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cluster.py
mikaylathompson: I see what you mean. We show these messages to users, so I care about more than just ValueError being thrown, I wanted to make sure that there was enough information to be actionable. So I started by testing for the pieces of the message I care about (is 'password' present, is 'password_from_secret_arn' present?), and ended up simplifying it to just match the whole message.

But you're right that it is more fragile. I'd rather have our tests break than suddenly start showing users a less helpful message without our awareness.

And if we're catching the exception thrown by cerberus, validating that it matches a specific string ("more than one value is present", for instance) and then rethrowing a custom exception, won't all the same tests break if they change their wording? I agree that it's better to have to fix it only in one spot, but I'm not sure it's _that_ much better.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
mikaylathompson: I agree that the class does both validation and functional work.

But to your point:
>  Is there a way to split out the validation so it's all accomplished before any work is done? It feels like that could make the code easier to read and test.


I think the validation is accomplished before work is done, and I think it's limited to a pretty small area of the class without any complex logic in it. Effectively, all of that logic is contained in the schema, which I believe basically functions as the Config object you're describing.

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -255,6 +256,14 @@ export class MigrationConsoleStack extends MigrationServiceCore {
             ]
         })
 
+        const getSecretsPolicy = new PolicyStatement({
+            effect: Effect.ALLOW,
+            resources: [props.servicesYaml.target_cluster.basic_auth?.password_from_secret_arn ?? ""],
+            actions: [
+                "secretsmanager:GetSecretValue"
+            ]
+        });
```
chelma, line 265: Can we please either wrap this in a conditional or use a cdk.CfnConditional so that this isn't generated in the event there is no ARN?  I can see a user (or security reviewer) being surprised to see this in the Cfn stack def if they haven't done anything w/ Secrets.

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -330,7 +330,14 @@ export class StackComposer {
             this.stacks.push(openSearchStack)
             servicesYaml.target_cluster = openSearchStack.targetClusterYaml;
         } else {
-            servicesYaml.target_cluster = { endpoint: targetEndpoint, no_auth: '' }
+            servicesYaml.target_cluster = { endpoint: targetEndpoint }
+            if (fineGrainedManagerUserName) {
```
chelma, line 334: We should probably check for both members (`fineGrainedManagerUserName`, `fineGrainedManagerUserSecretManagerKeyARN`) being present here, yeah?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cluster.py
chelma: Another thing that occurred to me - it probably isn't the best user experience to present this info to the user in this way.  Are we just killing the process and dumping the exception to stdout?  If I'm a user and I see a stack trace from the tool I just ran, I'd normally assume there was a bug or something broken inside the tool.  Typically, I'd expect to see log messages saying I did something wrong, not a stack trace.

It feels like our overall experience here isn't where we want it to be longer term, and that's fine for this PR, but we should be moving a different direction going forward and that's worth talking about.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
chelma: It sounds like we agree this type is performing two responsibilities then - presenting a config spec/validating adherence AND providing operations using that config.  The classic argument of why you'd split these into two types is if you need two very similar but different behaviors for the same config - such as different ways to interact with versioned Elasticsearch/Opensearch clusters.  This seems like an issue we will run into in the near future.

Let's talk about this one more; I don't think we need to resolve the issue in this PR but having some discussion on how we approach the topic going forward could save us time/effort later.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_cluster.py
mikaylathompson: If you have specific thoughts for improving it, happy to put them into a JIRA.

chelma: Per discussion w/ @mikaylathompson - it seems like there's a way to hook into `cerberus` at a lower level and present our own messaging for situations such as this.  This would strike a better balance between leveraging the library and retaining control over the user experience by removing the chance that an update to `cerberus` would break our tests and, potentially, user scripting relying on specific phrases provided by the library.

@mikaylathompson will make a followup task to investigate that approach in the future.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
chelma: Per discussion w/ @mikaylathompson - while I feel like splitting the responsibilities into separate types and using composition to re-incorporate them is the better approach long term due to the isolation and flexibility it provides:
* The current code isn't unreasonable or hard to read
* The tests are currently simple
* There is a path forward to provide different behavior for different versions w/ inheritance

For these reasons, we'll proceed with the current path for the time being until it becomes an issue.

chelma (approved):
LGTM, thanks @mikaylathompson

PR #863, opened 2024-08-03
[Console] Use secrets manager password for metadata & snapshot

mikaylathompson:
### Description
While #857 added support for password-containing-secrets in CDK and when making cluster commands via the Console Library, it did not support them for creating snapshots and migrating metadata. This PR updates the library to pass those through to those commands.

Specifically, the console library still pulls the secret and then passes the password itself to the java command.

It also masks the password in log messages.

### Issues Resolved
Partially [MIGRATIONS-1912](https://opensearch.atlassian.net/browse/MIGRATIONS-1912) (this also needs sigv4 to be complete), I don't think we have a task for the metadata side.

### Testing
Unit & manual

### Check List


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/metadata.py
```diff
@@ -134,80 +134,56 @@ def _init_from_fs_snapshot(self, snapshot: FileSystemSnapshot) -> None:
...
-            subprocess.run(command, stdout=None, stderr=None, text=True, check=True)
-            logger.info(f"Metadata migration for snapshot {self._snapshot_name} completed")
-            return CommandResult(success=True, value="Metadata migration completed")
-        except subprocess.CalledProcessError as e:
-            logger.error(f"Failed to migrate metadata: {str(e)}")
-            return CommandResult(success=False, value=f"Failed to migrate metadata: {str(e)}")
-
-    def _run_as_detached_process(self, command, log_file) -> CommandResult:
+        command_runner = CommandRunner(command_base, command_args,
+                                       password_field="--target-password",
```
AndreKurait, line 181: nit: in the future i can see metadata migrate needing access to source cluster in which case password_field can be renamed sensitive fields and be a list

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/metadata.py
```diff
@@ -134,80 +134,56 @@ def _init_from_fs_snapshot(self, snapshot: FileSystemSnapshot) -> None:
...
-            "--snapshot-name", self._snapshot_name,
-            "--target-host", self._target_cluster.endpoint,
-            "--min-replicas", str(self._min_replicas)
-        ]
+        logger.info("Starting metadata migration")
+        command_base = "/root/metadataMigration/bin/MetadataMigration"
+        command_args = {
+            "--snapshot-name": self._snapshot_name,
+            "--target-host": self._target_cluster.endpoint,
+            "--min-replicas": str(self._min_replicas)
```
AndreKurait, line 142: nit: Should we be converting all the values to string at the end instead?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_command_runner.py
AndreKurait, line 1: Can we test/support argument flags which don't have a value? e.g. `--debug`

How will this be represented in the dict? `{ "--debug" : ""}` or `{"--debug": {}}` or something else? Is there a different data structure which may make this more intuitive?

mikaylathompson: Yeah, I did it with `None`. Maybe I make a `FlagOnly` type that can passed through. I want the dictionary in large part because it makes it much easier to figure out where the password might be.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/command_runner.py
```diff
@@ -29,11 +33,15 @@ def run(self) -> CommandResult:
         return self._run_as_synchronous_process()
 
     def sanitized_command(self) -> List[str]:
-        if (not self.password_field) or (self.password_field not in self.command_args):
+        if not self.sensitive_fields:
             return self.command
```
AndreKurait, line 37: nit: should we return command.copy() to be consistent here?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/command_runner.py
```diff
@@ -29,11 +33,15 @@ def run(self) -> CommandResult:
...
         # Check whether the value of the password field is in the command
-        password_index = self.command.index(self.command_args[self.password_field])
-        return self.command[:password_index] + ["*" * 8] + self.command[password_index + 1:]
+        display_command = self.command.copy()
+        for field in self.sensitive_fields:
+            if field in display_command:
+                password_index = display_command.index(self.command_args[field])
```
AndreKurait, line 42: This is a bit unintuitive, we're taking the fields as inputs, but looking for the value and replacing it (instead of reconstructing the command_args dict and reconstructing the display command).

There's a slight bug in this logic in that we're sensitizing the first value that we find. e.g.

Given args: `--username admin --password admin`
Our Display_Command will be `--username ******** --password admin`

AndreKurait (changes requested):

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/command_runner.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def run(self) -> CommandResult:
+        if self.run_as_detached:
+            return self._run_as_detached_process(self.log_file)
+        return self._run_as_synchronous_process()
+
+    def sanitized_command(self) -> List[str]:
+        if not self.sensitive_fields:
+            return self.command
+        # Check whether the value of the password field is in the command
```
AndreKurait, line 38: nit: replace password with sensitive_value in this function

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/command_runner.py
mikaylathompson: Oh, great point. I'll add that as a test and rework it.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/command_runner.py
```diff
@@ -0,0 +1,70 @@
...
+    def __init__(self, command_root: str, command_args: Dict[str, Any], sensitive_fields: Optional[List[str]] = None,
+                 run_as_detatched: bool = False, log_file: Optional[str] = None):
+        self.command_args = command_args
+        self.command = [command_root]
+        for key, value in command_args.items():
+            self.command.append(key)
+            if value is not FlagOnlyArgument:
+                if type(value) is not str:
+                    value = str(value)
+                self.command.append(value)
```
AndreKurait, line 24: nit: do we need to add \"? Can the values contain spaces? Do we have a test that tests for this?

password: my password

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/command_runner.py
```diff
@@ -0,0 +1,70 @@
...
+            return self._run_as_detached_process(self.log_file)
+        return self._run_as_synchronous_process()
+
+    def sanitized_command(self) -> List[str]:
+        if not self.sensitive_fields:
+            return self.command
+        display_command = self.command.copy()
+        for field in self.sensitive_fields:
+            if field in display_command:
+                field_index = display_command.index(field)
```
AndreKurait, line 41: nit: I'm still a bit hesitant on this logic and would rather see us use command_args for this. It's unlikely to cause a bug, but the possibility is still there to log something sensitive if another bug existed in the system.

E.g. 1:
if command_root == sensitive_field, there will be an issue (shouldn't happen)

e.g 2:
If value == sensitive field, there will be an issue (a lot less likely than before, but could still happen due to other issue.

Say someone accidentally sets their username as "--target-password", then password will be logged in plain text.

AndreKurait (approved):
Please don't merge this until after #866

PR #865, opened 2024-08-05
Sigv4 console and cdk

mikaylathompson:
### Description
First, this PR adds sigv4 signing
1. as a concept in the services.yaml
2. to the documentation for the services.yaml
3. To the Cluster object
**TODO** Either in this PR after #863 is merged, or in an independent follow-up PR, the sigv4 credentials need to be passed to CreateSnapshot & MigrateMetadata.

This then led to the matter of how we know whether to add `sigV4` to the services.yaml. It's actually a surprisingly confusing situation even when we're creating the domain, but in discussion with @AndreKurait and @chelma we ended up deciding that if basic auth credentials aren't specified, we can do sigv4.

For a preexisting cluster, there's currently no way at all to specify it. I took a stab at adding a `prexisting target cluster` object to the cdk context and using that to specify all details of the pre-existing cluster. In order to create the SSM parameters that other stacks use, I ended up making a entire "pre-existing cluster" stack.

I'm far from convinced that this is the right solution on the CDK side, looking for feedback on this implementation.

### Issues Resolved
[[Console] Implement sigv4 signing for clusters](https://opensearch.atlassian.net/browse/MIGRATIONS-1759)

(in the todo) [Support basic auth and sigv4 for CreateSnapshot](https://opensearch.atlassian.net/browse/MIGRATIONS-1912)

### Testing

### Check List


deployment/cdk/opensearch-service-migration/options.md
```diff
@@ -76,24 +76,23 @@ These tables list all CDK context configuration values a user can specify for th
...
+| ------------------------------ | ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| prexistingTargetClusterDetails | object | <code>{"endpoint": "https://vpc-demo-opensearch-cluster-cv6hggdb66ybpk4kxssqt6zdhu.us-west-2.es.amazonaws.com:443",<\br>"basic_auth": {"username": admin, "password_from_secret_arn": "arn:aws:secretsmanager:us-east-1:12345678912:secret:master-user-os-pass-123abc"}<\br>}`</code> | This object configures all settings for a pre-existing target cluster. If this is provided, a domain will NOT be deployed. Full defintion of the fields is available [here](TODO). |
```
AndreKurait, line 84: The`<\br>` here doesn't look like it's showing up correctly

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -227,10 +228,10 @@ export class StackComposer {
             console.warn("Addon deployments will use the original deployment 'vpcId' regardless of passed 'vpcId' values")
         }
         let targetEndpoint
-        if (targetClusterEndpoint && osContainerServiceEnabled) {
+        if (prexistingTargetClusterDetails && osContainerServiceEnabled) {
             throw new Error("The following options are mutually exclusive as only one target cluster can be specified for a given deployment: [targetClusterEndpoint, osContainerServiceEnabled]")
```
AndreKurait, line 232: Fix error message to specify correct key

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -207,7 +209,6 @@ export class StackComposer {
         const captureProxyExtraArgs = this.getContextForType('captureProxyExtraArgs', 'string', defaultValues, contextJSON)
         const elasticsearchServiceEnabled = this.getContextForType('elasticsearchServiceEnabled', 'boolean', defaultValues, contextJSON)
         const kafkaBrokerServiceEnabled = this.getContextForType('kafkaBrokerServiceEnabled', 'boolean', defaultValues, contextJSON)
-        const targetClusterEndpoint = this.getContextForType('targetClusterEndpoint', 'string', defaultValues, contextJSON)
```
AndreKurait, line 210: Removing a parameter is a major change, let's make sure we get input from the whole team on this as it will break all existing cdk.context.json that we/customers have which specify targetClusterEndpoint. 

We should think about if we want to support both fields for a time period.

Another approach could be to add "targetAuthConfig" as a separate key from the endpoint and not change this key, which I'm in favor of as I see it as less intrusive.

If we move forward with removing this, let's make sure to scrub it from all our wiki's and documentation, i see a reference to this in the cdk README remaining.

AndreKurait (changes requested):

deployment/cdk/opensearch-service-migration/lib/opensearch-domain-stack.ts
```diff
@@ -121,7 +121,9 @@ export class OpenSearchDomainStack extends Stack {
     if (adminUserName) {
         targetCluster.basic_auth = new ClusterBasicAuth({ username: adminUserName, password_from_secret_arn: adminUserSecret?.secretArn })
     } else {
-      targetCluster.no_auth = ''
+        targetCluster.sigv4 = new ClusterSigV4Auth();
+        targetCluster.sigv4.region = this.region;
+        targetCluster.sigv4.service = "es";
```
AndreKurait, line 126: Can this not be overridden?

deployment/cdk/opensearch-service-migration/lib/service-stacks/prexisting-clusters-stack.ts
AndreKurait, line 1: Let's not make multiple stacks publish the same SSM parameters. That can cause deployment issues during updates when switching paradigms as one stack will try to create it before the other stack deletes it. Say someone wants to move from a MA deployed target to their own or vice versa.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
```diff
@@ -99,6 +106,17 @@ def _get_basic_auth_password(self) -> str:
         password = client.get_secret_value(SecretId=self.auth_details["password_from_secret_arn"])
         return password["SecretString"]
 
+    def _get_sigv4_details(self, force_region=False) -> tuple[str, str]:
+        """Return the service signing name and region name. If force_region is true,
+        it will instantiate a boto3 session to guarantee that the region is not None.
+        This will fail if AWS credentials are not available.
+        """
+        assert self.auth_type == AuthMethod.SIGV4
+        if force_region:
```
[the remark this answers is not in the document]
mikaylathompson: This is in preparation for the `CreateSnapshot` & `MigrateMetadata` commands, where region is a required parameter, so we'll want to get it no matter what.

For the cluster-specific calls, it's okay if we skip it because the auth object will figure it out itself.

deployment/cdk/opensearch-service-migration/lib/service-stacks/prexisting-clusters-stack.ts
mikaylathompson: Ooh, okay, that makes sense and is a good point. Assuming the SSM params have to be created from within a stack though, how would you recommend doing it?

deployment/cdk/opensearch-service-migration/lib/opensearch-domain-stack.ts
mikaylathompson: Well, if we're using this stack, we're creating the domain so we know it's `es`. Do you have any cases in mind where it might be something else?

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
mikaylathompson: Okay, yeah, I like that idea. I'll rework this to support `targetClusterEndpoint` and `targetClusterAuthConfig`

deployment/cdk/opensearch-service-migration/lib/service-stacks/prexisting-clusters-stack.ts
AndreKurait: I'd recommend adding it in `opensearch-domain-stack`/ network-stack like we do currently

mikaylathompson: So `opensearch-domain-stack` would be used whether we're creating a domain or using an existing one? My concern is basically that it'll effectively create two stacks with much more complex logic inside, even if technically they're combined.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/cluster.py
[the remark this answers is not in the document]
AndreKurait: What do you think about using `boto3.session.Session().region_name` to resolve it up front?

PR #868, opened 2024-08-06
CDK fixes for thousandeyes

mikaylathompson:
### Description
1. MSK is only deployed if we need it: any of the capture proxies OR the replayer.
2. Guarantee that the OS_USER_AND_SECRET parameter is created by sticking it in the network stack.

### Issues Resolved
Bugs found while deploying for the Thousand Eyes project.

### Testing
Unit tests are in place & pass, but I'm doing a full manual deploy & run through now.

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/default-values.json
+++ b/deployment/cdk/opensearch-service-migration/default-values.json
@@ -12,7 +12,7 @@
   "migrationAssistanceEnabled": true,
   "replayerOutputEFSRemovalPolicy": "DESTROY",
   "migrationConsoleServiceEnabled": true,
-  "trafficReplayerServiceEnabled": true,
+  "trafficReplayerServiceEnabled": false,
   "otelCollectorEnabled": true,
   "dpPipelineTemplatePath": "./dp_pipeline_template.yaml"
 }
--- a/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
@@ -24,6 +24,8 @@ export interface NetworkStackProps extends StackPropsExt {
     readonly migrationAPIEnabled?: boolean;
     readonly sourceClusterEndpoint?: string;
     readonly targetClusterEndpoint?: string;
+    readonly targetClusterUsername?: string;
+    readonly targetClusterPasswordSecretArn?: string;
     readonly albAcmCertArn?: string;
     readonly env?: { [key: string]: any };
 }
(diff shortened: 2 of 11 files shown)


mikaylathompson:
Deploys successfully, usage tests early in the am, need to fix the cdk tests

mikaylathompson:
(and total deployment time less than 4 minutes!)

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -23,6 +23,7 @@ import { OtelCollectorSidecar } from "./migration-otel-collector-sidecar";
 export interface MigrationConsoleProps extends StackPropsExt {
     readonly migrationsSolutionVersion: string,
     readonly vpc: IVpc,
+    readonly mskEnabled?: boolean,
     readonly streamingSourceType: StreamingSourceType,
```
AndreKurait, line 27: Let's use StreamingSourceType.DISABLED in place of mskEnabled==false in the stack here. We should start to think about BYO Kafka and that enum to fit naturally for this purpose

deployment/cdk/opensearch-service-migration/default-values.json
```diff
@@ -12,7 +12,7 @@
   "migrationAssistanceEnabled": true,
   "replayerOutputEFSRemovalPolicy": "DESTROY",
   "migrationConsoleServiceEnabled": true,
-  "trafficReplayerServiceEnabled": true,
+  "trafficReplayerServiceEnabled": false,
```
AndreKurait, line 15: With this change we need to add it as true in test/defaultMigrationContext.json and in the cdk demo deploy

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
```diff
@@ -257,6 +259,17 @@ export class NetworkStack extends Stack {
                     defaultDeployId: deployId,
                     parameter: MigrationSSMParameter.OS_CLUSTER_ENDPOINT
                 });
+                // This is a somewhat surprsing place for this non-network related set of parameters, but it pairs well with
+                // the OS_CLUSTER_ENDPOINT parameter and is helpful to ensure it happens. This probably isn't a long-term place
+                // for it, but is helpful for the time being.
+                if (props.targetClusterUsername && props.targetClusterPasswordSecretArn) {
+                    createMigrationStringParameter(this,
```
AndreKurait, line 266: This is currently created in opensearch-domain stack and container stack, if including here it needs to be removed from the other two otherwise there will be a deployment conflict from multiple stacks creating the same ssm parameter

mikaylathompson: It's only created here if `targetEndpoint`, which is the condition under which opensearch-domain is not created at all, so I don't think there's a conflict between these two.

Good catch on the os-container one--I've adjusted that to only create the SSM param in the case of demoAdmin, which is when this one won't be used.

This is a bit cobbled together and I think very due for an overhaul/simplification when we start passing in target cluster details more systematically.

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
mikaylathompson: oh, wow, that was already there and I totally missed it. Updated.

AndreKurait (approved):

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -153,10 +153,11 @@ export class MigrationConsoleStack extends MigrationServiceCore {
...
         });
-        const brokerEndpoints = getMigrationStringParameterValue(this, {
-            ...props,
-            parameter: MigrationSSMParameter.KAFKA_BROKERS,
-        });
+        const brokerEndpoints = props.streamingSourceType != StreamingSourceType.DISABLED ?
+            getMigrationStringParameterValue(this, {
+                ...props,
+                parameter: MigrationSSMParameter.KAFKA_BROKERS,
+            }) : "";
```
chelma, line 160: @mikaylathompson Can you help me understand why we we're performing cross-stack reference sharing via Parameter Store rather than passing object references we have in-memory at synth-time?

PR #871, opened 2024-08-06
Fix permissions for github release

AndreKurait:
### Description
Fixes github actions for release, the release workflow needed issue permission.
Fixes AndreKurait maintainer (the approval workflow is case sensitive) 

* Category: Bug Fix
* Why these changes are required? Fix release workflow
* What is the old behavior before changes and new behavior after changes? Release workflow wouldn't accept andre as approver, it also failed on permission creating issues.

### Issues Resolved
[MIGRATIONS-1570](https://opensearch.atlassian.net/browse/MIGRATIONS-1570)

### Testing
Tested in github actions by triggering workflow, was not able to test anything with assuming the role and after since it requires to be cut on the tag.

### Check List

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -12,6 +12,7 @@ env:
 permissions:
   id-token: write
   contents: read
+  issues: write
 
 jobs:
   draft-a-release:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -12,6 +12,7 @@ env:
 permissions:
   id-token: write
   contents: read
+  issues: write
```
mikaylathompson, line 15: Is write a superset of read & write? Presumably we need to do both, right?

mikaylathompson (approved):

.github/workflows/release-drafter.yml
AndreKurait: Write is what they actually have in their readme to use

PR #880, opened 2024-08-08
Add two new command line flags for RfsMigrateDocuments 

gregschohn:
New args are --documents-per-bulk-request and --max-connections.

Now users can specify:
* how many documents should be put into each bulk request to the target server.
* the maxConnection value to the ConnectionProvider that the reactor HttpClients will use.

### Testing
gradle testing

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/RestClient.java
+++ b/RFS/src/main/java/com/rfs/common/RestClient.java
@@ -26,13 +26,16 @@ import io.netty.handler.codec.http.HttpMethod;
 import io.netty.handler.ssl.SslContext;
 import io.netty.handler.ssl.SslContextBuilder;
 import io.netty.handler.ssl.util.InsecureTrustManagerFactory;
+import lombok.extern.slf4j.Slf4j;
 import reactor.core.publisher.Mono;
 import reactor.netty.Connection;
 import reactor.netty.http.client.HttpClient;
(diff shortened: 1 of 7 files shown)


RFS/src/main/java/com/rfs/common/RestClient.java
```diff
@@ -48,7 +51,18 @@ public class RestClient {
...
-        this(connectionContext, HttpClient.create());
+        this(connectionContext, 0);
+    }
+
+    /**
+     * @param maxConnections If &gt; 0, an HttpClient will be created with a provider
+     *                       that uses this value for maxConnections.  Otherwise, a client
+     *                       will be created with default values provided by Reactor.
+     */
+    public RestClient(ConnectionContext connectionContext, int maxConnections) {
```
peternied, line 62: Seems like the connection context should include the max connection

gregschohn: One is a pool configuration and the other is necessary for a single client.

peternied: I'm don't understand the distinction, what is the case for allowing them to be set separately?  Is this an issue with rfs use of connection context vs how the proxy/replayer use it?

peternied (approved):
Thanks for getting this out, should let us experiment with different settings for troubleshooting connection issues.

PR #882, opened 2024-08-08
Add OpenSearchClientTest to verify creation and retry behavior

peternied:
### Description
Add OpenSearchClientTest to verify creation and retry behavior

### Issues Resolved
- Follow up to https://github.com/opensearch-project/opensearch-migrations/pull/876

### Check List
- [X] New functionality includes testing
  - [X] All tests pass, including unit test, integration test and doctest
- [ ] ~New functionality has been documented~
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #886, opened 2024-08-09
Properly implement max connections through flux concurrency in RFS

AndreKurait:
### Description
Properly implement max connections through flux concurrency in RFS with flatmap(_, concurrency)
> @param concurrency the maximum number of in-flight inner sequences

Current implementation with client connection provider resulted in exceptions being thrown when the flux requested more than were available.

Also, fixed a bug where the backoff logic appeared to occur in two places.

Set's the max concurrency here to 50 by default (previously was reactor.util.concurrent.Queues.SMALL_BUFFER_SIZE which is default to 256

* Category: Bug Fix
* Why these changes are required? Correct backoff and concurrency, will allow for greater horizontal scaling
* What is the old behavior before changes and new behavior after changes? Concurrency was at 256 and threw errors if trying to set lower than that.

Note: @gregschohn will take care of unit tests for this stuff today

### Issues Resolved
Issues with ThousandEyes

### Testing

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -32,13 +32,11 @@ public class DocumentReindexer {
             .buffer(numDocsPerBulkRequest) // Collect until you hit the batch size
             .doOnNext(bulk -> logger.info("{} documents in current bulk request", bulk.size()))
             .map(this::convertToBulkRequestBody)  // Assemble the bulk request body from the parts
-            // Retry on failures accumulating requests
-            .retryWhen(Retry.backoff(3, Duration.ofSeconds(1)).maxBackoff(Duration.ofSeconds(5)))
             .flatMap(
                 bulkJson -> client.sendBulkRequest(indexName, bulkJson, context.createBulkRequest()) // Send the request
                     .doOnSuccess(unused -> logger.debug("Batch succeeded"))
                     .doOnError(error -> logger.error("Batch failed", error))
-                    // Prevent the error from stopping the entire stream
+                    // Prevent the error from stopping the entire stream, retries occurring within sendBulkRequest
                     .onErrorResume(e -> Mono.empty()),
                 maxConcurrentRequests)
             .doOnComplete(() -> logger.debug("All batches processed"))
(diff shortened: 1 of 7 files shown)


RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -31,13 +32,15 @@ public Mono<Void> reindex(
             .buffer(numDocsPerBulkRequest) // Collect until you hit the batch size
             .doOnNext(bulk -> logger.info("{} documents in current bulk request", bulk.size()))
             .map(this::convertToBulkRequestBody)  // Assemble the bulk request body from the parts
+            // Retry on failures accumulating requests
+            .retryWhen(Retry.backoff(3, Duration.ofSeconds(1)).maxBackoff(Duration.ofSeconds(5)))
```
gregschohn: This seems like it needs to be AFTER the sendBulkRequest call, where it was.  Why did you rotate it upward?

gregschohn (approved):

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
AndreKurait: Discussed offline, the actual retry mechanism is inside sendBulkRequest. I've since removed this

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -35,9 +34,9 @@ public Mono<Void> reindex(
                 bulkJson -> client.sendBulkRequest(indexName, bulkJson, context.createBulkRequest()) // Send the request
                     .doOnSuccess(unused -> logger.debug("Batch succeeded"))
                     .doOnError(error -> logger.error("Batch failed", error))
-                    .onErrorResume(e -> Mono.empty()) // Prevent the error from stopping the entire stream
-            )
-            .retryWhen(Retry.backoff(3, Duration.ofSeconds(1)).maxBackoff(Duration.ofSeconds(5)))
+                    // Prevent the error from stopping the entire stream, retries occurring within sendBulkRequest
```
peternied, line 37: @AndreKurait Lets add a test to confirm the expected behavior, recently I created a PR that verified a similar scenario [1].

- [1] https://github.com/opensearch-project/opensearch-migrations/pull/882

peternied: I'd recommend testing this because we previously had a bug around this logic that required this change.

AndreKurait: > Note: @gregschohn will take care of unit tests for this stuff today

PR #890, opened 2024-08-10
[Console] Add more flexibility around which clusters are defined

mikaylathompson:
### Description
We're currently working on a scenario in which a target cluster is defined, but the source cluster is not and all work proceeds off of a snapshot.
The console lib _tolerates_ this, but it doesn't handle it particularly well. This PR allows it to specifically handle undefined source clusters.

It's also plausible for user to want to use the create snapshot tooling without having a target cluster defined, so I'm supporting that scenario as well, but giving a warning log message to make it clear that functionality may be impaired without a target cluster (e.g. metadata migration, backfill status checks, etc.).

Example output with no source cluster defined:
```
(.venv) bash-5.2# console clusters connection-check
SOURCE CLUSTER
No source cluster defined.
TARGET CLUSTER
ConnectionResult(connection_message='Successfully connected!', connection_established=True, cluster_version='2.13.0')
```

### Issues Resolved
N/a, testing for Thousand Eyes scenario.

### Testing
Manually, via deployment to AWS.

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
@@ -50,10 +50,8 @@ def cli(ctx, config_file, json, verbose):
 @cli.group(name="clusters", help="Commands to interact with source and target clusters")
 @click.pass_obj
 def cluster_group(ctx):
-    if ctx.env.source_cluster is None:
-        raise click.UsageError("Source cluster is not set")
-    if ctx.env.target_cluster is None:
-        raise click.UsageError("Target cluster is not set")
+    if ctx.env.source_cluster is None and ctx.env.target_cluster is None:
+        raise click.UsageError("Neither source nor target cluster is defined.")
 
 
 @cluster_group.command(name="cat-indices")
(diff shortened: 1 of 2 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
```diff
@@ -50,10 +50,8 @@ def cli(ctx, config_file, json, verbose):
 @cli.group(name="clusters", help="Commands to interact with source and target clusters")
 @click.pass_obj
 def cluster_group(ctx):
-    if ctx.env.source_cluster is None:
-        raise click.UsageError("Source cluster is not set")
-    if ctx.env.target_cluster is None:
-        raise click.UsageError("Target cluster is not set")
+    if ctx.env.source_cluster is None and ctx.env.target_cluster is None:
+        raise click.UsageError("Neither source nor target cluster is defined.")
```
chelma, line 54: Non-blocking grammar nit: `is` => `are`

chelma (approved):
Seems reasonable - thank, Mikayla!

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/cli.py
mikaylathompson: equally nitpicky, but `is` is technically right because both are singular nouns (but both versions are common) (https://www.cjr.org/language_corner/either_wins.php)

PR #892, opened 2024-08-12
Dead Letter Queue for failed bulk requests

peternied:
### Description
Adding the logic to capture failed bulk requests logic into a dead letter queue that can be inspecting after RFS completes.  Infrastructure changes for this are still in flight.

Notably I've rewritten the approach of bulk requests so that as they are retried any documents that have already successfully been uploaded are removed from the bulk request, this should increase the chances for success on throttled systems.  This also means that after the final retry has failed the full response is likely small enough to put in the DLQ.

To parse the response from the BulkRequest I've implemented an iterative json parser, such be able to handle sufficiently large bulk responses compared to marshalling the whole data structure into a json object.

### Issues Resolved
- https://opensearch.atlassian.net/browse/MIGRATIONS-1939

### Testing
Unit tests + component tests.

### Check List


RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -89,6 +98,12 @@ public Optional<ObjectNode> createIndex(
         return createObjectIdempotent(targetPath, settings, context);
     }
 
+    Retry checkIfItemExistsRetryStrategy = Retry.backoff(3, Duration.ofSeconds(1))
```
AndreKurait, line 101: should these be private static final?

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -257,20 +270,51 @@ public Optional<ObjectNode> getSnapshotStatus(
...
     }
 
-    public Mono<BulkResponse> sendBulkRequest(String indexName, String body, IRfsContexts.IRequestContext context) {
+    Retry getBulkRetryStrategy() {
+        return Retry.backoff(10, Duration.ofSeconds(2)).maxBackoff(Duration.ofSeconds(60));
+    }
+
+    public Mono<BulkResponse> sendBulkRequest(String indexName, List<BulkDocSection> docs, IRfsContexts.IRequestContext context) {
         String targetPath = indexName + "/_bulk";
+        Map<String, BulkDocSection> docsMap = docs.stream().collect(Collectors.toMap(d -> d.getDocId(), d -> d));
```
AndreKurait, line 279: Can we defer constructing this docsMap to be inside the mono.defer? Possibly through an AtomicReference?

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -257,20 +270,51 @@ public Optional<ObjectNode> getSnapshotStatus(
         }
     }
 
-    public Mono<BulkResponse> sendBulkRequest(String indexName, String body, IRfsContexts.IRequestContext context) {
+    Retry getBulkRetryStrategy() {
+        return Retry.backoff(10, Duration.ofSeconds(2)).maxBackoff(Duration.ofSeconds(60));
```
AndreKurait, line 274: Can we add a comment here why you chose 10? I'm thinking based on seeing the 10 minute cluster write locks?
Though in this case, since maxBackoff is 60s, the total delay won't actually be 10 minutes.

2 + 4 + 8 + 16 + 32 + 60 + 60 + 60 + 60 + 60 = 6 min 2 sec

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -257,20 +270,51 @@ public Optional<ObjectNode> getSnapshotStatus(
...
+                    var successfulDocs = resp.getSuccessfulDocs();
+                    successfulDocs.forEach(docsMap::remove);
+                    log.info(
+                        "After bulk request on index '{}', {} more documents have succeed, {} remain",
+                        indexName,
+                        successfulDocs.size(),
+                        docsMap.size()
+                    );
+
+                    log.error(resp.getFailureMessage());
```
AndreKurait, line 303: Let's include the failure message here in the same log as above. Say my log level was error, it would be really strange to see this message and not the above log given it's not actually an issue

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -298,6 +342,15 @@ public boolean hasFailedOperations() {
             return matcher.find();
         }
 
+        public List<String> getSuccessfulDocs() {
+            try {
+                return BulkResponseParser.findSuccessDocs(body);
+            } catch (IOException ioe) {
+                log.warn("Unable to process bulk request for success", ioe);
```
AndreKurait, line 349: Should this be at error? It's going to mean the failedRequestsLogger has extra docs in it

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -257,20 +270,51 @@ public Optional<ObjectNode> getSnapshotStatus(
...
+            return client.postAsync(targetPath, body, context)
+                .flatMap(response -> {
+                    var resp = new BulkResponse(response.statusCode, response.statusText, response.headers, response.body);
+                    if (!resp.hasBadStatusCode() && !resp.hasFailedOperations()) {
+                        return Mono.just(resp);
+                    }
+                    // Remove all successful documents for the next bulk request attempt
+                    var successfulDocs = resp.getSuccessfulDocs();
+                    successfulDocs.forEach(docsMap::remove);
+                    log.info(
```
AndreKurait, line 296: This log may make more sense in a doAfterRetry so that we can include context on the current retry count

```
            .retryWhen(Retry.backoff(6, Duration.ofSeconds(2))
                .maxBackoff(Duration.ofSeconds(60))
                .doAfterRetry(retrySignal -> 
                    logger.error("Retrying bulk request after error: {}. Attempt: {}/{}",
                        retrySignal.failure().getMessage(),
                        retrySignal.totalRetriesInARow() + 1,
                        retrySignal.totalRetries())));
```

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -257,20 +270,51 @@ public Optional<ObjectNode> getSnapshotStatus(
...
-                }
-                return Mono.just(resp);
-            })
-            // In throttle cases, this will be low enough to get down to 1tps with 50 concurrency
-            .retryWhen(Retry.backoff(6, Duration.ofSeconds(2)).maxBackoff(Duration.ofSeconds(60)));
+                });
+        })
+        .retryWhen(getBulkRetryStrategy())
+        .doOnError(error -> {
+            if (!docsMap.isEmpty()) {
```
AndreKurait, line 309: docsMap should always have items here. Can we do an assert on that or at least have an error log on the else case?

RFS/src/main/java/org/opensearch/migrations/reindexer/FailedRequestsLogger.java
```diff
@@ -0,0 +1,58 @@
...
+        var responseBody = Optional.ofNullable(rootCause)
+            .filter(OperationFailed.class::isInstance)
+            .map(OperationFailed.class::cast)
+            .map(opFailed -> opFailed.response)
+            .map(response -> response.body);
+
+        if (responseBody.isPresent()) {
+            logger.atInfo()
+                .setMessage(
+                    "Bulk request failed for {} index on {} items, reason {}, bulk request body followed by response:\n{}\n{}"
```
AndreKurait, line 32: nit: Do we have the shard here? Would be helpful to know which shard to retry

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -106,6 +107,7 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
                 "RFS_TARGET_USER": targetUser,
                 "RFS_TARGET_PASSWORD": targetPassword,
                 "RFS_TARGET_PASSWORD_ARN": targetPasswordArn,
+                "TUPLE_DIR_PATH": `/shared-replayer-output/traffic-replayer-${deployId}`
```
AndreKurait, line 110: do you mean `traffic-replayer` here?

RFS/src/main/resources/log4j2.properties
```diff
@@ -0,0 +1,31 @@
...
+
+appender.console.type = Console
+appender.console.name = Console
+appender.console.target = SYSTEM_OUT
+appender.console.layout.type = PatternLayout
+appender.console.layout.pattern = %d{yyyy-MM-dd HH:mm:ss,SSS}{UTC} %p %c{1.} [%t] %m%n
+
+appender.FailedRequests.type = RollingRandomAccessFile
+appender.FailedRequests.name = FailedRequests
+appender.FailedRequests.fileName = ${logsDir}/failedRequests.log
```
AndreKurait, line 15: This looks like it's going to have issues with multiple RFS workers writing to the same path and files right?

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
peternied: Upped this to 15 , where it will handle +10 minutes of an issue.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
peternied: Interesting feature, but I'd rather keep the logic as is to allow for the retry policy to be subbed out in the test cases.

RFS/src/main/java/org/opensearch/migrations/reindexer/FailedRequestsLogger.java
peternied: Nope, it seems strange to pipe that through as is - we can revisit this as we iterate on logging

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
peternied: I'd rather place all the logs in once place even if its a little strange.  We can circle back on this and update the documentation after we have a cohesive strategy.

RFS/src/main/resources/log4j2.properties
peternied: Ugh - handling this scenario is going to require more work, its a little convoluted to extract the container identifier.  I'll create a follow up PR that handles the deployment changes.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -257,20 +280,54 @@ public Optional<ObjectNode> getSnapshotStatus(
...
+                    // Remove all successful documents for the next bulk request attempt
+                    var successfulDocs = resp.getSuccessfulDocs();
+                    successfulDocs.forEach(docsMap::remove);
+                    log.atWarn()
+                        .setMessage("After bulk request on index '{}', {} more documents have succeed, {} remain")
+                        .addArgument(indexName)
+                        .addArgument(successfulDocs::size)
+                        .addArgument(docsMap::size)
+                        .log();
                     return Mono.error(new OperationFailed(resp.getFailureMessage(), resp));
```
chelma, line 311: I'm curious why we're removing successful docs rather than making new request out of the ones that failed.  Is this a bet that, on the whole, there will be more failed docs than successful ones when there are any failed docs?  Also, and I could definitely be wrong here, it feels more efficient to construct a new array from a set of elements of known size rather than iteratively remove elements from an existing array.  Both aspects lead me to intuit that we might be better off taking the alternative approach.

What are your thoughts?

RFS/src/main/java/org/opensearch/migrations/parsing/BulkResponseParser.java
```diff
@@ -0,0 +1,120 @@
...
+import com.fasterxml.jackson.core.JsonToken;
+
+import lombok.Builder;
+import lombok.Getter;
+import lombok.experimental.UtilityClass;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+@UtilityClass
+public class BulkResponseParser {
```
chelma, line 18: Is there a speed or efficiency gain by parsing the JSON like this at a lower level rather than marshalling the response into objects?  If there isn't a big difference, it feels like it would be simpler to just do the marshalling and scan through that way.  We're taking a dependency on the version-specific structure of the response body either way.

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
peternied: In the happy path all docs succeed and we don't get to the winnowing code.

For the 'some docs failed case' Its easier to determine that a doc has been successfully uploaded - whereas if we do not get a full list of documents in the bulk response we need work backwards from the list of doc as we scan and make sure it isn't absent.  I'm optimizing this code towards reliability.

chelma: I'd love to take a look at some examples with you because it sounds like we understand the response format differently.  From my reading of the spec [1], we end up with an entry for every element of the request and it appears easy to check if any given element failed.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/7.10/docs-bulk.html

chelma: Per conversation - we don't have evidence to indicate how many failed requests will show up in a response when some do (e.g. is it usually a lot or very few?).  Additionally, the `docsMap` doesn't contain the bulk sections itself, it just contains references to them.  As a result, while I feel intuitively that we'd be better off making a new map using the deleted docs, the impact appears uncertain enough to keep the current approach and see how it works out.

RFS/src/main/java/org/opensearch/migrations/parsing/BulkResponseParser.java
peternied: No matter the speed difference, if we get a truncated response for whatever reason this parser will still be able to read the json document until it hits the end of what is available.  Added tests that prove this.

Additionally, this parser should be able to handle streaming input with minimal changes if/when that arrives.

Finally its ~19% faster than using readObject, I'll create a separate PR with the details.

```
Benchmark                                                                      Mode  Cnt   Score   Error  Units
JacksonParserVsReadObjectBenchmark.testJacksonMarshalling                     thrpt    2  56.948          ops/s
JacksonParserVsReadObjectBenchmark.testJacksonParser                          thrpt    2  67.714          ops/s
```

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
```diff
@@ -313,6 +320,12 @@ public Mono<BulkResponse> sendBulkRequest(String indexName, List<BulkDocSection>
                     () -> BulkDocSection.convertToBulkRequestBody(docsMap.values()),
                     error
                 );
+            } else {
+                log.atWarn()
```
AndreKurait, line 324: nit: atError given this is unexpected

AndreKurait (approved):

RFS/src/main/java/com/rfs/common/OpenSearchClient.java
peternied: I'll get this on a follow up PR - thanks!

PR #893, opened 2024-08-13
Remove MSKUtilityStack / Use MSK Alpha CDK Construct

lewijacn:
### Description
This change cleans up a lot of questionable pieces around our MSK cluster. Mainly it does the following key things:
1. Swaps out L1 construct with MSK alpha construct. This construct already has support for getting the broker endpoints via its own internal custom resource and no longer requires us to handle this logic. From my handful of attempts this has cleaned up nicely. It also has support for importing an existing MSK cluster, which we don't currently utilize, but with a very small change we can have support for.
2. Removes logic around MSK public endpoints. This support is not something we test for and likely something we don't want to promote as a common use case. The code around this was pretty rough as well, so I've seen this as a good opportunity to clean this up.
3. Removes the MSKUtilityStack and custom resource lambda code, as these are no longer needed with use of this construct
4. Upgrades CDK library version

### Testing
Jest tests and cloud testing

### Check List


deployment/cdk/opensearch-service-migration/options.md
```diff
@@ -95,10 +95,7 @@ These tables list all CDK context configuration values a user can specify for th
...
 | mskAZCount                       | number       | 2                                                                                                       | The number of Availability Zones for the MSK cluster to use. **NOTE** This value must be 2 or 3                                                                                                                                                                                                                                                    |
 | mskARN (Not currently available) | string       | `"arn:aws:kafka:us-east-2:12345678912:cluster/msk-cluster-test/81fbae45-5d25-44bb-aff0-108e71cc079b-7"` | Supply an existing MSK cluster ARN to use. **NOTE** As MSK is using an L1 construct this is not currently available for use                                                                                                                                                                                                                        |
```
peternied, line 101: Seems like we should remove this parameter while we are here.

peternied (approved):
Looks good to me

deployment/cdk/opensearch-service-migration/options.md
lewijacn: I plan to implement this in a very small follow-up PR

PR #894, opened 2024-08-13
Added RFS support for Deletes/Updates on ES 7.10+; _source doc bugfix

chelma:
### Description
* Updated the RFS code to handle deleted and updated docs in ES 7.10+ snapshots.  The code is positioned to handle 6.X snapshots as well, though that has not been tested
* Fixed a bug in how we assemble bulk requests from the _source doc that resulted in invalid requests.  The solution was to "minify" the _source JSON before adding to a bulk request.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1746
* https://opensearch.atlassian.net/browse/MIGRATIONS-1747

### Testing
Tested extensively using Docker Compose locally.  I turned my two test snapshots (one w/ soft deletes, the other without) into `test-resources` and used them in the updated unit tests for the `LuceneDocumentsReader`

### Check List


RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -28,32 +30,41 @@ public Mono<Void> reindex(
...
+
+        return action + "\n" + minifiedSource;
     }
 
     private String convertToBulkRequestBody(List<String> bulkSections) {
         StringBuilder builder = new StringBuilder();
         for (String section : bulkSections) {
             builder.append(section).append("\n");
         }
+        log.atInfo().log("Bulk request body: \n{}", builder.toString());
```
chelma, line 66: Oops - this should be `atDebug()`

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,83 @@
...
         });
     }
 
-    protected IndexReader openIndexReader(Path indexDirectoryPath) throws IOException {
-        return DirectoryReader.open(FSDirectory.open(indexDirectoryPath));
+    protected DirectoryReader wrapReader(DirectoryReader reader, boolean softDeletesEnabled, String softDeletesField) throws IOException {
+        if (softDeletesEnabled) {
+            return new SoftDeletesDirectoryReaderWrapper(reader, softDeletesField);
+        }
+        return reader;
```
chelma, line 68: This needs a comment explaining what it does and how it works; it's pretty magical

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,83 @@
...
+    public static Function<Path, LuceneDocumentsReader> getFactory(boolean softDeletesPossible, String softDeletesField) {
+        return path -> new LuceneDocumentsReader(path, softDeletesPossible, softDeletesField);
+    }
     public static final int NUM_DOCUMENTS_BUFFERED = 1024;
+
     protected final Path indexDirectoryPath;
+    protected final boolean softDeletesPossible;
+    protected final String softDeletesField;
 
     public Flux<Document> readDocuments() {
```
chelma, line 33: Need a higher-level comment explaining the various states the documents can be in, and how they'll be handled.

RFS/test-resources/inventory.md
```diff
@@ -13,3 +13,107 @@ An Elasticsearch 7.10 snapshot repo containing a single snapshot (`global_state_
 #### ES_7_10_Double
 An Elasticsearch 7.10 snapshot repo containing a two snapshots (`global_state_snapshot`, `global_state_snapshot_2`).  `global_state_snapshot` contains two indices (`posts_2023_02_25`, `posts_2024_01_01`), each with few documents in them, an index template, `posts_index_template`, and a composite template, `posts_template`.  `global_state_snapshot_2` is the same, except it doesn't have the index `posts_2023_02_25` in it.
 
+#### ES_7_10_Updates_Deletes_w_Soft
```
peternied, line 16: SnapshotStateTest allow for this kind of configuration, lets update those tests rather than check in artifacts

chelma: I think I'm gonna disagree and commit on this one.  The fewer containers we need to spin up in order to run our tests, the better.  When we use these instead of spinning up a Source cluster it runs in seconds instead of minutes.

RFS/src/test/java/com/rfs/common/LuceneDocumentsReaderTest.java
```diff
@@ -1,90 +1,121 @@
...
+import com.rfs.version_es_7_10.SnapshotRepoProvider_ES_7_10;
+import lombok.extern.slf4j.Slf4j;
 import reactor.core.publisher.Flux;
 import reactor.test.StepVerifier;
 
-import static org.mockito.Mockito.*;
+@Slf4j
+public class LuceneDocumentsReaderTest {
+    private static final ObjectMapper objectMapper = new ObjectMapper();
+    private Path tempDirectory;
```
peternied, line 30: Using the `@TempDir` annotation can remove the setup/teardown code

peternied (commented):
Looks good minor comments, nice find on `SoftDeletesDirectoryReaderWrapper`

peternied (approved):

RFS/test-resources/inventory.md
chelma: I'll make a JIRA for the test-resource generation idea; that makes a lot of sense!

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,116 @@
...
+    protected final boolean softDeletesPossible;
+    protected final String softDeletesField;
+
+    /**
+     * There are a variety of states the documents in our Lucene Index can be in; this method extracts those documents
+     * that would be considered "live" from the ElasticSearch/OpenSearch perspective.
+     *
+     * For context, when ElasticSearch/OpenSearch deletes a document, it doesn't actually remove it from the Lucene Index.
+     * Instead, what happens is that the document is marked as "deleted" in the Lucene Index, but it is still present in the
+     * Lucene segment on disk.  The next time that segment is merged, the deleted documents are removed from the Lucene Index.
```
gregschohn, line 39: Technically, the documents are not propagated into the new segment.  Documents are never removed from a segment.  From a nomenclature perspective (I'm always fuzzy on what's an index and a segment), wouldn't the document have been not-present both before and after merge in the index?  I guess the question is, does the index include the tombstones or not?

Having a link to a high-level explanation of the lucene storage structure could be helpful.

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -216,7 +216,8 @@ public static void main(String[] args) throws Exception {
                 );
 
                 run(
-                    LuceneDocumentsReader::new,
+                    LuceneDocumentsReader.getFactory(ElasticsearchConstants_ES_7_10.SOFT_DELETES_POSSIBLE,
```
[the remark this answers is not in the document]
chelma: Yep! Part of adding support for 6.8 will be having a way to know the source version, either by the user passing it in or us detecting it somehow. At that point we'll be able to construct the version-appropriate classes for dependency injection.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,116 @@
...
+     * added in a new Lucene segment.  This means that from an ES/OS perspective, you could have a single document that has
+     * been created, deleted, recreated, updated, etc. multiple times and only a single version of the doc would exist when
+     * you queried the ES/OS Index - but every single iteration of that doc might still exist in the Lucene Segments on disk,
+     * all of which have the same _id (from the ES/OS perspective).
+     *
+     * Additionally, Elasticsearch 7 introduced a feature called "soft deletes" which allows you to mark a document as
+     * "deleted" in the Lucene Index without actually removing it from the Lucene Index.  This works by having the
+     * application writing the Lucene Index define a field that is used to mark a document as "soft deleted" or not.  When
+     * a document is marked as "soft deleted", it is not returned in search results, but it is still present in the Lucene
+     * Index.  The status of whether any given document is "soft deleted" or not is stored in the Lucene Index itself.  By
```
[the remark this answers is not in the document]
chelma: The documentation here not very helpful [1], but I think the idea is that the soft deletes are only "resolved"/"closed" once the Elasticsearch cluster is confident that all replicas have successfully resolved them.  This provides an additional layer of safety beyond the translog on the primary so that if a replica node loses connectivity to the cluster, it can successfully resolve the deletes that have been performed while it was diconnected.  In that scenario, the primary may have already performed a Lucene Merge and truncated its translog, forcing the replica to pull a full new copy of the shard when it comes back into the cluster instead of being able to just replay the operations it missed.  Soft-deletes appears like an optimization to make that expensive operation less likely to be necessary.

[1] https://www.elastic.co/guide/en/elasticsearch/reference/7.10/index-modules-history-retention.html

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,116 @@
...
+     * default, Elasticsearch 7+ Indices have soft deletes enabled; this is an Index-level setting.
```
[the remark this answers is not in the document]
chelma: When we're reindexing from a snapshot, we don't need to send soft-deleted docs to the target cluster.  From the perspective of the shard primary on the source cluster, those docs have been deleted successfully.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,116 @@
...
+     *
+     * In order to retrieve only those documents that would be considered "live" in ES/OS, we use a few tricks:
+     * 1. We make sure we use the latest Lucene commit point on the Lucene Index.  A commit is a Lucene abstraction that
```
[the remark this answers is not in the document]
chelma: AFAIK, a Lucene Index can contain more than one commit point.  In the case of incremental snapshots, the fact they are incremental is completely invisible at the Lucene level and is an Elasticsearch abstraction.  According to my current understanding, restoring an incremental snapshot (or, pulling it to disk for RFS) results in a fully complete Lucene Index that is current to the moment the incremental snapshot was created.  I don't think there's a scenario where we'd ever want to use something other than the most recent commit point.

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -17,58 +21,116 @@
...
     public Flux<Document> readDocuments() {
-        return Flux.using(() -> openIndexReader(indexDirectoryPath), reader -> {
-            log.info(reader.maxDoc() + " documents found in the current Lucene index");
+        return Flux.using(() -> wrapReader(DirectoryReader.open(FSDirectory.open(indexDirectoryPath)), softDeletesPossible, softDeletesField), reader -> {
+            log.atInfo().log(reader.maxDoc() + " documents found in the current Lucene index");
+
+            return Flux.fromIterable(reader.leaves()) // Iterate over each segment
+                .flatMap(leaf -> {
+                    LeafReader leafReader = leaf.reader();
+                    Bits liveDocs = leafReader.getLiveDocs(); // Get live docs bits
```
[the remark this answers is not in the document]
chelma: Good question, I'm not sure actually.  I know there's a per-segment LiveDocs file [1].  From the API spec, it appears that this file would contain bit sets for each commit point on the Lucene Index:

```
void  files(SegmentCommitInfo info, Collection<String> files)
Records all files in use by this SegmentCommitInfo into the files argument.

Bits  readLiveDocs(Directory dir, SegmentCommitInfo info, IOContext context)
Read live docs bits.

void  writeLiveDocs(Bits bits, Directory dir, SegmentCommitInfo info, int newDelCount, IOContext context)
Persist live docs bits.
```

I'm guessing that part of the commit process is updating the bitset for each segment.


[1] https://lucene.apache.org/core/8_0_0/core/org/apache/lucene/codecs/lucene50/Lucene50LiveDocsFormat.html

RFS/src/main/java/com/rfs/version_es_6_8/ElasticsearchConstants_ES_6_8.java
```diff
@@ -22,6 +24,10 @@ public class ElasticsearchConstants_ES_6_8 {
         smileFactory.configure(JsonGenerator.Feature.AUTO_CLOSE_JSON_CONTENT, false);
         smileFactory.configure(JsonParser.Feature.STRICT_DUPLICATE_DETECTION, false);
         SMILE_FACTORY = smileFactory;
+
+        // Soft Deletes were added in 7.0
+        SOFT_DELETES_FIELD = "";
```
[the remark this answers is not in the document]
chelma: So, the Lucene API requires the application using it (Elasticsearch, in this case) to define the name of the field which will indicate whether a document is soft deleted or not.  You then supply this name to Lucene on read in order for it to be able to tell the soft-deleted from the not.  That is - this is not a built in, unchanging key on some object in the Lucene API spec, it's user-configurable.  No idea why at the moment.

RFS/test-resources/inventory.md
```diff
@@ -13,3 +13,107 @@ An Elasticsearch 7.10 snapshot repo containing a single snapshot (`global_state_
...
+    "bucket": "chelma-iad-rfs-local-testing",
+    "region": "us-east-1"
+  }
+}'
+
+curl -X PUT "localhost:19200/_snapshot/test_s3_repository/rfs_snapshot" -H "Content-Type: application/json" -d '{
+  "indices": "test_updates_deletes",
+  "ignore_unavailable": true,
+  "include_global_state": true
+}'
```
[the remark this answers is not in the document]
chelma: Our Docker Compose container currently has the S3 plugin installed, so there wasn't any work required on that front.  I agree that when we move to a better system for generating these snapshots we should use local disk.

PR #900, opened 2024-08-14
Add sourceClusterDisabled option to CDK

mikaylathompson:
### Description
This change adds a `sourceClusterDisabled` option to the CDK that gives the user an opportunity to explicitly opt out of having a source cluster or any services dependent on it.

If this option is provided along with any source-cluster related services, it will throw an error.
If none of the existing ways to define a source cluster are provided, and this option is left out, it will also throw an error.

While slightly redundant, the goal is to make sure that we're not letting users accidentally skip having a source cluster.

Note: this option is unusual in that it's defined as the negative (`disabled` instead of `enabled`). The alternate formulation of this is to define it as `sourceClusterEnabled` and set the default value to `true`, so the user will explicitly have to set `sourceClusterEnabled: false` in their cdk.context.json.  I'm ambivalent here, if reviewers feel strongly, I'm happy to flip it.

Because CDK is CDK and our CDK is our CDK, there are a few annoying workaround type spots to deal with parameters.

### Issues Resolved
[MIGRATIONS-1942](https://opensearch.atlassian.net/browse/MIGRATIONS-1942)

### Testing
Unit test added & manual testing.

### Check List


deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -19,6 +19,7 @@ import {Fn, RemovalPolicy} from "aws-cdk-lib";
 import {MetadataMigrationYaml, ServicesYaml} from "../migration-services-yaml";
 import {ELBTargetGroup, MigrationServiceCore} from "./migration-service-core";
 import { OtelCollectorSidecar } from "./migration-otel-collector-sidecar";
+import { assert } from "console";
```
lewijacn, line 22: Can we remove this unused import

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
```diff
@@ -229,8 +230,11 @@ export class NetworkStack extends Stack {
         }
 
         // Create Source SSM Parameter
-        if (props.sourceClusterEndpoint) {
-            createMigrationStringParameter(this, props.sourceClusterEndpoint, {
+        if (props.sourceClusterEndpoint || props.sourceClusterDisabled) {
+            // The migration console stack depends on this parameter, even if the source cluster is disabled, so
```
lewijacn, line 234: Why does the Migration Console stack depend on this, aren't we already controlling getting the SSM parameter with the sourceClusterDisabled paramter we are passing to the Migration Console stack?

lewijacn: Apart from this though it looks like the RFS and FetchMigration stacks use the SOURCE_CLUSTER_ENDPOINT SSM parameter. RFS looks like it actually doesn't use it and can be removed. And we should probably just include Fetch in the conditional to throw an error if sourceClusterDisabled is provided.

mikaylathompson: > Why does the Migration Console stack depend on this, aren't we already controlling getting the SSM parameter with the sourceClusterDisabled paramter we are passing to the Migration Console stack?


Yes, butttt CDK is only creating cloudformation and the cloudformation template includes it in the "parameter" section regardless of any conditionality around whether it's used. So as far as I can tell, we need to create this parameter if there are any stacks that _might_ depend on it, even if they don't actually.

mikaylathompson: ^ update, I was fortunately wrong about this!

lewijacn (approved):
LGTM!

PR #901, opened 2024-08-14
Add Benchmarks for Jackson Parser vs ReadObject

peternied:
### Description
Adds new benchmarks for testing BulkResponseParser speed vs using object marshalling.  Also includes documentation on how to run and author tests.  Using [Java Microbenchmark Harness (JHM)](https://github.com/openjdk/jmh) for benchmarks.

Example output in `${project.dir}/build/reports/jmh/results.txt`:

```
Benchmark                                                   Mode  Cnt    Score   Error  Units
JacksonParserVsReadObjectBenchmark.testJacksonMarshalling  thrpt    2  438.434          ops/s
JacksonParserVsReadObjectBenchmark.testJacksonParser       thrpt    2  709.433          ops/s

```

### Issues Resolved
- Related https://github.com/opensearch-project/opensearch-migrations/pull/892
- Resolves https://opensearch.atlassian.net/browse/MIGRATIONS-1947

### Testing
Ran locally, tests are also executed via CI

### Check List


RFS/src/test/java/org/opensearch/migrations/parsing/BulkResponseParserTest.java
```diff
@@ -39,7 +39,7 @@ void testFindSuccessDocs_truncatedResponse() throws IOException {
         var successDocId = "223";
```
peternied, line 39: Looks like I broke this test case when I changed the generated output slightly, need to fix this...

RFS/build.gradle
```diff
@@ -84,6 +85,9 @@ dependencies {
     testFixturesImplementation group: 'org.testcontainers', name: 'testcontainers'
 
     testFixturesImplementation group: 'org.hamcrest', name: 'hamcrest'
+
+    testImplementation 'org.openjdk.jmh:jmh-core:1.36'
+    testImplementation 'org.openjdk.jmh:jmh-generator-annprocess:1.36'
```
chelma, line 90: Is there a reason not to promote these package versions to the top level gradle dependency definition, like the rest of our dependencies?

RFS/src/test/java/org/opensearch/migrations/parsing/JacksonParserVsReadObjectBenchmark.java
```diff
@@ -0,0 +1,117 @@
...
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.Warmup;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+
+public class JacksonParserVsReadObjectBenchmark {
```
chelma, line 27: Thanks for testing this possible approach!  Since it appears the other approach has a substantial performance benefit and will therefore be part of our mainline code, I'm curious what your thinking is for whether/where to keep this unused code around longer term?  Do you envision a scenario where we'd want to use this class in our mainline code in the future, or are you thinking this will be useful as a reference for how to perform benchmarking?

RFS/build.gradle
peternied: Good call out, I'll move them up there.

RFS/src/test/java/org/opensearch/migrations/parsing/JacksonParserVsReadObjectBenchmark.java
peternied: I'd be curious the teams thoughts on this topic, this seems like a good Friday demo.

Keeping the unused solution allows revisiting this experiment, otherwise it would be lost to time as I have often seen.  I don't think there is value aside from having it available as an artifact.  If it becomes a burden that sounds like a good case to revisit.

I think have a reference cases is useful to allow us to quickly iterate.

chelma: I agree, having an example would be helpful.  What would be even better would be instrumenting our existing code for performance testing and tracking the results commit-by-commit.  That's definitely out of scope of this pro-bono PR, though.  Would you be willing to make a JIRA to track that work?  Once we have some in-use perf testing code we can remove this unused example.

peternied: Created broad issue - https://opensearch.atlassian.net/browse/MIGRATIONS-1948

chelma (approved):

PR #905, opened 2024-08-15
Throughput improvements

AndreKurait:
### Description
Currently, reindexing documents had inefficient processing and single thread operations causing a slowdown in max throughput (from 480 -> 170 10MB requests/min with AWS Fargate 2vcpu) from a few PRs ago.

This change alleviates these bottlenecks with the following:
* Refactored the LuceneDocumentReader to process segments and docs in parallel and set aside I/O bound reading on a dedicated elastic threadpool
* Modified source doc "minimizing" in DocumentReindexer to use string replace instead of ObjectMapper
* Modified BulkDocSection to optimistically construct and persist index strings
* Split up dedicated threadpools in DocumentReindexer based on I/O vs CPU tasks.

* Category: Enhancement
* Why these changes are required? Throughput improvements to reduce cost and risk with long running shard migrations. 
* What is the old behavior before changes and new behavior after changes? With Fargate 2cpu 4gb tasks, went from ~170 requests/min -> 580 10MB requests (14k docs)

### Issues Resolved

### Testing
Unit testing and cloud testing

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
+++ b/RFS/src/main/java/com/rfs/common/DocumentReindexer.java
@@ -15,11 +15,12 @@ import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Scheduler;
+import reactor.core.scheduler.Schedulers;
 
 @Slf4j
 @RequiredArgsConstructor
 public class DocumentReindexer {
-    private static final ObjectMapper objectMapper = new ObjectMapper();
     protected final OpenSearchClient client;
     private final int maxDocsPerBulkRequest;
     private final long maxBytesPerBulkRequest;
@@ -30,8 +31,17 @@ public class DocumentReindexer {
         Flux<Document> documentStream,
         IDocumentMigrationContexts.IDocumentReindexContext context
     ) {
+        // Create elastic scheduler for long-lived i/o bound tasks
+        Scheduler elasticScheduler = Schedulers.newBoundedElastic(maxConcurrentWorkItems, Integer.MAX_VALUE, "documentReindexerElastic");
+        // Create parallel scheduler for short-lived CPU bound tasks
+        Scheduler parallelScheduler = Schedulers.newParallel( "documentReindexerParallel");
+
         return documentStream
+            .parallel()
+            .runOn(parallelScheduler)
             .map(BulkDocSection::new)
+            .sequential()
+            .publishOn(parallelScheduler)
             .bufferUntil(new Predicate<>() {
                 private int currentItemCount = 0;
                 private long currentSize = 0;
@@ -58,12 +68,13 @@ public class DocumentReindexer {
                 maxConcurrentWorkItems, // Number of parallel workers, tested in reindex_shouldRespectMaxConcurrentRequests
                 maxConcurrentWorkItems // Limit prefetch for memory pressure
             )
+            .runOn(elasticScheduler, 1) // Use elasticScheduler for I/O bound request sending
             .flatMap(
                 bulkDocs -> client
                     .sendBulkRequest(indexName, bulkDocs, context.createBulkRequest()) // Send the request
                     .doFirst(() -> log.atInfo().log("{} documents in current bulk request.", bulkDocs.size()))
                     .doOnSuccess(unused -> log.atDebug().log("Batch succeeded"))
-                    .doOnError(error -> log.atError().log("Batch failed", error))
+                    .doOnError(error -> log.atError().log("Batch failed {}", error))
                     // Prevent the error from stopping the entire stream, retries occurring within sendBulkRequest
                     .onErrorResume(e -> Mono.empty()),
             false,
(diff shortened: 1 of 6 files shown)


RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -30,7 +30,13 @@ public Mono<Void> reindex(
         Flux<Document> documentStream,
         IDocumentMigrationContexts.IDocumentReindexContext context
     ) {
+        // Create elastic scheduler for long-lived i/o bound tasks
```
peternied, line 33: Could you write a test case that confirms the differences in processing speeds for the flux source?

peternied (commented):
Great to see the perf improvements, lets add some tests to verify the shape of the performance is improved with these changes.

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
```diff
@@ -58,48 +64,49 @@ public boolean test(BulkDocSection next) {
...
                     // Prevent the error from stopping the entire stream, retries occurring within sendBulkRequest
                     .onErrorResume(e -> Mono.empty()),
             false,
                 1, // control concurrency on parallel rails
                 1 // control prefetch across all parallel runners
             )
             .doOnComplete(() -> log.debug("All batches processed"))
-            .then();
+            .then()
+            .publishOn(Schedulers.single()) // replace with Standard scheduler before disposing schedulers
```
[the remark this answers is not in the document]
AndreKurait: restructured to remove this switch

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -64,26 +64,27 @@ public static Function<Path, LuceneDocumentsReader> getFactory(boolean softDelet
 
     */
     public Flux<Document> readDocuments() {
```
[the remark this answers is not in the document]
AndreKurait: Added a test to verify parallelization

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
AndreKurait: What are you envisioning for this? Is this a blocking comment given the test I added for parallelization checks?

peternied:
@AndreKurait I've got a PR [1] out that reorganizations the code in a way I _think_ is much easier to understand - would love to get your thoughts.  There is still a setting around parrallel that I'm unclear on - but the tests are great - thank you for that effort.

- [1] https://github.com/AndreKurait/opensearch-migrations/pull/4

RFS/src/main/java/com/rfs/common/DocumentReindexer.java
peternied: The new test cases you add cover this scenario - thanks

buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
```diff
@@ -65,7 +65,8 @@ class CommonUtils {
             }
 
             copyFile("jars", "/jars")
-            def jvmParams = "-XX:MaxRAMPercentage=80.0 -XX:+ExitOnOutOfMemoryError"
+            // Leave memory headroom for native execution which allocates memory outside jvm
+            def jvmParams = "-XX:+UseContainerSupport -XX:MaxRAMPercentage=60.0 -XX:+ExitOnOutOfMemoryError"
```
peternied, line 69: Revert this?

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -76,75 +78,91 @@ public static Function<Path, LuceneDocumentsReader> getFactory(boolean softDelet
...
+                } catch (IOException e) {
+                    log.atError().setMessage("Failed to close DirectoryReader").setCause(e).log();
+                    throw Lombok.sneakyThrow(e);
+                }
         });
     }
 
+    Publisher<Document> readDocsByLeavesInParallel(DirectoryReader reader) {
+        var segmentsToReadAtOnce = 5; // Arbitrary value
+        var maxDocumentsToReadAtOnce = 100; // Arbitrary value
```
peternied, line 98: I'm OK if these values give the throughput we are looking for - how does this all hold up with the latest perf numbers?

buildSrc/src/main/groovy/org/opensearch/migrations/common/CommonUtils.groovy
AndreKurait: Will leave this out for a separate PR

DocumentsFromSnapshotMigration/src/test/java/com/rfs/PerformanceVerificationTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyList;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+@Slf4j
+public class PerformanceVerificationTest {
```
peternied, line 35: Tuning this test might be hard - it will also operate differently on your machine, mine, github, and jenkins.  I'm happy to merge this change without a 'wall clock' based perf test.  What do you think about adding a disabled annotation and iterating on it in a future PR?

peternied (approved):
Thanks for the great work in this space.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/PerformanceVerificationTest.java
AndreKurait: Updated to a polling mechanism to not be dependent on the processing speed

PR #907, opened 2024-08-15
Remove unsupported AddOn features from deployment

peternied:
### Description
There was a way in the deployment to add multiple replayers via additional CDK deployment steps.  Since this was added there have been more approachable ways to scale up ECS.   Removing this unsupported and untested feature.

### Issues Resolved
N/A

### Testing
None! Going to depend on Jenkins workflows to verify,

### Check List


lewijacn:
Just at a high level glance I have concern with stripping out the default deploy id everywhere. When we deploy additional resources in the future (say an additional replayer service through the migration console) it seems like having this structure of naming that includes both the stage and deploy id could be useful, e.g. we call traffic replayer service `traffic-replayer-default`. Curious your thoughts here

peternied:
> When we deploy additional resources in the future

When you say deploy do you mean add/update/remove resources in our existing stacks? Resources are associated with a stack, I'd much rather we integrate stack name or some other unique identifier as part of resource naming.  When using identifier provided by users we have no way of being sure there isn't a collision.

The code is concatenating string which is inherently error prone instead of referencing ARNs.  Unless there is a pressing need to keep the deploy id, I'd rather we bring it back when we know the form it should take that will be most resilient.

_Note - I'm basing this on the assumption that for an aws account we will only support deploying one instance of Migration Assistant._

peternied:
>Note - I'm basing this on the assumption that for an aws account we will only support deploying one instance of Migration Assistant.

Turns out this assumption was not true, we need to keep supporting AddOn instances.  Until we have a replacement for adding multiple replayers we cannot touch this part of the workflow.

peternied:
We've been looking at removing the `MigrationSSMParameter` along with sliming down the number of stacks, this could be used as a jumping off point for that effort, but we'd need to create a new stack if we wanted to keep support for a multi-replayer option.

PR #915, opened 2024-08-26
Client Traffic Swinging Docs

AndreKurait:
### Description
Adds documentation around ALB configuration and setup

* Category: Documentation
* Why these changes are required? Give customers who want to deep dive a place to understand the capabilities of the system.
* What is the old behavior before changes and new behavior after changes? Adds markdown in docs.

### Issues Resolved
[MIGRATIONS-1962](https://opensearch.atlassian.net/browse/MIGRATIONS-1962)

### Testing
View docs on github

### Check List

### DIFF
--- a/docs/ClientTrafficSwinging.md
+++ b/docs/ClientTrafficSwinging.md
@@ -76,10 +76,12 @@ graph TD
     end
 
     subgraph MigrationAssistantInfrastructure
-        ALB
-        L1[Weighted Listener :9200]
-        L2[Source Listener :9201]
-        L3[Target Listener :9202]
+        ALB[Application Load Balancer]
+        subgraph ALB
+            L1[Weighted Listener :9200]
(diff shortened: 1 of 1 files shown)


sumobrian (changes requested):
There is something that may be unclear to an audience that doesn't understand how the capture proxy works. Its not clear that the capture proxy is operating in replication mode sending traffic to kafka. I would also clarify at what phase in a migration you would expect traffic to be directed each of these services. For example, before starting the migration, we would test the source and target listens directly but before starting a migration you would direct traffic to the weidghted listener with 100% of traffic going through traffic capture. If migrating existing data from the source, once existing data migration is complete 100% of traffic should be directed to target cluster bypassing the source and kafka.

sumobrian (approved):
Since this documentation is no longer being included in the user guide, but rather a reference is included, I'm approving and merging.

PR #918, opened 2024-08-27
Run some tests in total isolation and all others in parallel

gregschohn:
### Description

Gradle is now configured to run most tasks in parallel, with a special exemption for tests marked to run them in total isolation of anything else that the gradle parent process is doing.  Tasks that can be run in parallel include building targets and running tests.  Notice that typical dependency rules apply.  The number of tasks running concurrently will be limited by the maxWorkerCount that gradle is passed or configures (which is typically the # of CPUs).  

Each project within the project has the same base test configuration.  Targets include `test`, `slowTest`, `isolatedTest`, and `fullTest`, which are defined within the root `build.gradle` file.  Those targets are defined via the @Tag("NAME") attribute on each test (class or method).  A project's `test` tasks will be run as part of its `build` task.  

A summary of each project's target composition is as follows.  Notice that `allTests` exists at the top-level project to depend upon ALL of the test tasks across all of the projects.

| Target | Composition | Purpose |
|---|---|---|
|`slowTest`| `@Tag("longTest")`| Tests that are too slow to provide value for every build run |
|`isolatedTest`| `@Tag("isolatedTest")`| Tests that may skew more toward integration tests and may take seconds to minutes to run.  While these tests may require more time, they shouldn't require exhaustive use of resources or be sensitive to other tasks running in parallel |
|`test`| all other tests NOT marked with the tags above ("longTest" or "isolatedTest")| Tests that require significant or exclusive use of resources or have sensitive performance bounds |
|`fullTest`| a task dependent upon the tasks above | Convenience Task |

The `isolatedTest` task (for each project) will run each of the tagged tests in serial and will run the isolatedTest task itself in serial from all of the other tasks within the project.  While the `isolatedTest` task isn't marked as dependent upon the other tests, it is marked to run _after_ other tests if gradle is set to run them.  That eliminates the requirement that test or slowTest run BEFORE the isolatedTest target when a developer is trying to only run the isolatedTest target.  Likewise, `slowTest` isn't dependent upon `test`, but those two targets may run in parallel since there aren't isolation requirements.  Parallelization for the test runners IS configured for `test` and `slowTest` targets so that those tests may complete quicker on hardware with more capacity.

TrafficCapture overrides the test and slowTest targets to enable netty leak detection only for slowTest.  The regular test targets for the TrafficCapture subprojects sets and environment variable to disable leak detection and the `slowTest` target for those subprojects leaves the `disableMemoryLeakTests` unset but alters the tag definition to include all tests but those tagged with isolatedTest.

* Category Test ehnancements
* Why these changes are required? To improve developer productivity
* What is the old behavior before changes and new behavior after changes? the number of automatic tests run in test is a bit lower.  The number of iterations for tests requiring leak detection is shorter for some longer tests.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2000

### Testing
Lots of gradle tests

### Check List


AndreKurait:
Please test with removing these lines in TrafficCapture build.gradle (added in PR #928 ) https://github.com/opensearch-project/opensearch-migrations/blob/670b9997ae1df5930eb57ab2268de32d087fc2d6/TrafficCapture/build.gradle#L36-L47

Please include scan links and execution duration for `./gradlew :TrafficCapture:test --rerun-tasks` with and without this  change (with it not disabled in the TrafficCapture build.gradle)

AndreKurait (approved):
Given we're not applying the quicker netty leak tests broadly, i'm fine with this PR as is and we can revisit applying it and removing the traffic capture disable for the test task later

AndreKurait (dismissed):
Negating my approval as there have been significant changes since then

TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/ExpiringSubstitutableItemPoolTest.java
```diff
@@ -25,6 +26,7 @@
 
 @Slf4j
 @Isolated("Isolation based on temporal checks")
```
AndreKurait, line 28: nit: this `@Isolated` isn't needed anymore with isolatedTest

AndreKurait (approved):
Looks good, could you do two things before merging
1. Add the chart you've made in the PR description to TESTING.md so it's retained.
2. Rebase/Merge commit so that the GHA is re-ran with all the tests

gradle.properties
```diff
@@ -1,2 +1,7 @@
 org.gradle.caching=true
+org.gradle.configuration-cache=true
+#org.gradle.configureondemand=true
```
peternied, line 3: Remove commented out setting

peternied (approved):

PR #921, opened 2024-08-28
Added Elasticsearch 6.8 support to Document Migrations

chelma:
### Description
* Updated the Doc Migrations code to accept a source version CLI argument
* Centralized some behavior with how we construct parser classes specific to source versions
* Updated unit tests where appropriate
* Added some additional test snapshots.  One (`ES_6_8_Updates_Deletes_Native`) is currently used in our unit tests, while the other two (`ES_5_6_Updates_Deletes`, `ES_6_8_Updates_Deletes_Merged`) are not currently used, but will be used when we support Lucene Version hot-swapping (https://opensearch.atlassian.net/browse/MIGRATIONS-1965) and the ability to handle Indices originally created on an Elastic 5.X cluster in an Elastic 6.X source snapshot.
* Added the ability to spin up an Elastic 5.6 source cluster via docker compose

NOTE: There is some remaining work to add end-to-end unit tests for ES 6.8 Sources.  Our existing end-to-end tests either currently don't cover document migration (see: `EndToEndTest.java`) or support ES 6.8 (see: `ParallelDocumentMigrationsTest.java`, OpenSearch Benchmark doesn't support ES 6).

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1953

### Testing
Lots of manual testing with Docker Compose; unit testing

### Check List


DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -132,6 +131,11 @@ public static class Args {
             description = "Optional.  The maximum number of connections to simultaneously " +
                 "used to communicate to the target, default 10")
         int maxConnections = 10;
+
+        @Parameter(names = { "--source-version" }, description = ("Optional. Version of the source cluster.  Possible "
+            + "values include: ES_6_8, ES_7_10, ES_7_17.  Default: ES_7_10"), required = false,
+            converter = ClusterVersion.ArgsConverter.class)
+        public ClusterVersion sourceVersion = ClusterVersion.ES_7_10;
```
peternied, line 138: Won't we need the target version to support OS 1.3 clusters?

DocumentsFromSnapshotMigration/src/test/java/com/rfs/ParallelDocumentMigrationsTest.java
```diff
@@ -304,7 +304,8 @@ private int migrateDocumentsSequentially(
         String targetAddress,
         AtomicInteger runCounter,
         Random clockJitter,
-        DocumentMigrationTestContext testContext
+        DocumentMigrationTestContext testContext,
+        ClusterVersion parserVersion
```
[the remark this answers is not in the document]
peternied: Looks like this name got used in many places, this value is only parsed once after that it seems like its the source/target version number

RFS/src/main/java/com/rfs/common/SourceResourceProvider.java
```diff
@@ -0,0 +1,14 @@
+package com.rfs.common;
+
+import com.rfs.models.IndexMetadata;
+import com.rfs.models.ShardMetadata;
+
+public interface SourceResourceProvider {
```
peternied, line 6: Thanks for including this change

DocumentsFromSnapshotMigration/src/test/java/com/rfs/ParallelDocumentMigrationsTest.java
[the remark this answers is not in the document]
chelma: Tl;DR - Agreed, will switch

I went back/forth on this one.  The catch here is that the parser version can be different than the source version, and whether you care/know about the parser or source version is dependent on context.   As an example, we will probably end up using the ES 7.10 parsers for OS 1.3.  All that said, on further reflection, I think the bigger point here may be that knowing there is a difference between the parser version and source version is an abstraction leak.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/ParallelDocumentMigrationsTest.java
```diff
@@ -377,23 +380,25 @@ private DocumentsRunner.CompletionStatus migrateDocumentsWithOneWorker(
...
+            IndexMetadata.Factory indexMetadataFactory = sourceResourceProvider.getIndexMetadataFactory(repoDataProvider);
+            ShardMetadata.Factory shardMetadataFactory = sourceResourceProvider.getShardMetadataFactory(repoDataProvider);
             final int ms_window = 1000;
             final var nextClockShift = (int) (clockJitter.nextDouble() * ms_window) - (ms_window / 2);
             log.info("nextClockShift=" + nextClockShift);
 
 
-            Function<Path, LuceneDocumentsReader> readerFactory = path -> new FilteredLuceneDocumentsReader(path, ElasticsearchConstants_ES_7_10.SOFT_DELETES_POSSIBLE,
-                    ElasticsearchConstants_ES_7_10.SOFT_DELETES_FIELD, terminatingDocumentFilter);
+            Function<Path, LuceneDocumentsReader> readerFactory = path -> new FilteredLuceneDocumentsReader(path, sourceResourceProvider.getSoftDeletesPossible(),
```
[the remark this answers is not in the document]
chelma: I'd say if we develop a need for access to the details in the `SourceResourceProvider` via the `LuceneDocumentsReader` after instantiation, then that could make sense.  There doesn't appear a need for that currently, so my gut feeling is to keep things how they are.  I appreciate having the `SourceResourceProvider`'s job being "completed" at the end of the dependency injection setup phase.  There's an argument to be made for removing one parameter from the `LuceneDocumentsReader` but I think having the two parameters specifically called out makes it clearer what that class' dependencies actually are.

RFS/src/main/java/com/rfs/common/SourceResourceProvider.java
chelma: You're welcome!  Not a fan of the class name though, lol...  Gets the job done though!

RFS/src/test/java/com/rfs/common/TestResources.java
```diff
@@ -12,12 +12,18 @@ public static class Snapshot {
...
     }
 
+    public static final Snapshot SNAPSHOT_ES_6_8;
     public static final Snapshot SNAPSHOT_ES_7_10_W_SOFT;
     public static final Snapshot SNAPSHOT_ES_7_10_WO_SOFT;
 
     static {
         Path rfsBaseDir = Paths.get(System.getProperty("user.dir"));
 
+        SNAPSHOT_ES_6_8 = new Snapshot(
```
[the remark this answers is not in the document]
chelma: Ah, yep - good callout, will do.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/ParallelDocumentMigrationsTest.java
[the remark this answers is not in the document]
peternied: I think what you are calling out here is a bigger problem that we need to address.  There is the version information supplied by the user that might not be accurate vs the version of the snapshot artifacts or service responses.

I'm working on this in the metadata systems - since the scenario is a bit more interactive.  I'll iterate on this space in my upcoming PR and we can see how this aligns with what you are thinking.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
```diff
@@ -0,0 +1,277 @@
...
+            } catch (ExecutionException e) {
+                var child = e.getCause();
+                if (child instanceof ExpectedMigrationWorkTerminationException) {
+                    // This is what we expect to happen
+                } else {
+                    throw Lombok.sneakyThrow(child);
+                }
+            } catch (InterruptedException e) {
+                throw Lombok.sneakyThrow(e);
+            }
```
peternied, line 136: Since we aren't running the workers in parallel this can be greatly simplified down to something like the following:
```java
      var result = migrateDocumentsWithOneWorker(
          sourceRepo,
          snapshotName,
          List.of(),
          targetCluster.getUrl(),
          new AtomicInteger(),
          new Random(1),
          testDocMigrationContext,
          sourceCluster.getVersion().getSourceVersion()
      ));
      assertThat(result, equalTo(WORK_COMPLETED));
```

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
```diff
@@ -0,0 +1,277 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+
+public class EndToEndTest extends SourceTestBase {
+    @TempDir
+    private File localDirectory;
+
+    @ParameterizedTest(name = "Target OpenSearch {0}")
+    @ArgumentsSource(SupportedTargetCluster.class)
+    public void migrateFrom_ES_v6_8(final SearchClusterContainer.Version targetVersion) throws Exception {
+        var executorService = Executors.newFixedThreadPool(1);
```
peternied, line 41: It would be simpler to remove the executor by having RFS run on the test thread.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
```diff
@@ -0,0 +1,277 @@
...
+            final var targetCluster = new SearchClusterContainer(targetVersion)
+        ) {
+            // === ACTION: Set up the source/target clusters ===
+            var bothClustersStarted = CompletableFuture.allOf(
+                CompletableFuture.runAsync(() -> sourceCluster.start()),
+                CompletableFuture.runAsync(() -> targetCluster.start())
+            );
+            bothClustersStarted.join();
+
+            // Create a template
```
peternied, line 58: Since this test is focused on the document reindexing, seems like all the template logic/metadata migration could be removed as its duplicated in `MetadataMigration/src/test/java/org/opensearch/migrations/EndToEndTest.java`.  What do you think about separating the validation of the metadata toolset from the document?

Note; if we did remove this validation from this test case, then can you make sure the ES6.8 scenarios are working in those tests cases?

DocumentsFromSnapshotMigration/src/main/java/com/rfs/RfsMigrateDocuments.java
```diff
@@ -35,15 +36,13 @@
...
 import com.rfs.common.SnapshotShardUnpacker;
 import com.rfs.common.SourceRepo;
+import com.rfs.common.SourceResourceProvider;
+import com.rfs.common.SourceResourceProviderFactory;
 import com.rfs.common.TryHandlePhaseFailure;
 import com.rfs.common.http.ConnectionContext;
 import com.rfs.models.IndexMetadata;
 import com.rfs.models.ShardMetadata;
 import com.rfs.tracing.RootWorkCoordinationContext;
-import com.rfs.version_es_7_10.ElasticsearchConstants_ES_7_10;
```
peternied, line 43: 🔥

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
chelma: Good callout.  I stuck with the more complex `migrateDocumentsSequentially` because there might be multiple shards worth of data and we want to ensure they are all migrated, but our final check that all our docs are migrated will catch if they aren't.  I'll move to the simpler approach and we can re-add the more complex one if/when required.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
chelma: I added this because I needed to match the api of `migrateDocumentsSequentially`.  Not necessary now that I'll do `migrateDocumentsWithOneWorker` instead.  Will remove.

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
chelma: I'd like to have a end-to-end test in our unit tests, happy to have that live here or elsewhere, and have that be a simplified case.  I'm also fine having some overlap with the (see these and `ParallelDocumentMigrationsTest`, for example).  What if we keep these tests how they are and you add your own transformation tests with much more complex inputs in the `MetadataMigration` sub-project?

DocumentsFromSnapshotMigration/src/test/java/com/rfs/EndToEndTest.java
peternied: Lets keep the test methods focused on what is being tested, all of these exception handling and unverified locals aren't relevant to this test scenario.  I'd be happy if you refactored it to work in parallel or with a single worker.

peternied (approved):

PR #923, opened 2024-09-03
[Jenkins] Allow performing AWS actions with an assumed role and minor inconsistency fixes

lewijacn:
### Description
* Improves Jenkins posture by specifying an AWS account and role for given AWS actions
* Minor bug fixes with context options and some improper naming
* Moves away from providing VPC id as a parameter and allows CDK to create the VPC

### Issues Resolved
N/A

### Testing
Cloud jenkins testing

### Check List


vars/defaultIntegPipeline.groovy
```diff
@@ -28,29 +28,57 @@ def call(Map config = [:]) {
...
             stage('Checkout') {
                 steps {
-                    git branch: "${params.GIT_BRANCH}", url: "${params.GIT_REPO_URL}"
+                    script {
+                        // Allow overwriting this step
+                        if (config.checkoutStep) {
+                            config.checkoutStep()
+                        } else {
+                            git branch: "${params.GIT_BRANCH}", url: "${params.GIT_REPO_URL}"
+                        }
```
peternied, line 37:
```suggestion
                    script {
                        config.checkoutStep()
```

I'm not a fan of this pattern, can we have the checkoutStep have a default value be `git branch: "${params.GIT_BRANCH}", url: "${params.GIT_REPO_URL}"`

This changes the model so there is no if/else branch, we always execute the checkout step, but it can be customized.  What do you think of this?

vars/defaultIntegPipeline.groovy
```diff
@@ -76,7 +104,11 @@ def call(Map config = [:]) {
                                     if (skipCaptureProxyOnNodeSetup) {
                                         baseCommand += " --skip-capture-proxy"
                                     }
-                                    sh baseCommand
+                                    withCredentials([string(credentialsId: 'migrations-test-account-id', variable: 'MIGRATIONS_TEST_ACCOUNT_ID')]) {
+                                        withAWS(role: 'JenkinsDeploymentRole', roleAccount: "${MIGRATIONS_TEST_ACCOUNT_ID}", duration: 5400, roleSessionName: 'jenkins-session') {
```
peternied, line 108: Can we tweak down this duration, what is this tied to? 90 minutes is a long time, and its different than the time at line 137 of 60 minutes

vars/defaultIntegPipeline.groovy
peternied: > we don't currently use it anywhere

If we don't use it, why add it?  Can we revisit this later if/when its needed?

lewijacn: It has been useful for some adhoc testing in the UI to be able to override this directly rather than repeatedly specifying the parameters. I'd like to keep it for that sake, at least until we have a better way to test Jenkins changes locally.

PR #928, opened 2024-09-05
Unified code coverage

AndreKurait:
### Description
Generates a task mergeJacocoReports which runs all tests that run jacoco and generates one shared report.

Cleans up some of the duplicated definitions in build.gradle

Fixes an issue where some classes weren't showing adequate coverage when tested by a different project.

* Category: Bug Fix, Refactoring

### Issues Resolved
[MIGRATIONS-1976](https://opensearch.atlassian.net/browse/MIGRATIONS-1976)

### Testing
Missing coverage verified to exist
image

Overall coverage + 7%

### Check List

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -112,12 +112,13 @@ jobs:
         with:
           name: test-reports
           path: |
+            **/build/reports/tests/
             **/reports/jacoco/mergedReport/
 
       - name: Upload to Codecov
         uses: codecov/codecov-action@v4
         with:
-          file: "./build/reports/jacoco/mergedReport/jacocoMergedReport.xml"
+          files: ${{ github.workspace }}/build/reports/jacoco/mergedReport/jacocoMergedReport.xml
           flags: gradle-test
           fail_ci_if_error: true
           disable_search: true
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -103,7 +103,7 @@ jobs:
           OS_MIGRATIONS_GRADLE_SCAN_TOS_AGREE_AND_ENABLED: ''
 
       - name: Run Tests with Coverage
-        run: ./gradlew test -x TrafficCapture:dockerSolution:build -x spotlessCheck jacocoTestReport --scan --stacktrace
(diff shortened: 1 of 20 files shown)


.github/workflows/CI.yml
```diff
@@ -112,14 +112,15 @@ jobs:
         with:
           name: test-reports
           path: |
-            **/build/reports/tests/
+            **/reports/jacoco/mergedReport/
```
peternied, line 115: By updating this line it removed the junit test reports that include logging alongside the test class which I've always found really useful for debugging.

Or if you'd like seperate out the artifacts so there is one for the project reports (junit) and another for the project-wide reports (the merged jacoco report)

build.gradle
```diff
@@ -54,6 +62,45 @@ subprojects {
...
+            excludeTags 'longTest'
+        }
+        jacoco.enabled = true
+    }
+
+    tasks.register('slowTest', Test) {
+        useJUnitPlatform {
+            includeTags 'longTest'
+        }
+        systemProperty 'disableMemoryLeakTests', 'false'
```
peternied, line 102: Can we get rid of this environment variable and tag these tests with `longTest`?

peternied (commented):
Great to see those CC numbers go up!  It looks like the gradle check is now taking 30mins, whereas it was 20min before - do you know where the extra time is from?

.github/workflows/CI.yml
AndreKurait: I'll keep both

AndreKurait:
> Great to see those CC numbers go up! It looks like the gradle check is now taking 30mins, whereas it was 20min before - do you know where the extra time is from?

We're not always running all of the tests and using gradle caching. See other PR that modified more packages and had a longer run.
 
https://github.com/opensearch-project/opensearch-migrations/actions/runs/10598495021

build.gradle
peternied: [Offline] Yea, its more complex than I first thought.  I'd be OK leaving this enabled when running slowTests all the time

peternied (approved):

build.gradle
AndreKurait: False == not specifying at all, i'm removing false and only specifying it as true where needed

PR #937, opened 2024-09-08
Replayer Transformation support

gregschohn:
### Description

Json transformations can work when bodies aren't present and when the bodies are of 'ndjson' types (like OpenSearch bulk requests).

There was also a test (testMalformedPayload_andTypeMappingUri_IsPassedThrough) that sent malformed JSON in a payload.  The test didn't do much and a comment admitted that we weren't likely doing the right thing.  Moving a buffer to Unpooled caused the test to fail w/ a leak.  Upon closer inspection, errors while parsing JSON didn't cause the payload contents to be dumped into the inlinedBinaryBody key of the payload.  Now that content is captured and passed forward (plus, refCounted appropriately, though still as an Unpooled to minimize risk from transformations [see code comment]).  In the case of this existing test, which has been moved to HttpJsonTransformingConsumerTest, the JsonTransformerForOpenSearch23PlusTargetTransformerProvider that the test uses throws an exception when the json payload isn't present.  That was also not being handled properly.  Now it is handled by marking the transformation as an error w/ the exception wrapped in a "TransformationException".  In that case, the transformation status is marked as an error and the contents of the entire message will be empty.  It feels more appropriate that if a transformation threw to be as conservative as possible and not put anything on the wire, in case some part of it was sensitive.
Lastly, I refactored the AggregatedRawResponse into a base class for AggregatedRawResult.  We use that class to accumulate transformation results.  It was extremely confusing to see an HTTP response and the word response on so many of the fields.  Now the word 'result' is used and the usage in various contexts makes more sense.

* Category Enhancement/Bug fix
* Why these changes are required? So that users can transform contents of http requests.  This was a feature that was added nearly 18 months ago, but has been incomplete as shown by some simple OpenSearch Benchmark test runs.
* What is the old behavior before changes and new behavior after changes? Before, a number of requests would be garbled in various ways when payload transformations were present.  Now they should be passed through much more reasonably.  This at least gets us to the point where we can now see some non-trivial bugs.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1984

### Testing
Unit testing and docker compose testing.

### Check List


TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml
```diff
@@ -78,7 +78,7 @@ services:
...
...ic-topic logging-traffic-topic --kafka-traffic-group-id logging-group-default --otelCollectorEndpoint http://otel-collector:4317" #--transformer-config-base64 W3sgIkpzb25Kb2x0VHJhbnNmb3JtZXJQcm92aWRlciI6ClsKICB7CiAgICAic2NyaXB0IjogewogICAgICAib3BlcmF0aW9uIjogInNoaWZ0IiwKICAgICAgInNwZWMiOiB7CiAgICAgICAgIm1ldGhvZCI6ICJtZXRob2QiLAogICAgICAgICJVUkkiOiAiVVJJIiwKICAgICAgICAiaGVhZGVycyI6ICJoZWFkZXJzIiwKICAgICAgICAicGF5bG9hZCI6IHsKICAgICAgICAgICJpbmxpbmVkSnNvbkJvZHkiOiB7CiAgICAgICAgICAgICJ0b3AiOiB7CiAgICAgICAgICAgICAgInRhZ1RvRXhjaXNlIjogewogICAgICAgICAgICAgICAgIioiOiAicGF5bG9hZC5pbmxpbmVkSnNvbkJvZHkudG9wLiYiIAogICAgICAgICAgICAgIH0sCiAgICAgICAgICAgICAgIioiOiAicGF5bG9hZC5pbmxpbmVkSnNvbkJvZHkudG9wLiYiCiAgICAgICAgICAgIH0sCiAgICAgICAgICAiKiI6ICJwYXlsb2FkLmlubGluZWRKc29uQm9keS4mIgogICAgICAgICAgfQogICAgICAgIH0KICAgICAgfQogICAgfQogIH0sIAogewogICAic2NyaXB0IjogewogICAgICJvcGVyYXRpb24iOiAibW9kaWZ5LW92ZXJ3cml0ZS1iZXRhIiwKICAgICAic3BlYyI6IHsKICAgICAgICJVUkkiOiAiPXNwbGl0KCcvZXh0cmFUaGluZ1RvUmVtb3ZlJyxAKDEsJikpIgogICAgIH0KICB9CiB9LAogewogICAic2NyaXB0IjogewogICAgICJvcGVyYXRpb24iOiAibW9kaWZ5LW92ZXJ3cml0ZS1iZXRhIiwKICAgICAic3BlYyI6IHsKICAgICAgICJVUkkiOiAiPWpvaW4oJycsQCgxLCYpKSIKICAgICB9CiAgfQogfQpdCn1dCg=="
```
peternied, line 81: This command becomes unreadable with the base64 encoded here - can we instead load it from a file path to improve comprehension?

gregschohn: @peternied - I don't expect customers to be loading this from a file.  They could, but that would require them to get a file onto a replayer container.  For our CDK deployment, it's much easier to add an extra argument to pass to the replayer than to add an additional file.  As such, I was hoping to test something closer to what more customers would be doing.
@chelma - It's a comment for now.  I'm debugging through why its inclusion caused some of the E2E tests to fail.

TrafficCapture/trafficReplayer/README.md
```diff
@@ -140,8 +148,9 @@ To run only one transformer without any configuration, the `--transformer-config
 be set to the name of the transformer (e.g. 'JsonTransformerForOpenSearch23PlusTargetTransformerProvider', 
 without quotes or any json surrounding it).
 
-The user can also specify a file to read the transformations from using the `--transformer-config-file`, but can't use
-both transformer options.
+The user can also specify a file to read the transformations from using the `--transformer-config-file`.  Users can
```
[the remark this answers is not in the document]
gregschohn: I'm not w/ the CDK deployment today, which is why I put a Base64 option in there too.  Further out, it might make sense to pull it via a parameter store and write it as part of an init task.  I'm going to try not bother with changing things like container initialization until I'm done w/ the K8s investigation.

gregschohn:
> Thanks for the PR. This is my first time looking through this code (really should have investigated it earlier), and I'm struggling a bit to see how we'd take something like the `HttpJsonTransformingConsumer` and generalize its mechanisms to work for, say, the transformations we need for RFS. Curious what your thoughts are on that?

As a developer and a user, I would expect to see extensibility in a similar way for metadata migration and document migration.  That means having a simple dependency-free API from the metadata/document migration processes to call into hot-loadable classes (via ServiceProvider).  Beyond that, I'd like to see the same translation environments supported for each of our 3 migration applications.  

Having a shared CompositeTransformer for all of these could be a start.  That composite transformer may eventually handle conditionals and maybe is represented as a graph or tree.  At that point, it should be shared by all of the applications that need to support transformations.

On the other side - what is being transformed, I'm not sure what we'll want to share.  Ideally, it's a very consistent environment that develops.  For example, the CompositeTransformer and TransformationLoader (that configures transformations) should be consistently shared.  Admittedly, both of these are super simple/lightweight today, but complexity and value aren't necessarily correlated.

Further into the future, I'd suspect that we'll want to develop "document" migrations for metadata and backfill document migrations.  These document migrations should be able to be pulled into the replayer since that's going to need to do all of the transformations of the other two systems since RESTful calls can go into the source w/ those document types.  I suspect that we'll build up a library of transforms, each which may be configured.  That library will be usable from all of the applications and may also allow transformations to be embedded/extended, etc.

So - things to care about for today.  What are document types.  Right now, for the replayer there are 3 kinds of documents - json, ndjson, binary.  Is that enough?  I suspect that for document and metadata migration, those will likely just need to care about json transformations.  Ideally, the replayer will be able to consume those shared transformations in the future.

TrafficCapture/trafficReplayer/README.md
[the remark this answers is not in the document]
chelma: Makes sense, thanks.

TrafficCapture/dockerSolution/src/main/docker/docker-compose.yml
peternied: > I don't expect customers to be loading this from a file.

How do you expect customer will be building up the configuration?  I would expect them to be iterating in a text editor, does your workflow expect them to process then and then copy/paste into a cdk variable and then redeploy.

If we are using a file on a shared EFS volume that text editor could be vim on the migration console, this could be reloaded on boot of the container.  This means no deployed is required, just stop/start on the TC instance... or at least that is where my head is at.

gregschohn: Please file a new feature for setting up the EFS volume for that.  We can tackle that separately.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -199,9 +273,12 @@ private static Parameters parseArgs(String[] args) {
     private static String getTransformerConfig(Parameters params) {
         if (params.transformerConfigFile != null
             && !params.transformerConfigFile.isBlank()
+            && params.transformerConfigEncoded != null
+            && !params.transformerConfigEncoded.isBlank()
             && params.transformerConfig != null
             && !params.transformerConfig.isBlank()) {
-            System.err.println("Specify either --transformer-config or --transformer-config-file, not both.");
+            System.err.println("Specify only one of --transformer-config-base64, --transformer-config or " +
```
AndreKurait, line 280: This is currently only occurring if all 3 are specified, let's make sure this occurs if we have 2/3 specified

AndreKurait, line 280: This is currently only occurring if all 3 are specified, let's make sure this occurs if we have 2/3 specified

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/JsonAccumulator.java
```diff
@@ -30,28 +31,42 @@ public class JsonAccumulator {
...
      */
-    public Object consumeByteBuffer(ByteBuffer byteBuffer) throws IOException {
-        ByteBufferFeeder feeder = (ByteBufferFeeder) parser.getNonBlockingInputFeeder();
+    public Object consumeByteBufferForSingleObject(ByteBuffer byteBuffer) throws IOException {
+        consumeByteBuffer(byteBuffer);
+        return getNextTopLevelObject();
+    }
+
+    public void consumeByteBuffer(ByteBuffer byteBuffer) throws IOException {
         log.trace("Consuming bytes: " + byteBuffer.toString());
```
AndreKurait, line 65: Let's fix this trace message while we're here

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodyAccumulateHandler.java
```diff
@@ -22,30 +27,62 @@ public class NettyJsonBodyAccumulateHandler extends ChannelInboundHandlerAdapter
...
+                        throw new IndexOutOfBoundsException("JSON contents were too large " +
+                            jsonAccumulator.getTotalBytesFullyConsumed() + " for a single composite ByteBuf");
+                    }
+                    // skip the contents that were already parsed and included in the payload as parsed json
+                    accumulatedBody.readerIndex((int) jsonAccumulator.getTotalBytesFullyConsumed());
+                    // and pass the remaining stream
+                    capturedHttpJsonMessage.payload()
+                        .put(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY, accumulatedBody);
+                } else {
+                    accumulatedBody.release();
```
AndreKurait, line 82: Can we also do a release if the channel is unregistered and accumulatedBody is still present

AndreKurait (changes requested):

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodyAccumulateHandler.java
```diff
@@ -74,10 +109,12 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                             jsonAccumulator.getTotalBytesFullyConsumed() + " for a single composite ByteBuf");
                     }
                     // skip the contents that were already parsed and included in the payload as parsed json
-                    accumulatedBody.readerIndex((int) jsonAccumulator.getTotalBytesFullyConsumed());
                     // and pass the remaining stream
+                    var jsonBodyByteLength = jsonWasInvalid ? 0 : (int) jsonAccumulator.getTotalBytesFullyConsumed();
                     capturedHttpJsonMessage.payload()
-                        .put(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY, accumulatedBody);
+                        .put(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY,
+                            accumulatedBody.retainedSlice(jsonBodyByteLength,
```
AndreKurait, line 116: nit: should the first argument be accumulatedBody.getReaderIndex + jsonBodyByteLength?

AndreKurait: I'm fine with this since this the accumulatedBody is controlled by this class

AndreKurait (approved):

PR #938, opened 2024-09-09
Add Jenkins Webhook and RFS E2E job on PR

lewijacn:
### Description
This change is hopefully the final chapter in the Jenkins setup saga and a huge leap in the integration testing story. It adds support for triggering a Jenkins job through a configured Github Action and monitoring its completion to pass/fail the action. This design is one of many that can be taken, but a benefit of this approach is that it keeps the Github action in control of triggering Jenkins jobs and status updates, as well as doesn't require Jenkins to update the PR (though this may be a useful addition for logs or other data in the future). As a part of this pattern, a general python webhook was created to trigger Jenkins jobs and monitor for a terminal state, this could also potentially become its own GHA plugin separate from this repo.

Additionally this adds our first GHA which triggers a Jenkins job for our RFS E2E integration test. This GHA is planned to be monitored as any bugs/issues are worked out and should not currently be restrictive for PRs to be merged.

Jenkins pipeline files have also been updated to attain a lock on deployment stages to prevent different pipeline executions from using the same stage. The pipeline files have also been largely moved to the shared library, to allow any changes to the context options or other settings for the pipeline to be tested in PR actions.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1441

### Testing
See testing GHA on fork: https://github.com/lewijacn/opensearch-migrations/pull/18
Manual testing of webhook trigger

### Check List


jenkins/webhookTriggers/default_webhook_trigger.py
peternied, line 1: Could we reuse an existing action rather than rolling our own such as https://github.com/marketplace/actions/trigger-jenkins-job?

peternied (changes requested):
Nice that you got this working, but I think this introduces a maintainance burden for the jenkins trigger - it looks like the GHA marketplace has some actions that can handle our use case.  Can we us them instead?

peternied (commented):
Much cleaner, thanks.  Just waiting to see a clean run on your fork and then I'll approve.

jenkins/webhookTriggers/default_webhook_trigger.py
peternied: Moving to your own github action is just fine too, thanks!

.github/workflows/rfs_pr_e2e_test.yml
```diff
@@ -0,0 +1,40 @@
+name: Jenkins E2E
```
peternied, line 1:
```suggestion
name: Jenkins
```
E2E isn't really required here and uses acronyms that folks might not be familiar with.

.github/workflows/rfs_pr_e2e_test.yml
```diff
@@ -0,0 +1,40 @@
...
+    types: [opened, synchronize, reopened]
+
+env:
+  python-version: '3.11'
+
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
+jobs:
+  rfs-e2e-aws-test:
```
peternied, line 18: This doesn't read very clean to me, RFS is a technology, but the scenario is a backfill, what do you think of the following?
```suggestion
  migration-assistant-backfill-on-aws:
```

.github/workflows/rfs_pr_e2e_test.yml
```diff
@@ -0,0 +1,40 @@
...
+            pr_repo_url="https://github.com/${{ github.event.pull_request.head.repo.full_name }}.git"
+          else
+            branch_name="${{ github.ref_name }}"
+            pr_repo_url="https://github.com/${{ github.repository }}.git"
+          fi
+          echo "Running jenkins test on repo: $pr_repo_url and branch: $branch_name"
+          echo "branch_name=$branch_name" >> $GITHUB_OUTPUT
+          echo "pr_repo_url=$pr_repo_url" >> $GITHUB_OUTPUT
+      - name: Jenkins Job Trigger and Monitor
+        uses: lewijacn/jenkins-trigger@1.0.4
```
peternied, line 35: Tip for next time; while iterating you can point to `main` during development until you ship a version that you are sure works, then create a release/tag

`uses: lewijacn/jenkins-trigger@main`

peternied (approved):
Minor comments, you are good to merge when ready - nicely done!

.github/workflows/rfs_pr_e2e_test.yml
lewijacn: I'd like to have this match the pipeline name for now but I'm open to changing this in the future if we decide we don't like it

PR #947, opened 2024-09-10
DRAFT - DO NOT MERGE - Parallel cicd

AndreKurait:
### Description

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/KafkaConfigurationCaptureProxyTest.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/KafkaConfigurationCaptureProxyTest.java
@@ -129,17 +129,30 @@ public class KafkaConfigurationCaptureProxyTest {
             captureProxy.start();
             final int numberOfTests = 20;
 
-            // Performance is different for first few calls so throw them away
-            assertBasicCalls(captureProxy, 3);
+            // Performance is different for first call so throw away
+            assertBasicCalls(captureProxy, 1);
 
-            var averageBaselineDuration = assertBasicCalls(captureProxy, numberOfTests);
+            var totalBaselineDuration = Duration.ZERO;
+            var totalImpairedDuration = Duration.ZERO;
 
-            failureMode.apply(kafkaProxy);
+            for (int i = 0; i < numberOfTests; i++) {
+                // Run baseline case
+                totalBaselineDuration = totalBaselineDuration.plus(assertBasicCalls(captureProxy, 1));
 
-            // Calculate average duration of impaired calls
-            var averageImpairedDuration = assertBasicCalls(captureProxy, numberOfTests);
+                // Apply failure mode
+                failureMode.apply(kafkaProxy);
 
-            long acceptableDifference = Duration.ofMillis(25).toMillis();
+                // Run impaired case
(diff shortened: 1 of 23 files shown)


TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/KafkaConfigurationCaptureProxyTest.java
```diff
@@ -129,17 +129,30 @@ public void testCaptureProxyWithKafkaImpairedDoesNotAffectRequest_proxysRequest(
...
+                removeAllFailureModes(kafkaProxy);
+            }
 
-            // Calculate average duration of impaired calls
-            var averageImpairedDuration = assertBasicCalls(captureProxy, numberOfTests);
+            var averageBaselineDuration = totalBaselineDuration.dividedBy(numberOfTests);
+            var averageImpairedDuration = totalImpairedDuration.dividedBy(numberOfTests);
 
-            long acceptableDifference = Duration.ofMillis(25).toMillis();
+            long acceptableDifference = Duration.ofMillis(20).toMillis();
```
gregschohn, line 155: why did this get tightened?

build.gradle
```diff
@@ -78,6 +78,9 @@ subprojects {
             exclude project.property('excludeTests')
         }
 
+        // Allow running different test classes in the same project in parallel
+        maxParallelForks = Runtime.runtime.availableProcessors()
```
gregschohn, line 82: I think it's probably fair to leave all 'regular' tests run in process.  That might net a marginal savings for the number of projects.  If it needs to be in its own process, it's likely a slow test anyway (forking & IPC will create a minor hit)

build.gradle
```diff
@@ -95,6 +98,10 @@ subprojects {
     }
 
     tasks.register('slowTest', Test) {
+        // Limit maxParallelForks to half the CPU cores or 1
+        maxParallelForks = Math.min(maxParallelForks as Integer, Runtime.runtime.availableProcessors() / 2 as Integer)
```
gregschohn, line 102: what was it set to before?  Why half the cores for slow tests?  Won't each of these projects create its own forked set of runners?  I'm worried that 8 slow tests could create 8x(# cores)/2 different tests - and they might all be very big.

AndreKurait: max parallel forks creates a number of tasks that gradle submits to a worker pool, without increasing it via `--max-workers` command line flag, this will create `8x(# cores)/2` tasks, but only up to `# cores` could be executed at the same time.

Furthermore, unless `--parallel` is provided on the command line, we will not execute different types of tasks concurrently, so we will fan out for a specific test (up to it's allowed forks and global workers), but not execute different tasks at once (e.g. RFS:slowTest and replayer:slowTest)

TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/KafkaConfigurationCaptureProxyTest.java
AndreKurait: I made this more resilient with the other changes so this can be tightened.

build.gradle
AndreKurait: This is because different tasks are not running in parallel, e.g. RFS and CoreUtilities will not run at the same time.

If i did that, i ran into an issue where having different project's slowTasks running at once could overload the docker memory allocation

PR #956, opened 2024-09-13
Proxy header translations

gregschohn:
### Description

This PR introduces the --setHeader argument to the CaptureProxy (arity 2: "name value" --> "name: value").  This flag will remove any old occurrence of the given header and add a new header with the given name/value at the beginning of the headers block of the http request that is sent to the destination server.  The argument can be repeated for any number of headers ("--setHeader host actual.target.example --setHeader X-Capture-Proxy-Awareness-Flag captureOn").  Notice that the CAPTURED traffic will NOT include these extra headers.  We'll capture exactly what the client was sent.  There are no translations/manipulations on the responses coming back.

* Category Enhancement
* Why these changes are required? To support scenarios where there's a load balancer fronting proxies and the proxies are on separate hardware from the destination servers.
* What is the old behavior before changes and new behavior after changes?  This is a new feature.   Now users can add/set new headers on requests to the destination, for every request coming from clients.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1998

### Testing
Unit tests, including new ones.

### Check List


TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -299,6 +345,9 @@ protected static SslContext loadBacksideSslContext(URI serverUri, boolean allowI
     }
 
     protected static Map<String, String> convertPairListToMap(List<String> list) {
+        if (list == null) {
```
AndreKurait, line 348: Should we remove this or change to an asset if it's unexpected case?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -375,4 +420,35 @@ public static void main(String[] args) throws InterruptedException, IOException
...
+        var headers = convertPairListToMap(headerOverrides);
+        return new ProxyChannelInitializer(
+            rootContext,
+            backsideConnectionPool,
+            sslEngineSupplier,
+            connectionFactory,
+            headerCapturePredicate
+        ) {
+            @Override
+            protected void initChannel(@NonNull SocketChannel ch) throws IOException {
```
AndreKurait, line 440: Is this the right place for this logic?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -375,4 +420,35 @@ public static void main(String[] args) throws InterruptedException, IOException
...
+                super.initChannel(ch);
+                for (var kvp : headers.entrySet()) {
```
AndreKurait, line 442: Is this the right place for this logic?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -375,4 +420,35 @@ public static void main(String[] args) throws InterruptedException, IOException
...
+                    var lineBytes = (kvp.getKey() + ": " + kvp.getValue()).getBytes(StandardCharsets.UTF_8);
+                    ch.pipeline().addAfter(ProxyChannelInitializer.CAPTURE_HANDLER_NAME, "AddHeader-" + kvp.getKey(),
+                        new HeaderAdderHandler(Unpooled.unreleasableBuffer(Unpooled.wrappedBuffer(lineBytes))));
```
AndreKurait, line 445: Does this mean we're creating a new unreleasableBuffer for every single channel? Would it make more sense to create these once and share duplicates of them?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -299,6 +345,9 @@ protected static SslContext loadBacksideSslContext(URI serverUri, boolean allowI
     }
 
     protected static Map<String, String> convertPairListToMap(List<String> list) {
+        if (list == null) {
+            return Map.of();
+        }
         var map = new TreeMap<String, String>();
```
AndreKurait, line 351: Should we use a `LinkedHashMap` to preserve ordering?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -0,0 +1,52 @@
...
+                composite.addComponent(true, headerLineToAdd.duplicate());
+                composite.addComponent(true, (useCarriageReturn ? CRLF_BYTE_BUF : LF_BYTE_BUF).duplicate());
+                composite.addComponent(true, buf.retainedSlice(upToIndex, buf.readableBytes()-upToIndex));
+                buf.release();
+                super.channelRead(ctx, composite);
+                insertedHeader = true;
+                return;
+            }
+        }
+        buf.resetReaderIndex();
```
AndreKurait, line 49: I don't like that we're reseting the readerIndex here after potentially calling `buf.release`

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -375,4 +420,35 @@ public static void main(String[] args) throws InterruptedException, IOException
...
+            sslEngineSupplier,
+            connectionFactory,
+            headerCapturePredicate
+        ) {
+            @Override
+            protected void initChannel(@NonNull SocketChannel ch) throws IOException {
+                super.initChannel(ch);
+                for (var kvp : headers.entrySet()) {
+                    var lineBytes = (kvp.getKey() + ": " + kvp.getValue()).getBytes(StandardCharsets.UTF_8);
+                    ch.pipeline().addAfter(ProxyChannelInitializer.CAPTURE_HANDLER_NAME, "AddHeader-" + kvp.getKey(),
```
AndreKurait, line 444: With multiple AddHeaders, this will reverse the output of them, should we reverse iterate through the map instead?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    private static final ByteBuf CRLF_BYTE_BUF =
+        Unpooled.unreleasableBuffer(Unpooled.wrappedBuffer("\r\n".getBytes(StandardCharsets.UTF_8)));
+    private static final ByteBuf LF_BYTE_BUF =
+        Unpooled.unreleasableBuffer(Unpooled.wrappedBuffer("\n".getBytes(StandardCharsets.UTF_8)));
+    boolean insertedHeader = false;
+    private final ByteBuf headerLineToAdd;
+    boolean useCarriageReturn;
+
+    public HeaderAdderHandler(ByteBuf headerLineToAdd) {
+        this.headerLineToAdd = headerLineToAdd.retain();
```
AndreKurait, line 20: This feels strange to have retain here without a release.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -0,0 +1,52 @@
...
+        while (buf.isReadable()) {
+            var nextByte = buf.readByte();
+            if (nextByte == '\r') {
+                useCarriageReturn = true;
+            } else if (nextByte == '\n') {
+                final var upToIndex = buf.readerIndex();
+                var composite = Unpooled.compositeBuffer(4);
+                buf.resetReaderIndex();
+                composite.addComponent(true, buf.retainedSlice(0, upToIndex));
+                composite.addComponent(true, headerLineToAdd.duplicate());
```
AndreKurait, line 40: There doesn't seem to be a guarantee this is an unreleaseable buffer, should we call retain here?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: What's the problem with keeping it?  Right now, JCommander doesn't return null when the arg isn't included.  I was concerned that if that ever changed, we'd get NPEs where we'd really just want to know that there weren't elements defined.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: Hmmm - let me pull the unreleaeable buffer outside of this block

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: Sure

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: Only relevant if I'm using a LinkedHashMap, so yeah, I'll make the change.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @Override
+    public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+        if (!(msg instanceof ByteBuf) || insertedHeader) {
+            super.channelRead(ctx, msg);
+            return;
+        }
+        var buf = (ByteBuf) msg;
+        buf.markReaderIndex();
```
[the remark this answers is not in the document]
gregschohn: This is the only spot that we're using it.  The full documentation on mark & reset from the ByteBuf javadoc is 4 sentences long.  Considering that this prevents the code from having to alloc a new object, yes, I'd say that it's worth it.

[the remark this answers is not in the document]
gregschohn: Actually - there's a bug that I hadn't accounted for - we can remove payload content that matches the pattern.  I'll put a test in to prevent that & fix it.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
gregschohn: Yeah - that's a bug.  I'll add the release to the unregistered.  It didn't come up as an issue because the one that is sent in is a constant unreleasable buffer,.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: What do you mean?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
AndreKurait: i'm fine with keeping it in for now, would like this to be tested in the future

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -0,0 +1,59 @@
...
+        while (buf.isReadable()) {
+            var nextByte = buf.readByte();
+            if (nextByte == '\r') {
+                useCarriageReturn = true;
+            } else if (nextByte == '\n') {
+                final var upToIndex = buf.readerIndex();
+                var composite = Unpooled.compositeBuffer(4);
```
AndreKurait, line 38: Should you use ctx.alloc().compositeBuffer(4) here?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -0,0 +1,59 @@
...
+                buf.resetReaderIndex();
+                composite.addComponent(true, buf.retainedSlice(0, upToIndex));
```
AndreKurait, line 40: Use `buf.readerIndex()` instead of 0

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
AndreKurait: It feels weird to have this logic inside CaptureProxy.java, i'd rather keep more logic outside of that class. For the sake of this PR i'm fine with it being in here for now.

AndreKurait (approved):

PR #959, opened 2024-09-16
"Back to the Future" - Upgrading to Lucene 9 to read Lucene 6

chelma:
### Description
* This change enables reading Lucene indices v6+ by upgrading RFS to use Lucene 9 and leveraging the backwards compatibility codecs embedded in that release.
* Add a new test for the LuceneDocumentsReader that ensures it's able to read an ES 6.8 snapshot whose indices were created in ES 5.X
* Updated unit tests elsewhere in the code where necessary

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1965

### Testing
Unit tests; manually running RFS against a Snapshot created in ES 5 and upgraded to ES 6

### Check List

### DIFF
--- a/RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
+++ b/RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
@@ -160,7 +160,6 @@ public class LuceneDocumentsReader {
             BytesRef sourceBytes = null;
             try {
                 for (var field : document.getFields()) {
-                    log.atDebug().setMessage("Field: {}").addArgument(field.name()).log();
                     String fieldName = field.name();
                     switch (fieldName) {
                         case "_id": {
--- a/RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
+++ b/RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
@@ -2,9 +2,9 @@ package com.rfs.common;
 
 import java.io.IOException;
 import java.nio.file.Path;
+import java.util.List;
 import java.util.concurrent.Callable;
 import java.util.function.Function;
-import java.util.List;
 
 import org.apache.lucene.document.Document;
 import org.apache.lucene.index.DirectoryReader;
@@ -14,6 +14,7 @@ import org.apache.lucene.index.LeafReaderContext;
 import org.apache.lucene.index.SoftDeletesDirectoryReaderWrapper;
 import org.apache.lucene.store.FSDirectory;
 import org.apache.lucene.util.BytesRef;
+
 import org.opensearch.migrations.cluster.ClusterSnapshotReader;
 
 import lombok.Lombok;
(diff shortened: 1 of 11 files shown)


RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
```diff
@@ -135,18 +153,44 @@ protected DirectoryReader wrapReader(DirectoryReader reader, boolean softDeletes
...
             Document document = reader.document(docId);
-            BytesRef sourceBytes = document.getBinaryValue("_source");
-            String id;
+            String id = null;
+            BytesRef sourceBytes = null;
             try {
-                var idValue = document.getBinaryValue("_id");
-                if (idValue == null) {
+                for (var field : document.getFields()) {
+                    log.atDebug().setMessage("Field: {}").addArgument(field.name()).log();
```
peternied, line 163: This seems like trace level logging, is it still needed?  If you want to keep it, lets bump to trace level and could you include the index name / doc id with the field.

peternied (commented):
Nice work, couple of minor callouts! Can we updated the end to end test with a ES 5 cluster?

RFS/src/main/java/com/rfs/common/LuceneDocumentsReader.java
chelma: Oops, that was for debugging.  Will remove.

chelma:
> Nice work, couple of minor callouts! Can we updated the end to end test with a ES 5 cluster?

I'd love to, but - we don't support ES 5 source clusters; we support ES 6 source clusters which contain indices created on ES 5 😛

peternied:
> I'd love to, but - we don't support ES 5 source clusters; we support ES 6 source clusters which contain indices created on ES 5

Then could we have one of those tests then?  The unit test is good for coverage metrics, but if the `backward-codecs` in a patch release we would have no way of knowing.

chelma:
> Then could we have one of those tests then? The unit test is good for coverage metrics, but if the `backward-codecs` is busted in a patch release we would have no way of knowing.

Yeah, that's the test I added 😉

Check `LuceneDocumentsReaderTest > ReadDocuments_ES5_Origin_AsExpected()`

peternied:
> Check `LuceneDocumentsReaderTest > ReadDocuments_ES5_Origin_AsExpected()`

Thanks for the pointer - I missed the relevance of the `TestResources.SNAPSHOT_ES_6_8_MERGED` line that was used the prepared test resources - thanks!

RFS/src/test/java/com/rfs/common/LuceneDocumentsReaderTest.java
peternied, line 1: Nit: To prevent confusion about how these tests work, I'd recommend moving this class under the 'integration' package.

peternied (approved):

PR #961, opened 2024-09-17
Add inteligentTiering for logs efs

AndreKurait:
### Description

Logs in EFS are infrequently accessed and cost savings are big for IA vs Standard tier. 

10TB in EFS Standard for a month cost $3000, whereas with this policy will cost $450 (Assuming no reads for files after 1 day, cost slightly higher based on reads for files in IA)

* Category: Enhancement
* Why these changes are required? Reduce cost for tuples as it is currently one of the most expensive aspects of capture and replay
* What is the old behavior before changes and new behavior after changes? EFS provisioned with tiering.

### Issues Resolved
N/A

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
@@ -1,6 +1,6 @@
 import {RemovalPolicy, Stack} from "aws-cdk-lib";
 import {IVpc, Port, SecurityGroup, SubnetFilter, SubnetType} from "aws-cdk-lib/aws-ec2";
-import {FileSystem} from 'aws-cdk-lib/aws-efs';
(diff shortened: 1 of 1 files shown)


peternied (approved):
This is interesting, I like the idea of cost optimization, but if we are triggering the right dials/knobs.

For example, if someone is doing a detailed analysis of tuples stored in those logs they could iterated over the collection gigabytes of data 100x of times without knowing the impact.  I think this breaks down to a lot of 'it depends' use cases.

Do we have a break down of where the costs are that we could relate this to?  I'd love to be able to refer to a cost break down and also have an accounting for the different changes we've made to raise/lower cost.  I feel like 6 months from now decisions like this could be lost.

@AndreKurait I'm curious what you think about tracking cost management decisions?  

That said lets keep moving forward by merging this PR

PR #964, opened 2024-09-17
Design document for the Capture Proxy and Replayer.

gregschohn:
### Description

* Category Documentation

### Issues Resolved
part of https://opensearch.atlassian.net/browse/MIGRATIONS-1986

### Testing
No testing

### Check List


docs/TrafficCaptureAndReplayDesign.md
```diff
@@ -0,0 +1,271 @@
+# TrafficCapture and Replay
```
peternied, line 1:
```suggestion
# Traffic Capture and Replay Design
```
I think this document might be better classified as an overview instead of a design.

peternied (commented):
Did you need a detailed review of this content? It seems like a majority of of is coming from the aws solution document and so this is more about moving the doc rather than fine-grain review.

sumobrian (approved):

docs/TrafficCaptureAndReplayDesign.md
gregschohn: Made the partial change - since there's a question of whether its an Overview or Design, I'll leave the filename speak for itself (the ALB splitter doc doesn't mention what kind of doc it is within the title, so I didn't bother w/ that for consistency).

PR #966, opened 2024-09-18
Fixed SonarQube bugs in coreUtilities

AndreKurait:
### Description
Fixed SonarQube bugs in coreUtilities
* Category: Maintenance
* Why these changes are required? Improve maintainability

### Issues Resolved
[MIGRATIONS-2005](https://opensearch.atlassian.net/browse/MIGRATIONS-2005)

### Testing
GHA

### Check List

### DIFF
--- a/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
+++ b/coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
@@ -177,7 +177,6 @@ public class RootOtelContext implements IRootOtelContext {
         String spanName,
         Stream<Span> linkedSpans
     ) {
-        assert forScope.getCurrentSpan() == null;
         var forEnclosingScope = forScope.getEnclosingScope();
         var parentSpan = forEnclosingScope == null ? null : forEnclosingScope.getCurrentSpan();
         var spanBuilder = getOpenTelemetry().getTracer(scopeName).spanBuilder(spanName);
--- a/coreUtilities/src/main/java/org/opensearch/migrations/utils/ProcessHelpers.java
+++ b/coreUtilities/src/main/java/org/opensearch/migrations/utils/ProcessHelpers.java
@@ -7,7 +7,9 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class ProcessHelpers {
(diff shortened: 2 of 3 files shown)


coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
```diff
@@ -177,7 +177,6 @@ private static Span buildSpanWithParent(SpanBuilder builder, Span parentSpan, St
         String spanName,
         Stream<Span> linkedSpans
     ) {
-        assert forScope.getCurrentSpan() == null;
```
gregschohn, line 180: why are we removing asserts?

gregschohn (changes requested):

coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: For what it's worth, I've never been happy w/ how I initialize the spans this way - but this assert has fired a number of times during development.  It's been super valuable in the past.  If you'd like to explain it more, please feel free to w/ the : "..." syntax.

AndreKurait: This seems redundant because the getter is annotated with NonNull

```
    @NonNull
    Span getCurrentSpan();
```

coreUtilities/src/main/java/org/opensearch/migrations/utils/ProcessHelpers.java
```diff
@@ -7,7 +7,9 @@
 
 @Slf4j
 public class ProcessHelpers {
-    private static final String DEFAULT_NODE_ID = "generated_" + UUID.randomUUID().toString();
+    private static final String DEFAULT_NODE_ID = "generated_" + UUID.randomUUID();
+
+    private ProcessHelpers() {}
```
peternied, line 12: nit: Use `@UtilityClass` instead

peternied (approved):

coreUtilities/src/main/java/org/opensearch/migrations/tracing/RootOtelContext.java
gregschohn: Good catch.  I'm fine removing it.

PR #967, opened 2024-09-18
Parse config file w/ jackson and set better TLS defaults

gregschohn:
### Description

* Category  Bug fix
* Why these changes are required? Better security posture.  'Secure by default'
* What is the old behavior before changes and new behavior after changes? Users can now set TLS protocols and any other HTTPS settings that require lists or nested data structures.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2014

### Testing
manual testing with `curl -v --insecure https://localhost:9200 --tls-max 1.2` for various values of tls-max.  As configured w/ this PR, the proxy container supports 1.2 and 1.3 only.

### Check List


TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -191,19 +195,20 @@ static Parameters parseArgs(String[] args) {
...
-        builder.put("path.home", configParentDirStr);
-        return builder.build();
+        var httpsSettings =
+            objectMapper.convertValue(configMap, new TypeReference<Map<String, Object>>(){})
+                .entrySet().stream()
+                .filter(kvp -> kvp.getKey().startsWith(HTTPS_CONFIG_PREFIX))
+                .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        return Settings.builder().loadFromMap(httpsSettings)
+            .put(SSLConfigConstants.SECURITY_SSL_TRANSPORT_ENABLED, false)
```
mikaylathompson, line 209: This looks like we're always setting TLS to false -- is that true? Why?

TrafficCapture/dockerSolution/src/main/docker/elasticsearchWithSearchGuard/Dockerfile
```diff
@@ -19,7 +19,8 @@ RUN sed 's/searchguard/plugins.security/g' $ELASTIC_SEARCH_CONFIG_FILE |  \
 # but maybe not for a demo to show individual steps
 RUN /root/enableTlsConfig.sh $ELASTIC_SEARCH_CONFIG_FILE
 # Alter this config line to either enable(searchguard.disabled: false) or disable(searchguard.disabled: true) HTTP auth
-RUN echo "searchguard.disabled: false" >> $ELASTIC_SEARCH_CONFIG_FILE
+RUN echo -n "searchguard.disabled: false" >> $ELASTIC_SEARCH_CONFIG_FILE  && \
+    echo -n "plugins.security.ssl.http.enabled_protocols: ['TLSv1.2', 'TLSv1.3']" >> $PROXY_TLS_CONFIG_FILE
```
mikaylathompson, line 23: I see that this sets the allowable protocols to `TLSv1.2` and `TLSv1.3`, but this dockerfile only looks relevant for the `elasticsearchWithSearchGuard` stack, not the capture proxy in general. Is that the full scope of the concern here?

mikaylathompson:
Nit(-ish):
> ### Issues Resolved
>
> https://github.com/opensearch-project/opensearch-migrations/pull/967

just links back to itself -- do we have a real issue for this?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: Because that's for the 9300 port setup stuff, which the proxy doesn't use.  I'll add a comment to explain.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -191,19 +195,20 @@ static Parameters parseArgs(String[] args) {
 
     @SneakyThrows
     protected static Settings getSettings(@NonNull String configFile) {
```
peternied, line 197: Can we add a unit test since we need coverage + we are touching previously untested coded that apparently had bugs?

TrafficCapture/dockerSolution/src/main/docker/elasticsearchWithSearchGuard/Dockerfile
gregschohn: @AndreKurait - are we using this as a base image for the proxy that we deploy?

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
gregschohn: Parsing tests added.  Notice that we get some tests here from the E2E tests.  We just weren't setting compound configurations in the proxy invocation that the E2E scripts were using.  Once I turned that on, things went sideways.

mikaylathompson (approved):
Would like to see confirmation from @AndreKurait  on this point (https://github.com/opensearch-project/opensearch-migrations/pull/967/files#r1765741424), otherwise LGTM

peternied (approved):

PR #976, opened 2024-09-18
Remove unused snapshot metadata code

peternied:
### Description
The snapshot metadata and associated factory are not showing up in any code coverage reports and based on the removal not impacting any other files this looks like a safe change that our CI suite will verify.

### Issues Resolved
- Related https://opensearch.atlassian.net/browse/MIGRATIONS-2004

### Check List


chelma (changes requested):
I expect we'll use these in the future and would prefer to keep them.  Knowing whether a snapshot was successful before we start parsing it is something we should probably be checking, for example, and that's something these classes provide.

peternied:
> I expect we'll use these in the future and would prefer to keep them. Knowing whether a snapshot was successful before we start parsing it is something we should probably be checking, for example, and that's something these classes provide.

@chelma While I appreciate that they could be useful, I don't see value in unused code around.  These changes are still in git history and can be revived when needed.

If you would prefer we keep them, are you willing to write test for these files?

gregschohn:
To Peter's inquiry, writing tests w/out knowing the context that these files are used won't put us into a better place.
I've opened up a new [Jira task](https://opensearch.atlassian.net/browse/MIGRATIONS-2022) to resurrect the spirit of this code.  That task cites this PR so that future implementors can extend this code and integrate it fully into the codebase.

gregschohn (approved):
See https://opensearch.atlassian.net/browse/MIGRATIONS-2022 for the tracking ticket to bring this code back.

chelma:
> To Peter's inquiry, writing tests w/out knowing the context that these files are used won't put us into a better place. I've opened up a new [Jira task](https://opensearch.atlassian.net/browse/MIGRATIONS-2022) to resurrect the spirit of this code. That task cites this PR so that future implementors can extend this code and integrate it fully into the codebase.

Thanks for making the follow-up task, @gregschohn.  Happy to revisit this code later as long as we're not losing track of the capability.  I agree w/ your take that it needs some work to bring it up to our current standards so just adding some tests probably isn't the right answer.

PR #977, opened 2024-09-19
Eliminate some code smells in the replayer

gregschohn:
### Description

SonarLinting for the replayer

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2009

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -346,20 +347,20 @@ private static void setupShutdownHookForReplayer(TrafficReplayerTopLevel tr) {
...
                     + "The logging facilities may also be shutting down concurrently, "
-                    + "resulting in missing logs messages.";
-                log.atWarn().setMessage(beforeMsg).log();
-                System.err.println(beforeMsg);
-            }
+                    + "resulting in missing logs messages.")
+                .ifPresent(beforeMsg -> {
+                    log.atWarn().setMessage(beforeMsg).log();
+                    System.err.println(beforeMsg);
+                });
```
peternied, line 356: I'm not clear why we are using optional on a string that will never be null, is removing the delete the curly brackets enough?

peternied: I've created a PR [1] that does this more cleanly, what do you think?

- [1] https://github.com/gregschohn/opensearch-migrations/pull/7

peternied (approved):
For that outstanding comment, up to you shouldn't be blocking

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: I didn't want the scope of the variables to be commingled.  Basically, I'm using a lambda via Optional to scope them.

peternied: Thanks for the comment Greg. I see where you’re coming from with the use of Optional to control scoping, but I’m not sure that addressing variable ‘commingling’ is critical here, given the method’s simplicity.

Since the new version resolves some sonar lint issues, I’m good with moving forward and merging this. If we want to revisit scoping or clarify the rationale down the road I’m open to discussing it.

PR #978, opened 2024-09-19
More SonarQube fixes for the RFS version* code

chelma:
### Description
* More SonarQube fixes for the RFS version* code

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-2006

### Testing
* Ran the local test suites

### Check List


RFS/src/main/java/com/rfs/version_os_2_11/GlobalMetadataCreator_OS_2_11.java
```diff
@@ -144,24 +144,21 @@ private List<String> createTemplates(
 
         templatesToCreate.forEach((templateName, templateBody) -> {
             log.info("Creating {}: {}", templateType, templateName);
-            switch (mode) {
```
peternied, line 147: This is fine - but IMO this was a sonar lint rule that does the wrong thing.

When you use switch with a enum without `default`, if any additional values are added to the enum you get a compile time failure which tells you this block of code needs to support all potential paths.  When switching to if/else no such warning could happen.

Why doesn't this really matter - I doubt we will ever add another MigrationMode.

chelma: That makes sense; good Java tip!  I was wondering why the code was structured this way.

PR #1009, opened 2024-09-24
Add indexAllowlist to snapshot create command

mikaylathompson:
### Description
Add an `--index-allowlist` flag to the `CreateSnapshot` command, which passes the list of fields to the snapshot creation api call.

### Issues Resolved

### Testing
Snapshot tests added.

### Check List


CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    @Test
+    public void testSnapshotCreateWithIndexAllowlist() throws Exception {
+        var snapshotName = "my_snap";
+        var indexAllowlist = List.of("allowed_index_1", "allowed_index_2");
+
+        final FullHttpRequest[] createSnapshotRequest = new FullHttpRequest[1];
+        final String[] createSnapshotRequestContent = {""};
```
mikaylathompson, line 83: These arrays were IntelliJ's suggestion to fix the issue that you can't use a non-final in a lambda. Open to a better suggestion if this is not the right way to resolve it.

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
```diff
@@ -59,6 +61,10 @@ public static class Args {
             "--s3-role-arn" }, required = false, description = "The role ARN the cluster will assume to write a snapshot to S3")
         public String s3RoleArn;
 
+        @Parameter(names = {
+            "--index-allowlist" }, required = false, description = "A comma separated list of indices to include in the snapshot. If not provided, all indices are included.")
```
gregschohn, line 65: nit: It would be nice for this to be can you format these to put each annotation parameter its own line (like what's done for CaptureProxy & TrafficReplayer).  Like this
```
        @Parameter(
            required = false,
            names = {"--insecure" },
            arity = 0, description = "Do not check the server's certificate")
        boolean allowInsecureConnections;
```

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
gregschohn: Yeah - sorry.  This is a java-ism.  It makes me realize why the capture specifiers in C++ `[=foo,&bar]` really aren't that bad!

You can also do an AtomicReference/AtomicInteger, etc.  I usually do the AtomicReference thing because it seems more clear to me.  Our codebase isn't to the level of consistency where it matters, especially for test cases!

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
```diff
@@ -0,0 +1,115 @@
...
+            String registerRepoRequestContent = capturedRequestList.get(0).getRight();
+            FullHttpRequest createSnapshotRequest = capturedRequestList.get(1).getLeft();
+            String createSnapshotRequestContent = capturedRequestList.get(1).getRight();
+
+            assert registerRepoRequest.uri().equals("/_snapshot/migration_assistant_repo");
+            assert registerRepoRequest.method().toString().equals("PUT");
+            assert registerRepoRequestContent.equals("{\"type\":\"s3\",\"settings\":{\"bucket\":\"new-bucket\",\"region\":\"us-east-2\",\"base_path\":null}}");
+            assert createSnapshotRequest.uri().equals("/_snapshot/migration_assistant_repo/" + snapshotName);
+            assert createSnapshotRequest.method().toString().equals("PUT");
+            assert createSnapshotRequestContent.equals("{\"indices\":\"_all\",\"ignore_unavailable\":true,\"include_global_state\":true}");
```
gregschohn, line 73: these are java language assertions.  If the unit tests were run w/out -ea, these wouldn't be evaluated.  If one fails, we probably won't always get the same support that we get in all of our tools to give more hints about what the issues were.

Import Assertions and use `Assertions.assertEquals(EXPECTED, ACTUAL, OPTIONAL_AND_RARELY_USED_MESSAGE)` and the other many helpers w/in the Assertions class.  Just find "Assertions." for many examples.

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
```diff
@@ -0,0 +1,115 @@
+package org.opensearch.migrations;
+
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.lang3.tuple.Pair;
```
gregschohn, line 9: this is a bit risky since you're (I think) depending on a transitive dependency.  
For a test case, it's reasonable to take on extra dependencies that you normally wouldn't want to take on.
That said, the lack of a Pair<> type is well-known for Java devs.  Normally, we just do Map.Entry<> for the type and instantiate `new AbstractMap.SimpleEntry<>(V1, V2)`.  For test code, I'm fine w/ another dependency on this or vavr.io or whatever, it will probably eventually be refactored away to keep dependency graphs smaller.

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
AndreKurait, line 65: Do we want to align this with document migration which excludes system indicies?

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
AndreKurait: an AtomicReference/AtomicString would make more sense here

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
```diff
@@ -0,0 +1,115 @@
...
+            SnapshotRunner.run(snapshotCreator);
+
+            FullHttpRequest registerRepoRequest = capturedRequestList.get(0).getLeft();
+            String registerRepoRequestContent = capturedRequestList.get(0).getRight();
+            FullHttpRequest createSnapshotRequest = capturedRequestList.get(1).getLeft();
+            String createSnapshotRequestContent = capturedRequestList.get(1).getRight();
+
+            assert registerRepoRequest.uri().equals("/_snapshot/migration_assistant_repo");
+            assert registerRepoRequest.method().toString().equals("PUT");
+            assert registerRepoRequestContent.equals("{\"type\":\"s3\",\"settings\":{\"bucket\":\"new-bucket\",\"region\":\"us-east-2\",\"base_path\":null}}");
```
gregschohn, line 70: is base_path = null right?
You might want to normalize the json in the test and from the java class so that if one pretty-prints, you aren't in trouble with a difference.
hamcrest lets you do fancy deep mappings too - but it's another thing to learn and therefore has been a barrier to entry for me (others use it a lot)

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
mikaylathompson: I lean towards including them by default. I'm fine with not transferring them by default, but if a user ends up wanting to move some/all of them, it would be very annoying to discover that the snapshot didn't include them.

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
mikaylathompson: Okay, cool, thanks -- updated this to AtomicReferences

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
mikaylathompson: This is so much better!

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
```diff
@@ -18,8 +22,12 @@
 import org.opensearch.migrations.testutils.SimpleNettyHttpServer;
 
 import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http.HttpMethod;
 import lombok.extern.slf4j.Slf4j;
 
+import static org.hamcrest.CoreMatchers.containsString;
+import static org.hamcrest.MatcherAssert.assertThat;
```
gregschohn, line 29: nit: I hate all static imports (others disagree though).

gregschohn (approved):

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
gregschohn: I like the indenting job that you did.  Thanks

PR #1014, opened 2024-09-25
Updated the ES 7.X transformation code to handle arbitrary type names

chelma:
### Description
* Per user bug report, we were not handling transformation of index mappings correctly.  The issue appears to be that, while ES 7.X doesn't make it default to specify type names for mappings in an index, it is possible, and our code did not gracefully handle that situation.  The solution was to accept all type names.
* Added a new ES 7.10 Snapshot and unit tests to exercise the particular failure mode

### Issues Resolved
* N/A

### Testing
* Local testing w/ Docker containers; unit tests.
* Loaded the new test snapshot into a EC2 source cluster and migrated it to an AOS target cluster using the Migration Console:

```
(.venv) bash-5.2# console metadata migrate --component-template-allowlist fwc_settings,fwc_mappings --index-template-allowlist fwc_template
Command line arguments: --otel-collector-endpoint http://localhost:4317 migrate --snapshot-name rfs-snapshot --target-host https://vpc-os-cluster-aws-integ-a2z3xm6v3nzxcgll5oldkej7bi.us-east-2.es.amazonaws.com:443 --min-replicas 1 --s3-local-dir /tmp/migration-rfs-snapshot-n9hfh5gp --s3-repo-uri s3://migration-artifacts-672500112149-aws-integ-us-east-2/rfs-snapshot-repo --s3-region us-east-2 --component-template-allowlist fwc_settings,fwc_mappings --index-template-allowlist fwc_template

Starting Metadata Migration
Clusters:
   Source:
      Snapshot: ELASTICSEARCH 7.10.0 S3Repo(s3RepoUri=org.opensearch.migrations.bulkload.common.S3Uri@687eb455)

   Target:
      Remote Cluster: OPENSEARCH 2.11.0 ConnectionContext(uri=https://vpc-os-cluster-aws-integ-a2z3xm6v3nzxcgll5oldkej7bi.us-east-2.es.amazonaws.com:443, protocol=HTTPS, insecure=false, compressionSupported=false)


Migrated Items:
   Index Templates:
      fwc_template

   Component Templates:
      fwc_mappings, fwc_settings

   Indexes:
      bwc_index_1, fwc_index_1

   Aliases:
      bwc_alias, fwc_alias


Results:
   0 issue(s) detected

Command executed successfully
```

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
```diff
@@ -78,17 +78,24 @@ public static void removeIntermediateMappingsLevels(ObjectNode root) {
...
-            return (ObjectNode) actualMappingsRoot.get("doc").deepCopy();
-        } else if (actualMappingsRoot.has("audit_message")) {
-            return (ObjectNode) actualMappingsRoot.get("audit_message").deepCopy();
-        } else {
-            throw new IllegalArgumentException("Mappings root does not contain one of the expected keys");
+        ObjectNode actualMappingCopy = (ObjectNode) actualMappingsRoot.get(actualMappingsRoot.fieldNames().next()).deepCopy();
+
+        // Confirm the thing we retrieved has the shape we expect
+        if (!actualMappingCopy.has("properties")) {
+            throw new IllegalArgumentException("Mappings object does not contain the expected properties key");
```
mikaylathompson, line 95: Okay, I'm not sure I understand exactly what's happening here and how it maps to the customer case.

My understanding is that their data structure looked like:
```
  {
    "mappings" : {
      "properties" : {
        "account_id" : ...,
        "next_field_here":...
      }
    }
  }
```

So `mappingsRoot` is the `mappings` item. This code seems to say that there needs to be a layer (`actualMappingsRoot`) between `mappings` and `properties`. But the customer JSON doesn't have a layer there. Am I missing understanding something here?

AndreKurait: There's two things going on here, 1. in the API, when `?include_type_names=false` which is the default, what you showed is returned In the API, when `?include_type_names=true`, then the type key is included, in the snapshot, this is always included (plus wrapping it in an array)

@chelma, IIRC, @peternied demo'd functionality to migrate metadata from a cluster instead of a snapshot. can you confirm that works with this?

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
```diff
@@ -78,17 +78,24 @@ public static void removeIntermediateMappingsLevels(ObjectNode root) {
...
     // Extract the mappings from their single-member list, will start like:
     // [{"_doc":{"properties":{"address":{"type":"text"}}}}]
+    //
+    // There should only be a single member in the list because multi-type mappings were deprecated in ES 6.X and
+    // removed in ES 7.X.  This list structure appears to be a holdover from previous versions of Elasticsearch.
+    // The exact name of the type can be arbitrarily chosen by the user; the default is _doc.  We need to extract
+    // the mappings from beneath this intermediate key regardless of what it is named.
     public static ObjectNode getMappingsFromBeneathIntermediate(ArrayNode mappingsRoot) {
+        if (mappingsRoot.size() != 1) {
+            throw new IllegalArgumentException("Mappings root does not contain a single member list");
```
AndreKurait, line 88: Please include the number found in the message here

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
```diff
@@ -78,17 +78,20 @@ public static void removeIntermediateMappingsLevels(ObjectNode root) {
...
-            return (ObjectNode) actualMappingsRoot.get("doc").deepCopy();
-        } else if (actualMappingsRoot.has("audit_message")) {
-            return (ObjectNode) actualMappingsRoot.get("audit_message").deepCopy();
-        } else {
-            throw new IllegalArgumentException("Mappings root does not contain one of the expected keys");
+
+        if (mappingsRoot.size() != 1) {
+            throw new IllegalArgumentException("Mappings root does not contain a single member list");
         }
+        actualMappingsRoot.size();
```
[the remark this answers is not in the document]
chelma: Nothing; was a bad line of code that I pruned in a later commit.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
chelma: @AndreKurait Good call, lemme look into that.

chelma: > @chelma, IIRC, @peternied demo'd functionality to migrate metadata from a cluster instead of a snapshot. can you confirm that works with this?

Great catch; this behavior is still broken.  Addressing it will also address Mikayla's concern about "what if there isn't an intermediary".

mikaylathompson (approved):

PR #1016, opened 2024-09-25
Simplify CDK ExtraArg processing with jcommander overrides

AndreKurait:
### Description
Simplify CDK ExtraArg processing with jcommander overrides.

Now, given two parameters `--s3-region us-east-2 --s3-region us-west-2`, jcommander will use the latter one.

Fixed places where parameter values which may have spaces weren't wrapped in quotes.

Removed cdk dependency on yargs

Fixed an issue where List arguments would split on `,` incorrectly e.g. `--setHeader name "value1,value2"`

Note: There's two use cases missing from this PR. 1. being able to remove/set an argument to null. 2. correct behavior with extraArgs in RFS for arguments placed in the ENV variables

* Category: Bug Fix
* Why these changes are required? Currently --setHeader in extraArgs is broken along with special characters in the values.
* What is the old behavior before changes and new behavior after changes? CaptureProxy could crash on  startup with otherwise valid arguments.

### Issues Resolved

### Testing

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
@@ -33,6 +33,7 @@ import org.apache.kafka.clients.producer.ProducerConfig;
 import org.apache.kafka.common.config.SaslConfigs;
 
 import org.opensearch.common.settings.Settings;
+import org.opensearch.migrations.jcommander.NoSplitter;
 import org.opensearch.migrations.tracing.ActiveContextTracker;
 import org.opensearch.migrations.tracing.ActiveContextTrackerByActivityType;
 import org.opensearch.migrations.tracing.CompositeContextTracker;
@@ -158,12 +159,14 @@ public class CaptureProxy {
         public String otelCollectorEndpoint;
         @Parameter(required = false,
             names = "--setHeader",
+            splitter = NoSplitter.class,
             arity = 2,
             description = "[header-name header-value] Set an HTTP header (first argument) with to the specified value" +
                 " (second argument).  Any existing headers with that name will be removed.")
         public List<String> headerOverrides = new ArrayList<>();
         @Parameter(required = false,
             names = "--suppressCaptureForHeaderMatch",
+            splitter = NoSplitter.class,
             arity = 2,
             description = "The header name (which will be interpreted in a case-insensitive manner) and a regex "
                 + "pattern.  When the incoming request has a header that matches the regex, it will be passed "
--- a/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
@@ -4,78 +4,7 @@ import {CpuArchitecture} from "aws-cdk-lib/aws-ecs";
 import {RemovalPolicy, Stack} from "aws-cdk-lib";
 import { IStringParameter, StringParameter } from "aws-cdk-lib/aws-ssm";
 import * as forge from 'node-forge';
-import * as yargs from 'yargs';
 import { ClusterYaml } from "./migration-services-yaml";
-
-
-// parseAndMergeArgs, see @common-utilities.test.ts for an example of different cases
-export function parseAndMergeArgs(baseCommand: string, extraArgs?: string): string {
-    if (!extraArgs) {
-        return baseCommand;
-    }
-
-    // Extract command prefix
-    const commandPrefix = baseCommand.substring(0, baseCommand.indexOf('--')).trim();
-    const baseArgs = baseCommand.substring(baseCommand.indexOf('--'));
-
-    // Parse base command
-    const baseYargsConfig = {
-        parserConfiguration: {
-            'camel-case-expansion': false,
-            'boolean-negation': false,
-        }
-    };
-
-    const baseArgv = yargs(baseArgs)
-        .parserConfiguration(baseYargsConfig.parserConfiguration)
-        .parse();
-
-    // Parse extra args if provided
-    const extraYargsConfig = {
-        parserConfiguration: {
-            'camel-case-expansion': false,
-            'boolean-negation': true,
-        }
-    };
-
-    const extraArgv = extraArgs
-        ? yargs(extraArgs.split(' '))
-            .parserConfiguration(extraYargsConfig.parserConfiguration)
-            .parse()
-        : {};
-
-    // Merge arguments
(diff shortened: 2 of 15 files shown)


CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
```diff
@@ -74,7 +74,10 @@ public static class S3RepoInfo {
 
     public static void main(String[] args) throws Exception {
         Args arguments = new Args();
-        JCommander jCommander = JCommander.newBuilder().addObject(arguments).build();
+        JCommander jCommander = JCommander.newBuilder()
+            .allowParameterOverwriting(true)
```
gregschohn, line 78: I don't like this idea at all - we should throw.  As a user of these programs, I sometimes have long invocations and I might hop to the beginning or end of the list because it's convenient.  Allowing the user to put two in but then just use one means that you're ignoring something that the user tried to pass in.

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -159,12 +160,14 @@ public static class Parameters {
...
             names = "--setHeader",
             arity = 2,
+            splitter = NoSplitter.class,
             description = "[header-name header-value] Set an HTTP header (first argument) with to the specified value" +
                 " (second argument).  Any existing headers with that name will be removed.")
         public List<String> headerOverrides = new ArrayList<>();
         @Parameter(required = false,
             names = "--suppressCaptureForHeaderMatch",
             arity = 2,
+            splitter = NoSplitter.class,
```
gregschohn, line 170: Thanks.  I was unaware of the comma-splitting feature of JCommander!

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -4,79 +4,8 @@ import {CpuArchitecture} from "aws-cdk-lib/aws-ecs";
 import {RemovalPolicy, Stack} from "aws-cdk-lib";
 import { IStringParameter, StringParameter } from "aws-cdk-lib/aws-ssm";
 import * as forge from 'node-forge';
-import * as yargs from 'yargs';
 import { ClusterYaml } from "./migration-services-yaml";
 
-
-// parseAndMergeArgs, see @common-utilities.test.ts for an example of different cases
-export function parseAndMergeArgs(baseCommand: string, extraArgs?: string): string {
```
gregschohn, line 12: I do love deleting this though!

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -69,11 +68,12 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
         rfsCommand = props.clusterAuthDetails.sigv4 ? rfsCommand.concat(`--target-aws-service-signing-name ${props.clusterAuthDetails.sigv4.serviceSigningName} --target-aws-region ${props.clusterAuthDetails.sigv4.region}`) : rfsCommand
         rfsCommand = props.otelCollectorEnabled ? rfsCommand.concat(` --otel-collector-endpoint ${OtelCollectorSidecar.getOtelLocalhostEndpoint()}`) : rfsCommand
-        rfsCommand = props.sourceClusterVersion ? rfsCommand.concat(` --source-version ${props.sourceClusterVersion}`) : rfsCommand
-        rfsCommand = parseAndMergeArgs(rfsCommand, props.extraArgs);
+        rfsCommand = props.sourceClusterVersion ? rfsCommand.concat(` --source-version \"${props.sourceClusterVersion}\"`) : rfsCommand
+        // TODO: This approach with extraArgs may not work with the entryPoint env arg processing that is occurring here. https://opensearch.atlassian.net/browse/MIGRATIONS-2025
+        rfsCommand = props.extraArgs ? rfsCommand.concat(` ${props.extraArgs}`) : rfsCommand
```
gregschohn, line 76: This feels like the place to be responsible for overrides vs not.
```
rfsCommand += formatArgsIfNotOverridden(extraArgs, "--s3-region", "${this.region}");
...
```

```
appendIfNotOverridden(argsList, argName, defaultValues) {
  return argsList.contains(argName) ? argsList.getValue(argName) : defaultValues;
}
```
You can keep todays space delimited list for extra args and have a helper parse that (based upon '--' prefixed values to parse it into a map - and share each of these across all of the cdk).  You could also change the incoming type, or make it optionally either type (if a string, parse; if not, just use it as a map).

As a rule, we probably should ONLY be consing arguments that are required so that a user never needs to remove something.  Instead, they would just alter their configurations to NOT include something.

gregschohn (commented):
I love making command line parsing easier.  I don't like doing it at the expense of the UX for users using the command line directly.  I use the command line directly many times per day and usually I cannot see all of the args that I pass at once.

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
AndreKurait: Removed

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
AndreKurait: I've updated the logic to handle only including base arguments when not provided in the cdk

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -4,87 +4,66 @@ import {CpuArchitecture} from "aws-cdk-lib/aws-ecs";
...
+        // If not present, append the argument and value (only append value if it exists)
+        baseCommand = value !== null ? baseCommand.concat(" ", arg, " ", value) : baseCommand.concat(" ", arg);
     }
+    return baseCommand;
+}
 
-    // Extract command prefix
-    const commandPrefix = baseCommand.substring(0, baseCommand.indexOf('--')).trim();
-    const baseArgs = baseCommand.substring(baseCommand.indexOf('--'));
+export function parseArgsToDict(argString: string | undefined): Record<string, string[]> {
```
AndreKurait, line 32: We're not using the values now, only the keys (except in the unit tests), but i thought it could be useful to keep this in here in case we use it.

I had originally put it in for the RFS Env Variable stuff, but since we don't support overriding the secretArn, we didn't need to use it

AndreKurait:
Closing due to error - 2024-09-26 18:55:28,638 -- Report creating failed: {"message":"Token required because branch is protected"}

Tracking in https://github.com/codecov/codecov-action/issues/1580

PR #1019, opened 2024-09-26
Update argument parsing to remove dependency on yargs and make more robust

AndreKurait:
### Description
Simplify CDK ExtraArg processing to remove yargs and make more robust.

Fixed places where parameter values which may have spaces weren't wrapped in quotes.

Removed cdk dependency on yargs

Fixed an issue where List arguments would split on `,` incorrectly e.g. `--setHeader name "value1,value2"`
Fixed an issue where ExtraArgs wouldn't correctly handle multi arity commands e.g. `--setHeader name value`

Fixed a bug in how user supplied plaintext password was being supplied to the replayer

* Category: Bug Fix
* Why these changes are required? Currently --setHeader in extraArgs is broken along with special characters in the values.
* What is the old behavior before changes and new behavior after changes? CaptureProxy could crash on  startup with otherwise valid arguments.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2025

### Testing
Unit testing, AWS Testing

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
@@ -6,10 +6,10 @@ import { IStringParameter, StringParameter } from "aws-cdk-lib/aws-ssm";
 import * as forge from 'node-forge';
 import { ClusterYaml } from "./migration-services-yaml";
 
-export function getTargetPasswordAccessPolicy(targetPasswordSecretArn: string): PolicyStatement {
+export function getSecretAccessPolicy(secretArn: string): PolicyStatement {
     return new PolicyStatement({
         effect: Effect.ALLOW,
-        resources: [targetPasswordSecretArn],
+        resources: [secretArn],
         actions: [
             "secretsmanager:GetSecretValue"
         ]
--- a/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
@@ -4,78 +4,7 @@ import {CpuArchitecture} from "aws-cdk-lib/aws-ecs";
 import {RemovalPolicy, Stack} from "aws-cdk-lib";
 import { IStringParameter, StringParameter } from "aws-cdk-lib/aws-ssm";
 import * as forge from 'node-forge';
-import * as yargs from 'yargs';
 import { ClusterYaml } from "./migration-services-yaml";
-
-
-// parseAndMergeArgs, see @common-utilities.test.ts for an example of different cases
-export function parseAndMergeArgs(baseCommand: string, extraArgs?: string): string {
-    if (!extraArgs) {
-        return baseCommand;
-    }
-
-    // Extract command prefix
-    const commandPrefix = baseCommand.substring(0, baseCommand.indexOf('--')).trim();
-    const baseArgs = baseCommand.substring(baseCommand.indexOf('--'));
-
-    // Parse base command
-    const baseYargsConfig = {
-        parserConfiguration: {
-            'camel-case-expansion': false,
-            'boolean-negation': false,
-        }
-    };
-
-    const baseArgv = yargs(baseArgs)
-        .parserConfiguration(baseYargsConfig.parserConfiguration)
-        .parse();
-
-    // Parse extra args if provided
-    const extraYargsConfig = {
-        parserConfiguration: {
-            'camel-case-expansion': false,
-            'boolean-negation': true,
-        }
-    };
-
-    const extraArgv = extraArgs
-        ? yargs(extraArgs.split(' '))
-            .parserConfiguration(extraYargsConfig.parserConfiguration)
-            .parse()
-        : {};
-
-    // Merge arguments
-    const mergedArgv: { [key: string]: unknown } = { ...baseArgv };
-    for (const [key, value] of Object.entries(extraArgv)) {
-        if (key !== '_' && key !== '$0') {
-            if (!value &&
-                typeof value === 'boolean' &&
-                (
-                    typeof (baseArgv as any)[key] === 'boolean' ||
-                    (typeof (baseArgv as any)[`no-${key}`] != 'boolean' && typeof (baseArgv as any)[`no-${key}`])
-                )
-            ) {
-                delete mergedArgv[key];
-            } else {
-                mergedArgv[key] = value;
-            }
-        }
-    }
-
-    // Reconstruct command
-    const mergedArgs = Object.entries(mergedArgv)
-        .filter(([key]) => key !== '_' && key !== '$0')
-        .map(([key, value]) => {
-            if (typeof value === 'boolean') {
-                return value ? `--${key}` : `--no-${key}`;
-            }
-            return `--${key} ${value}`;
-        })
-        .join(' ');
-
-    let fullCommand = `${commandPrefix} ${mergedArgs}`.trim()
-    return fullCommand;
-}
 
 export function getTargetPasswordAccessPolicy(targetPasswordSecretArn: string): PolicyStatement {
     return new PolicyStatement({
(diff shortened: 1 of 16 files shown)


deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -4,87 +4,66 @@ import {CpuArchitecture} from "aws-cdk-lib/aws-ecs";
...
+        // If not present, append the argument and value (only append value if it exists)
+        baseCommand = value !== null ? baseCommand.concat(" ", arg, " ", value) : baseCommand.concat(" ", arg);
     }
+    return baseCommand;
+}
 
-    // Extract command prefix
-    const commandPrefix = baseCommand.substring(0, baseCommand.indexOf('--')).trim();
-    const baseArgs = baseCommand.substring(baseCommand.indexOf('--'));
+export function parseArgsToDict(argString: string | undefined): Record<string, string[]> {
```
AndreKurait, line 32: Note: We're not using the values (only keys) as of now (except the unit tests). I wanted to keep this since it's written, working, and tested

deployment/cdk/opensearch-service-migration/lib/service-stacks/traffic-replayer-stack.ts
```diff
@@ -80,21 +80,37 @@ export class TrafficReplayerStack extends MigrationServiceCore {
         });
         const groupId = props.customKafkaGroupId ? props.customKafkaGroupId : `logging-group-${deployId}`
 
-        let replayerCommand = `/runJavaWithClasspath.sh org.opensearch.migrations.replay.TrafficReplayer ${osClusterEndpoint} --insecure --kafka-traffic-brokers ${brokerEndpoints} --kafka-traffic-topic logging-traffic-topic --kafka-traffic-group-id ${groupId}`
+        let command = `/runJavaWithClasspath.sh org.opensearch.migrations.replay.TrafficReplayer ${osClusterEndpoint}`
+        const extraArgsDict = parseArgsToDict(props.extraArgs)
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--insecure")
```
AndreKurait, line 85: Note: In a followup, we need to figure out how a customer might exclude this, most likely through a cluster setting in cdk.context.json

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -4,87 +4,66 @@ import {CpuArchitecture} from "aws-cdk-lib/aws-ecs";
...
-export function parseAndMergeArgs(baseCommand: string, extraArgs?: string): string {
-    if (!extraArgs) {
-        return baseCommand;
+export function parseArgsToDict(argString: string | undefined): Record<string, string[]> {
+    const args: Record<string, string[]> = {};
+    if (argString === undefined) {
+        return args;
     }
+    // Split based on '--' at the start of the string or preceded by whitespace, use non-capturing groups to include -- in parts
+    const parts = argString.split(/(?=\s--|^--)/).filter(Boolean);
```
gregschohn, line 38: Worst case scenario here is that you could misparse something like 
```
--setHeader from-client \"Search Agent --alpine --classic\"
```

and you'd then drop any arguments that already existed that were named '--alpine' or '--classic', right?

I'd like to see the map eventually supported as just a map of keys and values (possibly the empty string) - and all of it is just concatenated together w/ spaces between.  That way, it's clear what keys are being used and what to strike.

Given that the code was even more problematic before, I'm fine w/ this being a follow up task.

deployment/cdk/opensearch-service-migration/lib/service-stacks/traffic-replayer-stack.ts
gregschohn: Seems like that's the way to go - we exclude kafkaConnection in a similar manner in the code above.

deployment/cdk/opensearch-service-migration/lib/service-stacks/traffic-replayer-stack.ts
```diff
@@ -80,21 +82,53 @@ export class TrafficReplayerStack extends MigrationServiceCore {
...
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--kafka-traffic-brokers", brokerEndpoints)
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--kafka-traffic-topic", "logging-traffic-topic")
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--kafka-traffic-group-id", groupId)
+
         if (props.clusterAuthDetails.basicAuth) {
-            replayerCommand = replayerCommand.concat(` --auth-header-user-and-secret "${props.clusterAuthDetails.basicAuth.username} ${props.clusterAuthDetails.basicAuth.password_from_secret_arn}"`)
+            let secret;
+            if (props.clusterAuthDetails.basicAuth.password) {
+                console.warn("Password passed in plain text, this is insecure and will leave" +
+                    "your password exposed.")
```
gregschohn, line 96: did we support this before?  I don't see it in the old patch & don't want to add it shortly before we cut a major release.

deployment/cdk/opensearch-service-migration/lib/service-stacks/capture-proxy-es-stack.ts
```diff
@@ -63,11 +62,22 @@ export class CaptureProxyESStack extends MigrationServiceCore {
...
-        command = props.streamingSourceType !== StreamingSourceType.DISABLED ? command.concat(`  --kafkaConnection ${brokerEndpoints}`) : command
-        command = props.streamingSourceType === StreamingSourceType.AWS_MSK ? command.concat(" --enableMSKAuth") : command
-        command = props.otelCollectorEnabled ? command.concat(` --otelCollectorEndpoint ${OtelCollectorSidecar.getOtelLocalhostEndpoint()}`) : command
-        command = parseAndMergeArgs(command, props.extraArgs);
+
+        let command = "/usr/local/bin/docker-entrypoint.sh eswrapper & /runJavaWithClasspath.sh org.opensearch.migrations.trafficcapture.proxyserver.CaptureProxy"
+        const extraArgsDict = parseArgsToDict(props.extraArgs)
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--destinationUri", "https://localhost:19200")
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--insecureDestination", "https://localhost:19200")
```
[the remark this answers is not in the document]
AndreKurait: Fixed, thanks

deployment/cdk/opensearch-service-migration/lib/service-stacks/traffic-replayer-stack.ts
AndreKurait: There's two places today we create plaintext secrets, as well as RFS supports it directly as an inline arg.

The experience without this PR/fix is that we concat the argument `--auth-header-user-and-secret username undefined` and the replayer fails on startup.

If you'd like instead we can throw an error here saying "Plaintext password in cdk.context.json with replayer is not supported" and have it be a build time exception.

https://github.com/search?q=repo%3Aopensearch-project%2Fopensearch-migrations+new+Secret%28this&type=code

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
AndreKurait: This would be a change in the customer cdk.context.json to have extra args be a map and not a string?

AndreKurait: Yes, that's the worst case, something like `--setHeader from-client \"Search Agent --destinationUri \"`
An existing `--classic` in your case wouldn't be dropped because of the `"` character after

gregschohn: correct - to accept the value as a dictionary/map and not just as a string (I'd support both using instanceof).  For later though.

deployment/cdk/opensearch-service-migration/lib/service-stacks/traffic-replayer-stack.ts
gregschohn: do we have a jira to stop accepting plaintext passwords and to stop accepting them on the command line.  With that, I'm fine moving forward.  As it is, I consider these major security risks and we shouldn't let the user store secrets in so many places unencrypted.  

As long as the secret is in a task definition, I suppose it doesn't matter if we put it here too - we just need to reverse this too when we lock down the other code

AndreKurait: https://opensearch.atlassian.net/browse/MIGRATIONS-1915

gregschohn (approved):
If you get a chance, can you please unresolve the comments that require future work (I've permalinked them from the ticket, so I don't think it's a big deal either way).

PR #1021, opened 2024-09-26
Set explicit defaults w/ placeholders in the cdk.context.json

mikaylathompson:
### Description
[Describe what this change achieves]
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)
* Why these changes are required?
* What is the old behavior before changes and new behavior after changes?

### Issues Resolved
[List any issues this PR will resolve]

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
[Please provide details of testing done: unit testing, integration testing and manual testing]

### Check List
- [ ] New functionality includes testing
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1022, opened 2024-09-26
Make cdk.context.json valid json

mikaylathompson:
### Description
My last PR, #1021, broke the build.  Having a not-quite-json `cdk.context.json` was bad anytime a `cdk` command was used, _even_ if it specifically specified a different context file.

So first, I just made it valid (if kind of ugly) JSON, but that broke the build in a different way because removing a lot of the default values changed the behavior of the cdk deployments in the CI pipelines. I copied the old values from the defaults into these blocks and have tested redeploying, successfully now.

### Check List


sumobrian (approved):
Passes CI.

PR #1023, opened 2024-09-27
Fixed bug in mapping parsing; added tests

chelma:
### Description
* Fixed a bug reported by a Cx; their mapping list in the snapshot was empty for a specific index, which broke our parsing code.  The only way this can occur (AFAIK) is if they have an index with no mappings and no documents added to it.  As soon as you add a document Elasticsearch adds a mapping to the index.

### Issues Resolved
* N/A

### Testing
* Added unit tests, including a new snapshot to exercise the code path
* Manually ran the MM tool against a local dockerized Target Cluster using that new test snapshot:

```
Command line arguments: migrate --snapshot-name rfs-snapshot --file-system-repo-path /Users/chelma/workspace/migration-assistant/opensearch-migrations/RFS/test-resources/snapshots/ES_7_10_BWC_Check --target-host http://localhost:29200 --source-version ES_7_10 --min-replicas 1 --component-template-allowlist fwc_settings,fwc_mappings --index-template-allowlist fwc_template

Starting Metadata Migration
Clusters:
   Source:
      Snapshot: ELASTICSEARCH 7.10.0 FileSystemRepo(repoRootDir=/Users/chelma/workspace/migration-assistant/opensearch-migrations/RFS/test-resources/snapshots/ES_7_10_BWC_Check)

   Target:
      Remote Cluster: OPENSEARCH 2.11.1 ConnectionContext(uri=http://localhost:29200, protocol=HTTP, insecure=false, compressionSupported=false)


Migrated Items:
   Index Templates:
      fwc_template

   Component Templates:
      fwc_mappings, fwc_settings

   Indexes:
      bwc_index_1, empty_mappings_no_docs, fwc_index_1, no_mappings_no_docs

   Aliases:
      bwc_alias, fwc_alias


Results:
   0 issue(s) detected
```

### Check List


AndreKurait (approved):

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
```diff
@@ -75,17 +76,21 @@ public static void removeIntermediateMappingsLevels(ObjectNode root) {
              * - [{"doc":{"properties":{"address":{"type":"text"}}}}]
              * - [{"audit_message":{"properties":{"address":{"type":"text"}}}}]
              *
-             * It's not impossible that the intermediate key is not present, in which case we should just extract the mappings:
+             * It's also possible for this list to be empty, in which case we should set the mappings to an empty node.
```
gregschohn, line 79: Why is it appropriate to replace [] with "[{}]".  Is this true for all versions of targets?

RFS/src/test/java/org/opensearch/migrations/bulkload/transformers/TransformFunctionsTest.java
```diff
@@ -0,0 +1,98 @@
...
+        ArrayNode mappingsNode1 = new ObjectMapper().createArrayNode();
+        ObjectNode docNode1 = new ObjectMapper().createObjectNode();
+        ObjectNode propertiesNode1 = new ObjectMapper().createObjectNode();
+        ObjectNode addressNode1 = new ObjectMapper().createObjectNode();
+        addressNode1.put("type", "text");
+        propertiesNode1.set("address", addressNode1);
+        docNode1.set(TransformFunctions.PROPERTIES_KEY_STR, propertiesNode1);
+        ObjectNode intermediateNode1 = new ObjectMapper().createObjectNode();
+        intermediateNode1.set("_doc", docNode1);
+        mappingsNode1.add(intermediateNode1);
```
gregschohn, line 28: could you replace this w/ inline json that gets parsed?
```
  var json = 
      "{" +
      "  \"mappings\": [{" +
      "    \"_doc\": {" +
      ...
```

It's a lot easier to read and to maintain.  You can copy-paste literal json from a text editor into your IDE and it should do all of the necessary escaping (IntelliJ does this, I'd hope eclipse & vscode do too).

Then you can just new `ObjectMapper().readTree(json)` and get the parsed JsonNode back.

RFS/test-resources/inventory.md
```diff
@@ -426,12 +426,36 @@ curl -X PUT "localhost:19200/_index_template/fwc_template" -H "Content-Type: app
...
+    "index": {
+      "number_of_shards": 1,
+      "number_of_replicas": 0
+    }
+  },
+  "mappings": {
+    "properties":{
+    }
+  }
+}'
```
gregschohn, line 457: I'd love to see us generate these on-the-fly and then cache them like we do for preloaded data images.  In fact, use the preloaded images, then just take a snapshot via a helper program.  Make the output of that something that gets cached as a build artifact and then it can become a dependency to the tests.  It seems like that could be 1-2 hours of work and removes the risks of many hours of work trying to figure out if there was a mismatch later on.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
chelma: This results in an index creation call that looks like:

```
curl -X PUT "localhost:92011/index_name" -H "Content-Type: application/json" -d '
{
  "settings": {
        <the settings>
  },
  "mappings": {
  }
}'
```

I'm not sure if the empty mappings blob is valid for ALL possible targets, but it does appear valid for the targets we currently consider - OS 1.X & OS 2.X

PR #1024, opened 2024-09-27
Remove arg printing from command line apps with secret values

AndreKurait:
### Description

CreateSnapshot, DocumentsFromSnapshotMigration, and MetadataMigration printed args which could contain secret value password

* Category: Bug Fix
* Why these changes are required? Password could be printed in plaintext
* What is the old behavior before changes and new behavior after changes? Password could be printed in plaintext

### Issues Resolved
N/A

### Check List

### DIFF
--- a/CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
+++ b/CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
@@ -73,6 +73,7 @@ public class CreateSnapshot {
     }
 
     public static void main(String[] args) throws Exception {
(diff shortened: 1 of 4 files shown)


sumobrian (approved):

gregschohn (changes requested):
How long do we think it would take to load this from an arn like we do in the replayer?

As previously indicated, putting passwords on the command line in plain text open it up to a number of eyes.  It will be in the shell history, which is often persisted to a file.  It's in the task definition, which is viewable by anybody w/ readonly access.

Command lines aren't places to store sensitive data.  Suppressing how a program was invoked may have its own post-forensic issues.  I'd vote to resolve this by pulling from the arn.  I wouldn't put a security seal on passing the password as an arg w/ any number of workarounds.

PR #1031, opened 2024-09-30
?v -> ?v=true to workaround the sigv4 bug

mikaylathompson:
### Description
There's a bug in our sigv4 dependency for `requests` that breaks when a url has a query param with no value. So `?v` fails but `?v=true` succeeds.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2023

### Testing
Manual

### Check List


peternied (approved):
I think this is a good work around, thanks for creating the issue on the source library [1]

- [1] https://github.com/andrewjroth/requests-auth-aws-sigv4/issues/20

PR #1034, opened 2024-09-30
Optimize the regex parser for InvalidResponse

peternied:
### Description
Rather than using a backtracking pattern which could have performance impact, switch to using explict character ranges which work for json path seperated with periods.

Mitigates sonar lint rule java:S5852 [1]

- [1] https://rules.sonarsource.com/java/RSPEC-5852/?search=Using%20slow%20regular%20expressions%20is%20security-sensitive

### Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
@@ -17,7 +17,7 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class InvalidResponse extends RfsException {
-    private static final Pattern UNKNOWN_SETTING = Pattern.compile("unknown setting \\[(.+?)\\].+");
+    private static final Pattern UNKNOWN_SETTING = Pattern.compile("unknown setting \\[([a-zA-Z0-9_.-]+)\\].+");
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private final transient HttpResponse response;
 


RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
```diff
@@ -17,7 +17,7 @@
 
 @Slf4j
 public class InvalidResponse extends RfsException {
-    private static final Pattern UNKNOWN_SETTING = Pattern.compile("unknown setting \\[(.+?)\\].+");
+    private static final Pattern UNKNOWN_SETTING = Pattern.compile("unknown setting \\[([a-zA-Z0-9_.-]+)\\].+");
```
AndreKurait, line 20: Would the issue also be resolved by changing `.+?` -> `.+`

peternied: Yes, but that expression wouldn't be correct since it could match the closing bracket or a host of invalid characters that opensearch would never use in a setting name.
- ✔️ `unknown setting [foo.bar] other stuff` -> `foo.bar` 
- ❌ `unknown setting [foo.bar] other stuff ]` -> `foo.bar] other stuff`

Note, this could also be mitigated by removing the trailing `.+` match, but this version is more correct and CI is nearly passed.

AndreKurait: I see.. the issue is that `]` shouldn't be a matching character in the inner match.

`"unknown setting \\[([^\]\n]+?)\\].+"` this would be the most similar to the original logic that doesn't have the bug

This change seems safe to set the characters inside the `[]` to those in this PR

AndreKurait (approved):

peternied:
Needed to retry the gradle tests, created Flaky Test: WorkCoordinatorTest.testAcquireLeaseForQuery [1]

- [1] https://opensearch.atlassian.net/browse/MIGRATIONS-2038

PR #1037, opened 2024-09-30
Continue to evaulate / migrate after failures

peternied:
### Description
Updates the behavior of the metadata processing for templates and indexes to collect all the results and then return failure details after all items have been processed.  When items are not created they will appear in the output with a reason and exception if needed be.  Here is the output from an updated evaulate test case where the index `already_exists` was already on the target cluster and won't be created.

```
Clusters:
   Source:
      Remote Cluster: ELASTICSEARCH 6.8.23 ConnectionContext(uri=http://localhost:36188, protocol=HTTP, insecure=false, compressionSupported=false)

   Target:
      Remote Cluster: OPENSEARCH 2.14.0 ConnectionContext(uri=http://localhost:36187, protocol=HTTP, insecure=false, compressionSupported=false)


Migration Candidates:
   Index Templates:
      simple_index_template

   Component Templates:
      <NONE FOUND>

   Indexes:
      blog_2023, movies_2023
      WARN - already_exists already exists

   Aliases:
      alias1, movies-alias


Results:
   0 issue(s) detected
```

### Check List


RFS/src/main/java/org/opensearch/migrations/metadata/CreationResult.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public boolean wasFatal() {
+        return Optional.ofNullable(getFailureType()).map(CreationFailureType::isFatal)
+            .orElse(false);
+    }
+
+    @AllArgsConstructor
+    @Getter
+    public static enum CreationFailureType {
```
peternied, line 27: Might be missing a way to show that we did not create an item because it wasn't included on an allow list, curious what others think of these categories

MetadataMigration/src/main/java/org/opensearch/migrations/cli/Items.java
```diff
@@ -31,25 +34,73 @@ public String asCliOutput() {
...
+    }
+
+    private void appendItems(StringBuilder sb, List<CreationResult> items) {
+        var successfulItems = items.stream()
+            .filter(r -> r.wasSuccessful())
+            .map(r -> r.getName())
+            .collect(Collectors.toList());
+
+        if (!successfulItems.isEmpty()) {
+            sb.append(Format.indentToLevel(2))
```
AndreKurait, line 69: nit: would prefer `2` to be a parameter to this function and appendFailures to be less errorProne if extending

RFS/src/main/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11.java
```diff
@@ -40,35 +43,45 @@ public boolean create(
...
                     throw invalidResponse;
                 }
 
-                var shortenedIllegalArgument = illegalArgument.replaceFirst("index.", "");
-                removeFieldsByPath(settings, shortenedIllegalArgument);
-            }
+                for (var illegalArgument : illegalArguments) {
+                    if (!illegalArgument.startsWith("index.")) {
+                        log.warn("Expecting all retryable errors to start with 'index.', instead saw " + illegalArgument);
+                        throw invalidResponse;
```
AndreKurait, line 70: Should we instead include an exception with  the message above, otherwise this log may be lost in the output

RFS/src/main/java/org/opensearch/migrations/metadata/CreationResult.java
AndreKurait: i'm good with these categories for now. Maybe to extend to stuff like the allowList, we shouldn't call this a FailureType.. maybe CreationSkipped?

RFS/src/main/java/org/opensearch/migrations/metadata/CreationResult.java
```diff
@@ -0,0 +1,46 @@
...
+        ALREADY_EXISTS(false, "already exists"),
+        DEPENDENCY_FAILURE(false, "dependency failure"),
```
AndreKurait, line 29: Can you describe DEPENDENCY_FAILURE and why it's not fatal?

RFS/src/main/java/org/opensearch/migrations/metadata/GlobalMetadataCreatorResults.java
```diff
@@ -1,14 +1,22 @@
...
-    private List<String> indexTemplates;
-    private List<String> componentTemplates;
+    private List<CreationResult> legacyTemplates;
+    private List<CreationResult> indexTemplates;
+    private List<CreationResult> componentTemplates;
+
+    public long fatalIssueCount() {
+        return Stream.of(getLegacyTemplates(), getIndexTemplates(), getComponentTemplates())
+            .flatMap(List::stream)
+            .filter(r -> !r.wasFatal())
```
AndreKurait, line 19: IIRC, filter includes elements that are true, so this will yield the nonFatalIssueCount

RFS/src/main/java/org/opensearch/migrations/metadata/CreationResult.java
peternied: I've removed this kind of failure since it was not as well defined as we would like.

RFS/src/main/java/org/opensearch/migrations/metadata/GlobalMetadataCreatorResults.java
peternied: Nice catch, added a test case too.

MetadataMigration/src/main/java/org/opensearch/migrations/cli/Items.java
peternied: I'll leave this as is for now, but if we invest more in the CLI experience in java this could use some more thought

AndreKurait (approved):
Thanks for the cleanup, i'm good for merging once the isolatedTests are fixed

PR #1041, opened 2024-10-01
RFS Workers now use EBS to store/unpack the snapshot

chelma:
### Description
* Updated the RFS Workers to use an EBS volume to store and unpack the snapshot.  Each worker gets a separate volume, which is provisioned on task launch.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-1750

### Testing
* Tested by performing a small migration in AWS and confirming separate, correctly-size volumes were used for each task

```
(.venv) bash-5.2# console clusters cat-indices

SOURCE CLUSTER
health status index       uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   bwc_index_1 QYzCFpW9RTeZ9xW3ht5eTA   1   0          1            0      3.8kb          3.8kb
green  open   fwc_index_1 Axns2r74RCiDQYt6ewLAeQ   1   0          1            0      3.8kb          3.8kb

TARGET CLUSTER
health status index                     uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   .migrations_working_state BUg_NWvsSE-a8kbdnPN6_Q   1   1          3            0     47.1kb          6.8kb
green  open   bwc_index_1               NR4nYu4rSiyfLNqRr3UQWw   5   1          1            0     11.5kb          5.7kb
green  open   fwc_index_1               xsBlWTtLT5-SJUUni3kBog   1   0          1            0      3.8kb          3.8kb
```
Screenshot 2024-10-01 at 1 54 28 PM

Screenshot 2024-10-01 at 1 55 30 PM

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/default-values.json
+++ b/deployment/cdk/opensearch-service-migration/default-values.json
@@ -4,5 +4,6 @@
   "migrationAssistanceEnabled": true,
   "migrationConsoleServiceEnabled": true,
   "trafficReplayerServiceEnabled": false,
-  "otelCollectorEnabled": true
+  "otelCollectorEnabled": true,
+  "reindexFromSnapshotMaxShardSizeGiB": 80
 }
--- a/deployment/cdk/opensearch-service-migration/default-values.json
+++ b/deployment/cdk/opensearch-service-migration/default-values.json
@@ -4,6 +4,5 @@
   "migrationAssistanceEnabled": true,
   "migrationConsoleServiceEnabled": true,
   "trafficReplayerServiceEnabled": false,
-  "otelCollectorEnabled": true,
-  "reindexFromSnapshotMaxShardSizeGiB": 80
+  "otelCollectorEnabled": true
 }
--- a/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
@@ -99,7 +99,8 @@ export class NetworkStack extends Stack {
         // General interface endpoints
         const interfaceEndpoints = [
             InterfaceVpcEndpointAwsService.CLOUDWATCH_LOGS, // Push Logs from tasks
-            InterfaceVpcEndpointAwsService.CLOUDWATCH_MONITORING, // Pull Metrics from Migration Console 
+            InterfaceVpcEndpointAwsService.CLOUDWATCH_MONITORING, // Pull Metrics from Migration Console
+            InterfaceVpcEndpointAwsService.EBS_DIRECT, // Enable EBS Volumes
             InterfaceVpcEndpointAwsService.ECR_DOCKER, // Pull Images on Startup
             InterfaceVpcEndpointAwsService.ECR, // List Images on Startup
             InterfaceVpcEndpointAwsService.ECS_AGENT, // Task Container Metrics
--- a/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
@@ -100,7 +100,6 @@ export class NetworkStack extends Stack {
         const interfaceEndpoints = [
             InterfaceVpcEndpointAwsService.CLOUDWATCH_LOGS, // Push Logs from tasks
             InterfaceVpcEndpointAwsService.CLOUDWATCH_MONITORING, // Pull Metrics from Migration Console
-            InterfaceVpcEndpointAwsService.EBS_DIRECT, // Enable EBS Volumes
             InterfaceVpcEndpointAwsService.ECR_DOCKER, // Pull Images on Startup
             InterfaceVpcEndpointAwsService.ECR, // List Images on Startup
             InterfaceVpcEndpointAwsService.ECS_AGENT, // Task Container Metrics
--- a/deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
@@ -149,7 +149,7 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
                 fileSystemType: FileSystemType.XFS,
                 tagSpecifications: [{
                     tags: {
-                        name: `rfs-snapshot-volume-${props.stage}`,
+                        Name: `rfs-snapshot-volume-${props.stage}`,
                     },
                     propagateTags: EbsPropagatedTagSource.SERVICE,
                 }],
--- a/deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
@@ -79,12 +79,14 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
         let command = "/rfs-app/runJavaWithClasspath.sh org.opensearch.migrations.RfsMigrateDocuments"
         const extraArgsDict = parseArgsToDict(props.extraArgs)
         const storagePath = "/storage"
+        const maxShardSizeBytes = props.maxShardSizeGiB ? `${props.maxShardSizeGiB * 1024 * 1024 * 1024}` : "8589934592"
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-local-dir", `"${storagePath}/s3_files"`)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-repo-uri", `"${s3Uri}"`)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-region", this.region)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--snapshot-name", "rfs-snapshot")
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--lucene-dir", `"${storagePath}/lucene"`)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--target-host", osClusterEndpoint)
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--max-shard-size-bytes", maxShardSizeBytes)
         if (props.clusterAuthDetails.sigv4) {
             command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--target-aws-service-signing-name", props.clusterAuthDetails.sigv4.serviceSigningName)
             command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--target-aws-region", props.clusterAuthDetails.sigv4.region)
@@ -130,20 +132,22 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
         const planningSize = props.maxShardSizeGiB ?? 80;
         const volumeSize = Math.max(
             Math.ceil(planningSize * 2 * 1.15),
-            125 // Minimum volume size for ST1
+            1
         )
+
+        if (volumeSize > 16000) {
+            // 16 TiB is the maximum volume size for GP3
+            throw new Error(`"Your max shard size of ${props.maxShardSizeGiB} GiB is too large to migrate."`)
+        }
 
         // Volume we'll use to download and unpack the snapshot
         const snapshotVolume = new ServiceManagedVolume(this, 'SnapshotVolume', {
             name: 'snapshot-volume',
             managedEBSVolume: {
                 size: Size.gibibytes(volumeSize),
-                volumeType: EbsDeviceVolumeType.ST1,
+                volumeType: EbsDeviceVolumeType.GP3,
                 fileSystemType: FileSystemType.XFS,
                 tagSpecifications: [{
-                    tags: {
-                    purpose: 'production',
-                    },
                     propagateTags: EbsPropagatedTagSource.SERVICE,
                 }],
             },
@@ -168,7 +172,6 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
             cpuArchitecture: props.fargateCpuArch,
             taskCpuUnits: 2048,
             taskMemoryLimitMiB: 4096,
-            ephemeralStorageGiB: 60, // Use for the RFS Process; not the snapshot & lucene storage
             environment: {
                 "RFS_COMMAND": command,
                 "RFS_TARGET_USER": targetUser,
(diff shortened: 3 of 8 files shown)


deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -210,6 +210,7 @@ export class StackComposer {
         const otelCollectorEnabled = this.getContextForType('otelCollectorEnabled', 'boolean', defaultValues, contextJSON)
         const reindexFromSnapshotServiceEnabled = this.getContextForType('reindexFromSnapshotServiceEnabled', 'boolean', defaultValues, contextJSON)
         const reindexFromSnapshotExtraArgs = this.getContextForType('reindexFromSnapshotExtraArgs', 'string', defaultValues, contextJSON)
+        const reindexFromSnapshotMaxShardSizeGiB = this.getContextForType('reindexFromSnapshotMaxShardSizeGiB', 'number', defaultValues, contextJSON)
```
mikaylathompson, line 213: What are the limits around this? How big can (should) we get? Can we check the bounds and throw an error earlier instead of later?

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +122,53 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
+        // Volume we'll use to download and unpack the snapshot
+        const snapshotVolume = new ServiceManagedVolume(this, 'SnapshotVolume', {
+            name: 'snapshot-volume',
+            managedEBSVolume: {
+                size: Size.gibibytes(volumeSize),
+                volumeType: EbsDeviceVolumeType.ST1,
+                fileSystemType: FileSystemType.XFS,
+                tagSpecifications: [{
+                    tags: {
+                    purpose: 'production',
```
mikaylathompson, line 145: what does this tag do/affect?

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +122,53 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
-            volumes: [sharedLogFileSystem.asVolume()],
-            mountPoints: [sharedLogFileSystem.asMountPoint()],
+            volumes: volumes,
+            mountPoints: mountPoints,
             taskRolePolicies: servicePolicies,
             cpuArchitecture: props.fargateCpuArch,
             taskCpuUnits: 2048,
             taskMemoryLimitMiB: 4096,
-            ephemeralStorageGiB: 200,
+            ephemeralStorageGiB: 60, // Use for the RFS Process; not the snapshot & lucene storage
```
AndreKurait, line 171: Can we set this as 20 which is the free size amount?

deployment/cdk/opensearch-service-migration/default-values.json
```diff
@@ -4,5 +4,6 @@
   "migrationAssistanceEnabled": true,
   "migrationConsoleServiceEnabled": true,
   "trafficReplayerServiceEnabled": false,
-  "otelCollectorEnabled": true
+  "otelCollectorEnabled": true,
+  "reindexFromSnapshotMaxShardSizeGiB": 80
```
AndreKurait, line 8: Can we move this to cdk.context.json? I'd like to increase visibility into this value

mikaylathompson: Oh, good catch, +1

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +122,53 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
+            Math.ceil(planningSize * 2 * 1.15),
+            125 // Minimum volume size for ST1
+        )
+
+        // Volume we'll use to download and unpack the snapshot
+        const snapshotVolume = new ServiceManagedVolume(this, 'SnapshotVolume', {
+            name: 'snapshot-volume',
+            managedEBSVolume: {
+                size: Size.gibibytes(volumeSize),
+                volumeType: EbsDeviceVolumeType.ST1,
```
AndreKurait, line 141: Let's use GP3 here. I'm concerned the throughput and iops will be a bottleneck for small volume sizes. This will also enable us to remove the minimumVolumeSize above


> with a baseline throughput of 40 MB/s per TB
https://aws.amazon.com/ebs/volume-types/

chelma: @AndreKurait Can you be more specific around the specific throughput parameters you expect the system to require?  There's also a significant cost difference.

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
chelma: Sure

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
chelma: Will try deleting and see what happens.

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
chelma: I think the stack will fail to synthesize if you go over the maximum EBS size, but I can add an explicit exception.  Good call.

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
AndreKurait, line 1: Add EBS to VPCe list otherwise the container startup won't be able to create the volume https://github.com/opensearch-project/opensearch-migrations/blob/f28e131aa4ee4b42d5120b557d0ca2e9292eb95c/deployment/cdk/opensearch-service-migration/lib/network-stack.ts#L103

AndreKurait, line 1: Can you post a screenshot of the time it takes to startup a new task with these settings? If it's more than 1 minute i'd like to explore using ephemeral storage for maxShardSizeGB <= 80

image

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
AndreKurait: Discussed offline, given ST1s baseline throughput of `40MBps per TB` and uncertainty around throughput with volumes much smaller than 1TB, we've aligned to go with GP3 despite the ~2x cost increase (GP3 being comparable to Ephemeral Storage cost)

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
chelma: I didn't experience a problem with EBS volumes attaching in my testing, but I'll add it to the list.

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -70,11 +78,12 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
         const extraArgsDict = parseArgsToDict(props.extraArgs)
-        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-local-dir", "/tmp/s3_files")
+        const storagePath = "/storage"
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-local-dir", `"${storagePath}/s3_files"`)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-repo-uri", `"${s3Uri}"`)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--s3-region", this.region)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--snapshot-name", "rfs-snapshot")
-        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--lucene-dir", "/lucene")
+        command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--lucene-dir", `"${storagePath}/lucene"`)
         command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--target-host", osClusterEndpoint)
```
AndreKurait, line 87: can we set the max shard size arg here too?

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
chelma: The startup time is less than a minute, and the volume attachment doesn't seem to add a noticeable delay.

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
chelma: Good catch!

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +122,53 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
+        // Calculate the volume size based on the max shard size
+        // Have space for the snapshot and an unpacked copy, with buffer
+        const planningSize = props.maxShardSizeGiB ?? 80;
+        const volumeSize = Math.max(
+            Math.ceil(planningSize * 2 * 1.15),
+            125 // Minimum volume size for ST1
+        )
+
+        // Volume we'll use to download and unpack the snapshot
+        const snapshotVolume = new ServiceManagedVolume(this, 'SnapshotVolume', {
```
AndreKurait, line 137: Is encryption an option?
image

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
AndreKurait: This path is exercised when the cdk.context.json does not specify a vpcid.

I had a compatible environment already spun up so just checked-out and deployed this PR, looks like it's working so no action is needed here 😄

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
chelma: Yes, it is an option; I'm baffled why the default is "False".  Another good catch!

deployment/cdk/opensearch-service-migration/lib/network-stack.ts
```diff
@@ -99,7 +99,8 @@ export class NetworkStack extends Stack {
         // General interface endpoints
         const interfaceEndpoints = [
             InterfaceVpcEndpointAwsService.CLOUDWATCH_LOGS, // Push Logs from tasks
-            InterfaceVpcEndpointAwsService.CLOUDWATCH_MONITORING, // Pull Metrics from Migration Console 
+            InterfaceVpcEndpointAwsService.CLOUDWATCH_MONITORING, // Pull Metrics from Migration Console
+            InterfaceVpcEndpointAwsService.EBS_DIRECT, // Enable EBS Volumes
```
AndreKurait, line 103: Sorry for the back and forth, but can we remove this? Looks like it's only for snapshot operations https://docs.aws.amazon.com/ebs/latest/APIReference/API_Operations.html

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +124,54 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
             servicePolicies.push(getSecretsPolicy);
         }
 
+        const volumes = [sharedLogFileSystem.asVolume()];
+        const mountPoints = [sharedLogFileSystem.asMountPoint()];
+
+        // Calculate the volume size based on the max shard size
+        // Have space for the snapshot and an unpacked copy, with buffer
+        const planningSize = props.maxShardSizeGiB ?? 80;
```
AndreKurait, line 132: nit: Can we move this (and possibly other logic here) the top of the function to align the default here maxShardSizeBytes so we don't have two defaults set?

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +124,54 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
+        }
+
+        // Volume we'll use to download and unpack the snapshot
+        const snapshotVolume = new ServiceManagedVolume(this, 'SnapshotVolume', {
+            name: 'snapshot-volume',
+            managedEBSVolume: {
+                size: Size.gibibytes(volumeSize),
+                volumeType: EbsDeviceVolumeType.GP3,
+                fileSystemType: FileSystemType.XFS,
+                tagSpecifications: [{
```
AndreKurait, line 150: EBS Console shows name which reflects the `Name` tag on the volume, can we populate it with something that says RFS and the stage

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
AndreKurait: Looks like the default is the ECS Cluster settings which we don't set

image

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
AndreKurait: it's case sensitive for `Name`
image
image

deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts
```diff
@@ -113,19 +125,57 @@ export class ReindexFromSnapshotStack extends MigrationServiceCore {
...
+                    tags: {
+                        name: `rfs-snapshot-volume-${props.stage}`,
```
AndreKurait, line 152: nit: name -> Name

https://github.com/opensearch-project/opensearch-migrations/pull/1041#discussion_r1783581674

AndreKurait (approved):
LGTM, see non blocking comment for tag name which can be handled in this or a follow up pr

PR #1043, opened 2024-10-01
Remove all dependencies from the migration-console stack

gregschohn:
That means that the console can come up and be used for diagnostics if the other stacks/clusters don't properly deploy.

### Description

This change allows the migration console to be deployed w/out any dependencies of other tasks that it's meant to interface with so that the console itself can be used as a diagnostic tool

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2037

### Testing
limited testing by redploying from my bootstrap box

### Check List


AndreKurait (commented):
I'm fine with the spirit of this change, though i'm a bit concerned with how this interacts with the SSM parameters dependencies since they would be added during deployment.

Have we tested with this PR on the first deployment in a new stage with concurrency set to a very high value?

AndreKurait (approved):
Please verify ^ test case before merge

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -621,8 +621,7 @@ export class StackComposer {
             })
             // To enable the Migration Console to make requests to other service endpoints with services,
             // it must be deployed after any connected services
-            this.addDependentStacks(migrationConsoleStack, [captureProxyESStack, captureProxyStack, elasticsearchStack,
-                openSearchStack, osContainerStack, migrationStack, kafkaBrokerStack])
+            this.addDependentStacks(migrationConsoleStack, [migrationStack])
```
peternied, line 624: This seems like the correct list, I think we accidentally inverted the dependencies when this was first added.

peternied (commented):
> I'm fine with the spirit of this change, though i'm a bit concerned with how this interacts with the SSM parameters dependencies since they would be added during deployment.
> 
> Have we tested with this PR on the first deployment in a new stage with concurrency set to a very high value?

@AndreKurait does the jenkins job verify this behavior so manually checks aren't needed for this change?

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
lewijacn: You can remove the comment above this as well. Originally when we used service connect to talk between ECS services, the Migration Console would need to be deployed after those services to have their service name mapping with service connect. We no longer user Service Connect I believe so that shouldn't block us. 

The other concern I have is loading in environment variables. If we are still loading in environment variables from other stacks that create a source or target (not our import case) say in the opensearch container stack case, we wanted to be sure that endpoint existed and could be passed. This seems like something CDK may be smart enough to detect without manually doing it, but it would be nice to confirm that's the case if we still want to support that use case.

gregschohn:
I'm currently focused on K8s

PR #1045, opened 2024-10-02
Add a doc about the architecture of the migration console

mikaylathompson:
### Description
Putting up a doc about the architecture of the migration console--both the box and the library, to be reviewed.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1755

### Testing
n/a

### Check List


docs/diagrams/migration-console-arch.svg
peternied, line 1: Its strange to have the bootstrap box and CloudWatch metrics outside any boundary, what about putting a container for the `Customer AWS Account`?

docs/diagrams/migration-console-library-source.md
peternied, line 1: https://excalidraw.com/ is where this was generated from? I wasn't able to import this file without getting an error `Error: invalid file`

docs/migration-console.md
peternied, line 1: Lots of AWS terminology in this document, what would you think about being agnostic of the deployment fabric?

docs/migration-console.md
```diff
@@ -0,0 +1,123 @@
...
+
+**Goals and Scope**:
+
+- Ensure centralized control and visibility into the migration process.
+- Provide interfaces for managing communication between the Migration Console and other components.
+- Operate statelessly with a transparent source of truth for predictable and configurable behavior.
+
+## System Architecture
+
+The Migration Console is deployed into the customer's VPC and maintains connections with effectively all other components of the Migration Assistant. The below diagram is not comprehensive and does not illustrate any of the data flow between various other components.
```
peternied, line 16: Lots of `ly` words, typically these can be removed to clarify the text in a less is more sense.  In this case `maintains connections with effectively all other components` could be `maintains connections with components`.

docs/migration-console.md
```diff
@@ -0,0 +1,123 @@
...
+
+![Migration Solution Architecture](diagrams/migration-console-arch.svg)
+[Source](https://tiny.amazon.com/jmayhucf/desia2zIMi)
+
+### Individual Connections
+
+#### Bootstrap Box
```
peternied, line 23: The bootstrap box isn't part of the console, but part of the solutions deployment, seems like it shouldn't be in this doc?

docs/migration-console.md
```diff
@@ -0,0 +1,123 @@
...
+- Operate statelessly with a transparent source of truth for predictable and configurable behavior.
+
+## System Architecture
+
+The Migration Console is deployed into the customer's VPC and maintains connections with effectively all other components of the Migration Assistant. The below diagram is not comprehensive and does not illustrate any of the data flow between various other components.
+
+![Migration Solution Architecture](diagrams/migration-console-arch.svg)
+[Source](https://tiny.amazon.com/jmayhucf/desia2zIMi)
+
+### Individual Connections
```
peternied, line 21:
```suggestion
### Components
```

docs/diagrams/migration-console-library-source.md
```diff
@@ -0,0 +1,261 @@
...
+Environment ^lrTOHtQJ
+
+- get an empty one
+- get one based on YAML
+
+All middleware layer functions
+get passed an Environment, plus
+args ^z9HycCiZ
+
+DMS API ^CTJ9Hu4D
```
peternied, line 55: Lets remove any references to a DMS API - REST/CLI should be enough

peternied: Since the interface isn't locked in and its effectively built using the REST interface it doesn't really impact the architecture of the MA console, what do you think?

docs/migration-console.md
```diff
@@ -0,0 +1,123 @@
...
+
+##### DMS API
+
+In the short term, a small subset of the functionality is necessary to meet the requirements of our in-progress collaboration with Database Migration Service (DMS). In some of these cases, DMS has specified an API contract that doesn't align with our plan for the API endpoints and data received.  In these cases, we will build out DMS specific endpoints that meet these exact purposes without trying to also match the general case.
+
+#### Generation from a spec
+
+For the CLI and (standard) web API, the intention is that the “frontend” should be very small—essentially just the necessary code to get the user’s data from the input source (either CLI commands or http requests) into the library itself. To accomplish this, the actual endpoints in both cases could be generated from an API spec or one of them from the other. This will be a helpful way to minimize the developer effort to support additional commands and ensure that behavior is consistent between the two modalities.
+
+#### Custom Frontends
```
peternied, line 83: What do you think about removing this section until we support a scenario like this?  If we don't already have this captured in a jira I think this would be a good candidate for one so the idea isn't lost (I really like the idea!)

peternied (commented):
Thanks for creating this - I've left a couple of different kinds of feedback, let me know if you'd like another/different pass on what you've written here.

docs/diagrams/migration-console-library-source.md
mikaylathompson: Yeah, I've replaced this with just discussing the Web API and additional frontends

docs/migration-console.md
mikaylathompson: I've reworked this to discuss the non-AWS deployment options fully in parallel to the AWS version. The caveat is that--while the library has the placeholders in--we haven't actually built out support for a lot of the non-AWS options (e.g. the only way to start the replayer is to call the ECS api, we didn't build the docker equivalent). So that's why a lot of this is AWS focused currently.

docs/migration-console.md
mikaylathompson: We actually do have one--the integ tests. I changed this to do a clearer job describing that and de-emphasized the future ideas. Let me know if you want them fully removed.

peternied (approved):
Thanks for for the updates, looks good to me!

PR #1047, opened 2024-10-02
Run the RFS Container's DocumentMigration application repeatedly as long as it's successful

gregschohn:
Unsuccessful means anything but a 0 exit code.  The application has been cleaned up to return 3 for no work available.  It used to return whatever a top-level exception would create.  That part wasn't necessary, but it seemed like a good idea to test for that to make sure that after running repeatedly, processes would eventually do THAT instead of returning 0, causing infinite loops in the containers.

### Description

This eliminates container startup costs while still cleaning out the java process so that we don't have to worry about runaway processes.

* Category Enhancement
* Why these changes are required? For clusters with very small indices, migrations could take a very long time. since every tiny shard would need to create a new container.  That should be drastically reduced now (~100x ?).

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2042

### Testing
Manual testing of the entrypoint.sh script
```
docker % export RFS_COMMAND="exit 1" 
docker % ./entrypoint.sh            
RFS_COMMAND: exit 1
RFS_TARGET_USER: 
RFS_TARGET_PASSWORD: <redacted>
RFS_TARGET_PASSWORD_ARN: 
Running command exit 1
```

```
export RFS_COMMAND="sleep 1"
docker % ./entrypoint.sh
RFS_COMMAND: sleep 1
RFS_TARGET_USER: 
RFS_TARGET_PASSWORD: <redacted>
RFS_TARGET_PASSWORD_ARN: 
Running command...
Running command...
^C

```

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/docker/entrypoint.sh
+++ b/DocumentsFromSnapshotMigration/docker/entrypoint.sh
@@ -43,5 +43,6 @@ if [[ $RFS_COMMAND != *"--target-password"* ]]; then
     fi
 fi
 
-echo "Executing RFS Command"
-eval $RFS_COMMAND
+[ -z "$RFS_COMMAND" ] && \
+{ echo "Warning: RFS_COMMAND is empty! Exiting."; exit 1; } || \
+until ! { echo "Running command $RFS_COMMAND"; eval "$RFS_COMMAND"; }; do :; done
--- a/DocumentsFromSnapshotMigration/docker/entrypoint.sh
+++ b/DocumentsFromSnapshotMigration/docker/entrypoint.sh
@@ -43,6 +43,20 @@ if [[ $RFS_COMMAND != *"--target-password"* ]]; then
     fi
 fi
 
+# Monitor how much aggregate free space is left, if we drop below a certain amount,
+# don't allow another process to run again
+space_drop_fraction=4
+initial_free_space=$(df --output=avail | awk 'NR>1 {sum+=$1} END {print sum}')
+still_enough_space_left() {
+  current_free=$(df --output=avail | awk 'NR>1 {sum+=$1} END {print sum}')
+  if (((initial_free_space - current_free) >= (initial_free_space / $space_drop_fraction))); then
+    return 1;
+  else
+    return 0;
+  fi
+}
+
+
 [ -z "$RFS_COMMAND" ] && \
 { echo "Warning: RFS_COMMAND is empty! Exiting."; exit 1; } || \
-until ! { echo "Running command $RFS_COMMAND"; eval "$RFS_COMMAND"; }; do :; done
+until ! { echo "Running command $RFS_COMMAND"; eval "$RFS_COMMAND" && still_enough_space_left; }; do :; done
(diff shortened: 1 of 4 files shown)


DocumentsFromSnapshotMigration/docker/entrypoint.sh
AndreKurait, line 1: We should be able to use testContainers to verify this behavior.

Do we need to delete and re-create the `--lucene-dir` and `--s3-local-dir` between runs?

gregschohn: testContainers would only work if I built the image. I didn't want to go through the gradle contortions to add a new test that was dependent upon the docker build. As a general rule, it would be nice to have all of our docker builds to go through a test.  That seems like it would be a nice investment to make across the board.  We don't do it today though & I think we should handle that build enhancement separately.

AndreKurait (approved):

DocumentsFromSnapshotMigration/docker/entrypoint.sh
AndreKurait: Approving since functionality is verified in rfs-e2e-aws-test

DocumentsFromSnapshotMigration/docker/entrypoint.sh
```diff
@@ -43,5 +43,6 @@ if [[ $RFS_COMMAND != *"--target-password"* ]]; then
     fi
 fi
 
-echo "Executing RFS Command"
-eval $RFS_COMMAND
+[ -z "$RFS_COMMAND" ] && \
+{ echo "Warning: RFS_COMMAND is empty! Exiting."; exit 1; } || \
+until ! { echo "Running command $RFS_COMMAND"; eval "$RFS_COMMAND"; }; do :; done
```
chelma, line 48: We should definitely delete the S3/Lucene directories between runs.  This enables us stay without our disk space boundaries (while we care about them) as well having a clean and repeatable starting point for the process.  The only tradeoff I can think of is the need to re-download some small snapshot metadata files, which feels non-impactful.

chelma (changes requested):

DocumentsFromSnapshotMigration/docker/entrypoint.sh
```diff
@@ -43,5 +43,47 @@ if [[ $RFS_COMMAND != *"--target-password"* ]]; then
...
+if [[ -n "$S3_LOCAL_DIR" ]]; then
+    echo "Will delete S3 local directory between runs: $S3_LOCAL_DIR"
+    rm -rf "$S3_LOCAL_DIR"
+    echo "Directory $S3_LOCAL_DIR has been deleted."
+else
+    echo "--s3-local-dir argument not found in RFS_COMMAND. Will not delete S3 local directory between runs."
+fi
+
+if [[ -n "$LUCENE_DIR" ]]; then
+    echo "Will delete lucene local directory between runs: $LUCENE_DIR"
```
chelma, line 59: Not a blocker, but - if we pre-delete the S3 directory (lines 52-23), why not pre-delete this one as well?

chelma (approved):
LGTM.  One question but I don't think it will substantively impact the application either way.

DocumentsFromSnapshotMigration/docker/entrypoint.sh
AndreKurait: removed lines 52-53. We'll run document migration once before deleting anything

AndreKurait (approved):

PR #1050, opened 2024-10-03
Add a script to determine the AWS resources used

peternied:
### Description
To help customers determine what resources are used by the Migration Assistant when deployed adding a shell script that can be used on the bootstrap machine to identify what there deployment will require based on the synthesized stacks.

Example output:
```shell
./aws-feature-usage.sh default
Synthesizing all stacks...
Synthesizing stack: networkStack-default
Synthesizing stack: migrationInfraStack
Synthesizing stack: reindexFromSnapshotStack
Synthesizing stack: migration-console
Finding resource usage from synthesized stacks...
-----------------------------------
IAM Policy Actions:
cloudwatch:GetMetricData
cloudwatch:ListMetrics
ec2:DescribeRouteTables
ec2:DescribeSubnets
ecr:BatchCheckLayerAvailability
...
-----------------------------------
Resources Types:
AWS::CDK::Metadata
AWS::EC2::InternetGateway
AWS::EC2::Route
AWS::EC2::RouteTable
AWS::EC2::SecurityGroup
...
```

### Testing
Ran locally, manually

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
@@ -144,10 +144,6 @@ export class StackComposer {
             throw new Error("Required context field 'contextId' not provided")
         }
         const contextJSON = this.parseContextBlock(scope, contextId)
-        console.log('Received following context block for deployment: ')
-        console.log(contextJSON)
-        console.log('End of context block.')
-
         const stage = this.getContextForType('stage', 'string', defaultValues, contextJSON)
 
         const domainName = this.getContextForType('domainName', 'string', defaultValues, contextJSON)
@@ -302,7 +298,6 @@ export class StackComposer {
         if (captureProxyServiceEnabled || captureProxyESServiceEnabled || trafficReplayerServiceEnabled || kafkaBrokerServiceEnabled) {
             streamingSourceType = determineStreamingSourceType(kafkaBrokerServiceEnabled)
         } else {
-            console.log("MSK is not enabled and will not be deployed.")
             streamingSourceType = StreamingSourceType.DISABLED
         }
 
--- a/deployment/cdk/opensearch-service-migration/.gitignore
+++ b/deployment/cdk/opensearch-service-migration/.gitignore
@@ -9,3 +9,4 @@ dist
 .cdk.staging
 cdk.out
 certs
+cdk-synth-output
--- a/deployment/cdk/opensearch-service-migration/aws-feature-usage.sh
+++ b/deployment/cdk/opensearch-service-migration/aws-feature-usage.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+set -e
+
+if [ -z "$1" ]; then
+    echo "Error: contextId is required. Please pass it as the first argument to the script."
+    echo "Usage: $0 <contextId>"
+    exit 1
+fi
+
+contextId="$1"
+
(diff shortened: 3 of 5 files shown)


deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -144,10 +144,6 @@ export class StackComposer {
             throw new Error("Required context field 'contextId' not provided")
         }
         const contextJSON = this.parseContextBlock(scope, contextId)
-        console.log('Received following context block for deployment: ')
-        console.log(contextJSON)
-        console.log('End of context block.')
```
mikaylathompson, line 149: Why are these being pulled out? They're pretty helpful for debugging

peternied: When calling `cdk list` or any other commands these messages were being printed out to the same console output that was preventing parsing.

Would you to want me to add this back in? If so I'd like to check for a flag that prints the output rather than doing it by default

lewijacn: This has been useful in the past for debugging. Is there a way we can turn this on only for deploy?

peternied: Roger that, I'll look into this.

peternied: I wasn't able to figure out a good 'switch' to print or not print this output, but instead I've added a deploy script, that allows the output to be printed and includes the other deployment defaults we use, example:

```
~/git/opensearch-migrations/deployment/cdk/opensearch-service-migration$ ./deploy.sh default
Context block for 'default':
---
{
   "stage": "dev",
   "targetCluster": {
      "endpoint": "http://TARGET_CLUSTER_ENDPOINT",
      "auth": {
         "type": "sigv4",
         "// basic auth documentation": "The next two lines are releavant for basic auth only",
         "username": "<USERNAME>",
         "passwordFromSecretArn": "<ARN_OF_SECRET_CONTAINING_PASSWORD>",
         "// sigv4 documentation": "The next two lines are releavant for sigv4 only",
         "region": "us-west-2",
         "serviceSigningName": "es"
      }
   },
   "sourceCluster": {
      "endpoint": "http://SOURCE_CLUSTER_ENDPOINT",
      "auth": {
         "type": "sigv4",
         "// sigv4 documentation": "The next two lines are releavant for sigv4 only",
         "region": "us-west-2",
         "serviceSigningName": "aoss"
      }
   },
   "reindexFromSnapshotServiceEnabled": true,
   "reindexFromSnapshotMaxShardSizeGiB": 80,
   "trafficReplayerServiceEnabled": false
}
---
```

lewijacn: I like for users who are accustomed to cdk to have the familiarity with seeing us recommend commands like `cdk <command>` at the top level instead of behind a script. I too haven't found a normalized way to do this in CDK though. What are you thoughts around having a different log type for these informational messages that is on by default but can be switched off when provided with some context variable like `disableAnalysisOutput`?

lewijacn (approved):

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
peternied: And you just approved - thanks!

PR #960, opened 2024-09-17
Add cluster version to yaml & rfs command

mikaylathompson:
### Description
Cluster version is an optional parameter for the `sourceCluster` and `targetCluster` objects in the CDK.

The source cluster version is passed (if provided) into the RFS command. It's also added to the metadata yaml and passed to the metadata command as `--source-version`.

### Issues Resolved
[MIGRATIONS-1951](https://opensearch.atlassian.net/browse/MIGRATIONS-1951)

### Testing
[Please provide details of testing done: unit testing, integration testing and manual testing]

### Check List
- [ ] New functionality includes testing
  - [x] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1051, opened 2024-10-07
Allow more versions to be in scope

peternied:
### Description
Loosen the ES 6.8 requirement to allow 6.X instead
Also fixes codesmell by replacing `peek` with `map` in stream usage.

### Issues Resolved
- Fixes https://opensearch.atlassian.net/browse/MIGRATIONS-2024

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
```diff
@@ -140,6 +143,9 @@ public static class TargetAdvancedArgs {
         @Parameter(names = {
             "--target-compression" }, description = "**Advanced**. Allow request compression to target", required = false)
         public boolean compressionEnabled = false;
+
+        @Parameter(names = { "--target-version-override" }, description = "**Advanced**. Allow overriding the version handling for the target cluster.", converter = VersionConverter.class)
```
AndreKurait, line 147: I'm not sure if this is an Advanced feature until 7.x is supported as a RemoteWriter or OS Compatibility flag is handled elsewhere

RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
AndreKurait, line 1: Let's update the readme at https://github.com/opensearch-project/opensearch-migrations/blob/main/DocumentsFromSnapshotMigration/README.md

RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
peternied: This feature should only be used when the target cluster has `override_main_response_version` enabled and metadata / rfs cannot trust the version number returned from the cluster, seems advanced to me.

AndreKurait, line 147: Before merging this, can we update the cdk to plum this to RFS workers since it already has the version key in the target_cluster object if specified

https://github.com/opensearch-project/opensearch-migrations/pull/960

AndreKurait (changes requested):

RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
peternied: I'm backing off exposing the target-version override for now, there are code smell fixes and the inclusion of ES 6.X that are still worthwhile to move forward with.

AndreKurait (approved):

RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
AndreKurait: do we have a jira for the target version override so we don't lose this feature?

peternied: Good call, created https://opensearch.atlassian.net/browse/MIGRATIONS-2101

PR #1054, opened 2024-10-08
Setup docker-compose like k8s deployment

gregschohn:
### Description

Longer term, Kubernetes (K8s) will allow us
1. To unify the single-node developer experience with and on prem experience and the cloud one.
2. Simplify and speed up deployment, especially around testing.
3. To enable auto-scaling

* Category Enhancement / New feature
* Why these changes are required? See above
* What is the old behavior before changes and new behavior after changes? CDK deployment should still work as it did before.

For this first cut...

### Installing

The k8s directory contains helm charts that can be installed with `helm install charts/aggregate/migrationAssistant`.  Optionally, `-n ma` can be added to add the helm installation and MA resources to the "ma" namespace (or whatever is specified).  aggregate/mockCustomerClusters is a helm chart for testing purposes to bring up the resources external to the migration assistant that are required to facilitate testing all of the migration assistant functionality.

Those aggregate charts pull other helm charts, either local to this repo (charts/components/(captureProxy|migrationConsole...) or 3rd party charts (opensearch, elasticsearch, prometheus, etc) into them.  Each chart defines default values in the values.yaml contained within the charts directory, alongside the Chart.yaml manifest file.  Those default values can be used to install a working version of the MA solution.  Today's functionality is limited and buggy, but doing a reasonable demo install w/out providing any value overrides will be an ongoing requirement.

Before running helm install, you'll likely need to run `helm dependency build CHART_DIR`, which downloads or copies charts into the CHART_DIR/charts directory.  There's a script (linkSubChartsToDependencies.sh), which still has some bugs, that attempts do do some of this manually w/ symlinks to local directories when possible.  Symlinks are preferable to tarballs because you don't need to keep rebuilding the dependencies, which can be time consuming.

### Configurations

Command line parameters are configured via helm through the values (such as values.yaml).  Here's sample contents for the bulkLoad deployment.
```
parameters:
  initialLeaseDuration:
    value: PT10M
  documentsPerBulkRequest:
    value: 1000
```

All of Migration Assistant's custom processes use the same paradigm to translate those yaml parameters into the command line parameters.  Shared helm charts (helmCommon) provide helper macros that allow consistent handling across our fleet of helm packages.  For command line parsing, helm charts create deployments, mostly via the common code that does the following.  1) Create config map stores for each of the parameters specified in the values.yaml file; 2) load the values from those config maps as environment variables into an init container; and then 3) that init container runs shell script commands to construct the arguments that will be passed into the main program.  For the last part, those arguments are passed via a file (vars.sh) within a shared mount and 4) the main container loads those variables before running the program.

The migration console has an extra init container that constructs the migration_services.yaml - which, like vars.sh is written into a shared container.  However, since pods can't refresh environment variables when config maps update and because we're bundling those into a single services yaml file, we have a separate init container to maintain that.  That container uses a custom shell script that uses the k8s client to watch config maps and formats the configured values themselves as yaml.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2287

### Testing
Manual testing only w/ minikube on my mac.  See `localTesting.sh` for some rough examples.

### Check List


TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/Pipfile
lewijacn, line 1: We also need a Pipfile.lock in this directory

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/services.yaml
lewijacn, line 1: We should probably revert changes to this file and have our own services.yaml we feed in for now as I'm not sure what impact this will cause

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -76,93 +76,93 @@ public static class Args {
         private boolean help;
 
         @Parameter(required = true,
-            names = { "--snapshot-name" },
+            names = { "--snapshot-name", "--snapshotName" },
```
peternied, line 79: This pattern of changes seems like it exposes a large risk for some commands to be missed which will create headaches when a parameter value isn't found.

We could ensure this behavior would be by wrapping the behavior at a higher level than on each individual parameter, or by having a way (test?) we are certain every parameter has a kabob-case and camelCase version.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/services.yaml
peternied, line 1: Revert unchanged file?

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/Dockerfile
peternied, line 1: Could this be created in a new high level location `${projectRoot}/images` or something?  If not lets move all the docker images out of this path since it makes a crazy dependency tree with the RFS worker images

peternied (commented):
Thanks for getting this out for review @gregschohn great work, still digesting the k8 directory

deployment/k8s/README.md
peternied, line 1: The project structure is throwing me a little bit here, I'm not sure why we need `.../k8s/charts/...`, it seems like we could get away with structuring, is there something I'm missing about what we put where?
```
deployment/
-> charts/
   -> README.md
   -> {chart deployment scripts}
   -> aggregates/...
   -> components/...
   -> sharedResources/...
```

deployment/k8s/aws/ack-resource-setup/.helmignore
peternied, line 1: Seems like the aws components are still in development, should these go in at the same time as everything else or can we pull them in when we use them?

peternied (approved):
While I've got outstanding comments, this is a great step forward thanks @gregschohn and @lewijacn

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
gregschohn: I'd love to investigate (& hopefully DO!) doing this at a higher level.  I honestly needed to try something out early on and each time that I had to do this, it was just ~5 minutes more - but in aggregate, it was considerable & the maintenance story sucks.
Here's a [jira](https://opensearch.atlassian.net/browse/MIGRATIONS-2314) to make this easier.

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/Dockerfile
gregschohn: I really, really, really, want to refactor all of these and lift them to the top directory.  I've already got a lot of scope creep at the moment & we have another set of changes for gradle at the moment that may have an impact on how we build any part of the system.

My plan is to do a lift in the near future - if only to be able to kill the buildDockerImages scripts in favor of `./gradlew buildDockerImages` at the root - and to be done with it!

I don't have a task for this, because the AC seems too squishy at the moment, but please add one if you have better specifics.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/services.yaml
gregschohn: +1 - dang.  I ALMOST completely reverted it :-P.  My bad (will add the last newline)

deployment/k8s/charts/components/captureProxy/templates/deployment.yaml
```diff
@@ -0,0 +1,45 @@
...
+        env: v1
+    spec:
+      initContainers:
+        {{- include "generic.setupEnvLoadInitContainer" (merge . (dict
+           "MountName" $mountName
+           "include" .Template.Include)) | nindent 8 }}
+        - name: wait-for-kafka
+          image: bitnami/kubectl:latest  # or any image with curl/kubectl
+          command: [ 'sh', '-c',
+                     'until kubectl wait --for=condition=Ready kafka/captured-traffic -n {{.Release.Namespace }} --timeout=10s; do echo waiting for kafka cluster is ready; sleep 1; done' ]
```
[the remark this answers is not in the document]
gregschohn: define different context?  Generally, yes - Kafka will need to be installed in the same namespace and it will need to have that specific name.  We'll want to improve this to support a bring-your-own-kafka at some point, but I don't know how big of a concern that is.  I'd like to have somebody ask for that.

deployment/k8s/charts/tests/testConsole/templates/deployment.yaml
```diff
@@ -0,0 +1,46 @@
...
+      serviceAccountName: configmap-watcher
+      initContainers:
+        {{- include "generic.setupEnvLoadInitContainer" (merge (dict
+              "MountName" $envVarMountName
+              "include" .Template.Include) .) | nindent 8 }}
+        {{- include "generic.setupServicesYamlContainer" (merge (dict
+                "MountName" "merged-config"
+                "include" .Template.Include) .) | nindent 8 }}
+      containers:
+        - name: console
```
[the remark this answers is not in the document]
gregschohn: done!

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/README.md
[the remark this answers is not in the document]
gregschohn: Yeah - that was early ideation. I'm removing it.  Once the shared config story is complete, I'll put a better README together.

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/configmap2yaml/config_watcher.py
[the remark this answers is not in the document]
gregschohn: config_watcher will at the very least change significantly once we start to extend the migration console to work directly with k8s.  At this point, config_watcher isn't really being tested (because we don't update config values at runtime).  Once we do - if we're still in need of services yaml, we'll want to put some integ tests together - and unit tests on the specific contract that we're converging to will also make sense.  At this point, the whole k8sConfigMapUtilScripts image should be considered an experimental bridge.

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/configmap2yaml/template.yaml.j2
```diff
@@ -0,0 +1,43 @@
...
+{%- if values.kafkaBroker is defined %}
+kafka:
+  {{ values | pop_value("kafkaBroker") | to_yaml | indent(2) }}
+{%- endif %}
+
+
+{%- if values %}
+extraValues:
+  {{ values | to_yaml | indent(2) }}
+{%- endif %}
```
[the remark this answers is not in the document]
gregschohn: This is here so that we don't lose anything - or so that we can find out what we missed.
I would consider READING and acting upon anything in extraValues a grave error, but writing to it seemed better than dropping configs on the floor.

deployment/k8s/README.md
gregschohn: Charts are a helm construct.  We'll also want to deploy EKS and AWS specific things for an EKS related deployment. We also might want tests, etc that are specific to k8s or helm.  Everything w/in charts today is specific to helm.  I can see it having siblings that are for k8s, but not specific to helm charts.  Does that help?

deployment/k8s/aws/ack-resource-setup/.helmignore
gregschohn: These were pulled forward from Tanner's branch.  Since I've merged everything else from it, I didn't want to remove it and then not be able to bring it back w/ full historical attribution.  I've considered labelling it experimental, etc - but that would somewhat imply that the other stuff wasn't.

I'm adding a README to explain this.

deployment/k8/.gitignore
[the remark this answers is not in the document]
lewijacn: I believe I had at least one section in the README.md that was referencing the `minikubeLocal.sh` script that I had. Since we have kept that script for now, I think we are fine

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/Dockerfile
peternied: Created this task for capturing and following up: [MIGRATIONS-2318](https://opensearch.atlassian.net/browse/MIGRATIONS-2318) | Move all docker container images into top level folder

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/configmap2yaml/config_watcher.py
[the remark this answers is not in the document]
peternied: I'm OK if we don't have a good test story so long as we add/reference a JIRA task that includes this testing as an AC.  What do you think?

TrafficCapture/dockerSolution/src/main/docker/k8sConfigMapUtilScripts/configmap2yaml/template.yaml.j2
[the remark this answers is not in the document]
peternied: Talked about this offline; this isn't meant to be a final design model for this template.  Lets iterate as we have a better picture of the system.

deployment/k8s/charts/components/captureProxy/templates/deployment.yaml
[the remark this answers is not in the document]
lewijacn: My main point here was that we probably need some way to provide the namespace that kafka is in otherwise it seems like it will fail if it continues to check its own namespace and kafka is in a different namespace

deployment/k8s/README.md
peternied: I certainly agree we might have AWS / non-AWS specific functionality.  I'm not sure I see the difference between k8s and helm, or to be more specific, I don't see k8s without helm.

When talking about support for other platforms my preference would be a single declaration for the canonical Migration Assistant.  Similarly there would be one component that represents the Migration Console.  Does this align with what you are thinking?

lewijacn (approved):

PR #1057, opened 2024-10-09
Increase the RFS default scale to 5

peternied:
### Description
After starting backfill it appears to be slow because a single worker node needs to spin up and down to process the default index with 5 shards.  By increasing the default this should move at least an index at a time causing the overall process to move much faster.

### Issues Resolved
- To be created

### Testing
Not tested

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_rfs.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_rfs.py
@@ -92,7 +92,7 @@ class ECSRFSBackfill(RFSBackfill):
         super().__init__(config)
         self.client_options = client_options
         self.target_cluster = target_cluster
-        self.default_scale = self.config["reindex_from_snapshot"].get("scale", 1)
+        self.default_scale = self.config["reindex_from_snapshot"].get("scale", 5)
 
         self.ecs_config = self.config["reindex_from_snapshot"]["ecs"]
         self.ecs_client = ECSService(cluster_name=self.ecs_config["cluster_name"],


AndreKurait:
Have you observed the same slowness after https://github.com/opensearch-project/opensearch-migrations/pull/1047 / 2.0.1?

peternied:
> Have you observed the same slowness after #1047 / 2.0.1?

Nope! Lets check in during standup if we think this is overkill.

mikaylathompson:
I don't think this is the right spot to change this

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_rfs.py
```diff
@@ -92,7 +92,7 @@ def __init__(self, config: Dict, target_cluster: Cluster, client_options: Option
         super().__init__(config)
         self.client_options = client_options
         self.target_cluster = target_cluster
-        self.default_scale = self.config["reindex_from_snapshot"].get("scale", 1)
```
mikaylathompson, line 95: This pulls `scale` from the services.yaml, let's use that and just set it to the right value in the yaml, which is built by the CDK.

Two places need to be updated:
- For whatever reason, we have the `scale` parameter defined in the Replayer YAML component in the CDK, but not the RFS YAML, so we should add it as a parameter there https://github.com/opensearch-project/opensearch-migrations/blob/main/deployment/cdk/opensearch-service-migration/lib/migration-services-yaml.ts#L64
- And then this is where a `scale` value should be added to that RFS yaml object: https://github.com/opensearch-project/opensearch-migrations/blob/main/deployment/cdk/opensearch-service-migration/lib/service-stacks/reindex-from-snapshot-stack.ts#L189

mikaylathompson (changes requested):

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/backfill_rfs.py
mikaylathompson: Or, per standup discussion, this could be added as a cdk parameter

peternied: Thanks for the pointer - I have changed both the default value in the cdk and I'd like to leave this default value to mirror it, any concerns with that approach?

deployment/cdk/opensearch-service-migration/lib/migration-services-yaml.ts
```diff
@@ -61,7 +61,7 @@ export class OSIBackfillYaml {
 
 export class ECSReplayerYaml {
     ecs: ECSService;
-    scale: number = 1;
```
mikaylathompson, line 64: ^ this is the replayer. We definitely don't want 5 replayers.

I was linking to this line to show what the RFS version should look like.

mikaylathompson (changes requested):

deployment/cdk/opensearch-service-migration/lib/migration-services-yaml.ts
mikaylathompson: I'd also somewhat rather the value is explicitly set in the RFS stack than in here -- if someone was looking to change it, that's where (I think) they'd look

mikaylathompson (approved):

AndreKurait (approved):
Ship It!

PR #1058, opened 2024-10-09
Make `--source-version` required

peternied:
### Description
By defaulting to ES 7.10 this caused confusion with the version parameter was not passed into RFS for different snapshot sources.  While this does require all customers to follow the same path it will prevent strange errors when the snapshot reader is unable to parse the snapshot.

### Issues Resolved
- Need to create

### Check List


deployment/cdk/opensearch-service-migration/cdk.context.json
AndreKurait, line 1: @peternied, i was looking at the options.md to see if it needed to be updated, and it looks like `version` has already been implemented as part of the sourceCluster and targetCluster above on a previous PR you reviewed.

Can we modify this implementation to suggest specifying it there and update the options.md to say it's required on the source and target if using compatibility mode?

It would also be nice to throw an error during parsing the sourceCluster object if it's not specified.

https://github.com/opensearch-project/opensearch-migrations/pull/960

AndreKurait (changes requested):

deployment/cdk/opensearch-service-migration/cdk.context.json
peternied: Thanks for looking into this, I've updated the options.md and examples to be correct following the format expected by the python system.

I did not throw an error on a missing value, it broke a sufficiently large number of test cases and I'd rather we reorganized how the validation of these parameters is handled that duplicate the logic in python typescript and java code in a future PR.

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -417,10 +417,10 @@ function parseAuth(json: any): ClusterAuth | null {
 
 export function parseClusterDefinition(json: any): ClusterYaml {
     const endpoint = json.endpoint
-    const version = json.version
     if (!endpoint) {
         throw new Error('Missing required field in cluster definition: endpoint')
     }
+    const version = json.version;
```
AndreKurait, line 423: can we do a check in stack-composer that this version exists on the source object since it's required so it's a synth failure?

peternied: I did not throw an error on a missing value, it broke a sufficiently large number of test cases and I'd rather we reorganized how the validation of these parameters is handled that duplicate the logic in python typescript and java code in a future PR.

AndreKurait (approved):

PR #1060, opened 2024-10-10
Controlling the Migration Assistant Control Plane

peternied:
### Description
I've taken a stab at writing a design, making changes to get metadata migration pulling from the services.yaml, updating how MA console handles tool arguments, and looking for some edge cases such as RFS where we do considerable work with parameters before the java process.

This was done quickly in the spirit of creating a discussions around interesting areas.

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/docker/entrypoint.sh
+++ b/DocumentsFromSnapshotMigration/docker/entrypoint.sh
@@ -3,78 +3,22 @@
 # Fail the script if any command fails
 set -e
 
-# Print our ENV variables
-if [[ $RFS_COMMAND != *"--target-password"* ]]; then
-    echo "RFS_COMMAND: $RFS_COMMAND"
-else
-    echo "RFS Target Cluster password found in RFS_COMMAND; skipping logging of the value"
-fi
+# Discussion needed: Container will have a minimum number of
+# parameters after on start services.yaml is loaded.
+#
+# This script was parsing the parameters that were feed into the
+# container, unless this script is pulling parameters from the
+# services.yaml its going to be out of sync. 
+#
+# Metadata & Snapshot will already need to know how to read the 
+# Secrets from ARNs so this seems like it aligns well to do this
+# at the same time.
 
-echo "RFS_TARGET_USER: $RFS_TARGET_USER"
-echo "RFS_TARGET_PASSWORD: <redacted>"
-echo "RFS_TARGET_PASSWORD_ARN: $RFS_TARGET_PASSWORD_ARN"
-
-# Check if the RFS Command already contains a username; only do special work if it does not
-if [[ $RFS_COMMAND != *"--target-username"* ]]; then
-    if [[ -n "$RFS_TARGET_USER" ]]; then
-        echo "Using username from ENV variable RFS_TARGET_USER.  Updating RFS Command with username."
-        RFS_COMMAND="$RFS_COMMAND --target-username \"$RFS_TARGET_USER\""
-    fi
-fi
-
-# Check if the RFS Command already contains a password; only do special work if it does not
-if [[ $RFS_COMMAND != *"--target-password"* ]]; then
-    PASSWORD_TO_USE=""
-
-    # Check if the password is available in plaintext; if, use it.  Otherwise, retrieve it from AWS Secrets Manager
-    if [[ -n "$RFS_TARGET_PASSWORD" ]]; then
-        echo "Using plaintext password from ENV variable RFS_TARGET_PASSWORD"
-        PASSWORD_TO_USE="$RFS_TARGET_PASSWORD"
-    elif [[ -n "$RFS_TARGET_PASSWORD_ARN" ]]; then
-        # Retrieve password from AWS Secrets Manager if ARN is provided
-        echo "Using password from AWS Secrets Manager ARN in ENV variable RFS_TARGET_PASSWORD_ARN"
-        PASSWORD_TO_USE=$(aws secretsmanager get-secret-value --secret-id "$RFS_TARGET_PASSWORD_ARN" --query SecretString --output text)
-    fi
-
-    # Append the username/password to the RFS Command if have an updated password
-    if [[ -n "$PASSWORD_TO_USE" ]]; then
-        echo "Updating RFS Command with password."
-        RFS_COMMAND="$RFS_COMMAND --target-password \"$PASSWORD_TO_USE\""
-    fi
-fi
-
-# Extract the value passed after --s3-local-dir
-S3_LOCAL_DIR=$(echo "$RFS_COMMAND" | sed -n 's/.*--s3-local-dir\s\+\("[^"]\+"\|[^ ]\+\).*/\1/p' | tr -d '"')
-# Extract the value passed after --lucene-dir
-LUCENE_DIR=$(echo "$RFS_COMMAND"  | sed -n 's/.*--lucene-dir\s\+\("[^"]\+"\|[^ ]\+\).*/\1/p' | tr -d '"')
-if [[ -n "$S3_LOCAL_DIR" ]]; then
-    echo "Will delete S3 local directory between runs: $S3_LOCAL_DIR"
-else
-    echo "--s3-local-dir argument not found in RFS_COMMAND. Will not delete S3 local directory between runs."
-fi
-
-if [[ -n "$LUCENE_DIR" ]]; then
-    echo "Will delete lucene local directory between runs: $LUCENE_DIR"
-else
-    echo "--lucene-dir argument not found in RFS_COMMAND. This is required."
-    exit 1
-fi
-
-cleanup_directories() {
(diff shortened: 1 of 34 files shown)


DocumentsFromSnapshotMigration/docker/entrypoint.sh
```diff
@@ -3,85 +3,28 @@
...
-
-    if [[ -n "$LUCENE_DIR" ]]; then
-        echo "Cleaning up Lucene local directory: $LUCENE_DIR"
-        rm -rf "$LUCENE_DIR"
-        echo "Directory $LUCENE_DIR has been cleaned up."
-    fi
-}
-
-
+# Discussion needed: Container will have a minimum number of
```
peternied, line 6: 🧵 For Discussion

DocumentsFromSnapshotMigration/docker/entrypoint.sh
```diff
@@ -3,85 +3,28 @@
...
+#
+# This script was parsing the parameters that were feed into the
+# container, unless this script is pulling parameters from the
+# services.yaml its going to be out of sync. 
+#
+# Metadata & Snapshot will already need to know how to read the 
+# Secrets from ARNs so this seems like it aligns well to do this
+# at the same time.
+
+# Discussion needed: Directly cleanup was being done by script based
```
peternied, line 17: 🧵 For Discussion

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
```diff
@@ -22,15 +23,62 @@
...
             .addObject(metadataArgs)
             .addCommand(migrateArgs)
             .addCommand(evaluateArgs)
-            .build();
-        jCommander.parse(args);
+            .build()
+            .parse(args);
+
+        // Then override with settings
+        var config = getConfig(args);
```
peternied, line 41: 🧵 For Discussion

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/metadata.py
```diff
@@ -62,108 +27,23 @@ def generate_tmp_dir(name: str) -> str:
...
-            if is_command(arg) and is_value(next_arg):
-                commands[arg] = next_arg
-                i += 2  # Move past the command and value
-            elif is_command(arg):
-                commands[arg] = None
-                i += 1  # Move past the command, its a flag
-            else:
-                logger.warning(f"Ignoring extra value {arg}, there was no command name before it")
-                i += 1
+    # Discussion Needed: Verification of the config should be done by
```
peternied, line 41: 🧵 For Discussion

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/metadata.py
```diff
@@ -13,47 +13,12 @@
...
-        }
-    },
-    # We _should_ have the check below, but I need to figure out how to combine it with a potentially
-    # nullable block (like this one)
-    # 'check_with': contains_one_of({'s3', 'fs'})
-
-}
-
 SCHEMA = {
-    "from_snapshot": FROM_SNAPSHOT_SCHEMA,
```
peternied, line 50: I think we might need to reorganize the layout of the existing services.yaml, e.g. the snapshot definition seems like it should be separate from the snapshot tool

gregschohn (changes requested):
Please don't merge this.  It creates a strong dependency between the migration console's global (union'ed) schema and every single one of our projects.  It forces people embedding just one program to need to know about all others.  It requires a formatting change within the migration console to be a recompile for the application layers.  The idea of the command line parameters is that they're self-documenting.  The boundary / surface exposed by each program is well-defined.

This tax of maintaining parameters lets us also describe the (minimal) interfaces and requirements automatically.  It also makes the tools much more accessible (think about the governance proxies that have been forked).  I think that this project has more interesting tasks if we think about independent components rather than bundling/coupling them together.

coreUtilities/build.gradle
```diff
@@ -41,6 +41,7 @@ dependencies {
 
     // JCommander
     compileOnly group: 'org.jcommander', name: 'jcommander'
+    implementation group: 'org.yaml', name: 'snakeyaml'
```
gregschohn, line 44: why do you need snakeyaml over jackson?  They both have a healthy stream of CVE's associated with them, but AFAIK currently, we don't have this as a dependency and we do have a dependency in pretty much every project on jackson.  Can we keep out maintenance costs down & stick to only jackson for yaml needs too?

coreUtilities/src/main/java/org/opensearch/migrations/config/Backfill.java
```diff
@@ -0,0 +1,7 @@
+package org.opensearch.migrations.config;
```
gregschohn, line 1: When I did the instrumentation work, I wanted to make sure that a change needed for one project wouldn't impact in any way any other project that didn't need to consume the change.  At the every-day level, this includes builds and build graphs that are as tight as possible.  It also self-documents the extent that datastructures and patterns are consumed.  If I see Backfill and Cluster, I'm going to presume that every single project that uses parameters will need these.  In fact though, some significant services don't - like the proxy or cluster have no use for many of these and never will.

We have a single repo for this project, but I would pretend that a different person were in the sole maintainer for each project.  If that was the case, would this aggregation be worth the overhead?

peternied: I agree that any change in coreUtilities requires a rebuild.  And as I understand it its purpose, we put shared functionality between all components in our migration ecosystem.  If the services.yaml is used by all the projects in the solution including the java projects, the place that seems logical to put this shared parsing code would be in the coreUtilities project.

I believe the relationship is worthwhile to be sure that all components can read the same services.yaml definition in the project.  How else would you have a shared definition, where else should it go?

coreUtilities/build.gradle
peternied: This is a prototype of what it would take to move in this direction where I quickly jumped to snakeyaml, make sense to use jackson when we implement the change.

gregschohn:
I'm continuing to try to wrap my head around this.  From the diagrams, it looks like there are two concerns being conflated.  The first is the separation of deployment vs runtime parameters.  The second is how applications are configured at runtime.

For the first case, as part of the K8s work, we're looking to redesign the deployment/runtime experience and split deployment parameters (vpcIds, service names) from runtime parameters (scale values, transforms to use, etc).  We can also make it so that there are two tiers of deployment parameters so that a change in target endpoint, etc doesn't cause the rest of the resources (console, replayer, etc) to need to be deployed if not applicable.  We can also determine what's best for the user and what should be deferred until runtime and is mutable vs what is set at deployment time and immutable and maybe what's set at deployment time but mutable.

I see the migration console (orchestrator from our architecture docs) managing all of this.  Each application takes what it needs from that orchestrator that understands every interface and how to stitch them together.

The second issue here is how to manage the datamodel for the migration console and the consistency between the orchestrator and its worker nodes.  The proposal here is unappealing to me because it doesn't scale immediately since every application needs to be aware of every other applications' interfaces.  Today, we have a yaml database (services.yaml) keep the union of all the configurations and the orchestrator uses that.  If we'd want to make it so that that file can be passed and unknown arguments were ignored, that's probably fine.  But I wouldn't want to see that functionality within the application depend upon specific infrastructure, like requiring a shared fs or access to secrets manager, etc.  That should go into the container definition for the application or outside of it.

If we want to make sure that the models are synchronized between the layers, we should build up model classes (like many web-services do) for each application and have the migration console take dependencies on all of those & concatenate them into one services.yaml spec.  That eliminates the need for everybody knowing about everything and still gets the migration console a full-view.

I'd be happy to see us continue to use some shared libraries for things like transformation configuration like what we're doing [here](https://github.com/opensearch-project/opensearch-migrations/blob/85d5846216002abfc8a437dc41356347d82b5a18/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java#L141-L145).  We can continue to extend the functionality of our parameter passing logic be easier to share & to have more features - like being able to read from a file or command line parameters (there's a @file annotation w/ JCommander).

gregschohn:
I think there's a good intention & I think there's a path forward, but I also doubt if this is a top priority over cleaning up the overall deployment story (K8s), scaling our services, or putting extensible transformations in place so that users can do more.  I don't see how we can derive enough short term benefit to warrant an investment here right now for the first point.  I haven't noticed a huge cost yet for the second point and it feels like it's mostly where it needs to be right now.

peternied:
This prototype served its purpose, had discussion with the team, notes are captured on the JIRA https://opensearch.atlassian.net/browse/MIGRATIONS-2075

Will follow up with this when it is critical path to future efforts.

PR #1074, opened 2024-10-14
Expand target support to  OpenSearch v1.3

peternied:
### Description
Expand target support to OpenSearch v1.3

Updated the java end to end test cases to read from the same list for
target platforms.

### Check List


gregschohn (approved):
This LGTM.  Since there's no documentation change that says that we support OS1.x in this PR, this change by itself makes the codebase better & I'm happy to approve it.

However, I'd like to see one more case added to the ParallelDocumentMigrationsTest (it's already setup to take a list of targets and do parameterized testing, so it's one line).  That test will take into account any differences that might occur in work coordination due to running on a different version of the cluster doing the work coordination.

PR #1076, opened 2024-10-16
Address code smells in DataGenerator

peternied:
### Description
Address code smells in DataGenerator
- Defined constants for reused literals
- Mark unmodified fields final
- Add a private constructor to hide the implicit public constructor

### Check List

### DIFF
--- a/DataGenerator/src/main/java/org/opensearch/migrations/data/workloads/Geonames.java
+++ b/DataGenerator/src/main/java/org/opensearch/migrations/data/workloads/Geonames.java
@@ -9,6 +9,9 @@ import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+import static org.opensearch.migrations.data.FieldBuilders.GEO_POINT;
+import static org.opensearch.migrations.data.FieldBuilders.INTEGER;
+import static org.opensearch.migrations.data.FieldBuilders.LONG;
 import static org.opensearch.migrations.data.FieldBuilders.createField;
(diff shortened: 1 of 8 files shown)


gregschohn (commented):
UtilityClass is one annotation I'd rather us not use.  It's value is marginal and it requires me to review the entire class/file, which is especially scary when looking at a diff since a reviewer won't see the whole file by default.

peternied:
> UtilityClass is one annotation I'd rather us not use. It's value is marginal and it requires me to review the entire class/file, which is especially scary when looking at a diff since a reviewer won't see the whole file by default.

Thanks for the review, I have addressed this comment.

DataGenerator/src/main/java/org/opensearch/migrations/data/workloads/Geonames.java
```diff
@@ -9,6 +9,9 @@
 import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
+import static org.opensearch.migrations.data.FieldBuilders.GEO_POINT;
+import static org.opensearch.migrations.data.FieldBuilders.INTEGER;
+import static org.opensearch.migrations.data.FieldBuilders.LONG;
```
gregschohn, line 14: Let's continue to not allow static imports outside of test code (this is in src, not testFixture)
Adding the classname will give names like "INTEGER" some more context and clarity

peternied: Thanks for the feedback, I don't agree with the style preference.  I would prefer to merge this change as is. 

I've created MIGRATIONS-2141 [1] to automate enforcement of style issues, lets use tooling to align our preferences.

- [1] https://opensearch.atlassian.net/browse/MIGRATIONS-2141

lewijacn (approved):

Issue #633, opened 2024-05-04
[BUG] Slf4j/Log4j2 Log Argument Formatting


AndreKurait:
### What is the bug?
Currently, when running the replayer, Log4j2 outputs `Unable to format msg` along with `java.lang.IllegalArgumentException: found 1 argument placeholders, but provided 0 for pattern`

### How can one reproduce the bug?
I've created unit tests to simulate the behavior in various instances. Including it here below: As you can see, it happens whenever objects are included in the "setMessage" function, but not when placed in the addArgument.

```java
package org.opensearch.migrations.replay;

import java.util.ArrayList;
import java.util.Collections;
import java.util.List;
import org.apache.logging.log4j.Level;
import org.apache.logging.log4j.LogManager;
import org.apache.logging.log4j.core.LogEvent;
import org.apache.logging.log4j.core.appender.AbstractAppender;
import org.apache.logging.log4j.core.impl.Log4jContextFactory;
import org.apache.logging.log4j.core.selector.ClassLoaderContextSelector;
import org.apache.logging.log4j.status.StatusData;
import org.apache.logging.log4j.status.StatusListener;
import org.apache.logging.log4j.status.StatusLogger;
import org.junit.jupiter.api.Assertions;
import org.junit.jupiter.api.Test;
import org.slf4j.LoggerFactory;

class LoggingTest {
    static {
        // Synchronize logging for assertions
        LogManager.setFactory(new Log4jContextFactory(new ClassLoaderContextSelector()));
    }

    @Test
    public void testLogInline_internalLoggerError() {
        try (var closeableLogSetup = new CloseableLogSetup()) {
            closeableLogSetup.testLogger.atInfo().setMessage("hello {} hello").log();
            Assertions.assertEquals(1, closeableLogSetup.logEvents.size());

            var contents = closeableLogSetup.logEvents.get(0);
            Assertions.assertEquals("hello {} hello", contents);

            var errorContents = closeableLogSetup.errorEvents.get(0);
            System.err.println(errorContents);
            Assertions.assertTrue(errorContents.contains("Unable to format msg") && errorContents.contains("java.lang.IllegalArgumentException"));
        }
    }

    @Test
    public void testLogSupplier_internalLoggerError() {
        try (var closeableLogSetup = new CloseableLogSetup()) {
            closeableLogSetup.testLogger.atInfo().setMessage(() -> "hello {} hello").log();
[...]

PR #1084, opened 2024-10-19
Reformat log statements so that they aren't susceptible to formatting errors and worse when {} are in the message

gregschohn:
### Description

This converts and canonicalizes all of our existing log usage for all java projects.  The rules followed are

* setCause() should be the first thing after setting the log level.
* setMessage() should never do any formatting & therefore never needs to create a lambda.
* addArgument() should only create lambdas if a method reference won't work (e.g. `() -> foo.get()` should be replaced with `foo:get`).
* addArguments and log() should be on separate lines if they can't fit onto one line.

See https://opensearch.atlassian.net/browse/MIGRATIONS-2161 for follow-up work to make these changes more durable.

---

* Category Bug fix
* Why these changes are required? For security and reliability.
* What is the old behavior before changes and new behavior after changes? In logging, if {} was within the contents of some strings being logged as messages, those would have thrown Exceptions.  Now "{}" will print out instead.  A class of security bugs should also be closed down with this change.

### Issues Resolved
[Jira issue](https://opensearch.atlassian.net/browse/MIGRATIONS-2121)
[Github Issue](https://github.com/opensearch-project/opensearch-migrations/issues/633)

### Testing
gradle testing

### Check List


DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -343,16 +343,10 @@ private static void confirmShardPrepIsComplete(
...
-                        () -> "After "
-                            + finalShardSetupAttemptNumber
-                            + "another process holds the lock"
-                            + " for setting up the shard work items.  "
-                            + "Waiting "
-                            + finalLockRenegotiationMillis
-                            + "ms before trying again."
-                    )
+                log.atInfo().setMessage("{}")
+                    .addArgument(() -> "After " + finalShardSetupAttemptNumber + "another process holds the lock"
```
AndreKurait, line 347: let's change this to use constant string with argument insertion like the rest of the log lines

AndreKurait (approved):
Please clean up the pr description and i'm good with this.

PR #1099, opened 2024-10-23
Add Basic Jenkins Integ Test for Demo Setup

lewijacn:
### Description
This change introduces a new Jenkins workflow for a full Migration workflow (Metadata, Backfill, C&R) using the Capture Proxy and Target Proxy ECS services and ALB as opposed to the previous workflow of having the capture proxy installed directly on the source nodes.

This is a basic integration test in the sense that it is only moving a few key documents that get added in different parts of the workflow and is verifying source and target connections within the ALB and outside of it. In its current state it should add a lot of value for checking this setup, but within the Epic of the Jira Issue linked, there are other tasks for adding more interesting data and checking an ALB cutover from source to target that would be great additions

Also updates were added after discussion to combine this with the previous ES 6.8 RFS only test and make this the default PR test that is executed

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2149

### Testing
~~Jenkins testing here: https://migrations.ci.opensearch.org/job/full-default-e2e-test/19/console~~
Update test here: https://migrations.ci.opensearch.org/job/full-es68source-e2e-test/4/

### Check List


jenkins/migrationIntegPipelines/fullDefaultE2ETestCover.groovy
```diff
@@ -0,0 +1,9 @@
+def gitBranch = params.GIT_BRANCH ?: 'main'
```
chelma, line 1: Why make a new Pipeline for this instead of use our existing ES 6.8 -> OS 2.X Pipeline?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/full_tests.py
```diff
@@ -25,6 +25,18 @@ def initialize(request):
     pytest.unique_id = unique_id
     source_cluster = pytest.console_env.source_cluster
     target_cluster = pytest.console_env.target_cluster
+    # If in AWS, modify source and target objects here to route requests through the created ALB to verify its operation
```
chelma, line 28: Wait - it's possible to run these tests \*not\* on AWS?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/full_tests.py
```diff
@@ -25,6 +25,18 @@ def initialize(request):
...
+    if 'AWS_EXECUTION_ENV' in os.environ:
+        logger.info("Detected an AWS environment")
+        stage = request.config.getoption("--stage")
+        logger.info("Checking original source and target endpoints can be reached, before using ALB endpoints for test")
+        direct_source_con_result: ConnectionResult = connection_check(source_cluster)
+        assert direct_source_con_result.connection_established is True
+        direct_target_con_result: ConnectionResult = connection_check(target_cluster)
+        assert direct_target_con_result.connection_established is True
+        source_cluster.endpoint = f"https://alb.migration.{stage}.local:9201"
+        target_cluster.endpoint = f"https://alb.migration.{stage}.local:9202"
```
chelma, line 38: This hardcoding is pretty fragile; is there a way we could pass these domains into the test?

lewijacn (commented):
Just pushed an update and responded to comments. Updated Jenkins run: https://migrations.ci.opensearch.org/job/full-default-e2e-test/20/

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/full_tests.py
lewijacn: Removed and passing in when running test command

jenkins/migrationIntegPipelines/fullDefaultE2ETestCover.groovy
chelma: A few thoughts:
* Why wouldn't we want to use a pre-made snapshot for all of our integ tests?
* Why don't we want to test both RFS and C&R for each of our integ tests?
* `see if both pipelines or a single pipeline is failing` -> Can you elaborate a bit more on this one?

lewijacn: I think its helpful for testing our workflow where the user performs the snapshot creation themselves from the migration console

In a similar vein, I like testing RFS and C&R together and separately as it affirms that these are separate independent pieces that can operate on their own, and that we haven't introduced some logic that say requires a snapshot to exist when we are only doing a C&R scenario

chelma: Per discussion, my takeaways:
* Combine w/ the existing ES 6.8 test
* Update GitHub actions to use this instead of the existing 7.10 RFS only test

lewijacn: Have updated this PR to include both these items. See recent Jenkins build here: https://migrations.ci.opensearch.org/job/full-es68source-e2e-test/4/

chelma (approved):
Thanks, Tanner!

PR #1104, opened 2024-10-25
Create RFS successor work items

mikaylathompson:
### Description
One component of sub-shard RFS is creating a mechanism to robustly create new work items from a given parent.
This PR implements a function in the work coordinator to take a parent work item id, and a list of one or more successor item ids.

`createSuccessorItemsAndMarkComplete` is called with the current work item id, and a list of successor item ids. It does the following:
1. format the list of successor item ids into a comma separated list and update the current work item's `successor_items` list, subject to the following criteria:
   a. client is using the correct script version
   b. the worker writing still holds the lease
   c. the work item's current `successor_items` list is either not set or the same as the one being pushed.
2. Send a bulk request to create each of the successor items as a new work item. This uses a `create` call that will only create the new items if their ids don't exist (e.g. it won't overwrite an item with the same id).
3. If the following step succeeds with only 201 (created) or 409 (already exists) error codes, mark the current item as completed.

The checks built in make this function idempotent. If it fails at any point, it can be run again without creating any inconsistent or conflicting state. Indeed, one of the expected behaviors is that if another work picks up a lease where 1 and any portion of 2 were completed, it will immediately re-run the function with the same parameters and this will attempt again to create the successors and complete the process.

If the first step (updating the item with the `successor_items` list) fails and the worker exits, the mechanism fails and the progress will be lost (TODO: I should have much more robust retries around this step).

It fits into the broader workflow as follows:
1. A work item is claimed, as currently done.
2. If it already has `successor_items`, jump to step 4.
3. Write documents until the lease has nearly expired or a SIGTERM signal is received.
4. Call `createSuccessorItemsAndMarkComplete` with a single item list containing a work item that defines the current shard starting where this worker is finishing.
    a. This creates one new work item, and marks the current one as completed.

In the future, this workflow can be modified to split the remaining work into multiple shards instead of one, or to split before starting a given shard, instead of after (e.g. a set of workers first runs through and splits every shard into sub-shards, and then new workers actually tackle writing the documents).

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2127

### Testing
Tests are added that test the basic behavior, the behavior in a contentious situation (40 simultaneous workers), and the error-recovery behavior (e.g. that the function can be run idempotently). There are also some tests of error-checking behavior (e.g. attempting to supply a new `successor_items` list.

I'd like to add a toxi-proxy test, a test (and check) to prevent adding one's self as a successor work item, and retries around setting the initial `successor_items` list.

### Check List


RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+        }
+
+        try (var workCoordinator = new OpenSearchWorkCoordinator(httpClientSupplier.get(), 3600, "claimItemWorker")) {
+            for (var i = 0; i < NUM_DOCS; ++i) {
+                String workItemId = getWorkItemAndVerify(testContext, "claimItemWorker", new ConcurrentHashMap<>(), Duration.ofSeconds(5), false, false);
+                var currentNumPendingItems = workCoordinator.numWorkItemsNotYetComplete(testContext::createItemsPendingContext);
+                ArrayList<String> successorWorkItems = new ArrayList<>();
+                for (int j = 0; j < NUM_SUCCESSOR_ITEMS; j++) {
+                    successorWorkItems.add(workItemId + "_successor_" + j);
+                }
```
gregschohn, line 205: nit: one liner for IntStream.range(0, NUM_SUCCESSOR_ITEMS).mapToObj(i->workItemId + "_successor_" + i).toList(); // or maybe collect(...toList)

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+                workCoordinator.createSuccessorWorkItemsAndMarkComplete(
+                        workItemId, successorWorkItems,
+                        testContext::createSuccessorWorkItemsContext
+                );
+                Assertions.assertTrue(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+                // One item marked as completed, and NUM_SUCCESSOR_ITEMS created.
+                Assertions.assertEquals(currentNumPendingItems - 1 + NUM_SUCCESSOR_ITEMS, workCoordinator.numWorkItemsNotYetComplete(testContext::createItemsPendingContext));
+            }
+
+            Assertions.assertEquals(NUM_SUCCESSOR_ITEMS * NUM_DOCS, workCoordinator.numWorkItemsNotYetComplete(testContext::createItemsPendingContext));
```
gregschohn, line 215: I'd love to see a loop after this make sure that we can get the NUM_SUCCESSOR_ITEMS * NUM_DOCS work items - to prove that the successor work items are completely valid and assignable work items.

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+
+        var originalWorkItemExpiration = Duration.ofSeconds(5);
+        try (var workCoordinator = new OpenSearchWorkCoordinator(httpClientSupplier.get(), 3600, "successorTest")) {
+            Assertions.assertFalse(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+            workCoordinator.createUnassignedWorkItem(docId, testContext::createUnassignedWorkContext);
+            Assertions.assertTrue(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+            // Claim the work item
+            getWorkItemAndVerify(testContext, "successorTest", new ConcurrentHashMap<>(), originalWorkItemExpiration, false, false);
+            var client = httpClientSupplier.get();
+            // Add tbe list of successors to the work item
```
gregschohn, line 273: typo 'tbe'

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+        }
+
+        var originalWorkItemExpiration = Duration.ofSeconds(5);
+        try (var workCoordinator = new OpenSearchWorkCoordinator(httpClientSupplier.get(), 3600, "successorTest")) {
+            Assertions.assertFalse(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+            workCoordinator.createUnassignedWorkItem(docId, testContext::createUnassignedWorkContext);
+            Assertions.assertTrue(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+            // Claim the work item
+            getWorkItemAndVerify(testContext, "successorTest", new ConcurrentHashMap<>(), originalWorkItemExpiration, false, false);
+            var client = httpClientSupplier.get();
```
gregschohn, line 272: would be nice to have an assertFalse(...workItemsNotYetComplete) for clarity.

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+                Assertions.assertTrue(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+                // One item marked as completed, and NUM_SUCCESSOR_ITEMS created.
+                Assertions.assertEquals(currentNumPendingItems - 1 + NUM_SUCCESSOR_ITEMS, workCoordinator.numWorkItemsNotYetComplete(testContext::createItemsPendingContext));
+            }
+
+            Assertions.assertEquals(NUM_SUCCESSOR_ITEMS * NUM_DOCS, workCoordinator.numWorkItemsNotYetComplete(testContext::createItemsPendingContext));
+        }
+    }
+
+    @Test
```
gregschohn, line 219: Which of these new tests are slowTests (and should be marked w/ the longTest annotation)?  look to see what the longest tests of the non-slow tests are running at and use that as your limit.

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+            // Okay, we're now in a state where the only document available is the original, incomplete one.
+            // We need to make sure that if we pick it up and call `createSuccessorWorkItemsAndMarkComplete`, it will complete successfully and create the two missing items.
+            var originalWorkItemId = getWorkItemAndVerify(testContext, "successorTest", new ConcurrentHashMap<>(), originalWorkItemExpiration, false, false);
+            Assertions.assertEquals(docId, originalWorkItemId);
+            workCoordinator.createSuccessorWorkItemsAndMarkComplete(originalWorkItemId, successorItems, testContext::createSuccessorWorkItemsContext);
+
+            // Now, we should be able to claim the other successor items but the _next_ call should fail because there are no available items
+            for (int i = 0; i < (N_SUCCESSOR_ITEMS - 1); i++) {
+                workItemId = getWorkItemAndVerify(testContext, "claimItem_" + i, new ConcurrentHashMap<>(), originalWorkItemExpiration, false, true);
+                Assertions.assertTrue(successorItems.contains(workItemId));
```
gregschohn, line 296: you need a slightly stronger set of tests here to confirm that you're not getting duplicates (reusing a ConcurrentHashMap above and taking the count will do that).

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+            }
+
+            Assertions.assertThrows(IllegalStateException.class, () -> {
```
gregschohn, line 299: Can you extend IllegalStateException and throw subtypes.  I don't know why the code within will throw.
You can also store the value from assertThrow & do some further checks (instanceOf, getMessage().contains(...)).

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+            var client = httpClientSupplier.get();
+            // Add an INCORRECT list of successors to the work item
+            var incorrectSuccessors = "successor_99,successor_98,successor_97";
+            var body = "{\"doc\": {\"successor_items\": \"" + incorrectSuccessors + "\"}}";
+            var response = client.makeJsonRequest("POST", ".migrations_working_state/_update/" + docId, null, body);
+            var responseBody = (new ObjectMapper()).readTree(response.getPayloadBytes());
+            Assertions.assertEquals(200, response.getStatusCode());
+
+            // Now attempt to go through with the correct successor item list
+            Assertions.assertThrows(IllegalArgumentException.class,
```
gregschohn, line 335: I know why this is throwing - but it would be nice to see some extra check that we know why...
I think that at its root this is from the server/painless script - so maybe just scan the getMessage() contents to confirm that it's helpful.

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+
+        try (var workCoordinator = new OpenSearchWorkCoordinator(httpClientSupplier.get(), 3600, "successorTest")) {
+            Assertions.assertFalse(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+            workCoordinator.createUnassignedWorkItem(docId, testContext::createUnassignedWorkContext);
+            Assertions.assertTrue(workCoordinator.workItemsNotYetComplete(testContext::createItemsPendingContext));
+            // Claim the work item
+            getWorkItemAndVerify(testContext, "successorTest", new ConcurrentHashMap<>(), Duration.ofSeconds(5), false, false);
+
+            // Now attempt to go through with the correct successor item list
+            Assertions.assertThrows(IllegalArgumentException.class,
```
gregschohn, line 358: This is completely the right behavior - but I can't recall from looking at the code... why will this fail?

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
```diff
@@ -180,12 +181,224 @@ public void testAcquireLeaseForQuery() throws Exception {
...
+            // We should NOT be able to claim the other successor items yet (since they haven't been created yet) or the original item
+            String workItemId = getWorkItemAndVerify(testContext, "claimSuccessorItem", new ConcurrentHashMap<>(), Duration.ofSeconds(60), false, false);
+            Assertions.assertEquals(successorItems.get(0), workItemId); // We need to ensure that the item we just claimed is the expected one.
+
+            // Sleep for the remainder of the original work item's lease so that it becomes available.
+            Thread.sleep(originalWorkItemExpiration.toMillis() + 1000);
+
+            // Okay, we're now in a state where the only document available is the original, incomplete one.
+            // We need to make sure that if we pick it up and call `createSuccessorWorkItemsAndMarkComplete`, it will complete successfully and create the two missing items.
+            var originalWorkItemId = getWorkItemAndVerify(testContext, "successorTest", new ConcurrentHashMap<>(), originalWorkItemExpiration, false, false);
```
gregschohn, line 289: As in another comment, I think that these lines should be transparently handled.  This test still has a lot of value, you'll just shed a few lines and make sure that you eventually fill up the "seenWorkIds" map.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -645,6 +665,145 @@ private WorkItemAndDuration getAssignedWorkItem(LeaseChecker leaseChecker,
...
+                + "      \"successorWorkItems\": \"" + SUCCESSOR_WORK_ITEM_IDS_TEMPLATE + "\"\n"
+                + "    },\n"
+                + "    \"source\": \""
+                + "      if (ctx._source.scriptVersion != \\\"" + SCRIPT_VERSION_TEMPLATE + "\\\") {"
+                + "        throw new IllegalArgumentException(\\\"scriptVersion mismatch.  Not all participants are using the same script: sourceVersion=\\\" + ctx.source.scriptVersion);"
+                + "      }"
+                + "      if (ctx._source." + LEASE_HOLDER_ID_FIELD_NAME + " != params.workerId) {"
+                + "        throw new IllegalArgumentException(\\\"work item was owned by \\\" + ctx._source."
+                +                        LEASE_HOLDER_ID_FIELD_NAME + " + \\\" not \\\" + params.workerId);"
+                + "      }"
```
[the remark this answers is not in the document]
mikaylathompson: Discussed with Greg -- because of the extra check on line 685, there's a risk that this could be picked up again, _but_ it will never cause this to spiral out of control. The second worker to grab it will never be able to create its own successor work items

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -228,14 +240,12 @@ AbstractedHttpClient.AbstractHttpResponse createOrUpdateLeaseForDocument(
             + "        throw new IllegalArgumentException(\\\"The current times indicated between the client and server are too different.\\\");"
             + "      }"
             + "      long newExpiration = params.clientTimestamp + (((long)Math.pow(2, ctx._source.numAttempts)) * params.expirationWindow);"
-            + "      if (params.expirationWindow > 0 && "
-            +                 // don't obtain a lease lock
-            "          ctx._source." + COMPLETED_AT_FIELD_NAME + " == null) {"
-            +              // already done
-            "        if (ctx._source." + LEASE_HOLDER_ID_FIELD_NAME + " == params.workerId && "
+            + "      if (params.expirationWindow > 0 && ctx._source." + COMPLETED_AT_FIELD_NAME + " == null) {"
+            +          // already done
```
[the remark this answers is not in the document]
mikaylathompson: I cannot figure out what this comment meant. I agree that it sure looks to me like this is a not-yet-done work item if we're at this point. I wrote out a somewhat more complete version of that.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -645,6 +665,145 @@ private WorkItemAndDuration getAssignedWorkItem(LeaseChecker leaseChecker,
...
+                                    (long) (Math.pow(2.0, (updateWorkItemRetries-1)) * CREATE_SUCCESSOR_WORK_ITEMS_RETRY_BASE_MS));
+                    ctx.addTraceException(e, false);
+                    log.atWarn().setCause(e)
+                            .setMessage("Couldn't complete work assignment due to exception. Backing off {} and trying again.")
+                            .addArgument(sleepBeforeNextRetryDuration).log();
+                    Thread.sleep(sleepBeforeNextRetryDuration.toMillis());
+                }
+            }
+
+            createUnassignedWorkItemsIfNonexistent(successorWorkItemIds);
```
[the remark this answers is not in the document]
mikaylathompson: I don't think it should continue indefinitely here. The first step is the critical one. As long as that first step has finished, the next worker to pick this up will just pick up where it left off in trying to create the successorWorkItems. This process is expected to be run from the `exitOnLeaseTermination` function, so it has a limited amount of time to work with before another worker will pick up the lease. If the cluster is running very slowly or with a high error rate, it's possible/likely that another worker will pick up this lease during the retries and attempt to create these successor work items at the same time. Due to the structure of the bulk request, that won't actually cause any problems, but it's extra work for both the workers and the cluster.

I think it's safer to give this a small number of attempts (I'm leaning towards one, but open to maybe 3?) and then move on so the task gets picked up by a different worker later.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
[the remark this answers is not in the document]
gregschohn: There may be a race condition in all of these calls since our final write isn't confirming that the document that we're writing was the same version.  I don't expect that we'll be using this datastore for the long term to do transactional coordination.  As Mikayla points out, this is a bounded efficiency problem and won't affect correctness, so this seems like a calculated risk that isn't worth putting more effort into at this point.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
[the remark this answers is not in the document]
gregschohn: Take a look at the counts for the top, I agree that it doesn't need to be indefinite since we'll die and another process can eventually drive. this to completion, but there IS a minor cost to not doing the splits ASAP (since workers can't grab the new work items, they might be spinning down & take longer to come back up, etc).

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -645,6 +733,167 @@ private WorkItemAndDuration getAssignedWorkItem(LeaseChecker leaseChecker,
...
+                                + workItemId
+                                + ".  Response: "
+                                + response.toDiagnosticString()
+                );
+            }
+        } catch (IllegalArgumentException e) {
+            var resultTree = objectMapper.readTree(response.getPayloadBytes());
+            System.out.println("resultTree = " + resultTree);
+            System.out.println(resultTree.get("error"));
+            System.out.println(resultTree.get("error").get("type"));
```
gregschohn, line 786: please remove these

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -935,7 +1031,13 @@ private void refresh(Supplier<IWorkCoordinationContexts.IRefreshContext> context
                     switch (obtainResult) {
                         case SUCCESSFUL_ACQUISITION:
                             ctx.recordAssigned();
-                            return getAssignedWorkItem(leaseChecker, ctx);
+                            var workItem = getAssignedWorkItem(leaseChecker, ctx);
+                            if (workItem.successorWorkItemIds != null) {
+                                // continue the previous work of creating the successors and marking this item as completed
+                                createSuccessorWorkItemsAndMarkComplete(workItem.workItemId, workItem.successorWorkItemIds, ctx::getCreateSuccessorWorkItemsContext);
+                                return new AlreadyCompleted();
```
gregschohn, line 1038: What's the downside to continuing?  AlreadyCompleted is a weird response for acquireNextWorkItem().  Why would it have been acquired if it was already done?

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
gregschohn: to discuss - why does this throw an illegal state exception?  That's a scary sounding exception.  The calls don't look that scary.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -114,6 +145,44 @@ public OpenSearchWorkCoordinator(
...
+                log.atError().setCause(e)
+                        .setMessage("Couldn't complete action due to a non-retryable exception.")
+                        .log();
+                Exception underlyingException = (Exception) e.getCause();
+                exceptionConsumer.accept(underlyingException);
+                throw new IllegalStateException(underlyingException);
+            } catch (Exception e) {
+                attempt++;
+                if (attempt > maxRetries) {
+                    log.atError().setCause(e)
```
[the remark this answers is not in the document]
mikaylathompson: I'm keeping the one with the retry times because that is opaque to the caller. But otherwise agreed.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -138,6 +207,10 @@ public void setup(Supplier<IWorkCoordinationContexts.IInitializeCoordinatorState
             + "      \"status\": {\n"
             + "        \"type\": \"keyword\",\n"
             + "        \"norms\": false\n"
+            + "      },\n"
+            + "     \"" + SUCCESSOR_ITEMS_FIELD_NAME + "\": {\n"
+            + "       \"type\": \"keyword\",\n"
```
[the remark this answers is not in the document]
mikaylathompson: I _think_ adding successor items would be okay, due to dynamic type mapping, but that's a setting that could be changed, and it's helpful for pulling this value out that I know that the field will always be there.

gregschohn (approved):

PR #847, opened 2024-07-26
DockerHub Build and Tag on Release

AndreKurait:
### Description
Adding a workflow on our release action to build and tag the docker images. I've commented out the section for engineering-effectiveness to add their secrets for the staging account.

Currently only builds the architecture the runner is on, will add cross arch build in a followup pr.

* Category: Enhancement
* Why these changes are required? Publishing to docker enables our solution to be consumed by other partners without building from source
* What is the old behavior before changes and new behavior after changes? Release action will build docker images, no publishing occurring at this step

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1570

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
Tested in GHA https://github.com/opensearch-project/opensearch-migrations/actions/runs/10115830331/job/27977434308

### Check List
- [x] New functionality includes testing
  - [x] All tests pass, including unit test, integration test and doctest
- [x] New functionality has been documented
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1118, opened 2024-11-05
Auto detect target cluster version

peternied:
### Description
Adding automatic version detection that supports both flavors of the Amazon Managed OpenSearch and Serverless offerings.  Also detecting if combability mode has been enabled that will extract the version number from the `_cat/plugins` api, specifically looking at the `repository-s3` plugin.

### Issues Resolved
- Resolves https://opensearch.atlassian.net/browse/MIGRATIONS-2187

### Testing
Tested with both a managed and serverless cluster locally.

```
> ./gradlew MetadataMigration:run --args='evaluate --source-host https://search-target-domain-p5keoxlvhl5slkqgildik64mce.us-west-2.es.amazonaws.com/ --source-aws-region us-west-2 --source-aws-service-signing-name es --target-host https://vwpt0d13nhrsiinwng90.us-west-2.aoss.amazonaws.com/ --target-aws-region us-west-2 --target-aws-service-signing-name aoss'

> Task :MetadataMigration:run
Starting Metadata Evaluation
Clusters:
   Source:
      Remote Cluster: AMAZON_MANAGED_OPENSEARCH 2.13.0 ConnectionContext(uri=https://search-target-domain-p5keoxlvhl5slkqgildik64mce.us-west-2.es.amazonaws.com/, protocol=HTTPS, insecure=false, compressionSupported=false, awsSpecificAuthentication=true)

   Target:
      Remote Cluster: AMAZON_SERVERLESS_OPENSEARCH 2.0.0 ConnectionContext(uri=https://vwpt0d13nhrsiinwng90.us-west-2.aoss.amazonaws.com/, protocol=HTTPS, insecure=false, compressionSupported=false, awsSpecificAuthentication=true)
````

### Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
@@ -4,6 +4,7 @@ import java.io.IOException;
 import java.net.HttpURLConnection;
 import java.time.Duration;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -14,6 +15,7 @@ import java.util.stream.Collectors;
 import org.opensearch.migrations.Flavor;
 import org.opensearch.migrations.Version;
 import org.opensearch.migrations.VersionMatchers;
+import org.opensearch.migrations.bulkload.common.OpenSearchClient.OperationFailed;
 import org.opensearch.migrations.bulkload.common.http.ConnectionContext;
 import org.opensearch.migrations.bulkload.common.http.HttpResponse;
 import org.opensearch.migrations.bulkload.tracing.IRfsContexts;
@@ -23,7 +25,6 @@ import org.opensearch.migrations.reindexer.FailedRequestsLogger;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import lombok.extern.slf4j.Slf4j;
 import reactor.core.publisher.Mono;
@@ -107,8 +108,8 @@ public class OpenSearchClient {
                 if (!hasCompatibilityModeEnabled) {
                     return Mono.just(versionFromRootApi);
                 }
-                return client.getAsync("_cat/plugins?format=json", null)
-                    .flatMap(this::getVersionFromPlugins)
+                return client.getAsync("_nodes/_all/nodes,version?format=json", null)
+                    .flatMap(this::getVersionFromNodes)
                     .doOnError(e -> log.error(e.getMessage()))
                     .retryWhen(CHECK_IF_ITEM_EXISTS_RETRY_STRATEGY);
             })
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java [hunk 1]
```diff
@@ -99,17 +129,76 @@ private Version versionFromResponse(HttpResponse resp) {
...
-            return versionBuilder.build();
+            return Mono.just(versionBuilder.build());
         } catch (Exception e) {
             log.error("Unable to parse version from response", e);
-            throw new OperationFailed("Unable to parse version from response: " + e.getMessage(), resp);
+            return Mono.error(new OperationFailed("Unable to parse version from response: " + e.getMessage(), resp));
         }
     }
 
+    Mono<Boolean> checkCompatibilityModeFromResponse(HttpResponse resp) {
```
chelma, line 143: Non-blocking nit-pick: May I suggest re-naming this to `isClusterInCompatibilityMode()`?  I think naming methods that answer a boolean question as a question helps make their intent clearer.  The current name is unclear on what the outcome/side-effect of the check is (IMO).

transformation/src/test/java/org/opensearch/migrations/FlavorTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+        var falseCases = List.of(Flavor.ELASTICSEARCH);
+        for (var testCase : falseCases) {
+            assertThat(testCase.isOpenSearch(), equalTo(false));
+        }
+    }
+
+    @Test
+    void uniqueValues() {
+        // Make sure values in the flavor enum are never marked as equal to one another
```
peternied, line 29: One could consider updating the `equal` in Flavor to match between different flavors, but this could cause all kinds of havok.  We need to ensure the uniqueness of Flavors, whereas `VersionMatchers` handles the similarity between these flavors.

transformation/src/main/java/org/opensearch/migrations/Version.java
```diff
@@ -27,6 +28,7 @@ public static Version fromString(final String raw) throws RuntimeException {
 
         var finalRemainingString = remainingString;
         var matchedFlavor = Arrays.stream(Flavor.values())
+            .sorted(Comparator.comparing((Flavor f) -> f.shorthand.length()).reversed())
```
peternied, line 31: Need to check for `AOSS` before `AOS` because the matching process checks that version string starts with this parameter, added a unit test `parseAllPossibleNames` to confirm that all parsing paths are viable.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
```diff
@@ -69,23 +79,43 @@ protected OpenSearchClient(RestClient client, FailedRequestsLogger failedRequest
...
-                    return Mono.just(versionFromResponse(resp));
-                } catch (Exception e) {
-                    return Mono.error(new OperationFailed(e.getMessage(), resp));
+                if (resp.statusCode == 200) {
+                    return versionFromResponse(resp);
+                }
+                // If the root API doesn't exist, the cluster is OpenSearch Serverless
+                if (resp.statusCode == 404) {
+                    return Mono.just(AMAZON_SERVERLESS_VERSION);
                 }
```
chelma, line 90: This is a pretty non-intuitive conclusion to reach just from performing a GET on the default REST path.  How confident are we that this is a durable condition?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
```diff
@@ -99,17 +129,76 @@ private Version versionFromResponse(HttpResponse resp) {
...
+            var body = (ArrayNode) objectMapper.readTree(resp.body);
+            var elemIterator = body.elements();
+            while (elemIterator.hasNext()) {
+                var current = elemIterator.next();
+                var isRepositoryS3 = Optional.ofNullable(current.get("component"))
+                    .filter(n -> !n.isNull())
+                    .map(n -> n.asText())
+                    .map(text -> "repository-s3".equals(text))
+                    .orElse(false);
+                if (isRepositoryS3) {
```
chelma, line 182: What if they don't have the S3 Repo plugin installed?

chelma (commented):
This process for finding the "correct" version of the source/target cluster feels complex and brittle, and I'm not sure I'm comfortable with it.  It seems like we're adding a fair amount of very specific logic around features of a cluster that may be different than we expect for not all that much gain.  I'm concerned we're creating code that will just break later at an unexpected moment when we could just have the user input the source/target version manually and avoid this entire process.

I'm curious to hear about the tradeoffs here, as you see them?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
peternied: Pretty confident.  The root API was intentionally removed from OpenSearch Serverless, so adding it back in would be very unexpected, it surfaces a lot of information that would be hard to rationalize - there is no option to get this information today.

I've created [MIGRATIONS-2209](https://opensearch.atlassian.net/browse/MIGRATIONS-2209) to track using an api and internally made a request for an API that indicates the wire compatibility version, when I've got the receipt for that request I'll add it to that task.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
chelma: Interesting - a few follow-ups, then:
* Are we thinking this API will only be called against AOS or AOSS clusters as opposed to self-managed?
* Because the plugin version can be anything, where are we gaining confidence that it will continue to truly report the server version?  Is this best-effort convention or something more durable?

peternied: Compatibility mode is a feature in the open-source, see [#847](https://github.com/opensearch-project/OpenSearch/pull/847), it can be enabled on self-managed clusters or amazon managed OpenSearch cluster.  

There are two reason that I trust repository-s3 plugin version value:
1) Repository-s3 plugin is one that is not modified for the Amazon deployment, its deployed as is (yay!).  Many other plugins are modified for special functionality in the AWS ecosystem (hence `unspecified` and `x.x.x.x` versions).
2) Repository-s3 plugin is built for bundling in the OpenSearch distributions where its version number will match OpenSearch's version number.

Maybe we tweak this over time after better APIs are added, but this seems like the best way to infer this information sans bug reports.

gregschohn (commented):
Thanks for this.  I know that we need something ASAP.  I didn't see this go through refinement, so sorry for suggesting some design changes late in the game...

To detect the version for clusters (rather than serverless), could we use something like `https://localhost:9200/_nodes/_all/nodes,version`?  

We'll need this functionality (very soon) to determine how to transform metadata from a source version to a target version, right?  That problem isn't constrained to just one java program.  It isn't even just constrained to our suite of java programs.  Even though, we'll likely host (at least) most of our transformations through those, setting them up will be done outside of those processes, in a web page or a CLI.

Could we bake auto-detection logic as some python, or maybe typescript so that it can be easily run from a webpage or our CLI?  The whole discovery feature could be a chain of two items - (request to run, pattern to extract the version) - keep working through the list until one matches.  That could then calculated on-demand or cached in a config if we like.  It would be pretty straightforward and easy for others to patch if necessary (as I'm sure that we'll find out new patterns as we go, especially for new targets).

peternied:
@gregschohn Thanks for the great suggestion, the nodes API is much more trust worthy IMO.  I've swapped out the plugins api for it, it was a relatively simple change.

For reusing this code elsewhere, I think we have a complicated story about how that can work today.  I'd prefer to move forward with this implementation and we can figure out how and where to make this logic available in other parts of the migration ecosystem and what the best way to do that is.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
peternied: Actually, nix this thread, I've switched to using the more reliable _cat/nodes api that also surfaces this information.

chelma:
> I'd prefer to move forward with this implementation and we can figure out how and where to make this logic available in other parts of the migration ecosystem

That makes sense to me, thanks for the work on this @peternied .  Thanks also to @gregschohn  for the suggestion on the Nodes API, it alleviates a lot of my concern about brittleness.

chelma (approved):
Thanks, Peter!

PR #1128, opened 2024-11-12
Metadata Support ES7+ Legacy Templates

AndreKurait:
### Description
Currently, legacy templates (created with `_template`) were only supported on ES 6. Since the API is deprecated and not removed, customers still use this type of template for ES 7 and above and it should be supported and tested.

This change mirrors the existing transformation logic that is in Index processing and applying to Legacy templates for ES7. E.g.
```
 * Starting state:
 * {
 *   "mappings": [
 *     {
 *       "foo": {
 *         "properties": {
 *           "field1": { "type": "text" },
 *           "field2": { "type": "keyword" }
 *         }
 *       }
 *     }
 *   ]
 * }
 *
 * Ending state:
 * {
 *   "mappings": {
 *     "properties": {
 *       "field1": { "type": "text" },
 *       "field2": { "type": "keyword" },
 *     }
 *   }
 * }
```

* Category: Bug fix
* Why these changes are required? Legacy template support on ES7+
* What is the old behavior before changes and new behavior after changes? Legacy templates wouldn't work and parse correctly.

### Issues Resolved
[MIGRATIONS-2217](https://opensearch.atlassian.net/browse/MIGRATIONS-2217)

### Testing
Added E2E tests across different criteria for legacy templates on all versions with snapshot and api retrieval

### Check List

### DIFF
--- a/RFS/src/test/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_7_10_OS_2_11Test.java
+++ b/RFS/src/test/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_7_10_OS_2_11Test.java
@@ -31,7 +31,7 @@ public class Transformer_ES_7_10_OS_2_11Test {
         GlobalMetadata transformedGlobalMetadata = transformer.transformGlobalMetadata(globalMetadata);
         GlobalMetadataData_OS_2_11 finalMetadata = new GlobalMetadataData_OS_2_11(transformedGlobalMetadata.toObjectNode());
 
-        String expectedBwcTemplates = "{\"bwc_template\":{\"order\":0,\"index_patterns\":[\"bwc_index*\"],\"settings\":{\"number_of_shards\":\"1\",\"number_of_replicas\":\"0\"},\"mappings\":[{\"arbitrary_type\":{\"properties\":{\"title\":{\"type\":\"text\"},\"content\":{\"type\":\"text\"}}}}],\"aliases\":{\"bwc_alias\":{}}}}";
+        String expectedBwcTemplates = "{\"bwc_template\":{\"order\":0,\"index_patterns\":[\"bwc_index*\"],\"settings\":{\"number_of_shards\":\"1\",\"number_of_replicas\":\"0\"},\"mappings\":{\"properties\":{\"title\":{\"type\":\"text\"},\"content\":{\"type\":\"text\"}}},\"aliases\":{\"bwc_alias\":{}}}}";
         String expectedIndexTemplates = "{\"fwc_template\":{\"index_patterns\":[\"fwc_index*\"],\"template\":{\"aliases\":{\"fwc_alias\":{}}},\"composed_of\":[\"fwc_mappings\",\"fwc_settings\"]}}";
         String expectedComponentTemplates = "{\"fwc_settings\":{\"template\":{\"settings\":{\"index\":{\"number_of_shards\":\"1\",\"number_of_replicas\":\"0\"}}}},\"fwc_mappings\":{\"template\":{\"mappings\":{\"properties\":{\"title\":{\"type\":\"text\"},\"content\":{\"type\":\"text\"}}}}}}";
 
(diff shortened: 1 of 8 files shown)


RFS/src/test/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_7_10_OS_2_11Test.java
```diff
@@ -31,7 +31,7 @@ public void transformGlobalMetadata_AsExpected() throws Exception {
         GlobalMetadata transformedGlobalMetadata = transformer.transformGlobalMetadata(globalMetadata);
         GlobalMetadataData_OS_2_11 finalMetadata = new GlobalMetadataData_OS_2_11(transformedGlobalMetadata.toObjectNode());
 
-        String expectedBwcTemplates = "{\"bwc_template\":{\"order\":0,\"index_patterns\":[\"bwc_index*\"],\"settings\":{\"number_of_shards\":\"1\",\"number_of_replicas\":\"0\"},\"mappings\":[{\"arbitrary_type\":{\"properties\":{\"title\":{\"type\":\"text\"},\"content\":{\"type\":\"text\"}}}}],\"aliases\":{\"bwc_alias\":{}}}}";
+        String expectedBwcTemplates = "{\"bwc_template\":{\"order\":0,\"index_patterns\":[\"bwc_index*\"],\"settings\":{\"number_of_shards\":\"1\",\"number_of_replicas\":\"0\"},\"mappings\":{\"properties\":{\"title\":{\"type\":\"text\"},\"content\":{\"type\":\"text\"}}},\"aliases\":{\"bwc_alias\":{}}}}";
```
gregschohn, line 34: This was existing code, but is this data easier or harder to read when it's on one line/not pretty printed?  For me as a reviewer, it's challenging.

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/http/ClusterOperations.java
```diff
@@ -158,7 +158,7 @@ public void createES6LegacyTemplate(final String templateName, final String patt
             "  }\r\n" + //
             "}";
 
-        final var createRepoRequest = new HttpPut(this.clusterUrl + "/_template/" + templateName);
+        final var createRepoRequest = new HttpPut(this.clusterUrl + "/_template/" + templateName + "?include_type_name=true");
```
gregschohn, line 161: It seems like this will create a template w/ some extra layers in the json.  From there, in the e2e test, we're trying to show that the intermediate type, which was hopefully there, no longer is, is that right?

Putting some assertions on the original template to show that it exhibited the behavior that you're trying to tickle would be helpful.

gregschohn (commented):
Could you also please provide the before/after of what the template from legacy to latest?  That could be in the PR or (better), the commit message.

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/http/ClusterOperations.java
AndreKurait: There's no behavior change with `?include_type_name=true` in this operation, we're just accepting that we're intentionally including the type in the put which will not log the warning message for this operation.

AndreKurait:
> Could you also please provide the before/after of what the template from legacy to latest? That could be in the PR or (better), the commit message.

Added context in the PR Overview. This isn't net-new transformation or functionality, just doing the same processing on the templates `mapping` key that we apply in ES7 indexes (and ES6 templates/indexes)

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/http/ClusterOperations.java
gregschohn: Code is parameterized, so putting a sanity check to assert that intermediate type mappings are present isn't feasible to do cleanly.

gregschohn (approved):

PR #1133, opened 2024-11-13
Capture the created role arn for snapshotting into the AWS yaml parameter and pass it through the command line

gregschohn:
### Description
Specifying 'managedServiceSourceSnapshotEnabled' in the cdk.context.json will now not just create the role for making a snapshot to s3, but it will add it to the services yaml file and the console will thread that through to the command line for snapshot create so that users don't have to.

* Category Enhancement

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2001

### Testing
Manual deployments w/ and without the cdk.context.json value and w/ and w/out the value in migration_services.yaml

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/snapshot.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/snapshot.py
@@ -23,6 +23,7 @@ SNAPSHOT_SCHEMA = {
                 'schema': {
                     'repo_uri': {'type': 'string', 'required': True},
                     'aws_region': {'type': 'string', 'required': True},
+                    'role': {'type': 'string', 'required': False}
                 }
             },
             'fs': {
@@ -109,6 +110,7 @@ class S3Snapshot(Snapshot):
         self.snapshot_name = config['snapshot_name']
         self.otel_endpoint = config.get("otel_endpoint", None)
         self.s3_repo_uri = config['s3']['repo_uri']
+        self.s3_role_arn = config['s3'].get('role')
         self.s3_region = config['s3']['aws_region']
 
     def create(self, *args, **kwargs) -> CommandResult:
@@ -131,6 +133,8 @@ class S3Snapshot(Snapshot):
             command_args["--no-wait"] = FlagOnlyArgument
         if max_snapshot_rate_mb_per_node is not None:
             command_args["--max-snapshot-rate-mb-per-node"] = max_snapshot_rate_mb_per_node
+        if self.s3_role_arn:
+            command_args["--s3-role-arn"] = self.s3_role_arn
         if extra_args:
             for arg in extra_args:
                 command_args[arg] = FlagOnlyArgument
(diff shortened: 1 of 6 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/snapshot.py
```diff
@@ -23,6 +23,7 @@
                 'schema': {
```
mikaylathompson, line 23: Ideally there would be two tests for this:
- one on the python side (if `role` is in the services.yaml, it should be included in the call to the java snapshot tool, a la [this test](https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_snapshot.py#L290) and the ones nearby)
- and one on the CDK side, where we set the `managedServiceSourceSnapshotEnabled` and make sure the policy gets created / added to the YAML (kind of like [this one](https://github.com/opensearch-project/opensearch-migrations/blob/main/deployment/cdk/opensearch-service-migration/test/reindex-from-snapshot-stack.test.ts#L311).

The python test is a low lift with decent helpful value to effectively document how role should effect the rest of the system. The CDK test is a much higher lift that should have been done in my previous PR, but wasn't. And is potentially going to be replaced in the near-ish future.

So I'd really like the python test, the CDK test is optional.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/snapshot.py
```diff
@@ -23,6 +23,7 @@
...
                     'repo_uri': {'type': 'string', 'required': True},
                     'aws_region': {'type': 'string', 'required': True},
+                    'role': {'type': 'string', 'required': False}
```
mikaylathompson, line 26: Can you document this here: https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/README.md#snapshot

deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -275,11 +266,37 @@ export class MigrationConsoleStack extends MigrationServiceCore {
...
         }
 
+        if (props.managedServiceSourceSnapshotEnabled &&
+            servicesYaml.snapshot &&
+            servicesYaml.snapshot.s3) {
+            servicesYaml.snapshot.s3.role =
+                createSnapshotOnAOSRole(this, artifactS3Arn, serviceTaskRole.roleArn,
+                    this.region, props.stage, props.defaultDeployId)
+                    .roleArn;
+            // HOW DO I GET THIS servicesYaml into the parameter AND get the service created?
```
mikaylathompson, line 282: I'm not sure what this question is -- what do you mean by "get the service created"?

mikaylathompson (approved):

PR #1147, opened 2024-11-19
Added initial CloudWatch Dashboard for RFS

chelma:
### Description
* Added a CloudWatch Dashboard for RFS that gets created when the RFS feature is enabled
* The Target Cluster section relies on having metrics provided by an Amazon OpenSearch Service domain to populate.  Additionally, the user must supply the name of their domain in the Dashboard.
* One quirk is the mechanic we're using to represent how many "active" RFS Workers there are.  We don't have a good way to represent the number of \*desired\* workers in a steady-state sense, so what we do is effectively report how many RFS Workers were alive in a given period by counting the number of separate instances of a CPU metric for our ECS containers.  During steady-state operation, we'd expect the metric in that graph to resemble closely the user's desired worker count.  However, if the workers are dying frequently (such as if there's no work), then new workers will be constantly popping up, reporting in, and dying.  Each such worker adds a new entry to this sample count, and so the number skews higher than the desired number.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-2132

### Testing
* Created the dashboard in my account and used to visualize a few simple migrations:

Screenshot 2024-11-19 at 9 13 26 AM

Screenshot 2024-11-19 at 9 13 42 AM

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/components/reindex-from-snapshot-dashboard.json
+++ b/deployment/cdk/opensearch-service-migration/lib/components/reindex-from-snapshot-dashboard.json
@@ -1,0 +1,292 @@
+{
+    "variables": [
+        {
+            "type": "property",
+            "property": "region",
+            "inputType": "input",
+            "id": "REGION",
+            "label": "Region",
+            "defaultValue": "us-east-1",
+            "visible": false
+        },
+        {
+            "type": "property",
+            "property": "DomainName",
+            "inputType": "input",
+            "id": "TC_DOMAIN_NAME",
+            "label": "Target Cluster Domain Name",
+            "defaultValue": "placeholder-name",
+            "visible": true
+        },
+        {
+            "type": "pattern",
+            "pattern": "MA_STAGE",
+            "inputType": "input",
+            "id": "MA_STAGE",
+            "label": "Migration Assistant Stage",
+            "defaultValue": "placeholder-stage",
+            "visible": false
+        },
+        {
+            "type": "pattern",
+            "pattern": "ACCOUNT_ID",
+            "inputType": "input",
+            "id": "ACCOUNT_ID",
+            "label": "ACCOUNT_ID",
+            "defaultValue": "ACCOUNT_ID",
+            "visible": false
+        }
+    ],
+    "widgets": [
+        {
+            "height": 1,
+            "width": 24,
+            "y": 0,
+            "x": 0,
+            "type": "text",
+            "properties": {
+                "markdown": "# Target Cluster\n",
+                "background": "transparent"
+            }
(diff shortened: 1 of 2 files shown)


deployment/cdk/opensearch-service-migration/lib/components/reindex-from-snapshot-dashboard.json
```diff
@@ -0,0 +1,295 @@
...
+            "height": 8,
+            "width": 12,
+            "y": 1,
+            "x": 0,
+            "type": "metric",
+            "properties": {
+                "view": "timeSeries",
+                "stacked": false,
+                "metrics": [
+                    [ { "expression": "METRICS()/1000", "id": "e1", "region": "REGION" } ],
```
AndreKurait, line 64: Should we need to do the same `/ PERIOD(m1)*60` here

deployment/cdk/opensearch-service-migration/lib/components/reindex-from-snapshot-dashboard.json
```diff
@@ -0,0 +1,295 @@
...
+                    [ "AWS/ES", "IndexingRate", "DomainName", "TC_DOMAIN_NAME", "ClientId", "ACCOUNT_ID", { "region": "region", "label": "Document Ingested per 60 seconds - MIN: ${MIN}, MAX: ${MAX}, AVG: ${AVG}", "id": "m1", "visible": false } ]
```
AndreKurait, line 65: Let's add a (including replicas) here on the label

deployment/cdk/opensearch-service-migration/lib/components/reindex-from-snapshot-dashboard.json
```diff
@@ -0,0 +1,295 @@
...
+            "type": "metric",
+            "properties": {
+                "metrics": [
+                    [ { "expression": "METRICS()/PERIOD(m1)*60", "id" : "e1", "region": "REGION" } ],
+                    [ "AWS/ECS", "CPUUtilization", "ServiceName", "migration-MA_STAGE-reindex-from-snapshot", "ClusterName", "migration-MA_STAGE-ecs-cluster", { "region": "REGION", "label": "RFS Workers - MIN - ${MIN}, MAX - ${MAX}, AVG - ${AVG}", "id": "m1", "visible": false } ]
+                ],
+                "period": 60,
+                "region": "REGION",
+                "stacked": false,
+                "title": "RFS Workers Reporting in During Period",
```
AndreKurait, line 251: nit: should we remove `During Period` from the title here?

AndreKurait (approved):

PR #1154, opened 2024-11-20
Type mappings transformation

gregschohn:
### Description

* Category - Enhancement, New feature
* Why these changes are required? - To allow users to more easily upgrade from an <= ES 6.8 cluster by remapping type mappings to new indices.
* What is the old behavior before changes and new behavior after changes?  New behaviors are only present if the type mapping sanitization transformer is specified at the command line of the replayer (**TODO** fill in the invocation string later).

These transformations are implemented via jinjava transformations that are configured with a couple maps as described below.  Notice that each type of request can be suppressed via feature flags, which are also an option that can be passed to the provider.  That allows users to opt-in to some transformations that might have value while not being forced to adopt every type of request transformation that this package does.  Users still have the ability to chain these transformations alongside any others using the existing transformation plugin framework.

From jsonTypeMappingsSanitizationTransformer/README.md
---
The structure of the documents and indices need to change.  Some options are to use separate indices, drop some of 
the types to make an index single-purpose, or to create an index that's the union of all the types' fields.

With a simple mapping directive, we can define each of these three behaviors.  The following yaml shows how to map 
documents into two different indices named users and posts: 
```
activity:
  user: new_users
  post: new_posts
```

To drop one type, just leave it out:
```
activity:
  user: only_users
```

To merge types together, use the same value:
```
activity:
  user: any_activity
  post: any_activity
```

Any _indices_ that are NOT specified won't be modified - all additions, changes, and queries on those other indices not
specified at the root level will remain untouched by the static mapping rewriter.  However, missing types from a 
specified index _**will**_ be removed.  To remove ALL the activity for a given index, specify an empty index with no
children types.
```
activity: {}
```

In addition to static source/target mappings, users can specify source and type pairs as a regex and use any captured 
groups in the target index name.  Regex rules take precedent _after_ the static rules and are only applied when there 
was no index match in the static mappings.  

Regex replacement is controlled via an ordered list of `[indexNamePattern, typeNamePattern, replacementString]`.  
The transformer will use the replacement for the first matched item found.  
If none are found, unlike missing indices for static mappings, the system presumes that the index and type are 
**NOT** to be propagated to the target - any reference to those types and their corresponding data will be suppressed.  
To preserve all items, a default rule will need to be included.  

The following sample shows how indices that start with 'time-' will be migrated and every other index and type not
already matched will be dropped.
```
[
  ["time-(.*)", "(cpu)", "time-$1-$2"]
]
```

The following example preserves all other non-matched items, 
merging all types into a single index with the same name as the source index.
```
[
  ["time-(.*)", "(cpu)", "time-$1-$2"],
  ["(.*)", ".*", "$1"]
]
```
---
From jsonTypeMappingsSanitizationTransformerProvider/README.md (to load such a transformation)
---
```
# Configuration Routing

See the [README for the Type Mappings Sanitization Transformer](../jsonTypeMappingsSanitizationTransformer/README.md) 
for specifics of how rules are evaluated.

This package loads a [Type Mappings Sanitization Transformer](../jsonTypeMappingsSanitizationTransformer/src/main/java/org/opensearch/migrations/transform/TypeMappingsSanitizationTransformer.java)
for that appropriate application variant (currently only for the REPLAYER) along with the configuration passed into
the provider.  This [Provider](./src/main/java/org/opensearch/migrations/transform/TypeMappingSanitizationTransformerProvider.java)
pulls the values for keys `featureFlags`, `staticMappings`, and `regexMappings` from the incoming configuration map
object so that the Type Mappings Sanitization Transformer can adjust requests for specific type mappings into the 
appropriate target index.  

The following example will load a Transformer to rewrite types as per the static mappings shown in the second key-value
(staticMappings), or if not present, will then default to the mappings in regexMappings.  Note that regexMappings will
only be checked if there are no entries in staticMappings.  
staticMappings index names (top-level key) and keys to their children maps will be evaluated literally, not as patterns.  
Patterns are ONLY supported via regexMappings.

```
{
    "staticMappings": {
        "indexA": {
            "type2": "indexA_2",
            "type1": "indexA_1"
        },
        "indexB": {
            "type2": "indexB",
            "type1": "indexB"
        },
        "indexC": {
            "type2": "indexC"
        }
    },
    "regexMappings": [
        [ "(time*)", "(type*)", "$1_And_$2" ]
    ]
}
```
---

### Issues Resolved

### Testing

### Check List

### CODE REVIEW

File: transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/main/java/org/opensearch/migrations/transform/JinjavaTransformer.java [hunk 1]
```diff
@@ -0,0 +1,90 @@
...
+
+        jinjava.getGlobalContext().registerFunction(new ELFunctionDefinition(
+            "",
+            "invoke_macro",
+            DynamicMacroFunction.class,
+            "invokeMacro",
+            String.class,
+            Object[].class
+        ));
+        jinjava.getGlobalContext().registerTag(new ThrowTag());
```
chelma, line 55: I think I got it. So these filters/functions/tags would become part of our formal interface specification to the LLM as options it can use when generating its Jinja, yeah?

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/main/resources/jinjava/common/featureEnabled.j2
```diff
@@ -0,0 +1,27 @@
+{%- macro is_enabled(features, path) -%}
+    {%- if features == None -%}
+      true
+    {%- else -%}
+        {%- set ns = namespace(value=features) -%}
+        {%- set debug = namespace(log=[]) -%}
```
chelma, line 6: How does this debug log get used/printed?

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/test/java/org/opensearch/migrations/transform/JinjavaTransformerTest.java
```diff
@@ -0,0 +1,88 @@
...
+
+import java.util.Map;
+
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.testcontainers.shaded.com.fasterxml.jackson.databind.ObjectMapper;
+
+@Slf4j
+class JinjavaTransformerTest {
```
chelma, line 11: A few comments explaining how this test operates would be helpful.  I find it hard to parse what was actually being tested, where, and how.

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/test/java/org/opensearch/migrations/transform/RouteTest.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Map;
+import java.util.TreeMap;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+@Slf4j
+public class RouteTest {
```
chelma, line 14: Same request for comments here explaining what this is testing, and how.  This is requiring more mental effort to grok than ideal.

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/test/java/org/opensearch/migrations/transform/RouteTest.java
```diff
@@ -0,0 +1,82 @@
...
+        var docB = Map.of(
+            "label", "B-hive",
+            "stuff", List.of(
+                "data1",
+                "data2"
+            ));
+        {
+            var resultMap = doRouting(null, docA);
+            Assertions.assertEquals(1, resultMap.size());
+            Assertions.assertEquals("_and more!", resultMap.get("matchedVal"));
```
chelma, line 70: Example of what I mean - why are these the expected results?  Part of this is that the routing macro isn't commented so we have to infer what it does via the tests, but the tests only get you so far in understanding things.

chelma (commented):
1st wave of comments, after having spent some time learning Jinja and looking through the JinjavaTransformer code.  I expect I will have more comments as I go through the TypeMappingsSanitizationTransformer itself, next.

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/main/java/org/opensearch/migrations/transform/JinjavaTransformer.java
gregschohn: This one unifies the 'throw' tag between jinja2 and jinjava.  Without this tag, valid jinja that throws wouldn't be able to be interpreted with jinjava.  So, to answer your question, in some cases we might need to lift this configuration.  I hope that we don't have to though so that there's good parity between jinja and jinjava - so that we don't need to think about hosting a python runtime for these.

However, I DO think we'll definitely want to lift the jinja library files into the context so that those snippets can be leveraged and composited to create more consistent solutions.  Hopefully, there's less room for error too.

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/main/resources/jinjava/common/featureEnabled.j2
gregschohn: In this case, we'll accumulate stuff in that list and then do nothing with it.  If we choose to, we can emit its value (or derivatives) in the output - which would have an impact on how the result of is_enabled would be interpreted.  It's pretty much a one-time use solution - once you emit something and muddy the signal, the rest of the templating will be off.

Another option is to pass arguments into the dynamic function, which gets them back into a java context.  From there, we could debug, etc.  That might be something very much worth adding (just thought of it).

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/replayer.j2
```diff
@@ -0,0 +1,20 @@
...
+
+
+{%- set source_and_mappings = {
+  'request': request,
+  'index_mappings': index_mappings,
+  'regex_index_mappings': regex_index_mappings}
+-%}
+{{- rscope.route(source_and_mappings, request.method + " " + request.URI, flags, 'preserve.make_keep_json',
+    [
+        ('(?:PUT|POST) /([^/]*)/([^/]*)/(.*)', 'rewrite_doc_request',  'rewrite_add_request_to_strip_types'),
```
chelma, line 15: Am I going crazy, or is `rewrite_add_request_to_strip_types` not defined anywhere else in the PR?  At least, CMD+F isn't finding it...

chelma (commented):
Overall, I like where this code is going.  Definitely feels like a step in the right direction, though I'm sure we'll iterate on it a bunch.  My biggest concern is finding ways to help make the Java/Jinja interface more understandable.  It's hard to piece things together just from looking at either the Java or the Jinja; comments would go a long way, but risk becoming out of date.  There might be room for more formal API between the two layers to enforce a contact (I have no clue what this would look like, just spitballing here).  It's hard to understand what the Jinja side is doing without knowing everything the Java side is passing it.

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/src/main/resources/jinjava/common/featureEnabled.j2
gregschohn: I now have functions to log values through the Slf4j configurations for the jinjava transformer package.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/PayloadAccessFaultingMap.java
```diff
@@ -80,8 +81,21 @@ public Object put(String key, Object value) {
     public Object get(Object key) {
         var value = super.get(key);
         if (value == null && !disableThrowingPayloadNotLoaded) {
-            throw PayloadNotLoadedException.getInstance();
+            throw makeFault();
         }
         return value;
     }
+
+    public boolean missingPaylaodWasAccessed() {
```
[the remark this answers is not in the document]
gregschohn: fixed

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/README.md
```diff
@@ -0,0 +1,152 @@
+This transformer converts routes for various requests (see below) to indices that used
```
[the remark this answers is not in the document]
gregschohn: The templates are in place to handle multi-type mappings.  I'm working on debugging through all of the bugs with that since the create index call's specification has changed over time.

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/replayer.j2
```diff
@@ -0,0 +1,20 @@
...
+{%- include "typeMappings/rewriteBulkRequest.j2" -%}
+{%- include "typeMappings/rewriteCreateIndexRequest.j2" -%}
+
+
+{%- set source_and_mappings = {
+  'request': request,
+  'index_mappings': index_mappings,
+  'regex_index_mappings': regex_index_mappings}
+-%}
+{{- rscope.route(source_and_mappings, request.method + " " + request.URI, flags, 'preserve.make_keep_json',
```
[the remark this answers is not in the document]
gregschohn: we can print out the full map when we start the template.
However, what's required of the map will be a the decision of the caller and it might even be deferred all the way until runtime.  The map needs to correspond to the template.  Both are parameters, so they can be anything.  That's also why things are being added in pieces - though the lambda sent into the Jinjava transformer should be picking up most of that load.

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/replayer.j2
gregschohn: That's the name of the feature.  If a user wanted to suppress it, that's the value that they'd reference in the feature map.

gregschohn (commented):
Thanks for the feedback.  On the better documentation, I'll loop back on that after I stabilize the last couple bugs and features that I've tossed in.  In the meantime, here's some of the responses to the feedback.

chelma (approved):
Solid start, makes good progress to where we want to be.  Per discussion w/ @gregschohn, there's a lot of future work to make the mechanics simpler and clearer as we iterate and improve the overall system.  Approving this change despite the outstanding request for comments/docs as this is PR is blocking priority work, we expect the items I requested more comments on to change substantively with future revisions, and there's a performance cost to just sticking comments in Jinja without preprocessing to strip them out at runtime.

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/java/org/opensearch/migrations/transform/TypeMappingsSanitizationTransformer.java
```diff
@@ -7,49 +7,58 @@
...
+    makeSourceWrapperFunction(SourceProperties sourceProperties,
+                              Map<String, Object> featureFlagsIncoming,
                               Map<String, Map<String, String>> indexMappingsIncoming,
                               List<List<String>> regexIndexMappingsIncoming)
     {
         var featureFlags = featureFlagsIncoming != null ? featureFlagsIncoming : Map.of();
         var indexMappings = indexMappingsIncoming != null ? indexMappingsIncoming : Map.of();
+        // By NOT including a backreference, we're a bit more efficient, but it also lets us be agnostic to what
+        // types of patterns are being used.
+        // This regex says, match the type part and reduce it to nothing, leave the index part untouched.
```
chelma, line 50: Blocker

I've spent a bit trying to understand the wider implications of this default, and I'm still not groking.  Can you please expand on the comments here and explore how this default fits into the bigger picture?

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/java/org/opensearch/migrations/transform/typemappings/SourceProperties.java
```diff
@@ -0,0 +1,21 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class SourceProperties {
+    private String type;
```
chelma, line 11: Should the `type` be a String or an enumeration?  It feels like we want a constrained and obvious set of values for this one.

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/documentBackfillItems.j2
```diff
@@ -0,0 +1,19 @@
+{# see https://github.com/opensearch-project/opensearch-migrations/pull/1110 for the format of these messages #}
```
chelma, line 1: Do we not have more formal documentation?

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/documentBackfillItems.j2
```diff
@@ -0,0 +1,19 @@
...
+{%- include "typeMappings/rewriteBulkRequest.j2" -%}
+{%- import "typeMappings/rewriteIndexForTarget.j2" as transidx -%}
+
+{%- set parameters = source_document.index -%}
+
+{%- set type_name = parameters['_type'] -%}
+{%- if type_name -%}
+    {%- set target_index = transidx.convert_source_index_to_target(parameters['_index'], type_name, input_map.index_mappings, input_map.regex_index_mappings) if type_name -%}
```
chelma, line 9: Why check if the `type_name` is defined again?

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/documentBackfillItems.j2
```diff
@@ -0,0 +1,19 @@
+{# see https://github.com/opensearch-project/opensearch-migrations/pull/1110 for the format of these messages #}
+{%- include "typeMappings/rewriteBulkRequest.j2" -%}
+{%- import "typeMappings/rewriteIndexForTarget.j2" as transidx -%}
```
chelma, line 3: General question - I'm curious if you've given thought to how "default" transformations will be incorporated into the wider story we'll have in 6+ months from now.  Like - will part of our assessment process be figuring out if the pre-canned transforms are sufficient for the Cx index vs. needing some manual/GenAI transforms?  Just thinking ahead a bit, and there seems to be complexity in the UX and design we're gonna need to tackle.

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/main/resources/jinjava/typeMappings/rewriteCreateIndexRequest.j2
```diff
@@ -43,16 +44,54 @@
...
+{%- macro uses_type_names(input_map) -%}
+    {%- set uri_flag_match = input_map.request.URI | regex_capture("[?&]include_type_name=([^&#]*)") -%}
+    {%- if uri_flag_match -%}
+        {{- uri_flag_match.group1 | lower -}}
+    {%- else -%}
+        {%- set major_version = ((input_map.properties |  default({})).version |  default({})).major -%}
+        {%- if major_version >= 7-%}
+            false
+        {%- elif major_version <= 6  -%}
+            true
```
chelma, line 57: How much do you expect we'll need to embed the conditional combinatoric complexity of `source version` -> `target version` mappings into our Jinja logic longer term?  I wonder if we're better off just having GenAI make these transformations as they are needed as per-migration "one-offs" rather than try to handle it all in our pre-canned transforms.

Taking it a step further... are we making our lives harder by having these hand-built, pre-canned transforms, longer-term?

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformer/src/test/java/org/opensearch/migrations/transform/TypeMappingsSanitizationDocBackfillTest.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @Test
+    public void test() throws Exception {
+        var testString = "{\n" +
+                "  \"index\": { \"_index\": \"performance\", \"_type\": \"network\", \"_id\": \"1\" },\n" +
+                "  \"source\": { \"field1\": \"value1\" }\n" +
+                "}";
+
+        var expectedString = "{\n" +
+            "  \"index\": { \"_index\": \"network\", \"_id\": \"1\" },\n" +
```
chelma, line 26: Blocker

This was a bit surprising, assuming this is the default behavior.  I'd have expected us to have the transformed index name be `performance_network` based on prior conversation/context.  Curious to hear a bit more about the motivation of this design choice, which seems to erase a clear link to the original index and set us up for a bigger chance of collisions (multiple indices could have types with the same name, and only by prefixing the new index name with the original would you keep those documents from all ending up in the same index on the target).

transformation/transformationPlugins/jsonMessageTransformers/jsonTypeMappingsSanitizationTransformerProvider/src/test/java/org/opensearch/migrations/replay/TypeMappingsSanitizationProviderTest.java
```diff
@@ -79,4 +87,79 @@ public void testSimpleTransform() throws JsonProcessingException {
...
+                Map.of("version",
+                    Map.of("major",  (Object) 6,
+                        "minor", (Object) 10)));
+        var transformer = new TypeMappingSanitizationTransformerProvider().createTransformer(fullTransformerConfig);
+        var resultObj = transformer.transformJson(OBJECT_MAPPER.readValue(testString, LinkedHashMap.class));
+        Assertions.assertEquals(JsonNormalizer.fromString(testString), JsonNormalizer.fromObject(resultObj));
+    }
+
+    @Test
+    public void testTypeMappingsWithSourcePropertiesWorks() throws Exception {
```
chelma, line 105: I'm struggling to understand the difference between this and the `testMappingWithoutTypesAndLatestSourceInfoDoesNothing` test in terms of what they're verifying.  I see that we're passing the regex in here and the version is different, but the result is the same either way?  Maybe I'm missing something here...

PR #1157, opened 2024-11-21
How to leverage Flux to preserve the sequence of bulk puts...

gregschohn:
### Description
This is a partial and rough implementation of AndreKurait's idea that sequences bulk put responses into a flux that's effectively a progress report.  We can then use that to monitor and update the last watermark that we've made it to for purposes of writing the work continuation record when a lease expires.

TODO: 
* Unit tests compile/pass
* sendBulkRequest takes in the segment/docIds (at least the last one) & that's wired to the SegmentDocumentCursor
* exitOnLeaseTimeout() is enhanced to do create a new work item from the lastDocIndexed
* Make this easily exitOnLeaseTimeout() easily transferable to [MIGRATIONS-2172](https://opensearch.atlassian.net/browse/MIGRATIONS-2172)

* Category Enhancement
* Why these changes are required? To support preserving backfill work even when a process is terminated.
* What is the old behavior before changes and new behavior after changes? See Jira.

### Testing
Tests break - this PR is to communicate the idea.

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
+++ b/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
@@ -6,6 +6,7 @@ import java.nio.file.Paths;
 import java.time.Clock;
 import java.time.Duration;
 import java.util.List;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
 import org.opensearch.migrations.bulkload.common.DefaultSourceRepoAccessor;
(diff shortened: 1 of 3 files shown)


RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
```diff
@@ -76,7 +75,7 @@ Mono<Void> sendBulkRequest(UUID batchId, List<BulkDocSection> docsBatch, String
                 .log())
             // Prevent the error from stopping the entire stream, retries occurring within sendBulkRequest
             .onErrorResume(e -> Mono.empty())
-            .then() // Discard the response object
+            .then(Mono.just(new SegmentDocumentCursor())) // Discard the response object
```
AndreKurait, line 78: note: this constructor would take in the last docsBatch

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -307,6 +310,7 @@ public static void main(String[] args) throws Exception {
             run(
                 LuceneDocumentsReader.getFactory(sourceResourceProvider),
```
AndreKurait, line 311: note: we'll need to make the LuceneDocumentsReader read in order too

AndreKurait:
Thanks greg, remaining work tracked in https://github.com/opensearch-project/opensearch-migrations/pull/1160

PR #1160, opened 2024-11-22

chelma:
### Description

* else

Added E2E test as follows:
* Create docs with workload generator, set up toxiproxy and leases to ensure will take 8 20 second leases to finish shard with checkpoints. Verify exit codes and docs migrated

### Issues Resolved

### Testing

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/common/BulkDocSection.java
```diff
@@ -34,16 +34,28 @@ public class BulkDocSection {
 
     @EqualsAndHashCode.Include
     @Getter
-    private final String docId;
+    private final int luceneSegId;
```
peternied, line 37: Individual bulk docs might be a little too granular for reporting the subshard items.  What do you think about creating a BulkDocs object that replaces List<BulkDocSecion> which knows the start/end positions of the docs within?

Report updates to the subshard progress based on a range rather that list of indices seems cleaner.

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexAndShardCursor.java
peternied, line 1: Organizationally it might make sense for a cursor to be created with an IndexReader, and then the cursor itself holds details on how to navigate its documents and if it has reached the end of the documents.

Potentially there is a sub object IndexAndShardCursorPosition that would represent the current place and can be feed into the constructor to 'set' the position.  I think this would be useful to break apart the complexity from the Lucene Document Reader into smaller chunks.  What do you think?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
```diff
@@ -76,8 +81,8 @@ Mono<Void> sendBulkRequest(UUID batchId, List<BulkDocSection> docsBatch, String
                 .log())
             // Prevent the error from stopping the entire stream, retries occurring within sendBulkRequest
             .onErrorResume(e -> Mono.empty())
-            .then() // Discard the response object
-            .subscribeOn(scheduler);
+            .then(Mono.just(new IndexAndShardCursor(indexName, shardNumber, lastDoc.getLuceneSegId(), lastDoc.getLuceneDocId()))
```
gregschohn, line 84: I've been having a hard time understanding why shardNumber is required here.  Now I understand - because indexName really shouldn't be required either.  This call path would be easier to understand and to maintain if it were to use an IndexAndShardCursor factory that took the lucene/docId in - or if those process global things weren't included in the tracking objects.

Since indexName/shardName are global to the reindexer (it never does more than one shard, right?) - these things shouldn't be here.  As it is now, there's a leaky abstraction.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
```diff
@@ -105,6 +109,28 @@ public Flux<RfsLuceneDocument> readDocuments(int startSegmentIndex, int startDoc
         });
     }
 
+    /**
+     * We need to ensure a stable ordering of segments so we can start reading from a specific segment and document id.
+     * To do this, we sort the segments by their name.
+     */
```
gregschohn, line 115: This won't work, IIRC - Mikayla was doing something like this and not everything will be a SegmentReader.  She was originally using 'name', but it wasn't always present.  We need some other way to put these into a canonical order - anything that's stored in the metadata that's stable and unique would work.  Size wouldn't be unique, so that could be a problem...

Does the metadata not have these in a stable order that's pulled the same way every time?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
```diff
@@ -161,17 +189,17 @@ protected DirectoryReader wrapReader(DirectoryReader reader, boolean softDeletes
...
+            document = reader.document(luceneDocId);
         } catch (IOException e) {
             log.atError().setCause(e).setMessage("Failed to read document at Lucene index location {}")
-                .addArgument(docId).log();
+                .addArgument(luceneDocId).log();
             return null;
         }
 
-        String id = null;
+        String osDocId = null;
```
gregschohn, line 202: Oh - I think that I just understood what this is - this is the id for the cluster, not for lucene.  I had thought that 'os' was a 'target' id, like what the value of the docId was after we put it to the target.
Can you make these names a bit longer/clearer?

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexAndShardCursor.java
gregschohn: what does "created with an IndexReader" mean?  From an IndexReader?  What should change in this class?
This class seems to be a POJO object that represents a serializable data type to be persisted.  How would that work/the code get simpler with what's described in this comment?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/BulkDocSection.java
peternied: Yes, the comment is a broad design statement.  I'm not clear on how everything is going into the flux, but the current organization seems have _implicit_ relationships on processing order which break down if going from sequential to a non-sequental processing such as a batch that is 'stuck' on a single document that persists on failing to upload where as many other batches could keep being processed around it.

The background of this was built to represent the individual bulk request item to include a document which was part of in a list where ordering was not relevant, since then it has been shifting and changing where is now is holding information that describes where it was from and what it means for this item to be completed or not completed.

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexAndShardCursor.java
peternied: I'm suggesting that we create create a class or reuse this class to encapsulate reading related docs and completion of work from the document reindexer.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/BulkDocSection.java
chelma: This has been refactored and no longer exists

DataGenerator/src/main/java/org/opensearch/migrations/data/WorkloadGenerator.java
```diff
@@ -48,12 +48,14 @@ private List<CompletableFuture<?>> generateDocs(String indexName, Workload workl
         client.createIndex(indexName, indexRequestDoc, null);
 
         var docIdCounter = new AtomicInteger(0);
+        var segmentId = 0;
```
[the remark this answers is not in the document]
chelma: Not needed anymore after a refactor.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/BulkDocSection.java
```diff
@@ -34,16 +34,28 @@ public class BulkDocSection {
...
+    
+    @EqualsAndHashCode.Include
+    @Getter
+    private final int luceneDocId;
+
+    @EqualsAndHashCode.Include
+    @Getter
+    private final String osDocId;
```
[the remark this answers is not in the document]
chelma: Removed this field

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
chelma: @AndreKurait and I found a simple solution here.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -370,14 +405,20 @@ public static DocumentsRunner.CompletionStatus run(Function<Path, LuceneDocument
...
+            (name, shard) -> {
+                var shardMetadata = shardMetadataFactory.fromRepo(snapshotName, name, shard);
+                log.info("Shard size: " + shardMetadata.getTotalSizeBytes());
+                if (shardMetadata.getTotalSizeBytes() > maxShardSizeBytes) {
+                    throw new DocumentsRunner.ShardTooLargeException(shardMetadata.getTotalSizeBytes(), maxShardSizeBytes);
+                }
+                return shardMetadata;
+            },
+            readerFactory,
+            progressCursor::set);
```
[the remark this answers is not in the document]
chelma: Handled a different way, but addressed.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -326,8 +341,27 @@ public static void main(String[] args) throws Exception {
...
+        if (progressCursor != null) {
+            var successorWorkItem = progressCursor.toWorkItemString();
+            ArrayList<String> successorWorkItemIds = new ArrayList<>();
+            successorWorkItemIds.add(successorWorkItem);
+
+            coordinator.createSuccessorWorkItemsAndMarkComplete(
+                workItemId,
+                successorWorkItemIds,
+                contextSupplier
+            );
```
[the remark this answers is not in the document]
chelma: Added more logging

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
chelma: Done

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -326,8 +341,31 @@ public static void main(String[] args) throws Exception {
...
+    private static void exitOnLeaseTimeout(
+            IWorkCoordinator coordinator,
+            String workItemId, 
+            AtomicReference<IndexAndShardCursor> progressCursorRef,
+            Supplier<IWorkCoordinationContexts.ICreateSuccessorWorkItemsContext> contextSupplier
+    ) {
         log.error("Terminating RfsMigrateDocuments because the lease has expired for " + workItemId);
+        var progressCursor = progressCursorRef.get();
+        if (progressCursor != null) {
+            log.error("Progress cursor: " + progressCursor.toString());
```
AndreKurait, line 354: nit: move to warn, use `{}` logger format

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
AndreKurait, line 1: Note: Currently there is an off-by-one behavior where the last doc committed will be sent again by the process that picks this up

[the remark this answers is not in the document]
AndreKurait: Addressed in https://github.com/opensearch-project/opensearch-migrations/pull/1160/commits/2c2a708458e47d19d6cb0b61898754942904827a

RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
AndreKurait: Shard Number and Index Name removed from here

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
AndreKurait: Fixed the off by one with using the global doc number

RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
```diff
@@ -47,24 +46,33 @@ Mono<Void> reindexDocsInParallelBatches(Flux<BulkDocSection> docs, String indexN
         return bulkDocsBatches
             .limitRate(bulkDocsToBuffer, 1) // Bulk Doc Buffer, Keep Full
             .publishOn(scheduler, 1) // Switch scheduler
-            .flatMap(docsGroup -> sendBulkRequest(UUID.randomUUID(), docsGroup, indexName, context, scheduler),
+            .flatMapSequential(docsGroup -> sendBulkRequest(UUID.randomUUID(), docsGroup, indexName, context, scheduler),
                 maxConcurrentWorkItems)
-            .doOnTerminate(scheduler::dispose)
-            .then();
```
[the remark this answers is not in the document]
AndreKurait: We're allowing the upstream caller to handle this. We're returning the flux now in this method which gives access to the emitted items (the checkpoints)

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
AndreKurait, line 1: @mikaylathompson, can you take a look at this test?

mikaylathompson: These look like just minor updates to the existing tests, right? Am I missing anything big?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
```diff
@@ -0,0 +1,264 @@
...
+    @Getter
+    private static class RunData {
+        Path tempDirSnapshot;
+        Path tempDirLucene;
+        ToxiProxyWrapper proxyContainer;
+    }
+
+    @Test
+    public void testProcessExitsAsExpected() {
+        // 2 Shards, for each shard, expect three status code 2 and one status code 0
```
mikaylathompson, line 56: Can you add a quick note that we expect three 2s because of the size of the shard + the toxiproxy settings?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
```diff
@@ -0,0 +1,264 @@
...
+                .host(esSourceContainer.getUrl())
+                .build()
+                .toConnectionContext()
+            );
+            var generator = new WorkloadGenerator(client);
+            var workloadOptions = new WorkloadOptions();
+
+            var sourceClusterOperations = new ClusterOperations(esSourceContainer.getUrl());
+
+            var shards = 2;
```
mikaylathompson, line 104: since we're also asserting more-or-less on this in the test, we should be passing it in.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
```diff
@@ -0,0 +1,264 @@
...
+        int finalExitCodePerShard = 0;
+        runTestProcessWithCheckpoint(continueExitCode, (migrationProcessesPerShard - 1) * shards,
+                finalExitCodePerShard, shards,
+            d -> runProcessAgainstToxicTarget(d.tempDirSnapshot, d.tempDirLucene, d.proxyContainer
+            ));
+    }
+
+    @SneakyThrows
+    private void runTestProcessWithCheckpoint(int initialExitCode, int initialExitCodes,
+                                              int eventualExitCode, int eventualExitCodeCount,
```
mikaylathompson, line 69: can you name the exit code counts consistently?

mikaylathompson: and maybe with `expected` so that the count variables in this test are more distinct?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
```diff
@@ -167,32 +167,32 @@ Publisher<RfsLuceneDocument> readDocsByLeavesFromStartingPosition(DirectoryReade
 
         // Create shared scheduler for i/o bound document reading
         var sharedSegmentReaderScheduler = Schedulers.newBoundedElastic(maxDocumentsToReadAtOnce, Integer.MAX_VALUE, "sharedSegmentReader");
-
+        int startDocIdInt = 2;
```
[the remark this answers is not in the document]
AndreKurait: This shouldn't be in the final version, we're taking it from the work item

RFS/src/test/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReaderTest.java
```diff
@@ -164,41 +163,40 @@ public void ReadDocuments_ES5_Origin_AsExpected() throws Exception {
...
             String actualId = doc.id;
 
-             String expectedType = "type2";
+             String expectedType = "type1";
              String actualType = doc.type;
 
-            String expectedSource = "{\"content\":\"Updated!\"}";
+            String expectedSource = "{\"title\":\"This is a doc with complex history. Updated!\"}";
             String actualSource = doc.source;
             assertDocsEqual(expectedId, actualId, expectedType, actualType,
```
mikaylathompson, line 201: just verifying that these orders changed because we are now sorting segments & docs?

RFS/src/test/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorTest.java
AndreKurait: Correct, just was making sure I didn't miss the spirit of any of the cases with this. I wasn't expecting anything big here.

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/DocumentsRunner.java
```diff
@@ -86,8 +91,34 @@ public CompletionStatus onAlreadyCompleted() throws IOException {
 
                 @Override
                 public CompletionStatus onAcquiredWork(IWorkCoordinator.WorkItemAndDuration workItem) {
-                    doDocumentsMigration(workItem.getWorkItem(), context);
-                    return CompletionStatus.WORK_COMPLETED;
+                    var docMigrationMono = setupDocMigration(workItem.getWorkItem(), context);
+                    var latch = new CountDownLatch(1);
+                    var disposable = docMigrationMono.subscribe( lastItem -> {},
```
mikaylathompson: very nit: extra space after the `(`(don't bother with this unless you're in the file anyways)

mikaylathompson (approved):

RFS/src/test/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReaderTest.java
AndreKurait: Correct

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/DocumentsRunner.java
```diff
@@ -57,14 +85,50 @@ public CompletionStatus migrateNextShard(
...
                 }
 
                 @Override
                 public CompletionStatus onAcquiredWork(IWorkCoordinator.WorkItemAndDuration workItem) {
-                    doDocumentsMigration(IndexAndShardCursor.valueFromWorkItemString(workItem.getWorkItemId()), context);
-                    return CompletionStatus.WORK_COMPLETED;
+                    var docMigrationCursors = setupDocMigration(workItem.getWorkItem(), context);
+                    var latch = new CountDownLatch(1);
+                    var finishScheduler = Schedulers.newSingle( "workFinishScheduler");
+                    var disposable = docMigrationCursors
```
gregschohn, line 97: Why is this not its own cancel/dispose (or autocloseable) method?  Are you trying to make sure that only the creator has access to cancel the class via this back-door hook?  That doesn't seem like it needs to be a critical concern and it makes the code much harder to maintain

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/DocumentsRunner.java
```diff
@@ -57,14 +85,50 @@ public CompletionStatus migrateNextShard(
...
+                        .subscribeOn(finishScheduler)
+                        .doFinally(s -> finishScheduler.dispose())
+                        .takeLast(1)
+                        .subscribe(lastItem -> {},
+                            error -> log.atError()
+                                    .setCause(error)
+                                    .setMessage("Error prevented all batches from being processed")
```
gregschohn, line 104: what are 'all batches'?  That sounds like it could be the whole run of data - but i don't think that it is

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/DocumentsRunner.java
```diff
@@ -57,14 +85,50 @@ public CompletionStatus migrateNextShard(
...
+                                    .log(),
+                            () ->  {
+                                log.atInfo().setMessage("Reindexing completed for Index {}, Shard {}")
+                                        .addArgument(workItem.getWorkItem().getIndexName())
+                                        .addArgument(workItem.getWorkItem().getShardNumber())
+                                        .log();
+                                latch.countDown();
```
gregschohn, line 111: does this get called if there's an error?  We await on the latch below - will we terminate in that case?

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -862,7 +885,7 @@ public void createSuccessorWorkItemsAndMarkComplete(
                     e -> ctx.addTraceException(e, true)
             );
             retryWithExponentialBackoff(
-                    () -> createUnassignedWorkItemsIfNonexistent(successorWorkItemIds),
+                    () -> createUnassignedWorkItemsIfNonexistent(successorWorkItemIds, successorNextAcquisitionLeaseExponent),
```
gregschohn, line 888: nit: newline

gregschohn (commented):
I haven't looked through the tests, but some of the comments that I've left are significant. I hope to see them acknowledged before they're released.

Issue #1162, opened 2024-11-24
[BUG] RFS backfill not considering shard routing from source cluster


kanatti:
### What is the bug?
In Elasticsearch and Opensearch, when you index a document, you can specify an optional routing value, which routes the document to a specific shard.  And instead of having to fan out a search request to all the shards in an index, the request can be sent to just the shard that matches the specific routing value,

Current RFS doesn't considering shard routing from source cluster. As a result the target indices are backward incompatible if source index was build with _routing and if search queries rely in routing .

### How can one reproduce the bug?
1. Index a few documents in source with `routing`. 
2. Do a search with `explain=true` to show `_routing` metadata per doc
3. Migrate to target with RFS
4. Do a search with `explain=true` on target OS. `_routing` wont be set. You will also notice that docs within a single shard have been shuffled now on target OS.

### What is the expected behavior?
RFS use `routing` from source while bulk indexing into target.

### What is your host/environment?
N/A

### Do you have any screenshots?
No

### Do you have any additional context?

Suggested solution:
 `_routing` field should be available within lucene document and can be set per document while calling bulk API.
 
 I have a working fix, I will raise a pull request soon.

AndreKurait:
Fixed with https://github.com/opensearch-project/opensearch-migrations/pull/1164

PR #1164, opened 2024-11-25
[RFS] Use shard routing from source while bulk indexing.

kanatti:
### Description

* Category - Bug Fix
* Why these changes are required? Not using routing from source breaks routing based queries on target index.
* What is the old behavior before changes and new behavior after changes? Old behavior was creating target docs without routing. New behaviour creates target docs with routing.

### Issues Resolved
https://github.com/opensearch-project/opensearch-migrations/issues/1162

Is this a backport?
No

### Testing
1. Updated unit tests for BulkDocSection
2. Updated EndToEndTest to verify routing
3. Manually tested with ES 6.8

Note: I will raise followup PR with test-resource snapshot with routing along with updated tests for LuceneDocumentsReaderTest

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
@@ -4,28 +4,30 @@ import java.io.File;
 import java.util.List;
 import java.util.Random;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Stream;
 
 import org.opensearch.migrations.bulkload.common.FileSystemRepo;
 import org.opensearch.migrations.bulkload.common.FileSystemSnapshotCreator;
 import org.opensearch.migrations.bulkload.common.OpenSearchClient;
+import org.opensearch.migrations.bulkload.common.RestClient;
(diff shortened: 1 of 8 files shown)


DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -97,24 +122,57 @@ private void migrationDocumentsWithClusters(
...
+
+            // Check that that docs were migrated with routing
+            checkDocsWithRouting(sourceCluster, testDocMigrationContext);
+            checkDocsWithRouting(targetCluster, testDocMigrationContext);
         } finally {
             deleteTree(localDirectory.toPath());
         }
     }
 
+    private void checkDocsWithRouting(
```
AndreKurait, line 157: Thanks for adding this function, though we are missing any assertion on the doc with routing on shard 2.

Can we take in indexName, queryString, and _routing as parameters to this function and return the hits?

Then we can do the assertion in the test above.

This function can also be moved to SourceTestBase

AndreKurait (changes requested):
Thanks for taking the time for this feature add and tests.

Implementation looks good, the testing comment shouldn't be too much work to modify then we should be good to go.

AndreKurait (approved):
Since the new code path is tested, merging this in and the testing refactor can be addressed in a followup PR

PR #1176, opened 2024-12-04
Flip the default should throw behavior for HttpJsonMessageWithFaultingPayload

gregschohn:
Flip the default should throw behavior for HttpJsonMessageWithFaultingPayload to opt-in rather than opt-out.

HttpJsonMessageWithFaultingPayload throws by default - so it could be run even if it wasn't within a transform - like from within a LoggingHandler, which is what I was observing when I added some more logging.  The worst part was that it created other errors which then caused the message to be processed in a very different way.

### Description

* Category Bug fix
* Why these changes are required? When logging levels go up, the system behaves as if logging weren't enabled!

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2242

### Testing
gradle/cicd

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodyAccumulateHandler.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodyAccumulateHandler.java
@@ -8,6 +8,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
+import java.util.function.Predicate;
 
 import org.opensearch.migrations.replay.datahandlers.JsonAccumulator;
 import org.opensearch.migrations.replay.tracing.IReplayContexts;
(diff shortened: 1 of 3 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodyAccumulateHandler.java
```diff
@@ -157,12 +163,13 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
...
 
-    private boolean isRequestContentTypeNotText(HttpJsonMessageWithFaultingPayload message) {
+    private boolean hasRequestContentTypeMatching(HttpJsonMessageWithFaultingPayload message,
+                                                  Predicate<String> cnotentTypeFilter) {
         // ContentType not text if specified and has a value with / and that value does not start with text/
         return Optional.ofNullable(capturedHttpJsonMessage.headers().insensitiveGet(HttpHeaderNames.CONTENT_TYPE.toString()))
             .map(s -> s.stream()
                 .filter(v -> v.contains("/"))
-                .filter(v -> !v.startsWith("text/"))
+                .filter(cnotentTypeFilter)
```
AndreKurait, line 172: fix spelling

AndreKurait (approved):
Please fix the spelling issue and this change looks good

PR #1178, opened 2024-12-04
Fix bug handle empty mappings when removing mapping types

### LINKED ISSUE #1177: [BUG] Running the solution is getting null error in metadata migration
### What is the bug?
After running the snapshot command, and waiting till completion, ive ran the command for metadata evaluate and got an null error

making a snapshot from ES_6.7 and then try to get metadata evaluated

A normal migration from ES_6.7 to OS_2.17

### What is your host/environment?
ES_6.7 to OS_2.17

### Do you have any screenshots?
image
image
image

peternied:
### Description
Fix bug handle empty mappings when removing mapping types

### Issues Resolved
- Resolves #1177

### Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
@@ -38,8 +38,16 @@ public class Transformer_ES_6_8_to_OS_2_11 implements Transformer {
             templatesRoot.fields().forEachRemaining(template -> {
                 var templateCopy = (ObjectNode) template.getValue().deepCopy();
                 var indexTemplate = (Index) () -> templateCopy;
-                transformIndex(indexTemplate, IndexType.TEMPLATE);
-                templates.set(template.getKey(), indexTemplate.getRawJson());
+                try {
+                    transformIndex(indexTemplate, IndexType.TEMPLATE);
+                    templates.set(template.getKey(), indexTemplate.getRawJson());
+                }  catch (Exception e) {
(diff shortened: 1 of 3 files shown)


RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -38,8 +38,16 @@ public GlobalMetadata transformGlobalMetadata(GlobalMetadata globalData) {
             templatesRoot.fields().forEachRemaining(template -> {
                 var templateCopy = (ObjectNode) template.getValue().deepCopy();
                 var indexTemplate = (Index) () -> templateCopy;
-                transformIndex(indexTemplate, IndexType.TEMPLATE);
-                templates.set(template.getKey(), indexTemplate.getRawJson());
+                try {
+                    transformIndex(indexTemplate, IndexType.TEMPLATE);
+                    templates.set(template.getKey(), indexTemplate.getRawJson());
+                }  catch (Exception e) {
+                    log.atError()
```
lewijacn, line 45: Do we currently populate anything in the summary to say a transformation didn't work?

lewijacn (approved):

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
peternied: Good callout - Nope, any items that are 'rejected' in this way will never show up in the UX (So at least there wouldn't be a false positive)

We will need to reorganize the data processing model to handle this cases :(

PR #1181, opened 2024-12-05
Address a number of sonarqube java issues

peternied:
### Description
Fixed issues:
- Remove this use of "put"; it is deprecated.
- Remove this unnecessary cast to "Map".
- Rename field "TYPE" to prevent any misunderstanding/clash with field "type" defined in superclass "org.opensearch.migrations.bulkload.models.MigrationItem".
- Reorder the modifiers to comply with the Java Language Specification.
- Change the visibility of this constructor to "protected".
- Replace this lambda with method reference '*Template.class::isInstance'.
- Replace this lambda with method reference '*Template.class::cast'.
- Remove this unused method parameter "templateType".
- Remove the parentheses around the "kvp" parameter

Exclusions for false positives:
- Set the credentials provider explicitly on this builder
- Set the region explicitly on this builder

### Issues Resolved
- Resolves `Fix Java SonarQube issues` https://opensearch.atlassian.net/browse/MIGRATIONS-2248

Note; I've not worked on any issues that are part of the critical paths for transformations or sub-shard processing.

#### From
```
✅ Build passed: Found 53 issues, which is within the baseline of 89.
```

#### To
```
✅ Build passed: Found 27 issues, which is within the baseline of 89.
```

### Check List


peternied:
Moving to draft while attempting to root cause a memory leak - https://opensearch.atlassian.net/browse/MIGRATIONS-2250

DataGenerator/src/main/java/org/opensearch/migrations/data/workloads/HttpLogs.java
```diff
@@ -83,8 +83,9 @@ public Stream<ObjectNode> createDocs(int numDocs) {
         return IntStream.range(0, numDocs)
             .mapToObj(i -> {
                 var random = new Random(i);
+                long randomTime = randomTime(currentTime, random);
                 return mapper.createObjectNode()
-                    .put("@timestamp", randomTime(currentTime, random))
+                    .put("@timestamp", randomTime)
```
lewijacn, line 88: Seeing similar fixes throughout this change, what was the issue here, too much complexity?

lewijacn (approved):
Are you going to lower the baseline now?

DataGenerator/src/main/java/org/opensearch/migrations/data/workloads/HttpLogs.java
peternied: > Remove this use of "put"; it is deprecated.

This one is silly - boo on Jackson for deprecating an overridden method, and boo on SonarQube for not identifying the correct override without a specific typed local.

PR #1183, opened 2024-12-06
Revert "Adding subshard work items on lease expiry"

mikaylathompson:
Reverts opensearch-project/opensearch-migrations#1160


peternied (approved):
Please amend the comment to include the DCO

PR #1185, opened 2024-12-09

AndreKurait:
### Description
Continuation on  #1160

* else

Added E2E test as follows:
* Create docs with workload generator, set up toxiproxy and leases to ensure will take 8 20 second leases to finish shard with checkpoints. Verify exit codes and docs migrated

### Issues Resolved

### Testing

### Check List


DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
```diff
@@ -0,0 +1,273 @@
...
+    @Tag("isolatedTest")
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    public void testProcessExitsAsExpected(boolean forceMoreSegments) {
+        // Sending 5 docs per request with 4 requests concurrently with each taking 0.125 second is 160 docs/sec
+        // will process 9760 docs in 61 seconds. With 20s lease duration, expect to be finished in 4 leases.
+        // This is ensured with the toxiproxy settings, the migration should not be able to be completed
+        // faster, but with a heavily loaded test environment, may be slower which is why this is marked as
+        // isolated.
+        // 2 Shards, for each shard, expect three status code 2 and one status code 0 (4 leases)
```
peternied, line 64: I suspect this tuning is going to break down between different hardware on dev machines and build labs, we might need to find another way to manage this work.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
```diff
@@ -0,0 +1,273 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.testcontainers.containers.Network;
+
+/**
+ * TODO - the code in this test was lifted from ProcessLifecycleTest.java
+ * Some of the functionality and code are shared between the two and should be refactored.
```
peternied, line 39: This comment is copypasta - the time to refactor is now - sorry.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
AndreKurait: Not exactly, this is because of toxiproxy and running as isolated.
By increasing the lease and the artificial delay, we can make it more consistent under system stress. What I added here was a buffer from the minimum that worked on my laptop and it worked first attempt on GHA.

AndreKurait: e.g. Say this entire workload can be done in 10s on jenkins and 15s on my laptop and 20 s on a slower laptop, by adding 3 minutes of artificial delay, it will take 3:10, 3:15, 3:20, which makes it possible to consistently assert on an expected time range (e.g. between 3:00 - 3:30)
(asserting on number of leases is roughly equivalent to the time range)

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
```diff
@@ -210,22 +318,19 @@ protected RfsLuceneDocument getDocument(IndexReader reader, int docSegId, boolea
                         break;
                 }
             }
-            if (id == null) {
-                log.atWarn().setMessage("Skipping document segment id {} from source {}, it does not have an referenceable id.")
-                    .addArgument(docSegId)
-                    .addArgument(indexDirectoryPath)
-                    .log();
+            if (openSearchDocId == null) {
+                log.atError().setMessage("Document with index {} does not have an id. Skipping")
```
peternied, line 322: Without that index directory path there might not be context to know how to refine the source of the failure - this was just added because CX reported this condition that was unactionable.

peternied: Unless there is something incorrect about the previous log messages, please keep these details in place.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
peternied, line 1: I'm looking for a test that starts migration, progress is reported, then we stop RFS (simulating something going wrong), and then start a fresh instance, sure that we restarted from the correct point in the process until completion.  Do we have a test that verifies this while reading a legit snapshot?  (Work coordinator could be mocked or real IMO)

peternied (commented):
Thanks for keeping up with this Andre.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
peternied: Thanks for including the thought process you are using here - my spider-sense is tingling but lets not take any additional action without a specific issue.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
AndreKurait: >then we stop RFS (simulating something going wrong)
we currently only save progress when a lease expires. What aspect of the system would this be verifying that isn't in LeaseExpirationTest?

peternied: Our most recent customer was running into OOM of the container- in that case we would never have updated the progress.  In another case I expect, if customers dialed the rfs work scale to 0 so they paused the migration, we wouldn't want to loose that progress either

How hard would it be to add 60 second check-ins?  🤞 I hope pretty easy

AndreKurait: > In another case I expect, if customers dialed the rfs work scale to 0 so they paused the migration, we wouldn't want to loose that progress either

That work is being tracked in https://opensearch.atlassian.net/issues/MIGRATIONS-2172

PR #1190, opened 2024-12-10
First cut of the Transformation Playground Frontend and Backend

chelma:
### Description
* Added an initial cut of the Transformation Playground backend
* The user is able to start up the Django webserver and POST against the `transforms/index/` path to create some Python code that will transform an input ES 6.8 Index Settings w/ multi-type mappings or ES 7.10 Index Settings into equivalent OpenSearch 2.X settings.  The API returns the transform code, the result of invoking the transform against the user's input, and the validation results.
* The backend takes the user's input JSON, runs it through Claude 3.5 Sonnet to generate some Python transformation code, loads the transform code, invokes the transform against the input, and creates/deletes each transformed Index setting against the OpenSearch cluster the user provided for testing.
* The user is also able to spin up a quick React/Cloudscape frontend and hit the backend in their web browser.  The frontend takes in the user's input JSON and request a GenAI recommendation.  They can clear their current transform as well.  There's dropdowns for different configuration options, though currently the only one with different options is the Source, which supports both ES 6.8 and ES 7.10.  The user is able to set static guidance for the GenAI recommendation to shape its output, as well as directly modify and test their own transformation code without involving GenAI.

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-2249
* https://opensearch.atlassian.net/browse/MIGRATIONS-2257
* https://opensearch.atlassian.net/browse/MIGRATIONS-2258
* https://opensearch.atlassian.net/browse/MIGRATIONS-2260

### Testing
* Added unit tests
* Ran manual tests.  Example:

```
(venv) chelma@80a9970a4d02 tp_backend % python3 manage.py runserver
Performing system checks...

System check identified no issues (0 silenced).
December 10, 2024 - 19:14:58
Django version 5.1.4, using settings 'tp_backend.settings'
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```

```
(venv) chelma@80a9970a4d02 LP04 % curl -X POST "http://127.0.0.1:8000/transforms/index/" -H "Content-Type: application/json" -d '
{
    "transform_language": "Python",
    "source_version": "Elasticsearch 6.8",
    "target_version": "OpenSearch 2.17",
    "input_shape": {
        "index_name": "test-index",
        "index_json": {
            "settings": {
                "index": {
                    "number_of_shards": 1,
                    "number_of_replicas": 0
                }
            },
            "mappings": {
                "type1": {
                    "properties": {
                        "title": { "type": "text" }
                    }
                },
                "type2": {
                    "properties": {
                        "contents": { "type": "text" }
                    }
                }
            }
        }
    },
    "test_target_url": "http://localhost:29200"
}'

{
    "output_shape": [
        {
            "index_name": "test-index-type1",
            "index_json": {
                "settings": {
                    "index": {
                        "number_of_shards": 1,
                        "number_of_replicas": 0
                    }
                },
                "mappings": {
                    "properties": {
                        "title": {
                            "type": "text"
                        }
                    }
                }
            }
        },
        {
            "index_name": "test-index-type2",
            "index_json": {
                "settings": {
                    "index": {
                        "number_of_shards": 1,
                        "number_of_replicas": 0
                    }
                },
                "mappings": {
                    "properties": {
                        "contents": {
                            "type": "text"
                        }
                    }
                }
            }
        }
    ],
    "transform_logic": "from typing import Dict, Any, List\nimport copy\n\n\"\"\"\nThis transformation function converts Elasticsearch 6.8 index settings to OpenSearch 2.17 compatible format.\nIt handles the removal of type-based mappings by creating separate indexes for each type.\n\"\"\"\n\ndef transform(source_json: Dict[str, Any]) -> List[Dict[str, Any]]:\n    result = []\n    index_name = source_json['index_name']\n    index_json = source_json['index_json']\n    \n    # Extract settings\n    settings = index_json.get('settings', {})\n    \n    # Extract mappings\n    mappings = index_json.get('mappings', {})\n    \n    # Create a separate index for each type\n    for type_name, type_mapping in mappings.items():\n        new_index_name = f\"{index_name}-{type_name}\"\n        new_index_json = {\n            'settings': copy.deepcopy(settings),\n            'mappings': type_mapping\n        }\n        \n        result.append({\n            'index_name': new_index_name,\n            'index_json': new_index_json\n        })\n    \n    return result",
    "validation_report": [
        "Attempting to load the transform function...",
        "Loaded the transform function without exceptions",
        "Attempting to invoke the transform function against the input...",
        "Invoked the transform function without exceptions",
        "The transformed output has 2 Index entries.",
        "Using target cluster for testing: http://localhost:29200",
        "Attempting to create & delete index 'test-index-type1' with transformed settings...",
        "Created index 'test-index-type1'.  Response: \n{\"acknowledged\": true, \"shards_acknowledged\": true, \"index\": \"test-index-type1\"}",
        "Deleted index 'test-index-type1'.  Response: \n{\"acknowledged\": true}",
        "Attempting to create & delete index 'test-index-type2' with transformed settings...",
        "Created index 'test-index-type2'.  Response: \n{\"acknowledged\": true, \"shards_acknowledged\": true, \"index\": \"test-index-type2\"}",
        "Deleted index 'test-index-type2'.  Response: \n{\"acknowledged\": true}"
    ],
    "validation_outcome": "PASSED"
}
```
The default view when you open the GUI, with an input ES 6.8 JSON filled in
Screenshot 2025-01-02 at 7 22 09 AM

The GenAI recommended transformation for that input JSON
Screenshot 2025-01-02 at 7 22 42 AM

The modal window for modifying the GenAI recommendations
Screenshot 2025-01-02 at 7 27 27 AM

The results of applying the user guidance in a new, GenAI recommend transform
Screenshot 2025-01-02 at 7 28 03 AM

### Check List


TransformationPlayground/README.md
```diff
@@ -0,0 +1,74 @@
+# Transformation Playground
+
+This package contains the work-in-progress code for the Transformation Playground, a way for Elasticsearch and OpenSearch users to rapidly visualize the data and metadata within a Source cluster, create and test transformations in a sandbox, perform/test toy-scale migrations in a sandbox, and then kick off an at-scale migration.
+
+### Running the code
+
+#### Backend
+To run the backend code locally, use a Python virtual environment.  You'll need AWS Credentials in your AWS Keyring, permissions to invoke Bedrock, and to have onboarded your account to use Claude 3.5 Sonnet (`anthropic.claude-3-5-sonnet-20240620-v1:0`) in `us-west-2`.
+
+```
```
chelma, line 10: Should be `bash`

TransformationPlayground/tp_backend/transform_api/models.py
```diff
@@ -0,0 +1,11 @@
...
+
+from django.db import models
+
+# class Transformation(models.Model):
+#     id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)  # Use UUID as the primary key
+#     input_shape = models.JSONField()  # Stores the starting shape of the object to be transformed
+#     test_target_url = models.URLField()  # URL of the test target
+#     output_shape = models.JSONField()  # Stores the shape of the object after transformation
+#     transform = models.TextField()  # Stores the transformation logic used to convert the input shape to the output shape
+#     created_at = models.DateTimeField(auto_now_add=True)  # Timestamp for when the request was processed
```
chelma, line 11: Just delete; not used right now and it will be a different shape when we do need it

TransformationPlayground/tp_backend/transform_api/views.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Perform the transformation
+        try:
+            transform_report = self._perform_transformation(transform_id, requestSerializer)
+            logger.info(f"Transformation successful")
+            logger.debug(f"Transformation output:\n{transform_report.task.output}")
+            logger.debug(f"Transformation logic:\n{transform_report.task.transform.to_file_format()}")
+        except TestTargetInnaccessibleError as e:
+            logger.error(f"Target cluster is not accessible: {str(e)}")
+            return Response({'error': str(e)}, status=status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
chelma: Will address

TransformationPlayground/tp_backend/transform_api/views.py
```diff
@@ -0,0 +1,98 @@
...
+        except Exception as e:
+            logger.error(f"Transformation failed: {str(e)}")
+            logger.exception(e)
+            return Response({'error': str(e)}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
chelma: Will address

TransformationPlayground/tp_backend/transform_expert/prompting/__init__.py
```diff
@@ -0,0 +1 @@
+from transform_expert.prompting.generation import get_system_prompt_factory
```
chelma, line 1: Add a comment here to explain why we have this import and nothing else

TransformationPlayground/tp_backend/transform_expert/tests/utils/test_transforms.py
```diff
@@ -0,0 +1,41 @@
...
+            imports="",
+            description="A transform that is not executable",
+            code="transform = 42"
+        )
+
+        with self.assertRaises(TransformNotExecutableError):
+            load_transform(transform)
+
+        
+
```
chelma, line 41: Delete extra lines

TransformationPlayground/tp_backend/transform_expert/tools.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+
+
+
+
+
+
+    
+
```
chelma, line 57: Delete extra lines

TransformationPlayground/tp_backend/transform_api/views.py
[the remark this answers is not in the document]
chelma: Confirmed - does not show stack track

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+          setOutputShape("");
+          setValidationReport("");
+        }}>
+          Clear Transform
+        </Button>
+        <FormField label="Generated Transformation Logic">
+          <Textarea
+            value={transformLogic}
+            readOnly
+            placeholder="The Python code for the transformation will appear here."
```
peternied, line 165: Seems like this text 'python' should be captured from the API response not encoded on the website

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+            <Select
+              selectedOption={transformLanguage}
+              onChange={({ detail }) => setTransformLanguage(detail.selectedOption!)}
+              options={transformLanguageOptions}
+              placeholder="Select Language"
+            />
+          </Grid>
+        </FormField>
+        <br />
+        <Button onClick={handleGetRecommendation}>
```
peternied, line 151: What happens if the transformation confidence is low or produced an error when issued against the target how is that handled?

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+          {isLoading ? <Spinner/> : "Get GenAI Recommendation"}
```
peternied, line 152: Can you use this tool without getting an API recommendation but manually creating the transformation logic?

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+            value={transformLogic}
+            readOnly
+            placeholder="The Python code for the transformation will appear here."
+            rows={30}
+          />
+        </FormField>
+      </Container>
+
+      {/* Testing/Output Column */}
+      <Container header="Testing & Output Panel">
```
peternied, line 172: How do you cycle back and forth between source / transformation logic / output?

TransformationPlayground/playground_frontend/src/app/page.tsx
peternied, line 1: This seems like 3 different experience, sourcing input json, transformation editing, and viewing output json.  Seems like these should be decoupled into different pages, but I'm not sure of the overall workflow.

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+            value={inputShape}
+            onChange={({ detail }) => setInputShape(detail.value)}
+            placeholder='{"index_name": "my-index", "index_json": { index settings and mappings }}'
+            rows={30}
+          />
+        </FormField>
+      </Container>
+
+      {/* Transformation Column */}
+      <Container header="Transformation Panel">
```
peternied, line 114: This seems like it should be refactored into its own control to reduce overall coupling

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+  const [outputShape, setOutputShape] = useState("");
+  const [validationReport, setValidationReport] = useState("");
+  const [testTargetUrl, setTestTargetUrl] = useState("");
+  const [isLoading, setIsLoading] = useState(false);
+
+  // Select options for dropdowns
+  const sourceVersionOptions: SelectProps.Options = [{ label: "Elasticsearch 6.8", value: "Elasticsearch 6.8" }];
+  const targetVersionOptions: SelectProps.Options = [{ label: "OpenSearch 2.17", value: "OpenSearch 2.17" }];
+  const transformTypeOptions: SelectProps.Options = [{ label: "Index", value: "Index" }];
+  const transformLanguageOptions: SelectProps.Options = [{ label: "Python", value: "Python" }];
```
peternied, line 20: Seems like this data should come from the API, or there should be a mapping of API data against the strings for visualization

TransformationPlayground/playground/manage.py
peternied, line 1: Lets use the existing jdango api server in console_api

peternied (changes requested):
Thanks for getting this in front of the team @chelma 

I did not get a chance to refine the associated jira's in advance of this change arriving - I expect there is going to be back and forth on a number of comments I've raised.

.gitignore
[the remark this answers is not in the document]
chelma: Good callout.  I have a bunch of existing docs that can be translated into this format.

TransformationPlayground/playground/playground/settings.py
```diff
@@ -0,0 +1,199 @@
...
+WSGI_APPLICATION = 'playground.wsgi.application'
+
+
+# Database
+# https://docs.djangoproject.com/en/5.1/ref/settings/#databases
+
+DATABASES = {
+    'default': {
+        'ENGINE': 'django.db.backends.sqlite3',
+        'NAME': BASE_DIR / 'db.sqlite3',
```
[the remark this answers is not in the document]
chelma: Good question.  In the future, I expect we'll store:
* The user-loaded inputs from the left panel (what if you have a snapshot or source cluster w/ 10k+ indices?  Probably should be stored client-side), which we'll then display portions of in the client.
* The transformation logic for each "input shape" that the user has made manual changes to, along with the full validation history and the LLM conversation.  This will enable us to deploy all the transformations in a bundle that replay/backfill can use, it will facilitate our (the team's) debugging efforts, facilitate a trackable history of what the LLM's actions were and how they were incorporated into the Cx migration (think for security owners), and it will facilitate LLM Model selection/evaluation as well as LLM Fine-Tuning (we could have a collection of real migrations to tune with).

TransformationPlayground/playground/playground/settings.py
```diff
@@ -0,0 +1,199 @@
...
+            'class': 'logging.FileHandler',
+            'filename': 'logs/transform_api.debug.log',
+            'formatter': 'verbose',
+        },
+        'transform_api_info_file': {
+            'level': 'INFO',
+            'class': 'logging.FileHandler',
+            'filename': 'logs/transform_api.info.log',
+            'formatter': 'verbose',
+        },
```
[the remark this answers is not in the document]
chelma: Yeah, it includes both sets of logs.  I don't remember the working group session, is there a link to its action items/artifacts?

TransformationPlayground/playground/transform_api/tests/test_serializers.py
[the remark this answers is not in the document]
chelma: Will investigate!

TransformationPlayground/playground/transform_expert/expert.py
```diff
@@ -0,0 +1,139 @@
...
+            'max_attempts': 20,  # Increase the number of retry attempts
+            'mode': 'adaptive'   # Use adaptive retry strategy for better throttling handling
+        }
+    )
+
+    # Define our Bedrock LLM and attach the tools to it
+    llm = ChatBedrockConverse(
+        model="anthropic.claude-3-5-sonnet-20240620-v1:0", # This is the older version of the model, could be updated
+        temperature=0, # Suitable for straightforward, practical code generation
+        max_tokens=4096,
```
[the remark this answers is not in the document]
chelma: In the future, I expect this to be fully configurable by the user - what if they want to use their own LLM, for example, rather than bedrock?  We'll need to figure out what the right level of configurability is for initial user feedback.  On the token front - that is the maximum number of output tokens, not input.  4k tokens is a LOT of code, I would be highly surprised if we ever had a transform that needed anywhere close to that.  For reference, the ~450 lines of text in the OS 2.7 "knowledge" file is only ~3500 tokens.

TransformationPlayground/playground/transform_expert/expert.py
```diff
@@ -0,0 +1,139 @@
...
+        read_timeout=120, # Wait 2 minutes for a response from the LLM
+        retries={
+            'max_attempts': 20,  # Increase the number of retry attempts
+            'mode': 'adaptive'   # Use adaptive retry strategy for better throttling handling
+        }
+    )
+
+    # Define our Bedrock LLM and attach the tools to it
+    llm = ChatBedrockConverse(
+        model="anthropic.claude-3-5-sonnet-20240620-v1:0", # This is the older version of the model, could be updated
```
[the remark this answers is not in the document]
chelma: Claude 3.5 Sonnet (v1) is industry-leading at general coding tasks of this sort.  In fact, I think it's probably \*too\* good at those tasks and may be overkill.  I'd like to swap to the new Claude Haiku model when it drops and see if we can get away w/ a smaller, faster, cheaper model here.  I chose Sonnet because it's a sledgehammer that I have high confidence will "drive the nail".  In the longer term, I'd like to have a model where the first pass of each transformation process is done w/ a small, fast, cheap model and we switch to a more capable model if we encounter issues in our GenAI test-driven development loop.

Regarding effort to swap models - I expect them to be highly interchangeable with the exception that our prompt template is tied to the model and version.  Anthropic models, like Claude, prefer to receive in-prompt directives via XML tags, where Meta models, like Llama, have their own syntax (I think they use `$`'s or something).  So swapping a model would entail updating this text here along with possibly re-writing and re-testing our system prompt.  We'll ideally want some benchmark tests we can run against prospective models/prompts to help us compare/onboard.

TransformationPlayground/playground/transform_expert/expert.py
```diff
@@ -0,0 +1,139 @@
...
+        return {
+            "transform_id": self.transform_id,
+            "response": self.response.to_json()
+        }
+
+
+def perform_inference(llm: Runnable[LanguageModelInput, BaseMessage], batched_tasks: List[InferenceTask]) -> List[InferenceResult]:
+    return asyncio.run(_perform_async_inference(llm, batched_tasks))
+
+# Inference APIs can be throttled pretty aggressively.  Performing them as a batch operation can help with increasing
```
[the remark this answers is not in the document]
chelma: That's a good question.  Basically, it depends on how your split up the overall work of inference.  In our scenario, I expect this batching to be used when the user wants to generate many different transforms simultaneously; think about them loading in a bunch of indices and hitting a general "gimme recommendations" button. Each index would have its own LLM conversation, etc.  To support this, the Django API would need a revamp to handle these sorts of batch operations, among other things.

I have this async code in here currently because it was a life-saver on some of my prototyping efforts (remember the document summarizer?) that required splitting up the overall workload into many chunks due to limitations around the number of output tokens I could receive back for each chunk.

TransformationPlayground/playground/transform_expert/parameters.py
```diff
@@ -0,0 +1,17 @@
...
+    ES_6_8 = "Elasticsearch 6.8"
+
+class TargetVersion(Enum):
+    OS_2_17 = "OpenSearch 2.17"
+
+class TransformType(Enum):
+    INDEX = "Index"
+
+class TransformLanguage(Enum):
+    PYTHON = "Python"
```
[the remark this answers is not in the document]
chelma: Good callout.  This PR does Python because I expect it to be pretty easy to swap to the language we want (Jinja) and it seemed more important to get a first cut of this stuff out for folks to see than wait for the Jinja swap.  Basically, moving to Jinja will entail updating the section of our system prompt that currently defines our Python interface and some of the validation code for loading the transform into memory.  It \*should\* all be addressable with the current abstractions in place (tried to plan ahead a bit for it).

TransformationPlayground/playground/transform_expert/prompting/knowledge/__init__.py
```diff
@@ -0,0 +1,33 @@
+from transform_expert.parameters import SourceVersion, TargetVersion, TransformType, TransformLanguage
+
+from transform_expert.prompting.knowledge.es_6_8 import INDEX_GUIDANCE as es_6_8_index_guidance, INDEX_KNOWLEDGE as es_6_8_index_knowledge
+from transform_expert.prompting.knowledge.os_2_17 import INDEX_GUIDANCE as os_2_17_index_guidance, INDEX_KNOWLEDGE as os_2_17_index_knowledge
+
+
+def get_source_guidance(source_version: SourceVersion, target_version: TargetVersion, input_shape_type: TransformType, transform_language: TransformLanguage) -> str:
```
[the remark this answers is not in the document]
chelma: Guidance is things like, "if you have multiple type mappings, prefer to split them into separate indices".  As a result - it's closely tied to the specific source/target versions.  Think of this as a set of default hints we provide the LLM when there's ambiguity on the multiple, valid routes it could take.

TransformationPlayground/playground/transform_expert/prompting/knowledge/os_2_17.py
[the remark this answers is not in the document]
chelma: Good question.  I've automated the process of distilling knowledge into a compact format (run a script, wait a minute or two, end up with markdown), but that process is still human-initiated.  We have a finite set of source/target versions we'll be working with and I don't expect we'll ever need to change the knowledge set once its been distilled and added to our library, so it feels like a manageable approach to handle things this way.

TransformationPlayground/playground/transform_expert/prompting/knowledge/os_2_17.py
```diff
@@ -0,0 +1,453 @@
+INDEX_GUIDANCE = """"""
+
+INDEX_KNOWLEDGE = """
```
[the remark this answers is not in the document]
chelma: I'm sure we'll iterate on the ideas here, but - I see "guidance" as the hints we provide the LLM when there are multiple valid choices it could pursue (separate indices vs union index when resolving multi-type mappings, for example).  The end user will be able to override this guidance with their own directives via the Playground UI, as well.

Knowledge is stuff that is fundamentally true (from the LLM's perspective) about the task at hand and unlikely to change.  This is stuff like the API spec for a specific version of the source/target.

TransformationPlayground/playground/transform_expert/utils/opensearch_client.py
```diff
@@ -0,0 +1,45 @@
+import logging
+from requests import HTTPError, ConnectionError
+from typing import Optional, Dict, Any
+
+from transform_expert.utils.rest_client import RESTClient
+
+logger = logging.getLogger("transform_expert")
+
+class OpenSearchClient():
```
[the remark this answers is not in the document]
chelma: Maybe - though I think they tend to be either very task specific (like what we have in RFS) or very lightweight (like what we have here).  If there's specific implementations you think we can/should re-use I'd be curious to know more.

TransformationPlayground/playground_frontend/README.md
```diff
@@ -0,0 +1,36 @@
+This is a [Next.js](https://nextjs.org) project bootstrapped with [`create-next-app`](https://nextjs.org/docs/app/api-reference/cli/create-next-app).
```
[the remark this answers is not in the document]
chelma: I put those details in the top-level readme, but I take your point.  Probably shouldn't have just the generic README here.  Good callout.

TransformationPlayground/playground_frontend/README.md
```diff
@@ -0,0 +1,36 @@
...
+## Learn More
+
+To learn more about Next.js, take a look at the following resources:
+
+- [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js features and API.
+- [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
+
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js) - your feedback and contributions are welcome!
+
+## Deploy on Vercel
```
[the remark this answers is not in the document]
chelma: Makes sense, lol

TransformationPlayground/playground_frontend/src/app/page.tsx
```diff
@@ -0,0 +1,206 @@
...
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(payload),
+      });
+
+      if (response.ok) {
+        const data = await response.json();
+
+        // Pretty-print JSON for output_shape
+        const prettyOutputShape = JSON.stringify(data.output_shape, null, 4);
```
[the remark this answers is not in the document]
chelma: Good callout.  Agreed, need to learn the best practices here and apply them.  I have ~5 hours of frontend experience in total at this point, so if you have specific ideas and approaches that you think are effective I'd love to hear them.

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: Good callout, all our hints/placeholder text should be dynamically driven.  Seems like the selection dropdowns are a good option to me?

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: Fantastic question!  There's a short-term answer, and a longer-term answer.

Short term, I imagine we'd clear out the transformation and output text fields and then have those errors show up in the validation report section.  I want there to be more prominent iconography that validation passed/failed/etc, and for the user to pop up something like a modal window and drill down into the specifics of the validation process.  The validation report should surface the stuff they care about.  One thing to keep in mind though... this is an inherently human-in-the-loop process.  The LLM is an assistant, even an expert assistant maybe, but it still needs a manager to direct the overall process and make high level decisions.

Longer term... the question that I wrestle with is stuff like, "if the user passes in guidance to the LLM on their task that causes the LLM to get confused and derails the transform process, how does the user gain insight into that"?  For the medium term, I don't envision this GUI as a true chat experience; it's really more like there's a machine shop in the back room that you can pass blueprints to and, optionally, a note to the machinist, then you wait and see what parts come back your way.  In the longer term, I can see us presenting a way to chat directly with an LLM about the history and context of a specific transformation job using the data we've been storing in the server-side DB.

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: Yep, that's one of the next things I want to implemented (should be real quick).  Basically, make the text area for the transformation writable and add buttons for "Test" and "Deploy" to the righthand panel.  "Test" would take the transform logic and run it through the same validation process we apply to the LLM output.  "Deploy" would basically do the same thing as "Test" but without the final cleanup step where we do stuff like delete the index off the test target.  This enables the user to use the Playground for small-scale migration efforts directly without needing the full Migration Assistant, as well as probably setting up the Playground to be how we do metadata migrations in general.

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: I haven't implemented this all yet (obviously), but here's the user journey I'm envisioning:
* The user loads in their input shapes.  Currently, that's just copy/pasting JSON into a text field, but soon I'd like to be able to load the contents of a snapshot into the Playground via a configurable process.  Imagine a dialogue where you select the S3 location of your snapshot, something extracts the templates, indices, and a configurable number of documents (that's a whole workstream right there), and presents them in the left panel so you can see everything in your snapshot, then click on specific items to see their JSON, etc.  Even once we've added the ability to read from snapshots, we'll want to leave open the possibility of adding manual input shapes.  Example - the user doesn't have access to a snapshot of their production cluster but they do have access to the settings JSON and a few documents, which they can manually enter as shapes into the Playground for testing.
* The user then uses the dropdowns to select what type of transformation they are performing.  The "transform type" (Index vs. template vs. documents) dropdown will disappear, because it will be obvious from context (the user selected an index on the input panel, so...).
* Once that is selected, we will have a pre-canned transform (we'll hand-craft and save, no GenAI) that is supplied based on those selections that will serve as a default.  The default will be populated into the UI and pinned against that shape in the backend.
* The user can then hit a "test" button to see how that default transform will work against their input shape (e.g. we kick off the validation process).
* If the user wants to include functional tests into validation (e.g. actually creating/deleting indices against a real test cluster), they can go through a dialogue to set up and test a connection to their target cluster.  Currently, that process is just "paste a URL into a text field and assume there's no auth to worry about").
* If the user likes the results of the default, pre-canned transform and it passes validation - great!  No further work needed.  If not, they can either modify the transform directly in the UI and run the validation process again, or they can ask the GenAI assistant to modify the existing transform in some way (not currently implemented, but expect to be easy).  Either way, the new transform is then pinned against the input shape in the backend so that we know there's something custom going on.
* The user goes through this process until everything transforming as they desire, then they can either hit a button to "deploy" their transformed data/metadata to the target cluster (think for metadata migrations which are inherently "low scale") or "bundle" the transformations for inclusion into the Migrations Assistant backfill/replay processes (which are anything but "low scale").  For the "deploy" button, this is basically just skipping the cleanup step of the validation process we've already performed (e.g. take the transformed output and PUT it against the target, but don't delete it afterwards).  For bundling, this is taking the stuff we have in the server-side DB, packaging it into a tarball/zip/whatever, and sticking it somewhere (S3?) so that the backfill and replayer processes can pick it up and load the transform objects.

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: I think there's a lot of benefit in visualizing the three things together; I outlined the user-journey I'm imagining above here.  Curious what you think after reading that.

[1] https://github.com/opensearch-project/opensearch-migrations/pull/1190#discussion_r1883812992

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: Would love more details!  But yeah - this whole page is ripe for refactoring.

TransformationPlayground/playground_frontend/src/app/page.tsx
chelma: +1.  I think this comes down to a shared spec between the frontend and backend; this also applies to the shape of the requests/responses that are going over the wire.  I know we've briefly discussed OpenAPI or something as a model format.

TransformationPlayground/playground/manage.py
chelma: I think there's a lot of benefit in separating this as a distinct project:
* The Playground is conceptually only loosely coupled to the rest of the Migration Assistant.  There's no reason the snapshot it's reading inputs from needs to come from a `console snapshot create` command.  Similarly, there's no reason the target cluster used for validation (or we do low-scale deployment against) needs to have been created as a part of the Migration Assistant setup process.  The best argument might be around the format of the transformation bundle being coupled to the Migration Assistant, but there's no reason we couldn't specify that format to be generic.  That said - you don't need to care about the rest of Migration Assistant at all for the Playground to be a useful way of creating/testing these transformations.
* The Playground is a very different experience than the general operator workflow.  It's inherently interactive and cyclical in a way that running basic configuration commands is not.
* The Playground is generically useful and extensible as a standalone product feature.  At it's core, it's a way to visualize data/metadata, have a GenAI assisted process for modifying/transforming it, testing the modifications/transformations, and some mechanisms for deployment.  Swapping out what specifically it's using as an input or testing against or has a desired output is inherently modular.  There's already multiple other use-cases we've discussed for where this could apply in other domains outside of Elasticsearch to OpenSearch migrations.

The key thing in my mind is that users of the Migration Assistant should have a cohesive overall experience; but I think tightly coupling the Playground to the Migration Assistant Operator API/GUI in a way that prevents re-use would be a mistake.

TransformationPlayground/playground/transform_api/views.py
```diff
@@ -0,0 +1,181 @@
...
+
+        # Perform the test
+        try:
+            validation_report = self._perform_test(transform_id, requestSerializer)
+            logger.info(f"Test complete")
+            logger.debug(f"Validation outcome:\n{validation_report.passed}")
+            logger.debug(f"Validation report entries:\n{validation_report.report_entries}")
+        except TestTargetInnaccessibleError as e:
+            logger.error(f"Target cluster is not accessible: {str(e)}")
+            return Response({'error': str(e)}, status=status.HTTP_400_BAD_REQUEST)
```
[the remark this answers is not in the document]
peternied: Can we tweak this code not to include the error in the response?  This will set off alarms in our other security scanning tools such as our internal sonarqube instance.

TransformationPlayground/playground/transform_api/views.py
```diff
@@ -0,0 +1,98 @@
...
+            logger.info(f"Transformation successful")
+            logger.debug(f"Transformation output:\n{transform_report.task.output}")
+            logger.debug(f"Transformation logic:\n{transform_report.task.transform.to_file_format()}")
+        except TestTargetInnaccessibleError as e:
+            logger.error(f"Target cluster is not accessible: {str(e)}")
+            return Response({'error': str(e)}, status=status.HTTP_400_BAD_REQUEST)
+        except Exception as e:
+            logger.error(f"Transformation failed: {str(e)}")
+            logger.exception(e)
+            return Response({'error': str(e)}, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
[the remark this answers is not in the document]
peternied: (Same as the other issue) Can we tweak this code not to include the error in the response? This will set off alarms in our other security scanning tools such as our internal sonarqube instance.

sumobrian:
Closing until SonarCube configuration changes are made. Once made this can be re-opened and merged.

PR #1198, opened 2024-12-12

AndreKurait:
### Description
Continuation on  #1160

Changes since #1160
* Added LeaseExpirationTest that refreshes after every bulk put to synthetically generate more segments for the test case.
* Added Binary search to more efficiently find the starting document segment
* Modified initial segment sorting with error logging on equality
____

* else

Added E2E test as follows:
* Create docs with workload generator, set up toxiproxy and leases to ensure will take 8 40 second leases to finish shard with checkpoints. Verify exit codes and docs migrated

### Issues Resolved

### Testing

### Check List


peternied (commented):
With the holidays coming up - lets hold off on merging this until after we've got a fuller crew in house.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
```diff
@@ -105,24 +112,109 @@ public Flux<RfsLuceneDocument> readDocuments(int startSegmentIndex, int startDoc
         });
     }
 
-    protected DirectoryReader getReader() throws IOException {// Get the list of commits and pick the latest one
+    /**
+     * We need to ensure a stable ordering of segments so we can start reading from a specific segment and document id.
+     * To do this, we sort the segments by their ID or name.
+     */
+    static class SegmentNameSorter implements Comparator<LeafReader> {
```
mikaylathompson, line 119: @AndreKurait do we still use sorting based on segment name?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
```diff
@@ -105,24 +112,109 @@ public Flux<RfsLuceneDocument> readDocuments(int startSegmentIndex, int startDoc
...
+        // Get the list of commits and pick the latest one
         try (FSDirectory directory = FSDirectory.open(indexDirectoryPath)) {
             List  <IndexCommit> commits = DirectoryReader.listCommits(directory);
             IndexCommit latestCommit = commits.get(commits.size() - 1);
-
             return DirectoryReader.open(
                 latestCommit,
                 6, // Minimum supported major version - Elastic 5/Lucene 6
-                null // No specific sorting required
+                new SegmentNameSorter()
```
peternied, line 171: @mikaylathompson FYI - This constructor isn't available in lucene7, we will need a workaround

AndreKurait: Workaround merged to main

RFS/src/main/java/org/opensearch/migrations/bulkload/common/LuceneDocumentsReader.java
AndreKurait: yes, Updated this to be more clear

mikaylathompson (approved):

PR #1200, opened 2024-12-13
Fix snapshot status based on demo

AndreKurait:
### Description
Fix snapshot status based on demo

During demo, snapshot status was returning an exceedingly high number, this fixes that by removing a reference on incremental. Also cleans up how early status reports throughput and duration

### Check List


peternied:
@jugal-chauhan Mind taking a look at this change, do you think it would be useful?

jugal-chauhan:
I see that this change simplifies the code logic - it now treats full and incremental snapshots as same. I am interested in understanding what led to this change. 

It is unlikely for Snapshot progress to report high numbers when customers just have a single full snapshot in their S3 Repo.

I am also curious about how often do customers use incremental snapshots. If they are common in our solution, I think we might be losing some accuracy in progress reporting. I did not find a test-case that covers implementing incremental snapshots with this change in this PR. If incremental snapshots are important to our use-case then are we handling them in a different way?

AndreKurait:
Closing, will address as part of https://opensearch.atlassian.net/issues/MIGRATIONS-2286

PR #1225, opened 2025-01-08
Metadata add 1:many custom transformations

AndreKurait:
### Description

1. Updates MetadataMigration to receive and handle List returned by IJsonTransformer, this adds support for 1:many custom transforms (e.g. multi type index split)
     * Note: A future change will add support for built in transform to handle the split case.
2. Updates the TypeSanitization built in transformer to throw an error when encountering multi type index and behavior is not set to Union. (Previous behavior was to return Unsupported which would ignore the transformation which would result in an error during cluster put) 

### Issues Resolved

[MIGRATIONS-2271](https://opensearch.atlassian.net/browse/MIGRATIONS-2271)

### Testing
Existing code paths tested by existing tests.

1:Many transformations will be tested during a followup PR which adds multi type split as a provided javascript transform.

### Check List

### DIFF
--- a/MetadataMigration/src/test/java/org/opensearch/migrations/MultiTypeMappingTransformationTest.java
+++ b/MetadataMigration/src/test/java/org/opensearch/migrations/MultiTypeMappingTransformationTest.java
@@ -81,6 +81,7 @@ class MultiTypeMappingTransformationTest extends BaseMigrationTest {
             var dataFilterArgs = new DataFilterArgs();
             dataFilterArgs.indexAllowlist = List.of(originalIndexName);
             arguments.dataFilterArgs = dataFilterArgs;
+            arguments.sourceVersion = upgradedSourceCluster.getContainerVersion().getVersion();
 
             // Use union method for multi-type mappings
             arguments.metadataTransformationParams.multiTypeResolutionBehavior = IndexMappingTypeRemoval.MultiTypeResolutionBehavior.UNION;
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/CompositeTransformer.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/CompositeTransformer.java
@@ -1,5 +1,9 @@
 package org.opensearch.migrations.bulkload.transformers;
 
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
 import org.opensearch.migrations.bulkload.models.GlobalMetadata;
 import org.opensearch.migrations.bulkload.models.IndexMetadata;
 
@@ -19,10 +23,13 @@ public class CompositeTransformer implements Transformer {
     }
 
     @Override
-    public IndexMetadata transformIndexMetadata(IndexMetadata indexData) {
-        for (Transformer transformer : transformers) {
-            indexData = transformer.transformIndexMetadata(indexData);
-        }
-        return indexData;
+    public List<IndexMetadata> transformIndexMetadata(IndexMetadata indexData) {
+        return Stream.of(transformers)
+                .reduce(
+                    Stream.of(indexData),
+                    (stream, transformer) -> stream.flatMap(data -> transformer.transformIndexMetadata(data).stream()),
+                    Stream::concat
+                )
+                .collect(Collectors.toList());
     }
 }
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer.java
@@ -1,5 +1,7 @@
 package org.opensearch.migrations.bulkload.transformers;
 
+import java.util.List;
+
 import org.opensearch.migrations.bulkload.models.GlobalMetadata;
 import org.opensearch.migrations.bulkload.models.IndexMetadata;
 
@@ -16,5 +18,5 @@ public interface Transformer {
     /**
      * Takes the raw JSON representing the Index Metadata of one version and returns a new, transformed copy of the JSON
      */
-    public IndexMetadata transformIndexMetadata(IndexMetadata indexData);
+    public List<IndexMetadata> transformIndexMetadata(IndexMetadata indexData);
 }
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
@@ -109,10 +109,8 @@ public class TransformerToIJsonTransformerAdapter implements Transformer {
 
     @Override
     public GlobalMetadata transformGlobalMetadata(GlobalMetadata globalData) {
-        var inputJson = objectNodeToMap(globalData.toObjectNode());
-        log.atInfo().setMessage("BeforeJsonGlobal: {}").addArgument(() -> printMap(inputJson)).log();
-        Object afterJson = transformer.transformJson(inputJson);
-        log.atInfo().setMessage("AfterJsonGlobal: {}").addArgument(() -> printMap(afterJson)).log();
+        log.atInfo().setMessage("BeforeJsonGlobal: {}")
+            .addArgument(() -> printMap(objectNodeToMap(globalData.toObjectNode()))).log();
 
         final List<LegacyTemplate> legacyTemplates = new ArrayList<>();
         globalData.getTemplates().fields().forEachRemaining(
@@ -150,13 +148,12 @@ public class TransformerToIJsonTransformerAdapter implements Transformer {
                 .map(ComponentTemplate.class::cast)
                 .collect(Collectors.toList());
 
-        assert transformedLegacy.size() + transformedIndex.size() + transformedComponent.size() == transformedTemplates.size();
-
         updateTemplates(transformedLegacy, globalData.getTemplates());
         updateTemplates(transformedIndex, globalData.getIndexTemplates());
         updateTemplates(transformedComponent, globalData.getComponentTemplates());
 
-        log.atInfo().setMessage("GlobalOutput: {}").addArgument(() -> printMap(objectNodeToMap(globalData.toObjectNode()))).log();
+        log.atInfo().setMessage("AfterJsonGlobal: {}")
+            .addArgument(() -> printMap(objectNodeToMap(globalData.toObjectNode()))).log();
         return globalData;
     }
 
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
@@ -24,7 +24,9 @@ import lombok.extern.slf4j.Slf4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
+/**
+ * Adapts an {@link IJsonTransformer} to a {@link Transformer}.
+ */
 @Slf4j
 public class TransformerToIJsonTransformerAdapter implements Transformer {
     public static final String OUTPUT_TRANSFORMATION_JSON_LOGGER = "OutputTransformationJsonLogger";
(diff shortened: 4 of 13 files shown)


RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/CompositeTransformer.java
```diff
@@ -19,10 +23,13 @@ public GlobalMetadata transformGlobalMetadata(GlobalMetadata globalData) {
...
 
     @Override
-    public IndexMetadata transformIndexMetadata(IndexMetadata indexData) {
-        for (Transformer transformer : transformers) {
-            indexData = transformer.transformIndexMetadata(indexData);
-        }
-        return indexData;
+    public List<IndexMetadata> transformIndexMetadata(IndexMetadata indexData) {
+        return Stream.of(transformers)
+                .reduce(
```
gregschohn, line 28: See https://docs.oracle.com/javase/8/docs/api/java/util/stream/Stream.html#reduce-T-java.util.function.BinaryOperator-

Reduce is "not constrained to execute sequentially."  

Is the first argument an identity value?  It seems like one could make the argument either way, but if ordering isn't defined, then bets are off.  Imagine if multiple patches of values were accumulated then combined together.  That would be drastically different than if one patch was executed sequentially with no combines required.

> The identity value must be an identity for the combiner function. This means that for all u, combiner(identity, u) is equal to u. Additionally, the combiner function must be compatible with the accumulator function; for all u and t, the following must hold:...

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/CompositeTransformer.java
```diff
@@ -19,10 +23,13 @@ public GlobalMetadata transformGlobalMetadata(GlobalMetadata globalData) {
...
+                    Stream.of(indexData),
+                    (stream, transformer) -> stream.flatMap(data -> transformer.transformIndexMetadata(data).stream()),
```
gregschohn, line 30: The spirit of this call - slicing the output into separate subsequent transforms is really weird.  For the JSON Transformer composite transformer code, we'll do direct composition.  In this case, there's an implicit fanout because the type for the per-index metadata transform calls is a scalar Index - not a collection of Indices.

My opinion here, from the limited code that I've seen, would be to abstain from allowing the RFS-MetaData CompositeTransformer to host transforms that can return lists.  Keep two types of transforms.  Merge the outputs of the list one with a specific class that does fanout across another sequence of 'simple' transforms.  That might be a few more lines of code, but it might be a lot clearer.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer.java
```diff
@@ -16,5 +18,5 @@ public interface Transformer {
     /**
      * Takes the raw JSON representing the Index Metadata of one version and returns a new, transformed copy of the JSON
      */
-    public IndexMetadata transformIndexMetadata(IndexMetadata indexData);
+    public List<IndexMetadata> transformIndexMetadata(IndexMetadata indexData);
```
gregschohn, line 21: See the comment above about the composability problem.  It's hard to get one's head around this once you know that we need to chain some transforms together.

What if you keep all of the "simple transforms" that are IndexMetadata->IndexMetadata.  For when you need to return a list or take a list in, use an adapter that does the wrapping and any fanout.  It feels capricious/fragile in the code below when we're boxing within a list.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
```diff
@@ -24,7 +24,9 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
+/**
+ * Adapts an {@link IJsonTransformer} to a {@link Transformer}.
+ */
 @Slf4j
 public class TransformerToIJsonTransformerAdapter implements Transformer {
```
gregschohn, line 31: Is the name of this class backward?  It feels like the IJsonTransformerAdapter is being transformed into a Metadata Transformer interface, not the other way around.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
```diff
@@ -74,12 +76,21 @@ private static String printMap(Object map) {
...
+        Object transformedResult = transformer.transformJson(MAPPER.convertValue(migrationItem, Map.class));
+        logTransformation(originalMap, transformedResult);
+
+        List<MigrationItem> transformedItems = new ArrayList<>();
+        if (transformedResult instanceof List) {
+            for (Object item : (List<Object>) transformedResult) {
+                transformedItems.add(MAPPER.convertValue(item, MigrationItem.class));
+            }
+        } else {
+            transformedItems.add(MAPPER.convertValue(transformedResult, MigrationItem.class));
```
gregschohn, line 91: nit: List.of() would work too

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
```diff
@@ -74,12 +76,21 @@ private static String printMap(Object map) {
...
         final Map<String, Object> originalMap = MAPPER.convertValue(migrationItem, Map.class);
-        Object transformedMigrationItem = transformer.transformJson(MAPPER.convertValue(migrationItem, Map.class));
-        logTransformation(originalMap, transformedMigrationItem);
-        return MAPPER.convertValue(transformedMigrationItem, MigrationItem.class);
+        Object transformedResult = transformer.transformJson(MAPPER.convertValue(migrationItem, Map.class));
+        logTransformation(originalMap, transformedResult);
+
+        List<MigrationItem> transformedItems = new ArrayList<>();
+        if (transformedResult instanceof List) {
+            for (Object item : (List<Object>) transformedResult) {
```
gregschohn, line 87: nit: map/collect would work too - this and the next comment could return immutable lists.  If you can't do that, you should add why (as I think of it, you might not be able to do that - so it is tempting).

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexRunner.java
```diff
@@ -33,44 +34,56 @@ public IndexMetadataResults migrateIndices(MigrationMode mode, ICreateIndexConte
...
         var indexMetadata = originalIndexMetadata.deepCopy();
+        List<CreationResult> creationResults = new ArrayList<>();
         try {
-            indexMetadata = transformer.transformIndexMetadata(indexMetadata);
-            return indexCreator.create(indexMetadata, mode, context);
+            List<IndexMetadata> transformedMetadataList = transformer.transformIndexMetadata(indexMetadata);
+            for (IndexMetadata transformedMetadata : transformedMetadataList) {
+                try {
+                    creationResults.add(indexCreator.create(transformedMetadata, mode, context));
+                } catch (Exception e) {
```
gregschohn, line 72: How did sonarqube not flag catching a raw Exception?

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexRunner.java
```diff
@@ -33,44 +34,56 @@ public IndexMetadataResults migrateIndices(MigrationMode mode, ICreateIndexConte
...
 
-    private CreationResult createIndex(String indexName, MigrationMode mode, ICreateIndexContext context) {
+    private List<CreationResult> createIndex(String indexName, MigrationMode mode, ICreateIndexContext context) {
         var originalIndexMetadata = metadataFactory.fromRepo(snapshotName, indexName);
         var indexMetadata = originalIndexMetadata.deepCopy();
+        List<CreationResult> creationResults = new ArrayList<>();
         try {
-            indexMetadata = transformer.transformIndexMetadata(indexMetadata);
-            return indexCreator.create(indexMetadata, mode, context);
+            List<IndexMetadata> transformedMetadataList = transformer.transformIndexMetadata(indexMetadata);
```
gregschohn, line 68: Agree w/ the SonarQube warning - if this fails, just return a List.of(...).
Then, if it succeeds, go into the for loop w/ the separate try-catch blocks

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexRunner.java
```diff
@@ -33,44 +34,56 @@ public IndexMetadataResults migrateIndices(MigrationMode mode, ICreateIndexConte
...
                 }
 
-                results.index(creationResult);
+                creationResults.forEach(results::index);
 
                 var indexMetadata = metadataFactory.fromRepo(snapshotName, index.getName());
                 indexMetadata.getAliases().fieldNames().forEachRemaining(alias -> {
                     var aliasResult = CreationResult.builder().name(alias);
-                    aliasResult.failureType(creationResult.getFailureType());
+                    aliasResult.failureType(creationResults.get(0).getFailureType());
```
gregschohn, line 56: nit, this code might be quicker to read if it were tighter (it would be easy to compress w/out the log statement - which could be moved to outside the construction.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
```diff
@@ -122,7 +132,8 @@ public GlobalMetadata transformGlobalMetadata(GlobalMetadata globalData) {
                 indexTemplates.stream()),
                 componentTemplates.stream()
                 )
-                .map(this::transformMigrationItem).collect(Collectors.toList());
+                .flatMap(item -> transformMigrationItem(item).stream())
+                .collect(Collectors.toList());
```
gregschohn, line 134: There's some risk here. If a custom transform maps some components to OS2 and leaves others behind.  We'd really want/need to let version transforms run for some things and not others.  As discussed w/ Andre, adding a version key to each item would help us identify the version in the transform and also let subsequent transforms know that it's already been updated to that schema.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
```diff
@@ -122,7 +132,8 @@ public GlobalMetadata transformGlobalMetadata(GlobalMetadata globalData) {
                 indexTemplates.stream()),
                 componentTemplates.stream()
                 )
-                .map(this::transformMigrationItem).collect(Collectors.toList());
+                .flatMap(item -> transformMigrationItem(item).stream())
```
gregschohn, line 133: Converting this calling pattern to be a transform on the whole list would be safer (no collisions), give the user more flexibility, & probably also make it easier for genAI to create complete transforms.

gregschohn (approved):
Provisionally approved with the understanding that there's a task filed to have all of the global metadata items (including indices) go through a single transformation as an index.

MetadataMigration/src/test/java/org/opensearch/migrations/MultiTypeMappingTransformationTest.java
```diff
@@ -81,6 +81,7 @@ public void multiTypeTransformationTest_union() {
             var dataFilterArgs = new DataFilterArgs();
             dataFilterArgs.indexAllowlist = List.of(originalIndexName);
             arguments.dataFilterArgs = dataFilterArgs;
+            arguments.sourceVersion = upgradedSourceCluster.getContainerVersion().getVersion();
```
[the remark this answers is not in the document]
AndreKurait: no, this is right, sourceClusterToUpgrade has been removed at this point

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/CompositeTransformer.java
AndreKurait: Thanks! Fixed

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerToIJsonTransformerAdapter.java
AndreKurait: Thanks, will address in https://opensearch.atlassian.net/browse/MIGRATIONS-2301

PR #1229, opened 2025-01-09
[Console Library] Allow console library completion everywhere

lewijacn:
### Description
Removes requirement that migrations `services.yaml` has to be loaded in from environment in order to get console library bash completion

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2236

### Testing
Local testing

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
@@ -61,10 +61,9 @@ RUN echo '. /etc/profile.d/bash_completion.sh' >> ~/.bashrc && \
     echo '. /etc/profile.d/venv.sh' >> ~/.bashrc && \
     echo 'echo Welcome to the Migration Assistant Console' >> ~/.bashrc && \
     echo 'eval "$(register-python-argcomplete cluster_tools)"' >> ~/.bashrc && \
-    echo 'PS1="(\t) \[\e[92m\]migration-console \[\e[0m\](\w) -> "' >> ~/.bashrc && \
-    source /.venv/bin/activate && \
-    console --config-file=/root/lib/console_link/services.yaml completion bash > /usr/share/bash-completion/completions/console
-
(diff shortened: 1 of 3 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
```diff
@@ -61,7 +61,9 @@ RUN echo '. /etc/profile.d/bash_completion.sh' >> ~/.bashrc && \
     echo '. /etc/profile.d/venv.sh' >> ~/.bashrc && \
     echo 'echo Welcome to the Migration Assistant Console' >> ~/.bashrc && \
     echo 'eval "$(register-python-argcomplete cluster_tools)"' >> ~/.bashrc && \
-    echo 'PS1="(\t) \[\e[92m\]migration-console \[\e[0m\](\w) -> "' >> ~/.bashrc
+    echo 'PS1="(\t) \[\e[92m\]migration-console \[\e[0m\](\w) -> "' >> ~/.bashrc && \
+    source /.venv/bin/activate && \
+    console --config-file=/root/lib/console_link/services.yaml completion bash > /usr/share/bash-completion/completions/console
```
gregschohn, line 66: why are we running this file w/ a services.yaml file at docker image build time?  Could auto-completion pickup values from that file?

Is it safer to put this script in the shell initialization (.bashrc?)

lewijacn: We need some config-file for the `completion bash` function to be successful currently, so I've used the existing file we have there. Our subsequent commands to the console library will load the default config file at `/etc/migration_services.yaml` if it exists or throw error if it doesn't, so there shouldn't be a chance that these values are picked up in some way.
This is only something we want to run once, I'm not sure if we'll hit any issue having it in the .bashrc but we can investigate

lewijacn: Pushed an update to have this in `.bashrc`

gregschohn (approved):
Thanks Tanner

PR #1226, opened 2025-01-09
Added version-specific OpenSearchWorkCoordinators

chelma:
### Description
* Checkpoint.  This PR updates the RFS code to use a version-specific OpenSearchWorkCoordinator.  Currently, there's only one supported version (OS 2), which presents the existing behavior.  E.g. there's no functional changes in this PR, it just sets up the plumbing to add Coordinators for new versions (like ES 6.8).

### Issues Resolved
* https://opensearch.atlassian.net/browse/MIGRATIONS-2285

### Testing
Ran the unit tests

### Check List
- [X] New functionality includes testing
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1230, opened 2025-01-10
Add an OpensearchClient for ES 6.x

mikaylathompson:
### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
```diff
@@ -30,7 +32,7 @@
 import reactor.util.retry.Retry;
 
 @Slf4j
-public class OpenSearchClient {
+public abstract class OpenSearchClient {
```
peternied, line 35: Please address this new sonarqube issue below

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
```diff
@@ -269,6 +271,30 @@ public boolean hasIndex(String indexName) {
         return hasObjectCheck(indexName, null);
     }
 
+//    private String appendUrlParameters(String path, Map<String, String> urlParameters) {
```
peternied, line 274: Uncomment?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
```diff
@@ -0,0 +1,45 @@
...
+import org.opensearch.migrations.bulkload.common.http.ConnectionContext;
+import org.opensearch.migrations.bulkload.version_es_6_8.OpenSearchClient_ES_6_8;
+import org.opensearch.migrations.bulkload.version_os_2_11.OpenSearchClient_OS_2_11;
+import org.opensearch.migrations.reindexer.FailedRequestsLogger;
+
+@RequiredArgsConstructor
+@Slf4j
+public class OpenSearchClientFactory {
+    // Version can be null, and if so, the "default" client, for OS_2_11 will be provided (matching the pre-factory behavior)
+    private final Version version;
```
peternied, line 17: The version should be detectable - so we don't need to pass a parameter to this factory, here is an example https://github.com/opensearch-project/opensearch-migrations/pull/1226/commits/163d4829346dc228acc5519aa8257082b9380645

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public OpenSearchClient get(
+        ConnectionContext connectionContext
+    ) {
+        if (version == null || VersionMatchers.isOS_1_X.test(version) || VersionMatchers.isOS_2_X.test(version) || VersionMatchers.isES_7_X.test(version)) {
```
peternied, line 22: Using `or(...)` with predicates is much more readable IMO
```suggestion
        if (VersionMatchers.isOS_1_X.or(VersionMatchers.isOS_2_X).or(VersionMatchers.isES_7_X).test(version)) {
```

peternied (commented):
Look good so far - thanks for doing this work @mikaylathompson

Looks like there are still a couple of other changes in flight on this PR.  I'll re-review when this is published for review.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
mikaylathompson: @peternied I've been struggling with this point a bit, and I don't think that example is quite relevant. I agree that there's probably a better way to do this, but I don't think it's quite that easy, because you need a client to detect the version, but you can't create a client without a version

mikaylathompson: So we're in a chicken-egg situation. We could have the client factory create a temporary client of the default (latest) version, check the version on the cluster, and then create a "real" client of the correct version, but I don't love that creating a Java class would then inherently put a load on the cluster--that seems like a pretty hidden side effect that could have consequences.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
mikaylathompson: I didn't know this syntax! Thanks!

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
chelma: @mikaylathompson I see your point, and I wish I was a bit more Java-savvy to propose a more elegant solution, but the general approach of having a "bootstrap" client that's our initial guess on the right one and finds the correct answer for us seems like the best overall direction.  The question is how to surface the potentially surprising side effect.  Maybe something like:

```java
var factory = new OpenSearchClientFactory()
factory.loadCorrectVersion()
var targetClient = factory.build() // Throws an exception if the loadCorrectVersion() method hasn't been invoked
```

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
```diff
@@ -135,7 +135,8 @@ public static void main(String[] args) throws Exception {
     private ICreateSnapshotContext context;
 
     public void run() {
-        var client = new OpenSearchClient(arguments.sourceArgs.toConnectionContext());
+        var clientFactory = new OpenSearchClientFactory(null);
```
chelma, line 138: Passing in `null` here seems a bit wonky.  See my comment about an alternative interface, curious what you think.

chelma (commented):
Overall looking good; just the one concern about the factory interface.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
mikaylathompson: re: discussion with Peter, proposal is to update this to move the version-checking code (with the RestClient, not an OpenSearchClient) into the factory, and have an api a la:
```
var factory = new OpenSearchClientFactory();
var targetClient = factory.determineVersionAndCreate(connectionContext);
```

RFS/src/test/java/org/opensearch/migrations/bulkload/version_es_6_8/OpenSearchClient_ES_6_8_Test.java
```diff
@@ -0,0 +1,117 @@
...
+    @Test
+    void testBulkRequestHasType_doc() {
+        // This test gets stuck! I can't figure out why. Somewhere in the `sendBulkRequest`, it's presumably waiting for
+        // an async to return, but I can't figure out what. Running in debug hasn't turned anything up.
+        var docId1 = "tt1979320";
+        var docId2 = "tt0816711";
+
+        var successResponse = bulkItemResponse(false, List.of(itemEntry(docId1), itemEntry(docId2)));
+        var bulkDocs = List.of(createBulkDoc(docId1), createBulkDoc(docId2));
+        when(restClient.postAsync(any(), any(), any())).thenReturn(Mono.just(successResponse));
```
peternied, line 96: This was causing the test to hang be cause postAsync(3 args) was never called, whereas postAsync(4 args) is on the sendBulkRequest method.  
```suggestion
        when(restClient.postAsync(any(), any(), any(), any())).thenReturn(Mono.just(successResponse));
```

peternied (commented):
Added suggestion that should get `testBulkRequestHasType_doc` unblocked

RFS/src/test/java/org/opensearch/migrations/bulkload/version_es_6_8/OpenSearchClient_ES_6_8_Test.java
```diff
@@ -0,0 +1,120 @@
...
+        assertThat(result.get().toPrettyString(), containsString(rawJson));
+        // The interface is to send back the passed json if on success
+        assertThat(result.get().toPrettyString(), not(containsString(createdItemRawJson)));
+
+        Mockito.verify(restClient).putAsync("indexName?include_type_name=false", "{}", null);
+    }
+
+    @Test
+    void testBulkRequestHasType_doc() {
+        // This test gets stuck! I can't figure out why. Somewhere in the `sendBulkRequest`, it's presumably waiting for
```
peternied, line 91: Nit: Please remove any comments that are no longer valid.  Comments that are outdated can create considerable confusion.

peternied (approved):

PR #1237, opened 2025-01-16
Correctly read compatibility mode from cluster settings when it was unset

jugal-chauhan:
### Description
This change updates the `_cluster/settings` API call in the OpenSearchClient to include default settings. The modification ensures that the compatibility mode check accurately reflects the current cluster state, even when the setting is at its default value. This addresses an issue where the compatibility mode value could remain `true` even after being disabled via the Migration Console.

Specifically, this change:

- Updates the `getClusterVersion()` method in OpenSearchClient.java to use `_cluster/settings?include_defaults=true`.
- Modifies corresponding test cases in OpenSearchClientTest.java to reflect this change.

By including default settings in the API call, we can correctly interpret the compatibility mode status in both scenarios.

### Issues Resolved
This PR addresses the issue described in [MIGRATIONS-2325](https://opensearch.atlassian.net/browse/MIGRATIONS-2325), where the compatibility mode status was not accurately detected when set to its default value.

### Testing
Updated unit tests in OpenSearchClientTest.java to cover the modified API call.
Ran the following test suites, all of which completed successfully:

- `./gradlew :RFS:test`
- `./gradlew :RFS:isolatedTest`
- `./gradlew :RFS:slowTest`
- `./gradlew test `

### Check List

### DIFF
--- a/CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
+++ b/CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshot.java
@@ -64,7 +64,7 @@ public class TestCreateSnapshot {
                 fl -> {
                     if (Objects.equals(fl.uri(), "/")) {
                         return new SimpleHttpResponse(ROOT_RESPONSE_HEADERS, ROOT_RESPONSE_ES_7_10_2, "OK", 200);
-                    } else if (Objects.equals(fl.uri(), "/_cluster/settings")) {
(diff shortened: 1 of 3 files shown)


peternied (approved):

gregschohn:
For my own future knowledge... The calls to get settings return the overridden values and assume that the caller already knows about the defaults that would have been present for this particular version of ES/OS.  Given that default settings have changed a lot over time and that it may not be immediately obvious what default values will be present, we want to get the settings w/ an override to include ALL settings.

PR #1248, opened 2025-01-23
[K8s] Initial cut of working RFS local backfill with shared snapshot volume

lewijacn:
### Description
This change modifies our k8s deployment for MA and MCC to use a shared snapshot volume on the local filesystem to allow local E2E testing of RFS in a k8s cluster.

#### Key Implications
1. This does introduce a dependency that MA and MCC are deployed into the same namespace since PVCs cannot conventionally span across namespaces, and the snapshot volume is needed by both the source cluster and MA services (RFS, Migration Console)
2. This change will require MA to already be deployed for the MCC source cluster to start up successfully (currently will stay in Pending state looking for the PVC until it exists), as MA is creating the shared snapshot volume (unless MCC disables the snapshot volume)
3. The current additional parameters in the MA `values.yaml` for RFS and the usage of `migration_services.yaml` are temporary until shared config map changes are in place

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2304

### Testing
Local k8s testing

### Check List


deployment/k8s/charts/components/bulkLoad/templates/deployment.yaml
```diff
@@ -34,6 +34,15 @@ spec:
           volumeMounts:
             - name: {{ $envMountName }}
               mountPath: /shared
+            {{- if .Values.snapshotVolumeEnabled }}
```
peternied, line 37: I see checks for the shared log volume and the snapshot volume to be enabled, why are check if they are enabled?  I would assume that if we are deploying any part of MA the shared log volume is always enabled, is that correct?

deployment/k8s/README.md
```diff
@@ -49,22 +58,24 @@ The full environment helm charts consists of:
...
 
-**Note**: For first-time deployments and deployments after changes have been made to a dependent helm package, such as the `migration-console` chart, the following command is needed to update dependent charts
+**Note**: For first-time deployments and deployments after changes have been made to a dependent helm package, such as the `migrationConsole` chart, the following command is needed to update dependent charts
 ```shell
-helm dependency update migration-assistant
+helm dependency update
 ```
 
-The full environment helm chart can be deployed with the helm command
+The full Migration Assistant environment with Mock Customer Clusters can be deployed with the following helm commands
```
peternied, line 66: [Clarity] I could see cases for creating shorthands, but Mock Customer Clusters -> `mcc` is too much for me, can we name this test-clusters or something.

While MigrationAssistant -> `ma` is clearer, and we will use ma quite often in our systems can we avoid the acronym?

peternied (commented):
Thanks for the update - did a quick pass on how some of this works, looking forward to deploying it myself!

deployment/k8s/charts/components/bulkLoad/templates/deployment.yaml
lewijacn: This component chart could potentially be deployed independently, in which case you may want to use something like the local filesystem instead of mounting a separate volume here

deployment/k8s/README.md
lewijacn: Personally I like the shorthands to type commands quickly and these both map to their chart names abbreviated so I haven't felt any confusion around them. My anticipation is that normal users will be abstracted away from this naming completely, and its more developers or advanced users that will be playing with these commands.

deployment/k8s/charts/components/bulkLoad/templates/deployment.yaml
peternied: Can this be written in such a way that this chart doesn't need to know if its a local file system or volume mount?  Seems like a concern outside of this chart.

lewijacn: I don't think so, but maybe we're talking past each other. I saw this as leaving the option to mount a snapshot volume or not, in which case it seems like the Chart will need to know whether to mount the volume in the container definition or have no volume mount

peternied: Maybe I'm thinking the charts works in a way it doesn't - lets chat offline thanks!

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
```diff
@@ -19,24 +20,37 @@ capture-proxy:
       yamlPath: "allow_insecure"
 
 migration-console:
+  snapshotVolumeEnabled: true
+  snapshotVolumePvc: "snapshot-volume-pvc"
   parameters:
     nothingImportant:
       value: false
```
gregschohn, line 27: darn, just FYI - these should be deleted

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
```diff
@@ -19,24 +20,37 @@ capture-proxy:
...
     targetHost:
-      value: https://opensearch-cluster-master.mcc:9200/
+      value: https://opensearch-cluster-master:9200
+      allowRuntimeOverride: true
+    targetUsername:
+      value: admin
+      allowRuntimeOverride: true
+    targetPassword:
+      value: myStrongPassword123!
       allowRuntimeOverride: true
```
gregschohn, line 47: once the shared config stuff merges in, it would be good to put these into shared configs.  Longer term, it will be cool to have separate targets in the shared config - like a targets map for each of the different target migrations that we're testing (then select out which one we want).  We need some migration console support for that UX first though.

deployment/k8s/charts/components/bulkLoad/templates/deployment.yaml
gregschohn: It's the chart's concern to do this wiring.  If this wasn't enabled, then we'd be using a local FS.
If it wasn't the chart's responsibility to know how to setup the pod and its mounts, whose responsibility would it be?

deployment/k8s/charts/components/bulkLoad/templates/deployment.yaml
```diff
@@ -34,6 +34,15 @@ spec:
...
+            - name: snapshot-volume
+              mountPath: /snapshot
+            {{- end }}
       volumes:
         - name: {{ $envMountName }}
           emptyDir: {}
+        {{- if .Values.snapshotVolumeEnabled }}
```
gregschohn, line 44: Could you store this value in a local variable?  Maybe that isn't a good idea, but that's the pattern I was following, but I'm happy to stop ;).

deployment/k8s/charts/components/migrationConsole/migration_services.yaml
```diff
@@ -0,0 +1,32 @@
+source_cluster:
```
gregschohn, line 1: can you file a jira bug to fix the creation of this file.  I was starting a fix but got hung up on three other things.  The last issue that I was looking at was the config_watcher.py file wasn't normalizing the casing of the configmap key names into the jinja template properly.

deployment/k8s/charts/sharedResources/snapshotVolume/templates/pvc.yml
```diff
@@ -1,6 +1,3 @@
-# Only create PVC, if PVC with same name does not already exist
-{{- $pvc := lookup "v1" "PersistentVolumeClaim" .Release.Namespace .Values.volume.pvcClaimName -}}
-{{- if not $pvc }}
```
gregschohn, line 3: this went away because now we are creating the volume as its own sub-chart?

gregschohn (approved):

deployment/k8s/README.md
```diff
@@ -49,22 +58,24 @@ The full environment helm charts consists of:
 * Target cluster
 * Migration services
 
-**Note**: For first-time deployments and deployments after changes have been made to a dependent helm package, such as the `migration-console` chart, the following command is needed to update dependent charts
+**Note**: For first-time deployments and deployments after changes have been made to a dependent helm package, such as the `migrationConsole` chart, the following command is needed to update dependent charts
```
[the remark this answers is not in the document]
lewijacn: Done

deployment/k8s/charts/components/bulkLoad/templates/deployment.yaml
lewijacn: Changed, open to changing the pattern as well if we don't like it

deployment/k8s/charts/components/migrationConsole/migration_services.yaml
lewijacn: Raised: https://opensearch.atlassian.net/browse/MIGRATIONS-2383

deployment/k8s/charts/sharedResources/snapshotVolume/templates/pvc.yml
lewijacn: I don't think any component chart should have a sub-chart dependency on a shared resource. I think I have found this structure to be a bit hack-ish and don't want to continue it. We were trying to have this PVC be something that could be shared between components as well as individually created by a component. I've taken the stance now to have this just as a shared PVC that the aggregate chart will create. If we find this is problematic for deploying individual charts I think there are alternatives but this feels cleaner to me

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
lewijacn: These got deleted with your latest PR 👍

Issue #1247, opened 2025-01-23
[BUG] RFS Fails to read ES 5 created indices on ES 6 Snapshot


AndreKurait:
### What is the bug?
A Elasticsearch 6 cluster may have indices originally created on ES 5 (e.g. via snapshot restore from an ES 5 snapshot). These indices will contain segments which were written on ES 5 (and thus Lucene 6) and may exist indefinitely on the cluster.

When RFS runs on the index, when it encounters segments written on ES 5, the following error occurs preventing the migration from proceeding

```
Caused by: java.lang.IllegalArgumentException: An SPI class of type org.apache.lucene.codecs.Codec with name 'Lucene62' does not exist. You need to add the corresponding JAR file supporting this SPI to your classpath. The current classpath supports the following names: [Lucene99, Lucene80, Lucene84, Lucene86, Lucene87, Lucene70, Lucene90, Lucene91, Lucene92, Lucene94, Lucene95]
at org.apache.lucene.util.NamedSPILoader.lookup(NamedSPILoader.java:113) ~[lucene-core-9.11.1.jar:9.11.1 
```

### What are your migration environments?
Source: ES 6.8, with index created on ES 5.6
Target: OS 2.17 (irrelevant)

### How can one reproduce the bug?
Given a snapshot from an ES 5.x cluster, this can be restored onto an ES 6 cluster. Then a snapshot can be taken and reindex-from-snapshot can be initiated encountering the error.

An automated test reproducing the issue has been created [here](https://github.com/opensearch-project/opensearch-migrations/blob/eedfeb848eb0dd8ac8e9c16e728e5dcdc5a2852b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SnapshotAndRestoreES5Test.java).

### What is the expected behavior?
RFS reads the segments without issue.

AndreKurait:
## Temporary workaround

We've pushed a branch forcing lucene 7 in order to enable compatibility with lucene 6 segments on branch ``.

With Migration Assistant, this can be used with the following commands on the bootstrap box:
```
cd /opensearch-migrations
git stash
git fetch
git checkout es-6-full-support
git stash pop
cd deployment/cdk/opensearch-service-migration/
./buildDockerImages.sh
./deploy.sh <CONTEXT_ID e.g. default>
```

AndreKurait:
> ## Temporary workaround
[...]

PR #1253, opened 2025-01-28
Support for ES6 using indexes created in ES5

peternied:
### Description
Support loading both Lucene 7 & 9 inside of RFS using shadow jar

### Issues Resolved
- Resolves https://github.com/opensearch-project/opensearch-migrations/issues/1247
- Resolves [MIGRATIONS-2370](https://opensearch.atlassian.net/browse/MIGRATIONS-2370) Support for ES5 created indices on ES6 cluster

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/common/ByteArrayIndexInput.java
```diff
@@ -3,7 +3,7 @@
 import java.io.EOFException;
 import java.io.IOException;
 
-import org.apache.lucene.store.IndexInput;
+import shadow.lucene9.org.apache.lucene.store.IndexInput;
```
chelma, line 6: What are your thoughts on how to handle when support for Lucene 10/OS 3.X becomes a requirement?  What if instead of having `lucene9` to represent the "latest" Lucene version, we do something like `lucenelatest`?  That way, we can just swap the version number in gradle and not have to update all our imports.  If it turns out we need to explicitly specify a version here, that's a whole different can of worms.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
```diff
@@ -66,7 +66,7 @@ public OpenSearchClient determineVersionAndCreate(RestClient restClient, FailedR
     private Class<? extends OpenSearchClient> getOpenSearchClientClass(Version version) {
         if (VersionMatchers.isOS_1_X.or(VersionMatchers.isOS_2_X).or(VersionMatchers.isES_7_X).test(version)) {
             return OpenSearchClient_OS_2_11.class;
-        } else if (VersionMatchers.isES_6_X.test(version)) {
+        } else if (VersionMatchers.isES_6_X.or(VersionMatchers.isES_5_X).test(version)) {
```
chelma, line 69: I'm confused - are we adding support for ES5 indicies in ES6, or support for ES5, in this PR?

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader.java
```diff
@@ -0,0 +1,39 @@
...
+public interface LuceneDocumentsReader {
+    Flux<RfsLuceneDocument> readDocuments(int startSegmentIndex, int startDoc);
+
+    @Slf4j
+    @AllArgsConstructor
+    public static class Factory {
+        public ClusterSnapshotReader snapshotReader;
+
+        public LuceneDocumentsReader getReader(Path path) {
+            if (VersionMatchers.isES_5_X.or(VersionMatchers.isES_6_X).test(snapshotReader.getVersion())) {
```
chelma, line 22: Unless I'm missing something, we shouldn't use a Lucene 7 reader to read an ES 5 index, because that ES5 index may contain Lucene 5 segments.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+                }
+        });
+    }
+
+    protected DirectoryReader getReader() throws IOException {
+        try (var directory = FSDirectory.open(indexDirectoryPath)) {
+            var commits = DirectoryReader.listCommits(directory);
+            var latestCommit = commits.get(commits.size() - 1);
+
+            return DirectoryReader.open(latestCommit, 0, null);
```
chelma, line 99: Do we want to set the minimum Lucene version here to Lucene 8, given we're explicitly fullfilling Lucene 7 and before with separate readers?  Feels like a good safety mechanism to ensure we aren't "surprised" by the wrong LuceneDocumentsReader class being used.

build.gradle
```diff
@@ -4,6 +4,7 @@ plugins {
     id "com.diffplug.spotless" version '6.25.0'
     id 'io.freefair.lombok' version '8.6' apply false
     id 'jacoco'
+    id 'com.github.johnrengelman.shadow' version '8.1.1' apply false
```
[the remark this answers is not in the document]
chelma: Should we be using the `com.gradleup.shadow` version instead?

Per the GitHub page:
Previously this plugin was developed by [@johnrengelman](https://github.com/johnrengelman) and published under the ID [com.github.johnrengelman.shadow](https://plugins.gradle.org/plugin/com.github.johnrengelman.shadow) before maintenance was transferred to the [GradleUp organization](https://github.com/GradleUp) to ensure future development, see https://github.com/GradleUp/shadow/issues/908.

If you are still using the old plugin ID in your build script, we recommend to switch to the new plugin ID [com.gradleup.shadow](https://plugins.gradle.org/plugin/com.gradleup.shadow) and update to the latest version to receive all the latest bug fixes and improvements.

chelma (commented):
Thanks for the work on this! Have a few questions.

chelma:
@peternied Oh, also - do we have a test to confirm this enables us to read Lucene 6 segments?  I didn't see one, and don't want to repeat the last time we had a PR claiming to solve this issue. 😛

peternied:
@chelma The EndToEnd tests cover this scenario by using RFS with a ES5 cluster, I confirmed by disabling the Lucene7 reader and got the same error when it was reported.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
peternied: Good eye - I'm doing a little ES5 work here that makes the testing easier.

For RFS's purposes the lucene version of the segments in the snapshot is the key difference.  By adding ES5 in the EndToEnd tests we get the testing coverage rather than building a new test that creates an ES5 cluster, creates indices and documents, upgrades to ES6, and then migrates.

Note; I've updated https://opensearch.atlassian.net/browse/MIGRATIONS-2276 's AC to include adding lucene6 when handing ES5 clusters because we will have the same issue with ES2 -> ES5 upgraded clusters.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader.java
peternied: Good callout, I've updated https://opensearch.atlassian.net/browse/MIGRATIONS-2276 's AC to include adding lucene6 when handing ES5 clusters because we will have the same issue with ES2 -> ES5 upgraded clusters.

build.gradle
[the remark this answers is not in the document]
peternied: I'll look into doing the gradle upgrade, 🤞 it isn't too involved.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/ByteArrayIndexInput.java
chelma: Per conversation w/ @peternied, the plan is to assume all of these classes are inherently versioned and begin moving them to the new `org.opensearch.migrations.bulkload.lucene` package when that becomes an issue.  E.g. we'll lazily resolve that issue.

build.gradle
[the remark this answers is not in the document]
peternied: Note; I'd really prefer to take the gradle upgrade out of band from this PR - I'll be working on it as a fast follow for a second PR.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClientFactory.java
chelma: Per convo, this change means that we will unofficially be able to handle ES 5 source domains for RFS.  That is, if someone points RFS at such a domain, it may just work even though it's not in our compatibility matrix.  Metadata migrations will fail due to lack of a transform path.  This all seems fine, as we're adding full ES 5 support next anyways.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader.java
chelma: TBH, I was a bit nervous here but we don't officially support ES 5 sources yet so any issues that arise will be because folks are using RFS off-label.  The task to add official ES 5 support will resolve this gap.  Fine w/ me.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+import shadow.lucene9.org.apache.lucene.index.DirectoryReader;
+import shadow.lucene9.org.apache.lucene.index.IndexReader;
+import shadow.lucene9.org.apache.lucene.index.LeafReaderContext;
+import shadow.lucene9.org.apache.lucene.index.SoftDeletesDirectoryReaderWrapper;
+import shadow.lucene9.org.apache.lucene.store.FSDirectory;
+import shadow.lucene9.org.apache.lucene.util.BytesRef;
+
+@RequiredArgsConstructor
+@Slf4j
+public class LuceneDocumentsReader9 implements LuceneDocumentsReader {
```
chelma, line 27: I missed in the previous review that we duplicated the majority of the behavior from the other reader so now we have two copies.  How much was changed in this version vs. the other?  If it's not too much work, maybe we can centralize the core behavior in this PR; seems like that should be a light lift.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/ByteArrayIndexInput.java
peternied: I'd rather we are explicit about which version of lucene is used, this should be trivial since the source versions is expected before we read a snapshot.  For some of these lingering places where we have lucene references, I think we should cleanup/consolidate them.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
peternied: I've created a task https://opensearch.atlassian.net/browse/MIGRATIONS-2373 for following up, it should be a prereq for lucene6 support.

build.gradle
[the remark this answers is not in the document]
peternied: Created follow up task https://opensearch.atlassian.net/browse/MIGRATIONS-2375 for upgrading gradle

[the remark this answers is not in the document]
chelma: Yeah, that seems out of scope for this task.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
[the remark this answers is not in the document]
chelma: Hadn't considered it, but that does make sense.  Just making a bunch of interfaces to abstract everything doesn't feel like automatic right answer here either; this feels like some investigation/conversation is required.  Agree to move this to a follow-up task to resolve the Cx issue.

chelma (commented):
There's a couple new SonarQube violations in the new/modified code; resolve those and I'll approve!

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+            .flatMap(ctx -> getReadDocCallablesFromSegments(ctx,
+                    // Only use startDocId for the first segment we process
+                    ctx.ord == startSegmentIndex ? startDocId : 0))
+            .flatMap(c -> Mono.fromCallable(c)
+                            .subscribeOn(sharedSegmentReaderScheduler), // Scheduler to read documents on
+                    maxDocumentsToReadAtOnce) // Don't need to worry about prefetch before this step as documents aren't realized
+            .doOnTerminate(sharedSegmentReaderScheduler::dispose);
+    }
+
+    Publisher<Callable<RfsLuceneDocument>> getReadDocCallablesFromSegments(LeafReaderContext leafReaderContext, int startDocId) {
```
jugal-chauhan, line 128: I've been reading through the Apache Lucene documentation, and I understand how Lucene handles deleted, soft-deleted, and updated documents, ensuring that only live documents are returned in search results, while keeping non-live at dish until merge. 

Question = Since Lucene excludes non-live documents from searches, what was the reason for adding extra logic to filter out these documents again when retrieving them for migration? Wouldn't it have been more efficient to modify the logic such that "non-live" documents that are hard-deleted or updated are immediately purged from the Lucene index and from disk instead of keeping older versions on disk?

RFS/src/test/java/org/opensearch/migrations/bulkload/PerformanceVerificationTest.java
```diff
@@ -7,27 +7,27 @@
...
-import org.apache.lucene.index.IndexWriter;
-import org.apache.lucene.index.IndexWriterConfig;
-import org.apache.lucene.store.ByteBuffersDirectory;
-import org.apache.lucene.util.BytesRef;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
 import reactor.core.publisher.Mono;
 import reactor.core.scheduler.Schedulers;
+import shadow.lucene9.org.apache.lucene.document.Document;
```
jugal-chauhan, line 23: Question = Now that we have implemented shadow jars to manage and isolate lucene 7 and lucene 9, does this mean we standardize on using "shadow.lucene9.org.apache.lucene" for all new code that requires Lucene functionality in the future and never use "org.apache.lucene" ?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_7_10/ShardMetadataData_ES_7_10.java
```diff
@@ -16,7 +16,7 @@
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.Getter;
-import org.apache.lucene.util.BytesRef;
+import shadow.lucene9.org.apache.lucene.util.BytesRef;
 
 @Getter
 public class ShardMetadataData_ES_7_10 implements ShardMetadata {
```
jugal-chauhan, line 22: Question = Is it okay to assume that abstraction layer will detect version-specific API calls and by itself adapt and determine if it should use Lucene 7 or 9 ?

chelma (approved):
Ship it!  Thanks Peter.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+     * removed from the Lucene Index.  A similar thing happens when a document is updated; the old document is marked as
+     * "deleted" in the Lucene segment and the new version of the document is added in a new Lucene segment.  Until a merge
+     * occurs, both the old and new versions of the document will exist in the Lucene Index in different segments, though only
+     * the new version will be returned in search results.  This means that from an ES/OS perspective, you could have a single
+     * document that has been created, deleted, recreated, updated, etc. multiple times at the Elasticsearch/OpenSearch level
+     * and only a single version of the doc would exist when you queried the ES/OS Index - but every single iteration of that
+     * document might still exist in the Lucene segments on disk, all of which have the same _id (from the ES/OS perspective).
+     *
+     * Additionally, Elasticsearch 7 introduced a feature called "soft deletes" which allows you to mark a document as
+     * "deleted" in the Lucene Index without actually removing it from the Lucene Index.  From what I can gather, soft deletes
```
gregschohn, line 53: Instead of "From what I gather" - you can reference the link below, as you do, and just say that it's used for certain shard synchronization or management issues, then cite the link.  Farther below, it's worth calling out that soft-deleted docs WILL be eventually FULLY-deleted lucene docs & that's why we don't want to load them onto the target.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+     * still present in the Lucene Index.  The status of whether any given document is "soft deleted" or not is stored in the
+     * Lucene Index itself.  By default, Elasticsearch 7+ Indices have soft deletes enabled;  this is an Index-level setting.
+     * Just like deleted documents and old versions of updated documents, we don't want to reindex them agaisnt the target
+     * cluster.
+     *
+     * In order to retrieve only those documents that would be considered "live" in ES/OS, we use a few tricks:
+     * 1. We make sure we use the latest Lucene commit point on the Lucene Index.  A commit is a Lucene abstraction that
+     *     comprises a consistent, point-in-time view of the Segments in the Lucene Index.  By default, a DirectoryReader
+     *     will use the latest commit point.
+     * 2. We use a concept called "liveDocs" to determine if a document is "live" or not.  The liveDocs are a bitset that
```
gregschohn, line 71: Is this a bitset or compressed bitset?

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * Soft deletes works by having the application writing the Lucene Index define a field that is used to mark a document as
+     * "soft deleted" or not.  When a document is marked as "soft deleted", it is not returned in search results, but it is
+     * still present in the Lucene Index.  The status of whether any given document is "soft deleted" or not is stored in the
+     * Lucene Index itself.  By default, Elasticsearch 7+ Indices have soft deletes enabled;  this is an Index-level setting.
+     * Just like deleted documents and old versions of updated documents, we don't want to reindex them agaisnt the target
+     * cluster.
+     *
+     * In order to retrieve only those documents that would be considered "live" in ES/OS, we use a few tricks:
+     * 1. We make sure we use the latest Lucene commit point on the Lucene Index.  A commit is a Lucene abstraction that
```
gregschohn, line 68: If there were two snapshots to a repo - a full one that successfully finished and a partial one that didn't, would this be true?  Could we get the latest commit point for some shards in an index from the second snapshot and snapshots further back in time from the first snapshot for other shards?

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+     * https://www.elastic.co/guide/en/elasticsearch/reference/7.10/index-modules-history-retention.html
+     *
+     * Soft deletes works by having the application writing the Lucene Index define a field that is used to mark a document as
+     * "soft deleted" or not.  When a document is marked as "soft deleted", it is not returned in search results, but it is
+     * still present in the Lucene Index.  The status of whether any given document is "soft deleted" or not is stored in the
+     * Lucene Index itself.  By default, Elasticsearch 7+ Indices have soft deletes enabled;  this is an Index-level setting.
+     * Just like deleted documents and old versions of updated documents, we don't want to reindex them agaisnt the target
+     * cluster.
+     *
+     * In order to retrieve only those documents that would be considered "live" in ES/OS, we use a few tricks:
```
gregschohn, line 67: s/tricks/strategies/ - or something else like algorithms.  Tricks imply that there's something non-obvious or maybe something that's only true in certain cases.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
```diff
@@ -0,0 +1,230 @@
...
+@RequiredArgsConstructor
+@Slf4j
+public class LuceneDocumentsReader9 implements LuceneDocumentsReader {
+
+    protected final Path indexDirectoryPath;
+    protected final boolean softDeletesPossible;
+    protected final String softDeletesField;
+ 
+    /**
+     * There are a variety of states the documents in our Lucene Index can be in; this method extracts those documents
```
gregschohn, line 34: Could you move this to the class documentation?
The below comments still hold, even if you didn't write them.  If you did - awesome, thanks for the work.  If you didn't, can you 1) make sure that these are only in one place (like an interface that all implementations inherit from)?  You can setup a follow-up task for it too if you like.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader9.java
gregschohn: Lucene keeps the old files on disk for the same reason that a purely functional data structures don't make modifications.  It's easier to reason with immutable data and its MUCH easier to distribute immutable data around in parallel.  That also means that it's a lot cheaper too (cache invalidations and doubts around race conditions kill performance of modern programs).

For more information, check out this page on [LSM Trees](https://en.wikipedia.org/wiki/Log-structured_merge-tree).  You can search for "Lucene lsm tree" for some articles about the similarities between Lucene and LSM trees.  Martin Kleppmann's "Designing Data-Intensive Applications" book is a great reference too.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader7.java
```diff
@@ -1,45 +1,35 @@
...
-public class LuceneDocumentsReader {
-
-    public static Function<Path, LuceneDocumentsReader> getFactory(ClusterSnapshotReader snapshotReader) {
-        return path -> new LuceneDocumentsReader(
-            path,
-            snapshotReader.getSoftDeletesPossible(),
-            snapshotReader.getSoftDeletesFieldData()
-        );
-    }
+public class LuceneDocumentsReader7 implements LuceneDocumentsReader {
```
gregschohn, line 27: For later, it we're able to isolate lucene7 vs lucene9 code and we only need one version in the runtime at once, it should be a chip-shot to load the things from the ServiceLoader interface (and not require shadowing at all).

It would also be nice to not have to require as much copied code (e.g. just building/linking for different versions, or injecting different version specific overrides)

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneDocumentsReader.java
```diff
@@ -0,0 +1,39 @@
...
+
+import org.opensearch.migrations.VersionMatchers;
+import org.opensearch.migrations.bulkload.common.RfsLuceneDocument;
+import org.opensearch.migrations.cluster.ClusterSnapshotReader;
+
+import lombok.AllArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import reactor.core.publisher.Flux;
+
+public interface LuceneDocumentsReader {
```
gregschohn, line 13: ILuceneDocumentsReader to show that it's an interface

gregschohn (commented):
Thanks for doing this.  I recognize that this is merged, but I think that a followup could address some of these minor issues.

PR #1263, opened 2025-01-31
Kafka clients upgrade for CVE-2024-31141

gregschohn:
### Testing
CICD

### Check List


lewijacn:
In the past we have tried to sync kafka versions across our whole system, kafka clients, MSK, kafka jar we bring into the Migration Console image, docker compose kafka version

That being said the Kafka clients should be backwards compatible with an older Kafka cluster version, and this upgrade is only for kafka clients. I also didn't see the exact 3.7.2 version in MSK to be able to sync across the whole system. So I think I'm ending up if we think there is enough severity just upgrading the clients should be fine

PR #1264, opened 2025-01-31
[Bug] Jenkins Replayer test needs updated transformation arg name

lewijacn:
### Description
See title

```
[INFO ] 2025-01-31 12:28:28,702788 [main] TrafficReplayer - Starting Traffic Replayer with id=ip-12-0-2-118.ec2.internal
Only one main parameter allowed but found several: "https://<target-cluster>:443" and "--transformer-config-base64"
```
I removed target cluster endpoint here

### Issues Resolved
N/A

### Testing
Jenkins testing

### Check List

### DIFF
--- a/vars/trafficReplayDefaultE2ETest.groovy
+++ b/vars/trafficReplayDefaultE2ETest.groovy
@@ -56,7 +56,7 @@ def call(Map config = [:]) {
         "domainRemovalPolicy": "DESTROY",
         "artifactBucketRemovalPolicy": "DESTROY",
         "trafficReplayerServiceEnabled": true,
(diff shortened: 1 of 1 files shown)


gregschohn (approved):
Should we put multiple aliases for "encoded" and "base64" on this & other options?  I think that we had a bifurcation somewhere between the two across different processes or flavors, hence the confusion

lewijacn:
> Should we put multiple aliases for "encoded" and "base64" on this & other options? I think that we had a bifurcation somewhere between the two across different processes or flavors, hence the confusion

I think we should just do a detailed sweep and get these on the same page, I'll try to raise a JIRA around this

PR #1266, opened 2025-01-31
[K8s] Introduce Helmfiles for MA deployment

lewijacn:
### Description
Change includes adding support deploying our component charts no longer under the umbrella of a parent chart with its limitations around dependency updating, but rather placing them in a Helmfile which can perform many Helm deployments under the hood and remove our nested Chart structure

### Issues Resolved
N/A

### Testing
Local Minikube deployments

### Check List

### DIFF
--- a/deployment/k8s/helmfile/helmfile.yaml
+++ b/deployment/k8s/helmfile/helmfile.yaml
@@ -1,0 +1,78 @@
+values:
+  - values-dev.yaml
+
+repositories:
+  - name: strimzi-kafka-operator
+    url: "https://strimzi.io/charts/"
+  - name: elasticsearch-repo
+    url: "https://helm.elastic.co"
+  - name: opensearch-repo
+    url: "https://opensearch-project.github.io/helm-charts/"
+
+releases:
+  - name: shared-configs
+    version: "0.1.0"
+    namespace: ma
+    chart: "../charts/sharedResources/sharedConfigs"
+  - name: snapshot-volume
+    version: "0.1.0"
+    namespace: ma
+    chart: "../charts/sharedResources/snapshotVolume"
+    condition: conditionalPackageInstalls.snapshotVolume.enabled
+  - name: captured-traffic-kafka-cluster
+    version: "0.1.0"
+    namespace: ma
+    chart: "../charts/sharedResources/baseKafkaCluster"
+    values:
+      - {{ toYaml .Values.capturedTrafficKafkaCluster | nindent 8 }}
+    condition: conditionalPackageInstalls.kafka.enabled
+
+  - name: migration-console
+    version: "0.1.0"
+    namespace: ma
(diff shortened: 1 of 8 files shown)


deployment/k8s/helmfile/values-dev.yaml
```diff
@@ -0,0 +1,172 @@
+conditionalPackageInstalls:
+  snapshotVolume:
+    enabled: true
```
gregschohn, line 3: Do these need this many layers?  Could we do w/ 2?

deployment/k8s/helmfile/values-dev.yaml
```diff
@@ -0,0 +1,172 @@
...
+      configMapName: "source-cluster"
+      configMapKey: "allowInsecure"
+    kafkaConnection:
+      source: otherConfig
+      configMapName: "kafka-brokers"
+      configMapKey: "brokers"
+
+
+migrationConsole:
+  snapshotVolumeEnabled: true
```
gregschohn, line 58: would be nice if we could have a single-source of truth for this value since it's shared in at least 2 other places.  Same for other occurrences of the pattern.  Is that possible w/ a helmfile?

deployment/k8s/helmfile/helmfile.yaml
```diff
@@ -0,0 +1,78 @@
...
+    needs:
+      - snapshot-volume
+    condition: conditionalPackageInstalls.source.enabled
+  - name: opensearch
+    version: "2.23.1"
+    namespace: ma
+    chart: "opensearch-repo/opensearch"
+    values:
+      - {{ toYaml .Values.target | nindent 8 }}
+    condition: conditionalPackageInstalls.target.enabled
```
gregschohn, line 78: I'd rather never install test resources along w/ production resources.  The risk of long term assumptions and entanglements is too great.  Even short term developer workflows could be negatively impacted.

gregschohn (commented):
I like how we don't have to adjust any of the existing charts.  Because of that, I'd be find including this support w/ helmfiles in our repo.  However, we need to continue to support the migrationAssistant chart so that we can eventually publish that to the artifact hub and elsewhere.  From what it looks like, helmfile doesn't have nearly the exposure that helm alone has so losing that distribution channel and inertia would be foolish - and more friction for users.

deployment/k8s/helmfile/values-dev.yaml
lewijacn: This was a weird requirement that I had from the helmfile to structure this way at least to my knowledge

lewijacn:
Closing this for now as the distribution factor of helmfile is our main concern, while this may be useful in the future for developer efficiency, distributing this would follow a non-typical pattern that is bound to cause friction. For now will remain with this umbrella chart approach

PR #1298, opened 2025-02-17
Reinstate gradle 8.12

peternied:
### Description
This changes reinstates gradle 8.12 for our project. Previously the gradle update was reverted because of an issue discovered during maven publishing that was unverified during our normal CI workflow.

- Add CI check that verifies maven publishing flow to detect this problem
- Adopt [management to library catalog](https://docs.gradle.org/current/userguide/version_catalogs.html), replaces commonDependencyVersionConstraints.
- Updated gradle dependencies to use catalog members

**Outstanding issues before leaving draft**
- [x] Make sure dependencies resolution has correct way to manage/track CVEs
- [x] Clean up any conflict resolution and filtering into centralized location and processes, document how this should be done.
- [x] Look into grouping common dependencies as bundles such as testcontainers, junit / mockito / hamcrest, and others
  - We could do this, which would be kind of nice, but it feels like feature creep at this point.  IMO this is a good starting point and we can build off it as needed. 

### Issues Resolved
- Resolves https://opensearch.atlassian.net/browse/MIGRATIONS-2396

### Check List


peternied:
Waiting on better time to get this change merged, closing until that time arrives.

DocumentsFromSnapshotMigration/build.gradle
```diff
@@ -48,29 +45,29 @@ dependencies {
...
 
 }
 
 application {
-    mainClassName = 'org.opensearch.migrations.RfsMigrateDocuments'
+    mainClass.set('org.opensearch.migrations.RfsMigrateDocuments')
 }
 
 // Cleanup additional docker build directory
+def dockerBuildDir = file("./docker/build")
```
gregschohn, line 68: why did you need to do this?

RFS/build.gradle
```diff
@@ -57,23 +55,23 @@ dependencies {
...
-    lucene7 "org.apache.lucene:lucene-backward-codecs:${luceneVersion7}"
+    lucene7 libs.lucene.v7.core
+    lucene7 libs.lucene.v7.backward.codecs
 
-    lucene9 "org.apache.lucene:lucene-core:${luceneVersion9}"
-    lucene9 "org.apache.lucene:lucene-analysis-common:${luceneVersion9}"
-    lucene9 "org.apache.lucene:lucene-backward-codecs:${luceneVersion9}"
+    lucene9 libs.lucene.v9.core
+    lucene9 libs.lucene.v9.analysis.common
+    lucene9 libs.lucene.v9.backward.codecs
```
gregschohn, line 74: This seems a lot nicer - clearer to see the intention

TrafficCapture/build.gradle
```diff
@@ -39,7 +36,6 @@ subprojects {
 
 allprojects {
     apply plugin: 'java'
-    apply plugin: 'org.owasp.dependencycheck'
```
gregschohn, line 42: Have we lost anything with this going away?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,156 @@
...
+graalvm = "24.1.1"
+reactor-test = "3.7.3"
+reactor-netty = "1.1.18"
+lucene7 = '7.7.3'
+lucene9 = "9.11.1"
+jni-java = "2.7.3"
+errorprone = "2.32.0"
+commons-collections = "4.4"
+jenkins-pipeline-unit = "1.23"
+groovy = "3.0.23"
```
gregschohn, line 42: can we make these alphabetical to make it easier to see dups/conflicts?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,156 @@
...
+reactor-netty-http = {module = "io.projectreactor.netty:reactor-netty-http", version.ref = "reactor-netty" }
+
+lucene-v7-core = { module = "org.apache.lucene:lucene-core", version.ref = "lucene7" }
+lucene-v7-backward-codecs = { module = "org.apache.lucene:lucene-backward-codecs", version.ref = "lucene7" }
+
+lucene-v9-core = { module = "org.apache.lucene:lucene-core", version.ref = "lucene9" }
+lucene-v9-analysis-common = { module = "org.apache.lucene:lucene-analysis-common", version.ref = "lucene9" }
+lucene-v9-backward-codecs = { module = "org.apache.lucene:lucene-backward-codecs", version.ref = "lucene9" }
+
+jni-java = { module = "com.hubspot.jinjava:jinjava", version.ref = "jni-java" }
```
gregschohn, line 148: let's fix this one to jinjava - jni and jin are very different things.  The hyphen also breaks the 'jinja' part

transformation/transformationPlugins/jsonMessageTransformers/jsonJinjavaTransformer/build.gradle
```diff
@@ -6,17 +6,17 @@ plugins {
...
     testImplementation testFixtures(project(path: ':testHelperFixtures'))
     testImplementation testFixtures(project(path: ':TrafficCapture:trafficReplayer'))
 
-    testImplementation group: 'org.junit.jupiter', name:'junit-jupiter-api'
-    testImplementation group: 'org.junit.jupiter', name:'junit-jupiter-params'
-    testImplementation group: 'org.slf4j', name: 'slf4j-api'
-    testRuntimeOnly group:'org.junit.jupiter', name:'junit-jupiter-engine'
+    testImplementation libs.junit.jupiter.api
+    testImplementation libs.junit.jupiter.params
+    implementation libs.slf4j.api
```
gregschohn, line 20: Can you rotate this upward for grouping

gregschohn (commented):
Thanks!  This is great to see the 'by-luck' (or whatever the undocumented feature way) api dependency version locks become something easier to understand.  Generally, almost every change in this PR is easier to understand (like foo = blah rather than foo blah).

2 top level comments.  1) how much of this is specific to 8.12?  If that's just a couple lines, it would be nice to see a change for the version catalog for the current gradle, then upgrade gradle.  If it's more than 15 minutes, I wouldn't worry about it.  2) I'd also love to see some library groups, but I think that's something that an enterprising soul can take on in a subsequent task.

TrafficCapture/build.gradle
peternied: I didn't see anything in CI, other task dependencies, and in documentation.  It looks like this is effectively another CVE scanner and its duplicated with Dependabot's scanner & SonarQube instance.

gradle/libs.versions.toml
peternied: Yup - I'll sort and then make edits there are a couple names such as 'aws' that I figure we should better name and group

gradle/libs.versions.toml
peternied: Thank you - this dyslexic developer sees nothing wrong :D

DocumentsFromSnapshotMigration/build.gradle
peternied: Yup, it was for the config cache.
```
* What went wrong:
Execution failed for task ':DocumentsFromSnapshotMigration:clean'.
> Cannot reference a Gradle script object from a Groovy closure as these are not supported with the configuration cache.
```

build.gradle
```diff
@@ -177,6 +182,7 @@ subprojects {
             systemProperty 'log4j2.contextSelector', 'org.apache.logging.log4j.core.selector.BasicContextSelector'
             // Verify assertions in tests
             jvmArgs = ['-ea', '-XX:+HeapDumpOnOutOfMemoryError']
+            systemProperty "junit.jupiter.execution.timeout.default", "5m"
```
peternied, line 185: This is a new behavior, tests taking longer than 5 minutes will fail with this timeout.  I've seen a couple of runs with no console output but seemingly there is a hung test case.

gregschohn: Thanks for calling this change out.  I'm fine w/ this behavior, but would also be happy to approve another change where this read a variable, etc to allow devs w/ slower machines to override.

gregschohn (approved):

PR #1300, opened 2025-02-18
testDocumentGenerator fetches endpoint and auth from config file

jugal-chauhan:
### Description
This change eliminates the need to manually provide endpoint and authentication details. The script now fetches these details from the `migration-services.yaml` file. It also leverages `boto3` to obtain AWS session credentials from within the migration console, improving security and ease of use.

### Testing
flake8
./gradlew test 
./gradlew spotlessCheck
./gradlew spotlessApply

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
@@ -5,10 +5,11 @@ import time
 import argparse
 from datetime import datetime
 import urllib3
-import os
 from collections import deque
 import logging
 import json
+from console_link.environment import Environment
+from console_link.models.cluster import Cluster
 
 # Disable InsecureRequestWarning
 urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
@@ -59,9 +60,6 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
 def parse_args():
     """Parse command line arguments."""
     parser = argparse.ArgumentParser()
-    parser.add_argument("--endpoint", help="Cluster endpoint e.g. http://test.elb.us-west-2.amazonaws.com:9200.")
-    parser.add_argument("--username", help="Cluster username.")
-    parser.add_argument("--password", help="Cluster password.")
     parser.add_argument("--enable_multi_type", action='store_true',
                         help="Flag to enable sending documents to a multi-type index.")
     parser.add_argument("--no-clear-output", action='store_true',
@@ -69,6 +67,9 @@ def parse_args():
                              "Helpful for piping to a file or other utility.")
     parser.add_argument("--requests-per-sec", type=float, default=10.0, help="Target requests per second to be sent.")
     parser.add_argument("--no-refresh", action='store_true', help="Flag to disable refresh after each request.")
+    parser.add_argument("--cluster", type=str, default="source", choices=["source", "target"],
+                        help="Specify 'source' or 'target' cluster.")
+    parser.add_argument("--config-file", default="/etc/migration_services.yaml", help="Path to config file")
     return parser.parse_args()
 
 
@@ -121,11 +122,17 @@ def calculate_sleep_time(request_timestamps, target_requests_per_sec):
     return max(0, sleep_time)
 
 
+def get_cluster_and_auth(config_file, cluster_type):
+    env = Environment(config_file)
+    cluster: Cluster = env.source_cluster if cluster_type == "source" else env.target_cluster
+    auth = cluster._generate_auth_object()
+    return cluster.endpoint, auth
+
+
 def main():
     args = parse_args()
-
-    url_base = args.endpoint or os.environ.get('SOURCE_DOMAIN_ENDPOINT', 'https://capture-proxy:9200')
-    auth = (args.username, args.password) if args.username and args.password else None
+    
+    url_base, auth = get_cluster_and_auth(args.config_file, args.cluster)
 
     session = requests.Session()
     keep_alive_headers = {'Connection': 'keep-alive'}


TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -121,11 +123,77 @@ def calculate_sleep_time(request_timestamps, target_requests_per_sec):
...
+
+    # Determine which cluster to target as endpoint, SOURCE or TARGET
+    cluster_type = args.target
+    cluster_config = get_cluster_config(config, cluster_type)
 
-    url_base = args.endpoint or os.environ.get('SOURCE_DOMAIN_ENDPOINT', 'https://capture-proxy:9200')
-    auth = (args.username, args.password) if args.username and args.password else None
+    # Extract endpoint and authentication details
+    url_base = cluster_config['endpoint']
+    auth = get_auth(cluster_config)
```
lewijacn, line 196: I'd really like to avoid duplicating some work we already have in the console library to determine what auth and endpoint to use. I would suggest looking into importing the class needed to make these commands into this script

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -59,16 +62,15 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
...
-    parser.add_argument("--password", help="Cluster password.")
+    # Removed endpoint, username, password
     parser.add_argument("--enable_multi_type", action='store_true',
                         help="Flag to enable sending documents to a multi-type index.")
     parser.add_argument("--no-clear-output", action='store_true',
                         help="Flag to not clear the output before each run. " +
                              "Helpful for piping to a file or other utility.")
     parser.add_argument("--requests-per-sec", type=float, default=10.0, help="Target requests per second to be sent.")
     parser.add_argument("--no-refresh", action='store_true', help="Flag to disable refresh after each request.")
+    parser.add_argument("--target", type=str, default="source", help="Specify 'source' or 'target' cluster.")
```
lewijacn, line 73: I would rename this to something like `cluster-type` as this could be pretty confusing

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -59,16 +62,15 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
 def parse_args():
     """Parse command line arguments."""
     parser = argparse.ArgumentParser()
-    parser.add_argument("--endpoint", help="Cluster endpoint e.g. http://test.elb.us-west-2.amazonaws.com:9200.")
```
lewijacn, line 62: I would suggest leaving the endpoint argument as an optional field that can be used to overwrite the endpoint. This seems useful in the cases where we want to use ALB endpoint, but source endpoint is the actual cluster endpoint

lewijacn (changes requested):

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -121,11 +123,77 @@ def calculate_sleep_time(request_timestamps, target_requests_per_sec):
...
+        with open(yaml_path, 'r') as f:
+            config = yaml.safe_load(f)
+        return config
+    except FileNotFoundError:
+        logger.error(f"Configuration file not found: {yaml_path}")
+        sys.exit(1)
+    except yaml.YAMLError as e:
+        logger.error(f"Error parsing YAML file: {e}")
+        sys.exit(1)
+
```
mikaylathompson, line 138: I'd reallly like us to avoid manually parsing the services.yaml in this tool. We already do this in the `cluster_link` library, and leveraging that means that any schema changes or improvements will automatically be used here, instead of needing to keep track that we need to do this in multiple places.

mikaylathompson (changes requested):

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
mikaylathompson: We had some slack discussion about this and about using the example of the cluster_tools library (like https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/cluster_tools/src/cluster_tools/base/main.py#L101 and https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/cluster_tools/src/cluster_tools/base/utils.py#L7), but happy to discuss more or work through issues.

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: Done

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: Changed it to '--cluster', as similar command exists (for ex: "console clusters clear-indices --cluster source/target")

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: I have made another commit importing console_link for environment and auth

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: Agreed! Refactored in next commit

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -59,16 +60,17 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
...
-    parser.add_argument("--password", help="Cluster password.")
+    # Removed endpoint, username, password
     parser.add_argument("--enable_multi_type", action='store_true',
                         help="Flag to enable sending documents to a multi-type index.")
     parser.add_argument("--no-clear-output", action='store_true',
                         help="Flag to not clear the output before each run. " +
                              "Helpful for piping to a file or other utility.")
     parser.add_argument("--requests-per-sec", type=float, default=10.0, help="Target requests per second to be sent.")
     parser.add_argument("--no-refresh", action='store_true', help="Flag to disable refresh after each request.")
+    parser.add_argument("--cluster", type=str, default="source", help="Specify 'source' or 'target' cluster.")
```
mikaylathompson, line 71: Ideally we would have some checking around this. Right now, if the user put in `sourcc` or whatever typo, it would treat it as the target cluster and do so basically silently.  I would like it if it threw an error if the value wasn't 'source' or 'target', for instance.

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -121,11 +123,19 @@ def calculate_sleep_time(request_timestamps, target_requests_per_sec):
...
+
 def main():
     args = parse_args()
-
-    url_base = args.endpoint or os.environ.get('SOURCE_DOMAIN_ENDPOINT', 'https://capture-proxy:9200')
-    auth = (args.username, args.password) if args.username and args.password else None
+    
+    url_base, auth = get_cluster_and_auth(args.config_file, args.cluster)
+    if args.endpoint:
+        url_base = args.endpoint
```
mikaylathompson, line 138: I think that if the user called this with a custom endpoint, it would grab the auth scheme from either the source or target cluster and use it, correct? Is that the expected behavior?
I can see that there might be some use for that (Tanner's example of talking to the ALB instead of the source directly), but I think we should have some logging here to make that behavior clear.

Especially if I, as the user, didn't specify a `--cluster` arg, so it defaulted to the source, I would be quite surprised if I was expect it to use no_auth and instead it used the source cluster's auth. Having a log message saying approximately "talking to cluster at [endpoint] using the auth for cluster [source/target] of type [sigv4/basic/no auth] specified in [config file]" would give the user a lot more context.l (flip side is that we don't want to log actual auth details like the password)

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: A choice containing 2 options have been added

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: This issue can be solved by a few checks. If ALB's endpoint matches the source cluster's endpoint, it will use source cluster's auth and vice versa.  Also having an additional condition that if no match found, use no auth.

Agreed! Logging to be enabled for better clarity for the endpoint routing, also stating the auth-type without any sensitive information

lewijacn (approved):
Thanks for changes

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -59,16 +60,23 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
...
     parser.add_argument("--no-refresh", action='store_true', help="Flag to disable refresh after each request.")
+    parser.add_argument("--cluster", type=str, default="source", choices=["source", "target"],
+                        help="Specify 'source' or 'target' cluster.")
+    parser.add_argument("--config-file", default="/etc/migration_services.yaml", help="Path to config file")
+    # parser.add_argument("--endpoint", help="Cluster endpoint e.g. http://test.elb.us-west-2.amazonaws.com:9200.")
+
+    # Temporary workaround for users who need to pass the ALB endpoint to capture live traffic:
+    # - Users should strictly use No-Auth OR Basic-Auth (no other auth methods) on both source and target clusters.
+    # - For Basic-Auth, keep Auth details (username and password) same for both the source and target clusters
+    #   if performing a switchover, to avoid disrupting the live traffic simulation.
```
mikaylathompson, line 78: A different workaround: a user can create their own `services.yaml` that just defines one cluster (I believe `target` one that's required to be present) to their specification (e.g. the right endpoint and auth method) and passes that in as the config file.

jugal-chauhan: Addressed

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -59,16 +60,30 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
...
+    # capture live traffic:
+    # - Users should strictly use No-Auth OR Basic-Auth (no other auth
+    #   methods) on both source and target clusters.
+    # - For Basic-Auth, keep Auth details (username and password) the same
+    #   for both the source and target clusters if performing a switchover,
+    #   to avoid disrupting the live traffic simulation.
+    # - Alternatively, users can create a custom config file (e.g.,
+    #   `custom_services.yaml`) that defines only the required cluster
+    #   (typically the target) with the appropriate endpoint and auth method,
+    #   then pass it in using `--config-file`.
```
peternied, line 85: Nit: Please do not include commented code.  If it something that shouldn't be forgotten, please capture the details in a new jira task, or keep a branch in your fork.

peternied (approved):

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
jugal-chauhan: New Jira Task has been created [here](https://opensearch.atlassian.net/browse/MIGRATIONS-241).

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -59,16 +60,18 @@ def send_multi_type_request(session, index_name, type_name, payload, url_base, a
...
                         help="Flag to enable sending documents to a multi-type index.")
     parser.add_argument("--no-clear-output", action='store_true',
                         help="Flag to not clear the output before each run. " +
                              "Helpful for piping to a file or other utility.")
     parser.add_argument("--requests-per-sec", type=float, default=10.0, help="Target requests per second to be sent.")
     parser.add_argument("--no-refresh", action='store_true', help="Flag to disable refresh after each request.")
+    parser.add_argument("--cluster", type=str, default="source", choices=["source", "target"],
+                        help="Specify 'source' or 'target' cluster.")
+    parser.add_argument("--config-file", default="/etc/migration_services.yaml", help="Path to config file")
+    # parser.add_argument("--endpoint", help="Cluster endpoint e.g. http://test.elb.us-west-2.amazonaws.com:9200.")
```
peternied, line 73: Please remove this commented out line

peternied (approved):

TrafficCapture/dockerSolution/src/main/docker/elasticsearchTestConsole/testDocumentGenerator.py
```diff
@@ -121,11 +122,19 @@ def calculate_sleep_time(request_timestamps, target_requests_per_sec):
...
+
 def main():
     args = parse_args()
-
-    url_base = args.endpoint or os.environ.get('SOURCE_DOMAIN_ENDPOINT', 'https://capture-proxy:9200')
-    auth = (args.username, args.password) if args.username and args.password else None
+    
+    url_base, auth = get_cluster_and_auth(args.config_file, args.cluster)
+    if args.endpoint:
+        url_base = args.endpoint
```
mikaylathompson, line 137: One last thing -- we no longer support an endpoint arg, right? Can we delete these two lines?

jugal-chauhan: Agreed!

mikaylathompson (approved):

PR #1306, opened 2025-02-24
Fix transformations with gzipped payload

AndreKurait:
### Description
Fixes transformation pipeline when applied on gzipped request payloads. Users do not need to do anything different to process requests with gzipped payloads, they will get uncompressed and processed as other requests and their uncompressed payloads will be available for transformations.

Updated the [NettyDecodedHttpRequestConvertHandler](https://github.com/opensearch-project/opensearch-migrations/pull/1306/files#diff-683d7876a64899e933620a1e583ef138bc1e53f3df78d9802b9d809b38d8b225) which feeds into the [NettyDecodedHttpRequestPreliminaryTransformHandler.java](https://github.com/opensearch-project/opensearch-migrations/pull/1306/files#diff-b2757a1cb0051fababe702fd43d5b1fb46d34d3a66301613b5fbe19b5c14a1b5) to support this. The change was to add a constructor parameter to the NettyDecodedHttpRequestConvertHandler to allow it to operate in two modes, an original request preservation mode which sends the original netty http request headers alongside the converted one, and the existing behavior. Also updated the NettyDecodedHttpRequestPreliminaryTransformHandler to pass through the original headers. What this achieves is that when the headers are modified that affect body parsing (transfer or content encoding) we pass the original request back through the processing pipleine, or when the body is accessed before it is propogated.

Updates the StrictCaseInsensitiveHttpHeadersMap to support case insensitive KeySet which was needed for the case insensitive remove operation.

### Issues Resolved
[MIGRATIONS-2069](https://opensearch.atlassian.net/browse/MIGRATIONS-2069)

### Testing
Added unit tests, see GHA

### Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/StrictCaseInsensitiveHttpHeadersMap.java
```diff
@@ -80,4 +92,39 @@ public int size() {
...
+
+            @Override
+            public boolean remove(Object key) {
+                String keyStr = (key instanceof String) ? (String) key : key.toString();
+                return lowerCaseToUpperCaseAndValueMap.remove(keyStr.toLowerCase()) != null;
+            }
+        };
+    }
+
+    private static String getCaseInsensitiveKey(Object key) {
```
lewijacn, line 124: Where is this used?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
```diff
@@ -180,7 +180,7 @@ private static Map<String, Object> convertRequest(
                         HttpByteBufFormatter.processHttpMessageFromBufs(
                             HttpMessageType.REQUEST,
                             byteBufStream,
-                            new NettyDecodedHttpRequestConvertHandler(transformationCtx),
+                            new NettyDecodedHttpRequestConvertHandler(transformationCtx, false),
```
lewijacn, line 183: What is the different use case of the handler being created here vs when we add as an initial handler in RequestPipelineOrchestrator?

peternied (commented):
Thanks for making this change, it isn't clear to me what users need to do be able to take advantage, could you include details as to what user facing changes are needed to process a gzip'd request with transformations?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestConvertHandler.java
```diff
@@ -38,6 +41,10 @@ public void channelRead(@NonNull ChannelHandlerContext ctx, @NonNull Object msg)
                 .log();
             var httpJsonMessage = parseHeadersIntoMessage(request);
             ctx.fireChannelRead(httpJsonMessage);
+            if (propagateOriginalRequest) {
+                // Send both, may need to re-drive the request through again depending on initial processing of httpJsonMessage
+                ctx.fireChannelRead(request);
```
lewijacn, line 46: What problem is being solved by also sending the original request through that has not been altered? Is there a later handler that will not work with the HttpJsonRequestWithFaultingPayload object?

lewijacn (commented):
Adding some comments for my understanding

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
AndreKurait: This is for tuple processing, we already have the full request and payload present so we don't need the double processing we do in the request flow (once with headers only, once with headers and payload)

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestConvertHandler.java
AndreKurait: The HttpJsonRequestWithFaultingPayload may have had modified headers from a partially processed transformation, we need the original request to re-run the transformation a second time with a fully parsed body

AndreKurait:
> Thanks for making this change, it isn't clear to me what users need to do be able to take advantage, could you include details as to what user facing changes are needed to process a gzip'd request with transformations?

Added a note in the overview. This is a bug-fix so users do not need to do anything special to take advantage. If their transformation attempts to access the payload, it will be decompressed and parsed out.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/StrictCaseInsensitiveHttpHeadersMap.java
AndreKurait: This is not used anymore, removed

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestConvertHandler.java
lewijacn: Ah okay that clears up some confusion for me here

lewijacn (approved):

peternied:
> > Thanks for making this change, it isn't clear to me what users need to do be able to take advantage, could you include details as to what user facing changes are needed to process a gzip'd request with transformations?
> 
> Added a note in the overview. This is a bug-fix so users do not need to do anything special to take advantage. If their transformation attempts to access the payload, it will be decompressed and parsed out.

Thanks for the details - that sounds great!

PR #1316, opened 2025-02-28
Add ES 2.x and 5.x clusters for source options in K8s

lewijacn:
### Description
This change adds pre-configured cluster environments files for creating source and target test clusters after experimenting with images and ES settings/configurations to have these be successfully created. It is expected that as integration tests are used for k8s these existing environments will also be utilized to spin up the necessary source and target clusters

This change also renames the `mockCustomerClusters` chart to `testClusters` after receiving feedback from the team which should also reduce our character count for naming around this

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2339

### Testing
Local K8s testing

### Check List


deployment/k8s/README.md
peternied, line 1: ❤️

deployment/k8s/README.md
```diff
@@ -74,11 +74,16 @@ By default, all of these components will be deployed, but if a user is certain t
...
 ```shell
 helm install ma -n ma charts/aggregates/migrationAssistant --create-namespace
-helm install mcc -n ma charts/aggregates/mockCustomerClusters
+helm install tc -n ma charts/aggregates/testClusters
+```
 
+Alternatively, specific test cluster configurations can be deployed instead of the default source and target configuration. Existing alternative configurations can be found in the `charts/aggregates/testClusters/environments` directory and deployed with helm commands similar to below
+```shell
+helm install tc-source -n ma charts/aggregates/testClusters -f charts/aggregates/testClusters/environments/es-5-6-single-node-cluster.yaml
+helm install tc-target -n ma charts/aggregates/testClusters -f charts/aggregates/testClusters/environments/os-2-latest-single-node-cluster.yaml
```
mikaylathompson, line 86: This is cool! Is there a way to specify a alternative source and target? (can you just chain the `-f YAML_PATH` syntax?)

deployment/k8s/charts/aggregates/testClusters/environments/es-2-4-single-node-cluster.yaml
```diff
@@ -0,0 +1,36 @@
...
+  sysctlInitContainer:
+    enabled: false
+  esConfig:
+    elasticsearch.yml: |
+      bootstrap.system_call_filter: false
+      xpack.security.enabled: false
+      network.host: 0.0.0.0
+  extraEnvs:
+    - name: "cluster.initial_master_nodes" # Remove setting unknown to ES 2.x
+      value: ""
```
mikaylathompson, line 25: by "remove", this means set to empty string? (basically, is this a todo comment or this is the implementation?)

deployment/k8s/charts/aggregates/testClusters/environments/es-2-4-single-node-cluster.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - name: "node.roles" # Remove setting unknown to ES 5.x
```
mikaylathompson, line 26: minor, update to `ES 2.X` or `ES 5.x and earlier`

deployment/k8s/charts/aggregates/testClusters/environments/es-5-6-single-node-cluster.yaml
```diff
@@ -0,0 +1,31 @@
+conditionalPackageInstalls:
+  proxy: false
+  source: true
+  target: false
+
+source:
```
mikaylathompson, line 6: I think the way this is set up is that this is _always_ a source cluster when it's included with the aggregate test cluster chart. Is that true? What kind of an error is thrown if you specify two source clusters? I'm fine with this being only a source, but we should be clear about it (and it's a bigger deal for those with overlap, like 7.10 or OS 1, which could be a source or a target--how are those handled?)

deployment/k8s/README.md
mikaylathompson: Yep, the first half is what I was wondering about but what are/why do you need: `--set conditionalPackageInstalls.source=true --set conditionalPackageInstalls.target=true`?

deployment/k8s/charts/aggregates/testClusters/environments/es-5-6-single-node-cluster.yaml
mikaylathompson: Not a blocker for now, but worth having a note/chart on what combos currently work, and maybe a JIRA task about what has to be done to accommodate it.

mikaylathompson (approved):

PR #1319, opened 2025-03-04
Update Sigv4 and ALB according to Opensearch logic

AndreKurait:
### Description
Opensearch Managed Service for Sigv4 auth requires the sigv4 signature to not have a port specified, this complicates adding the capture proxy between the service if the listener is on a non-443 port.

For the Migration Console, we will update our cluster sigv4 logic to exclude the port on the host prior to signing.
For the ALB, we will move the standard weighted listener to the 443 port instead of the current 9200 one.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2421

### Testing
Verified with the migration console that sigv4 works with a vpc cluster on port 9201 (the capture proxy direct listener)

### Check List


lewijacn (commented):
This seems like it should be fine. Separately, will our existing non-weighted 9201 and 9202 listeners work for SigV4?

AndreKurait:
> This seems like it should be fine. Separately, will our existing non-weighted 9201 and 9202 listeners work for SigV4?

it'll work with migration console, maybe not with other clients depending on how they are implemented

PR #1320, opened 2025-03-04
Add doc and diagram for lucene compatibility

mikaylathompson:
### Description
The off-by-one-ness of Lucene and Elasticsearch versioning is throwing me off on a daily basis, especially as we discuss which versions of Lucene are necessary to support snapshots from specific versions of Elasticsearch.

This pulls in and expands a chart that @peternied has on a branch: https://github.com/peternied/opensearch-migrations/blob/expand-reindexing/docs/expanding-reindexing.md#lucene-support-over-time

### Issues Resolved
n/a

### Testing
n/a

### Check List


docs/diagrams/Lucene_compatibility_chart.svg
peternied, line 1: I'm hesitant to document that Lucene 10 can read back 2 versions without knowing what is making that possible or having verified this behavior.

peternied:
Thanks for writing this down somewhere other than my branch, but maybe this means I should make a PR or RFC around that doc so it isn't lost :D

docs/diagrams/Lucene_compatibility_chart.svg
mikaylathompson: That's fair. I'm basing it off of the backwards compatibility codecs included [here](https://lucene.apache.org/core/9_9_0/backward-codecs/index.html) vs [here](https://lucene.apache.org/core/10_0_0/backward-codecs/index.html), but I haven't seen an authoritative source saying that they're intentionally supporting back to Lucene 8, and I haven't tested it myself.

peternied: According to those, we should be able to read lucene6 indices with lucene9 but that doesn't seem to be the case.  I'd rather remove the 'extra' hops if we don't need to represent them in our documentation and I'm open to what you are thinking.  How would you like to move forward?

peternied (approved):

docs/diagrams/Lucene_compatibility_chart.svg
mikaylathompson: I don't think that's a correct reading of the pages.

If you open up each of those high level items, they contain different constintuent classes, and for Lucene 10, the only class in the Lucene70Codec is [Lucene70SegmentInfoFormat](https://lucene.apache.org/core/10_0_0/backward-codecs/org/apache/lucene/backward_codecs/lucene70/package-summary.html), whereas for Lucene 9, the [Lucene70Codec](https://lucene.apache.org/core/9_9_0/backward-codecs/org/apache/lucene/backward_codecs/lucene70/package-summary.html) contains several other classes -- notably the [Lucene70Codec](https://lucene.apache.org/core/9_9_0/backward-codecs/org/apache/lucene/backward_codecs/lucene70/Lucene70Codec.html) class.

The classes contained in Lucene 10's `Lucene80Codec` match what's available for Lucene 9's `Lucene70Codec`. While I agree that it's untested, I think this is a reasonably strong indicator that the backwards compatibility extends two versions.

peternied:
@mikaylathompson Maybe tomorrow we should grab some time to walk through what you propose.  I've approved this PR, so its up to you if you'd like to wait for us to sync or merge as you prefer.

PR #1333, opened 2025-03-07
[Console Lib] Add SearchContainer and tests to catch source cluster behavior differences

mikaylathompson:
### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_source_cluster_multiversion.py
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_source_cluster_multiversion.py
@@ -1,0 +1,139 @@
+import itertools
+import logging
+import os
+import tempfile
+import json
+import time
+
+import yaml
+import pytest
+
+from console_link.environment import Environment
+from console_link.models.cluster import Cluster, HttpMethod
+from tests.search_containers import SearchContainer, Version, CLUSTER_SNAPSHOT_DIR
+import console_link.middleware.clusters as clusters_
+import console_link.middleware.snapshot as snapshot_
+
+SUPPORTED_SOURCE_CLUSTERS = [Version("ELASTICSEARCH", 5, 6, 16),
+                             Version("ELASTICSEARCH", 6, 8, 23),
+                             Version("ELASTICSEARCH", 7, 10, 2)]
+
+TEST_INDEX_NAME = "test_index"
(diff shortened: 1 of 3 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_source_cluster_multiversion.py
```diff
@@ -0,0 +1,139 @@
...
+    setup_snapshot_repo_and_snapshot(env.source_cluster)
+
+    yield env
+    # Stop the container and clean up the temporary services.yaml file after tests complete
+    container.stop()
+    os.remove(temp_config_path)
+
+
+@pytest.mark.parametrize("env_with_source_container,refresh,json",
+                         itertools.product(SUPPORTED_SOURCE_CLUSTERS, [True, False], [True, False]),
```
peternied, line 105: Rather that have a flag to refresh or not, can we minimize the parameterized values so that we reuse the same cluster to do that true case and the false case?  This will lower the runtime considerably for each of these pairs that we remove.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_source_cluster_multiversion.py
```diff
@@ -0,0 +1,139 @@
...
+    env = env_with_source_container
+    assert type(env.source_cluster) is Cluster
+    result = clusters_.connection_check(env.source_cluster)
+    assert result.connection_established
+
+
+@pytest.mark.parametrize("env_with_source_container,deep_status_check",
+                         itertools.product(SUPPORTED_SOURCE_CLUSTERS, [False, True]),
+                         indirect=["env_with_source_container"])
+def test_snapshot_status(env_with_source_container: Environment, deep_status_check: bool):
```
peternied, line 134: I've got beef with 'deep' vs 'normal' output, can change the behavoir to only have a singular output that we verify - or maybe only prioritize the deep scenario because will we fine a bug in the difference?

peternied (approved):
Really like this, have some optional comments

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/test_source_cluster_multiversion.py
mikaylathompson: Oh, yeah, basically run with refresh first, and then non-refresh in the same test case? 👍

PR #1343, opened 2025-03-11
Update most tests to use OS_LATEST image, currently 2.19.1

mikaylathompson:
### Description
The documentation, as of #1336, refers to migrating to 2.19. This PR updates our tests to use the 2.19.1 image.

Because it's kind of silly to have to go change our code in N places every time we are testing against a new version, I also added an `OS_LATEST` alias so that as we add a new image, we only have to add its image info and update the alias in one file, and we'll automatically test against it in most places.

I did _not_ change our `SupportedCluster` matrix to use `OS_LATEST` because I think that's a helpful spot to be very explicit about exactly what we test against. Open to feedback if others disagree.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2430

### Testing
Primarily a test-related change.

### Check List

### DIFF
--- a/RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
+++ b/RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
@@ -70,6 +70,7 @@ public class SearchClusterContainer extends GenericContainer<SearchClusterContai
             new ImmutableMap.Builder<String, String>().putAll(BASE.getEnvVariables())
                 .put("plugins.security.disabled", "true")
                 .put("OPENSEARCH_INITIAL_ADMIN_PASSWORD", "SecurityIsDisabled123$%^")
+                .put("search.insights.top_queries.exporter.type", "debug")
                 .build()
         );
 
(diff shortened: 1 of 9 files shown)


RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
```diff
@@ -68,6 +69,11 @@ private enum INITIALIZATION_FLAVOR {
                 .build()),
         OPENSEARCH(
             new ImmutableMap.Builder<String, String>().putAll(BASE.getEnvVariables())
+                .put("plugins.security.disabled", "true")
+                .put("OPENSEARCH_INITIAL_ADMIN_PASSWORD", "SecurityIsDisabled123$%^")
+                .build()),
+        OPENSEARCH_2_19(
+        new ImmutableMap.Builder<String, String>().putAll(BASE.getEnvVariables())
```
peternied, line 76: Can we add the new entry for OPENSEARCH rather than create a new category?

peternied (approved):

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
mikaylathompson: No, if it's present in earlier opensearch versions it prevents the cluster from initializing

PR #1351, opened 2025-03-13
Metadata migration support for ES2 indices

peternied:
### Description
This is a proof of concept for writing a JavaScript transformation that will update indices and index templates that are using the string type, see `docs/breaking-changes/StringMappingType.md` included in this PR for additional details.

MultiTypeMappingTransformationTest.java is augmented to use the new transformation.  Run this scenario yourself with the following command.  It will 'minify' [1] the javascript into the json definition, then start the test that will load use that configuration.

```bash
python MetadataMigration/src/test/resources/updatejson.py && ./gradlew MetadataMigration:isolatedTest --tests *MultiTypeMappingTransformationTest.migrateFromUpgrade`
```

### Issues
- Related [MIGRATIONS-2425 | Test lucene5 indices on an Elasticsearch 5 cluster](https://opensearch.atlassian.net/browse/MIGRATIONS-2425)

### Check List


MetadataMigration/src/test/resources/es2-transforms.json
peternied, line 1: By adding this for a test case, the test is unblocked, but if you are coming from ES2 you will get an error and then have to figure out where/how to get this fix.  I'm unclear on what the appropriate path is for this and other options.  @AndreKurait @gregschohn could you provide guidance on mechanically how we should do this?

#### Some potential ideas
- If we included this as a resource in the jars, then its practically as inaccessible as the java code :/
- If we update the console to have a default config file, it will be hard to read - Even if you aren't coming from ES2 we need this available.  Secondarily it means that the java projects will require this default config which is a new dependency relationship that will be a headache.
- New projects or new files in existing projects also seem like an option, but I'm not sure how we 'expose' this default out.

transformation/src/main/java/org/opensearch/migrations/transformation/rules/StringMappingType.java
peternied, line 1: For side by side comparison with JavaScript implementation.

MetadataMigration/src/test/resources/es2-transforms.json
```diff
@@ -0,0 +1,8 @@
...
+            "bindingsObject": "{}"
+        }
+    }
+]
```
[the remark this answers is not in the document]
mikaylathompson: Or would it be a follow up task to put a copy of this file in something like a `transformations/es2/stringToText.json` directory from which a user can add it to the params?

MetadataMigration/src/test/resources/es2-transforms.json
mikaylathompson, line 1: I'd love to have a test for this, but I'm okay waiting until we have @gregschohn's new JS-with-jest setup.

mikaylathompson (approved):

MetadataMigration/src/test/resources/es2-transforms.json
peternied: What do you mean, we do have a test - It will fail without this code.  But rather to your point, I'd rather we have a 'production ready' version of this code that is better reviewed and unit-tested.

MetadataMigration/src/test/resources/es2-transforms.json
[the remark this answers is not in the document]
peternied: Nope, its only available to the test cases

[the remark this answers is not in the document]
peternied: I think there is an ideal version which we wouldn't know the shape of until we have clarity from [MIGRATIONS-2435 | Define user experience for handling many breaking changes]( https://opensearch.atlassian.net/browse/MIGRATIONS-2435).

This is a technically functional tests aren't blocked version, which I think could be useful to build on if we get more requests. 

I don't think it is a simple as a parameter in a number for the `ignore_above` value, as there are string types that might be the more appropriate version, see [String field types](https://opensearch.org/docs/latest/field-types/supported-field-types/string/) which has 6 different options 🤯

MetadataMigration/src/test/resources/es2-transforms.json
mikaylathompson: Yes, good point! I think a narrower test that shows the before and after clearly would be helpful and clarifying, especially as we build up a library of these.

PR #1354, opened 2025-03-13
Add Integration Tests for E2E Multi-type transformations

lewijacn:
### Description
WIP

### Check List


peternied (commented):
Too a brief look, added some high level comments.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/multi_type_tests.py
```diff
@@ -91,9 +91,9 @@ def perform_operations_after_backfill_migration(self):
 
     def perform_operations_during_live_capture_migration(self):
         self.target_operations.get_document(cluster=self.target_cluster, index_name=self.index_name,
-                                            doc_id=self.doc_id3, max_attempts=10, delay=3.0)
+                                            doc_id=self.doc_id3, max_attempts=20, delay=3.0)
```
peternied, line 94: Why are `max_attempts` and `delay` on every request, shouldn't these be in within common thresholds?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/multi_type_tests.py
```diff
@@ -91,9 +91,9 @@ def perform_operations_after_backfill_migration(self):
...
         self.target_operations.get_document(cluster=self.target_cluster, index_name=self.index_name,
-                                            doc_id=self.doc_id4, max_attempts=10, delay=3.0)
+                                            doc_id=self.doc_id4, max_attempts=20, delay=3.0)
```
peternied, line 96: Why do we have a pattern of so many retries, shouldn't the workflow be deterministic enough that each step should reasonable complete on the first attempt?

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
```diff
@@ -32,7 +32,7 @@ shared-configs:
         basicAuthPassword: "myStrongPassword123!"
         # Should be in sync with above username/password
         basicAuthHeader: "Basic YWRtaW46bXlTdHJvbmdQYXNzd29yZDEyMyE="
-        version: "2.16.0"
+        version: "OS_2.16"
```
peternied, line 35: I'd prefer that these are more 'type-safe', can we switch to using an enum of something similar to prevent strange scenarios?

deployment/k8s/charts/components/migrationConsole/values.yaml
```diff
@@ -1,3 +1,4 @@
+developerModeEnabled: false
```
peternied, line 1: What do you think about not having this flag and we always deploy all the tests systems?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/ma_test_base.py
```diff
@@ -0,0 +1,108 @@
...
+        pass
+
+    def start_backfill_migration(self):
+        backfill_start_result: CommandResult = self.backfill.start()
+        assert backfill_start_result.success
+        # small enough to allow containers to be reused, big enough to test scaling out
+        backfill_scale_result: CommandResult = self.backfill.scale(units=2)
+        assert backfill_scale_result.success
+
+    def perform_operations_during_backfill_migration(self):
```
gregschohn, line 82: What does it mean to run 'during'?  Will you run continuously?  I suspect that this is really just guaranteed to run before the next phase is run and after the previous one has finished - which isn't quire the same as during since a backfill might not have started backfilling or it could have been done.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/multi_type_tests.py
gregschohn: This one is added AFTER backfill has finished, so it will take time for it to land, is that right?  Are those exponential backoffs? How many retries (and more importantly, lost seconds > optimal
) do you expect to happen, given the other parameters for this test?

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
```diff
@@ -9,17 +9,19 @@ conditionalPackageInstalls:
   replayer: true
   sharedConfigs: true
   snapshotVolume: true
+  logsVolume: true
```
gregschohn, line 12: I suspect that we'll want to use otel-collector to lift and ship logs to a final resting place using compression.  This is a good quick-and-dirty fix that we can eventually make into false for prod and leave it as true for devs.

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
```diff
@@ -9,17 +9,19 @@ conditionalPackageInstalls:
...
 
 
 shared-configs:
   globalParameters:
     sourceCluster:
       allowRuntimeOverride: true
       object:
-        endpoint: "http://elasticsearch-master:9200"
+        endpoint: "http://ma-capture-proxy:9200"
+        directEndpoint: "http://elasticsearch-master:9200"
```
gregschohn, line 21: why do we have this split now?

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
```diff
@@ -103,8 +109,18 @@ bulk-document-loader:
       source: otherConfig
       configMapName: "target-cluster"
       configMapKey: "basicAuthPassword"
+    sourceVersion:
+      source: parameterConfig
+      value: "ES_7.10"
+      allowRuntimeOverride: true
+    docTransformerConfigFile:
+      source: parameterConfig
+      value: "/shared-logs-output/test-transformations/transformation.json"
```
gregschohn, line 118: See https://opensearch.atlassian.net/browse/MIGRATIONS-2401 for a better way to send configs/arguments into processes.

deployment/k8s/charts/components/bulkLoad/values.yaml
```diff
@@ -1,3 +1,5 @@
+sharedLogsVolumeEnabled: false
+sharedLogsPvc: ""
```
gregschohn, line 2: hmmm - this isn't very obvious (to clear the pvc) - what happens if this isn't done?

deployment/k8s/charts/components/migrationConsole/values.yaml
gregschohn: See this section above
```
{{- if $developerModeEnabled  }}
        - name: local-console-libs
          hostPath:
            path: /opensearch-migrations/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib
            type: Directory
```

That mounts the LOCAL HOST directory from the developer's machine into the running container.

deployment/k8s/minikubeLocal.sh
```diff
@@ -7,18 +7,28 @@ usage() {
...
 start() {
   helm repo add opensearch-operator https://opensearch-project.github.io/opensearch-k8s-operator/
   helm repo add strimzi https://strimzi.io/charts/
 
   minikube start
+  minikube mount .:/opensearch-migrations > /dev/null 2>&1 &
 }
 
 pause() {
+  kill_minikube_processes
```
gregschohn, line 26: When do you typically do this in your workflows?  I usually leave mine running and don't think much about it.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/conftest.py
```diff
@@ -2,18 +2,32 @@
...
 
-def pytest_configure(config):
-    # Configure logging, avoid sensitive data at lower logs levels
-    logging.basicConfig(level=logging.INFO,
-                        format='%(asctime)s - %(levelname)s - %(message)s',
-                        datefmt='%Y-%m-%d %H:%M:%S')
+logger = logging.getLogger(__name__)
+
+# Dynamically collect all test case classes that have been imported above
+ALL_TEST_CASES = [cls for cls in globals().values() if isinstance(cls, type) and
```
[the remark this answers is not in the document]
lewijacn: Yes let me add some details around this

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/default_operations.py
```diff
@@ -0,0 +1,185 @@
...
+            if index_prefix_ignore_list is None or valid_index:
+                # "To get an accurate count of Elasticsearch documents, use the cat count or count APIs."
+                # See https://www.elastic.co/guide/en/elasticsearch/reference/7.10/cat-indices.html
+
+                count_response = execute_api_call(cluster=cluster, path=f"/{index_name}/_count?format=json", **kwargs)
+                index_dict[index_name] = count_response.json()
+                index_dict[index_name]['index'] = index_name
+        return index_dict
+
+    def check_doc_counts_match(self, cluster: Cluster,
```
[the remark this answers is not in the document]
lewijacn: Yes agreed, this was copied from the existing integ test `common_operations.py` file that got removed.  Once we remove the existing integration tests I'd like to remove passing the test case to any of these functions and allow them to just throw an error if a check fails.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/ma_workflow_test.py
```diff
@@ -0,0 +1,88 @@
...
+    target_cluster = test_case.target_cluster
+    source_con_result: ConnectionResult = connection_check(source_cluster)
+    assert source_con_result.connection_established is True
+    target_con_result: ConnectionResult = connection_check(target_cluster)
+    assert target_con_result.connection_established is True
+    kafka: Kafka = test_case.console_link_env.kafka
+
+    # Cleanup generated transformation files
+    try:
+        shutil.rmtree("/shared-logs-output/test-transformations")
```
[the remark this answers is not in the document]
lewijacn: It's probably a small effort, especially if the teardown function has some way of determining success or failure. It seems like we really want a lot of things to get generated somewhere persistent, such as service logs, cluster logs, test_output, etc. Currently items aren't cleaned up until another test case is executed.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/multi_type_tests.py
peternied: Until we do, can we have a default pre-backfill complete action that monitors for the backfill to be done, or better yet something like that in the console, `console backfill wait-until-complete` or whatever the cleanest pattern is?

deployment/k8s/charts/aggregates/migrationAssistant/values.yaml
peternied: While that makes sense it creates fragility in our system where it is non-obvious why something isn't working if these values are slightly off.  I hope that doesn't often occur.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/multi_type_tests.py
lewijacn: I created this task to track work for reporting a COMPLETED status for backfill, which we can then easily check in integration tests to wait for a completed state: https://opensearch.atlassian.net/browse/MIGRATIONS-2449

PR #1358, opened 2025-03-17
Support OpenSearch 2.x as a source

### LINKED ISSUE #1038: Upgrading from OpenSearch 2
### Is your feature request related to a problem?
Amazon OpenSearch Service uses the blue/green deployment method by default for version upgrades. However, in some cases, you may need to keep the existing cluster and roll back if necessary. To support this, Amazon OpenSearch Service 2.x needs to support upgrading to a higher 2.x version as the target.

### What solution would you like?
Supporting Amazon OpenSearch Service 2.x as the source cluster

### What alternatives have you considered?
N/A

This is not a migration from open source to a managed service, but rather a feature that supports migration between managed services for upgrading

mikaylathompson:
### Description
We have increasing customer interest in using OpenSearch 2.X as a source and a target (e.g. keeping the version the same and moving from one cluster to another). This PR unblocks that path, which was primarily blocked by 1/ a few version checks on our end, 2/ a not-quite-up-to-date-enough version of Lucene (for OS 2.19).

This PR adds OS2->OS2 to our Document Bulkload and Metadata Migration end-to-end tests.

### Issues Resolved
#1038 

### Testing
As mentioned above, end-to-end tests for documents and metadata.

### Check List


MetadataMigration/src/test/java/org/opensearch/migrations/EndToEndTest.java
```diff
@@ -37,15 +37,15 @@
 class EndToEndTest extends BaseMigrationTest {
 
     private static Stream<Arguments> scenarios() {
```
peternied, line 39: This is pretty messy, can we refactor so `SupportedClusters` returns a definitive list of source and targets?  Then this function only adds test specific properties?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -42,6 +42,7 @@ private static Stream<Arguments> scenarios() {
                 scenarios.add(Arguments.of(sourceCluster, targetCluster));
             }
         }
+        scenarios.add(Arguments.of(SearchClusterContainer.OS_LATEST, SearchClusterContainer.OS_LATEST));
```
peternied, line 45: Lets hardcode these values to the `OS_V2_19_1` version, unless that alias is really refers to the latest 2.x version, in which case lets adjust the naming.

peternied (approved):

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
```diff
@@ -114,7 +114,9 @@ public SearchClusterContainer(final ContainerVersion version) {
     public void copySnapshotData(final String directory) {
         try {
             // Execute command to list all files in the directory
-            final var result = this.execInContainer("sh", "-c", "find " + CLUSTER_SNAPSHOT_DIR + " -type f");
+            // `find` is not available on some versions of the containers, in which case it falls back to a ls/grep loop.
+            final var result = this.execInContainer("sh", "-c", "find " + CLUSTER_SNAPSHOT_DIR + " -type f" + " || " +
```
peternied, line 118: Can you open an issue in https://github.com/opensearch-project/opensearch-build this seems like an unintentional regression

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
mikaylathompson: Good point, it will likely eventually be a 3.x verison

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
mikaylathompson: I assume it's because they're trying to minimize the container size and therefore is an intentionally exclusion. So far as I know, it's all 2.x images, not just 2.19.1. Do you have additional reason to believe it's unintentional?

peternied: I mean that seems like a reasonable reason for them to push back, but `find` seems pretty standard IMO.  I say unintentional because they might not realize the impact.  If grep went away we'd also have trouble even if it was for the same reason.

PR #1372, opened 2025-03-20
[Sonarqube] java:S3776 Refactor NettyJsonBodySerializeHandler to reduce complexity

peternied:
### Description
java:S3776 Refactor NettyJsonBodySerializeHandler to reduce complexity

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
@@ -26,18 +26,22 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter 
         if (!(msg instanceof HttpJsonRequestWithFaultingPayload)) {
             super.channelRead(ctx, msg);
         } else {
-            ctx.fireChannelRead(msg);
-
+            // Copy the payload before clearing it from the message
             var jsonMessage = (HttpJsonRequestWithFaultingPayload) msg;
             var payload = jsonMessage.payload();
             jsonMessage.setPayloadFaultMap(null);
 
-            processInlineDocuments(ctx, payload);
+            var messageWithoutPayload = jsonMessage;
+            ctx.fireChannelRead(messageWithoutPayload);
+
+            parsePayloadAndAttachToChannel(ctx, payload);
+            
+            ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
         }
     }
 
     @SuppressWarnings("unchecked")
-    private void processInlineDocuments(ChannelHandlerContext ctx, Map<String, Object> payload) throws Exception {
+    private void parsePayloadAndAttachToChannel(ChannelHandlerContext ctx, Map<String, Object> payload) throws Exception {
         if (payload.containsKey(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY)) {
             serializePayload(ctx, payload.get(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY));
         } else if (payload.containsKey(JsonKeysForHttpMessage.INLINED_NDJSON_BODIES_DOCUMENT_KEY)) {
@@ -57,7 +61,6 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter 
             body.writeCharSequence(bodyString, StandardCharsets.UTF_8);
             fireHttpContentOrRelease(ctx, body);
         }
-        ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
     }
 
     private void fireHttpContentOrRelease(ChannelHandlerContext ctx, ByteBuf buf) {
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
@@ -25,14 +25,15 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter 
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
         if (!(msg instanceof HttpJsonRequestWithFaultingPayload)) {
             super.channelRead(ctx, msg);
+        } else {
+            ctx.fireChannelRead(msg);
+
+            var jsonMessage = (HttpJsonRequestWithFaultingPayload) msg;
+            var payload = jsonMessage.payload();
+            jsonMessage.setPayloadFaultMap(null);
+
+            processInlineDocuments(ctx, payload);
         }
-        ctx.fireChannelRead(msg);
-
-        var jsonMessage = (HttpJsonRequestWithFaultingPayload) msg;
-        var payload = jsonMessage.payload();
-        jsonMessage.setPayloadFaultMap(null);
-
-        processInlineDocuments(ctx, payload);
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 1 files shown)


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
```diff
@@ -22,40 +23,46 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
...
+    private void processInlineDocuments(ChannelHandlerContext ctx, Map<String, Object> payload) throws Exception {
+        if (payload.containsKey(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY)) {
+            serializePayload(ctx, payload.get(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY));
+        } else if (payload.containsKey(JsonKeysForHttpMessage.INLINED_NDJSON_BODIES_DOCUMENT_KEY)) {
+            var shouldAddLastNewline = !payload.containsKey(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY) &&
+                !payload.containsKey(JsonKeysForHttpMessage.INLINED_TEXT_BODY_DOCUMENT_KEY);
+            serializePayloadList(ctx,
+                (List<Object>) payload.get(JsonKeysForHttpMessage.INLINED_NDJSON_BODIES_DOCUMENT_KEY),
+                shouldAddLastNewline);
+        } else if (payload.containsKey(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY)) {
```
peternied, line 48: Could I get a double check here? Previously the organization of these if statements was 2 separate groups with json/ndjson checked and then a separate checks for binary/text.  That doesn't seem correct, it seems like a document should only include one of json/ndjson/binary/text.

See @38 on the original file for this behavoir

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
```diff
@@ -22,40 +23,46 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
...
-                    ctx.fireChannelRead(new DefaultHttpContent(body));
-                } else {
-                    ReferenceCountUtil.release(body);
-                }
-            }
-            ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
-        } else {
+        if (!(msg instanceof HttpJsonRequestWithFaultingPayload)) {
             super.channelRead(ctx, msg);
         }
```
AndreKurait, line 28: Did you mean to include an else here?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
```diff
@@ -22,40 +23,46 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
...
+        ctx.fireChannelRead(msg);
```
AndreKurait, line 29: The existing code performed this action after jsonMessage.setPayloadFaultMap(null), but now it's happening before

gregschohn (changes requested):
I'm not sure that this class is worth the time given the amount of complexity within any parsing logic.  The following patch gets sonar to stop complaining.
```
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
@@ -22,11 +22,17 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
 
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
-        if (msg instanceof HttpJsonRequestWithFaultingPayload) {
-            var jsonMessage = (HttpJsonRequestWithFaultingPayload) msg;
+        if (!(msg instanceof HttpJsonRequestWithFaultingPayload)) {
+            super.channelRead(ctx, msg);
+        } else {
+            parseMessage(ctx, (HttpJsonRequestWithFaultingPayload) msg);
+        }
+    }
+
+    private void parseMessage(ChannelHandlerContext ctx, HttpJsonRequestWithFaultingPayload jsonMessage) throws Exception {
             var payload = jsonMessage.payload();
             jsonMessage.setPayloadFaultMap(null);
-            ctx.fireChannelRead(msg);
+            ctx.fireChannelRead(jsonMessage);
             if (payload.containsKey(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY)) {
                 serializePayload(ctx, payload.get(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY));
             } else if (payload.containsKey(JsonKeysForHttpMessage.INLINED_NDJSON_BODIES_DOCUMENT_KEY)) {
@@ -53,9 +59,6 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
                 }
             }
             ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
-        } else {
-            super.channelRead(ctx, msg);
-        }
     }
 ```

File: TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java (hunk 1 above)
Comment (peternied) [reply TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java]: Thanks! I've corrected this, just pushed an update.

File: TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java [hunk 4]
```diff
@@ -22,39 +23,48 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
...
+        if (payload.containsKey(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY)) {
+            var rawBody = (ByteBuf) payload.get(JsonKeysForHttpMessage.INLINED_BINARY_BODY_DOCUMENT_KEY);
+            fireHttpContentOrRelease(ctx, rawBody);
+        } else if (payload.containsKey(JsonKeysForHttpMessage.INLINED_TEXT_BODY_DOCUMENT_KEY)) {
+            var bodyString = (String) payload.get(JsonKeysForHttpMessage.INLINED_TEXT_BODY_DOCUMENT_KEY);
+            ByteBuf body = ctx.alloc().buffer();
+            body.writeCharSequence(bodyString, StandardCharsets.UTF_8);
+            fireHttpContentOrRelease(ctx, body);
+        }
+        ctx.fireChannelRead(LastHttpContent.EMPTY_LAST_CONTENT);
```
AndreKurait, line 60: can we move this into the top level function

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodySerializeHandler.java
```diff
@@ -22,39 +23,48 @@ public class NettyJsonBodySerializeHandler extends ChannelInboundHandlerAdapter
 
     @Override
     public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
-        if (msg instanceof HttpJsonRequestWithFaultingPayload) {
+        if (!(msg instanceof HttpJsonRequestWithFaultingPayload)) {
+            super.channelRead(ctx, msg);
+        } else {
+            ctx.fireChannelRead(msg);
```
AndreKurait, line 29: This should happen after `jsonMessage.setPayloadFaultMap(null);` since we're intending to send it through without the payload (since the payload is transformed in this handler)

peternied: Thanks for digging in on this! That wasn't really clear with the previous implementation

peternied:
@gregschohn I hear your concerns about investing time in this code.  As I was unfamiliar there were many nuances to how this function worked that were not obvious, looking at the change history it has been edited 27 times.  Can we agree this code is in an improvement state and is worthwhile to merge?

AndreKurait (approved):
Change looks like it retains all functionality and reduces duplication with fireHttpContentOrRelease

PR #1391, opened 2025-03-26
[CVE] Updated forced versions for transitive dependencies

peternied:
### Description
[CVE] Updated forced versions for transitive dependencies
   Also sorted all items by name

### Check List


build.gradle
```diff
@@ -111,20 +111,32 @@ subprojects {
...
+
     configurations.all {
         resolutionStrategy {
             // Force specific versions of transitive dependencies to handle CVEs
-            force 'net.minidev:json-smart:2.5.2'
-            force 'xalan:xalan:2.7.3'
+            force "com.fasterxml.jackson.core:jackson-databind:${libs.versions.jackson.get()}"
+            force "com.google.guava:guava:${libs.versions.guava.get()}"
+            force "commons-io:commons-io:2.14.0"
+            force "io.netty:netty-codec:${libs.versions.netty.get()}"
```
mikaylathompson, line 122: it looks like `nettyVersion` is already a defined variable -- is there a reason why we repeat the call from line 114 for each of these?

peternied: Oh thanks, removing

peternied:
Verified locally that the latest commit works on during failing test cases, 🤞

AndreKurait (approved):
This change makes sense to fix our current CVEs, for the future, i'd like to look into a more holistic strategy for aligning direct and transitive dependencies across the stack even when differences don't result in a CVE

PR #1394, opened 2025-03-27
Automatically update replica count for az compatibility

mikaylathompson:
### Description
This updates how the metadata migration tool sets replica counts in cases where the cluster in question (usually, but not necessarily in the AWS managed service) has enforced balancing across awareness attributes. Usually, awareness attributes means availability zones, though it could be other factors. See this page for more information on how it works in opensearch: https://opensearch.org/docs/latest/tuning-your-cluster/#advanced-step-6-configure-shard-allocation-awareness-or-forced-awareness

In this PR, we take an approach to updating replica counts where we begin with the value from the metadata command arguments. If we see an incompatible replica count error, we increment the presumed dimensionality used by the transformer (which sets the minimum replica count for each index to dimensionality - 1) and try migrating the indices again. We repeat this up to 4 times, if necessary, but exit as soon as we no longer see an incompatible replica count error.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2185

### Check List

### DIFF
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
@@ -2,7 +2,6 @@ package org.opensearch.migrations.commands;
 
 import org.opensearch.migrations.MigrateOrEvaluateArgs;
 import org.opensearch.migrations.MigrationMode;
-import org.opensearch.migrations.bulkload.common.IncompatibleReplicaCountException;
 import org.opensearch.migrations.metadata.tracing.RootMetadataMigrationContext;
 
 import com.beust.jcommander.ParameterException;
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
@@ -2,7 +2,7 @@ package org.opensearch.migrations.commands;
 
 import org.opensearch.migrations.MigrateOrEvaluateArgs;
 import org.opensearch.migrations.MigrationMode;
-import org.opensearch.migrations.bulkload.version_universal.IncompatibleReplicaCountException;
+import org.opensearch.migrations.bulkload.common.IncompatibleReplicaCountException;
 import org.opensearch.migrations.metadata.tracing.RootMetadataMigrationContext;
 
 import com.beust.jcommander.ParameterException;
@@ -35,10 +35,6 @@ public class Migrate extends MigratorEvaluatorBase {
                     .exitCode(INVALID_PARAMETER_CODE)
                     .errorMessage("Invalid parameter: " + pe.getMessage())
                     .build();
-        } catch (IncompatibleReplicaCountException e) {
-            arguments.minNumberOfReplicas++;
-            log.atError().setMessage("Incompatible replica count detected, increasing min replica count to {}").addArgument(arguments.minNumberOfReplicas).log();
-            return execute(context);
         } catch (Throwable e) {
             log.atError().setCause(e).setMessage("Unexpected failure").log();
             migrateResult
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
@@ -2,6 +2,7 @@ package org.opensearch.migrations.commands;
 
 import org.opensearch.migrations.MigrateOrEvaluateArgs;
 import org.opensearch.migrations.MigrationMode;
+import org.opensearch.migrations.bulkload.version_universal.IncompatibleReplicaCountException;
 import org.opensearch.migrations.metadata.tracing.RootMetadataMigrationContext;
 
 import com.beust.jcommander.ParameterException;
@@ -31,9 +32,13 @@ public class Migrate extends MigratorEvaluatorBase {
         } catch (ParameterException pe) {
             log.atError().setCause(pe).setMessage("Invalid parameter").log();
             migrateResult
-                .exitCode(INVALID_PARAMETER_CODE)
-                .errorMessage("Invalid parameter: " + pe.getMessage())
-                .build();
+                    .exitCode(INVALID_PARAMETER_CODE)
+                    .errorMessage("Invalid parameter: " + pe.getMessage())
+                    .build();
+        } catch (IncompatibleReplicaCountException e) {
+            arguments.minNumberOfReplicas++;
+            log.atError().setMessage("Incompatible replica count detected, increasing min replica count to {}").addArgument(arguments.minNumberOfReplicas).log();
+            return execute(context);
         } catch (Throwable e) {
             log.atError().setCause(e).setMessage("Unexpected failure").log();
             migrateResult
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
@@ -9,6 +9,7 @@ import org.opensearch.migrations.bulkload.transformers.FanOutCompositeTransforme
 import org.opensearch.migrations.bulkload.transformers.TransformFunctions;
 import org.opensearch.migrations.bulkload.transformers.Transformer;
 import org.opensearch.migrations.bulkload.transformers.TransformerToIJsonTransformerAdapter;
+import org.opensearch.migrations.bulkload.version_universal.IncompatibleReplicaCountException;
 import org.opensearch.migrations.bulkload.worker.IndexMetadataResults;
 import org.opensearch.migrations.bulkload.worker.IndexRunner;
 import org.opensearch.migrations.bulkload.worker.MetadataRunner;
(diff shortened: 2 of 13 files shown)


MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
```diff
@@ -31,9 +31,9 @@ public MigrateResult execute(RootMetadataMigrationContext context) {
         } catch (ParameterException pe) {
             log.atError().setCause(pe).setMessage("Invalid parameter").log();
             migrateResult
-                .exitCode(INVALID_PARAMETER_CODE)
-                .errorMessage("Invalid parameter: " + pe.getMessage())
-                .build();
+                    .exitCode(INVALID_PARAMETER_CODE)
```
mikaylathompson, line 34: Ugh, this is my IDE being out of sync with the usual settings. I'll go through and clean up the indents.

MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
```diff
@@ -118,16 +124,33 @@ private GlobalMetadataCreatorResults migrateGlobalMetadata(MigrationMode mode, C
...
+                    clusters.getSource().getIndexMetadata(),
+                    clusters.getTarget().getIndexCreator(),
+                    transformer,
+                    arguments.dataFilterArgs.indexAllowlist
+            );
+            var indexResults = indexRunner.migrateIndices(mode, context.createIndexContext());
+            // Check whether any indices failed with an incompatibleReplicaCount
+            boolean incompatibleReplicaCountSeen = indexResults.getIndexes().stream().anyMatch(result -> result.wasFatal() && result.getFailureType().equals(CreationResult.CreationFailureType.INCOMPATIBLE_REPLICA_COUNT_FAILURE));
+            if (incompatibleReplicaCountSeen) {
+                if (presumedClusterDimensionality >= arguments.minNumberOfReplicas + MAX_REPLICA_ADJUSTMENT_LOOPS) {
```
AndreKurait, line 142: Given cluster.routing.allocation.awareness.force.zone.values can be arbritraily high, can we parse out the response and skip straight to the  next-higher valid value?

RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
```diff
@@ -85,4 +86,36 @@ private static Map.Entry<String, String> getUnknownSetting(JsonNode json) {
             return Map.entry(entry.getKey().asText(), matcher.group(1));
         }).orElse(null);
     }
+
+    /** Awareness attribute exceptions (when the replica count doesn't match the number of zones) present slightly differently
+     in different versions. The message is the same (`"Validation Failed: 1: expected total copies needs to be a multiple of total awareness attributes [3];"`,
+     for instance), but the type is different (either `validation_exception` or `illegal_argument_exception`). For this reason,
+     we're matching based on a regex against the message instead of also checking the type.
+    **/
+    public Optional<String> containsAwarenessAttributeException() {
```
peternied, line 95: Does the message need to be parsed and returned by this function, that would aid the user if we displayed it to them?  If not, this function could be reduced to a return a `boolean` value.

RFS/src/main/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11.java
```diff
@@ -72,6 +76,13 @@ private void createInner(IndexMetadata index,
                 result.failureType(CreationFailureType.ALREADY_EXISTS);
             }
         } catch (InvalidResponse invalidResponse) {
+            var potentialAwarenessAttributeException = invalidResponse.containsAwarenessAttributeException();
```
peternied, line 79: Naming is a little messy, not sure if this was caught in a bulk rename.

```suggestion
            var hasInvalidReplicaCountForZoneSize = invalidResponse.containsAwarenessAttributeException();
```

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
```diff
@@ -0,0 +1,78 @@
...
+    private static final SearchClusterContainer.ContainerVersion SOURCE_VERSION = SearchClusterContainer.ES_V7_10_2;
+    private static final SearchClusterContainer.ContainerVersion TARGET_VERSION = SearchClusterContainer.OS_LATEST;
+
+    private static Stream<Arguments> scenarios() {
+        return Stream.of(
+                Arguments.of(1),
+                Arguments.of(2),
+                Arguments.of(3)
+        );
+    }
```
peternied, line 32: I don't think there is a test coverage benefit to running this more than once, seems like one test that detect this scenario would be good enough?

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
```diff
@@ -0,0 +1,78 @@
...
+            arguments.targetArgs.host = targetCluster.getUrl();
+
+            MigrationItemResult result = executeMigration(arguments, MetadataCommands.MIGRATE);
+            assertThat(result.getExitCode(), equalTo(0));
+            // Ensure that the same number of indices were created (successfully) on the target cluster
+            verifyIndexesExistOnTargetCluster(indexNames);
+        }
+    }
+
+    void verifyIndexesExistOnTargetCluster(List<String> indexNames) {
```
peternied, line 72: Doesn't look like the new INCOMPATIBLE_REPLICA_COUNT_FAILURE is encountered during this test, is that scenario possible?

RFS/src/main/java/org/opensearch/migrations/metadata/CreationResult.java
```diff
@@ -30,6 +30,7 @@ public enum CreationFailureType {
         ALREADY_EXISTS(false, "already exists"),
         UNABLE_TO_TRANSFORM_FAILURE(true, "failed to transform to the target version"),
         TARGET_CLUSTER_FAILURE(true, "failed on target cluster"),
+        INCOMPATIBLE_REPLICA_COUNT_FAILURE(true, "failed due to incompatible replica count for awareness attribute count"),
```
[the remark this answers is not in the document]
peternied: Revisiting after finishing the review, maybe this doesn't get sent to users?

MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
mikaylathompson: Yes, we definitely could. I went with the approach that we defined in the ticket refinement, but this would also be possible.

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.params.provider.MethodSource;
+
+import static org.hamcrest.CoreMatchers.equalTo;
+import static org.hamcrest.MatcherAssert.assertThat;
+
+@Tag("isolatedTest")
+public class ReplicaCountWithAZsTest extends BaseMigrationTest{
+
+    private static final SearchClusterContainer.ContainerVersion SOURCE_VERSION = SearchClusterContainer.ES_V7_10_2;
+    private static final SearchClusterContainer.ContainerVersion TARGET_VERSION = SearchClusterContainer.OS_LATEST;
```
peternied, line 24: Its strange that this is unused, maybe the zoned search cluster is a static class for only this test case

peternied (commented):
Thanks for revisiting this, couple of thought

MetadataMigration/src/main/java/org/opensearch/migrations/commands/Migrate.java
peternied: I had a PR once upon a time for an intellij formatter [1], there is a VS code in the project too.

- https://github.com/opensearch-project/opensearch-migrations/pull/846

RFS/src/main/java/org/opensearch/migrations/metadata/CreationResult.java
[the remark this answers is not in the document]
mikaylathompson: It will only get sent to users if we can't find a working replica count after 4 increments. By the `CLI value to add` -- what do you mean? The right replica count to set?

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
mikaylathompson: Not for the number of zones + max number of loops we're testing here. I could set up a test with 5+ zones that would cause it to be emitted, but I didn't because 1/ that's going to be a very slow test, and 2/ while a possible situation for a self-managed service cluster, I don't think it's possible in AOS (and unlikely no matter what)

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
mikaylathompson: I think the reason to do it is primarily that these match the scenarios customers may have and the behavior--at least for 1 AZ vs more-than-one is different. I'm open to removing the 2 AZ case.

peternied: 1 AZ doesn't seem like an interesting case, because any number will match, no?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11.java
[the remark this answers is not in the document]
peternied: That seems like a bug in OpenSearch, is there one open for it?

PR #1411, opened 2025-04-02
Experimental: Migration as a workflow

gregschohn:
### Description
Model a migration as an argo workflow.

There are some changes to the existing java programs to support running different workflows in parallel and in our minikube setup.  Notably...
1) RFS now accepts a --sessionName argument that it will append to ".migrations_working_state" for the index name within the target cluster.  That allows multiple RFS instances fleets to work on different indices in parallel.  The console python code has been updated at the lowest (library) levels to have that understanding.
2) RFS now accepts an --s3Endpoint argument so that we can send use localstack as the S3 provider.  Notice that the endpoint will likely need to be specified as IP address when running with this PR because working out the nested domain names for the deployment was a non-trivial amount of work.
3) The elasticsearchWithSearchGuard's image (which we use only for test, the proxy now builds off of an amazon linux base image) now has a keystore initialized.  That keystore now includes credentials ("test") for the S3 plugin to use when connecting to S3.  If we'd want to use this image to write values to a real S3 bucket, we'd need to override these values, probably via the configuration for the S3 plugin.
3) The capture proxy and the replayer now accepts a topic name

Other changes present include
1) Migration console now installs etcd (from github) to get the etcdctl.
2) Argo workflows templates are copied onto the migration console.  Notice that there's not yet an automated way to install those into the cluster though I plan to get to that with a job that pulls the templates from the migration console.
3) Included in the argo templates are templates for the migrationConsole.  The templates take configurations and produce yaml configurations that can be used by the console.  This is used repeatedly by workflows to kickoff snapshots and also to give the user a window into a specific step.  E.g. ere's the console that's configured with the kafka cluster & topic being used by this replayer, and the config also includes source & target so that all of the observability tools are present (cat-indices, describe-topic, etc)
4) OPA Gatekeeper is present, but I'm not planning to finish its configuration until we get closer to a security review.  My plan is to use it to allow us to only install workflows from a specific repository.
5) Otel collector is configured as a daemonset by default.  There's support for the opentelemetry-operator, but I was having a lot of issues with CRD teardown and management when things went wrong with the cluster.  For our current needs, the simple daemonset seems like a simpler approach than the operator.
6) I have brought in a prometheus operator, which also deploys a daemonset (and grafana, which we'll disable for production).  I don't see the metrics getting from the otel-collector to prometheus yet, but I'll work on that.

Onto why those changes were present and how to play with them!

After helm installing the migration chart, one can test a migration `kc create -f createFullMigrationWorkflow.yaml` (from `TrafficCapture/dockerSolution/src/main/docker/migrationConsole`).  This migration will automatically create a snapshot and once the snapshot is complete, run RFS until complete before moving onto the Replayer step.

Currently snapshotting and RFS are fully-active, using the localstack or S3 configuration passed into helm.  Metadata isn't being performed and the replayer step is a dummy one.  There's a separate workflow to spinup Kafka, the proxy, and the replayer.  There's no checks yet for when the replayer is done - that'll probably be work much further out.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2454
https://opensearch.atlassian.net/browse/MIGRATIONS-2455

### Testing
```
./deployment/cdk/opensearch-service-migration/buildDockerImages.sh
pushd deployment/k8s/charts/aggregates
helm install -n ma ma migrationAssistant
kubectl -n ma exec --stdin --tty  ma-migration-console-0 -- /bin/bash
kubectl create -f workflows/createFullMigrationWorkflow.yaml
```

localTesting.sh has some commands to run to get things setup.  You can run  open up the argo UI (`kubectl port-forward svc/argo-server 2746:2746` ... `http://localhost:2746`) and navigate to the workflow that was just deployed.  `kubectl get <WORKFLOW_ID>` will also show progress.

### Check List


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/middleware/clusters.py
```diff
@@ -58,7 +58,7 @@ def run_test_benchmarks(cluster: Cluster):
 
 # As a default we exclude system indices and searchguard indices
 def clear_indices(cluster: Cluster):
-    clear_indices_path = "/*,-.*,-searchguard*,-sg7*,.migrations_working_state"
+    clear_indices_path = "/*,-.*,-searchguard*,-sg7*,.migrations_working_state*"
```
peternied, line 61: Nit: What happens if an RFS worker is running, will it die with an understandable explanation?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/README.md
```diff
@@ -0,0 +1,246 @@
...
+The start capture workflow is worth describing because it's an example of a simple model that helps to facilitate reasoning.  In this case "all proxies are healthy" sounds like an easy and obvious step.  However, how many proxies have been deployed?  How does the rest of the system know the global health?  We don't have that global state today, but with workflow modeling, we can zoom in on steps, let them express more granular conditions, and test them for correctness! 
+
+```mermaid
+flowchart
+    initialize_kafka_cluster --> initialize_kafka_topic --> signal_proxy_to_write
+    rollout_capture_proxy_across_fleet --> all_proxies_are_healthy_and_not_capturing --> signal_proxy_to_write
+    signal_proxy_to_write --> all_proxies_are_capturing
+```
+
+## Example User-Driven Workflow (including A/B testing)
```
peternied, line 48: I'm lost past this point, maybe trim this down to what is going in right now or I'm missing out on how the workflow management systems operates.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/createFullMigrationWorkflow.yaml
peternied, line 1: Not specific to this file but it looks like the 'deployment' part of the infrastructure is leaking out into the codebase, could you help me understand how we know when/where we augment functionality in the project?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/createFullMigrationWorkflow.yaml
```diff
@@ -0,0 +1,66 @@
...
+                      "basic_auth": {
+                          "username": "admin",
+                          "password": "admin"
+                      }
+                  },
+                  "snapshot-and-migration-configs": [
+                      {
+                          "indices": [
+                              "index_3",
+                              "index_4"
```
peternied, line 27: Hard coding two specific indices into this workflow, is this an example?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/createProcessSnapshotWorkflow.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: argoproj.io/v1alpha1
```
peternied, line 1: Note, really don't love `yaml` files getting scattered throughout the codebase, is there another extension we can use to differentiate argo vs k8 vs helm vs other declarative files?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/captureReplay.yaml
```diff
@@ -0,0 +1,196 @@
...
+            valueFrom:
+              expression: "inputs.parameters['provided-kafka-k8s-name'] == '' ? inputs.parameters['created-kafka-name'] : inputs.parameters['provided-kafka-k8s-name']"
+          - name: bootstrap-servers
+            valueFrom:
+              expression: "inputs.parameters['provided-kafka-bootstrap-servers'] == '' ? inputs.parameters['created-bootstrap-servers'] : inputs.parameters['provided-kafka-bootstrap-servers']"
+
+    - name: get-user-satisfied-confirmation
+      suspend: {}
+
+    - name: id-generator
```
peternied, line 183: Can these functions(?) be pulled out into their own files? really easy to get lost in these files as is

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/captureReplay.yaml
peternied, line 1: Maybe add a decomposition of how these workflow definitions can be read, or if there is a guide / tutorial that I should review first happy to take a look.  Not sure how to 'mentally scope' what these workflows govern.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/captureReplay.yaml
```diff
@@ -0,0 +1,196 @@
...
+          - name: kafka-cluster-setup
+            templateRef:
+              name: kafka-setup
+              template: cluster-deploy
+            when: "'{{inputs.parameters.provided-kafka-bootstrap-servers}}' == ''"
+            dependencies: [ id-generator ]
+            arguments:
+              parameters:
+                - name: kafka-name
+                  value: "{{inputs.parameters.kafka-prefix}}-{{=last(split(tasks['id-generator'].id, '-'))}}"
```
peternied, line 53: How can we validate this value is constructed correctly?  What happens if we get the task reference wrong?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/createSnapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+spec:
+  entrypoint: snapshot-workflow
+  templates:
+    - name: snapshot-workflow
+      inputs:
+        parameters:
+          - name: snapshot-name
+            value: "migration-assistant-snapshot"
+          - name: source-config
+            value: "{ \"endpoint\": \"http://ma-capture-proxy:9200\" }"
```
peternied, line 15: Hmm, won't this cause a snapshot to be created during replay?  We could avoid this by hitting the target directly, but I'm not sure that is correct either.  This might be out of scope for this change.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/createSnapshot.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - name: config-contents
+      retryStrategy:
+        limit: "200"
+        retryPolicy: "Always"
+        backoff:
+          duration: "5"
+          factor: 2
+          maxDuration: "300"
+      steps:
+        - - name: check-snapshot-completion
```
peternied, line 73: Is this creating a new MA console each execution, or reusing the existing one?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/documentBulkLoad.yaml
```diff
@@ -0,0 +1,404 @@
...
+                  value: "{{inputs.parameters.session-name}}"
+                  # TODO - parse the other stuff out and into the args.
+
+    - name: run-bulk-load
+      inputs:
+        parameters:
+          # Required parameters (no default names)
+          - name: session-name
+            description: "Name of the unique session to disambiguate coordinating indices"
+          - name: target-host
```
peternied, line 55: There seems to be a lot of duplicate declaration for parameters and arguments, are there ways to avoid this pattern?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/examples/fanIn.yaml
peternied, line 1: Should this be reviewed/included?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/Chart.yaml
peternied, line 1: Could you add a separate README.md for how this chart functions?  Seems like its a lot of connective tissue for giving Argo access to spin up/down resources, but I'm out of my depth if there is something interesting that we should be looking into over here.

peternied (commented):
Thanks for putting this out for review, I've got some structural questions around the declaration of the workflows and how they are related to one another.  Lots of comments inline, hopefully this helps give guidance on where I am coming from and isn't too disruptive.

Look forward to seeing more!

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -0,0 +1,420 @@
...
+  configs:
+    metrics:
+      - prometheus
+      - cloudwatch
+    traces:
+      - jaeger
+      - xray
+
+
+charts:
```
lewijacn, line 67: I'm assuming these charts should be considered standard charts that we always deploy, whereas an argo workflow will deploy charts like Replayer/RFS. I assume things like localstack will be disabled by default or a separate values.yaml will be made for an EKS setup

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/createS3SnapshotBucket.yaml
```diff
@@ -0,0 +1,92 @@
+# templates/create-s3-bucket-job.yaml
+{{- if .Values.snapshotBucketConfiguration.create }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: create-s3-bucket-{{ .Release.Name }}
```
lewijacn, line 6: Hoping ACK can handle this logic for us both in EKS and locally with LocalStack but we will see

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/createFullMigrationWorkflow.yaml
lewijacn, line 1: I assume this are reference workflows that have been pre-filled in with details from the default environment? Curious what your future plans are with these?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/templates/fullMigration.yaml
```diff
@@ -0,0 +1,322 @@
...
+      - name: s3-snapshot-configmap
+        value: "s3-snapshot-config"
+
+  templates:
+    # Main workflow template
+    - name: main
+      inputs:
+        parameters:
+          - name: source-migration-configs
+#                {\"source\":{\"endpoint\":\"https://sourceA.example.com\",\"allow_insecure\":true,\"basic_auth\":{\"username\":\"admin\",\"password\":\"admin\"}},\"snapshot-and-migration-configs\":[{\"indices\":[\"index_1\",\"index_2\"],\"migrations\":[{\"metadata\":{\"mappings\":{\"properties\":{}},\"settings\":{}},\"documentBackfillConfigs\":[{\"indices\":[\"index_1\"]},{\"indices\":[\"index_2\"]}]}]}],\"replayer-config\":{\"batchSize\":1000,\"concurrency\":4}},
```
lewijacn, line 33: nit: Seeing commented out sources in a few places in these templates

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/templates/replayer.yaml
```diff
@@ -0,0 +1,220 @@
...
+            replicas: {{inputs.parameters.replicas}}
+            selector:
+              matchLabels:
+                app: replayer
+            template:
+              metadata:
+                labels:
+                  app: replayer
+              spec:
+                containers:
```
lewijacn, line 95: I'd like to see us make this pattern for argument passing a bit cleaner for proxy and replayer. Ideally we would be able to loop over the input parameters but don't know if that is supported

lewijacn:
Performed a test run just now. Was able to see workflow kick off and backfill complete, but argo workflow errored out once the RFS pods started failing since they had finished their work. Below are the commands I executed:

```
cd deployment/k8s
./buildDockerImagesMini.sh
kubectl apply -f https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/main/example/prometheus-operator-crd/monitoring.coreos.com_servicemonitors.yaml
helm install -n ma ma charts/aggregates/migrationAssistantWithArgo/
helm install -n ma tc charts/aggregates/testClusters/
kubectl --namespace ma port-forward svc/argo-server 2746:2746
kubectl -n ma exec --stdin --tty  migration-console-0 -- /bin/bash
// From the console
curl -XPUT https://elasticsearch-master:9200/my-test-index/_doc/1 -H "Content-Type: application/json" -d '{"message": "Hello, world!"}' --insecure -u admin:admin
kubectl create -f workflows/createFullMigrationWorkflow.yaml
```

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/templates/fullMigration.yaml
gregschohn: deleted those

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
```diff
@@ -173,6 +172,11 @@ public static class Parameters {
                 "E.g. '(.* /ephemeral/*|GET /_cat/.*)' to ignore capturing all traffic for '/ephemeral' AND " +
                 "all GET requests to /_cat/.*")
         public String suppressMethodAndPath;
+        @Parameter(required = false,
+            names = { "--kafkaTopic" },
+            arity = 1,
+            description = "Name of the topic to write captured traffic to.")
+        public String kafakTopicName = KafkaCaptureFactory.DEFAULT_TOPIC_NAME_FOR_TRAFFIC;
```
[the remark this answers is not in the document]
gregschohn: My only hesitation is that the proxy and replayer may be dealing with multiple topics in the near future.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/OpenSearchWorkCoordinator.java
```diff
@@ -120,6 +120,7 @@ static class WorkItemWithPotentialSuccessors {
         final List<String> successorWorkItemIds;
     }
 
+    public final String indexName;
```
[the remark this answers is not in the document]
gregschohn: Made this protected.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/createFullMigrationWorkflow.yaml
gregschohn: This is a sample workflow that uses the template that must first be installed (and is installed in the top-level helm chart).  There are some sample values within the workflow templates that we'll want to remove and require callers to specify, but for testing through things now, it was more convenient to leave some hardcoded constants.  Think of values moving from inline literals up the stack.  Eventually all of these will be defined exclusively in the top-level configs OR be default configs that could work for anybody. For now, it isn't obvious yet where different parameters will land with regards to needing to be configured vs not.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/configs/s3/createS3SnapshotBucket.yaml
```diff
@@ -0,0 +1,92 @@
...
+                  name: s3-snapshot-config
+                  key: ENDPOINT
+          command:
+            - /bin/sh
+            - -c
+            - |
+              set -x
+
+              # Setting up variables
+              export AWS_DEFAULT_REGION=AWS_REGION
```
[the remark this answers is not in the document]
lewijacn: What I'm getting at here is that I don't think you want to set the value of AWS_DEFAULT_REGION to the string 'AWS_REGION'

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/models/snapshot.py
```diff
@@ -27,7 +27,8 @@
                 'schema': {
                     'repo_uri': {'type': 'string', 'required': True},
                     'aws_region': {'type': 'string', 'required': True},
-                    'role': {'type': 'string', 'required': False}
+                    'role': {'type': 'string', 'required': False},
+                    'endpoint': {'type': 'string', 'required': False}
```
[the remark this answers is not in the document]
lewijacn: Yes it should only be the java processes that use this currently. We only call the create snapshot java process that I am aware of from the console library which probably doesn't need this. One potential spot would possibly be our kafka export logic, but it is not currently linked into the console library to use the services.yaml.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -123,6 +124,12 @@ public static class Args {
                 + " also supply --s3-local-dir and --s3-repo-uri.  Mutually exclusive with --snapshot-local-dir."))
         public String s3Region = null;
 
+        @Parameter(required = false,
+            names = { "--s3-endpoint", "--s3Endpoint" },
+            description = ("The endpoint URL to use for S3 calls.  " +
+                "For use when the default AWS ones won't work for a particular context."))
```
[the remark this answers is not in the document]
peternied: What do you think about updating the description to mention VPC Endpoints and then link to the AWS doc - without that context it wasn't clear how this could be used.

RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/WorkCoordinatorFactory.java
```diff
@@ -39,12 +48,26 @@ public OpenSearchWorkCoordinator get(
...
+                workerId,
+                clock,
+                workItemConsumer);
         } else {
             throw new IllegalArgumentException("Unsupported version: " + version);
         }
     }
 
+
+    public String getFinalIndexName() {
```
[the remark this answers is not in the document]
peternied: I'd stil like to hide all other index names from possibly being referenced instead of the correct 'final' one, what do you think about making that change?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/middleware/clusters.py
peternied: By deleting the working state we could have running RFS workers do some very confusing things.  If we want to bucket this into 'advanced' options that we won't protect our users from that is ok with me, but I'd just like to be sure that is our expectation.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/README.md
peternied: Is that happening in this PR or future PR?  I'd prefer less, more trusted content in our readme.md otherwise its easy to get stale and confusing for our future selves.

lewijacn (approved):

PR #1412, opened 2025-04-02
Update replica count for az compatibility

mikaylathompson:
### Description

This PR changes/fixes our behavior around replica counts in situations where users have routing balance enforced with multiple awareness attributes (generally zones).

The evaluate command now catches these cases by querying the settings and comparing it to the replica count, giving the user a message like:

> ERROR - index_1_0 failed due to incompatible replica count for awareness attribute count: A replica count of 0 is not compatible with 2 awareness attributes (usually zones). The metadata migration tool can automatically remedy this by increasing the replica count to a compatible number if run with the command line parameter `--cluster-awareness-attributes 2`

When run with the parameter described (which is a renamed version of our previous `--min-num-replicas` param), it will adjust the replica count before attempting to create the indices.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2185

### Testing
Testing included.

### Check List


MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
```diff
@@ -46,9 +46,11 @@ public class MigrateOrEvaluateArgs {
...
 
     // https://opensearch.org/docs/2.11/api-reference/cluster-api/cluster-awareness/
-    @Parameter(names = {"--min-replicas" }, description = "Optional.  The minimum number of replicas configured for migrated indices on the target."
-            + " This can be useful for migrating to targets which use zonal deployments and require additional replicas to meet zone requirements.  Default: 0")
-    public int minNumberOfReplicas = 0;
+    @Parameter(names = {"--cluster-dimensionality" }, description = "Optional. This sets up a transformation for all indices to ensure that the number of replicas"
+            + "is compatible with the dimensionality of the cluster (usually the number of zones). This transformation strictly increases the number of replicas for each"
+            + "index to meet the cluster's dimensionality.  This can be useful for migrating to targets which use zonal deployments and require additional replicas to "
+            + "meet zone requirements.  Default: 0 (which does not apply any transformation)")
+    public int clusterDimensionality = 0;
```
AndreKurait, line 53: Would it make more sense for this to be a default of 1

RFS/src/main/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11.java
```diff
@@ -71,7 +95,17 @@ private void createInner(IndexMetadata index,
             if (alreadyExists) {
                 result.failureType(CreationFailureType.ALREADY_EXISTS);
             }
+
+            if (mode == MigrationMode.SIMULATE) {
```
AndreKurait, line 99: Should we only make this call on evaluate, or on both? What is the experience if someone skips evaluate and just does migrate?

MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
mikaylathompson: Yeah, it won't actually be a behavior difference, but conceptually makes more sense.

RFS/src/main/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11.java
mikaylathompson: Then it'll attempt to actually make the call and (assuming we correctly interpreted the settings) fail with the `IncompatibleReplicaCountException` that we're checking the response for below.

mikaylathompson: I think if the user asks to migrate, it's better for us to attempt and give them the real error than to predict the error and block the attempt

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
```diff
@@ -0,0 +1,92 @@
...
+            MigrateOrEvaluateArgs arguments = new MigrateOrEvaluateArgs();
+            arguments.sourceArgs.host = sourceCluster.getUrl();
+            arguments.targetArgs.host = targetCluster.getUrl();
+
+            // Evaluate first and check that errors were thrown appropriately
+            MigrationItemResult initialEvaluateResult = executeMigration(arguments, MetadataCommands.EVALUATE);
+            if (availabilityZoneCount == 1) {
+                Assertions.assertTrue(initialEvaluateResult.getItems().getIndexes().stream().allMatch(CreationResult::wasSuccessful));
+            } else {
+                Assertions.assertFalse(initialEvaluateResult.getItems().getIndexes().stream().allMatch(CreationResult::wasSuccessful));
```
AndreKurait, line 70: Can we do an assertion on the error message here?

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
```diff
@@ -0,0 +1,92 @@
...
+                        });
+                    }
+            );
+
+            MigrateOrEvaluateArgs arguments = new MigrateOrEvaluateArgs();
+            arguments.sourceArgs.host = sourceCluster.getUrl();
+            arguments.targetArgs.host = targetCluster.getUrl();
+
+            // Evaluate first and check that errors were thrown appropriately
+            MigrationItemResult initialEvaluateResult = executeMigration(arguments, MetadataCommands.EVALUATE);
```
AndreKurait, line 66: Can we add a test to run migrate with the invalid setting and assert on the error message given?

MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
```diff
@@ -46,9 +46,11 @@ public class MigrateOrEvaluateArgs {
     public DataFilterArgs dataFilterArgs = new DataFilterArgs(); 
 
     // https://opensearch.org/docs/2.11/api-reference/cluster-api/cluster-awareness/
-    @Parameter(names = {"--min-replicas" }, description = "Optional.  The minimum number of replicas configured for migrated indices on the target."
-            + " This can be useful for migrating to targets which use zonal deployments and require additional replicas to meet zone requirements.  Default: 0")
-    public int minNumberOfReplicas = 0;
+    @Parameter(names = {"--cluster-dimensionality" }, description = "Optional. This sets up a transformation for all indices to ensure that the number of replicas"
```
AndreKurait, line 49: The term dimensionality is throwing me off here, i'm not sure if it's used in the opensearch documentation. Perhaps "zones" would be a more common term

peternied (commented):
> index_1_0 failed due to incompatible replica count for awareness attribute count: A replica count of 0 is not compatible with 2 awareness attributes

Can we improve this message? IMO it isn't as important to know what is wrong but how the user can fix it.

MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
mikaylathompson: Yep, I agree with this. I'm trying to figure out where I got the term dimensionality, but I don't see it in the documentation, so I agree with switching to the widely used terms

MetadataMigration/src/test/java/org/opensearch/migrations/ReplicaCountWithAZsTest.java
mikaylathompson: We can -- are you looking for something specific in the error message or confirming that it was a `IncompatibleReplicaCount` exception?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11.java
```diff
@@ -72,7 +72,10 @@ private void checkForReplicaCountIncompatibility(ObjectNode settings, AwarenessA
         // To be compatible with the number of awareness attributes (usually zones), the awareness attribute must be divisible
         // by the replica count + 1
         if ((replicaCount + 1) % awarenessAttributes != 0) {
-            throw new IncompatibleReplicaCountException("A replica count of " + replicaCount  + " is not compatible with " + awarenessAttributes + " awareness attributes", null);
+            var replicaCountMessage = ("A replica count of %d is not compatible with %d awareness attributes (usually zones). " +
```
peternied, line 75: 👍

mikaylathompson:
Updated evaluate output:
```

Migration Candidates:
   Index Templates:
      <NONE FOUND>

   Component Templates:
      <NONE FOUND>

   Indexes:
      index_1_1, index_1_3, index_5_1, index_5_3
      ERROR - index_1_0 failed due to incompatible replica count for awareness attribute count: A replica count of 0 is not compatible with 2 awareness attributes (usually zones). The metadata migration tool can automatically remedy this by increasing the replica count to a compatible number if run with the command line parameter `--cluster-awareness-attributes 2`
      ERROR - index_1_2 failed due to incompatible replica count for awareness attribute count: A replica count of 2 is not compatible with 2 awareness attributes (usually zones). The metadata migration tool can automatically remedy this by increasing the replica count to a compatible number if run with the command line parameter `--cluster-awareness-attributes 2`
      ERROR - index_5_0 failed due to incompatible replica count for awareness attribute count: A replica count of 0 is not compatible with 2 awareness attributes (usually zones). The metadata migration tool can automatically remedy this by increasing the replica count to a compatible number if run with the command line parameter `--cluster-awareness-attributes 2`
      ERROR - index_5_2 failed due to incompatible replica count for awareness attribute count: A replica count of 2 is not compatible with 2 awareness attributes (usually zones). The metadata migration tool can automatically remedy this by increasing the replica count to a compatible number if run with the command line parameter `--cluster-awareness-attributes 2`

   Aliases:
      <NONE FOUND>
```

RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
```diff
@@ -85,4 +86,36 @@ private static Map.Entry<String, String> getUnknownSetting(JsonNode json) {
             return Map.entry(entry.getKey().asText(), matcher.group(1));
         }).orElse(null);
     }
+
+    /** Awareness attribute exceptions (when the replica count doesn't match the number of zones) present slightly differently
```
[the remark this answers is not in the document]
mikaylathompson: ^ addressed

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
```diff
@@ -73,6 +77,68 @@ public Version getClusterVersion() {
...
+
+        Optional.ofNullable(getSettingFromPersistentOrDefaults(balanceAttributeSetting, settings))
+            .ifPresent(attributes -> {
+                attributes.forEach(attributeName -> {
+                    Optional.ofNullable(getSettingFromPersistentOrDefaults(
+                            balanceAttributeValues + attributeName.asText() + ".values", settings))
+                        .map(JsonNode::asText)
+                        .map(text -> text.split(","))
+                        .map(Arrays::asList)
+                        .ifPresent(attributeValues::addAll);
```
AndreKurait, line 131: Is this behavior accurate with more than one attribute? It looks like we'd need to take the product of the size of all the forced attributes.

e.g.

```
cluster.routing.allocation.awareness.attributes: rack,zone
cluster.routing.allocation.awareness.force.rack.values: rack1,rack2,rack3
cluster.routing.allocation.awareness.force.zone.values: zone1,zone2
```

Would have a cardinality of 6.

MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
```diff
@@ -46,9 +46,12 @@ public class MigrateOrEvaluateArgs {
     public DataFilterArgs dataFilterArgs = new DataFilterArgs(); 
 
     // https://opensearch.org/docs/2.11/api-reference/cluster-api/cluster-awareness/
-    @Parameter(names = {"--min-replicas" }, description = "Optional.  The minimum number of replicas configured for migrated indices on the target."
-            + " This can be useful for migrating to targets which use zonal deployments and require additional replicas to meet zone requirements.  Default: 0")
-    public int minNumberOfReplicas = 0;
+    @Parameter(names = {"--cluster-awareness-attributes" }, description = "Optional. This sets up a transformation for all indices to ensure"
```
AndreKurait, line 49: Sorry for going back and forth on this, but reading this in context with the possibility having multiple awareness attributes doesn't seem to fit, may i suggest

```
clusterAwarenessAttributeCombinationCount / 
--cluster-awareness-attribute-combination-count
```

Alternative to combination-count, while more technical, we could use the term `cardinality`

mikaylathompson: I worry that we're getting so accurate that we're losing all semblance of readability

mikaylathompson: I think having multiple awareness attributes is really an edge case. I don't think it's possible in AWS. While it should be supported for self-managed targets (meaning, in part, that I agree with your other comment about multiplying the attributes), I really don't think we need to cater our naming and documentation around it and make the experience worse for the far-more-common use case.

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
mikaylathompson: I _think_ you're right about this. The documentation is rather lacking, but it seems like the most likely case. Updating now.

MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
```diff
@@ -46,9 +46,12 @@ public class MigrateOrEvaluateArgs {
     public DataFilterArgs dataFilterArgs = new DataFilterArgs(); 
 
     // https://opensearch.org/docs/2.11/api-reference/cluster-api/cluster-awareness/
```
AndreKurait, line 48: nit: https://opensearch.org/docs/2.11/tuning-your-cluster/#forced-awareness

MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
AndreKurait: Sounds good, this is certainly better than what we had it

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
mikaylathompson: Andre found a better description of the behavior:

> The routing.allocation.awareness.balance setting is false by default. When it is set to true, the total number of shards for the index must be a multiple of the highest count for any awareness attribute. For example, consider a configuration with two awareness attributes—zones and rack IDs. Let’s say there are two zones and three rack IDs. The highest count of either the number of zones or the number of rack IDs is three. Therefore, the number of shards must be a multiple of three. If it is not, OpenSearch throws a validation exception.

I've implemented this behavior and checked it with a test that validates the behavior.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
```diff
@@ -20,21 +20,21 @@ private TransformFunctions() {}
     public static Transformer getTransformer(
         Version sourceVersion,
         Version targetVersion,
-        int dimensionality,
+        int clusterDimensionality,
```
AndreKurait, line 23: nit: clusterAwarenessAttributes

MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
```diff
@@ -68,19 +68,23 @@ protected Transformer getCustomTransformer() {
         return new TransformerToIJsonTransformerAdapter(transformer);
     }
 
-    protected Transformer selectTransformer(Clusters clusters) {
+    protected Transformer selectTransformer(Clusters clusters, int clusterDimensionality) {
```
AndreKurait, line 71: nit: clusterAwarenessAttributes

RFS/src/main/java/org/opensearch/migrations/cluster/ClusterSnapshotReader.java
```diff
@@ -24,4 +25,8 @@ public interface ClusterSnapshotReader extends ClusterReader {
 
     /** gets the soft deletes can field data */
     String getSoftDeletesFieldData();
+
+    default AwarenessAttributeSettings getAwarenessAttributeSettings() {
```
AndreKurait, line 29: I just realized this is added on the reader, do we use this for `ClusterReader` or would it make more sense to just have this method on `ClusterWriter`

AndreKurait (approved):
Looks great! Thanks

PR #1422, opened 2025-04-03
BYOS - Bring Your Own Snapshot for CDK

lewijacn:
### Description
This change introduces the ability to specify an existing snapshot with CDK and have that be utilized by downstream services such as RFS and the Migration Console

Sample CDK context option:
```
{
    "snapshot": {
        "snapshotName": "test-snapshot",
        "s3Uri": "s3://snapshot-bucket-123456789012-us-east-2/snapshot-repo",
        "s3Region": "us-east-2"
    }
}
```
~Still verifying this works on a deployed environment and seeing if any issues exist to use this cross-region or cross-account as well as adding official documentation~

Documentation PR to review here: https://github.com/opensearch-project/documentation-website/pull/9573

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2219

### Testing
CDK tests

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
@@ -468,11 +468,11 @@ export function parseClusterDefinition(json: any): ClusterYaml {
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export function parseSnapshotDefinition(json: any): SnapshotYaml {
-    const snapshotName = json.snapshot_name
-    const s3Region = json.s3_region
-    const s3Uri = json.s3_uri
+    const snapshotName = json.snapshotName
+    const s3Region = json.s3Region
+    const s3Uri = json.s3Uri
     if (!snapshotName || !s3Region || !s3Uri) {
-        throw new Error('Missing at least one of the required snapshot fields: snapshot_name, s3_region, s3_uri');
+        throw new Error('Missing at least one of the required snapshot fields: snapshotName, s3Region, s3Uri');
     }
     const snapshotYaml = new SnapshotYaml()
     snapshotYaml.snapshot_name = snapshotName
--- a/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
@@ -4,7 +4,7 @@ import {ContainerImage, CpuArchitecture} from "aws-cdk-lib/aws-ecs";
 import {RemovalPolicy, Stack} from "aws-cdk-lib";
 import { IStringParameter, StringParameter } from "aws-cdk-lib/aws-ssm";
 import * as forge from 'node-forge';
-import { ClusterYaml } from "./migration-services-yaml";
+import {ClusterYaml, SnapshotYaml} from "./migration-services-yaml";
 import { CdkLogger } from "./cdk-logger";
 import { mkdtempSync, writeFileSync } from 'fs';
 import { join } from 'path';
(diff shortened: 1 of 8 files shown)


deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -466,6 +466,20 @@ export function parseClusterDefinition(json: any): ClusterYaml {
     return new ClusterYaml({endpoint, version, auth})
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export function parseSnapshotDefinition(json: any): SnapshotYaml {
+    const snapshotName = json.snapshotName
+    const s3Region = json.s3Region
+    const s3Uri = json.s3Uri
```
mikaylathompson, line 473: double checking for our snapshot definition params: is there anywhere in the (pre-existing) code where we assume a specific snapshot repo name? Do we need to collect that as well?

mikaylathompson (commented):
No concerns with how this looks!

I would love if you could run a test (doesn't have to be in the code necessarily) to prove that we can correctly distinguish between multiple snapshots in the same repo and multiple repos in the same s3 bucket (as well as the cross-region and cross-account tests you mentioned)

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -378,8 +382,18 @@ export class StackComposer {
             servicesYaml.client_options.user_agent_extra = props.migrationsUserAgent
         }
 
-        // There is an assumption here that for any deployment we will always have a target cluster, whether that be a
-        // created Domain like below or an imported one
+        const existingSnapshotDefinition = this.getContextForType('snapshot', 'object', defaultValues, contextJSON)
+        let snapshotYaml
+        if (existingSnapshotDefinition) {
```
AndreKurait, line 387: Should we still be deploying the S3 bucket we create for artifacts if this is given?

deployment/cdk/opensearch-service-migration/options.md
```diff
@@ -49,6 +49,22 @@ In all other cases, the required components of each cluster object are:
     3. Basic auth with plaintext password (only supported for the source cluster and not recommended): `{"type": "basic", "username": "admin", "password": "admin123"}`
     4. Basic auth with password in secrets manager (recommended): `{"type": "basic", "username": "admin", "passwordFromSecretArn": "arn:aws:secretsmanager:us-east-1:12345678912:secret:master-user-os-pass-123abc"}`
 
+### Snapshot Definition Options
```
AndreKurait, line 52: What experience do we want if someone runs `console snapshot status` or `console snapshot delete` or `console snapshot create` with external snapshots provided?

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
lewijacn: We could potentially use the S3 bucket for other use cases like exporting Kafka records, so I did not remove

deployment/cdk/opensearch-service-migration/options.md
lewijacn: Right now unless the source is configured properly, these should fail since they all go through the source cluster. That failure may not be very clear though, and there is potentially a follow-up item here where we could add something to identify the snapshot as external and fail early on these console commands

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
lewijacn: From my testing, snapshot repo name doesn't seem to matter when reading directly from S3 and Metadata and RFS seem to just need the details you highlighted as well as the version of the cluster.  It could be a problem if the user actually wanted to use the console snapshot commands on that external snapshot, but that seems a little odd since the snapshot should already be existing

deployment/cdk/opensearch-service-migration/lib/stack-composer.ts
```diff
@@ -378,8 +382,18 @@ export class StackComposer {
...
+            snapshotYaml = parseSnapshotDefinition(existingSnapshotDefinition)
+        } else {
+            snapshotYaml = new SnapshotYaml();
+            snapshotYaml.snapshot_name = "rfs-snapshot"
+            const s3Uri = `s3://migration-artifacts-${account}-${stage}-${region}/rfs-snapshot-repo`;
```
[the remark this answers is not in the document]
lewijacn: Pushed an update to pull this from the Migration Assistance stack

deployment/cdk/opensearch-service-migration/options.md
lewijacn: Sure created a task here: https://opensearch.atlassian.net/browse/MIGRATIONS-2471

mikaylathompson (approved):

PR #1434, opened 2025-04-08
Fix HeaderAdder/RemoverHandler over reused connections by resetting on response

AndreKurait:
### Description
Currently, the HeaderAdderHandler and HeaderRemoverHandler only correctly performs their intended operation on the first request per connection. This change modifies them to reset between each request by triggering a reset when responses are received. A client should not send a second request until this occurs on HTTP/1.1.

### Testing
Added unit test. Verified functionality end to end too.

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
@@ -4,11 +4,23 @@ import java.nio.charset.StandardCharsets;
 
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.Unpooled;
+import io.netty.channel.ChannelDuplexHandler;
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.channel.ChannelInboundHandlerAdapter;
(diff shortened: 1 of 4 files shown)


TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
```diff
@@ -57,4 +68,14 @@ public void channelUnregistered(ChannelHandlerContext ctx) throws Exception {
         ReferenceCountUtil.release(headerLineToAdd);
         super.channelUnregistered(ctx);
     }
+
+    @Override
+    public void flush(ChannelHandlerContext ctx) throws Exception {
```
gregschohn, line 73: are the flush calls guaranteed to happen before another request can be processed (in a well-formed HTTP/1.x traffic stream)?

gregschohn (approved):

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/HeaderAdderHandler.java
AndreKurait: Updated to write since this wasn't guaranteed with flush

PR #1463, opened 2025-04-18
Replayer logging changes

mikaylathompson:
### Description
Behavior is as follows:
- all categories of logs are written to a local directory until they're ready to be rolled over
- then they're compressed and written to the shared logs volume

This PR initially included code that looked for an env var to decide which log4j2 properties file to use. It turns out that [exists natively in Log4j](https://logging.apache.org/log4j/2.x/manual/systemproperties.html#log4j2.configurationFile), so we're using that instead.

### Testing

Testing needed! Discussed with @gregschohn to add tests that validate the new logging properties file path behavior, especially when the file doesn't exist.

### Check List


TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
```diff
@@ -24,18 +26,21 @@ appender.ReplayerLogFile.policies.time.interval = 60
...
+appender.ReplayerLogFile.strategy.compressionLevel = 9
 appender.ReplayerLogFile.immediateFlush = false
 
 appender.OUTPUT_TUPLES.type = RollingRandomAccessFile
 appender.OUTPUT_TUPLES.name = OUTPUT_TUPLES
-appender.OUTPUT_TUPLES.fileName = ${tuplesFileNamePrefix}.log
-appender.OUTPUT_TUPLES.filePattern = ${tuplesFileNamePrefix}-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz
+appender.OUTPUT_TUPLES.fileName = ${tempDir}/tuples.log
+appender.OUTPUT_TUPLES.filePattern = ${logsDir}/${hostName}/tuples/tuples-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz
+appender.OUTPUT_TUPLES.tempCompressedFilePattern = ${tempDir}/tuples-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz.tmp
```
gregschohn, line 36: This won’t matter for network IO, I’m just surprised… does Log4j2 require writing a temporary file before copying?  I suppose that the move is an atomic FS operation, which makes these safer operations… if they’re in the same volume, it’s nearly a no-op, so it makes as to why it would be engineered that way.

TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
```diff
@@ -14,8 +15,9 @@ appender.console.direct = true
 
 appender.ReplayerLogFile.type = RollingRandomAccessFile
 appender.ReplayerLogFile.name = ReplayerLogFile
-appender.ReplayerLogFile.fileName = logs/replayer.log
-appender.ReplayerLogFile.filePattern = logs/%d{yyyy-MM}{UTC}/replayer-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz
+appender.ReplayerLogFile.fileName = ${tempDir}/replayer.log
+appender.ReplayerLogFile.filePattern = ${logsDir}/${hostName}/replayer-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz
+appender.ReplayerLogFile.tempCompressedFilePattern = ${tempDir}/replayer-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz.tmp
```
AndreKurait, line 20: I'm getting 
```
2025-04-19T13:02:20.526718719Z main ERROR RollingRandomAccessFile contains an invalid element or attribute "tempCompressedFilePattern"
```

AndreKurait: seems like it may be need to be put next to compressionLevel

TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
```diff
@@ -24,18 +26,21 @@ appender.ReplayerLogFile.policies.time.interval = 60
...
+appender.OUTPUT_TUPLES.fileName = ${tempDir}/tuples.log
+appender.OUTPUT_TUPLES.filePattern = ${logsDir}/${hostName}/tuples/tuples-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz
+appender.OUTPUT_TUPLES.tempCompressedFilePattern = ${tempDir}/tuples-%d{yyyy-MM-dd-HH-mm}{UTC}-%i.log.gz.tmp
 appender.OUTPUT_TUPLES.layout.type = PatternLayout
 appender.OUTPUT_TUPLES.layout.pattern = %m%n
 appender.OUTPUT_TUPLES.policies.type = Policies
 appender.OUTPUT_TUPLES.policies.size.type = SizeBasedTriggeringPolicy
 appender.OUTPUT_TUPLES.policies.size.size = 10 MB
 appender.OUTPUT_TUPLES.strategy.type = DefaultRolloverStrategy
+appender.OUTPUT_TUPLES.strategy.compressionLevel = 9
```
AndreKurait, line 43: Thanks for adding this, let's set this to 1

gregschohn: I was thinking 9 (I thought that was the max) would make the most sense since we're likely network constrained.  If we can't decide, the default is probably the best way to go for now.

TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
AndreKurait: The tempCompressedFilePattern should probably be in the EFS volume, and with a different name, e.g `temp-tuples...`

TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
AndreKurait: We can keep this at the default, but be explicit (e.g. 7) so that customers can know it's tweakable if they are network vs cpu constrained

AndreKurait: It certainly uses more cpu for 9 and we don't have metrics on the increase in compression ratio, i'm good to leave it at 7 which is where it's at now

AndreKurait:
@mikaylathompson, i've made some updates for the default config here: https://github.com/mikaylathompson/opensearch-migrations/pull/6 Let me know what you think

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -333,6 +355,8 @@ private static Parameters parseArgs(String[] args) {
 
     public static void main(String[] args) throws Exception {
         System.err.println("Got args: " + String.join("; ", args));
+        reloadConfiguration(System.getenv(LOG4J2_PROPERTIES_FILE));
```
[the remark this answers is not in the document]
mikaylathompson: Update from testing locally: this works perfectly, exactly as the new env var was intended to. Stripping out the code related to this.

TrafficCapture/trafficReplayer/src/main/resources/log4j2.properties
mikaylathompson: fixed

peternied (commented):
Two considerations, but if @AndreKurait and @mikaylathompson are happy with this I'm fine to merge.

1/ Should these changes be replicated in the RFS logger?
2/ The complexity of this file is growing, almost seems like having a script/programmatic construction would be useful for reuse

AndreKurait (approved):

PR #1469, opened 2025-04-21
Enable migration console role to pass user-provided Snapshot role

jugal-chauhan:
### Description
This PR expands the IAM permissions granted to the Migration Console Role used in OpenSearch migrations. Previously, this role could only pass a specifically created snapshot role. With this update, it is now allowed to pass any IAM role within the account.

This change supports customer-supplied snapshot roles at runtime, a requirement for advanced use cases like cross-region snapshot operations, where snapshots may be stored in or restored from S3 buckets managed by roles in different regions. By loosening the pass role permissions, the Migration Assistant becomes more flexible in multi-region environments.

This does not alter existing behavior for standard in-region migrations but provides the necessary permissions for extended snapshot workflows.

### Issue Resolved
- Related [MIGRATIONS-2507](https://opensearch.atlassian.net/browse/MIGRATIONS-2507)

### Check List


deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -260,8 +260,13 @@ export class MigrationConsoleStack extends MigrationServiceCore {
 
         const openSearchPolicy = createOpenSearchIAMAccessPolicy(this.partition, this.region, this.account)
         const openSearchServerlessPolicy = createOpenSearchServerlessIAMAccessPolicy(this.partition, this.region, this.account)
+        const passRolePolicy = new PolicyStatement({
```
AndreKurait, line 263: We should already have pass role for the existing managed snapshot, instead of adding a new policy, can we update the existing one?

jugal-chauhan: Found and modified `requestingRole` to grant broader permissions

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -211,9 +211,16 @@ export function createSnapshotOnAOSRole(scope: Construct, artifactS3Arn: string,
         resources: [`${artifactS3Arn}/*`],
     }));
 
-    // The Migration Console Role needs to be able to pass the snapshot role
+    // The Migration Console Role needs to be able to pass the snapshot role as well as any other role
     const requestingRole = Role.fromRoleArn(scope, 'RequestingRole', migrationConsoleTaskRoleArn);
     snapshotRole.grantPassRole(requestingRole);
+    
+    // Also grant broader permission to pass any role (needed for other operations like multiplier test for cross-region snapshot creation)
```
lewijacn, line 218: nit: A test case normally isn't good grounds for loosening IAM permissions, supporting cross-region snapshot itself makes more sense

jugal-chauhan: I have made the necessary changes in PR Title and description, Jira task and in-line comment.

lewijacn (approved):

PR #1470, opened 2025-04-21
Add support for ES 8.x local K8s testing

lewijacn:
### Description
This changes adds support for executing MA workflow tests against an Elasticsearch 8.18 cluster. It also adds a Jenkins pipeline which will execute this workflow. Currently the test cases fail with the following expected error as we have not yet added the support in MA for this version:
```
Exception in thread "main" java.lang.IllegalArgumentException: Unsupported version: ELASTICSEARCH 8.18.0
	at org.opensearch.migrations.bulkload.common.OpenSearchClientFactory.getOpenSearchClientClass(OpenSearchClientFactory.java:75)
	at org.opensearch.migrations.bulkload.common.OpenSearchClientFactory.determineVersionAndCreate(OpenSearchClientFactory.java:45)
	at org.opensearch.migrations.CreateSnapshot.run(CreateSnapshot.java:139)
	at org.opensearch.migrations.CreateSnapshot.main(CreateSnapshot.java:131) 
```

As this version support is added these test cases should pass and new test cases should be added. For now this setup should be helpful for developers adding this support to test locally.  From the `libraries/testAutomation` directory a command like the following can be used to perform a simple backfill test with this setup:
```
pipenv run app --source-version=ES_8.x --test-ids=0001
```

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2500

### Testing
Local and Jenkins testing. See staging Jenkins pipeline for this

### Check List


jenkins/migrationIntegPipelines/elasticsearch5xK8sLocalTestCover.groovy
```diff
@@ -6,4 +6,8 @@ library identifier: "migrations-lib@${gitBranch}", retriever: modernSCM(
          remote: "${gitUrl}"])
 
 // Shared library function (location from root: vars/elasticsearch5xK8sLocalTest.groovy)
-elasticsearch5xK8sLocalTest()
+elasticsearch5xK8sLocalTest(
```
peternied, line 9: This cause issues with the jenkins test cases?
```
lso:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: b416532e-acfc-4c2c-86d1-57400cb032ef
java.lang.RuntimeException: The gitUrl argument must be provided to k8sLocalDeployment()
	at k8sLocalDeployment.call(k8sLocalDeployment.groovy:4)
	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.each(CpsDefaultGroovyMethods:2125)
	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.each(CpsDefaultGroovyMethods:2110)
	at com.cloudbees.groovy.cps.CpsDefaultGroovyMethods.each(CpsDefaultGroovyMethods:2151)
	at k8sLocalDeployment.call(k8sLocalDeployment.groovy:2)
```

source https://migrations.ci.opensearch.org/job/elasticsearch-5x-k8s-local-test/947/console

libraries/testAutomation/es-8-to-os-2-values.yaml
```diff
@@ -0,0 +1,201 @@
...
+      object:
+        repoPath: "/snapshot"
+        snapshotName: "migration-assistant-snapshot"
+        snapshotType: "fs"
+    metadataMigration:
+      allowRuntimeOverride: true
+      object:
+        clusterAwarenessAttributes: 1
+        metadataType: "from_snapshot"
+        sourceClusterVersion: "ES_8.18"
```
peternied, line 52: Shouldn't this be picked up from line 24?

libraries/testAutomation/es-8-to-os-2-values.yaml
```diff
@@ -0,0 +1,201 @@
+conditionalPackageInstalls:
```
peternied, line 1: This seems like a lot of copy pasta configuration, do we have a task to consolidate this?  Not a blocker for this change, but I'm not sure where we are converging here

libraries/testAutomation/README.md
```diff
@@ -12,7 +12,7 @@ pipenv install
 
 To run the test automation library, a command like below can be used:
 ```shell
-pipenv run app
+pipenv run app --source-version=ES_5.6
```
peternied, line 15: Looks like this isn't required, it uses the default which is the same as what you've included here.  Lets add a new section to the documentation about this.

libraries/testAutomation/testAutomation/test_runner.py
```diff
@@ -212,6 +214,18 @@ def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(
         description="Process inputs for test automation runner"
     )
+    parser.add_argument(
```
peternied, line 217: I'm getting a code smell for a leaky abstraction here where tests define matrix but the test runner also chooses sources/targets.  See Joel on Software's post The Law of Leaky Abstractions [1] for more details.

I like the idea of the test runner discovering and deploying all configurations that are needed rather than having an external system that has to know the matrix of what it will run.  What are you thinking longer term for these tests?

- [1] https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/

libraries/testAutomation/testAutomation/test_runner.py
```diff
@@ -247,18 +261,24 @@ def main() -> None:
...
-    test_cluster_env = TestClusterEnvironment(source_version="ES_5.6",
-                                              source_helm_values_path=es_5_6_values,
+    if args.source_version == "ES_5.6":
+        ma_chart_values_path = "es-5-to-os-2-values.yaml"
+        source_values = (f"{helm_charts_base_path}/components/elasticsearchCluster/"
+                         f"environments/es-5-6-single-node-cluster.yaml")
+    elif args.source_version == "ES_8.18":
+        ma_chart_values_path = "es-8-to-os-2-values.yaml"
+        source_values = (f"{helm_charts_base_path}/components/elasticsearchCluster/"
+                         f"environments/es-8-latest-single-node-cluster.yaml")
```
peternied, line 271: Can we use a convention that avoids hardcoding all these values, like we example the file to be named flavor 
'es', major '5', minor '6', and then just have the paths of these items follow those rules?

peternied (commented):
Nice progress here!

jenkins/migrationIntegPipelines/elasticsearch5xK8sLocalTestCover.groovy
lewijacn: I've update to verify in Jenkins. Before this gets merged in we should change the reference to my branch

libraries/testAutomation/testAutomation/test_runner.py
peternied: @lewijacn Not a blocker, What do you think of adding a diagram of what parts of this system drive what environments to spin up and select what tests to run?

IMO this would really our understanding of the test orchestration that will get more important when we add in argo or other layers of abstraction in the deployment & execution flow in MA.

lewijacn: I agree some sort of documentation/diagram could add some clarity here, my gut reaction is to wait until we have more clarity on the argo piece before investing too much in this structure right now as this piece could change a good amount

PR #1477, opened 2025-04-26
[Transformation Playground] Handling and Displaying Input Docs

mikaylathompson:
### Description
This PR handles pretty much everythign we need to get going (I think) for input docs.
You can now:
- upload one or more input documents
  - they are checked for valid json or ndjson
  - json docs are saved immediately; ndjson docs are split by line and saved as individual docs
  - all docs are persisted to local storage
- view input docs, pretty-printed
- delete input docs
- edit input doc content and title
  - also with json validation before saving

The trickiest thing I ran into is that it turns out that local storage has a ~10mb size limit (may vary by browser, etc.). This certainly blocks our ability to handle some customers' 100mb documents, but I think this is okay as a limitation as long as it's clear to the user what's happening and how to fix it (use a smaller file). I've attempted to make sure that's the case and have done quite a bit of local testing around this piece.

Note that if you're testing this out locally, the webpack/ace stuff is very weird and annoying and can (partially) impact editing. You may have to do `npm run build`, not just `npm run dev`, to get it packed up and available. If you do that and it still doesn't work, that sounds like a bug.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2489 (but I think this PR goes beyond exactly what's described there)

### Testing
Unit tests added for a lot of the json handling stuff as well as file uploads

### Check List

-------------

A few screenshots, for the sake of getting a quick summary:

1. Documents dragged into upload section, but not confirmed yet
Screenshot 2025-04-26 at 2 02 33 AM

2. Documents uploaded successfully
Screenshot 2025-04-26 at 2 02 38 AM

3. Viewing a document
Screenshot 2025-04-26 at 2 02 44 AM

4. Editing a document
Screenshot 2025-04-26 at 2 02 51 AM


frontend/src/utils/jsonUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+};
+
+export const validateJsonContent = (content: string): string | null => {
+  try {
+    JSON.parse(content);
+    return null;
+  } catch {
+    // If not regular JSON, check if it's newline-delimited JSON
+    return validateNewlineDelimitedJson(content);
```
lewijacn, line 17: Will this catch block still produce valid error messages if it is not a newline delimited json and just a malformed normal json?

frontend/src/context/PlaygroundProvider.tsx
```diff
@@ -33,20 +43,75 @@ export const PlaygroundProvider: React.FC<{ children: React.ReactNode }> = ({
...
+    const stateToSave = {
+      inputDocuments: state.inputDocuments,
+      transformations: state.transformations,
+    };
+
+    const stateJson = JSON.stringify(stateToSave);
+
+    // Update the total storage size
+    const sizeInBytes = new Blob([stateJson]).size;
+    setTotalStorageSize(sizeInBytes);
```
lewijacn, line 64: I was a bit confused why we were calculating size and storing in both effect functions. Do we just not store this value in state?

lewijacn (approved):
Looks good! I was able to run locally and provide and see validation of json documents. Left a couple minor comments

frontend/src/context/PlaygroundProvider.tsx
mikaylathompson: We do store it in state, but not in local storage.
The first `useEffect` (lines 25-44) is only run when the provider first renders, so it calculates how much space we're currently using from whatever is loaded from local storage.

The second `useEffect` (lines 53-94) runs when inputDocuments or transformations changes, and it theoretically could use the saved `storageSize` and just modify it by whatever the change was (e.g. add the size of the latest document), but it's implemented such that it doesn't know what the change was (e.g. which document was added, updated, or removed), it just has the new state in its entirety.

PR #1479, opened 2025-04-28
Refactored snapshot handling to make it easier to add new Lucene sources

gregschohn:
### Description

WIP - I've refactored snapshot handling throughout shard setup and document retrieval to make it easier to add new Lucene document sources.

I'm also in the process of adding new parameters -
1. There are parameters for the coordination cluster, of the same type as what we use for source and target, so that the target cluster doesn't need to be used for coordination.  A wholly independent cluster can be setup just for coordination.  There's a patch on the argo branch to support flexible index names too, so one could imagine running many deployments with the same coordination index.
2. snapshotLoader* is in place, though only partially implemented, to allow alternate ways to get shard info and create document streams.  I intend to load the specified class in via ServiceLoader, similar to how Transformations work today.
3. initialSubdivision hasn't been implemented yet, but I think it would be pretty easy and super useful, especially for the case where users want to reshard an index.  Rather than use a single worker at a time for a source shard, this would pull the shard down to find the number of documents, then split the work up into that many shards so that the migration could be completed by that factor faster (since there wouldn't be idling target shards).  This should be easy to implement by adding an end document index so that we can split a shard in parallel.  It would also be easy to detect the first one since the initial work item would have a -1 for its end index.

Everything else was just shuffling old code around into constructors and the newly created ElasticsearchSnapshotDocumentRepository, which wraps up the S3-centric objects and metadata that RFS had since been using beforehand. The EncryptedShardBucketCollectionDocumentRepository also has stub code that won't do anything.  Since we don't have a spec for reading encrypted files yet, this will be hived off, to be provided via a hot-loaded jar at some point in the future.

### Check List


peternied (commented):
High level, this looks good, thanks for cleaning up some of the snapshot reading interface.

RFS/src/main/java/org/opensearch/migrations/bulkload/ManuallyManagedLuceneShardRepository.java
```diff
@@ -0,0 +1,128 @@
...
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.Getter;
+import lombok.NonNull;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class ManuallyManagedLuceneShardRepository extends LuceneBasedDocumentRepository {
```
peternied, line 25: I'm not sure what to make of this class, it seems like it belongs in its own top level directory as it seems to be an independent tool.  What do you think of adding an explanation / data flow diagram of how this will be used and in what case in conjunction with the Metadata/RFS workflow?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_universal/RemoteDataProvider.java
```diff
@@ -21,15 +21,15 @@ public List<Snapshot> getSnapshots() {
     }
 
     @Override
-    public List<Index> getIndicesInSnapshot(String snapshotName) {
+    public List<Index> getIndicesInSnapshot() {
```
peternied, line 24: Really appreciate this refactor - for your consideration breakout the 'fix' for the snapshot name into its own PR so we could fast-track merging of that change and isolate it from the other work that you are doing.

peternied:
@gregschohn This has been open for more than a month, need any help with next steps or should this be closed out?

AndreKurait:
I connected with @gregschohn, we aligned to close out this PR and focus on a different way to cut this in https://opensearch.atlassian.net/browse/MIGRATIONS-2564

PR #1485, opened 2025-04-28
Update capture-replay-dashboard with rename

AndreKurait:
### Description
Dashboards cannot share the same name across all regions in an account. This change fixes deployment failures when the same stage name is used across regions along with adding simplification to the capture and replay dashboard, primarily fixing metric discovery with SEARCH, adding calculations for speedup observed and catchup time, and removing metrics that were unclear or added confusion.

### Issues Resolved

[MIGRATIONS-1608](https://opensearch.atlassian.net/issues/MIGRATIONS-1608)

### Testing
Jenkins run to verify successful deployment along with verifying json in AWS under a migration

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/constructs/migration-dashboard.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/constructs/migration-dashboard.ts
@@ -11,7 +11,7 @@ interface DashboardBody {
 }
 
 export interface MigrationDashboardProps {
-    readonly dashboardName: string;
+    readonly dashboardQualifier: string;
     readonly stage: string;
     readonly account: string;
     readonly region: string;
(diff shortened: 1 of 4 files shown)


deployment/cdk/opensearch-service-migration/lib/constructs/migration-dashboard.ts
```diff
@@ -11,20 +11,21 @@ interface DashboardBody {
...
-    constructor(scope: Construct, id: string, props: MigrationDashboardProps) {
-        super(scope, id);
+    constructor(scope: Construct, props: MigrationDashboardProps) {
+        const dashboardName = `MigrationAssistant_${props.dashboardQualifier}_${props.stage}_${props.region}`
+        super(scope, dashboardName);
 
         const dashboard = this.setDashboardVariables(props.dashboardJson, props);
-        new CfnDashboard(this, 'Dashboard', {
-            dashboardName: props.dashboardName,
+        new CfnDashboard(this, dashboardName, {
```
lewijacn, line 27: FYI I think this change in id name here will cause a stack conflict with our existing CDK deployments in jenkins, which is fine if it makes this clearer, but we may need to do some cleaning up of stacks until it is consistent across branches

AndreKurait: I'm not sure. Since the old value wasn't exported, i don't think there's any conflict to rename this.

The jenkins deployment in GHA had succeeded which points to this, but i'll keep this in mind post-merge

AndreKurait: I'm not sure. Since the old value wasn't exported, i don't think there's any conflict to rename this.

The jenkins deployment in GHA had succeeded which points to this, but i'll keep this in mind post-merge

lewijacn (approved):

PR #1489, opened 2025-04-30
[Transformation Playground] Transformation UI

mikaylathompson:
### Description
This PR basically gets all of the UI for the center column (transformations) functional.
You can now:
- add new transformations
- reorder transformations
- resize the container for each transformation (e.g. collapse or expand it)
- change the transformation's name
- edit the content, with syntax highlighting, error catching, etc.
- delete a transformation
- apply formatting to a transformation
- see a visual indicator of whether it's saved to local storage

The one thing you can't do in this PR is see the result of running them.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2491 (and beyond)

### Testing
I haven't actually added any tests for this one because it does a minimal amount of changing state (and those state changes are already pretty well tested). Let me know if you think this leaves gaps.

### Check List

A few screenshots:

Basic overview:
Screenshot 2025-04-30 at 3 15 55 PM

Dragging to reorder:
Screenshot 2025-04-30 at 3 16 09 PM

Showing unsaved state indicator:
Screenshot 2025-04-30 at 3 16 31 PM


lewijacn (commented):
Walked through this with Mikayla. We noticed one bug with the saving feature of the transformation that would cause inconsistent saving, but Mikayla is aware and working on fix for this. Other than this had no issue creating transformations, resizing them, moving them around, and deleting them 👍

mikaylathompson:
Updated the editor internals to fix the saving error and show clearer messages about when the transformation was saved or if it wasn't saved
Screenshot 2025-05-01 at 12 26 22 PM
Screenshot 2025-05-01 at 1 00 30 PM

frontend/src/components/playground/AceEditorComponent.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+// Import workers
+import jsonWorkerUrl from "ace-builds/src-noconflict/worker-json";
+import javascriptWorkerUrl from "ace-builds/src-noconflict/worker-javascript";
+
+// Configure Ace to use the imported workers
+ace.config.setModuleUrl("ace/mode/json_worker", jsonWorkerUrl);
+ace.config.setModuleUrl("ace/mode/javascript_worker", javascriptWorkerUrl);
+
+interface AceEditorComponentProps {
```
peternied, line 29: [Design] I've got a preference towards requiring all parameters, could we removal the `?` from many of these members?  This way we have fewer edges cases to handle within the component

frontend/src/components/playground/AceEditorComponent.tsx
```diff
@@ -0,0 +1,256 @@
...
+    }
+  }, [transformation, onSaveStatusChange]);
+
+  // Save the current content to local storage
+  const saveContent = useCallback(() => {
+    // Immediately set content to the exact content from the editor ref (our state may be stale)
+    const editorContent = editorRef.current?.editor.getValue();
+    const activeContent = editorContent ?? content;
+    // Skip update if transformation doesn't exist or if content is unchanged.
+    if (!transformation || activeContent === transformation.content) return;
```
peternied, line 104:
```suggestion
    if (activeContent === transformation?.content) return;
```

frontend/src/components/playground/AceEditorComponent.tsx
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    return () => {
+      if (editor) {
+        editor.container.removeEventListener("keydown", handleKeyDown);
+      }
+    };
+  }, [handleKeyDown]);
+
+  // Handle content change and save (debounce is handled internally by AceEditor)
```
peternied, line 174: Perfect comment to mention de-bouncing :D

frontend/src/components/playground/SaveStatusIndicator.tsx
```diff
@@ -0,0 +1,74 @@
...
+        return (
+          "Saved" +
+          (state.savedAt ? ` ${formatDistanceToNow(state.savedAt)} ago` : "")
+        );
+      case SaveStatus.BLOCKED:
+        return (
+          "Blocked" +
+          (state.errors.length
+            ? ` (${state.errors.length} error${
+                state.errors.length > 1 ? "s" : ""
```
peternied: I'm not sure I agree with sonarqube on this one, I think it could be abused, but this use is fine.  I'd be ok if you wanted to add an except for this rule on all .tsx files

frontend/src/components/playground/TransformationItem.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }, []);
+
+  const handleEditNameChange = (
+    event: CustomEvent<ButtonProps.ClickDetail>,
+  ) => {
+    event.preventDefault();
+    event.stopPropagation();
+
+    if (isEditing) {
+      if (editName.trim()) {
```
peternied, line 47: Nit: what do you think about 'pre-trim'ing the name, leading/trailing whitespace is confusing on identifiers

frontend/src/components/playground/TransformationItem.tsx
```diff
@@ -0,0 +1,149 @@
...
+                    setEditName(item.data.name); // Reset to original
+                    setIsEditing(false);
+                  }
+                }}
+                autoFocus
+              />
+            ) : (
+              <>
+                {item.data.name}
+                <div style={{ marginLeft: "10px" }}>
```
peternied, line 131: Is there a way we can get this formatting using the Cloudscape layout primitives instead of CSS?

peternied: If we've got to push pixels, I'd rather they were only inside the AceEditorComponent, and everything out side it could cleanly use the cloudscape theming.  What do you think?

frontend/src/components/playground/boardI18nStrings.tsx
```diff
@@ -0,0 +1,71 @@
+"use client";
+import { BoardProps } from "@cloudscape-design/board-components";
+import { Transformation } from "@/context/PlaygroundContext";
+
+export const boardI18nStrings: BoardProps.I18nStrings<Transformation> = (() => {
```
peternied, line 5: Nit:
```suggestion
export const transformationBoardLayoutStrings: BoardProps.I18nStrings<Transformation> = (() => {
```

frontend/src/components/playground/boardItemI18nStrings.tsx
```diff
@@ -0,0 +1,9 @@
+"use client";
+export const boardItemI18nStrings = {
```
peternied, line 2: Nit: also can we associate the type for the boardItemStrings?
```suggestion
export const transformationBoardItemStrings = {
```

frontend/src/components/playground/SaveStatusIndicator.tsx
peternied, line 1: Maybe worthwhile for some superficial UX tests for this component.

frontend/src/components/playground/TransformationSection.tsx
peternied, line 1: Moving items around on the board seems like it might be a good candidate for UX adding tests to confirm the behavior.

frontend/src/components/playground/AceEditorComponent.tsx
peternied, line 1: I'm curious if it is easy to see this break as you've been dev'ing on this component, trying to get an understanding of if we should have UX tests to verify callback/intervals being triggered correctly.

peternied (commented):
@mikaylathompson Thanks for getting this out there, did a deep dive on the Ace integration and the react components.   Called out a couple of files where I'd be curious for your thoughts are on testing.

frontend/src/components/playground/AceEditorComponent.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+export default function AceEditorComponent({
+  itemId,
+  mode = "json",
+  formatRef,
+  onSaveStatusChange,
+}: Readonly<AceEditorComponentProps>) {
+  const { state } = usePlayground();
+  const { updateTransformation } = usePlaygroundActions();
+  const [content, setContent] = useState("");
```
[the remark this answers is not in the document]
mikaylathompson: No, it ends up not mattering because we pull the transformation from storage and effectively initialize this variable then (in the `useEffect` below). When we pull from storage, we want to overwrite any data that might be there already.

frontend/src/components/playground/SaveStatusIndicator.tsx
mikaylathompson: Already fixed it, but I don't disagree with you. I was considering it for this and the other one in the file.

frontend/src/components/playground/TransformationItem.tsx
mikaylathompson: This particular check is basically just "is it empty", so I'm not going to change this, but I agree with you that preserving leading/trailing whitespace is a bad idea, so I've trimmed it in the `update` call on the line below.

frontend/src/components/playground/TransformationItem.tsx
mikaylathompson: Yes, probably, I'll play around with it.

frontend/src/components/playground/AceEditorComponent.tsx
mikaylathompson: https://opensearch.atlassian.net/browse/MIGRATIONS-2520

lewijacn (approved):
Bug I faced fixed after updates, thanks!

PR #1490, opened 2025-05-01
adding MSK EBS storage autoscaling policy

cauneacm:
### Description
This change introduces MSK EBS volume autoscaling support in the migration-assistance-stack.ts.
An Application Auto Scaling policy for MSK EBS volume size is conditionally created.
The policy is enabled by setting the enableStorageAutoscaling flag in the storage block of cdk.context.json.
The maximum volume capacity can be configured via maxCapacity in the same block.
The policy targets the kafka:broker-storage:VolumeSize scalable dimension and uses target tracking scaling based on Kafka broker storage utilization

Example:

    "MskEbsStorage":{
      "enableStorageAutoscaling": false,
      "maxCapacity": 16384  // - max value
    },

### Testing
The stack was successfully deployed in the Sandbox account.
EBS Autoscaling for MSK brokers was verified to be enabled.
The Application Auto Scaling policy was created with a target value of 70% for KAFKA_BROKER_STORAGE_UTILIZATION.
The scalable dimension kafka:broker-storage:VolumeSize was applied to the MSK cluster and verified via the Auto Scaling console.

### Check List


deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
```diff
@@ -6,6 +6,7 @@ import {CfnConfiguration} from "aws-cdk-lib/aws-msk";
 import {Cluster} from "aws-cdk-lib/aws-ecs";
 import {LogGroup, RetentionDays} from "aws-cdk-lib/aws-logs";
 import {Bucket, BucketEncryption} from "aws-cdk-lib/aws-s3";
+import * as appscaling from 'aws-cdk-lib/aws-applicationautoscaling';
```
peternied, line 9: Can we switch the use of `*` to follow the explicit import pattern?

deployment/cdk/opensearch-service-migration/cdk.context.json
```diff
@@ -26,6 +26,10 @@
         "serviceSigningName": "es | aoss"
       }
     },
+    "MskEbsStorage":{
```
peternied, line 29: Can you update the tables in options.md in this directory to include the new options?

deployment/cdk/opensearch-service-migration/cdk.context.json
```diff
@@ -26,6 +26,10 @@
...
+      "enableStorageAutoscaling": false,
```
peternied, line 30: What would you think about leaving this enabled by default?  IMO that seems like a better experience in general

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
```diff
@@ -32,14 +33,22 @@ export interface MigrationStackProps extends StackPropsExt {
...
     artifactBucketName: string;
+    clusterArn: string;
+    clusterName: string;
+    bootstrapBrokers: string;
+
 
     createMSKResources(props: MigrationStackProps, streamingSecurityGroup: SecurityGroup) {
+        const storageContext = this.node.tryGetContext('MskEbsStorage') || {};
+        const enableAutoscaling = storageContext.enableStorageAutoscaling === true;
+        const maxCapacity = storageContext.maxCapacity ?? props.maxCapacity ?? 1500;
```
peternied, line 51: Kind of strange that the default value doesn't align with the value in the cdk.context.json that you've added, shouldn't it be the same?

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
```diff
@@ -32,14 +33,22 @@ export interface MigrationStackProps extends StackPropsExt {
     readonly mskBrokersPerAZCount?: number,
     readonly replayerOutputEFSRemovalPolicy?: string
     readonly artifactBucketRemovalPolicy?: string
+    readonly maxCapacity?: number
```
peternied, line 36: Is this used?

peternied (commented):
Thanks for the contribution, a couple of comments.

deployment/cdk/opensearch-service-migration/cdk.context.json
AndreKurait: I think we can remove the ability to disable autoscaling altogether.

The only setting i would make configurable is default ebs size

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
```diff
@@ -83,6 +92,23 @@ export class MigrationAssistanceStack extends Stack {
             removalPolicy: RemovalPolicy.DESTROY
```
AndreKurait, line 92: Let's set an initial scale to handle 50MBps.  When capacity policy is breached, MSK scales up 10% of the configured capacity with a cooldown period of 6 hours. 50MBps is 175GB so our default scale should be 1750GB to catch up.

```
ebsStorageInfo {
   volumeSize: 1750 // initialSize
}
```

AndreKurait, line 92: Here is some calculations and reasoning for scaling values.

When MSK scales when the target value is breached and scales up 10% (or 10GB) of the current capacity with a cooldown period of 6 hours. This means we should set our initial size such that 10% of it is at least enough scale to hold the expected average incoming traffic for 6 hours.

50MBps over 6 hours would result in needing 1750GB cluster to keep up with the scale.

Lets set this in `ebsStorageInfo`

deployment/cdk/opensearch-service-migration/cdk.context.json
cauneacm: MskEbsStorage option was added to the options.md file on the second commit

deployment/cdk/opensearch-service-migration/cdk.context.json
cauneacm: removed the option to disable autoscaling

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
cauneacm: removed the prop as it's superfluous in the context

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
cauneacm: I was thinking of following a more conservative line, but I understand it can cause unnecessary confusion, I updated the default value to be the same as in the cdk.context.json file

deployment/cdk/opensearch-service-migration/lib/migration-assistance-stack.ts
cauneacm: Thank you for the insight, I added the ebsStorageInfo in the MSK cluster values with the 1750 as initial value. Please let me know if you have any comments.

PR #1492, opened 2025-05-02
[Playground] Transformation Execution

mikaylathompson:
### Description
This adds execution to the transformations.
When transformations are saved (dependent on having no validation errors), the chain of transformations is automatically executed on all input documents, and output documents are displayed.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2493

### Testing
Lots of manual testing, and unit tests added.

### Check List


peternied (commented):
Could you provide some sample input documents?  I tried using the following doc and I get back `Error in document 'test-doc.json' at transformation "Transformation 1":Transformation doesn't provide a valid entrypoint`.

#### Input doc
```json
{ "id": "1", "name": "Test", "content": "{}" }
```

frontend/src/components/playground/AceEditorComponent.tsx
```diff
@@ -72,17 +72,18 @@ export default function AceEditorComponent({
     return () => resizeObserver.disconnect();
   }, []);
 
-  // Initialize content from the transformation
+  // Initialize content from the transformation only once
   useEffect(() => {
-    if (transformation) {
-      setContent(transformation.content || defaultContent);
+    // Only initialize content if it hasn't been initialized yet or if the transformation ID changes
```
peternied, line 77: This two comments are in contradiction of one another?

frontend/src/components/playground/AceEditorComponent.tsx
```diff
@@ -43,15 +43,15 @@ export default function AceEditorComponent({
...
+  // Use refs for validation errors to prevent re-renders
   const validationErrorsRef = useRef<IAnnotation[]>([]);
-
-  // Refs and dimension state used for resizing the AceEditor instance
   const editorRef = useRef<AceEditor>(null);
   const containerRef = useRef<HTMLDivElement>(null);
   const [dimensions, setDimensions] = useState({ width: 500, height: 300 });
 
+  // Track if content has been initialized
+  const contentInitializedRef = useRef<boolean>(false);
```
peternied, line 53: This never seems to be set to true anywhere?

frontend/src/components/playground/OutputDocumentSection.tsx
```diff
@@ -2,23 +2,66 @@ import Container from "@cloudscape-design/components/container";
...
 import { usePlayground } from "@/context/PlaygroundContext";
 import { DocumentItemWithPopoverCodeView } from "./DocumentItemWithPopoverCodeView";
+import { ReactNode } from "react";
 
-// Inner component that uses the usePlayground hook
 export default function OutputDocumentSection() {
   const { state } = usePlayground();
+  const { isProcessingTransformations, transformationErrors } = state;
+
+  const renderOutputContent = (): ReactNode => {
```
peternied, line 15: Curious why you choose not to inline this code?

frontend/src/components/playground/OutputDocumentSection.tsx
```diff
@@ -2,23 +2,66 @@ import Container from "@cloudscape-design/components/container";
...
+              {transformationErrors.map((error) => {
+                const inputDoc = state.inputDocuments.find(
+                  (doc) => doc.id === error.documentId,
+                );
+                const transformation = state.transformations.find(
+                  (t) => t.id === error.transformationId,
+                );
+
+                return (
+                  <li key={inputDoc?.id ?? error.documentId}>
```
peternied, line 51: Wouldn't it be easier to always use the docId from the error?

frontend/src/components/playground/OutputDocumentSection.tsx
```diff
@@ -2,23 +2,66 @@ import Container from "@cloudscape-design/components/container";
...
   return (
     <Container header={<Header variant="h3">Output Documents</Header>}>
       <SpaceBetween size="m">
-        {state.outputDocuments.length === 0 ? (
-          <Box>No output documents.</Box>
-        ) : (
-          state.outputDocuments.map((doc) => (
-            <DocumentItemWithPopoverCodeView key={doc.id} document={doc} />
-          ))
+        {transformationErrors.length > 0 && (
```
peternied, line 39: I'm not sure theming is supported on UL/LI can we just put these in their own `<Box>` to make sure they better use horizontal space? 
image

frontend/src/hooks/useTransformationExecutor.ts
```diff
@@ -0,0 +1,80 @@
...
+        } else {
+          // Record error
+          errors.push({
+            documentId: inputDoc.id,
+            transformationId: result.transformationId ?? "",
+            message: result.error ?? "Unknown error",
+          });
+        }
+      } catch (error) {
+        console.warn(`Error processing document ${inputDoc.id}:`, error);
```
peternied, line 56: Lets change this to error, IMO this should never happen, no?

frontend/src/utils/transformationExecutor.ts
```diff
@@ -0,0 +1,101 @@
...
+    };
+  }
+}
+
+// A DocumentTransformer is a function that takes a document (object or string)
+// and returns a transformed document (object or string).
+// We need all transformations to conform to this type.
+export type DocumentTransformer = (
+  document: object | string,
+) => object | string;
```
peternied, line 40: Rather than accept all these combinations of input/output can we expect `object` for both and make the downstream callers conform?

frontend/src/utils/transformationExecutor.ts
```diff
@@ -0,0 +1,101 @@
...
+  try {
+    // Use Function constructor to evaluate the code safely
+    // This can be replaced with isolated-vm or web workers later for better isolation
+    // The way of extracting the entrypoint function is a bit specific to our default transformation format,
+    // this could be broadened.
+    const result = new Function(`
+      ${code}
+      return (document) => {
+        try {
+          const entrypoint = ${code.includes("(() => main)()") ? "(() => main)()" : "null"};
```
peternied, line 53: I might not be reading this correctly, why do we need to look for this exact statement? Shouldn't we try to the execute `code` and then make sure we got back a function that can be used for the transformation?

frontend/__tests__/utils/transformationExecutor.test.ts
```diff
@@ -0,0 +1,154 @@
...
+        },
+      ];
+
+      const document = { value: 42 };
+      const result = await executeTransformationChain(transformations, document);
+
+      expect(result.success).toBe(false);
+      expect(result.error).toBe("Test error");
+      expect(result.transformationId).toBe("2");
+      expect(result.document).toEqual({ value: 42, step1: true }); // Only first transformation applied
```
peternied, line 151: Nit: Also add a negative check for `step3: true`

frontend/src/components/playground/OutputDocumentSection.tsx
mikaylathompson: Because sonarqube complained about the depth 😭 
I can add a comment to that effect

frontend/src/components/playground/OutputDocumentSection.tsx
mikaylathompson: Done -- this is the new appearance

Screenshot 2025-05-14 at 4 24 57 PM

frontend/src/utils/transformationExecutor.ts
mikaylathompson: Okay, after discussion with Andre, I loosened this up. The way our transformations are defined doesn't give us an unambiguous definition of "got back a function". So now it looks for this default entrypoint definition and uses that if it can. If not, it takes the last expression of the block and tries to use that as an entrypoint function. That means you can do other approaches like,
```
  function newMain(context) {
    return (document) => {
      document.transformed = true;
      return document;
    };
  }
  newMain;
  ```
  and it will work correctly.

peternied (approved):
Just made some minor fixes so this can be merged.

PR #1516, opened 2025-05-14
Update Pipfiles across subdirectories

jugal-chauhan:
### Description
This PR updates all Pipfile.lock files in the repository to ensure consistent dependency versions across submodules, specifically aligning them with Python 3.11.

### Issues Resolved
During the Version Bump Release for Migration Assistant 2.3.7 workflow, we encountered an error caused by version mismatches of the six library across multiple Pipfiles. This inconsistency led to unexpected behavior in dependency resolution and required a uniform lock file update.

### Temporary one-time fix to unblock the release has been implemented over here
This PR follows a manual updation as a temporary fix. The following command was run to regenerate all Pipfile.lock files in the TrafficCapture/dockerSolution directory:
`find . -type f -name Pipfile -execdir bash -c 'PIPENV_IGNORE_VIRTUALENVS=true pipenv lock --python 3.11' \;`

### Long Term fix for this issue would be to have a centralized Piplock
- New Task has been drafted : [MIGRATIONS-2533](https://opensearch.atlassian.net/browse/MIGRATIONS-2533)
to prevent this issue from happening again and to enhance the detection of dependency divergence.


peternied (changes requested):
I'm not sure this is a good approach, while this change synchronizes everything today, we don't have a process that ensures we are using the same definition across all projects - we've got dependabot updates and CVE scanners that work by individual Pipfile 

Isn't the solution to this to centralize on a single Pipfile at the shard root directory of all these project so there is only one lock file?

jugal-chauhan:
> Thanks for making this change that starts this discussion
> 
> I'm not sure this is a good approach, while this change synchronizes everything today, we don't have a process that ensures we are using the same definition across all projects - we've got dependabot updates and CVE scanners that work by individual Pipfile
> 
> Isn't the solution to this to centralize on a single Pipfile at the shard root directory of all these project so there is only one lock file?

This change was brought in since a mismatch in 'six' library was spotted acros different PipFiles. 
More information on how this error was detected is [over here](https://github.com/opensearch-project/opensearch-migrations/actions/runs/14983649365/job/42093275982) (in the workflow failure for a version bump release)

Does this mean that the dependabot scanners failed to update it across all files ? Curious to know if a fix for dependabot/CVE scanners is needed.

jugal-chauhan:
@peternied I have created a new JIRA task to tackle a longer term fix to prevent divergence in dependencies. I have updated the PR description to point to the longer term fix which also requires better definition. We can have a detailed discussion under refinement of this task and I am happy to kick off work on creating a new composite workflow. 

If you agree with this Temporary change coming in and unblocking a new release for 2.3.7 of MA, I'd request an approval on this PR.

peternied (approved):

PR #1546, opened 2025-05-28
Add experimental path for EKS solutions deployments

lewijacn:
### Description
This change adds an additional experimental template for our solutions CDK to generate a CFN stack that utilizes an EKS cluster. I expect this to change a good bit as we discover more about how we should configure EKS and what additional resources we actually need there, but should provide developers of this feature a good starting point.

The template can be generated with the following command:
```
cdk synth "Migration-Assistant-Infra-Create-VPC-v3"
```

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2474

### Testing
AWS deployment of template

### Check List

### DIFF
--- a/deployment/migration-assistant-solution/lib/eks-infra.ts
+++ b/deployment/migration-assistant-solution/lib/eks-infra.ts
@@ -1,0 +1,205 @@
+import {Construct} from 'constructs';
+import {CfnCluster, CfnPodIdentityAssociation} from 'aws-cdk-lib/aws-eks';
+import {IVpc, Port, SecurityGroup} from 'aws-cdk-lib/aws-ec2';
+import {
+    CfnPolicy,
+    CfnRole,
+    Effect,
+    ManagedPolicy,
+    PolicyStatement,
+    Role,
+    ServicePrincipal,
+} from "aws-cdk-lib/aws-iam";
+import {RemovalPolicy, Tags} from "aws-cdk-lib";
+import {Repository} from "aws-cdk-lib/aws-ecr";
+
+
+export interface EKSInfraProps {
+    vpc: IVpc;
+    clusterName: string;
+    ecrRepoName: string;
+    stackName: string;
+    namespace?: string;
+    serviceAccountName?: string;
+}
+
+export class EKSInfra extends Construct {
+    public readonly cluster: CfnCluster;
+    public readonly ecrRepo: Repository;
+
+    constructor(scope: Construct, id: string, props: EKSInfraProps) {
+        super(scope, id);
+
+        const namespace = props.namespace ?? 'ma';
+        const serviceAccountName = props.serviceAccountName ?? 'migrations-sa';
+
+        const migrationSecurityGroup = new SecurityGroup(this, 'MigrationsSecurityGroup', {
+            vpc: props.vpc,
+            allowAllOutbound: true,
+            allowAllIpv6Outbound: true,
+        })
+        migrationSecurityGroup.addIngressRule(migrationSecurityGroup, Port.allTraffic());
+
+        this.ecrRepo = new Repository(this, 'MigrationsECRRepository', {
+            repositoryName: props.ecrRepoName,
+            removalPolicy: RemovalPolicy.DESTROY,
+            emptyOnDelete: true
+        });
+
+        const clusterRole = new Role(this, 'MigrationsEKSClusterRole', {
+            assumedBy: new ServicePrincipal('eks.amazonaws.com'),
+            managedPolicies: [
+                ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSClusterPolicy'),
+                ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSBlockStoragePolicy'),
+                ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSComputePolicy'),
+                ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSLoadBalancingPolicy'),
+                ManagedPolicy.fromAwsManagedPolicyName('AmazonEKSNetworkingPolicy'),
+            ],
+        });
+        clusterRole.assumeRolePolicy?.addStatements(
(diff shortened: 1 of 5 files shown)


peternied (changes requested):
Can we create a completely separate cdk project? There is a lot of governance and information passing that I would hope we can completely remove with the EKS deployment - I'd prefer we can cleanly start the new deployment configuration

lewijacn:
> Could we create a completely separate cdk project for this effort? There is a lot of governance and information passing that I would hope we can completely remove with the EKS deployment. By building inside the existing project it makes this harder to detect.

It feels strange to me to make a new package for this as we want our existing solutions setup to continue to use this same structure, and we may want to have a structure in near future that has templates for the old and new pattern. I definitely agree though that once we get to a spot where we can remove the old pattern, we should remove all that logic, and for now make sure we are not putting unneeded things in the new stack. What pieces are most concerning to you?

deployment/migration-assistant-solution/lib/solutions-stack-eks.ts
```diff
@@ -0,0 +1,236 @@
...
+    readonly codeBucket: string;
+    readonly createVPC: boolean;
+    readonly stackNameSuffix?: string;
+}
+
+interface ParameterLabel {
+    default: string;
+}
+
+function applyAppRegistry(stack: Stack, stage: string, infraProps: SolutionsInfrastructureStackEKSProps): string {
```
peternied, line 37: Can we move the shared functionality into a separate file?  This is already duplicated in the other stack.

deployment/migration-assistant-solution/lib/eks-infra.ts
```diff
@@ -0,0 +1,205 @@
...
+                    },
+                ],
+            },
+            managedPolicyArns: [
+                ManagedPolicy.fromAwsManagedPolicyName('AmazonEC2ContainerRegistryFullAccess').managedPolicyArn,
+            ],
+            description: 'Migrations IAM role assumed by pods via EKS Pod Identity',
+        });
+
+        new CfnPolicy(this, 'MigrationsPodIdentityPolicy', {
```
peternied, line 139: Looks like we are using the low level policy resource in CFN, can we switch to `Policy` , `PolicyStatements`, `Role`.  This will let us write them with policy `attachToRole(...)` & `addStatements(...)`

Nit: It would be cleaner if we had separate policy statements that were named in a way that indicated what they are used to access even if it is for our own knowledge.

deployment/migration-assistant-solution/lib/eks-infra.ts
```diff
@@ -0,0 +1,205 @@
...
+                    enabled: true
+                }
+            }
+        });
+        migrationSecurityGroup.addIngressRule(
+            SecurityGroup.fromSecurityGroupId(this, "MigrationsEKSClusterDefaultSG", this.cluster.attrClusterSecurityGroupId),
+            Port.allTraffic()
+        );
+
+        const podIdentityRole = new CfnRole(this, 'MigrationsPodIdentityRole', {
```
peternied, line 116: Switch to Role

peternied (commented):
Thanks for discussing this, my earlier comment wasn't accurate, I was thinking this was part of the other cdk stack instead of the solutions stack.  Some comments mostly using higher level cfn objects

deployment/migration-assistant-solution/lib/eks-infra.ts
```diff
@@ -0,0 +1,205 @@
...
+            ],
+        });
+
+        const subnetIds = []
+        for (const subnet of props.vpc.privateSubnets) {
+            Tags.of(subnet).add(`kubernetes.io/cluster/${props.clusterName}`, 'shared');
+            Tags.of(subnet).add('kubernetes.io/role/internal-elb', '1');
+            subnetIds.push(subnet.subnetId)
+        }
+        this.cluster = new CfnCluster(this, 'MigrationsEKSCluster', {
```
peternied, line 82: Can we make sure that we are using the upgrade policy is standard and not extended?

image

lewijacn: I don't believe this is something we directly control, rather the version of EKS you use determines this policy

deployment/migration-assistant-solution/lib/solutions-stack-eks.ts
lewijacn: Added shared functionality to separate class

deployment/migration-assistant-solution/lib/eks-infra.ts
lewijacn: I take it back, it looks like the lower level construct does have support for this: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-eks-cluster.html#cfn-eks-cluster-upgradepolicy
```
    interface UpgradePolicyProperty {
        /**
         * If the cluster is set to `EXTENDED` , it will enter extended support at the end of standard support.
         *
         * If the cluster is set to `STANDARD` , it will be automatically upgraded at the end of standard support.
         *
         * [Learn more about EKS Extended Support in the *Amazon EKS User Guide* .](https://docs.aws.amazon.com/eks/latest/userguide/extended-support-control.html)
         *
         * @see http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-eks-cluster-upgradepolicy.html#cfn-eks-cluster-upgradepolicy-supporttype
         */
        readonly supportType?: string;
    }
```

deployment/migration-assistant-solution/lib/eks-infra.ts
```diff
@@ -80,6 +80,9 @@ export class EKSInfra extends Construct {
         this.cluster = new CfnCluster(this, 'MigrationsEKSCluster', {
             name: props.clusterName,
             version: '1.32',
+            upgradePolicy: {
```
peternied, line 83: ❤️

PR #1550, opened 2025-05-28
[Transformations] Add a 'template' for metadata updater function

peternied:
### Description
Adding a JavaScript transformation example that can be iterated on by customer to handle mapping type conversions.  So long as this looks like a good foundation I'll adapt it to a simplified version that I'll post on the documentation website.

Can be used with string -> text, flattened -> flat_object, constant_keyword -> keyword fields as quick examples.

I'd expect the version that is posted into the documentation site to look something like the following:

```javascript
function main(context) {
  /* Define rules here */
  const rules = [{
    when: { type: "string" }, /* Match all the key/value pairs here */
    set: { type: "text" }, /* Replace them with these key/value pairs */
  }, {
    when: { type: "flattened" }, /* Match all the key/value pairs here */
    set: { type: "flat_object" }, /* Replace them with these key/value pairs */
    remove: ["index"], /* Any sibling keys that should be removed */
  }];

  function applyRulesToMap(when, set, remove, map) {
    const matches = Object.entries(when).every(([k, v]) => map.get(k) === v);
    if (matches) {
      Object.entries(set).every(([k, v]) => map.set(k, v));
      remove.forEach((key) => map.delete(key));
    }
  }

  function applyRulesToObject(when, set, remove, obj) {
    const matches = Object.entries(when).every(([k, v]) => obj[k] === v);
    if (matches) {
      Object.assign(obj, set);
      remove.forEach((key) => delete obj[key]);
    }
  }

  function applyRules(node, rules) {
    if (Array.isArray(node)) {
      node.forEach((child) => applyRules(child, rules));
    } else if (node instanceof Map) {
      for (const { when, set, remove = [] } of rules) {
        applyRulesToMap(when, set, remove, node);
      }
      // recurse
      for (const child of node.values()) {
        applyRules(child, rules);
      }
    } else if (node && typeof node === "object") {
      for (const { when, set, remove = [] } of rules) {
        applyRulesToObject(when, set, remove, node);
      }
      // recurse
      Object.values(node).forEach((child) => applyRules(child, rules));
    }
  }

  return (doc) => {
    if (doc && doc.type && doc.name && doc.body) {
      applyRules(doc.body, context.rules);
    }
    return doc;
  };
}
(() => main)();
```

### Issues Resolved
- Resolves

### Testing
Added new JavaScript tests that verify the new 

### Check List
    https://github.com/opensearch-project/documentation-website/pull/10010

### DIFF
--- a/transformation/standardJavascriptTransforms/src/metadataUpdater.js
+++ b/transformation/standardJavascriptTransforms/src/metadataUpdater.js
@@ -1,0 +1,38 @@
+function applyRules(node, rules) {
+  if (Array.isArray(node)) {
+    node.forEach((child) => applyRules(child, rules));
+  } else if (node && typeof node === "object") {
+    for (const { when, set, remove = [] } of rules) {
+      const matches = Object.entries(when).every(([k, v]) => node[k] === v);
+      if (matches) {
+        // apply the rule
+        Object.assign(node, set);
+        remove.forEach((key) => delete node[key]);
+      }
+    }
+    // recurse
+    Object.values(node).forEach((child) => applyRules(child, rules));
+  }
+}
+
+function main(context) {
+  if (!context.rules) {
+    throw Error(
+      "Expected rules to be defined in the context.  Example: " +
+        JSON.stringify(
+          { rules: [{ when: { type: "foo" }, set: { type: "bar" } }] },
+          null,
+          2
+        )
+    );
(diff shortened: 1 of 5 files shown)


AndreKurait (approved):
Looks good! Thanks for the updates to pull in a script from a file!

PR #1558, opened 2025-06-02
Adding ES 7.1 and 7.4 as extended supported sources

jugal-chauhan:
### Description
This PR extends support for Elasticsearch 7.1 and 7.4 as source clusters for document migration via snapshot, using the existing SnapshotReader for ES 6.8. These versions are now officially supported as extended sources when migrating to OpenSearch (target OS 3.x).

Code changes include:
- SupportedClusters.extendedSources() now includes ES 7.1 and ES 7.4
- E2E backfill test and metadata test now test snapshot-based document migration from these versions to OS 3.x
- Introduced new matcher VersionMatchers.isES_7_0_to_7_9
- Updated SnapshotReader_ES_6_8 to treat ES 7.0–7.9 as strictly compatible
- SnapshotReader_ES_7_10 remains the primary reader for ES 7.10+ and OS clusters

### Issues Resolved
[MIGRATIONS-2550](https://opensearch.atlassian.net/browse/MIGRATIONS-2550)

### Testing
- `./gradlew :MetadataMigration:isolatedTest --tests "*EndToEndTest"`
- `./gradlew :DocumentsFromSnapshotMigration:isolatedTest --tests "*EndToEndTest"`

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
@@ -53,6 +53,21 @@ public class EndToEndTest extends SourceTestBase {
         }
     }
 
+    private static Stream<Arguments> extendedScenarios() {
+        return SupportedClusters.extendedSources().stream().map(s -> Arguments.of(s));
+    }
+    @ParameterizedTest(name = "Source {0} to Target OS 3.0")
+    @MethodSource(value = "extendedScenarios")
+    public void extendedMigrationDocuments(
+            final SearchClusterContainer.ContainerVersion sourceVersion) {
+        try (
+                final var sourceCluster = new SearchClusterContainer(sourceVersion);
+                final var targetCluster = new SearchClusterContainer(SearchClusterContainer.OS_V3_0_0)
+        ) {
+            migrationDocumentsWithClusters(sourceCluster, targetCluster);
+        }
+    }
+
     @SneakyThrows
     private void migrationDocumentsWithClusters(
         final SearchClusterContainer sourceCluster,
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
@@ -154,11 +154,6 @@ public class EndToEndTest extends SourceTestBase {
                     "[{\"NoopTransformerProvider\":{}}]" // skip transformations including doc type removal
                     : null;
 
-            // Add --allow-loose-version-matching flag for loose compatibility
-            if (VersionMatchers.isES_7_0_to_7_9.test(sourceVersion)) {
-                testDocMigrationContext.allowLooseVersionMatches = true;
-            }
-
             // ExpectedMigrationWorkTerminationException is thrown on completion.
             var expectedTerminationException = waitForRfsCompletion(() -> migrateDocumentsSequentially(
                     sourceRepo,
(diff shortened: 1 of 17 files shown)


RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_6_8/SnapshotReader_ES_6_8.java
```diff
@@ -25,7 +25,7 @@ public boolean compatibleWith(Version version) {
     @Override
     public boolean looseCompatibleWith(Version version) {
         return UnboundVersionMatchers.isBelowES_6_X
-            .or(VersionMatchers.isES_6_X)
```
jugal-chauhan, line 28: Removed `VersionMatchers.isES_6_X` from LooseCompatible as it is present already in CompatibleWith method above (strictly compatible)

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/SupportedClusters.java
```diff
@@ -24,6 +24,13 @@ private static List<ContainerVersion> sources() {
         );
     }
 
+    public static List<ContainerVersion> extendedSources() {
```
peternied, line 27: Could you add overview comment for why some sources should be on this list versus the one above?

transformation/src/main/java/org/opensearch/migrations/VersionMatchers.java
```diff
@@ -12,6 +12,12 @@ public class VersionMatchers {
     public static final Predicate<Version> isES_7_X = VersionMatchers.matchesMajorVersion(Version.fromString("ES 7.10"));
     public static final Predicate<Version> isES_7_10 = VersionMatchers.matchesMinorVersion(Version.fromString("ES 7.10.2"));
     public static final Predicate<Version> isES_8_X = VersionMatchers.matchesMajorVersion(Version.fromString("ES 8.17"));
+    public static final Predicate<Version> isES_7_0_to_7_9 =
+            v -> v != null
```
peternied, line 16: Lets use or create a helper function instead of doing this inline like is done with the other matchers

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_6_8/SnapshotReader_ES_6_8.java
peternied: This doesn't seem correct, `compatibleWith` needs to be updated to include ES 7.1 and ES 7.4.  looseCompatibleWith is checked seperately so by removing the isES_6_X check it will cause some new scenarios to fail.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -139,6 +154,11 @@ private void migrationDocumentsWithClusters(
                     "[{\"NoopTransformerProvider\":{}}]" // skip transformations including doc type removal
                     : null;
 
+            // Add --allow-loose-version-matching flag for loose compatibility
+            if (VersionMatchers.isES_7_0_to_7_9.test(sourceVersion)) {
+                testDocMigrationContext.allowLooseVersionMatches = true;
+            }
```
peternied, line 160: This parameters forces customers to do a parameter update/redeploy and we know for both of these versions it will work.  Since we are now testing this versions and know they work lets not require the flag for these cases.

MetadataMigration/src/test/java/org/opensearch/migrations/EndToEndTest.java
```diff
@@ -150,10 +174,11 @@ private void metadataCommandOnClusters(TransferMedium medium,
                 throw new RuntimeException("Invalid Option");
         }
 
-        // If the target is not part of  supported target matrix enable loose version matching
-        if (!(SupportedClusters.supportedTargets(false)
-            .stream()
-            .anyMatch(v -> v.equals(targetCluster.getContainerVersion().getVersion())))) {
+
+        // If target is ES and the target is not part of  supported target matrix enable loose version matching
```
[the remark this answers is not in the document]
jugal-chauhan: Acknowledged after our discussion offline. Will revert with cleaner changes.

PR #1566, opened 2025-06-06
Fix TrafficReplayer behavior with HEAD and CONNECT method requests.

AndreKurait:
### Description
TrafficReplayer's HttpByteBufFormatter, used for tuples, decodes the request and response in separate Embedded Channels. This limits the ability for the decoder to detect rfc9112 6.3.1-2 behavior where responses from HEAD and CONNECT requests are terminated after the headers ignoring content-length fields.

We now achieve correct behavior for these requests by writing a LastHttpContent.EMPTY_LAST_CONTENT in the channel if the output was empty after the entire request/response was sent. This triggers the end of the message and output is written.

This change converts from using HttpClientCodec to HttpResponseDecoder for initial response decoding and HttpServerCodec to HttpRequestDecoder for request decoding to simplify the decoding and make clearer the responsibilities. HttpClientCodec misleadingly handled this case for HEAD and CONNECT requests as it seeks to maintain state for these responses, but to do so, it requires the request be sent through the outbound of the channel. Doing this would add unnecessary complexity to couple the request and response decoding for the HttpByteBufFormatter especially in cases of multiple responses during retrys where we would need to decode a single request several times to maintain this state for each of the response parsing.

### Issues Resolved
[MIGRATIONS-2560](https://opensearch.atlassian.net/browse/MIGRATIONS-2560)

### Testing
Unit testing

### Check List

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/HttpByteBufFormatter.java
@@ -258,8 +258,8 @@ public class HttpByteBufFormatter {
         try {
             byteBufStream.forEachOrdered(b -> channel.writeInbound(b.retainedDuplicate()));
             T output = channel.readInbound();
-            if (output == null) {
-                log.debug("HTTP message was not processed after decoding all bytes." +
+            if (output == null && HttpMessageType.RESPONSE.equals(msgType)) {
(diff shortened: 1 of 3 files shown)


gregschohn:
For posterity, the HTTP RFS says this on HEAD responses...
> Any response to a HEAD request and any response with a 1xx (Informational), 204 (No Content), or 304 (Not Modified) status code is always terminated by the first empty line after the header fields, regardless of the header fields present in the message, and thus cannot contain a message body or trailer section.

gregschohn (approved):

PR #1432, opened 2025-04-08
vpc subnet selection support

jugal-chauhan:
### Description
This PR adds support for users to specify both a VPC ID and specific subnet IDs when deploying Migration Assistant into an existing VPC. Previously, when users provided a VPC ID, all subnets in that VPC were used automatically. This enhancement gives users more control over subnet selection.

### Issues Resolved
- Resolves #1430
- Resolves [MIGRATIONS-2472](https://opensearch.atlassian.net/browse/MIGRATIONS-2472) 

### Check List
- [x] New functionality includes testing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1588, opened 2025-06-16
Containerize and deploy the website

peternied:
### Description
As part of the build process, bundles up the website and installed it on a small nginx image.  This is then deployed via either our k8 stack or as the ECS task for the MA console.  Includes basic guide for remote access, will build this up when the feature is ready for general availability.

### Issues Resolved
- Resolves [MIGRATIONS-2540 | Deploy website on MA Console](https://opensearch.atlassian.net/browse/MIGRATIONS-2540)

### Testing
Verified the running container image, running inside of k8, and deployed in AWS.

### Check List


deployment/cdk/opensearch-service-migration/lib/service-stacks/migration-console-stack.ts
```diff
@@ -330,6 +330,7 @@ export class MigrationConsoleStack extends MigrationServiceCore {
         }
 
         this.createService({
+            websiteEnabled: true,
```
AndreKurait, line 333: Can this be configurable with default false. Http api could be a security issue on ecs especially if deployed in public subnet

peternied: I don't see how that could be an issue, the SG rules to do not allow ingress in.  We don't support public subnets only. With the BYOVPC option we require 2 private subnets.

AndreKurait: IIRC https://github.com/opensearch-project/opensearch-migrations/pull/1432 it allows the user to specify public subnets. You are right that the SG rules should block ingress from external actors. I'll defer to the team on launching this as the default without a security review.

deployment/k8s/charts/components/migrationConsole/templates/deployment.yaml
```diff
@@ -27,6 +27,11 @@ spec:
            "MountName" "merged-config"
            "include" .Template.Include) .) | nindent 8 }}
       containers:
+        - name: website
+          image: migrations/website:latest
+          imagePullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
peternied: I'm not sure if there is any action I can take here without attempting to update how this is broadly handled, is there a jira ticket to track this problem, if not could you create one?  Feels out of scope to address it with this PR.

PR #1602, opened 2025-06-25
Extended support for older ES 5.x minor versions

jugal-chauhan:
### Description
This PR adds support for ES 5.4, ES 5.3, ES 5.2, ES 5.1 and ES 5.0.
This makes our tool MA utilize a newly created `SnapshotReader_ES_5_4` and supports these versions as Source Cluster to migrating from.

Code changes to address broken changes in these older versions : 

1. a new SnapshotReader_ES_5_4
- Unlike ES 5.5+, where the top-level snapshots field is a `List<String>`, ES 5.4 and older use a list of objects, each containing snapshot metadata. We now define `RawSnapshot` and deserialize as `List<RawSnapshot>`.
- In ES 5.4 and below, `indices/INDEX_ID/index-N` stores snapshot names only (not UUIDs). However, logic in `SnapshotReader_ES_6_8` assumes that per-index metadata will reference snapshot UUIDs.
- Due to this `SnapshotReader_ES_6_8` incorrectly tries to validate index snapshot membership using UUIDs, causing failures. This new reader fixes this by matching snapshot names instead, as recorded in index-N files.

Note : To avoid duplication, tested behavior for metadata deserialization, newly created `SnapshotReader_ES_5_4` delegates the following methods to `SnapshotReader_ES_6_8`:
- GlobalMetadata.Factory getGlobalMetadata()
- IndexMetadata.Factory getIndexMetadata()
- ShardMetadata.Factory getShardMetadata()
because the deserialization logic for reading global metadata from the snapshot files is same as done for ES 5.5-ES 6.8.
This helps us as the `GlobalMetadataFactory_ES_6_8` simply reads and parses the `index-<n>` and `snap-<uuid>.dat` JSON files. It does not interpret or transform the internal structure of mappings. The actual logic that handles the mapping structure differences is the Transformer class.

Although the new snapshot reader resolved snapshot parsing for all ES 5.x versions, ES 5.4 specifically encountered a unique failure due to which new transformer was created.

2. a new Transformer_ES_5_4_to_OS_2_19 
- This is due to ES 5.4 preserving richer type mappings `"copy_to"` under multiple mapping types, which causes IndexMappingTypeRemoval to throw on strict equality mismatch.
- This transformer extends `Transformer_ES_6_8_to_OS_2_11` and replaces the default one for ES 5.4 and older via `Transformer_ES_5_4_to_OS_2_19`

3. a new IndexMappingTypeRemovalWithMergedSupport (to resolve ES 5.4 issues)
- extends `IndexMappingTypeRemoval`
- relaxes its conflict detection logic
- performs safe merging of properties like `"copy_to"`, when other field attributes of same type are compatible

### Issues Resolved
- [MIGRATIONS-2552](https://opensearch.atlassian.net/browse/MIGRATIONS-2552)
- [MIGRATIONS-2553](https://opensearch.atlassian.net/browse/MIGRATIONS-2553)
- Additionally also added ES 5.4, ES 5.2, ES 5.0 as extended supported sources

### Testing
- `./gradlew :MetadataMigration:isolatedTest --tests "*EndToEndTest"`
- `./gradlew :DocumentsFromSnapshotMigration:isolatedTest --tests "*EndToEndTest"`

### Check List


transformation/src/main/java/org/opensearch/migrations/VersionMatchers.java
```diff
@@ -68,4 +70,12 @@ static Predicate<Version> equalOrLessThanMinorVersion(final Version version) {
             .and(other2 -> version.getMinor() >= other2.getMinor())
             .test(other);
     }
+
+    static Predicate<Version> isBetweenInclusive(Version lower, Version upper) {
+        return other ->
+                other.getFlavor().equals(lower.getFlavor())
+                        && other.getMajor() == lower.getMajor()
```
AndreKurait, line 77: This should handle where lower major version != upper major version.

transformation/src/main/java/org/opensearch/migrations/VersionMatchers.java
```diff
@@ -68,4 +70,12 @@ static Predicate<Version> equalOrLessThanMinorVersion(final Version version) {
             .and(other2 -> version.getMinor() >= other2.getMinor())
             .test(other);
     }
+
+    static Predicate<Version> isBetweenInclusive(Version lower, Version upper) {
```
AndreKurait, line 74: This only works when the Flavor of lower and upper are the same, can we add an assertion here for that?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_6_8/SnapshotReader_ES_6_8.java
```diff
@@ -17,15 +16,15 @@ public class SnapshotReader_ES_6_8 implements ClusterSnapshotReader {
...
+        return VersionMatchers.equalOrBetween_ES_5_5_and_5_6
+            .or(VersionMatchers.isES_6_X)
             .or(VersionMatchers.equalOrBetween_ES_7_0_and_7_8)
             .test(version);
     }
 
     @Override
     public boolean looseCompatibleWith(Version version) {
-        return UnboundVersionMatchers.isBelowES_6_X
+        return VersionMatchers.equalOrBetween_ES_5_5_and_5_6
```
AndreKurait, line 27: Should this be 5.5 and above? If there was a unknown `5.7` version we'd want to use this right?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotReader_ES_5_4.java
```diff
@@ -0,0 +1,90 @@
...
+
+    @Override
+    public ClusterSnapshotReader initialize(Version version) {
+        this.version = version;
+        return this;
+    }
+
+    @Override
+    public GlobalMetadata.Factory getGlobalMetadata() {
+        return new GlobalMetadataFactory_ES_6_8(getSnapshotRepo());
```
AndreKurait, line 45: I'm confused by this alongside your overview comment
>Instead of merged mappings, we have per-type mappings nested under { "mappings": { "doc": { "properties": { ... } } } } whereas ES 5.5 and ES 5.6 follow { "mappings": { "properties": { ... } } }
If they are different, how could it work to have the same global metadata factory?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotReader_ES_5_4.java
```diff
@@ -0,0 +1,90 @@
+package org.opensearch.migrations.bulkload.version_es_5_4;
+
+import org.opensearch.migrations.UnboundVersionMatchers;
+import org.opensearch.migrations.Version;
+import org.opensearch.migrations.VersionMatchers;
+import org.opensearch.migrations.bulkload.common.SnapshotRepo;
+import org.opensearch.migrations.bulkload.common.SourceRepo;
+import org.opensearch.migrations.bulkload.models.*;
```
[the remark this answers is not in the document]
AndreKurait: Update your IDE to stop doing this automatically

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_6_8/SnapshotReader_ES_6_8.java
AndreKurait: Any reason this shouldn't just be `return compatibleWith(version)`?

jugal-chauhan: Unlikely about ES 5.7. 

By removing `UnboundVersionMatchers.isBelowES_6_X`, we are not breaking the SnapshotReader_ES_6_8. We are simply limiting it to start invoking from ES 5.5.0 onwards.

I have added the "Unbound" version matcher `isBelowES_5_X` in the loose compatibility of new Snaphsot Reader ES_5_4. This means any if source cluster is untested ES version older than ES 5.0, it will utilise the SnapshotReader_ES_5_4 (for example ES 2.x)

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotReader_ES_5_4.java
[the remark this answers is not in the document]
jugal-chauhan: This has been corrected

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotReader_ES_5_4.java
jugal-chauhan: Great question. I will edit the overview appropriately. I am brining in a refactor to place this logic in a different transformer class.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -23,14 +23,22 @@ public class Transformer_ES_6_8_to_OS_2_11 implements Transformer {
 
     private final int awarenessAttributes;
 
-    public Transformer_ES_6_8_to_OS_2_11(int awarenessAttributes, MetadataTransformerParams params) {
+    public Transformer_ES_6_8_to_OS_2_11(
```
jugal-chauhan, line 26: New constructor has been created to leverage this class in `RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_5_4_to_OS_2_19.java`

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -23,14 +23,24 @@ public class Transformer_ES_6_8_to_OS_2_11 implements Transformer {
 
     private final int awarenessAttributes;
 
-    public Transformer_ES_6_8_to_OS_2_11(int awarenessAttributes, MetadataTransformerParams params) {
+    // Note: `params` is unused here but retained to match subclass constructor signatures
+    @SuppressWarnings("unused")
+    public Transformer_ES_6_8_to_OS_2_11(
```
AndreKurait, line 28: I'd make this constructor `Protected` since we don't intend it to be used except for inheriting superclasses

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotRepoData_ES_5_4.java
```diff
@@ -0,0 +1,127 @@
...
+
+        @JsonIgnore
+        public List<String> getSnapshots() {
+            if (rawSnapshots == null) return List.of();
+            return rawSnapshots.stream()
+                    .map(o -> {
+                        if (o instanceof String) {
+                            return (String) o;
+                        } else if (o instanceof RawSnapshot) {
+                            return ((RawSnapshot) o).getName();
```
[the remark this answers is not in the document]
jugal-chauhan: For ES 5.0-5.4, it is always a MAP

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
jugal-chauhan: Done

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -23,14 +23,24 @@ public class Transformer_ES_6_8_to_OS_2_11 implements Transformer {
...
+            int awarenessAttributes,
+            MetadataTransformerParams params,
```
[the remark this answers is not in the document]
jugal-chauhan: Now we have a SonarQube failure due to unused `ignoreParams`

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotRepoData_ES_5_4.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    @Getter
+    @AllArgsConstructor
+    @NoArgsConstructor
+    public static class RawIndex {
+        private String id;
+
+        @JsonProperty("snapshots")
+        private List<Object> rawSnapshots;
```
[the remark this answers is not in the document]
jugal-chauhan: Done

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
```diff
@@ -23,14 +23,23 @@ public class Transformer_ES_6_8_to_OS_2_11 implements Transformer {
 
     private final int awarenessAttributes;
 
-    public Transformer_ES_6_8_to_OS_2_11(int awarenessAttributes, MetadataTransformerParams params) {
+    // Note: `ignoredParams` is retained to match subclass constructor signatures
+    protected Transformer_ES_6_8_to_OS_2_11(
+            int awarenessAttributes,
+            MetadataTransformerParams ignoredParams,
```
AndreKurait, line 29: You should just be able to remove this parameter

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
```diff
@@ -137,6 +137,68 @@ private static Map<String, String> overrideAndRemoveEnv(
...
     );
+    public static final ContainerVersion ES_V5_4 = new ElasticsearchVersion(
+            "docker.elastic.co/elasticsearch/elasticsearch:5.4.2",
+            Version.fromString("ES 5.4.2")
+    );
+    public static final ContainerVersion ES_V5_3 = new Elasticsearch5Version(
+        "elasticsearch:5.3.2",
+        Version.fromString("ES 5.3.2"),
+        "/usr/share/elasticsearch/config/elasticsearch.yml",
+        String.join("\n",
```
peternied, line 148: Please refactor these duplicate settings following DRY principles https://en.wikipedia.org/wiki/Don%27t_repeat_yourself

peternied: Seems like there is alot of unneeded / redundant details in this file.  Please do a pass over each part to see if it is needed - less lines of code is easier to maintain and improve on in the future.

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformerMapper.java
```diff
@@ -58,7 +68,13 @@ private Transformer strictTransformerMapping(int awarenessAttributes, MetadataTr
...
     private Transformer looseTransformerMapping(int awarenessAttributes, MetadataTransformerParams metadataTransformerParams) {
         if (UnboundVersionMatchers.anyOS.or(UnboundVersionMatchers.isGreaterOrEqualES_6_X).test(targetVersion)) {
-            if (UnboundVersionMatchers.isBelowES_6_X.test(sourceVersion)) {
+            if (UnboundVersionMatchers.isBelowES_5_X.test(sourceVersion)) {
+                return new Transformer_ES_5_4_to_OS_2_19(awarenessAttributes, metadataTransformerParams);
+            }
+            if (VersionMatchers.equalOrBetween_ES_5_0_and_5_4.test(sourceVersion)) {
+                return new Transformer_ES_5_4_to_OS_2_19(awarenessAttributes, metadataTransformerParams);
+            }
+            if (VersionMatchers.equalOrBetween_ES_5_5_and_5_6.test(sourceVersion)) {
```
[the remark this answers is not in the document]
peternied: This is about good file hygine - it is not that I think we need to support ES 5.7, but its that we should have predicable outcome if a user comes in with any ES or OS version from 0.0 -> 9999.9999.  ES 5.11 exists in a gap in that mapping which could cause undefined behavoir.

transformation/src/main/java/org/opensearch/migrations/VersionMatchers.java
```diff
@@ -8,6 +8,10 @@
 public class VersionMatchers {
     public static final Predicate<Version> isES_2_X = VersionMatchers.matchesMajorVersion(Version.fromString("ES 2.4"));
     public static final Predicate<Version> isES_5_X = VersionMatchers.matchesMajorVersion(Version.fromString("ES 5.6"));
+    public static final Predicate<Version> equalOrBetween_ES_5_0_and_5_4 = VersionMatchers.equalOrGreaterThanMinorVersion(Version.fromString("ES 5.0.0"))
```
peternied, line 11: Please do not use patch version `99` that expects that a version will never have more than 100 patch release which isn't true - it can be an unbound number of patches (Even if that would be insane).  Just list these versions as the 2 digit code, e.g. `5.4`

transformation/src/main/java/org/opensearch/migrations/VersionMatchers.java
```diff
@@ -58,14 +62,20 @@ static Predicate<Version> matchesMinorVersion(final Version version) {
     }
 
     static Predicate<Version> equalOrGreaterThanMinorVersion(final Version version) {
-        return other -> matchesMajorVersion(version)
-            .and(other2 -> version.getMinor() <= other2.getMinor())
-            .test(other);
+        return other -> {
+            if (other == null || version == null) return false;
+            if (!matchesFlavor(version).test(other)) return false;
```
peternied, line 67: Lets not reimplement matches major version in this code, I don't understand what wasn't working with its initial implementation?

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/Transformer_ES_6_8_to_OS_2_11.java
jugal-chauhan: I’ll remove it from the signature and let the protected constructor only accept the transformation rule lists directly. That avoids confusion about params being relevant in the base class.

transformation/src/main/java/org/opensearch/migrations/VersionMatchers.java
jugal-chauhan: Reverted

RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
```diff
@@ -40,6 +42,31 @@ private static Map<String, String> overrideAndRemoveEnv(
         return Collections.unmodifiableMap(merged);
     }
 
+    private static final List<String> ES_5_COMMON_CONFIG_LINES = List.of(
+        "network.host: 0.0.0.0",
+        "http.port: 9200",
+        "transport.tcp.port: 9300",
+        "discovery.zen.ping.unicast.hosts: []",
+        "discovery.zen.minimum_master_nodes: 1",
+        "node.max_local_storage_nodes: 2",
```
peternied, line 51: These settings were not needed when running ES 2.4 or ES 5.5, can you check each individual setting to see if it is needed?

jugal-chauhan: I have declared these common settings due to the change in the way of initializing these containers. I am later using `ES_5_0_AND_5_1_CONFIG_YML` and `ES_5_2_AND_5_3_CONFIG_YML` for the older ES 5.x using the same common List of Strings

For ES 5.5, it follows and accepts the class `ElasticsearchVersion` like ES 5.6/

For ES 2.4, it follows and accepts the class `OlderElasticsearchVersion` as few of these settings are not acceptable.

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotRepoProvider_ES_5_4.java
```diff
@@ -0,0 +1,91 @@
...
+    public List<SnapshotRepo.Index> getIndicesInSnapshot(String snapshotName) {
+        List<SnapshotRepo.Index> matchedIndices = new ArrayList<>();
+
+        var targetSnapshot = getRepoData().getSnapshots().stream()
+                .filter(snapshot -> snapshotName.equals(snapshot.getName()))
+                .findFirst()
+                .orElse(null);
+
+        if (targetSnapshot == null) {
+            log.warn("No snapshot found with name [{}]", snapshotName);
```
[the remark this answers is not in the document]
jugal-chauhan: Added an exception

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_5_4/SnapshotRepoProvider_ES_5_4.java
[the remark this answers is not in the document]
jugal-chauhan: Done

RFS/src/main/java/org/opensearch/migrations/cluster/ClusterProviderRegistry.java
```diff
@@ -57,7 +59,7 @@ public ClusterSnapshotReader getSnapshotReader(Version version, SourceRepo repo,
             });
 
         snapshotProvider.initialize(repo);
-        log.info("Found snapshot resource reader for version: " + version);
+        log.info("Found snapshot resource reader {} for version: {}", snapshotProvider.getClass().getSimpleName(), version);
```
AndreKurait, line 62: Update to log.atInfo(...

Issue #1597, opened 2025-06-24
Use self-signed ssl


dnull88:
We use self-signed certificate in opensearch configuration. When I try connect throw console - receive error:
`
> console clusters connection-check
...
TARGET CLUSTER
ConnectionResult(connection_message="Unable to connect to cluster with error: HTTPSConnectionPool(host='opensearch-qa.**-dev.int', port=9200): Max retries exceeded with url: / (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1016)')))", connection_established=False, cluster_version=None)`

How I can add my ca.crt to stack?

sumobrian:
This option is not currently configurable through our CDK interface. As a temporary workaround, please modify /config/migration_services.yaml. Under the source and/or target clusters (depending on your source and cluster certification configuration), add the following key:

```
target_cluster:
    allow_insecure: true
```

Once you save that change, run `console clusters connection-check` to verify the fix.

sumobrian:
Being tracked internally with ticket: https://opensearch.atlassian.net/browse/MIGRATIONS-1977

dnull88:
ok, what about backfill process? It doesnt work, in reindex-from-snapshot-logs:
`javax.net.ssl.SSLHandshakeException: PKIX path building failed: sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target`

gregschohn:
Fixed in https://github.com/opensearch-project/opensearch-migrations/pull/1603.

Once 2.4.2 is released (see [here](https://github.com/opensearch-project/opensearch-migrations/releases/tag/2.4.2)), you'll be able to deploy the CDK so that your containers/processes can connect to hosts with self-signed certs.  Redeploying with a cdk.context.json that does NOT include `{ "allow_insecure": false }` for the target cluster (or source cluster), will cause the migration_console's /config/migration_services.yaml to include allow_insecure: true AND for RFS, the capture proxy, and replayer all to include command line flags to ignore if root certs aren't signed by a trusted entity.

dnull88:
great! but in aws cloudformation templates the last version of assistant is 2.3.0. Is it possible use last version with CF template?

lewijacn:
[...]

PR #1603, opened 2025-06-25
Propagate the allow_insecure flag from the cluster definition in the cdk.context.json file into the console's migration_services.yaml file.

gregschohn:
### Description
Propagate the allow_insecure flag from the cluster definition in the cdk.context.json file into the console's migration_services.yaml file.

### Issues Resolved
https://github.com/opensearch-project/opensearch-migrations/issues/1597
https://opensearch.atlassian.net/browse/MIGRATIONS-1977

### Testing
Manually tested.
Added `"allow_insecure": true,` to the source_cluster definition of the cdk.context.json file. and ran the following commands...

```
% cdk deploy "*" --c contextId=default  --concurrency 5
 % ./accessContainer.sh migration-console $STAGE $REGION
...
 The Session Manager plugin was installed successfully. Use the AWS CLI to start a session.


Starting session with SessionId: ecs-execute-command-3tfbtqc58n5sh534loyh427xie

Welcome to the Migration Assistant Console
(17:30:11) migration-console (~) -> cat /config/migration_services.yaml
source_cluster:
  endpoint: http://target.example.com:80
  version: ES_7.10
  allow_insecure: false
  no_auth: ""
...
```

### Check List
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.  Documentation is inline within the cdk.context.json.  I'm not sure that this change warrants a larger documentation change.


deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -467,12 +467,13 @@ export function parseClusterDefinition(json: any): ClusterYaml {
         throw new Error('Missing required field in cluster definition: endpoint')
     }
     endpoint = validateAndReturnFormattedHttpURL(endpoint)
+    let allowInsecure = json.allow_insecure;
```
lewijacn, line 470: Check that we should use `const` here seems valid

deployment/cdk/opensearch-service-migration/cdk.context.json
```diff
@@ -15,6 +15,8 @@
     },
     "sourceCluster": {
       "endpoint": "<SOURCE_CLUSTER_ENDPOINT>",
+      "// allow_insecure documentation": "Set this to true for clients to ignore when the source cluster's HTTP certificate was not signed with a trusted CA Root Certificate.  Default: false",
```
AndreKurait, line 18: Let's add this on the target too

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -467,12 +467,13 @@ export function parseClusterDefinition(json: any): ClusterYaml {
...
     const version = json.version;
     const auth = parseAuth(json.auth)
     if (!auth) {
         throw new Error(`Invalid auth type when parsing cluster definition: ${json.auth.type}`)
     }
-    return new ClusterYaml({endpoint, version, auth})
+    return new ClusterYaml({endpoint, version, allowInsecure, auth})
```
AndreKurait, line 476: let's string this through capture proxy, replayer, and backfill to ensure we pass the required arguments there when this is true

gregschohn: We have this line in place already for the capture proxy's CDK.  If we were to pull this value from the services.yaml file, which should have precedence?  I'd rather make the comment for the services.yaml a little bit tighter right now than to break any existing contract that we have for the proxy deployments.  Since we're working on deploying an entirely new way, I'd also rather us not invest in more error checking upfront to make sure that values are not inconsistent.
```
command = appendArgIfNotInExtraArgs(command, extraArgsDict, "--insecureDestination")
```

deployment/cdk/opensearch-service-migration/cdk.context.json
AndreKurait: I'm fine with that, should be straight forward enough for someone to try it out if they need it

deployment/cdk/opensearch-service-migration/lib/migration-services-yaml.ts
```diff
@@ -4,17 +4,20 @@ import * as yaml from 'yaml';
...
         this.endpoint = endpoint;
         this.auth = auth;
+        this.allowInsecure = allowInsecure;
         this.version = version;
     }
     toDict() {
         return {
             endpoint: this.endpoint,
             version: this.version,
+            allow_insecure: this.allowInsecure,
```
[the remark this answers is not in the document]
gregschohn: false would always put a value in the services.yaml file.  As it is, the code will fold down to a missing allow_insecure field for when it was missing in 'this'.  That seems a bit more appropriate to me rather than having yet another spot where we find a default value.

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
gregschohn: Looking back, the original ticket is for an insecure TARGET, so this is the way that I went.  See the latest updates.

AndreKurait (approved):
LGTM, thanks for the updates!! Big improvement in usability here

PR #1620, opened 2025-07-02
Synchronize Secret Format and Usage

lewijacn:
### Description
This change makes our usage of AWS secrets for basic auth consistent across the solution (minus K8s which will have a separate task for its secret management). It also includes various improvement for the MA CDK library such as removing some invalid context options, making task names readable and predictable, and removing the OS container stack which was causing additional complexity and is not being used. This change also provides a way for RFS and Replayer services to load specified arguments from environment variables, as would be the case when a secret is provided to ECS or EKS as this is how it provides the values to the container

**Note**: This includes a breaking change to secret format
Previously:
We expected an AWS secret to contain only the password value as plaintext with no structure and would expect the username through a separate plaintext argument
```
myStrongPassword123!
```

Now:
We follow the standard AWS secret format of containing both the username and password as Key-Value pairs within the secret 
```
{"username":"admin","password":"myStrongPassword123!"}
```
The console library will produce errors like below when using an improperly structured secret:
```
migration-console (~) -> console clusters cat-indices 

WARNING: Cluster information may be stale. Use --refresh to update.

SOURCE CLUSTER
health status index                           uuid                   pri rep docs.count docs.deleted store.size pri.store.size
green  open   test_e2e_0001_integ_full_jul9_2 U--CcY4FQVGQx30DY7q7LA   3   0          3            0     13.3kb         13.3kb

TARGET CLUSTER
Error: Unable to perform cat-indices command with message: Expected secret arn:aws:secretsmanager:us-east-1:123456789012:secret:test-jul2-secret-Y8JbPm to be a JSON object with username and password fields
```
Or a secret with missing keys:
```
migration-console (~) -> console clusters curl target_cluster /
Error: Unable to perform cluster command with message: Secret arn:aws:secretsmanager:us-east-1:123456789012:secret:test-jul2-secret-Y8JbPm is missing required key(s): username, password
```

### Other Fixes
Task roles in AWS now follow a consistent pattern like below: 
```
OSMigrations-full-es68-3-us-east-1-capture-proxy-task
OSMigrations-full-es68-3-us-east-1-reindex-from-snapshot-task
OSMigrations-full-es68-3-us-east-1-target-cluster-proxy-task
OSMigrations-full-es68-3-us-east-1-traffic-replayer-default-task
OSMigrations-full-es68-3-us-east-1-migration-console-task
```

* Removed printing RFS args twice on every startup
* Added basic logic for RFS and Replayer to redact password fields when printing args at startup
* Removed printing stack traces for middleware layer for cat-indices and cluster curl commands

### Bug Fixes to check
We were using using source_cluster insecure here, but have changed to target cluster: `skipClusterCertCheck: servicesYaml.target_cluster?.allowInsecure,`
```
        let targetClusterProxyStack
        if (targetClusterProxyServiceEnabled && networkStack && migrationStack) {
            targetClusterProxyStack = new CaptureProxyStack(scope, "target-cluster-proxy", {
                serviceName: "target-cluster-proxy",
                vpcDetails: networkStack.vpcDetails,
                destinationConfig: {
                    endpointMigrationSSMParameter: MigrationSSMParameter.OS_CLUSTER_ENDPOINT,
                    securityGroupMigrationSSMParameter: MigrationSSMParameter.OS_ACCESS_SECURITY_GROUP_ID,
                },
                otelCollectorEnabled: false,
                skipClusterCertCheck: servicesYaml.target_cluster?.allowInsecure,
```

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-1787
https://opensearch.atlassian.net/browse/MIGRATIONS-1915

### Testing
Unit testing
Cloud testing in progress...

### Check List
- [X] Public documentation issue/PR [created]: https://github.com/opensearch-project/documentation-website/pull/10186


DocumentsFromSnapshotMigration/docker/entrypoint.sh
```diff
@@ -36,43 +36,14 @@ trap cleanup SIGTERM SIGINT SIGQUIT
 
 
 # Print our ENV variables
-if [[ $RFS_COMMAND != *"--target-password"* && $RFS_COMMAND != *"--targetPassword"* ]]; then
-    echo "RFS_COMMAND: $RFS_COMMAND"
-else
-    echo "RFS Target Cluster password found in RFS_COMMAND; skipping logging of the value"
-fi
+echo "RFS_COMMAND: $RFS_COMMAND"
```
AndreKurait, line 39: This will print the password if set by `--target-password` right?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -254,6 +257,28 @@ public String getTransformerConfigParameterArgPrefix() {
         private String transformerConfigFile;
     }
 
+    public static class EnvParameters {
+
+        private EnvParameters() {
+            throw new IllegalStateException("EnvParameters utility class should not instantiated");
+        }
+
+        public static void injectFromEnv(Args args) {
```
gregschohn, line 266: This can wait, but this could be generalized into the solution for https://opensearch.atlassian.net/browse/MIGRATIONS-2634.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -494,45 +530,32 @@ private static String formatAuthArgFlagsAsString() {
...
-        if (params.removeAuthHeader
-            && params.authHeaderValue != null
-            && params.useSigV4ServiceAndRegion != null
-            && params.awsAuthHeaderUserAndSecret != null) {
+        long authOptionsSpecified = Stream.of(
+            params.removeAuthHeader,
+            params.authHeaderValue != null,
+            params.useSigV4ServiceAndRegion != null,
+            params.targetUsername != null || params.targetPassword != null
+        ).filter(b -> b).count();
```
gregschohn, line 545: nit - there's an Function.identity that you can use instead.  FWIW, I am partial the 4 characters that you've used :).

coreUtilities/src/main/java/org/opensearch/migrations/utils/ArgLogUtils.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public static List<String> getRedactedArgs(String[] args) {
+        List<String> redactedArgs = new ArrayList<>();
+        boolean shouldCensorNext = false;
+
+        for (String arg : args) {
+            if (shouldCensorNext) {
+                redactedArgs.add(CENSORED_VALUE);
+                shouldCensorNext = false;
+            } else if (arg.equals("--target-password") || arg.equals("--targetPassword")) {
```
gregschohn, line 22: These should be passed in.  We may have passwords for work coordination or sources too.

deployment/cdk/opensearch-service-migration/lib/common-utilities.ts
```diff
@@ -170,11 +183,26 @@ export function createAwsDistroForOtelPushInstrumentationPolicy(): PolicyStateme
...
 
-export function createDefaultECSTaskRole(scope: Construct, serviceName: string): Role {
-    const serviceTaskRole = new Role(scope, `${serviceName}-TaskRole`, {
+export function createECSTaskRole(scope: Construct, serviceName: string, region: string, stage: string): Role {
+    let taskRoleName = `OSMigrations-${stage}-${region}-${serviceName}-task`
+    const excessCharacters = taskRoleName.length - MAX_IAM_ROLE_NAME_LENGTH
+    if (excessCharacters > 0) {
+        if (excessCharacters > serviceName.length) {
+            throw Error(`Unexpected ECS task role name length for proposed name: '${taskRoleName}' could not be reasonably truncated 
+                below ${MAX_IAM_ROLE_NAME_LENGTH} characters`)
```
gregschohn, line 192: Thanks!

gregschohn (commented):
Why is there an empty osi_utils.py file?
I went through the non-ts code and left some comments.  Let me know if you need a thorough review on the CDK parts.

gregschohn:
One other idea - could you scan the cdk's extra args for usernames and passwords and emit a warning?

deployment/cdk/opensearch-service-migration/lib/opensearch-domain-stack.ts
```diff
@@ -38,8 +34,7 @@ export interface OpensearchDomainStackProps extends StackPropsExt {
   readonly openAccessPolicyEnabled?: boolean
   readonly useUnsignedBasicAuth?: boolean,
   readonly fineGrainedManagerUserARN?: string,
```
peternied, line 36: Can we remove this?

lewijacn:
> Why is there an empty osi_utils.py file? I went through the non-ts code and left some comments. Let me know if you need a thorough review on the CDK parts.

Thanks for catching have removed the file now

gregschohn:
> One other idea - could you scan the cdk's extra args for usernames and passwords and emit a warning?

Since our CDK implementation's days are numbered, we should just do this in the configure commands for the K8s workflow commands... Even better, route those values directly to secrets (even if they come in via a file).

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -265,13 +264,13 @@ private EnvParameters() {
 
         public static void injectFromEnv(Args args) {
             List<String> addedEnvParams = new ArrayList<>();
-            if (args.targetArgs.username == null && System.getenv(TARGET_USERNAME_ENV_VAR) != null) {
-                args.targetArgs.username = System.getenv(TARGET_USERNAME_ENV_VAR);
-                addedEnvParams.add(TARGET_USERNAME_ENV_VAR);
+            if (args.targetArgs.username == null && System.getenv(ArgNameConstants.TARGET_USERNAME_ENV_ARG) != null) {
```
gregschohn, line 267: If you added the env variables BEFORE the JCommander ones, would you need to check for null?
Should one take precedent over the other or should you throw an error if they're both specified (and different)?

lewijacn: My stance (and what seems like should be expected) has been that command line arguments have priority over environment variables. I think we will get to the point where we are using command line arguments less, and they will be more an override than the norm

DocumentsFromSnapshotMigration/docker/entrypoint.sh
```diff
@@ -35,45 +35,11 @@ cleanup() {
...
-    # Append the username/password to the RFS Command if have an updated password
-    if [[ -n "$PASSWORD_TO_USE" ]]; then
-        echo "Updating RFS Command with password."
-        RFS_COMMAND="$RFS_COMMAND --target-password \"$PASSWORD_TO_USE\""
-    fi
+SAFE_PRINT_COMMAND="$RFS_COMMAND"
+if [[ "$RFS_COMMAND" == *"--target-password"* || "$RFS_COMMAND" == *"--targetPassword"* ]]; then
+    SAFE_PRINT_COMMAND=$(echo "$RFS_COMMAND" | sed -E 's/--target(-)?[Pp]assword[ =]"?[^"[:space:]]+"?/--target-password=******/g')
 fi
+echo "Executing RFS_COMMAND: $SAFE_PRINT_COMMAND"
```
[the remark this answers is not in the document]
lewijacn: I have kept it as a safeguard in case there is some issue with the `entrypoint.sh` before we start the RFS app

TrafficCapture/trafficReplayer/build.gradle
```diff
@@ -38,8 +38,7 @@ dependencies {
     implementation libs.log4j.slf4j2.impl
     implementation libs.slf4j.api
     implementation libs.disruptor
-    implementation libs.aws.arns
-    implementation libs.aws.secretsmanager
+    implementation libs.aws.auth
```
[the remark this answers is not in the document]
lewijacn: Yes was included in the previous libraries. Specifically the Replayer still has need of this import for SigV4: `import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;`

deployment/cdk/opensearch-service-migration/cdk.context.json
```diff
@@ -7,9 +7,8 @@
       "allow_insecure": true,
       "auth": {
         "type": "none | basic | sigv4",
-        "// basic auth documentation": "The next two lines are relevant for basic auth only",
-        "username": "<USERNAME>",
-        "passwordFromSecretArn": "<ARN_OF_SECRET_CONTAINING_PASSWORD>",
+        "// basic auth documentation": "The next line is relevant for basic auth only",
```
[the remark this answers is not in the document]
lewijacn: I've gone back and forth with this. The `options.md` does list the options for users more intent on using plaintext username/password. However, as a default I'd like to steer users to using a secret. I'm open to changing this though if we find this to be problematic

PR #1624, opened 2025-07-03
Remove the extra log messages around metadata tool and include a logpath

peternied:
### Description
- Remove the log4j configuration statements from the metadata startup/shutdown process.
- Fix an issue where inner exceptions were not being returned to users during metadata migration workflow.
- Added a log line that indicates where the current run's metadata log was created for faster troubleshooting.

#### New
```
Starting Metadata Migration
Results:
   Issue(s) detected
Issues:
   Unexpected failure: Retries exhausted: 3/3, inner cause: Failed to resolve 'hostname' [A(1)]
   No source was defined
   No target was defined

Consult /home/ubuntu/git/opensearch-migrations/MetadataMigration/./logs/ip-172-31-18-157/metadata/metadata2025-07-03_17-39-08.log to see detailed logs for this run
```

#### Previous
```
2025-07-03T17:40:05.050985425Z main INFO Starting configuration org.apache.logging.log4j.core.config.properties.PropertiesConfiguration@40cb8df7...
2025-07-03T17:40:05.052605364Z main INFO Configuration org.apache.logging.log4j.core.config.properties.PropertiesConfiguration@40cb8df7 started.
2025-07-03T17:40:05.054522271Z main INFO Stopping configuration org.apache.logging.log4j.core.config.DefaultConfiguration@32464a14...
2025-07-03T17:40:05.054804970Z main INFO Configuration org.apache.logging.log4j.core.config.DefaultConfiguration@32464a14 stopped.
Starting Metadata Migration
Results:
   Issue(s) detected
Issues:
   Unexpected failure: Retries exhausted: 3/3
   No source was defined
   No target was defined

2025-07-03T17:40:13.009890461Z pool-1-thread-1 INFO Stopping configuration org.apache.logging.log4j.core.config.properties.PropertiesConfiguration@40cb8df7...
2025-07-03T17:40:13.010343766Z pool-1-thread-1 INFO Configuration org.apache.logging.log4j.core.config.properties.PropertiesConfiguration@40cb8df7 stopped.
```

### Issues Resolved
- Resolves [[MIGRATIONS-2632 | Clean up logging output from metadata migration]](https://opensearch.atlassian.net/browse/MIGRATIONS-2632)

### Testing
Added unit tests for the exception messaging.

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
+++ b/deployment/cdk/opensearch-service-migration/lib/network-stack.ts
@@ -134,7 +134,7 @@ export class VpcDetails {
                 throw new Error(`Not enough AZs (${azCount} unique AZs detected) used for private subnets to meet the ${azCount} AZ requirement. Alternatively subnets can be manually specified with the 'vpcSubnetIds' option and the AZ requirement set with the 'vpcAZCount' option`)
             }
             const desiredSubnetIds = uniqueAzPrivateSubnets.subnetIds
-                .toSorted((a, b) => a.localeCompare(b))
+                .sort((a, b) => a.localeCompare(b))
                 .slice(0, azCount);
             this.subnetSelection = vpc.selectSubnets({
                 subnetFilters: [
--- a/deployment/cdk/opensearch-service-migration/tsconfig.json
+++ b/deployment/cdk/opensearch-service-migration/tsconfig.json
@@ -3,7 +3,7 @@
     "outDir": "dist",
     "target": "ES2020",
     "module": "commonjs",
-    "lib": ["es2023"],
+    "lib": ["es2020"],
     "resolveJsonModule": true,
     "declaration": true,
     "strict": true,


MetadataMigration/src/main/resources/log4j2.properties
```diff
@@ -1,4 +1,4 @@
-status = INFO
+status = WARN
```
peternied, line 1: FYI - This is the common level used in all the other log4j2.properties files

MetadataMigration/src/test/java/org/opensearch/migrations/commands/EvaluateTest.java
```diff
@@ -8,10 +8,24 @@
...
+    @Test
+    void evaluate_failsInvalidParameters() {
+        var args = new MigrateArgs();
+        var context = mock(RootMetadataMigrationContext.class);
+        var meta = new MetadataMigration();
+
+        var results = meta.evaluate(args).execute(context);
+
+        assertThat(results.getExitCode(), equalTo(Migrate.INVALID_PARAMETER_CODE));
+        assertThat(results.getErrorMessage(), equalTo("Invalid parameter: No details on the source cluster found, please supply a connection details or a snapshot"));
```
gregschohn, line 26: I've never looked at the Result class before now.  The class doesn't describe its data in any way - there's no structure to pull important details from it.  I don't see any implementations for getErrorMessage() other than within test code.  Should this pattern be reconsidered?

You should also take a look at CloseableLogSetup.  That will let you test your log outputs without needing to create your own abstractions.  The advantage is that you won't need to carry any test code through production code.

MetadataMigration/src/test/java/org/opensearch/migrations/commands/EvaluateTest.java
```diff
@@ -24,5 +38,21 @@ void evaluate_failsUnexpectedException() {
...
+    }
+
+    @Test
+    void evaluate_failsUnexpectedExceptionInnerMessage() {
+        var args = new EvaluateArgs();
+        var meta = new MetadataMigration();
+        var context = mock(RootMetadataMigrationContext.class);
+ 
+        var evaluate = spy(meta.evaluate(args));
+        doThrow(new RuntimeException("Outer", new RuntimeException("Inner"))).when(evaluate).createClusters();
```
gregschohn, line 51: As per the above, I'll make the argument that when you have a RuntimeException that isn't a known more specific type, you definitely should NOT be eating the stacktraces.

MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
```diff
@@ -137,4 +138,10 @@ private IndexMetadataResults migrateIndices(MigrationMode mode, Clusters cluster
         var indexResults = indexRunner.migrateIndices(mode, context.createIndexContext());
         log.info("Index copy complete.");
         return indexResults;
-    } }
+    }
+
+    protected String createUnexpectedErrorMessage(Throwable e) {
+        var causeMessage = Optional.of(e).map(Throwable::getCause).map(Throwable::getMessage).orElse(null);
+        return "Unexpected failure: " + e.getMessage() + (causeMessage == null ? "" : ", inner cause: " + causeMessage);
```
gregschohn, line 145: If we have *any* kind of exception, we'll NOT show a stack trace?  If there's a NullPointerException, I certainly want to see the stack trace for that.  You should scope down 'expected' vs 'unexpected' exceptions and treat them differently.

MetadataMigration/src/test/java/org/opensearch/migrations/commands/EvaluateTest.java
peternied: getErrorMessage() is used in Evaluate/Migrate. You are right that the clear boundary of the presentation layer components isn't clear - in this case EvaluateResult is a presentation component.

Testing the logger isn't needed for this scenario - its a unit test that ensures the presentation format is prepared in expected way. I've seen CloseableLogSetup, it is used in other tests.

MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
peternied: This is presentation layer, user are not suppose to see callstacks - this is the intention.  Actual exceptions are logged and available in the metadata run log.

gregschohn:
What does "[A(1)]" mean in the sample output in the PR description?

There's a bigger question here that makes it hard to review this PR.  Most of the rest of our Java code is for long-running processes that will not be invoked directly by the user at the command line.  In all of those other cases, managing logs as files that are eventually aggregated is going to be the way that a user gets feedback and all of it will be asynchronously.

If a user is running a short or interactive utility on the console, should we be thinking about logfiles for user-feedback at all?  (logfiles could still be useful for us to recover from them to diagnose bugs and unforeseen issues).  If metadata migrate was run in a separate container, but still interactively, would using log4j2 and/or loggers make the most sense to provide feedback?  I'm on the fence, also considering hardcoding stdout/stderr within java and a also having a --verbose flag.  Hardcoding stdout/stderr could also be done in the container w/ the log4j properties file - but w/out another utility to upload a file directly to a ticket, etc - does keeping a bunch of logfiles around and telling the user about it provide a better experience than --verbose that might print most of those contents out to the console (& can be sent where the user chooses or filtered, etc).

Metadata migrate & most of the other console commands (like those for interfacing with tuples) feel like they could be more unix-inspired, but I'm also talking as a shell-piping fan.  I'd lean toward continuing to use log4j, but also provide a --verbose (or --info/--debug, etc) flag to wire up different log4j properties that we pass into the process from the console python code to change its behavior.  I wouldn't have the program itself make any advertisements about its logging situation & let the caller orchestrate that.  I'm on the fence about the importance of logfiles for a user - I don't think of going to a logfile for interactive console programs when I run them.  E.g. emacs, jq, curl, (honestly, even gradle, though I now know better), etc may all take a while to run or encounter runtime issues, but I don't normally think, oh - where do I look at the logs

peternied:
@gregschohn Thanks for the review.  After talking this over it looks like this is a decent approach in the near term, while there are trade off such as not having a crisp interface boundary - I created https://opensearch.atlassian.net/browse/MIGRATIONS-2642 to track those improvements so we can deliver on this incremental improvement.

Mind taking another look?

MetadataMigration/src/main/java/org/opensearch/migrations/commands/Evaluate.java
```diff
@@ -38,10 +37,9 @@ public EvaluateResult execute(RootMetadataMigrationContext context) {
                 .build();
         } catch (Throwable e) {
```
gregschohn, line 38: This should most definitely NOT be throwable.  Unless you have very good reason to handle Errors, don't catch Throwables.  I've only know of one place where we need to catch one (while shutting down the replayer, which can uniquely die from > 1 signal/error).

peternied: - Resolving this out of band in https://github.com/opensearch-project/opensearch-migrations/pull/1631

PR #1633, opened 2025-07-10
Checkpoint for feedback.

gregschohn:
3 psuedocode files to show what the structure would be to get 60-90% of the way through our argo use cases

### Issues Resolved
Related to https://opensearch.atlassian.net/browse/MIGRATIONS-2532, but it has a long way to go

### DIFF
--- a/orchestrationSpecs/userConfigSchemas.txt
+++ b/orchestrationSpecs/userConfigSchemas.txt
@@ -1,0 +1,59 @@
+HTTP_AUTH_BASIC: {
+  username: string
+  password: string
+}
+
+HTTP_AUTH_SIGV4: {
+  region:  string
+  service?: string | "es" // default
+}
+
+CLUSTER_CONFIG: {
+  name:            string
+  endpoint?:       string // optional for source and maybe metadata evaluate commands, but required for most others
+  allow_insecure?: bool
+  version?:        string // required in some contexts?
+  authConfig?:     HTTP_AUTH_BASIC | HTTP_AUTH_SIGV4
+}
+
+TARGET_CLUSTER_CONFIG: CLUSTER_CONFIG {
+  endpoint: string
+}
+
+SNAPSHOT_MIGRATION_CONFIG: {
+	indices: [string]
+	migrations: [{
+		metadata: {
+			mappings: {
+				properties: {...}
+			},
+			//
+			documentBackfillConfigs: {
+				indices: [...string],
+				config: {
+					batchSize: int,
+					initialReplicas: int
+				}
+			}
+		}
+	}]
+}
+
+REPLAYER_CONFIG: {
+	speedupFactor: float
+	initialReplicas: int
+}
+
+SOURCE_MIGRATION_CONFIG: {
+  source: #CLUSTER_CONFIG,
+  snapshotAndMigrationConfigs: [SNAPSHOT_MIGRATION_CONFIG],
+  replayerConfig: REPLAYER_CONFIG
+}
+
+MIGRATION_CONFIG: {
+  sessionName: string,
+  sourceMigrations: SOURCE_MIGRATION_CONFIG,
+  targets: [TARGET_CLUSTER_CONFIG]
+}
+
+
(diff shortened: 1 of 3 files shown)


orchestrationSpecs/userConfigSchemas.txt
```diff
@@ -0,0 +1,59 @@
...
+  endpoint: string
+}
+
+SNAPSHOT_MIGRATION_CONFIG: {
+	indices: [string]
+	migrations: [{
+		metadata: {
+			mappings: {
+				properties: {...}
+			},
```
peternied, line 29: What does this model in the mappings?  Is this another way to declare transformations?

orchestrationSpecs/userConfigSchemas.txt
```diff
@@ -0,0 +1,59 @@
...
+			//
+			documentBackfillConfigs: {
+				indices: [...string],
```
peternied, line 32: t seems like we will need to have more complex validation on this code where as a model expected multiple snapshot migration configs would be less error prone.

Consider these questions, why is there an outer indices and inner indices object?  Is this about allocating the indices to specific backfill parameters then it is possible some are included multiple times or accidentally not included?

orchestrationSpecs/userConfigSchemas.txt
```diff
@@ -0,0 +1,59 @@
...
+					batchSize: int,
+					initialReplicas: int
+				}
+			}
+		}
+	}]
+}
+
+REPLAYER_CONFIG: {
+	speedupFactor: float
```
peternied, line 43: Why is the speedup factor a constant?  Should a user be able to dial this up and down during an execution of a workflow?

orchestrationSpecs/userConfigSchemas.txt
```diff
@@ -0,0 +1,59 @@
...
+	initialReplicas: int
```
peternied, line 44: Replicas of what - is this conflating features of the captured traffic system (kafka queue replicas?) into the replayer configuration? What is a case where a user would want this to be zero or 100?

orchestrationSpecs/userConfigSchemas.txt
peternied: Slightly further - why does a user set this value at all, shouldn't we auto-scale the replay speedup based on the target cluster performance just like RFS auto-scaling?

orchestrationSpecs/userConfigSchemas.txt
```diff
@@ -0,0 +1,59 @@
...
+HTTP_AUTH_SIGV4: {
+  region:  string
+  service?: string | "es" // default
+}
+
+CLUSTER_CONFIG: {
+  name:            string
+  endpoint?:       string // optional for source and maybe metadata evaluate commands, but required for most others
+  allow_insecure?: bool
+  version?:        string // required in some contexts?
```
peternied, line 15: This should be typed - use/reimplement `transformation/src/main/java/org/opensearch/migrations/Version.java`

orchestrationSpecs/userConfigSchemas.txt
```diff
@@ -0,0 +1,59 @@
+HTTP_AUTH_BASIC: {
+  username: string
+  password: string
+}
+
+HTTP_AUTH_SIGV4: {
+  region:  string
+  service?: string | "es" // default
+}
```
peternied, line 9: I'd elect to remove this entirely, hidden from the user can't we determine if is AOS or AOSS from the endpoint so the user doesn't have to do this work.

peternied (commented):
Thanks for getting this out there for us to take a look at, left many design related comments and a couple of more tactically focused feedback on `argoConfigRenderer.ts`.

lewijacn (approved):
README is very helpful thanks, commands worked as expected for me

nit: We should add `orchestrationSpecs/..tsbuildinfo` to `.gitignore`

orchestrationSpecs/createMigration.yaml
```diff
@@ -36,7 +36,7 @@ spec:
                           },
                           "migrations": [
                               {
-                                  "metadata": {
+                                  "NOTmetadata": {
```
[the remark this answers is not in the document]
gregschohn: Yes - it's so that I can easily turn that config off to make sure that skipping the step works

PR #1636, opened 2025-07-10
Add support for metadata migration from ES 2.4 snapshot

jugal-chauhan:
### Description
Our tool MA supports various versions starting from ES 5.0, uptil ES 8.x
Working backwards with this PR, our tool will now also support metadata migration from ES 2.4 clusters/snapshot.
I have added ES 2.4.6 container in our supported clusters matrix (and not extended sources)

Metadata Migration E2E test modifications
- Testing data in the documents have been changed because we do not want any test cases with `"type": "string"/"keyword"`
- ES 2.x allows "type": "string" instead of "type": "keyword" as other ES 5+ versions
- We already solve this part using a javascript transformer, which is well documented
- our Metadata E2E test cannot inject javascript transformer, hence I have removed any testing data of such data-type.

What was different in ES 2.4 (vs ES 5.0) : 
- Index structure = ES 2.4 snapshot “index” files contain only snapshot names, without any mapping of indices → snapshot IDs or UUIDs (unlike ES 5.4+ which maintain index UUIDs and snapshot references)
- Snapshot Repo Layout = There is no per-snapshot UUID-based substructure as introduced in ES 5.x. `index-N` files were simply named `index`. The metadata files are pre-fixed with `meta-<snapshot-name>.dat` instead of `meta-<uuid>.dat`
- Snapshot format = ES 2.4 snapshots contain older JSON schemas with legacy fields (ex, “template” instead of “index_patterns”) and simpler layouts requiring custom parsing and normalization in the reader; whereas ES 5.6 snapshots were already using a standardized JSON format (Smile-encoded) that largely aligns with modern OpenSearch expectations, requiring only minimal wrapping for transformer compatibility.

New files added in codebase : 

version_es_2_4/
- SnapshotReader_ES_2_4: ClusterSnapshotReader implementation for ES 2.4 with modified compatibility matrix
- SnapshotRepoProvider_ES_2_4: Provides access to ES 2.4 snapshot repository contents by navigating its simpler directory layout, listing available indices, and reading their metadata files.
- SnapshotRepoData_ES_2_4: Parses the top-level index file in an ES 2.4 repository to extract and expose the list of available snapshot names. (Map between snapshots and indices is not present)
- SnapshotRepo_ES_2_4: Defines a specialized interface extension of SnapshotRepo.Provider for ES 2.4, adding methods to list index directories and retrieve raw metadata file contents directly.
- GlobalMetadataFactory_ES_2_4: similar to GlobalMetadataFactory of other readers
- GlobalMetadataData_ES_2_4: Holds ES 2.4 global cluster metadata as an ObjectNode tree. Exposes JSON Pointers for accessing templates, index templates, and aliases.
- IndexMetadataFactory_ES_2_4: Factory for creating IndexMetadataData_ES_2_4 instances by decoding raw Smile-encoded binary index files.
- IndexMetadataData_ES_2_4: Represents ES 2.4 index-level metadata parsed from Smile, exposing mappings, settings, and index name.

### Issues Resolved
[MIGRATIONS-2481](https://opensearch.atlassian.net/browse/MIGRATIONS-2481)

### Testing
`./gradlew :MetadataMigration:isolatedTest --tests "*EndToEndTest" --info`

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/models/AliasMetadata.java
AndreKurait, line 1: Is this specific to ES 2, if so, should it be in that versioned folder?

RFS/src/main/java/org/opensearch/migrations/bulkload/models/CompressedMapping.java
AndreKurait, line 1: Is this specific to ES 2?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_2_4/GlobalMetadataData_ES_2_4.java
```diff
@@ -0,0 +1,124 @@
...
+import org.opensearch.migrations.bulkload.models.IndexMetadata;
+import org.opensearch.migrations.bulkload.models.IndexTemplate;
+
+import com.fasterxml.jackson.core.JsonPointer;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import lombok.Data;
+
+@Data
+public class GlobalMetadataData_ES_2_4 implements GlobalMetadata {
```
AndreKurait, line 16: This file follows a completely different structure than `GlobalMetadataData_ES_6_8`, `GlobalMetadataData_ES_7_10`, and `GlobalMetadataData_OS_2_11`. Why was this design choice made and if it has benefits should we adopt a similar pattern across the other ones?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_2_4/SnapshotReader_ES_2_4.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public ShardMetadata.Factory getShardMetadata() {
+        throw new UnsupportedOperationException(
+            "Reading ShardMetadata for ES 2.4 snapshots is not yet implemented."
+        );
+    }
+
+    @Override
+    public boolean getSoftDeletesPossible() {
```
AndreKurait, line 62: Should we use a ElasticsearchConstants_ES_2_4.SOFT_DELETES_POSSIBLE along with the other fields?

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_2_4/IndexMetadataData_ES_2_4.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import static org.opensearch.migrations.bulkload.version_es_2_4.ElasticsearchConstants_ES_2_4.*;
```
AndreKurait, line 13: Let's remove * imports

RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_2_4/GlobalMetadataData_ES_2_4.java
jugal-chauhan: I have made the structures similar now

RFS/src/main/java/org/opensearch/migrations/bulkload/models/CompressedMapping.java
```diff
@@ -0,0 +1,11 @@
+package org.opensearch.migrations.bulkload.models;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class CompressedMapping {
+    private String type;
+    private byte[] source;
```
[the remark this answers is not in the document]
jugal-chauhan: Removed this file completely as not using it

RFS/src/main/java/org/opensearch/migrations/bulkload/models/CompressedMapping.java
jugal-chauhan: Both of those files have been removed

jugal-chauhan:
I will change the overview before merging

PR #1665, opened 2025-07-16
Support Metadata migration and Backfill from ES 1.7 snapshots as source

jugal-chauhan:
### Description
This PR brings in metadata migration and backfill supporting ES 1.7 as a source cluster in MA.

From the [previous PR](https://github.com/opensearch-project/opensearch-migrations/pull/1636) (supporting metadata migration for ES 2.4) we had already modified our testing data in E2E test to avoid injecting the javascript transformer. Using this same `migrations/EndToEndTest.java` as a point of validation, this PR brings in a new Snapshot Reader _ES_1_7 which helps users migrate their metadata from a ES 1.7 snapshot structure.

Enlisting the differences between ES_1_7 and ES_2_4 snapshot reader:
1. Index Metadata files are named differently
`ES 1.7 = indices/{indexName}/snapshot-{snapshotName}`
`ES 2.4 = indices/{indexName}/meta-{snapshotName}.dat`

2. Global Metadata files are also named differently
`ES 1.7 = metadata-{snapshotName}`
`ES 2.4 = meta-{snapshotName}.dat`

3. In ES 1.7 we have no SMILE encoding, it is purely JSON, whereas in ES 2.4 we had smile encoding.
4. `SnapshotRepoES17` requires `snapshotName` parameter since filenames have `snapshot-<snapshotName>` format. Whereas in `SnapshotRepoES24`, we used the internally determined snapshot name.

To reference the entire structure of ES 1.x snapshots, please view the description of the JIRA attached below.

### NOTE
Both the metadata migration support PRs (this + [previous](https://github.com/opensearch-project/opensearch-migrations/pull/1636) one for ES 2.x) only support File System migration. This logic will fail on S3. Another [work item in JIRA](https://opensearch.atlassian.net/browse/MIGRATIONS-2650) has been created to achieve complete metadata migration support.

Reason for anticipated failure on S3 : Without a straightforward path to the per-index metadata file, S3 Repo will not download it making it impossible to fetch a Map for snapshots and indicies in snapshot. Our S3Repo currently follows a static path to these metadata files, which no longer exists.

Breaking changes in snapshot repo : In ES 1.7 and in ES 2.4, the top-level `index/index-0` file does not contain a map of indices present in a snapshot. This map is currently created by looking at each index's metadata file to see which snapshot does it relate to - this is tedious. 

### Issues Resolved
[MIGRATIONS-2483](https://opensearch.atlassian.net/browse/MIGRATIONS-2483)
[MIGRATIONS-2482](https://opensearch.atlassian.net/browse/MIGRATIONS-2482)

### Testing
`./gradlew :MetadataMigration:isolatedTest --tests "*EndToEndTest" --info`
`./gradlew :DocumentsFromSnapshotMigration:isolatedTest --tests "*EndToEndTest" --info`

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/version_es_2_4/SnapshotRepoProvider_ES_2_4.java
```diff
@@ -129,6 +81,50 @@ public SnapshotRepo.Provider getDelegateRepo() {
...
+                if (f.isDirectory()) {
+                    indexNames.add(f.getName());
+                }
+            }
+        }
+        return indexNames;
+    }
+
+    @Override
+    public byte[] getIndexMetadataFile(String indexName) {
```
AndreKurait, line 101: This function is problematic since it does not filter by the snapshot name and thus could retrieve the metadata from a previous (or later) snapshot in a repo. Where is it used? Let's remove it from everywhere if we can

AndreKurait (approved):

Issue #1684, opened 2025-07-23
[BUG] ES 7.17.12 -> OS 2.19 backfill error: "Completion84" PostingsFormat is not supported


n-logn:
### What is the bug?
When running the backfill migration from ES 7.17.12 to OS 2.19, an error occured:
```
An SPI class of type shadow.lucene9.org.apache.lucene.codecs.PostingsFormat with name 'Completion84' does not exist.
You need to add the corresponding JAR file supporting this SPI to your classpath.
The current classpath supports the following names: [Lucene50, Lucene84, Lucene90, Lucene99, Lucene912, ES812Postings, ES87BloomFilter]
```

### What are your migration environments?
ES 7.17.12
OS 2.19
OSMA 2.4.5

### How can one reproduce the bug?
- In ES 7.17.12: add an index with a field of type "completion", populate index with document(s)
```json
"properties": {
  "en": {
    "type": "text",
    "fields": {
    "completion": {
      "type": "completion",
      "analyzer": "simple",
      "preserve_separators": true,
      "preserve_position_increments": true,
      "max_input_length": 50
    }
  }
}
```
- Migrate to OS 2.19 using [guide](https://docs.opensearch.org/docs/latest/migration-assistant/deploying-migration-assistant/getting-started-data-migration/)
- Backfill hangs in status RUNNING
```
BackfillStatus.RUNNING
Running=5
Pending=0
Desired=5
Terminating=0
Work items total: 2
Work items completed: 1
Work items incomplete: 1
Work items in progress: 1
Work items unclaimed: 0
```
- There is an error in ECS service task logs (below)

### What is the expected behavior?
Migration passed successfully

### Do you have any additional context?
ECS service task log:
```
ERROR o.o.m.b.w.DocumentsRunner [DocumentBatchReindexer-1] Error prevented some batches from being processed
  java.lang.IllegalArgumentException:
    An SPI class of type shadow.lucene9.org.apache.lucene.codecs.PostingsFormat with name 'Completion84' does not exist.
    You need to add the corresponding JAR file supporting this SPI to your classpath.
    The current classpath supports the following names: [Lucene50, Lucene84, Lucene90, Lucene99, Lucene912, ES812Postings, ES87BloomFilter]
  at shadow.lucene9.org.apache.lucene.util.NamedSPILoader.lookup(NamedSPILoader.java:113)
[...]

PR #1698, opened 2025-07-24
Add support for completion codec available in ES 5/6/7

jugal-chauhan:
### Description
This PR address a recent issue discovered due to missing Codec.

Code changes include:
- Adding lucene-suggest as a runtime dependency to the Lucene 9/7/6/5 shadow jar
- Introduced createIndexWithCompletionField() in ClusterOperations to create a completion field index with correct mappings/settings across ES versions.
- Added CompletionCodecTest to validate document migration involving completion fields from ES 5.6, 6.8, and 7.17 to OS targets.

### Issues Resolved
[Github Issue](https://github.com/opensearch-project/opensearch-migrations/issues/1684)
[MIGRATIONS-2654](https://opensearch.atlassian.net/browse/MIGRATIONS-2654)

### Testing
CompletionCodecTest has been added and used to validate

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/CompletionCodecTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/CompletionCodecTest.java
@@ -1,0 +1,162 @@
+package org.opensearch.migrations.bulkload;
+
+import java.io.File;
+import java.util.List;
+import java.util.Random;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.stream.Stream;
+
+import org.opensearch.migrations.VersionMatchers;
+import org.opensearch.migrations.bulkload.common.FileSystemRepo;
+import org.opensearch.migrations.bulkload.common.FileSystemSnapshotCreator;
+import org.opensearch.migrations.bulkload.common.OpenSearchClientFactory;
+import org.opensearch.migrations.bulkload.common.http.ConnectionContextTestParams;
+import org.opensearch.migrations.bulkload.framework.SearchClusterContainer;
+import org.opensearch.migrations.bulkload.http.ClusterOperations;
+import org.opensearch.migrations.bulkload.worker.SnapshotRunner;
+import org.opensearch.migrations.reindexer.tracing.DocumentMigrationTestContext;
+import org.opensearch.migrations.snapshot.creation.tracing.SnapshotTestContext;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.testcontainers.lifecycle.Startables;
+
+@Slf4j
+@Tag("isolatedTest")
+public class CompletionCodecTest extends SourceTestBase {
+    @TempDir
+    private File localDirectory;
+
+    private static Stream<Arguments> completionFieldScenario() {
+        return Stream.of(
+                Arguments.of(SearchClusterContainer.ES_V7_17, SearchClusterContainer.OS_V1_3_16),
+                Arguments.of(SearchClusterContainer.ES_V7_17, SearchClusterContainer.OS_V2_19_1),
+                Arguments.of(SearchClusterContainer.ES_V7_17, SearchClusterContainer.OS_V3_0_0),
+                Arguments.of(SearchClusterContainer.ES_V6_8_23, SearchClusterContainer.OS_V1_3_16),
+                Arguments.of(SearchClusterContainer.ES_V6_8_23, SearchClusterContainer.OS_V2_19_1),
+                Arguments.of(SearchClusterContainer.ES_V6_8_23, SearchClusterContainer.OS_V3_0_0),
+                Arguments.of(SearchClusterContainer.ES_V5_6_16, SearchClusterContainer.OS_V1_3_16),
+                Arguments.of(SearchClusterContainer.ES_V5_6_16, SearchClusterContainer.OS_V2_19_1),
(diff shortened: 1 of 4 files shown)


DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/CompletionCodecTest.java
```diff
@@ -0,0 +1,162 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.testcontainers.lifecycle.Startables;
+
+@Slf4j
+@Tag("isolatedTest")
+public class CompletionCodecTest extends SourceTestBase {
```
peternied, line 33: Please do not create a new set of tests like this for a operation that could be added onto our existing test cases, lets just add a new index with completion codec for the given version ranges.

peternied (changes requested):
While these tests are useful for confirming this scenario, this isn't a maintainable strategy for every feature we support to add ~5 minutes of test run time.  I'm pretty sure we can augment the existing tests cases that would handle this scenario.  What do you think of that approach?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -63,14 +66,15 @@ public void extendedMigrationDocuments(
...
+            migrationDocumentsWithClusters(sourceCluster, targetCluster, false);
         }
     }
 
     @SneakyThrows
     private void migrationDocumentsWithClusters(
         final SearchClusterContainer sourceCluster,
-        final SearchClusterContainer targetCluster
+        final SearchClusterContainer targetCluster,
+        final boolean includeCompletionFieldIndex
```
peternied, line 77: Lets always test with these fields, no need for an option
```suggestion
```

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -104,6 +108,18 @@ private void migrationDocumentsWithClusters(
             sourceClusterOperations.createIndex(indexName, body);
             targetClusterOperations.createIndex(indexName, body);
 
+            if (includeCompletionFieldIndex && sourceVersion != UnboundVersionMatchers.isBelowES_5_X) {
```
peternied, line 111: Move this check into a function `sourceSupportsCompletionFields()` and use that during the setup and verification steps, then we don't have to tweak the parameter generation logic

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -104,6 +109,19 @@ private void migrationDocumentsWithClusters(
             sourceClusterOperations.createIndex(indexName, body);
             targetClusterOperations.createIndex(indexName, body);
 
+            // Create and verify a 'completion' index only for ES 5.x and above
+            if (sourceSupportsCompletionFields(sourceVersion, targetVersion)) {
```
AndreKurait, line 113: sourceSupportsCompletionFields takes in targetVersion?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -104,6 +109,19 @@ private void migrationDocumentsWithClusters(
...
+                String completionIndex = "completion_index";
+                sourceClusterOperations.createIndexWithCompletionField(completionIndex, numberOfShards);
+                String completionDoc =
+                "{" +
+                "    \"completion\": \"openai\" " +
```
AndreKurait, line 118: let's have a more generic term here, `bananas`

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/version_9/IgnorePsmPostings.java
```diff
@@ -38,16 +41,14 @@ public FieldsConsumer fieldsConsumer(SegmentWriteState state) throws IOException
...
-                throw new UnsupportedOperationException(
-                    String.format(
-                        "Detected unsupported .psm file in segment [%s]. The index is using an unrecognized format.",
-                        state.segmentInfo.name
-                    )
-                );
-            }
+        String[] files = dir.listAll();
+
+        boolean foundPsm = Arrays.stream(files).anyMatch(f -> f.endsWith(".psm"));
```
AndreKurait, line 46: Why are we looking for these psm files only to ignore it?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
jugal-chauhan: will update

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/EndToEndTest.java
```diff
@@ -168,13 +186,25 @@ private void migrationDocumentsWithClusters(
...
+            int totalShards = supportsCompletion ? 2 * numberOfShards : numberOfShards;
+            Assertions.assertEquals(totalShards + 1, expectedTerminationException.numRuns);
 
             // Check that the docs were migrated
             checkClusterMigrationOnFinished(sourceCluster, targetCluster, testDocMigrationContext);
             boolean isSourceES1x = VersionMatchers.isES_1_X.test(sourceCluster.getContainerVersion().getVersion());
             boolean isTargetES1x = VersionMatchers.isES_1_X.test(targetCluster.getContainerVersion().getVersion());
 
+            if (sourceSupportsCompletionFields(sourceVersion, targetVersion)) {
+                var res = targetClusterOperations.get("/completion_index/_doc/1");
```
AndreKurait, line 199: let's move this logic out to a function for validation

PR #1719, opened 2025-07-31
Metadata output format support for json

peternied:
### Description
This change updates the CLI output classes so they can also include JSON output which is needed for communicating with other systems.

Example output with `--output json` flag:
```json
{
    "clusters": {
        "source": {
            "type": "MockRemoteReader",
            "version": "OPENSEARCH 54.0.0",
            "uri": "http://remote.source",
            "protocol": "HTTP",
            "insecure": true,
            "compressionSupported": false,
            "awsSpecificAuthentication": false
        },
        "target": {
            "type": "MockClusterWriter",
            "version": "OPENSEARCH 2000.0.0",
            "uri": "http://remote.target",
            "protocol": "HTTP",
            "insecure": false,
            "compressionSupported": false,
            "awsSpecificAuthentication": false
        }
    },
    "items": {
        "dryRun": true,
        "indexTemplates": [
           { "name": "it1", "successful": true },
           { "name": "it2", "successful": true }
        ],
        "componentTemplates": [
           { "name": "ct1", "successful": true },
           { "name": "ct2", "successful": true }
        ],
        "indexes": [
           { "name": "i1", "successful": true },
           { "name": "i2", "successful": true }
        ],
        "aliases": [
           { "name": "a1", "successful": true },
           { "name": "a2", "successful": true }
        ],
        "errors": []
    },
    "transformations": {
        "transformers": []
    },
    "errors": [],
    "errorCode": 0
}
```

### Issues Resolved
- Related [[MIGRATIONS-2593 | View session page]](https://opensearch.atlassian.net/browse/MIGRATIONS-2593)

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
```diff
@@ -29,13 +31,20 @@ public enum Protocol {
         HTTPS
     }
 
+    @JsonProperty("uri")
```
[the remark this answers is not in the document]
peternied: Dev guide shows an example of this, here is a snippet:

```
        "source": {
            "type": "MockRemoteReader",
            "version": "OPENSEARCH 54.0.0",
            "uri": "http://remote.source",
            "protocol": "HTTP",
            "insecure": true,
            "compressionSupported": false,
            "awsSpecificAuthentication": false
        },
```

Note; we don't really need this for the website, but it felt wrong not to have minimal parity with the existing CLI output.

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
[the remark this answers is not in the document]
peternied: There is a guard around the call to the function via `if (humanReadableOutput)`, this isn't pretty, but this logger is special only for this class to write to standard out, all other loggers write to the metadata run log.  I'll look into this if maybe I can do something a little cleaner.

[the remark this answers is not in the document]
peternied: Updated with better protections, lmk what you think.

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
```diff
@@ -102,7 +123,19 @@ public Migrate migrate(MigrateOrEvaluateArgs arguments) {
...
-    private static void printTopLevelHelp(JCommander commander) {
+    protected void writeOutput(String humanReadableOutput) {
+        if (outputFormat.get() == OutputFormat.HUMAN_READABLE) {
+            log.atInfo().setMessage(humanReadableOutput).log();
+        }
+    }
+
+    protected void writeOutput(JsonNode output) {
+        if (outputFormat.get() == OutputFormat.JSON) {
+            log.atInfo().setMessage(output.toPrettyString()).log();
```
AndreKurait, line 134: We need to modify these logs to be

```
log.atInfo().setMessage("{}").addArgument(output.toPrettyString()).log();
```

to prevent any log4j parsing of the output special characters (say, if it included `{}`)

Ideally, these `writeOutput` would take in a string supplier so the concatenation can be deferred.

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
```diff
@@ -22,7 +25,13 @@
 @Slf4j
```
AndreKurait, line 25: This would be a bit safer to remove `@Slf4j` or use a different logger for the console logging than the class one to ensure this type of logging is only done in the writeOutput method

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
peternied: Good call

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
peternied: I'd rather defer on this.  This is an abstract problem and I don't think we are going to be touching high-level metadata output for a while yet.  If we were investing in the CLI output this might be worthwhile to invest in more - but I think we are going to remain fairly static for a while and the website is were we will be pushing for more interaction and details via the existing json data.

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
```diff
@@ -102,7 +123,19 @@ public Migrate migrate(MigrateOrEvaluateArgs arguments) {
         return new Migrate(arguments);
     }
 
-    private static void printTopLevelHelp(JCommander commander) {
+    protected void writeOutput(String humanReadableOutput) {
+        if (outputFormat.get() == OutputFormat.HUMAN_READABLE) {
+            log.atInfo().setMessage(humanReadableOutput).log();
```
AndreKurait, line 128: Please add in the argument format here too for safety to align with the current behavior

AndreKurait (changes requested):

AndreKurait (approved):

PR #1729, opened 2025-08-04
Add automatic flattened to flat_object transformation

AndreKurait:
### Description
Add support to automatically resolve breaking change during ES 7.3+ to OS 2.7+ migrations by migrating `flattened` data type to flat_object https://docs.opensearch.org/latest/field-types/supported-field-types/flat-object/

### Testing
Added test

### Check List

### DIFF
--- a/transformation/src/main/java/org/opensearch/migrations/UnboundVersionMatchers.java
+++ b/transformation/src/main/java/org/opensearch/migrations/UnboundVersionMatchers.java
@@ -16,8 +16,12 @@ public class UnboundVersionMatchers {
     public static final Predicate<Version> isBelowES_8_X = belowMajorVersion(Version.fromString("ES 8.0.0"));
     public static final Predicate<Version> isGreaterOrEqualES_6_X = greaterOrEqualMajorVersion(Version.fromString("ES 6.0.0"));
     public static final Predicate<Version> isGreaterOrEqualES_7_X = greaterOrEqualMajorVersion(Version.fromString("ES 7.0.0"));
+    public static final Predicate<Version> isGreaterOrEqualES_7_3 = greaterOrEqualMajorVersion(Version.fromString("ES 7.3.0"));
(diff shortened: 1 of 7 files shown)


transformation/standardJavascriptTransforms/src/metadataUpdater.js
peternied, line 1: Want to give this a better name, what do you think of fieldTypeConverter?

transformation/standardJavascriptTransforms/src/metadataUpdater.js
```diff
@@ -46,6 +49,15 @@ function main(context) {
     );
   }
```
peternied, line 51: Rather than make a whole new block, check if context is a map and then do `context.get("rules")` or do `context.rules` to reduce the duplicated code.

peternied (approved):

transformation/src/main/java/org/opensearch/migrations/UnboundVersionMatchers.java
```diff
@@ -16,8 +16,12 @@ public class UnboundVersionMatchers {
     public static final Predicate<Version> isBelowES_8_X = belowMajorVersion(Version.fromString("ES 8.0.0"));
     public static final Predicate<Version> isGreaterOrEqualES_6_X = greaterOrEqualMajorVersion(Version.fromString("ES 6.0.0"));
     public static final Predicate<Version> isGreaterOrEqualES_7_X = greaterOrEqualMajorVersion(Version.fromString("ES 7.0.0"));
+    public static final Predicate<Version> isGreaterOrEqualES_7_3 = greaterOrEqualMajorVersion(Version.fromString("ES 7.3.0"));
```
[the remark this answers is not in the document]
AndreKurait: Yeah, flattened was added in 7.3. Would make sense to start thinking about a "FeatureRegistry" where we can encode all the versions specific features were added/removed that we have business logic on

PR #1742, opened 2025-08-05
Generate openapi spec via docker container

peternied:
### Description
Generate openapi spec via docker container

### Issues Resolved
 - Closes [[MIGRATIONS-2663 | Enable automatic generation of sdk client]](https://opensearch.atlassian.net/browse/MIGRATIONS-2663)

### Testing
 - [x] **Verified** Deploy an MA bootstrap box and make sure this image can be built.

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
@@ -1,0 +1,21 @@
+FROM python:3.11-slim
+
+# Set working directory
+WORKDIR /app
+
+# Copy dependency files
+COPY Pipfile Pipfile.lock ./
+
+# Install pipenv and dependencies
+RUN pip install pipenv && \
+    pipenv install --system --deploy
+
+# Copy source code
+COPY console_link/ ./console_link/
+COPY setup.py ./
+
+# Install the package in development mode
+RUN pip install -e .
+
+# Set the entrypoint to generate OpenAPI spec
+ENTRYPOINT ["python", "-m", "console_link.scripts.generate_openapi"]
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/build.gradle
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/build.gradle
@@ -26,7 +26,16 @@ tasks.register('buildOpenApiDockerImage') {
         def proc = ['docker', 'images', '-q', dockerImageName].execute()
         proc.waitFor()
         def imageId = proc.text.trim()
-        return !imageId.isEmpty()
+        // Check if Docker image exists, with error handling
+        try {
+            def proc = ['docker', 'images', '-q', dockerImageName].execute()
+            proc.waitFor()
+            def imageId = proc.text.trim()
+            return !imageId.isEmpty()
+        } catch (Exception e) {
+            logger.error("Failed to check if Docker image exists. Is Docker installed and running? Error: ${e.message}")
+            return false
+        }
     }
 
     doLast {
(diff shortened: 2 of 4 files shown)


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/build.gradle
```diff
@@ -7,36 +7,60 @@ def openapiOutFile = layout.buildDirectory.file("openapi.json")
...
+    inputs.files(pythonSources)
     inputs.files(dependencyFiles)
-    outputs.dir(venv)
+    outputs.upToDateWhen {
+        // Check if Docker image exists
+        def proc = ['docker', 'images', '-q', dockerImageName].execute()
+        proc.waitFor()
+        def imageId = proc.text.trim()
+        return !imageId.isEmpty()
+    }
```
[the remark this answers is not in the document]
peternied: This could be more robust, but it isn't needed since the python source files and dependency files are listed as inputs.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
AndreKurait, line 1: Is there a way we can use this without keeping this image around? So a docker build for output file (would use "RUN" instead of entrypoint) to reduce the size of the images we create on the docker context

```
docker build --output type=local,dest=./out .
```

peternied: Good idea, looking into this.

peternied: Done!

peternied: Unfortunately, this works on my dev machine but is broken on the bootstrap box, there isn't a clear reason why it doesn't work, just hangs :/ 

I'd recommend lets move forward with the existing functional state, and if we want to look at better image management we've got a wholistic issue rather than just this task/PR.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
```diff
@@ -0,0 +1,21 @@
+FROM python:3.11-slim
```
AndreKurait, line 1: Can we update this to Amazon Linux 2023
It has 0 vs 4 critical CVEs (10 vs 96 total)

Amazon Linux:
https://hub.docker.com/layers/library/amazonlinux/latest/images/sha256-939e9079a959d007ce849ddf6bcda893e342630438614ee26b43d1f461c2152a

Debian based python:3.11-slim
https://hub.docker.com/layers/library/python/3.11-slim/images/sha256-7ae2d10e4bdc6f69ba2daf031647568fec08f3191621d7a5c8760abb236d16ab?context=explore

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
AndreKurait: I can live without the docker build file output for this PR

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
peternied: When comparing architecture to architecture, I see 3 for both.  I don't see why I'd used amazonlinux - as oppose to the purpose built python3.11 image.

peternied: Furthermore, amazonlinux also only comes with python3.9 we'd need to bake installation of 3.11 into the image which seems like overkill vs using a standard image.

AndreKurait: We can keep this and keep an eye on rate limits on the bootstrap box

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/Dockerfile.openapi
AndreKurait: https://github.com/opensearch-project/opensearch-migrations/commit/f8561e3dec9968442c53bcb631a3431fe005fd3b

PR #1746, opened 2025-08-07
Set request timeout and shorten connection timeout for rest client

AndreKurait:
### Description

Reduce default RestClient connection and request timeout

We have seen some worker coordination client/server time out of sync. Upon investigation
this could occur during connection timeout issues since a reqeust could be delayed to be sent
until after a connection timesout (prior default 30s) which would cause a clock skew exception 
with the current value of 15 seconds. This change sets the value to 5 and further failure will delegate to 
the retry logic surrounding a call which would be able to reconstruct the request as needed.

Set request timeout and shorten connection timeout for rest client.
Prior default connection timeout is 30 seconds and prior to this PR there is no request timeout.
 
### Check List


gregschohn (commented):
I'd like to see a git log that shows why this change is being made.  There's no explanation in the commit history & no logs anywhere else.  W/out that, we risk thrashing back and forth on heuristic tweaks.

PR #1758, opened 2025-08-12
Disable using compression for source cluster

AndreKurait:
### Description
Disable using compression for source cluster.

This can help speed up client creation when the source cluster's cluster settings information is not available (e.g. Amazon Managed Service ES 1-5). Previously the request would retry a couple times with backoff before defaulting to compression off when the source is ES 1-5 on Amazon Managed OpenSearch Service

### Issues Resolved
N/A

### Check List


RFS/src/main/java/org/opensearch/migrations/bulkload/common/http/ConnectionContext.java
```diff
@@ -297,7 +297,8 @@ public static class SourceArgs implements IParams {
 
         @Override
         public boolean isDisableCompression() {
-            return false;
+            // No need to interrogate source cluster for compression information
+             return true;
```
peternied, line 301: Can we add a UT that would verify we aren't retrying this scenario described?

> Previously the request would retry a couple times with backoff before defaulting to compression off when the source is ES 1-5 on Amazon Managed OpenSearch Service

gregschohn (approved):

PR #1762, opened 2025-08-13
Create test to be used for Delta Snapshot Restore

AndreKurait:
Signed-off-by: Andre Kurait <akurait@amazon.com>


RFS/src/test/java/org/opensearch/migrations/bulkload/lucene/LiveDocsConverterTest.java
```diff
@@ -0,0 +1,315 @@
...
+        assertTrue(result.get(3));
+        assertFalse(result.get(4));
+        assertFalse(result.get(5));
+        assertFalse(result.get(6));
+        assertTrue(result.get(7));
+        assertFalse(result.get(8));
+        assertTrue(result.get(9));
+        
+        // Verify that getBits() was called (fast path for FixedBitSet)
+        verify(spyFixed, times(1)).getBits();
```
AndreKurait, line 67: This is important we're verifying this, since all uses in ES/OS should use FixedBitSet versus the other implementations so we want to ensure we're using the fast path for the copies

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LiveDocsConverter.java
```diff
@@ -0,0 +1,145 @@
...
+            getBitsFromFixed,
+            getLength,
+            getBitAt,
+            nextSetBitFactory
+        );
+        var convertedBitSet = converter.convert(liveDocs);
+        return convertedBitSet != null ? new LengthDisabledBitSet(convertedBitSet) : null;
+    }
+
+    public static class LengthDisabledBitSet extends BitSet {
```
AndreKurait, line 132: Open to suggestions here. I was surprised to find out that the bitset.length() was not the length i created the bitset with (it's one more than the highest enabled bit). So i made this wrapper to disable calls to length while delegating for all the other methods

PR #1153, opened 2024-11-20
Hardcode AMIs for use in the bootstrap machine

peternied:
### Description
With the current latest AL2023 image `al2023-ami-2023.6.20241111.0-kernel-6.1-x86_64` npm install hangs indefinitely.  We are switching back to an older version that works and providing a way to incrementally update to AMIs that we have tested.

### Issues Resolved
- initBootstrap.sh hangs on npm install tasks https://opensearch.atlassian.net/browse/MIGRATIONS-2227
- Related https://github.com/amazonlinux/amazon-linux-2023/issues/840

### Testing
Deployed the updated cloudformation manually and verified the build succeeds.

### Check List
- [ ] New functionality includes testing
  - [ ] All tests pass, including unit test, integration test and doctest
- [X] New functionality has been documented
- [X] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1767, opened 2025-08-14
feat(migration-assitant-solution): fetch latest AMI from SSM parameter store

spacelatte:
This is more of a refactoring. During a recent change, I realised the AMI IDs
referenced (generated by the `create-ami-map.sh`) were quite outdated.
(ie. The image was deprecated)

As a solution, I refactored this code to use SSM Parameter store to fetch
respective AMI ID. As those parameters are public, and contains the current
AMI IDs, this will always fetch the latest AMI ID.

This change will also remove the need for the `create-ami-map.sh` script.

These are documented in the AL2023 EC2 documentation as well as the SSM parameter store one:
- https://docs.aws.amazon.com/linux/al2023/ug/ec2.html
- https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-public-parameters-ami.html

Signed-off-by: Mert Akengin <5755154+spacelatte@users.noreply.github.com>

### Check List


spacelatte (commented):
**Testing:** Not sure how can I _effectively_ test this changes. Thought of maybe including a test case which validates a non-empty AMI specification in the Instance/LaunchTemplate.

peternied (changes requested):
I appreciate the change; however, I don't think we would adopt it because it introduces another parameter that might be hard to explain for our customers.  We've seen a lot of customer pain around correctly setting and troubleshooting incorrect parameters. 

More context when this change was introduced:
- https://github.com/opensearch-project/opensearch-migrations/pull/1153

PR #1819, opened 2025-09-11
Add OTEL metrics and traces to EKS CloudWatch

lewijacn:
### Description
This change modifies the Migration Assistant Helm chart to setup OTEL metrics and traces to be sent to CloudWatch in EKS. It also wires the otel-collector endpoint through the `full-migration` Argo workflow so that application metrics are being sent to OTEL. The one stipulation to this is the metadata-migration Argo workflow which still expects the otel-collector endpoint to come from the top level migration config, as it currently has a tight affinity with the `services.yaml` config, but I suspect this structure will change in the near future.

Additionally this change:
1. Modifies the `aws-bootstrap.sh` script to use public MA images as a default
2. Keeps the service monitor up for OTEL, so that Prometheus is now able to collect the captured OTEL metrics

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2605

### Testing
Local and EKS testing

### Check List

### DIFF
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -278,7 +278,7 @@ helm install "$namespace" "${ma_chart_dir}" \
   --set stageName="${STAGE}" \
   --set aws.region="${AWS_CFN_REGION}" \
   --set aws.account="${AWS_ACCOUNT}" \
-  "$IMAGE_FLAGS" \
+  $IMAGE_FLAGS \
   || { echo "Installing Migration Assistant chart failed..."; exit 1; }
 
 kubectl -n "$namespace" wait --for=condition=ready pod/migration-console-0 --timeout=300s
--- a/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/DEVELOPER_GUIDE.md
+++ b/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/DEVELOPER_GUIDE.md
@@ -2,12 +2,11 @@
 Guidance for developing components of the Migration Assistant helm chart
 
 
-
 ### Configuring OpenTelemetry (OTEL) metrics and traces
 The OTEL configuration for this chart can be modified by changing the `otelConfiguration` value this chart defines. It includes a `collectorConfig` value which is the OTEL core configuration, and allows the user to update aspects like which processors and exporters should be used.
 
 
-#### AWS X-Ray
+#### AWS X-Ray traces
 Traces, as a default, are not enabled for the EKS deployment configured by `valuesEks.yaml` in this directory. The following reference could be used to modify these values and enable tracing to AWS X-Ray
 ```yaml
     processors:


TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
```diff
@@ -104,7 +105,8 @@ def prepare_workflow_snapshot_and_migration_config(self):
         snapshot_and_migration_configs = [{
             "migrations": [{
                 "metadata": {
-                    "from_snapshot": None
+                    "from_snapshot": None,
+                    "otel_endpoint": OTEL_COLLECTOR_ENDPOINT
```
gregschohn, line 109: Sorry if  this contradicts anything that I've said before - but shouldn't there be just one collector endpoint for the whole system (workflow)?  If that's the case, we'll just push it down through the workflow with each process.
I don't think that there's anything necessarily to change right now for this PR since https://github.com/opensearch-project/opensearch-migrations/pull/1633 has work to send configurations through different layers much more elegantly.  We'd just need to update where we set the value in the config...

But should we even require setting the value within the config or should it come from a configmap that was set at helm install-time (and available to this test).

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/workflows/templates/createSnapshot.yaml
```diff
@@ -21,7 +22,7 @@ spec:
             value: ""
             description: "Which indices should be snapshot.  Default empty.  Empty string means all."
           - name: snapshot-config
-            value: "{ \"snapshot_name\": \"{{inputs.parameters.snapshot-name}}\", \"s3\": {{inputs.parameters.s3-config}} }"
+            value: "{ \"snapshot_name\": \"{{inputs.parameters.snapshot-name}}\", \"otel_endpoint\": \"{{inputs.parameters.otel-collector-endpoint}}\" , \"s3\": {{inputs.parameters.s3-config}} }"
```
gregschohn, line 25: regardless of anything else, why would this be part of snapshot-config?  IIRC, this value is passed along later on to RFS.  Why would RFS want to know the otel-endpoint?  Again, parameter passing is changing considerably, so I don't think that any changes will be necessary because we'll refine it in https://github.com/opensearch-project/opensearch-migrations/pull/1633.

buildImages/README.md
```diff
@@ -34,5 +34,5 @@ The following gradle command can be used after setting up the previous steps to
 
 Or customized to use a specific registry endpoint and architecture
 ```shell
-./gradlew buildImagesToRegistry -PregistryEndpoint=123456789012.dkr.ecr.us-west-2.amazonaws.com/my-ecr-repo -PimageArch=arm64
+./gradlew buildImagesToRegistry -PregistryEndpoint=123456789012.dkr.ecr.us-west-2.amazonaws.com/my-ecr-repo -PimageArch=amd64
```
gregschohn, line 37: What H/W do we deploy by default, amd64?  If so, this makes sense.

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -22,9 +22,13 @@ base_dir="./opensearch-migrations"
 build_images_chart_dir="${base_dir}/deployment/k8s/charts/components/buildImages"
 ma_chart_dir="${base_dir}/deployment/k8s/charts/aggregates/migrationAssistantWithArgo"
 namespace="ma"
-skip_image_build=false
+skip_image_build=true
```
gregschohn, line 25: It would be nice if there was a command line option here - or what if we just had separate scripts & maybe another script to tie these things together.

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -22,9 +22,13 @@ base_dir="./opensearch-migrations"
...
+use_public_images=true
```
gregschohn, line 26: Should this be set to false when skip_image_build=false?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/otel/simpleCollectorAsIndependentDaemonset.yaml
```diff
@@ -24,6 +24,10 @@ spec:
               value: "true"
             - name: OTEL_LOG_LEVEL
               value: "debug"
+            - name: QUALIFIER
+              value: {{ .Values.stageName }}
+            - name: TRACE_SAMPLING_PERCENTAGE
+              value: "1"
```
gregschohn, line 30: is this 100%, 1%?  I think that we should turn all traces off.  There's a severe performance degradation risk with them being on & customers have never asked about traces, nor have we ever documented them.
This isn't different from our ECS story - but I'd like to eliminate the need to do helm redeployments & I'd rather have people need to redeploy for more debugging info rather than redeploying to fix a performance issue.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/otel/simpleCollectorAsIndependentDaemonset.yaml
```diff
@@ -61,55 +65,8 @@ metadata:
...
-          exporters: [ prometheus ]
-        traces:
-          receivers: [ otlp ]
-          processors: [ batch ]
-          exporters: [ otlp/jaeger ]
-        logs:
-          receivers: [ otlp ]
-          processors:
-          exporters: [ debug ]
+{{ .Values.otelConfiguration.collectorConfig | indent 4 }}
```
gregschohn, line 68: I think that this makes sense, given the deployment architecture of the OTEL collector endpoint.  We should document how customers can tweak this afterward (either redeploy or by altering a configmap)

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/createS3Bucket.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
     "helm.sh/hook-delete-policy": hook-succeeded
 spec:
   ttlSecondsAfterFinished: 600
-  backoffLimit: 6
+  backoffLimit: 2
```
gregschohn, line 12: Why adjust this?  3 total tries seems reasonable (is this 2 or 3 total?).
What's the failure mode overall when this step fails?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/deleteS3Bucket.yaml
```diff
@@ -100,7 +100,6 @@ spec:
                       echo "Tip: Set defaultBucketConfiguration.emptyBeforeDelete=true to force-delete all objects before bucket deletion."
                     fi
                   fi
-                  exit 1
```
gregschohn, line 103: Does this return with any error (non-zero) code?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/valuesEks.yaml
```diff
@@ -25,6 +25,69 @@ defaultBucketConfiguration:
...
+      awsxray:
+        index_all_attributes: true
+      prometheus:
+        endpoint: "0.0.0.0:8889"
+        send_timestamps: true
+        metric_expiration: 5m
+        enable_open_metrics: true
+      otlp/jaeger: # Jaeger supports OTLP directly. The default port for OTLP/gRPC is 4317
+        endpoint: jaeger-collector:4317
+        tls:
```
gregschohn, line 77: This should be disabled for AWS

gregschohn (commented):
Only one big comment which is around tracing for EKS - don't include jaeger in the config & consider disabling xray too.
Other still is optional or a check on how we'll want the final otel-collector endpoint configuration to look like once we merge 1633 (Argo-DSL)

PR #1846, opened 2025-09-25
[Fix] ES8VectorFieldMappingsTransformationTest to have no replicas

jugal-chauhan:
### Description
This test was failing in CI due to unassigned primary shards preventing snapshot creation. The error occurred because the test created an index without specifying replica settings, causing Elasticsearch to default to 1 replica in a single-node cluster.

This PR explicitly adds 0 replicas to the index settings to ensure all shards can be assigned in the single-node test.environment.

### Issues Resolved
No Jira

### Testing
Local : ./gradlew :MetadataMigration:isolatedTest --tests "ES8VectorFieldMappingsTransformationTest"
Also validating with all GHA checks

### Check List

### DIFF
--- a/MetadataMigration/src/test/java/org/opensearch/migrations/ES8VectorFieldMappingsTransformationTest.java
+++ b/MetadataMigration/src/test/java/org/opensearch/migrations/ES8VectorFieldMappingsTransformationTest.java
@@ -57,6 +57,9 @@ class ES8VectorFieldMappingsTransformationTest extends BaseMigrationTest {
         startClusters();
 
         String requestBody = "{\n" +
+                "  \"settings\": {\n" +
+                "    \"number_of_replicas\": 0\n" +
+                "  },\n" +
                 "  \"mappings\": \n" +
                 "    {\n" +
                 "        \"dynamic\": \"strict\",\n" +


RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
```diff
@@ -160,7 +160,7 @@ public enum INITIALIZATION_FLAVOR {
             "ES_JAVA_OPTS", "-Xms2g -Xmx2g",
             "cluster.routing.allocation.disk.watermark.low", "99%",
             "cluster.routing.allocation.disk.watermark.high", "99%",
-            "cluster.routing.allocation.disk.watermark.flood_stage", "99%"
+            "cluster.routing.allocation.disk.watermark.flood_stage", "99.99%"
```
AndreKurait, line 163: What about the low/high watermark? Should we just set all of these to 100% to disable this behavior during these tests?

jugal-chauhan: Yes, increased all watermark level settings to 100% for all images

PR #1863, opened 2025-10-04
Command line JCommander parameters parsed via Environment Variables

gregschohn:
### Description
For the Argo workflows/k8s resources, it's easier to configure environment variables than command line arguments.  Arguments are a list, and it's hard to say "this item is empty" because that won't be coalesced.  It's easier to say - this env var just isn't set (or is set to empty, in which case just ignore it).

I still need to go back and confirm all of the edge cases - when arity is not equal to 1 and when the value is empty (because there wasn't a value from the workflow)

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2634

### Testing
No new tests, but this will be well tested with integ tests.

### Check List


gregschohn:
See https://github.com/opensearch-project/opensearch-migrations/pull/1883

PR #1876, opened 2025-10-09
Add workflow submit, approve, status, stop commands

AndreKurait:
### Description
Implements workflow cli operations `status`, `approve`, `stop`, `status` with integration with the orchestrationSpecs migration workflow and test workflow for quick developer validation.

See UX in videos.

### Issues Resolved

[MIGRATIONS-2702](https://opensearch.atlassian.net/browse/MIGRATIONS-2702)
[MIGRATIONS-2703](https://opensearch.atlassian.net/browse/MIGRATIONS-2703)
[MIGRATIONS-2705](https://opensearch.atlassian.net/browse/MIGRATIONS-2705)
[MIGRATIONS-2706](https://opensearch.atlassian.net/browse/MIGRATIONS-2706)

### Testing
Tested mostly with placeholder test workflow, with some validation of the orchestrationSpecs workflow in eks.

For UX, see:

https://github.com/user-attachments/assets/52233ac3-3ccb-4899-98d6-7e092c386a32

https://github.com/user-attachments/assets/b5b11702-ba2e-45e7-819e-5d91e33258e0

https://github.com/user-attachments/assets/ef20abf1-e994-41d2-bc84-8846f63d6924

### Check List

### DIFF
--- a/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
+++ b/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
@@ -129,14 +129,6 @@ RUN echo '. /etc/profile.d/bash_completion.sh' >> ~/.bashrc && \
     echo 'workflow util completions bash > /usr/share/bash-completion/completions/workflow' >> ~/.bashrc && \
     echo 'PS1="(\t) \[\e[92m\]migration-console \[\e[0m\](\w) -> "' >> ~/.bashrc
 
-# Install YQ
-RUN set -eux; \
-  arch="$(uname -m)"; \
-  case "$arch" in x86_64) yq_arch=amd64 ;; aarch64|arm64) yq_arch=arm64 ;; *) echo "unsupported arch: $arch" >&2; exit 1 ;; esac; \
-  curl -fsSL "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_${yq_arch}" -o /usr/local/bin/yq; \
-  chmod +x /usr/local/bin/yq; \
-  /usr/local/bin/yq --version
-
 COPY nodeStaging/configProcessor /root/configProcessor
 
 # Copy test workflow scripts and resources with same structure as configProcessor


orchestrationSpecs/packages/config-processor/scripts/sampleMigration.yaml
```diff
@@ -12,8 +12,7 @@
       "snapshotRepo": {
         "aws_region": "us-east-2",
         "endpoint": "localstack://localstack.ma.svc.cluster.local:4566",
-        "s3RepoPathUri": For localstack testing, replace with `aws --endpoint-url=http://$(kubectl get svc localstack -o jsonpath='{.spec.clusterIP}'):4566 s3 ls`
-      },
+        "s3RepoPathUri": For localstack testing, replace with `aws --endpoint-url=http://$(kubectl get svc localstack -o jsonpath='{.spec.clusterIP}'):4566 s3 ls`      },
```
gregschohn: nit - spacing

orchestrationSpecs/packages/config-processor/src/RunMigrationInitializer.ts
```diff
@@ -59,7 +59,6 @@ async function main() {
     let etcdEndpoints = process.env.ETCD_ENDPOINTS;
     let etcdUser = process.env.ETCD_USER;
     let etcdPassword = process.env.ETCD_PASSWORD;
-    let prefix = process.env.PREFIX;
```
gregschohn: nit: why drop this & not the others?  I've been setting up env variables because in some contexts, those can be easier to set up on containers than args (e.g. what if the prefix were in a configmap).

orchestrationSpecs/packages/config-processor/sample.yaml
```diff
@@ -0,0 +1,14 @@
+# Sample workflow configuration
```
gregschohn: Can this file be removed - or moved elsewhere.  There's already another sampleMigration.yaml file anc we have a limited ability to create a sample from the schema.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+        workflow approve my-workflow
+        workflow approve --acknowledge
+        workflow approve --argo-server https://10.105.13.185:2746 --insecure
+    """
+
+    try:
+        service = WorkflowService()
+
+        # Auto-detect workflow if name not provided
+        if not workflow_name:
```
gregschohn: Should this block be a common function - maybe in workflowService or somewhere else?  Won't you need it for every workflow command?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/workflow_service.py
```diff
@@ -0,0 +1,860 @@
...
+            nodes: Dictionary of workflow nodes
+
+        Returns:
+            List of step dictionaries with name, phase, type, and started_at
+        """
+        steps = []
+        for node in nodes.values():
+            node_type = node.get("type", "")
+            # Include Pod and Suspend steps, plus Skipped steps to show conditional logic
+            if node_type in ["Pod", "Suspend", "Skipped"]:
```
gregschohn: What do you want to filter out?  You'll filter out Resources, HttpRequests, Steps & Dags themselves.  What do you need to do this processing for?  For approve, you could just apply a filter on the whole set of nodes. I suspect that a lot of other usages might be like that too.  As it is now, some things might be skipped or otherwise never show up if they weren't skipped in a different run, which could be odd - making something not-skipped look like it was never even created at all.

I wouldn't call these steps - that's a very specific argo term.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+                step_name = step['name']
+                step_phase = step['phase']
+
+                # Collect suspended steps for approval selection
+                if step_type == 'Suspend' and step_phase == 'Running':
+                    suspended_steps.append(step_name)
+                    click.echo(f"  > {step_name} [{step_type}] - {step_phase} (WAITING FOR APPROVAL)")
+                else:
+                    status_icon = "+" if step_phase == "Succeeded" else "*"
+                    click.echo(f"  {status_icon} {step_name} [{step_type}] - {step_phase}")
```
gregschohn: why is approve showing a bunch of other results - this will include items that aren't suspends too - right?  This might make it harder to convert the overall output into a machine readable format that's easily consumable downstream.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+                click.echo(f"\nApproving step: {selected_step}")
+            except ValueError:
+                click.echo(
+                    f"Error: Invalid input. Please enter a number (0-{len(suspended_steps) - 1}) or 'c'",
+                    err=True
+                )
+                ctx.exit(ExitCode.FAILURE.value)
+
+        # Resume the workflow
+        result = service.approve_workflow(
```
gregschohn: if there are 3 suspensions, this will approve all of them?  Maybe its flag should include the word 'all'

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+                ctx.exit(ExitCode.SUCCESS.value)
+
+            # Validate numeric choice
+            try:
+                choice_idx = int(choice)
+                if choice_idx < 0 or choice_idx >= len(suspended_steps):
+                    click.echo(f"Error: Invalid choice. Please select 0-{len(suspended_steps) - 1} or 'c'", err=True)
+                    ctx.exit(ExitCode.FAILURE.value)
+
+                selected_step = suspended_steps[choice_idx]
```
gregschohn: are you missing filtering the resume down to just one node?  I expected to see resume called with an node identifier.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/configure.py
```diff
@@ -197,3 +187,40 @@ def clear_config(ctx, confirm):
     except Exception as e:
         logger.exception(f"Failed to clear configuration: {e}")
         raise click.ClickException(f"Failed to clear configuration: {e}")
+
+
+@configure_group.command(name="sample")
+@click.option('--format', type=click.Choice(['yaml', 'json']), default='yaml',
+              help='Output format')
+@click.option('--load', is_flag=True, help='Load sample into current session')
```
gregschohn: If you made load and show different commands, it might make it simpler to know what happens when you don't load (will the user think that something has happened when they type `...sample`?).  I also would like to kick the user right into an editting session once they load the sample.  If they just wanted to store it, they could run ...sample | configure store.

I don't want the user to run sample and forget to come back to it - dropping them into an editor immediately forces them to start taking action before committing it.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+    is_flag=True,
+    default=False,
+    help='Skip TLS certificate verification'
+)
+@click.option(
+    '--token',
+    help='Bearer token for authentication'
+)
+@click.option(
+    '--all',
```
gregschohn: Is there a command to show me the list of all the workflows, present and historical?  Can we easily make tab-complete expand with the list of workflows - maybe with a --previous NAME option?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+                                workflow_statuses.append(wf_result)
+
+                        # Sort by finished_at timestamp (most recent first)
+                        if workflow_statuses:
+                            workflow_statuses.sort(key=lambda x: x['finished_at'], reverse=True)
+                            _display_workflow_status(workflow_statuses[0])
+                        elif completed_result['workflows']:
+                            # Fallback to first workflow if no finish times available
+                            result = service.get_workflow_status(
+                                workflow_name=completed_result['workflows'][0],
```
gregschohn: Are these returned in sorted order?  Will they always be?
The auto-discovery/scanning pattern is making me anxious.  I'd like to see us use something like a label or even a configmap.  We could think about various ways to drive a transaction to keep the current/last workflow easily discoverable.
Can you submit a README file alongside this code that explains the workflow/state machine that we're using to manage workflows.  I want to make sure that we aren't missing anything.
One tool that could be useful would be [argo mutexes](https://argo-workflows.readthedocs.io/en/latest/synchronization/).  We might also want to consider calling the full-migration workflow within another workflow that does some other bookkeeping (including acquiring a mutex).

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+    elif step_phase == 'Running':
+        return '  |' if step_type == 'Suspend' else '  *'
+    elif step_phase in ('Failed', 'Error'):
+        return '  -'
+    elif step_phase == 'Pending':
+        return '  >'
+    elif step_phase == 'Skipped':
+        return '  ~'
+    else:
+        return '  ?'
```
gregschohn: Do these need to be a single character?  How about 
```
    if step_phase == 'Succeeded':
        return '  +'
    elif step_phase == 'Running':
        return '  Paused' if step_type == 'Suspend' else '  *'
    elif step_phase in ('Failed', 'Error'):
        return '  x'
    elif step_phase == 'Pending':
        return '  Running'
    elif step_phase == 'Skipped':
        return '  Skipped'
    else:
        return '  ?'
```

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+    Returns:
+        Symbol character for the phase
+    """
+    phase_symbols = {
+        'Running': '*',
+        'Succeeded': '+',
+        'Failed': '-',
+        'Error': '-',
+        'Pending': '>',
+        'Stopped': 'X',
```
gregschohn: Can we use words here and below?  It's hard for me to understand what's the difference between X and - and _.  If somebody sees one of these for the first time after some familiarity with the tools, it will be jarring.  Same if we'd ever want to add new states.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+        return f"[{color}]{symbol} {step_name} ({step_phase})[/{color}]"
+
+
+def _display_workflow_tree(step_tree: list):
+    """Display workflow steps in tree format using Rich.
+
+    Args:
+        step_tree: List of WorkflowNode dictionaries with hierarchy
+    """
+    if not step_tree:
```
gregschohn: where/how does this get constructed?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+    help='Bearer token for authentication'
+)
+@click.option(
+    '--acknowledge',
+    is_flag=True,
+    default=False,
+    help='Skip interactive confirmation and approve immediately'
+)
+@click.pass_context
+def approve_command(ctx, workflow_name, argo_server, namespace, insecure, token, acknowledge):
```
gregschohn: Especially for testing purposes, it would be nice to be able to send a single suspended task name in to approve.  I could see a user getting the status tree as json & walking it for the task to pass in to this.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+                # Find parent and add as child
+                parent_id = node.get('parent')
+                if parent_id and parent_id in node_to_tree:
+                    node_to_tree[node['id']] = node_to_tree[parent_id].add(label)
+                else:
+                    # Fallback: add to root if parent not found
+                    node_to_tree[node['id']] = tree.add(label)
+
+    click.echo("")
+    console.print(tree)
```
gregschohn: I like the tree output.  I'd also like to see json output so that I could do some jpath (jq) queries to get names that I could then check for or do other scripting with.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/script_runner.py
```diff
@@ -0,0 +1,391 @@
...
+    def __init__(self, script_dir: Optional[Path] = None):
+        """
+        Initialize with script directory path.
+
+        Args:
+            script_dir: Optional path to scripts. If None, uses CONFIG_PROCESSOR_DIR environment variable.
+                       Raises ValueError if neither is provided or if the directory doesn't exist.
+        """
+        if script_dir is None:
+            config_processor_dir = os.environ.get('CONFIG_PROCESSOR_DIR')
```
gregschohn: nit: It might be better to not have a default script_dir - or maybe break this class into two - one to run scripts and another to run scripts from a specific place.  I don't want to need to put many different kinds of scripts all into the same directory.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/script_runner.py
```diff
@@ -0,0 +1,391 @@
...
+        blank starter configuration template instead.
+
+        Returns:
+            YAML content as string (either from sample.yaml or blank starter)
+
+        Raises:
+            IOError: If sample.yaml exists but cannot be read
+        """
+        logger.info("Getting sample configuration")
+        sample_path = self.script_dir / "sample.yaml"
```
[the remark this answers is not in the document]
gregschohn: I'm still annoyed that I can't figure out how to comment/uncomment a region in emacs like I can in vscode or intellij (my meta key doesn't seem to be bound properly).

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/script_runner.py
```diff
@@ -0,0 +1,391 @@
...
+        try:
+            with open(sample_path, 'r') as f:
+                content = f.read()
+            logger.debug(f"Loaded sample config from {sample_path} ({len(content)} bytes)")
+            return content
+        except IOError as e:
+            logger.error(f"Failed to read sample configuration: {e}")
+            raise
+
+    def transform_config(self, config_data: str) -> str:
```
gregschohn: Why do you need this function?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/script_runner.py
```diff
@@ -0,0 +1,391 @@
...
+        logger.info(f"Submitting workflow with prefix: {prefix}, namespace: {namespace}")
+
+        script_path = self.script_dir / "createMigrationWorkflowFromUserConfiguration.sh"
+
+        if not script_path.exists():
+            raise FileNotFoundError(f"Workflow submission script not found: {script_path}")
+
+        # Create temporary file with config data
+        with tempfile.NamedTemporaryFile(mode='w', suffix='.yaml', delete=False) as temp_file:
+            temp_file.write(config_data)
```
gregschohn: FWIW, both the initializer node script & the shell script can handle stdin via '-', IIRC

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/workflow_service.py
```diff
@@ -0,0 +1,860 @@
...
+            WorkflowListResult dict with success status, workflows list, count, and error
+        """
+        try:
+            headers = self._prepare_headers(token)
+            url = f"{argo_server}/api/v1/workflows/{namespace}"
+
+            logger.info(f"Listing workflows in namespace {namespace} (exclude_completed={exclude_completed})")
+            logger.debug(f"List request URL: {url}")
+
+            response = requests.get(url, headers=headers, verify=not insecure)
```
gregschohn: I'm not sure if argo can do the same sort of filtering on metadata that k8s does.  Other than consistency, is there any reason that you would need to continue using argo directly to get workflows?  I suspect that it might be less impactful to do argo.  Just for my own knowledge.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
```diff
@@ -128,8 +129,24 @@ RUN echo '. /etc/profile.d/bash_completion.sh' >> ~/.bashrc && \
     echo 'workflow util completions bash > /usr/share/bash-completion/completions/workflow' >> ~/.bashrc && \
     echo 'PS1="(\t) \[\e[92m\]migration-console \[\e[0m\](\w) -> "' >> ~/.bashrc
 
+# Install YQ
```
gregschohn: nit: python can do yq transforms pretty well too.  yq isn't as ubiquitous as jq or python.

gregschohn (commented):
I've reviewed everything but the tests.  Let's find some time to walk through these and figure out which comments make sense, which don't, and anything that we might want to change - or open up as a ticket - before merging.

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
```diff
@@ -128,8 +129,25 @@ RUN echo '. /etc/profile.d/bash_completion.sh' >> ~/.bashrc && \
...
 COPY nodeStaging/configProcessor /root/configProcessor
 
+# Copy test workflow scripts and resources with same structure as configProcessor
+COPY lib/console_link/tests/workflow-tests/resources/scripts /root/testConfigProcessor
+RUN chmod +x /root/testConfigProcessor/*.sh
+RUN chmod +x /root/configProcessor/*.sh
+
+# To use test workflows, flip the comments
+ENV CONFIG_PROCESSOR_DIR=/root/configProcessor
+#ENV CONFIG_PROCESSOR_DIR=/root/testConfigProcessor
```
gregschohn: why is there a shadow test variant here?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
gregschohn: What do you need yq for?

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+    is_flag=True,
+    default=False,
+    help='Skip TLS certificate verification'
+)
+@click.option(
+    '--token',
+    help='Bearer token for authentication'
+)
+@click.option(
+    '--acknowledge',
```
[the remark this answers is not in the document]
AndreKurait: Follow up, will update in the future to be `approve all`

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
AndreKurait: Future improvement to move this to a `getSession` function that is applied for all the workflow commands

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -0,0 +1,205 @@
...
+            click.echo(f"Started: {status_result['started_at']}")
+        click.echo("=" * 60)
+
+        # Display steps and collect suspended steps
+        suspended_steps = []
+        if status_result['steps']:
+            click.echo("\nWorkflow Steps:")
+            click.echo("-" * 60)
+            for step in status_result['steps']:
+                step_type = step['type']
```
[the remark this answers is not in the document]
AndreKurait: Future, add comment around user facing experience expected here

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
AndreKurait: When we are printing we are logging the notation, the word, and the color

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
AndreKurait: Possibly right now, let's add a followup task to handle workflows with multiple concurrent suspensions across all commands (status, approve, etc.)

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/configure.py
AndreKurait: We can remove load

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -0,0 +1,376 @@
...
+    """Display workflow steps in tree format using Rich.
+
+    Args:
+        step_tree: List of WorkflowNode dictionaries with hierarchy
+    """
+    if not step_tree:
+        return
+
+    console = Console()
+    tree = Tree("📋 [bold]Workflow Steps[/bold]")
```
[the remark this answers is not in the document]
AndreKurait: removed

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
AndreKurait: todo on future work to better test and handle parallels

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/console_link/workflow/services/script_runner.py
```diff
@@ -0,0 +1,391 @@
...
+            output_prefix = result.stdout.strip()
+            logger.info(f"Workflow initialized with prefix: {output_prefix}")
+            return output_prefix
+
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Workflow initialization failed with exit code {e.returncode}")
+            logger.error(f"stderr: {e.stderr}")
+            raise
+
+    def submit_workflow(
```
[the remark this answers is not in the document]
AndreKurait: i've rewritten to run createMigrationWorkflowFromUserConfiguration

TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile
AndreKurait: This is a simple workflow which let's you run through the workflow commands without having a source/target cluster setup

PR #1888, opened 2025-10-15
Changes to properly run RFS in the argo workflow (w/ all options)

gregschohn:
### Description

Typecheck that manifests are properly defined.  There are a couple new functions (makeStringTypeProxy/makeDirectTypeProxy) to convert BaseExpressions of either type string or boolean|number to the flat (non Expression) TYPE (for manifest agreement) and expression to wrap the value in an indicator that will be used to exclude putting the value into quotes during rendering.  Notice that the way that that's currently working is somewhat hacky because I'm using an off-the-shelf yaml renderer, so I need to use a regex.  It would be less brittle to emit the yaml directly, notice that tag & proceed accordingly.

K8s resources themselves are now being imported from the cluster (via kubectl) through a prebuild script (scripts/fetchK8sSchemas.sh and generateK8sTypes.ts) for argo-workflow-builders.  That pulls the k8s schemas and then converts them into TS types, scraping the names of resources so that it's pretty convenient to refer to containers, replicasets, etc.  Notice that fetching the k8s schemas is done manually now - how that gets plumbed in automatically (& what the final solution is for the whole workflow template distribution) is still TBD.  A couple new devDependencies have come into the workflow builders project to support the type construction.

Output has been switched from json to yaml.  There's a nasty impossibility, as far as hours of testing have shown, for argo manifests and interpolation (aka substitution).  In YAML, I can write `foo: 1`, `foo: hello`, or `foo: {...`.  In JSON, if I want to write an expression, the only way that I can do it is to wrap it in quotes because '{' is an obviously special character.  Argo's YAML parsers seem to accept '{{...' sequences as values, so Argo happily accepts the YAML, leaves the quotes away, then the interpolation is performed, which may result in a boolean or number, with no quotes - leaving the final YAML form correct.  Unfortunately, if we try to do this in JSON, we have to wrap the expression in a string.  Even after interpolation, those quotes are still there and numbers/booleans look like strings instead of the correct types as per the manifest, causing resource deployment failures.  SO - everything was switched to use YAML output.  Quotes are prevented with a hack that eats sentinel bookends around serialized content.  As mentioned above, emitting yaml as we walk the tree would make this safer, but emitting yaml is painful.

nullCoalesce has been removed because I couldn't figure out any way to get that function to work.  I tried just nullCoalesce, sprig.nullCoalesce.  I couldn't find any examples - and simple isEmpty checks work just as well for where we were using them.  In a similar spirit, MISSING_FIELD was removed as a possibility for the migration console's config values.  Now those are optional variables & everything just works.  Type somersaults no longer need to be done.  That all became noticeable when I had to start parsing input fields for numbers and booleans.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2620

### Testing
```
rm k8sResources/*yaml ; \
kc delete workflows `kc get workflow 2>&1 | tail -n +2  | grep -v "No resources"  | cut -f 1 -d \  ` ; \
kc delete workflowtemplates `kc get workflowtemplates 2>&1 | tail -n +2  | grep -v "No resources"  | cut -f 1 -d \  ` ; \
npm run make-templates -- --outputDirectory ${PWD}/k8sResources && \
kc create -f k8sResources && 
./scripts/createSampleMigration.sh
```

Screenshot 2025-10-15 at 3 22 34 PM

### Check List


orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonWorkflowTemplates.ts
```diff
@@ -86,30 +85,30 @@ export function setupTestCredsForContainer(
...
+                    name: makeStringTypeProxy(expr.literal("localstack-test-creds")),
+                    optional: makeDirectTypeProxy(expr.not(useLocalStack))
                 }
             }
         ],
         container: {
             ...restOfContainer,
             env: [
-                ...safeSpread(env),
+                ...(env === undefined ? [] : env),
```
AndreKurait: Seems more readable to me to use `...(env ?? [])` plus this handles env = null

.gitignore
```diff
@@ -37,3 +37,6 @@ orchestrationSpecs/node_modules/
 orchestrationSpecs/packages/**/dist/
 orchestrationSpecs/packages/**/node_modules/
 orchestrationSpecs/packages/migration-workflow-templates/k8sResources
+orchestrationSpecs/k8s-schemas
```
AndreKurait: We'll remove this right? I'm fine making that in a separate PR

AndreKurait (approved):

PR #1913, opened 2025-10-22
Fix knn settings during json conflict

### LINKED ISSUE #1821: [BUG] OS 1.3.7 -> OS 2.19 transformation error TextNode cannot be cast to ObjectNode
### What is the bug?
When we do console metadata evaluate we get the following error
ERROR - test_index failed to transform to the target version: Transformation for index index test_index failed due to class `com.fasterxml.jackson.databind.node.TextNode` cannot be cast to class `com.fasterxml.jackson.databind.node.ObjectNode `
(`com.fasterxml.jackson.databind.node.TextNode` and `com.fasterxml.jackson.databind.node.ObjectNode` are in unnamed module of loader 'app') 

Source: OS 1.3.7
Target: OS 2.19
MigrationAssistant: 2.5.0

Create an index with the following setting:
```json
{
    "settings": {
      "index": {
        "number_of_shards": "1",
        "provided_name": "test-migrations-reindex-fixed",
        "knn.space_type": "cosinesimil",
        "knn": "true",
        "creation_date": "1757078780188",
        "number_of_replicas": "1",
        "uuid": "_5TwzpZUQKa37bAjmf6EAQ",
        "version": {
          "created": "135248527",
          "upgraded": "136407827"
        }
      }
    }
}
```

Migration assistant should transform this correctly and send this for indexing.

I suspect it is happening because of these two settings:
```
"knn.space_type": "cosinesimil",
"knn": "true"
```
Migration assistant tries to construct a map like this:
knn: {
   space_type: cosinesimil
}
When it reads "knn": "true", the textnode collides with the object node.

AndreKurait:
### Description
Address bug when index settings conflict in metadata when `knn` and `knn.nested_setting` both are present. (and other similar conflicts)

### Issues Resolved
#1821 

### Testing
Created failing test prior to fix, verified test passed with fix along with more targeted unit tests.

### Check List

### DIFF
--- a/MetadataMigration/src/test/java/org/opensearch/migrations/KnnSettingsTransformationTest.java
+++ b/MetadataMigration/src/test/java/org/opensearch/migrations/KnnSettingsTransformationTest.java
@@ -52,11 +52,14 @@ class KnnSettingsTransformationTest extends BaseMigrationTest {
         }
     }
 
+    record IndexConfig(String name, String settingsJson) {}
+
     @SneakyThrows
     private void performKnnSettingsTest() {
         startClusters();
 
-        String flatDottedFormat = "{\n" +
+        var indexConfigs = List.of(
+            new IndexConfig("test-knn-1", "{\n" +
                 "  \"settings\": {\n" +
                 "    \"index\": {\n" +
                 "      \"number_of_shards\": \"1\",\n" +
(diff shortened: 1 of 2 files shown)


MetadataMigration/src/test/java/org/opensearch/migrations/KnnSettingsTransformationTest.java
```diff
@@ -0,0 +1,112 @@
...
+    @SneakyThrows
+    private void performKnnSettingsTest() {
+        startClusters();
+
+        String flatDottedFormat = "{\n" +
+                "  \"settings\": {\n" +
+                "    \"index\": {\n" +
+                "      \"number_of_shards\": \"1\",\n" +
+                "      \"knn.algo_param.ef_search\": 100,\n" +
+                "      \"knn\": \"true\",\n" +
```
gregschohn: Can you set up this test to run twice, with these two lines in different orders.  I could see a refactoring of the code that does one-pass instead of two and not doing the right thing

RFS/src/main/java/org/opensearch/migrations/bulkload/transformers/TransformFunctions.java
```diff
@@ -16,27 +16,108 @@ private TransformFunctions() {}
 
     /* Turn dotted index settings into a tree, will start like:
      * {"index.number_of_replicas":"1","index.number_of_shards":"5","index.version.created":"6082499"}
+     * 
+     * Special handling for conflicting keys: When a scalar value like "knn": "true" exists
+     * alongside nested properties like "knn.space_type": "l2", we keep the settings flat
+     * to preserve both values. This means the output will have both "knn": "true" and
+     * "knn.space_type": "l2" as separate flat keys rather than nesting space_type under knn.
```
gregschohn: Adding 2-3 more sentences describing an overview of the strategy would make it much easier for the reader to understand what you're trying to achieve.

gregschohn (approved):
Please see the comments, but either way, this is an improvement.

PR #1915, opened 2025-10-22
[Introduce] Continuous Worker mode in the Reindex From Snapshot process

jugal-chauhan:
### Description
RFS workers are currently short-lived. Each worker processes a single shard and then terminates.

While this model works well for ECS tasks, it causes unnecessary pod churn in EKS deployments, since Kubernetes marks pods that exit within 5 minutes as “failed” and places them in BackOff state.As a result, even successfully completed RFS workers were briefly treated as failures, introducing restart delays and slowing overall migration throughput.

This PR introduces continuous mode to enable long-running workers that can process multiple shards sequentially, while preserving existing ECS/test behavior. Changes introduced : 

1. New `--continuous-mode` for long running deployments
- CLI default: `false` (preserve non-continuous mode)
- Switched EKS deployment to use `true` for long living pods
- EKS pods should remain alive across shard boundaries to reduce restarts and operational overhead. Now in continuous mode, workers will acquire, process, re-acquire work indefinitely, stopping only when no migration work remains. 
- In single-run mode (ECS/tests), behavior remains unchanged where worker processes one shard and exits.

Previously, the worker’s main path did a single pass of work. That made continuous operation awkward and lease timers could leak across items with `--continuous-mode`. This PR introduced two layers:
- `runWorkItemLoop(...)` which owns the overall lifecycle loop for continuous mode and calls `processWorkItem(...)` once per work item attempt.
- `processWorkItem(...)` which owns the single-item attempt, including the per-item lease timer `LeaseExpireTrigger`, calls into existing `run(...)`, and returns a `boolean` telling the outer loop whether to continue.

2. Prevent work item double finalization during concurrent lifecycle events
- Long-lived workers increase the likelihood of race conditions between lease timeouts and shutdown hooks. This PR ensures clean transitions between work items and prevents concurrent finalization paths from interfering by
a. Adding `lastWorkItemFinalized` tracking to detect and skip a re-finalization
b. Scoping each work item to its own `LeaseExpireTrigger`

3. Simplifies CI/CD behavior by making process outcomes predictable across ECS, EKS, and test environments by defining consistent exit semantics:
- 0 = successful completion, 
- 1 = no work remaining, 
- 2 = lease timeout/processing failure
Replaces previous `NO_WORK_LEFT_EXIT_CODE = 3` for consistency.

4. New `ContinuousWorkerTest`
This new test explicitly validates contract to prevent regressions using 4 test cases:
- `No-work-left` exits with appropriate codes in both deployment modes
- `SIGTERM` during active shard processing triggers clean shutdown without double finalization
- Lease timeout resilience under network latency (ToxiProxy simulation) maintains process health in continuous mode
- Verify previous shard lease timers never fire during subsequent shard processing

### Check List

### DIFF
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
@@ -76,42 +76,40 @@ public class ContinuousWorkerTest extends SourceTestBase {
     @TempDir
     Path tempDirLucene;
 
-    private static Stream<Arguments> scenarios() {
-        return Stream.of(
-                org.junit.jupiter.params.provider.Arguments.of(true,  SearchClusterContainer.ES_V7_10_2, SearchClusterContainer.OS_V2_19_1),
-                org.junit.jupiter.params.provider.Arguments.of(false, SearchClusterContainer.ES_V7_10_2, SearchClusterContainer.OS_V2_19_1)
-        );
-    }
-
-    private static String[] args(boolean continuous, String sourceVersion, String... more) {
-        var baseArgs = continuous 
-                ? new String[] { "--continuous-mode", "--source-version", sourceVersion }
-                : new String[] { "--source-version", sourceVersion };
+    // Always run in continuous mode
+    private static String[] contArgs(String sourceVersion, String... more) {
+        var baseArgs = new String[] { "--continuous-mode", "--source-version", sourceVersion };
         var result = new String[baseArgs.length + more.length];
         System.arraycopy(baseArgs, 0, result, 0, baseArgs.length);
         System.arraycopy(more, 0, result, baseArgs.length, more.length);
         return result;
     }
 
+    private static Stream<Arguments> fixedPathOnly() {
+        return Stream.of(Arguments.of(
+                SearchClusterContainer.ES_V7_10_2, SearchClusterContainer.OS_V2_19_1
+        ));
+    }
+
     @SneakyThrows
     private static void waitUntilAcquiredOrTimeout(Process p, long ms) {
         // Note: This helper reads from stdout only. If relevant output appears on stderr,
         // ensure ProcessBuilder.redirectErrorStream(true) is called before starting the process.
-        var br = new BufferedReader(new InputStreamReader(p.getInputStream()));
-        long deadline = System.currentTimeMillis() + ms;
-        String line;
-        while (System.currentTimeMillis() < deadline && (line = br.readLine()) != null) {
-            if (line.contains("Acquired") || line.contains("Processing shard")) break;
+        try (var br = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
+            long deadline = System.currentTimeMillis() + ms;
+            String line;
+            while (System.currentTimeMillis() < deadline && (line = br.readLine()) != null) {
+                if (line.contains("Acquired") || line.contains("Processing shard")) break;
+            }
         }
     }
 
-    // Test: Worker exits with non-zero code when no migration work remains
-    @ParameterizedTest(name = "[no-work-left → exit=3] continuous={0} {1}->{2}")
-    @MethodSource("scenarios")
+    // Test: Continuous worker exits with code 3 when no migration work remains
+    @ParameterizedTest(name = "[no-work-left → exit=3] {0}->{1}")
+    @MethodSource("fixedPathOnly")
     @Timeout(value = DEFAULT_TIMEOUT_MINUTES, unit = TimeUnit.MINUTES)
     @SneakyThrows
-    public void exitsOnNoWorkLeft_inBothModes(
-            boolean continuousMode,
+    public void exitsOnNoWorkLeft_continuousOnly(
             SearchClusterContainer.ContainerVersion sourceClusterVersion,
             SearchClusterContainer.ContainerVersion targetClusterVersion
     ) {
@@ -128,27 +126,37 @@ public class ContinuousWorkerTest extends SourceTestBase {
                     CompletableFuture.runAsync(osTarget::start)
             ).join();
 
+            var targetOps = new ClusterOperations(osTarget);
+            var healthResponse = targetOps.get("/_cluster/health?wait_for_status=yellow&timeout=30s");
+            assertTrue(healthResponse.getValue().contains("yellow") || healthResponse.getValue().contains("green"),
+                    "Target cluster not ready");
+
             createSnapshot(esSource, SNAPSHOT_NAME, testSnapshotContext);
             esSource.copySnapshotData(tempDirSnapshot.toString());
-            
-            var additionalArgs = args(continuousMode, sourceClusterVersion.getVersion().toString());
+
+            var additionalArgs = contArgs(
+                    sourceClusterVersion.getVersion().toString(),
+                    "--index-allowlist", "geonames"
+            );
             var pb = setupProcess(tempDirSnapshot, tempDirLucene, osTarget.getUrl(), additionalArgs);
             var process = runAndMonitorProcess(pb);
 
-            assertTrue(process.waitFor(DEFAULT_TEST_TIMEOUT.toSeconds(), TimeUnit.SECONDS), "worker timed out");
-            int exitCode = process.exitValue();
-            assertEquals(3, exitCode, "Expected NO_WORK_LEFT (3) when no migration work remains");
+            try {
+                assertTrue(process.waitFor(DEFAULT_TEST_TIMEOUT.toSeconds(), TimeUnit.SECONDS), "worker timed out");
+                assertEquals(3, process.exitValue(), "Expected NO_WORK_LEFT (3) when no migration work remains");
+            } finally {
+                process.destroyForcibly();
+            }
         }
     }
 
 
-    // Test: Worker handles SIGTERM gracefully during active migration work
-    @ParameterizedTest(name = "[sigterm] continuous={0} {1}->{2}")
-    @MethodSource("scenarios")
+    // Test: Continuous worker handles SIGTERM gracefully during active migration work
+    @ParameterizedTest(name = "[sigterm] {0}->{1}")
+    @MethodSource("fixedPathOnly")
     @Timeout(value = DEFAULT_TIMEOUT_MINUTES, unit = TimeUnit.MINUTES)
     @SneakyThrows
-    public void sigtermLeadsToCleanShutdownDuringActiveWork(
-            boolean continuousMode,
+    public void sigtermLeadsToCleanShutdownDuringActiveWork_continuousOnly(
             SearchClusterContainer.ContainerVersion sourceClusterVersion,
             SearchClusterContainer.ContainerVersion targetClusterVersion
     ) {
@@ -168,7 +176,6 @@ public class ContinuousWorkerTest extends SourceTestBase {
                     ConnectionContextTestParams.builder().host(esSource.getUrl()).build().toConnectionContext()
             ).determineVersionAndCreate();
 
-            // Generate test data for migration
             var generator = new WorkloadGenerator(client);
             var opts = new WorkloadOptions();
             opts.setTotalDocs(100);
@@ -176,11 +183,19 @@ public class ContinuousWorkerTest extends SourceTestBase {
             opts.getIndex().indexSettings.put(IndexOptions.PROP_NUMBER_OF_SHARDS, 1);
             generator.generate(opts);
 
+            var targetOps = new ClusterOperations(osTarget);
+            var healthResponse = targetOps.get("/_cluster/health?wait_for_status=yellow&timeout=30s");
+            assertTrue(healthResponse.getValue().contains("yellow") || healthResponse.getValue().contains("green"),
+                    "Target cluster not ready");
+
             createSnapshot(esSource, SNAPSHOT_NAME, testSnapshotContext);
             esSource.copySnapshotData(tempDirSnapshot.toString());
 
-            var additionalArgs = args(continuousMode, sourceClusterVersion.getVersion().toString(),
-                    "--initial-lease-duration", "PT30s");
+            var additionalArgs = contArgs(
+                    sourceClusterVersion.getVersion().toString(),
+                    "--index-allowlist", "geonames",
+                    "--initial-lease-duration", "PT30s"
+            );
             var pb = setupProcess(tempDirSnapshot, tempDirLucene, osTarget.getUrl(), additionalArgs);
 
             pb.redirectErrorStream(true);
@@ -188,21 +203,24 @@ public class ContinuousWorkerTest extends SourceTestBase {
 
             var process = pb.start();
 
-            waitUntilAcquiredOrTimeout(process, 8000);
-            // Send SIGTERM to test graceful shutdown
-            process.destroy();
+            try {
+                waitUntilAcquiredOrTimeout(process, 8000);
+                // Send SIGTERM to test graceful shutdown
+                process.destroy();
 
-            assertTrue(process.waitFor(DEFAULT_TEST_TIMEOUT.toSeconds(), TimeUnit.SECONDS),
-                    "worker did not terminate after SIGTERM");
-            // Optional sanity check: make sure we didn't take the lease-timeout path
-            assertNotEquals(2, process.exitValue(), "Process should not exit via lease-timeout on SIGTERM");
+                assertTrue(process.waitFor(DEFAULT_TEST_TIMEOUT.toSeconds(), TimeUnit.SECONDS),
+                        "worker did not terminate after SIGTERM");
+                assertNotEquals(2, process.exitValue(), "Process should not exit via lease-timeout on SIGTERM");
+            } finally {
+                process.destroyForcibly();
+            }
 
         }
     }
 
 
     // Test: Worker handles lease timeout gracefully with network latency in continuous mode
-    @ParameterizedTest(name = "[lease-timeout-continuous] {0}->{1}")
+    @ParameterizedTest(name = "[lease-timeout-continuous: phase1=exit=2, phase2=exit=0|3] {0}->{1}")
     @MethodSource("fixedPathOnly")
     @Timeout(value = DEFAULT_TIMEOUT_MINUTES, unit = TimeUnit.MINUTES)
     @SneakyThrows
@@ -255,7 +273,8 @@ public class ContinuousWorkerTest extends SourceTestBase {
 
             try {
                 // Phase 1: force a timeout (expect exit code 2)
-                var timeoutArgs = args(true, sourceClusterVersion.getVersion().toString(),
+                var timeoutArgs = contArgs(sourceClusterVersion.getVersion().toString(),
+                        "--index-allowlist", "geonames",
                         "--documents-per-bulk-request", "1000",
                         "--max-connections", "1",
                         "--initial-lease-duration", "PT3s");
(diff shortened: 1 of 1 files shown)


DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -427,11 +430,15 @@ public static void main(String[] args) throws Exception {
             }));
 
             MDC.put(LOGGING_MDC_WORKER_ID, workerId); // I don't see a need to clean this up since we're in main
-            DocumentReindexer reindexer = new DocumentReindexer(targetClient,
+
+            // Target reindexer is safe to reuse across iterations
```
AndreKurait: This comment can be removed

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -448,37 +455,44 @@ public static void main(String[] args) throws Exception {
...
-                unpackerFactory,
-                arguments.maxShardSizeBytes,
-                context,
-                cancellationRunnableRef,
-                workItemTimeProvider);
-            cleanShutdownCompleted.set(true);
-        } catch (NoWorkLeftException e) {
-            log.atInfo().setMessage("No work left to acquire.  Exiting with error code to signal that.").log();
+            if (arguments.continuousMode) {
+                log.atInfo().setMessage("Running in continuous mode. RFS will process multiple work items until SIGTERM").log();
```
AndreKurait: This log doesn't add much value and is inaccurate

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
```diff
@@ -217,7 +225,8 @@ private static int runProcessAgainstToxicTarget(
             "--documents-per-bulk-request", "10",
             "--max-connections", "2",
             "--initial-lease-duration", "PT20s",
-            "--source-version", sourceClusterVersion.getVersion().toString()
+            "--source-version", sourceClusterVersion.getVersion().toString(),
+            "--continuous-mode", "false"
```
AndreKurait: Why does this test need to change at all if --continuous-mode is false?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * Validates the continuous-worker behavior for RFS:
+ *  - runs with --continuous-mode (loop)
+ *  - processes all available work items
+ *  - remains alive (idles) when no work is left
+ *  - exits cleanly when interrupted (simulated SIGTERM)
+ */
+@Slf4j
+@Disabled("Test disabled from Gradle builds")
```
AndreKurait: Is this going to be re-enabled before merging in?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
```diff
@@ -0,0 +1,219 @@
...
+        try (
+            final var sourceCluster = new SearchClusterContainer(SearchClusterContainer.ES_V5_6_16);
+            final var targetCluster = new SearchClusterContainer(SearchClusterContainer.OS_V2_19_1)
+        ) {
+            runContinuousHappyPath(sourceCluster, targetCluster, sampleDocs(), 2);
+        }
+    }
+
+    @Test
+    void testContinuous_ES710_to_OS13() throws Exception {
```
AndreKurait: Can you imagine any bug getting introduced that would occur by-version with this?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
```diff
@@ -0,0 +1,219 @@
...
+        final var snapshotContext = SnapshotTestContext.factory().noOtelTracking();
+        final var testDocMigrationContext = DocumentMigrationTestContext.factory().noOtelTracking();
+
+        try {
+            // 1) Bring clusters up and define matching index
+            Startables.deepStart(sourceCluster, targetCluster).join();
+            var sourceOps = new ClusterOperations(sourceCluster);
+            var targetOps = new ClusterOperations(targetCluster);
+
+            String indexBody = "{\"settings\":{\"number_of_shards\":1,\"number_of_replicas\":0}}";
```
AndreKurait: Could we have this with 2 shards to make sure it continues through both?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
```diff
@@ -0,0 +1,219 @@
...
+    ) {
+        List<String> args = new ArrayList<>(List.of(
+            "--target-host", targetCluster.getUrl(),
+            "--snapshot-name", snapshotName,
+            "--snapshot-local-dir", localDirectory.toString(),
+            "--lucene-dir", localDirectory.toString() + "/lucene",
+            "--source-version", sourceCluster.getContainerVersion().getVersion().toString(),
+            "--session-name", sessionName,
+            "--index-allowlist", index,
+            "--no-work-retry-delay-seconds", String.valueOf(noWorkDelaySeconds)
```
AndreKurait: What is this argument?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -756,4 +771,85 @@ private static void confirmShardPrepIsComplete(
...
+                throw (RuntimeException) e;
+            }
+            throw new IOException("Error processing work item", e);
+        } finally {
+            progressCursor.set(null);
+            workItemRef.set(null);
+            // Ensure nothing stale can be invoked later
+            cancellationRunnableRef.set(null);
+        }
+        return arguments.continuousMode && !Thread.currentThread().isInterrupted() && !cleanShutdownCompleted.get();
```
AndreKurait: I think `!Thread.currentThread().isInterrupted()` is redundant here since you return false above

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
AndreKurait: I'd like to see logic from LeaseExpirationTest with tested with continuous mode on and off, either here or in the LeaseExpirationTest

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -756,4 +771,85 @@ private static void confirmShardPrepIsComplete(
...
+                sourceResourceProvider.getShardMetadata(),
+                unpackerFactory,
+                arguments.maxShardSizeBytes,
+                context,
+                cancellationRunnableRef,
+                workItemTimeProvider);
+
+            if (completionStatus == CompletionStatus.NOTHING_DONE) {
+                if (arguments.continuousMode) {
+                    shortIdle();
```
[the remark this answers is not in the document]
jugal-chauhan: shortidle() has been removed entirely

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SourceTestBase.java
```diff
@@ -5,14 +5,12 @@
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStreamReader;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.Paths;
+import java.nio.file.*;
```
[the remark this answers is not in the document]
jugal-chauhan: reverted the change

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
jugal-chauhan: reverted the change.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SourceTestBase.java
```diff
@@ -355,15 +352,38 @@ public static CompletionStatus migrateDocumentsWithOneWorker(
     }
 
     public static void deleteTree(Path path) throws IOException {
-        try (var walk = Files.walk(path)) {
-            walk.sorted(Comparator.reverseOrder()).forEach(p -> {
+        if (path == null) return;
```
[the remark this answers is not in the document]
jugal-chauhan: Reverted this change for now

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -448,43 +449,75 @@ public static void main(String[] args) throws Exception {
...
             var unpackerFactory = new SnapshotShardUnpacker.Factory(
                 repoAccessor,
                 luceneDirPath,
                 sourceResourceProvider.getBufferSizeInBytes()
             );
 
-            run(
-                new LuceneIndexReader.Factory(sourceResourceProvider),
-                reindexer,
+            // Unified path: always run at least once; continue if continuousMode=true.
```
AndreKurait: remove comment

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -448,43 +449,75 @@ public static void main(String[] args) throws Exception {
...
+                    workItemTimeProvider,
+                    cleanShutdownCompleted,
+                    context,
+                    cancellationRunnableRef,
+                    sourceResourceProvider,
+                    reindexer,
+                    unpackerFactory
+                );
+            } catch (IOException e) {
+                log.atError().setCause(e).setMessage("IO error during work item processing").log();
```
AndreKurait: I'm not sure what this is adding beyond the caller's `Unexpected error running RfsWorker` when exceptions are thrown

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -756,4 +799,80 @@ private static void confirmShardPrepIsComplete(
...
+            }
+            throw new IOException("Error processing work item", e);
+        } finally {
+            progressCursor.set(null);
+            workItemRef.set(null);
+            // Ensure nothing stale can be invoked later
+            cancellationRunnableRef.set(null);
+        }
+        // loop continuation logic for continuous mode
+        return arguments.continuousMode && !cleanShutdownCompleted.get();
```
AndreKurait: In what case would you expect this line to run and cleanShutdownCompleted == true?

jugal-chauhan: This line will get executed when the RFS worker's iteration was normal (made some progress/completed work). 

This line executes when the `run(....)` method returned a status other than
- all above described exceptions like NoWorkLeftException or IO or InterruptedException
- CompletionStatus.NOTHING_DONE

It returns `true` to continue with the outer loop
- when continuous mode == true
- and cleanShutdownCompleted == false

It returns `false` to stop the loop
- when continuous mode == false
- or cleanShutdownCompleted == true

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -502,11 +562,22 @@ private static void executeCleanShutdownProcess(
             return;
         }
         var workItemAndDuration = workItemRef.get();
-        log.atInfo().setMessage("Marking progress: " + workItemAndDuration.getWorkItem().toString() + ", at doc " + progressCursor.get().getProgressCheckpointNum()).log();
```
gregschohn: did you get a warning for this line?  I had thought that with ErrorProne these were supposed to be build errors.  I'm wondering how it persisted

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -449,48 +451,106 @@ public static void main(String[] args) throws Exception {
...
-        } catch (NoWorkLeftException e) {
-            log.atInfo().setMessage("No work left to acquire.  Exiting with error code to signal that.").log();
+                sourceResourceProvider,
+                reindexer,
+                unpackerFactory
+            );
             cleanShutdownCompleted.set(true);
-            System.exit(NO_WORK_LEFT_EXIT_CODE);
+
+            if (arguments.continuousMode) {
```
AndreKurait: Can you help me understand this branch? Both sides of this if run when `shouldContinue=false` and i have a hard time seeing how you would know this was succesful

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -449,48 +451,106 @@ public static void main(String[] args) throws Exception {
...
                 luceneDirPath,
                 sourceResourceProvider.getBufferSizeInBytes()
             );
 
-            run(
-                new LuceneIndexReader.Factory(sourceResourceProvider),
-                reindexer,
-                progressCursor,
-                workCoordinator,
+            runWorkItemLoop(
```
AndreKurait: Probably more straight forward to embed the do while here.

```
do {
} while(processWorkItem(
                initialLeaseDuration,
                continuousMode,
                snapshotName,
                previousSnapshotName,
                experimentalDeltaMode,
                indexAllowlist,
                maxShardSizeBytes,
                workItemRef,
                progressCursor,
                workCoordinator,
                workItemTimeProvider,
                cleanShutdownCompleted,
                lastWorkItemFinalized,
                context,
                cancellationRunnableRef,
                sourceResourceProvider,
                reindexer,
                unpackerFactory
 ));
```

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -569,10 +643,19 @@ private static void exitOnLeaseTimeout(
...
+
+        // On success
+        if (!continuousMode) {
+            cleanShutdownCompleted.set(true);
+            log.info("Exiting due to lease-timeout in non-continuous mode with code {}", PROCESS_TIMED_OUT_EXIT_CODE);
             System.exit(PROCESS_TIMED_OUT_EXIT_CODE);
         }
-        cleanShutdownCompleted.set(true);
-        System.exit(PROCESS_TIMED_OUT_EXIT_CODE);
+        // continuous mode: keep shutdown hook active for next item checkpointing
```
AndreKurait: please revert

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/ContinuousWorkerTest.java
```diff
@@ -0,0 +1,430 @@
...
+    @MethodSource("scenarios")
+    @Timeout(value = 6, unit = TimeUnit.MINUTES)
+    @SneakyThrows
+    public void exitsOnNoWorkLeft_inBothModes(
+            boolean continuousMode,
+            SearchClusterContainer.ContainerVersion sourceClusterVersion,
+            SearchClusterContainer.ContainerVersion targetClusterVersion
+    ) {
+        final var testSnapshotContext = SnapshotTestContext.factory().noOtelTracking();
+        var tempDirSnapshot = Files.createTempDirectory("rfs_no_work_left_snap");
```
AndreKurait: We should be able to use annotated temp directory in the class

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -387,33 +394,25 @@ public static void main(String[] args) throws Exception {
         var workItemTimeProvider = new WorkItemTimeProvider();
         var coordinatorFactory = new WorkCoordinatorFactory(targetVersion, arguments.indexNameSuffix);
         var cleanShutdownCompleted = new AtomicBoolean(false);
+        var lastWorkItemFinalized = new AtomicReference<String>();
```
gregschohn: Can you please add a comment describing how the value here transitions through the different execution paths and how this value is read to influence the execution?  That will help me understand this much better.

I'm also trying to figure out if you're trying to use this for concurrency control or just as a reference (the split get and set makes me thing just as a reference).

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -757,4 +839,93 @@ private static void confirmShardPrepIsComplete(
...
+                context,
+                cancellationRunnableRef,
+                workItemTimeProvider);
+
+            if (completionStatus == CompletionStatus.NOTHING_DONE) {
+                if (continuousMode) {
+                    return true;
+                } else {
+                    log.atInfo().setMessage("No progress made (non-continuous mode). Exiting with error code (no-work-left) to signal that.").log();
+                    System.exit(NO_WORK_LEFT_EXIT_CODE);
```
gregschohn: Generally try to exit ONLY at the top level and in as few places as possible.  Calling exit means that you won't be able to test that function.  This is another reason to flatten the logic more.

Notice that we DO call exit within the lease expiration handler - that's purposeful, since we can't reliably get back onto the main thread and into the main() method to do it there.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -72,8 +73,9 @@
 
 @Slf4j
 public class RfsMigrateDocuments {
+    public static final int SUCCESS_EXIT_CODE = 0;
+    public static final int NO_WORK_LEFT_EXIT_CODE = 1;
     public static final int PROCESS_TIMED_OUT_EXIT_CODE = 2;
-    public static final int NO_WORK_LEFT_EXIT_CODE = 3;
```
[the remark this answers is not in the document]
jugal-chauhan: I have reverted this change to make this exit code remain 3

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -502,11 +562,22 @@ private static void executeCleanShutdownProcess(
             return;
         }
         var workItemAndDuration = workItemRef.get();
-        log.atInfo().setMessage("Marking progress: " + workItemAndDuration.getWorkItem().toString() + ", at doc " + progressCursor.get().getProgressCheckpointNum()).log();
+        var currentWorkItemId = workItemAndDuration.getWorkItem().toString();
+
+        // Check if this work item was already finalized by exitOnLeaseTimeout
+        if (currentWorkItemId.equals(lastWorkItemFinalized.get())) {
+            log.atInfo().setMessage("Work item {} already finalized by lease timeout handler").addArgument(currentWorkItemId).log();
```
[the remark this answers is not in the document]
jugal-chauhan: I added coverage via the lease timeout and SIGTERM tests in ContinuousWorkerTest, which simulates a clean shutdown and timeout finalization. The guard ensures a work item isn’t marked complete twice by both the lease handler and shutdown hook.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
jugal-chauhan: Fixed. All System.exit() calls have been moved to top-level control flow in main()

Lower level methods (processWorkItem, run, etc.) now return result enums instead of calling System.exit().

The only remaining direct exit is in exitOnLeaseTimeout(), where it’s required for reliability.

jugal-chauhan:
Update: I’ve removed `lastWorkItemFinalized` entirely.

The lease timeout path now sets a single, authoritative flag `cleanShutdownCompleted.set(finalized)` before `System.exit(PROCESS_TIMED_OUT_EXIT_CODE)`.

The shutdown hook early returns if:
	1.	`cleanShutdownCompleted.get()` is true, or
	2.	`workItemRef.get() == null || progressCursor.get() == null`

With those two guards, the shutdown hook won’t attempt to mark completion after a timeout path has already finalized (or when there’s nothing to finalize). This makes the extra “last finalized ID” state redundant, so I removed it.

jugal-chauhan:
After further discussions with the team, we've decided to proceed with the changes proposed in PR #1946 and PR #1939 instead. These two PRs will bring in the functionality originally intended by this PR in a more comprehensive manner.

Thank you for the review and feedback!

PR #1927, opened 2025-10-29
Add transformation for adding `max_ngram_diff` when needed

peternied:
### Description
Customers have reported issues with templates containing ngram analzyers, they needed a transformation to include max_ngram_diff when the difference was larger than 1.  Noticed this scenario was unique and slightly untested - template transformation haven't been verified before.   Added a new javascript transformer and used it in a new test case that verifies both index and templates are updated as expected.

### Testing
Added a new test case to verify the scenario.

### Check List
  - https://github.com/opensearch-project/documentation-website/issues/11451

### DIFF
--- a/transformation/standardJavascriptTransforms/test/ngram-diff-setting.test.js
+++ b/transformation/standardJavascriptTransforms/test/ngram-diff-setting.test.js
@@ -224,4 +224,54 @@ describe('NGramDiffSetting Transformer', () => {
       expect(result).toBe(doc);
     });
   });
+
+  describe('Plain Object Support', () => {
+    test('adds max_ngram_diff to plain object body', () => {
+      const transformer = main(null);
+      const doc = {
+        type: 'template',
+        name: 'test-template',
+        body: {
+          settings: {
+            analysis: {
+              tokenizer: {
+                my_tokenizer: {
+                  type: 'ngram',
+                  min_gram: 2,
+                  max_gram: 5
+                }
+              }
+            }
+          }
+        }
+      };
+      
+      const result = transformer(doc);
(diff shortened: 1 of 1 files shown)


transformation/standardJavascriptTransforms/src/ngram-diff-setting.js
```diff
@@ -0,0 +1,148 @@
...
+    return (type === "template" || type === "indexTemplate" || type === "componentTemplate");
+  };
+  
+  const isIndexCreationRequest = (doc) => {
+    if (!isObj(doc)) return false;
+    if (!(hasKey(doc, "type") && hasKey(doc, "name") && hasKey(doc, "body"))) {
+      return false;
+    }
+    const type = getKey(doc, "type");
+    return (type === "index");
```
AndreKurait: Does this work? I thought we found the type was different than index and was instead the classname?

transformation/standardJavascriptTransforms/src/ngram-diff-setting.js
AndreKurait: Can we get a documentation/breaking changes update on this?

transformation/standardJavascriptTransforms/src/ngram-diff-setting.js
AndreKurait: With the test that's added in this PR, the updated target template `ngram_template` will get applied to index `ngram_test_index` even if this transformation does not apply on the index itself. Can we add a new index on the test with ngram that wouldn't get the setting from the created template?

peternied: In the test a template is created and then an index that uses the template, both of these are using ngram.  The test makes sure that both the template an the index are updated separately during the migration.  We wouldn't get any better coverage by adding another ngram tokenizing index, do I have that right?

PR #1938, opened 2025-11-03
Bump netty from 4.1.124 -> 4.1.128

peternied:
### Description
Bump netty from 4.1.124 -> 4.1.128

As part of the netty fixes addressing https://github.com/netty/netty/security/advisories/GHSA-jq43-27x9-3v86 this tightened the strictness of checks during request parsing which the test started failing.  By switching from LF -> CRLF this fixed the out of compliance configuration.

Note; this could have implications for requests that were previously in compliance and are now being rejected, we might want to consider adding a flag to enable the lenient parsing mode.

### Check List
- [ ] ~New functionality includes testing~
- [ ] ~Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.~

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1942, opened 2025-11-05
Netty 4.1.128 with support for lenient default line ending

AndreKurait:
**Description**  
Follow-up to Proposed PR #1938, which upgraded Netty from 4.1.124 → 4.1.128 to address GHSA-jq43-27x9-3v86 and GHSA-fghv-69vj-qj49.

The previous bump introduced stricter request-parsing checks in Netty 4.1.125, causing possible failures where clients use `LF` instead of `CRLF`. That change aligned behavior with the current HTTP spec but could break compatibility with some older clients.

This PR adds configuration support for **lenient line-ending parsing by default**, disabling strict line checks to preserve backward compatibility with legacy clients. This ensures we can continue handling replayed or proxied traffic that doesn’t fully conform to modern HTTP formatting.

Netty Reviewed CVE: GHSA-fghv-69vj-qj49 (Marked as Low)
Original Netty Issue: https://github.com/netty/netty/issues/15522
Original Netty PR: https://github.com/netty/netty/pull/15611

**Key Changes**
- Retains Netty 4.1.128 for security compliance (GHSA-jq43-27x9-3v86 and GHSA-fghv-69vj-qj49).  
- Leverage's netty's built in system property flag `io.netty.handler.codec.http.defaultStrictLineParsing` to disable strict line parsing by default.  
- Defaults to lenient mode for proxy/replayer use cases.  

Note: Includes an update to the JIB building to include the JVM flags we've been applying to our existing docker builds.

### Testing
Existing tests broke with the change, adding system property during test runs reverts to existing behavior shown by the tests passing

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/build.gradle
+++ b/TrafficCapture/trafficCaptureProxyServer/build.gradle
@@ -172,6 +172,11 @@ tasks.register("verifyCompatibilityWithJDK11", VerifyCompatibilityWithJDK11Task)
     dependsOn tasks.named("assemble")
 }
 
+tasks.withType(Test).configureEach {
+    // Allow Lenient line parsing for compatibility with older clients and tests
+    systemProperty 'io.netty.handler.codec.http.defaultStrictLineParsing', false
+}
+
 // Add class version verification to slowTest
 tasks.named("slowTest") {
     finalizedBy("verifyCompatibilityWithJDK11")
--- a/TrafficCapture/trafficReplayer/build.gradle
+++ b/TrafficCapture/trafficReplayer/build.gradle
@@ -82,3 +82,8 @@ jar {
         attributes 'Main-Class': application.mainClass
     }
 }
+
+tasks.withType(Test).configureEach {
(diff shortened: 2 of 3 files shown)


peternied:
@AndreKurait thanks for spending the time to look into this.  Is there a way to do this without using jvm parameters?  While its nice that we can keep the more lenient default, it creates yet another configuration path for how netty operates that seems previously unused.

buildImages/build.gradle
```diff
@@ -19,12 +19,22 @@ if (!imageArch) {
...
+
 def jibProjects = [
         "TrafficCapture:trafficReplayer": [
                 baseImageName : "amazoncorretto",
                 baseImageTag  : "17-al2023-headless",
                 imageName     : "traffic_replayer",
                 imageTag      : "latest",
+                jvmFlags      : [
+                        // By default enable transparent proxy with older clients for maximum compatibility
+                        "-Dio.netty.handler.codec.http.defaultStrictLineParsing=false",
```
gregschohn: Same question as before - how does jib wire up the java properties - are they env variables?  How would a user override this?

PR #1962, opened 2025-11-12
Add data stream backing index transformation for RFS migration

AndreKurait:
## Description

This PR adds support for data stream migration in RFS with three key features:

### 1. Data Stream Backing Index Transformation
Adds dataStreamBackingIndexTransform.js to automatically transform `.ds-*` backing indexes during migration:

Extracts data stream name using regex matching for both formats:
`.ds-<data-stream>-<yyyy.MM.dd>-<generation>` (e.g., `.ds-logs-2024.01.15-000034`)
`.ds-<data-stream>-<generation>` (e.g., `.ds-logs-000034`)
Replaces backing index name with extracted data stream name and sets op_type to create
### 2. Document Exception Allowlist for Idempotent Migrations
Adds DocumentExceptionAllowlist to treat specific document-level exceptions as successful operations:

Adds --allowed-doc-exception-types CLI parameter (e.g., version_conflict_engine_exception)
Enables safe re-runs when documents already exist on target cluster
Extends BulkResponseParser, DocumentReindexer, and OpenSearchClient with allowlist support
### 3. Regex Pattern Matching for Index Allowlists
Adds AllowlistEntry class to support flexible index filtering:

Literal matching: Entries like my-index match exactly
Regex matching: Entries prefixed with regex: enable pattern-based matching (e.g., regex:logs-.*-2024)
Supports complex patterns with character classes, quantifiers, and alternation
Validates regex patterns at construction time with clear error messages
Updates FilterScheme.filterByAllowList() to use the new matching logic
This allows users to specify index allowlists more flexibly:

#### Exact matches
```
my-index
production-logs
```

#### Regex patterns
```
regex:logs-.*-2024
regex:test-\d+
regex:(logs|metrics)-.*
````

### Important Limitations

⚠️ **Migrated data streams will have wider timestamp ranges per backing index**, which may reduce search efficiency compared to the original time-based partitioning. This occurs because documents from multiple source backing indexes are consolidated into target backing indexes.

## Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2639

## Testing
- Added `DataStreamMigrationTest` with comprehensive validation:
  - Data stream creation and document indexing
  - Backing index transformation and migration
  - Exception allowlist functionality for idempotent operations
  - Document skip behavior for pre-existing documents

## Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/AllowlistEntry.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/AllowlistEntry.java
@@ -19,7 +19,6 @@ public class AllowlistEntry {
     private static final String REGEX_PREFIX = "regex:";
     
     private final String originalValue;
-    private final boolean isRegex;
     private final Pattern pattern;
     
     /**
@@ -39,7 +38,6 @@ public class AllowlistEntry {
         this.originalValue = value;
         
         if (value.startsWith(REGEX_PREFIX)) {
-            this.isRegex = true;
             String patternString = value.substring(REGEX_PREFIX.length());
             
             if (patternString.isEmpty()) {
@@ -55,7 +53,6 @@ public class AllowlistEntry {
                 );
             }
         } else {
-            this.isRegex = false;
             this.pattern = null;
         }
     }
@@ -74,7 +71,7 @@ public class AllowlistEntry {
             return false;
         }
         
-        if (isRegex) {
+        if (pattern != null) {
             return pattern.matcher(testValue).matches();
         } else {
             return originalValue.equals(testValue);
@@ -88,14 +85,5 @@ public class AllowlistEntry {
      */
     public String getOriginalValue() {
         return originalValue;
-    }
-    
-    /**
-     * Returns whether this entry is a regex pattern.
-     * 
-     * @return true if this entry uses regex matching, false for literal matching
-     */
-    public boolean isRegex() {
-        return isRegex;
     }
 }
--- a/RFS/src/test/java/org/opensearch/migrations/bulkload/common/AllowlistEntryTest.java
+++ b/RFS/src/test/java/org/opensearch/migrations/bulkload/common/AllowlistEntryTest.java
@@ -12,7 +12,6 @@ public class AllowlistEntryTest {
     void testLiteralEntry_exactMatch() {
         var entry = new AllowlistEntry("my-index");
         
-        assertThat(entry.isRegex(), equalTo(false));
         assertThat(entry.getOriginalValue(), equalTo("my-index"));
         assertThat(entry.matches("my-index"), equalTo(true));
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java [hunk 1]
```diff
@@ -446,7 +452,11 @@ public Mono<BulkResponse> sendBulkRequest(String indexName, List<? extends BulkO
                     log.atDebug().setMessage("Response has some errors...: {}").addArgument(response.body).log();
                     log.atDebug().setMessage("... for request: {}").addArgument(body).log();
                     // Remove all successful documents for the next bulk request attempt
-                    var successfulDocs = resp.getSuccessfulDocs();
+                    var successfulDocs = resp.getSuccessfulDocs(allowlist);
+                    // If all documents have been successfully processed (including allowlisted errors), treat as success
+                    if (docsMap.isEmpty()) {
```
gregschohn: how is docsMap getting updated?

transformation/standardJavascriptTransforms/src/dataStreamBackingIndexTransform.js
```diff
@@ -0,0 +1,52 @@
+/**
+ * Transforms data stream backing index operations for migration by replacing the
+ * backing index name with the data stream name and setting op_type to "create".
```
gregschohn: Can you expand upon this a little bit more.  Assume that somebody finds this, doesn't know what datastreams are at all, and decides that this sounds like it's exactly what they need - even though it isn't!

RFS/src/main/java/org/opensearch/migrations/bulkload/common/OpenSearchClient.java
AndreKurait: Updated

RFS/src/main/java/org/opensearch/migrations/bulkload/common/AllowlistEntry.java
```diff
@@ -0,0 +1,101 @@
...
+                this.pattern = Pattern.compile(patternString);
+            } catch (PatternSyntaxException e) {
+                throw new IllegalArgumentException(
+                    "Invalid regex pattern in allowlist entry '" + value + "': " + e.getMessage(),
+                    e
+                );
+            }
+        } else {
+            this.isRegex = false;
+            this.pattern = null;
```
gregschohn: nit: Why not use the null-ness of pattern to determine if it was a regex or not?

gregschohn (approved):

RFS/src/main/java/org/opensearch/migrations/bulkload/common/AllowlistEntry.java
AndreKurait: That would be simpler, i'll adjust before merging

PR #2007, opened 2025-11-27
Refactor the migrationConsole to be a top-level directory

mikaylathompson:
### Description
The major refactor I've been dreaming of; hopefully executed in a minimally complicated way.

The migration console is moved from `TrafficCapture/dockerSolution/src/main/docker/migrationConsole` to just `migrationConsole`.

It has a dockerfile, which is then used in a multi-stage docker build to fully assemble the "final migration console" with the compiled java stuff, etc.

### Issues Resolved
The long-ago https://opensearch.atlassian.net/browse/MIGRATIONS-1849

### Testing
All existing tests, plus a lot of manual testing.

### Check List

### DIFF
--- a/deployment/cdk/opensearch-service-migration/package-lock.json
+++ b/deployment/cdk/opensearch-service-migration/package-lock.json
@@ -15,7 +15,7 @@
         "@types/aws-lambda": "^8.10.159",
         "aws-lambda": "^1.0.7",
         "esbuild": "^0.27.0",
-        "node-forge": "^1.1.0",
+        "node-forge": "^1.3.2",
         "source-map-support": "^0.5.21",
         "yaml": "^2.8.1"
       },
@@ -943,6 +943,7 @@
       "integrity": "sha512-bXYxrXFubeYdvB0NhD/NBB3Qi6aZeV20GOWVI47t2dkecCEoneR4NPVcb7abpXDEvejgrUfFtG6vG/zxAKmg+g==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@ampproject/remapping": "^2.2.0",
         "@babel/code-frame": "^7.27.1",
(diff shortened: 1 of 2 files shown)


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/migrationConsole.yaml
```diff
@@ -86,11 +86,11 @@ spec:
     {{- if $developerModeEnabled  }}
         - name: local-console-libs
           hostPath:
-            path: /opensearch-migrations/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib
+            path: /opensearch-migrations/migrationConsole/lib
             type: Directory
         - name: local-test-config-processor
           hostPath:
-            path: /opensearch-migrations/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/lib/console_link/tests/workflow-tests/resources/scripts
+            path: /opensearch-migrations/migrationConsole/lib/console_link/tests/workflow-tests/resources/scripts
```
gregschohn: This should be here for this PR, but pulling this scripts directory any futher at all is an anti-pattern - it holds some fake mock scripts that look pretty similar to the real scripts.

AndreKurait:
> There are two sonarqube complaints:
> 
> 1. line 26 of the new dockerfile uses a curl command. This is the same as the [existing dockerfile](https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/dockerSolution/src/main/docker/migrationConsole/Dockerfile#L24), which I assume handles it by using a sonarqube exception. I think it makes sense to transfer that exception to this file.
> 2. line 103 of the new dockerfile is a RUN command that sonarqube would like to be combined with an subsequent RUN command. While generally in agreement, these are both already very long concatenated RUN commands, and they handle very different things. I also think this should also be given an exception.

All sonarqube exceptions are here https://github.com/opensearch-project/opensearch-migrations/blob/main/sonar-project.properties do you have time to take a look and add that change to this PR?

This would also be addressed by renaming `migrationConsole/Dockerfile.base` to `migrationConsole/Dockerfile`

sonar-project.properties
AndreKurait: @mikaylathompson, new rules need to be added on this list, which can be avoided if you do the consolidation as the comment says below

gregschohn (approved):

PR #2023, opened 2025-12-03
Add JSON logger for RFS in Kubernetes

AndreKurait:
### Description
Configures RFS (Reindex From Snapshot) to output logs in JSON format when running in Kubernetes. This enables better log parsing and aggregation in container environments.

Example log output:
```json
{"@timestamp":"2025-12-03T20:19:35.339Z","log.level":"INFO","message":"Starting RfsMigrateDocuments","log.logger":"org.opensearch.migrations.RfsMigrateDocuments"}
```

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2732

### Testing
- Verified JSON log output in Kubernetes pod logs via `kubectl logs`

- Verified without env variable logs show as expected:
```
2025-12-03 20:48:23,615 INFO o.o.m.RfsMigrateDocuments [main] After 0 another process holds the lock for setting up the shard work items.  Waiting 1000 ms before trying again.
```

### Check List


RFS/src/main/resources/log4j2.properties
```diff
@@ -21,12 +21,26 @@ logger.FailedRequestsLogger.level = ${env:FAILED_REQUESTS_LOGGER_LEVEL:-info}
...
+
+appender.routing.type = Routing
+appender.routing.name = RoutingConsole
+appender.routing.routes.type = Routes
+appender.routing.routes.pattern = ${env:CONSOLE_LOG_FORMAT:-pattern}
+appender.routing.routes.route1.type = Route
+appender.routing.routes.route1.key = json
+appender.routing.routes.route1.ref = ConsoleJson
+appender.routing.routes.route2.type = Route
+appender.routing.routes.route2.ref = Console
```
gregschohn: Should these key names be ConsoleText and ConsoleJson.  The descriptors don't seem to be in agreement right now.
I'm also curious what happens if a user specifies a different value.  Will a default one be used - will none be used?

AndreKurait: > Otherwise, the default Route is selected. The key of the default Route is determined by the [Routing script](https://logging.apache.org/log4j/2.x/manual/appenders/delegating.html#RoutingAppender-element-AbstractScript) or is null (lack of [key attribute](https://logging.apache.org/log4j/2.x/manual/appenders/delegating.html#Route-attr-key)) if the script is absent.

since there is no `.key` on route2, it is the default route

gregschohn (approved):

PR #2065, opened 2025-12-16
Fix S3 URI regex to allow optional path after bucket name

AndreKurait:
### Description
Fix the s3RepoPathUri regex in userSchemas.ts to allow optional path component after the bucket name. The previous regex `^s3:\/\/[a-z0-9][a-z0-9.-]{1,61}[a-z0-9]$` only matched bucket names without paths, but the description indicated it should support `s3://BUCKETNAME/PATH`.

The minikube integ test now validates this path support by using a unique S3 path per migration workflow (using workflow UID suffix) to prevent overlap between concurrent test runs.

### Issues Resolved
N/A

### Testing
- Updated snapshot tests pass
- Regex now correctly matches:
  - `s3://my-bucket` (bucket only)
  - `s3://my-bucket/` (bucket with trailing slash)
  - `s3://my-bucket/path/to/repo` (bucket with path)
- Minikube integ test uses unique S3 paths per workflow

### Check List


gregschohn (approved):
thanks for adding the test

PR #2081, opened 2025-12-16
Fix Sync task deleting runtimeJars directory

AndreKurait:
## Description

This PR fixes the release build failure where the `reindex_from_snapshot` Docker image build was failing because `/build/runtimeJars` was not found.

**Root Cause:**
The `copyVersionFile_*` task used Gradle's `Sync` task which syncs to the same directory (`docker/build`) as `copyDockerRuntimeJars`. Since `Sync` deletes files in the destination that aren't in the source, it deleted the `runtimeJars` directory when it ran.

**Fix:**
Move the VERSION file to its own subdirectory so each Sync task has its own directory and won't conflict:
- `DocumentsFromSnapshotMigration`: `docker/build/version/VERSION`
- `frontend`: `build/version/VERSION`

**Changes:**
- Renamed task from `copyVersionFile_*` to `syncVersionFile_*` for clarity
- Updated destination paths to use `version/` subdirectory
- Updated Dockerfiles to COPY from the new `version/` subdirectory
- Disabled Gradle scan in release-drafter workflow

**Directory structure after fix:**
```
docker/build/
├── runtimeJars/  (from copyDockerRuntimeJars - preserved!)
└── version/
    └── VERSION   (from syncVersionFile - isolated!)
```

## Issues Resolved

Fixes release build failure for reindex-from-snapshot image (run 20281168457).

## Testing

- Verified locally that `runtimeJars` directory is preserved after both tasks run
- Verified Docker build proceeds past the COPY stage

## Check List

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -75,6 +75,8 @@ jobs:
             HEAD
       - name: Build and Publish to Maven
         if: ${{ !github.event.act }}
+        env:
+          OS_MIGRATIONS_GRADLE_SCAN_TOS_AGREE_AND_ENABLED: ''
         run: |
           ./gradlew publishMavenJavaPublicationToMavenRepository -Dbuild.snapshot=false -Dbuild.version=0.${{ steps.get_data.outputs.version }} && tar -C build -cvf opensearch-migrations-artifacts.tar.gz repository
       - name: Configure AWS Credentials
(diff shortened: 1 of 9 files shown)


gregschohn (approved):
We already have copyDockerRuntimeJars using its own subdirectory with its sync task
    def buildDir = project.file("./docker/build/runtimeJars")

PR #2101, opened 2025-12-22
Workflow status with deep checks

gregschohn:
### Description

Rework the workflow commands for observability of a live migration.  Some of those changes required new features in the underlying workflows, which in turn required further changes within the argo-ts builder package.  These changes should make parallel workflow tracks first class citizens/completely natural and automatic.

#### Argo Workflow Changes

Working in reverse, containerBuilder.ts was simplified to use brands (or unions of simple objects) and a static record to prevent duplicative calls but without the expensive overhead of maintaining so many different types.  That make the complexity (LOC) of maintaining the class linear with the number of features rather than squared as we don't pass a separate type parameter to every function to add a feature value now.  That was required because there are now ways to add metadata (labels & annotations) to the pods that argo spins up.  For good measure, all of the possible container fields should now be supported. 

Testing parallel snapshots showed that older versions of Elasticsearch can't perform multiple snapshots simultaneously against the same repo.  To workaround that limitation, the workflow uses argo semaphores to limit the concurrency.  Semaphore support is now available for templates.  Argo requires a config map to already be constructed (and initialized) so the migrationInitializer script takes care of that.  There was some refactoring between the wrapper shell script and the TS initializer.  Now the TS code writes a set of files to a directory and the script simply calls kubectl apply on all the files in the directory.  Notice that if the source is a newer version of OpenSearch, we'll STILL use semaphores BUT we'll have a separate semaphore for EACH SNAPSHOT, not for each repo.  That keeps all of the code within the workflow identical, just the semaphore names being passed around are different.

Another change in the builder library is to change the suspend tasks to be constructed as full TemplateBuilders so that they can take on input parameters.  They do that so it's convenient to address different suspension/approval steps now. 

The migration workflows take advantage of those new features by passing label information from the workflow config, like the source name, target name, etc down the call graph, either explicitly or included with other configs that it was sending anyway.  Those values are then set as pod labels so that the kubernetes control plane can filter exactly WHICH step was being run, especially relevant when there are multiple branches within the workflow.

The migration-workflow templates now follow a number of conventions that are used to signal hints to python tooling in the console/workflow libraries/commands.  Some hints indicate that nodes should be visually grouped together (by default, intermediary nodes will be excised) or the syntax to do further live checks or to set a context.

#### Workflow command changes

Several workflow commands have undergone major changes.  Workflows are no longer automatically detected.  If a workflow name isn't specified, it defaults to migration-workflow.  Where it makes sense there's an --all-workflows flag that will show all of the workflows.  The output and approve commands no longer have output components for the user to discover what is available.  There's a new `workflow manage` command that is a full-blown UI that renders the tree as it grows, allowing the user to view logs, tail logs, approve steps, and check on progress.

The output of `workflow status` change to remove "Failed" from the output of check/test tasks.  It replaces 'Failed' with "checked" and a static status summary at the time of the check.  The status command also now has a `--live-status` flag that will call `console status...` for each of the active resources.

`workflow output` and `workflow approve` no longer have interactive components to them.  Instead, they show the output or approve for specific resources/steps (see the previous section), specified by the command line.  Both of those programs have command-line autocompletion, which isn't super-fast, but is helpful.  Output now allows users to specify pod labels as filters, which will be source, target, snapshot, from-snapshot-migration, and task.  If multiple pods match the filter we'll show logs from multiple matches.  They'll be sorted in chronological order.  `-f` is still an option to tail and without it only the historical logs will be shown.  Any extra args are passed to the underlying commands that are pulling the logs -`kubectl log` and `stern`, which is now being installed onto the base image.

`workflow approve` now takes the name (or a glob pattern of a name).  Suspend steps now have names that match the same labels from the workflow config that were also used on the pods.  Now 

Here's the overall rundown of the command line options for the affected commands

#### Status
```
% workflow status --help
Usage: workflow status [OPTIONS] [WORKFLOW_NAME]

  Show detailed status of workflows.

  Displays workflow progress, completed steps, and approval status. By
  default, only shows running workflows. Use --all to see completed workflows
  too.

  Example:
      workflow status
      workflow status my-workflow
      workflow status --all-workflows
      workflow status --all

Options:
  --all-workflows     Show status for all workflows
  --argo-server TEXT  Argo Server URL (default: ARGO_SERVER env var, or
                      ARGO_SERVER_SERVICE_HOST:ARGO_SERVER_SERVICE_PORT)
  --namespace TEXT    Kubernetes namespace for the workflow (default: ma)
  --insecure          Skip TLS certificate verification
  --token TEXT        Bearer token for authentication
  --all               Show all workflows including completed ones (default:
                      only running)
  --live-status       Run a current status check for each snapshot and
                      backfill still running
  --help              Show this message and exit.
```

#### Output

```
% workflow output --help
Usage: workflow output [OPTIONS]

  View or tail workflow logs.

Options:
  --workflow-name TEXT
  --all-workflows       Show output for all workflows
  --argo-server TEXT    Argo Server URL (default: ARGO_SERVER env var, or
                        ARGO_SERVER_SERVICE_HOST:ARGO_SERVER_SERVICE_PORT)
  --namespace TEXT
  --insecure
  --token TEXT          Bearer token for authentication
  --prefix TEXT         Label prefix for filters
  -l, --selector TEXT   Label selector (e.g. source=a,target=b)
  -f, --follow
  --timestamps
  --help                Show this message and exit.
```

#### Approve
```
% workflow approve --help
Usage: workflow approve [OPTIONS] TASK_NAMES...

  Approve suspended workflow steps matching TASK_NAMES.

  Each TASK_NAME can be an exact name or glob pattern (e.g.,
  *.metadataMigrate).

  Example:     workflow approve source.target.metadataMigrate     workflow
  approve "*.metadataMigrate"     workflow approve step1 step2 step3

Options:
  --workflow-name TEXT
  --argo-server TEXT    Argo Server URL
  --namespace TEXT
  --insecure
  --token TEXT          Bearer token for authentication
  --help                Show this message and exit.
```

#### Manage

```
% workflow manage --help 
Usage: workflow manage [OPTIONS]

Options:
  --workflow-name TEXT
  --argo-server TEXT    Argo Server URL (default: ARGO_SERVER env var, or
                        ARGO_SERVER_SERVICE_HOST:ARGO_SERVER_SERVICE_PORT)
  --namespace TEXT
  --insecure
  --token TEXT
  --help                Show this message and exit.
```

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2788

### Testing
```
% workflow status      
[*] Workflow: migration-workflow
  Phase: Running
  Started: 2026-01-23T23:42:12Z

Workflow Steps
└── ▶  2026-01-23 23:42:12: Migration (source to target) (Running) 
    ├── ▶  2026-01-23 23:42:12: Get Snapshot Then Migrate Snapshot (t4Migration) (Running) 
    │   ├── ✓  2026-01-23 23:42:38: createSnapshot (Succeeded) 
    │   ├── ✓  2026-01-23 23:42:53: Wait For Completion (checks) (Succeeded): Snapshot completed successfully 
    │   │   └── ✓  2026-01-23 23:42:46: checkSnapshotCompletion (Succeeded): Snapshot completed successfully 
    │   └── ▶  2026-01-23 23:42:53: Migrate From Snapshot (snapshot-migration-0) (Running) 
    │       ├── ✓  2026-01-23 23:43:01: evaluateMetadata (Succeeded) 
    │       └── ⟳  2026-01-23 23:43:03: approveEvaluate - WAITING FOR APPROVAL OF 'source.target.t4Migration.snapshot-migration-0.evaluateMetadata' 
    └── ▶  2026-01-23 23:42:12: Get Snapshot Then Migrate Snapshot (t2Migration) (Running) 
        ├── ✓  2026-01-23 23:42:18: createSnapshot (Succeeded) 
        ├── ✓  2026-01-23 23:42:31: Wait For Completion (checks) (Succeeded): Snapshot completed successfully 
        │   └── ✓  2026-01-23 23:42:25: checkSnapshotCompletion (Succeeded): Snapshot completed successfully 
        └── ▶  2026-01-23 23:42:32: Migrate From Snapshot (everything) (Running) 
            ├── ✓  2026-01-23 23:42:41: evaluateMetadata (Succeeded) 
            └── ⟳  2026-01-23 23:42:43: approveEvaluate - WAITING FOR APPROVAL OF 'source.target.t2Migration.everything.evaluateMetadata' 
...

To view step outputs, run: workflow output migration-workflow
```

### Check List

### New commands in action
workflowManage


migrationConsole/lib/console_link/console_link/workflow/tui/manage_injections.py
```diff
@@ -0,0 +1,205 @@
...
+import subprocess
+import threading
+
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, List, Optional
+
+from console_link.workflow.services.workflow_service import WorkflowApproveResult, logger, WorkflowService
+from console_link.workflow.tree_utils import clean_display_name
+
+
```
AndreKurait: Nice use of dependency injection here. The dataclass interfaces make this very testable without needing real K8s/Argo connections.

deployment/k8s/localTesting.sh
```diff
@@ -34,9 +34,23 @@ if [ "${BUILD_IMAGES_WITHOUT_DOCKER:-false}" = "true" ]; then
...
+      *)
+        echo "Unsupported architecture: $ARCH"
+        exit 1
+        ;;
+    esac
+
+    gradlew :buildImages:buildImagesToRegistry_$PLATFORM
 else
-    echo "Configuring Docker environment variables to point to minikube"
+    echo "Configuring Docker environment variables to point to minikube - legacy/not supported"
```
AndreKurait: Minor: This legacy branch (lines 53-55) should be removed as a fast follow since it's marked as not supported.

PR #2103, opened 2025-12-24
Enable EKS integration test with multi-arch support and workflow orchestration improvements

AndreKurait:
## Description

Enables the EKS integration test in CI, adds multi-arch Docker build support, and includes several workflow orchestration improvements for reliability.

### CI/CD Improvements

**EKS Integration Test Enablement**
- Added `eks-integ-test` job to GitHub Actions (`.github/workflows/jenkins_tests.yml`)
- Added 2-hour timeout for EKS integration test
- Updated default repo/branch from `lewijacn/opensearch-migrations` to `opensearch-project/opensearch-migrations` and `main`

**Multi-arch Docker Builds**
- Added QEMU setup for cross-architecture builds (arm64 emulation on x86_64 hosts):
  ```groovy
  sh "docker run --privileged --rm tonistiigi/binfmt --install all"
  ```

**Workflow Concurrency**
- Updated cancel-in-progress rules to allow cancellation on non-main branches while protecting main branch pushes
- Applied to: `CI.yml`, `codeql.yml`, `jenkins_tests.yml`, `sonar-qube.yml`

### Workflow Orchestration Improvements

**etcd Client Reliability**
- Added `etcdClientHelper.sh` with retry logic, timeouts, and safe wrappers for etcd operations
- Replaced inline etcd commands with robust helper functions in `decrement.sh` and `cleanup.sh`
- Removed unused `init.sh`

**Template Renaming for Clarity**
- Renamed `createReplicaset` → `startHistoricalBackfill`
- Renamed `deleteReplicaSet` → `stopHistoricalBackfill`
- Renamed `checkRfsCompletion` → `checkHistoricalBackfillCompletion`

**Snapshot Workflow Fix**
- Added `noWait: true` to CreateSnapshot to allow workflow to manage polling separately via `checkSnapshotStatus`

**Resource Optimization**
- Reduced migration console CLI CPU limit from 1000m to 500m

**Monitor Script Fix**
- Fixed workflow status check to properly retry on connection errors instead of silently continuing

### Other Changes
- Updated Pipfile.lock dependencies
- Added kiro-cli to gradle settings
- Fixed cleanup script to handle missing servicemonitor resource gracefully

## Testing

EKS integration test runs automatically on pushes to main repo branches.

### DIFF
--- a/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/createS3Bucket.yaml
+++ b/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/createS3Bucket.yaml
@@ -37,7 +37,8 @@ spec:
               valueFrom:
                 configMapKeyRef:
                   name: migrations-default-s3-config
-                  key: ENDPOINT
+                  key: ENDPOINT_HTTP
+                  optional: true
             - name: BUCKET_NAME
               valueFrom:
                 configMapKeyRef:
--- a/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/deleteS3Bucket.yaml
+++ b/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/deleteS3Bucket.yaml
@@ -37,7 +37,8 @@ spec:
               valueFrom:
                 configMapKeyRef:
                   name: migrations-default-s3-config
-                  key: ENDPOINT
+                  key: ENDPOINT_HTTP
+                  optional: true
             - name: BUCKET_NAME
               valueFrom:
                 configMapKeyRef:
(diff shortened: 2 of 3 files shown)


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/createS3Bucket.yaml
```diff
@@ -37,7 +37,8 @@ spec:
               valueFrom:
                 configMapKeyRef:
                   name: migrations-default-s3-config
-                  key: ENDPOINT
+                  key: ENDPOINT_HTTP
```
gregschohn: It would probably be cleaner to make the ENDPOINT always use HTTP only (ditching 'localstack') and then having another key that says if it's localstack vs s3 (vs minIO?).  It really doesn't matter except for one of our test scenarios (self-managed ES, since managed won't work....) that requires mapping the domain name to the IP address before passing it in, which is done by the initialization script... That initialization script could check the extra key.  The advantage with that scenario is that the weirdness is isolated to just where it's needed.

PR #2115, opened 2026-01-05
Consolidate CI triggers to reduce duplicate workflow runs on PRs

AndreKurait:
## Overview

This PR consolidates CI workflow triggers to eliminate duplicate runs when pushing to PR branches.

### CI, SonarQube, CodeQL workflows
- `push` trigger now only runs on `main` branch
- PRs still trigger via `pull_request` event
- Eliminates duplicate runs (previously both `push` and `pull_request` triggered on PR commits)

### Jenkins workflow
- `push` trigger now only runs on `main` branch (removed `branches-ignore`)
- Added `labeled` to `pull_request_target` types
- `eks-integ-test` now uses label-based triggering:
  - Always runs on push to main
  - Runs on PRs with `run-eks-tests` label
  - Adding the label triggers the job
  - Subsequent commits run while label is present

## Testing

- Verified `eks-integ-test` shows as `skipped` on PRs without label
- Label-based triggering will be fully testable after this merges to main

## Impact

This should roughly halve the number of workflow runs on PRs from branches within the repo.

### DIFF
--- a/.github/workflows/jenkins_tests.yml
+++ b/.github/workflows/jenkins_tests.yml
@@ -8,7 +8,7 @@ on:
     types: [opened, synchronize, reopened, labeled]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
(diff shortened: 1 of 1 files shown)


AndreKurait:
We won't be able to fully validate this until after merge since `pull_request_target` changes run on version of workflow on `main`

PR #2122, opened 2026-01-07
Replace etcd chart with Aenix etcd-operator and fix Kyverno policy

AndreKurait:
## Description

Fixes EKS/K8s integration test failures caused by etcd pod rescheduling crashes and Kyverno policy issues.

### Changes

1. **Replace groundhog2k etcd chart with Aenix etcd-operator**
   - Uses `oci://ghcr.io/aenix-io/charts/etcd-operator:0.4.2`
   - Adds `dependsOn: cert-manager` and `waitForInstallation: true` to fix install race condition
   - Disables VPA for clusters without VerticalPodAutoscaler CRD

2. **Fix Kyverno zero-resource-requests policy**
   - Changed from `selector.matchExpressions` (matches labels) to `preconditions` (matches spec fields)
   - Kyverno installs in `ma` namespace (deleted with namespace cleanup)
   - Override webhook config to only exclude `kube-system`, allowing Kyverno to mutate `ma` namespace pods

3. **Increase etcd client retry timeout to 5 minutes**
   - Changed from ~25s to ~5 minutes for resilience during cluster instability

4. **Improve cleanup**
   - Added 60s timeout and force finalizer removal for stuck etcd CRs

### Testing

Build #10694 passed all tests with etcd-operator fix.

### DIFF
--- a/migrationConsole/lib/console_link/console_link/models/argo_service.py
+++ b/migrationConsole/lib/console_link/console_link/models/argo_service.py
@@ -92,13 +92,21 @@ class ArgoService:
         return result
 
     def print_workflow_details(self, workflow_name: str) -> None:
-        """Print workflow details for debugging failed workflows."""
+        """Print workflow details and namespace state for debugging failed workflows."""
         try:
             workflow_data = self._get_workflow_status_json(workflow_name)
             logger.info(
                 f"Workflow {workflow_name} full status:\n{json.dumps(workflow_data.get('status', {}), indent=2)}")
         except Exception as e:
             logger.error(f"Failed to get workflow details: {e}")
+
+        # Print namespace resource summary
+        try:
+            logger.info(f"===== Namespace {self.namespace} resource summary =====")
+            get_args = {"get": "pods,services,deployments,statefulsets,workflows", "--namespace": self.namespace}
+            self._run_kubectl_command(get_args)
+        except Exception as e:
+            logger.error(f"Failed to get namespace resources: {e}")
 
     def save_namespace_diagnostics(self, output_dir: str) -> Optional[str]:
         """Save detailed namespace diagnostics to a file for artifact collection."""


gregschohn (commented):
I completely agree that we need to find a better solution for our etcd deployment.

I'm torn between doing that through an operator or not.  Operators have a non-negligible tax, especially noticeable for local development - though we do use strimzi.  I don't think that we'd be able to deploy scalable kafka clusters w/out that though.  For etcd, we might need to vertically scale at some point for VERY wide migrations (hundreds to thousands of parallel workstreams - like RFS sessions - which may not be possible to do with the current argo-workflow implementation).  An operator might help with that.  In the case of disaster recovery (losing 2 or more nodes), I'd think that restarting would probably be a better option - or running w/ 5 replicas.  If we have a problem with operator reliability, we have an even bigger problem... it's something else that we need to monitor too.

Anyway - let's say that the operator makes sense - is this the right operator?  There's another 'aenix' operator.  Neither that nor druid look to be widespread.  druid is part of SAP's ecosystem and aenix sounds like it's a more humble open-source project.  

[Here's](https://gemini.google.com/share/23df6b4f1bf0) what gemini had to say comparing the different approaches.  I guess that for the medium term, I'd lean toward the aenix operator, given how we got burned by the changes to the etcd chart we were using before.  That said, any of the three options (druid, aenix, roll-our-own) are all likely an improvement over what we have and on the road to something better.  Let me know what you think about choosing between the two operators.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/etcd-cluster.yaml
```diff
@@ -0,0 +1,77 @@
...
+          
+          echo "Creating EtcdCluster..."
+          cat <<EOF | kubectl apply -f -
+          apiVersion: etcd.aenix.io/v1alpha1
+          kind: EtcdCluster
+          metadata:
+            name: etcd
+            namespace: {{ .Release.Namespace }}
+          spec:
+            replicas: {{ .Values.etcdCluster.replicas | default 1 }}
```
gregschohn: can you leave the default out of here.  I'd rather see us fail during templating if the replicas count wasn't in one of the values passed into helm

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -642,6 +610,7 @@ charts:
   kyverno:
     version: "3.5.2"
     repository: "https://kyverno.github.io/kyverno/"
+    namespace: kyverno
```
gregschohn: how will this work w/ the helm install when a user specifies the namespace?  This will always override it?  Mostly curious for now - what would happen if 2 MA charts were installed to separate namespaces, how would all of that work out?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/etcd-cluster.yaml
```diff
@@ -0,0 +1,77 @@
...
+      containers:
+      - name: create-etcd
+        image: {{ .Values.images.installer.repository }}:{{ .Values.images.installer.tag }}
+        command:
+        - /bin/sh
+        - -c
+        - |
+          set -e
+          echo "Waiting for etcd-operator CRD to be available..."
+          until kubectl get crd etcdclusters.etcd.aenix.io 2>/dev/null; do
```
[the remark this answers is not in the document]
AndreKurait: updated

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
AndreKurait: updated to use release namespace

PR #2136, opened 2026-01-08
Fix K8s local test reliability: Kyverno CRD race condition and cleanup improvements

AndreKurait:
### Description
This PR fixes several issues affecting K8s local test reliability:

1. **Kyverno CRD race condition**: The `ma-apply-kyverno-zero-resources` job was failing because it tried to apply a ClusterPolicy before the Kyverno CRDs were established. Added explicit wait for `crd/clusterpolicies.kyverno.io` before applying the policy.

2. **Stale webhook cleanup**: Kyverno registers cluster-scoped webhooks that persist after namespace deletion. When tests re-run, these stale webhooks cause failures like:
   ```
   failed calling webhook "mutate.kyverno.svc-fail": service "kyverno-svc" not found
   ```
   Added cleanup of Kyverno mutating/validating webhook configurations before namespace deletion.

3. **Pre-run cleanup**: Added namespace cleanup at the start of test runs to ensure a clean slate, removing any stale resources from previous interrupted runs.

4. **Simplified cleanup**: Replaced multi-step cleanup (helm uninstall → wait for pods → delete PVCs) with single namespace deletion, which is more reliable and faster.

5. **Bug fixes**:
   - Fixed PVC timeout error message that was trying to join PVC objects instead of their names
   - Fixed Kyverno policy preconditions to handle pods without serviceAccountName

### Issues Resolved
Fixes K8s local test failures caused by:
- Kyverno CRD not ready when policy job runs
- Stale Kyverno webhooks blocking pod creation after test reruns

### Testing
- Verified fix on Jenkins CI job elasticsearch-5x-k8s-local-test
- Kyverno installs correctly in release namespace (ma)
- ClusterPolicy `zero-resource-requests` applies successfully
- Workflow pods are mutated with zero resource requests
- Cleanup properly removes webhooks before namespace deletion

### Check List

### DIFF
--- a/libraries/testAutomation/testAutomation/k8s_service.py
+++ b/libraries/testAutomation/testAutomation/k8s_service.py
@@ -163,7 +163,7 @@ class K8sService:
             elapsed = time.time() - start_time
             if elapsed > timeout_seconds:
                 raise TimeoutError(f"Timeout reached: Not all PVCs were deleted within {timeout_seconds} seconds. "
-                                   f"Remaining PVCs: {', '.join(remaining_pvcs)}")
+                                   f"Remaining PVCs: {[pvc.metadata.name for pvc in remaining_pvcs]}")
 
             logger.info(f"Waiting for PVCs to be deleted. Remaining: {[pvc.metadata.name for pvc in remaining_pvcs]}")
             time.sleep(poll_interval)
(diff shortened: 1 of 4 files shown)


libraries/testAutomation/testAutomation/k8s_service.py
```diff
@@ -182,6 +182,16 @@ def create_namespace(self, namespace: str) -> CompletedProcess | None:
...
 
+    def delete_namespace(self) -> None:
+        logger.info(f"Deleting namespace '{self.namespace}'")
+        # Delete cluster-scoped kyverno webhooks that reference this namespace
+        self.run_command(["kubectl", "delete", "mutatingwebhookconfigurations",
+                         "-l", "app.kubernetes.io/instance=kyverno", "--ignore-not-found"], ignore_errors=True)
+        self.run_command(["kubectl", "delete", "validatingwebhookconfigurations",
+                         "-l", "app.kubernetes.io/instance=kyverno", "--ignore-not-found"], ignore_errors=True)
+        self.run_command(["kubectl", "delete", "namespace", self.namespace,
+                         "--ignore-not-found", "--grace-period=0", "--force"])
```
gregschohn: All of these delete calls are blocking, right?

AndreKurait: yes

gregschohn (approved):

PR #2138, opened 2026-01-09
Fix duplicate key error in delta document migration

AndreKurait:
### Description

Fixes a bug in delta snapshot migrations where the same document ID appearing in both deletions and additions streams would cause a bulk request failure:

```
IllegalStateException: Duplicate key 996376421 (attempted merging values 
DeleteOp(...) and IndexOp(...))
  at OpenSearchClient.sendBulkRequest(OpenSearchClient.java:434)
  at DocumentReindexer.sendBulkRequest(DocumentReindexer.java:124)
```

This occurred because deletions and additions were concatenated into a single stream and processed in parallel batches, allowing a delete and index for the same document to end up in the same bulk request batch.

**Root cause**: When a document is deleted and then re-created between snapshots, the delta engine emits both a DeleteOp and an IndexOp for the same document ID. Previously these were merged via `Flux.concat()` and batched together, but `OpenSearchClient.sendBulkRequest()` uses `Collectors.toMap()` which throws on duplicate keys.

**Solution**: Process all deletions first, wait for completion, then process all additions. This ensures delete operations complete before any index operations begin.

### Changes

- `DocumentReaderEngine.prepareChangeset()` (renamed from `readDocuments()`) now returns `DocumentChangeset` record with separate `deletions` and `additions` Flux streams plus a `cleanup` Runnable
- `DocumentReindexer.reindex()` processes deletions first via `thenMany()`, ensuring sequential execution
- Added `LuceneDirectoryReader.getCleanupRunnable()` helper for consistent reader cleanup across engines
- `LuceneReader.readDocsByLeavesFromStartingPosition()` now returns `Flux` directly instead of `Publisher`

### Refactoring

Renamed classes and methods to better reflect document change semantics:
- `RfsLuceneDocument` → `LuceneDocumentChange` - represents a change operation, not just a document
- `RfsDocumentOperation` → `DocumentChangeType` - clearer naming for INDEX/DELETE enum
- `DocumentsResult` → `DocumentChangeset` - better describes the additions/deletions container
- `readDocuments()` → `streamDocumentChanges()` in `LuceneIndexReader`
- `readDocuments()` → `prepareChangeset()` in `DocumentReaderEngine`

### Issues Resolved

N/A - discovered during production migration

### Testing

- Existing unit tests updated and passing
- Manual verification with delta migration scenario

### Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
@@ -9,7 +9,7 @@ import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
-import org.opensearch.migrations.bulkload.common.DocumentReaderEngine.DocumentsResult;
+import org.opensearch.migrations.bulkload.common.DocumentReaderEngine.DocumentChangeset;
 import org.opensearch.migrations.bulkload.common.bulk.BulkNdjson;
 import org.opensearch.migrations.bulkload.worker.WorkItemCursor;
 import org.opensearch.migrations.reindexer.tracing.IDocumentMigrationContexts.IDocumentReindexContext;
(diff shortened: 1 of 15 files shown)


RFS/src/main/java/org/opensearch/migrations/bulkload/common/DocumentReindexer.java
```diff
@@ -58,7 +59,14 @@ public DocumentReindexer(OpenSearchClient client,
         this.allowlist = allowlist;
     }
 
-    public Flux<WorkItemCursor> reindex(String indexName, Flux<RfsLuceneDocument> documentStream, IDocumentReindexContext context) {
+    public Flux<WorkItemCursor> reindex(String indexName, DocumentsResult documents, IDocumentReindexContext context) {
+        // Process deletions first, then additions, cleanup always runs at the end
+        return reindexStream(indexName, documents.deletions(), context)
```
gregschohn: what exactly are the types of elements in the deletion stream?  How does reindexStream know that the docs should be scrubbed?

gregschohn: I'll ask the same question as for the other comment - is this the right name for the class?  Operation is a verb, so the document now becomes something else - like a change.  I haven't looked at these models in months and it's apparent that they need a bit more work.  The concepts are fine, but to make sure that they're maintainable, we should invest in making the programming models match the concepts well

gregschohn (approved):

PR #2150, opened 2026-01-12
Add _source reconstruction from stored fields, doc_values, and Points (BKD trees)

AndreKurait:
## Description

When `_source` is disabled or has excluded fields in the source cluster, RFS migration would previously lose that field data. This PR adds automatic reconstruction of document `_source` from stored fields, doc_values, and Points (BKD trees) using type-aware conversion based on index mappings.

### Key Changes

**Core Infrastructure:**
- Add `SourceReconstructor` to rebuild `_source` with four-tier fallback: stored fields → doc_values → Points → Terms
- Add `EsFieldType` enum classifying 11 ES field type groups for conversion logic
- Add `FieldMappingInfo` record with type, mappingType, format, scalingFactor, and docValues flag
- Add `FieldMappingContext` to parse index mappings once per work item
- Add `DocValueFieldInfo` interface with boolean detection via T/F indexed terms

**Lucene Version Support:**
- Implement doc_values accessors for all Lucene versions (5, 6, 7, 9)
- Add `getPointValues()` to `LuceneLeafReader` interface for BKD tree access
- Add `getValueFromTerms()` for boolean recovery via terms index
- Implement Points extraction in LeafReader5/6/7/9

**Type-Aware Conversion:**

| Type | Conversion Logic |
|------|------------------|
| Float/double | Decode from sortable long bits |
| Half-float | 16-bit IEEE 754 binary16 encoding |
| IP | 16-byte IPv6-mapped format; ES 2.x 32-bit IPv4 |
| Date | Epoch millis to ISO format (respects `epoch_millis` format) |
| Date_nanos | Epoch nanos to ISO format with nanosecond precision |
| Scaled_float | Divide by `scaling_factor` from mapping |
| Unsigned_long | Handle values > Long.MAX_VALUE via BigInteger |
| Geo_point | Morton decoding (ES 2+); binary format (ES 1.x - 2 doubles) |
| Boolean | "T"/"F" in stored/terms; 0/1 in doc_values |
| Binary | Base64 encode |
| Wildcard | Base64 decoding for doc_values |
| Token_count | Stored as integer (token count from analyzer) |

**Recovery Priority:** stored fields (fastest) → doc_values → Points (BKD trees) → Terms index

### Recovery Methods by Source Version

| Source Version | Stored Fields | Doc Values | Points | Terms | Notes |
|----------------|---------------|------------|--------|-------|-------|
| ES 1.x | ✅ | ✅* | ❌ | ✅ | *Requires explicit `doc_values: true` in mapping |
| ES 2.x | ✅ | ✅ | ❌ | ✅ | Full stored/doc_values support |
| ES 5.x | ✅ | ✅ | ✅ | ✅ | Full support including Points |
| ES 6.x | ✅ | ✅ | ✅ | ✅ | Full support |
| ES 7.x | ✅ | ✅ | ✅ | ✅ | Full support |
| ES 8.x | ✅ | ✅ | ✅ | ❌ | No terms index for boolean (Lucene 9) |
| OS 1.x | ✅ | ✅ | ✅ | ✅ | Full support |
| OS 2.x | ✅ | ✅ | ✅ | ✅ | Full support |

### Field Type Support Matrix

| Field Type | Doc Values | Points | Stored | Terms | Version |
|------------|------------|--------|--------|-------|---------|
| `byte`, `short`, `integer`, `long` | ✅ | ✅ ES5+ | ✅ | ❌ | All |
| `float`, `double`, `half_float` | ✅ | ✅ ES5+ | ✅ | ❌ | ES 5+ |
| `scaled_float` | ✅ | ❌ | ✅ | ❌ | ES 5+ |
| `unsigned_long` | ✅ | ❌ | ✅ | ❌ | OS 2.8+ |
| `token_count` | ✅ | ✅ ES5+ | ✅ | ❌ | ES 5+ |
| `boolean` | ✅ ES2+ | ❌ | ✅ | ✅* | All |
| `date` | ✅ | ✅ ES5+ | ✅ | ❌ | ES 2+ |
| `date_nanos` | ✅ | ✅ | ✅ | ❌ | ES 7+ |
| `ip` | ✅ | ✅ ES5+ | ✅ | ❌ | ES 2+ |
| `geo_point` | ✅ | ❌ | ❌ | ❌ | All |
| `keyword` | ✅ | ❌ | ✅ | ❌ | ES 5+ |
| `string` (ES 1.x-4.x) | ✅ | ❌ | ✅ | ❌ | ES 1.x-4.x |
| `wildcard` | ✅ | ❌ | ✅ | ❌ | ES 7.11+ |
| `binary` | ✅ ES2+ | ❌ | ✅ | ❌ | All |
| `constant_keyword` | ❌** | ❌ | ❌ | ❌ | ES 7.11+ |

\* Terms index not available in ES 8+ (Lucene 9)
\*\* constant_keyword value is in mapping, not index - NOT recoverable

### Field Types Tested

All field types tested across ES 1.7, 2.4, 5.6, 6.8, 7.10, 7.17, 8.17, OS 1.3, OS 2.19 → OS 3.0:

- **String types:** string (ES 1.x-4.x), keyword, wildcard
- **Numeric types:** byte, short, integer, long, float, double, half_float, scaled_float, unsigned_long, token_count
- **Date types:** date (ISO and epoch_millis formats), date_nanos
- **Other types:** boolean, binary, ip, ipv6, geo_point

### Known Limitations

- **constant_keyword:** Value stored in mapping, NOT recoverable from Lucene index
- **Range types:** Complex binary format, not supported
- **Object/nested:** Flattened structure, cannot reconstruct
- **geo_shape, vectors, etc.:** Complex types not supported

## Testing

- Comprehensive parameterized test (`NoStoredSourceMigrationTest`) covering all version pairs
- Unit tests for `EsFieldType`, `FieldMappingInfo`, `FieldMappingContext`
- Tests all permutations of doc_values (enabled/disabled) × store (true/false)
- Tests mergeWithDocValues for _source with excluded fields
- Tests unrecoverable document handling

## Documentation

- Updated `RFS/docs/FIELD_TYPE_CONVERSION.md` with detailed recovery reference

## Check List


build.gradle
```diff
@@ -216,7 +216,7 @@ subprojects {
                 systemProperty 'junit.jupiter.extensions.autodetection.enabled', true
                 // Verify assertions in tests
                 jvmArgs += ['-ea', '-XX:+HeapDumpOnOutOfMemoryError', '-Xmx2g']
-                systemProperty "junit.jupiter.execution.timeout.default", "5m"
+                systemProperty "junit.jupiter.execution.timeout.default", "10m"
```
AndreKurait: These added tests are fast, this is just for time to download the container images

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/NoStoredSourceMigrationTest.java
```diff
@@ -0,0 +1,709 @@
...
+            Arguments.of(SearchClusterContainer.ES_V8_17, SearchClusterContainer.OS_V3_0_0),
+            Arguments.of(SearchClusterContainer.OS_V1_3_20, SearchClusterContainer.OS_V3_0_0),
+            Arguments.of(SearchClusterContainer.OS_V2_19_4, SearchClusterContainer.OS_V3_0_0)
+        );
+    }
+
+    @TempDir
+    private File localDirectory;
+
+    private static boolean isTypeAvailable(FieldTypeConfig config, ContainerVersion version) {
```
gregschohn: This method gives me some significant concerns about our taxonomy to organize feature capabilities.  If a new feature comes in, do we have to know to put a bisection in everywhere that has code like this?  All of might be cleaner with a virtual and a factory that adds/overrides capabilities as the version is considered at progressively tighter lenses.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/NoStoredSourceMigrationTest.java
```diff
@@ -0,0 +1,709 @@
...
+            sourceCluster.start();
+            targetCluster.start();
+
+            var sourceOps = new ClusterOperations(sourceCluster);
+            var targetOps = new ClusterOperations(targetCluster);
+
+            String indexName = "source_reconstruction_test";
+            String docType = needsDocType(sourceVersion) ? "doc" : null;
+
+            StringBuilder props = new StringBuilder();
```
gregschohn: can this be a map that then gets stringified by jackson?  The appends and backslashes are tough to read and take away from the intent.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/NoStoredSourceMigrationTest.java
```diff
@@ -0,0 +1,709 @@
...
+
+            sourceOps.createIndex(indexName, indexBody);
+            sourceOps.createDocument(indexName, "1", doc, null, docType);
+            sourceOps.post("/_refresh", null);
+
+            var snapshotCtx = SnapshotTestContext.factory().noOtelTracking();
+            createSnapshot(sourceCluster, "snap", snapshotCtx);
+            sourceCluster.copySnapshotData(localDirectory.toString());
+
+            StringBuilder targetProps = new StringBuilder();
```
gregschohn: consider breaking this & other parts into a separate pieces to make the code easier to read.
I suspect that this might be one of those rare tests that we show to others outside of this project so that they can understand how this tooling and a migration will play out.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/NoStoredSourceMigrationTest.java
```diff
@@ -0,0 +1,709 @@
...
+            // included_field should be from _source
+            assertEquals("included_value", source.get("included_field").asText());
+            // excluded_field should be recovered from doc_values via mergeWithDocValues
+            assertEquals(42, source.get("excluded_field").asInt());
+        }
+    }
+
+    /**
+     * Tests that documents with no recoverable fields are skipped.
+     * Uses binary field with store:false (no doc_values, no Points, no stored field).
```
gregschohn: This is surprising, but maybe there's no other option?  If I ingest 5 documents w/ different binary fields, I'll see that RFS runs and has moved 0 documents?  That just doesn't seem right...
From the test, it looks like you're counting the returned docs for a search.  What does the index claim to have?  Would it have those 5 documents?

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/version_5/LeafReader5.java
```diff
@@ -1,17 +1,34 @@
...
+import shadow.lucene5.org.apache.lucene.index.Terms;
+import shadow.lucene5.org.apache.lucene.index.TermsEnum;
+import shadow.lucene5.org.apache.lucene.store.ByteArrayDataInput;
 import shadow.lucene5.org.apache.lucene.util.Bits;
+import shadow.lucene5.org.apache.lucene.util.BytesRef;
 import shadow.lucene5.org.apache.lucene.util.FixedBitSet;
 import shadow.lucene5.org.apache.lucene.util.SparseFixedBitSet;
 
+@Slf4j
 public class LeafReader5 implements LuceneLeafReader {
```
gregschohn: Is there a place that I can find out which of these methods is called for each doc vs ones that are called just once for the index?

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneField.java
```diff
@@ -10,4 +10,13 @@ public interface LuceneField {
 
     public String utf8ToStringValue();
 
+    /** Returns numeric value if field is numeric, null otherwise */
```
gregschohn: Can you document if this can hold any kind of number?  My expectation is that it can.

gregschohn (commented):
I'd like to see instrumentation for the different field types and retrieval strategies.  When looking into the speed or quality of a backfill, were there lots of doc-value lookups?  Metrics will help us understand immediately some of the traits of a migration and help diagnose future issues.  I'd like to see some tests to assert that those values are being updated too (just by instrumenting existing tests).

RFS/src/test/java/org/opensearch/migrations/bulkload/lucene/FieldMappingContextTest.java
```diff
@@ -0,0 +1,136 @@
...
+        JsonNode mappings = MAPPER.readTree(json);
+        FieldMappingContext context = new FieldMappingContext(mappings);
+
+        assertEquals(2, context.getFieldNames().size());
+        assertEquals(EsFieldType.STRING, context.getFieldInfo("name").type());
+        assertEquals(EsFieldType.NUMERIC, context.getFieldInfo("age").type());
+    }
+
+    @Test
+    void testTypedMappings() throws Exception {
```
gregschohn: thanks - hard to believe that it took this long to model and test this stuff

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -0,0 +1,542 @@
...
+    /**
+     * Reconstructs _source JSON from stored fields and doc_values for a document.
+     * Uses mapping context for type-aware conversion when available.
+     */
+    public static String reconstructSource(LuceneLeafReader reader, int docId, LuceneDocument document, FieldMappingContext mappingContext) {
+        try {
+            Map<String, Object> reconstructed = new LinkedHashMap<>();
+            
+            // First, read from stored fields (more performant)
+            for (var field : document.getFields()) {
```
gregschohn: Could you precompute this list based upon the index configuration - you should know in the next loop if the field should have a value or not, right?  For other cases, you could probably do some upfront work too.  It might make the code easier to read and debug too. It seems like you might be doing some of that, but I'd expect to see no continue statements in these loops - just grab exactly the right data as per the plan - or throw, etc - and move on.

RFS/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -0,0 +1,542 @@
...
+        } catch (Exception e) {
+            return Base64.getEncoder().encodeToString(bytes);
+        }
+    }
+
+    // Constants for Morton decoding (ES 2.x geo_point format)
+    private static final long[] MAGIC = {
+        0x5555555555555555L, 0x3333333333333333L, 0x0F0F0F0F0F0F0F0FL,
+        0x00FF00FF00FF00FFL, 0x0000FFFF0000FFFFL, 0x00000000FFFFFFFFL
+    };
```
gregschohn: these blocks should be in separate helper classes to reduce scope.  MAGIC and SHIFT could loop appealing for other purposes.

AndreKurait:
Jira: [MIGRATIONS-2965](https://opensearch.atlassian.net/browse/MIGRATIONS-2965)

sumobrian:
This PR is stale. Support would need to be added for _source-less Elastic.

PR #2168, opened 2026-01-20
[Update] index template exclusion list

jugal-chauhan:
### Description
A recent discovery revealed that AWS Managed Elasticsearch 6.8 domains create default index templates with names like:
- `kibana_index_template:.kibana`
- `kibana_index_template:.kibana_*`

These templates were not being filtered during migration because they don't start with `.` , rather they start with `kibana_index_template:.`. When Migration Assistant attempted to create these on an OpenSearch 2.x target, the metadata migration step failed as OpenSearch 2.x+ does not allow `*` in template names, and these Kibana system templates should not be migrated regardless.
```
invalid_index_template_exception: index_template [kibana_index_template:.kibana_*] invalid, 
cause [Validation Failed: 1: name must not contain a '*';]
```

WIth this PR, Migration Assistant will ignore anything starting with `kibana` , such as :
- `kibana_index_template:.kibana_*`
- `kibana_logs`
- `kibana`

Index and Template names containing `kibana` elsewhere is still allowed to migrate, such as :  
- `my_kibana_logs`
- `data_kibana`

### Issues Resolved
N/A

### Testing
- Added a unit test for this test case at `RFS/src/test/java/org/opensearch/migrations/bulkload/common/FilterSchemeTest.java`

### Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/FilterScheme.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/FilterScheme.java
@@ -19,7 +19,7 @@ public class FilterScheme {
             "elastic-connectors-",
             "elastic_agent.",
             "ilm-history-",
-            "kibana_index_template:",
+            "kibana",
             "logs-elastic_agent",
             "logs-endpoint.",
             "logs-index_pattern",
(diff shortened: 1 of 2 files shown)


RFS/src/main/java/org/opensearch/migrations/bulkload/common/FilterScheme.java
```diff
@@ -19,6 +19,7 @@ private FilterScheme() {}
             "elastic-connectors-",
             "elastic_agent.",
             "ilm-history-",
+            "kibana_index_template:",
```
AndreKurait: I'd prefer this without the : for a wider match, honestly even `kibana_` or `kibana`

PR #2199, opened 2026-01-30
Cross platform eks gradle builds

gregschohn:
### Description
Fix cross-platform builds with buildkit when we're backed by EKS

For minikube builds, we continue to do cross-platform docker builds with qemu emulation in a single buildkit pod.  If we're in EKS, we don't deploy a pod directly, but use docker buildx's k8s driver (`--driver=kubernetes`) to manage the buildkit pods in the k8s cluster.

In either the minikube or EKS case, the expectation is that the buildImages chart has been installed with the name 'buildkit' (in the 'buildkit' namespace).  The chart itself deploys different components depending on its variables.  For EKS, the chart is configured to deploy a nodepool with arm64 & amd64 instances but does NOT deploy any pods.  For minikube, we don't deploy a nodepool, but DO deploy a pod that runs buildkit.  There is still an option for the chart to run an installation job to build buildImagesToRegistry from a repo/branch (when skipBuildJob=false).

After the chart has been installed, setupK8sBuilder.sh can be called to configure the docker builder.  This is separate because the install job, when enabled, will run this as well.  When the script detects if the K8s cluster is cloud-based, it creates two docker builders and merges them together (--append).  Each of the builders is configured for one of the platforms and is deployed to a specific architecture type, specified by tolerations.  For these cloud-based deployments, we don't manage the pods at all, but let docker buildx do that with its kubernetes driver.

Unfortunately, buildkit when deployed in minikube has some DNS issues so we can't use the kubernetes driver.  For minikube docker builds, we still rely upon deploying a pod, typically (if following the instructions in the READMEs) from the buildImages chart (buildkit installation).

The aws-bootstrap script, when --build-images-locally is passed, will now run setUpK8sImageBuildServices.sh to configure buildkit appropriately (as described above).

aws-bootstrap.sh has also added the ability to pass more helm values and the ability to DISABLE the general-purpose node pool.  That combination allows one to disable the general pool and replace it with another pool (--set cluster.useCustomKarpenterNodePool=true).  The custom node pool, but default will support both amd64 & arm64.  The general purpose one ONLY supports amd64.  Notice that the migration console nodepool ALREADY supports both graviton/amd64 (which is why we needed cross-platform builds).  The custom 'general-work-pool' has a priority over the general-purpose pool, so it will be preferred.  The custom node pool will be created will limit the TOTAL CPU/Memory that the pool can consume (change workloadsNodePool.limits to adjust).  Those limits could be very valuable for deployments concerned about limiting the total spend (at least per hour).  However, the limits aren't useful as long as there's a general-purpose pool that pods will spill-over to if the scheduler runs out of capacity on the custom nodepool, hence the reason that the general purpose pool can be disabled.

When the general-purpose pool is disabled, the aws-bootstrap script will enable it to do the helm deployments and will then disable it at the end of the script.  If the custom nodepool isn't installed yet and there's no general purpose pool, there won't be any way to run any jobs to install more resources, so for bootstrapping, it's required.  The custom pool features aren't documented yet because we probably have some more tuning to do on the defaults for the pool.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-2867

### Testing
Manually ran instructions in aws/README.md to construct a new EKS cluster, etc.  Manually ran localTesting.sh on my laptop after deleting minikube and starting fresh.

### Check List

### DIFF
--- a/buildImages/setupK8sBuilders.sh
+++ b/buildImages/setupK8sBuilders.sh
@@ -27,7 +27,7 @@ if [[ "$CONTEXT" =~ (eks:|gke_|aks-) ]]; then
     --node=builder-amd64 \
     --driver-opt="namespace=${NAMESPACE}" \
     --driver-opt="nodeselector=kubernetes.io/arch=amd64" \
-    --driver-opt="tolerations=key=build-nodepool,value=true,effect=NoSchedule"
+    --driver-opt='"tolerations=key=build-nodepool,value=true,effect=NoSchedule"'
 
   docker buildx create \
     --append \
@@ -37,7 +37,7 @@ if [[ "$CONTEXT" =~ (eks:|gke_|aks-) ]]; then
     --node=builder-arm64 \
     --driver-opt="namespace=${NAMESPACE}" \
     --driver-opt="nodeselector=kubernetes.io/arch=arm64" \
-    --driver-opt="tolerations=key=build-nodepool,value=true,effect=NoSchedule"
+    --driver-opt='"tolerations=key=build-nodepool,value=true,effect=NoSchedule"'
 else
   echo "Detected local K8s, using remote driver with port-forwards"
   
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -12,6 +12,9 @@
 #   Run directly: curl -s https://raw.githubusercontent.com/opensearch-project/opensearch-migrations/main/deployment/k8s/aws/aws-bootstrap.sh | bash
 #   Save & run:   curl -s -o aws-bootstrap.sh https://raw.githubusercontent.com/opensearch-project/opensearch-migrations/main/deployment/k8s/aws/aws-bootstrap.sh && chmod +x aws-bootstrap.sh && ./aws-bootstrap.sh
 # -----------------------------------------------------------------------------
+
+set -euo pipefail
+
 # --- defaults ---
 org_name="opensearch-project"
 repo_name="opensearch-migrations"
--- a/buildImages/setUpK8sImageBuildServices.sh
+++ b/buildImages/setUpK8sImageBuildServices.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 MIGRATIONS_REPO_ROOT_DIR=$(git rev-parse --show-toplevel)
 
 echo "Installing buildImages helm chart for nodepool..."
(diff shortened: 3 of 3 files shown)


gregschohn:
Further arguing the deprecation of --build-images...

[Here's](https://opensearch.atlassian.net/browse/MIGRATIONS-2867?focusedCommentId=20408) a way to build an image and distribute it as a tarball for somebody else to load it.  We can also distribute helm charts as tarballs too.
I would argue that we should also be able to get our CICD systems to be able to build - and release off of branches, that would alleviate this too.
Branches + tarball images in the releases (from release drafter) would mean that any time that we wanted to give the user a branch, they'd be able to install the build with a couple steps, even if the image wasn't ready to get pushed up as a full public release.  If a user needs to make a proprietary change, they can follow the developer scenario - we don't need to invent a solution for that problem.

AndreKurait:
**Bug: Wrong builder name in aws-bootstrap.sh:370**

```bash
if docker buildx inspect k8s-builder --bootstrap &>/dev/null; then
```

Should be `local-remote-builder` to match what `setupK8sBuilders.sh` creates. This check will always fail, causing unnecessary re-setup every time.

buildImages/setupK8sBuilders.sh
```diff
@@ -0,0 +1,64 @@
...
+
+  docker buildx create \
+    --append \
+    --name="$BUILDER_NAME" \
+    --driver=kubernetes \
+    --platform=linux/arm64 \
+    --node=builder-arm64 \
+    --driver-opt="namespace=${NAMESPACE}" \
+    --driver-opt="nodeselector=kubernetes.io/arch=arm64" \
+    --driver-opt='"tolerations=key=build-nodepool,value=true,effect=NoSchedule"'
```
AndreKurait: Same quoting issue.

```suggestion
    --driver-opt="tolerations=key=build-nodepool,value=true,effect=NoSchedule"
```

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -330,17 +364,33 @@ if [[ "$skip_git_pull" == "false" ]]; then
   popd > /dev/null || exit
 fi
 
+EXTRA_VALUES_FLAG=""
+if [[ -n "$extra_helm_values" ]]; then
+  EXTRA_VALUES_FLAG="-f $extra_helm_values"
+fi
```
AndreKurait: Remove this and use `${extra_helm_values:+-f "$extra_helm_values"}` inline at line 484 instead. Handles spaces in paths correctly.

```suggestion
```

PR #2226, opened 2026-02-12
[Add] RFS coordinator tests to validate retries on RFS

jugal-chauhan:
### Description
Adds an integration test for RFS with a dedicated OpenSearch coordinator cluster (separate from the target), validating behavior when the coordinator becomes unavailable during document migration.

ToxiProxy injects deterministic coordinator faults via a `ScheduledExecutorService` while the target data path remains unaffected. RFS runs via `RfsMigrateDocuments.run()` with separate connection contexts for target and coordinator. Timing is controlled by `1000ms upstream latency` on the target proxy (~1 doc/sec, 60 docs ≈ 60s migration), with coordinator outage injected at t=30s mid migration.

Also adds `coordinatorAddress/coordinatorVersion` parameters to `SourceTestBase.migrateDocumentsWithOneWorker` so any future test can use a dedicated coordinator cluster (defaults to target when null, all existing callers unchanged).

### Test 1: `allDocsMigratedButCoordinatorUnavailableAtCompletion` ✅ Active, passing

Scenario: Coordinator goes down at t=30s (mid-migration) and never comes back.

Observed RFS behavior:
- All 60 docs successfully migrate to target despite coordinator outage
- RFS completes document reindexing, then fails when attempting to mark work item as completed
- Coordinator working state contains incomplete work items (no completedAt field)
- RFS throws 
```
OpenSearchWorkCoordinator$RetriesExceededException caused by io.netty.channel.unix.Errors$NativeIoException: readAddress(..) failed with error(-54): Connection reset by peer
```

Assertions: Timeline ordering (rfsStarted < outageInjected < rfsEnded), all docs on target, incomplete coordinator work items, RetriesExceededException type check.

### Test 2: `allDocsMigratedButCoordinatorLongRestartsAtCompletion` 🔜 @Disabled

Scenario: Coordinator goes down at t=30s, comes back at t=150s (120s restart window)

Observed RFS behavior:
- All 60 docs successfully migrate to target
- RFS exhausts its retry budget (~30s of retries) before the coordinator returns at t=150s
- Throws 
```
OpenSearchWorkCoordinator$RetriesExceededException: reactor.core.Exceptions$ReactiveException: io.netty.channel.unix.Errors$NativeIoException: readAddress(..) failed with error(-54): Connection reset by peer                                                                                                                                 

Fails on assertion: Expected RFS to recover and complete cleanly after coordinator returns ==> expected: <null> but was: <RetriesExceededException>
```
Assertions: Full timeline ordering (rfsStarted < outageInjected < reEnabled < rfsEnded), all docs on target, no exception, all coordinator work items completed.

Currently fails because RFS exhausts its retry budget (~1.3s) before the coordinator returns. This test documents the desired recovery behavior and will be enabled by the retry/backoff fix in [MIGRATIONS-2862](https://opensearch.atlassian.net/browse/MIGRATIONS-2862)

### Implementation
- Calls `RfsMigrateDocuments.run()` directly with separate `ConnectionContext` objects for `target` and `coordinator`
- `1000ms` upstream latency on target proxy == ~1 doc/sec (60 docs ≈ 60s), coordinator outage at t=30s hits mid-migration
- Disabled coordinator's connection at 30s using ToxiProxy

### Issues Resolved
[MIGRATIONS-2861](https://opensearch.atlassian.net/browse/MIGRATIONS-2861)

### Testing
```
gradlew :DocumentsFromSnapshotMigration:isolatedTest --tests org.opensearch.migrations.bulkload.RfsOpenSearchCoordinatorOutageTest --no-daemon --rerun-tasks
```

### Check List

### DIFF
--- a/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/configManagementHelpers.ts
+++ b/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/configManagementHelpers.ts
@@ -44,7 +44,7 @@ export const ConfigManagementHelpers = WorkflowBuilder.create({
             .addImageInfo(b.inputs.imageMigrationConsoleLocation, b.inputs.imageMigrationConsolePullPolicy)
             .addInputsAsEnvVars({prefix:"", suffix: ""})
             .addCommand(["sh", "-c"])
-            .addResources(DEFAULT_RESOURCES.MIGRATION_CONSOLE_CLI)
+            .addResources(DEFAULT_RESOURCES.SHELL_MIGRATION_CONSOLE_CLI)
             .addArgs([decrementTlhScript])
 
             .addPathOutput("shouldFinalize", "/tmp/should-finalize", typeToken<boolean>())
@@ -58,7 +58,7 @@ export const ConfigManagementHelpers = WorkflowBuilder.create({
             .addImageInfo(b.inputs.imageMigrationConsoleLocation, b.inputs.imageMigrationConsolePullPolicy)
             .addInputsAsEnvVars({prefix: "", suffix: ""})
             .addCommand(["sh", "-c"])
-            .addResources(DEFAULT_RESOURCES.MIGRATION_CONSOLE_CLI)
+            .addResources(DEFAULT_RESOURCES.SHELL_MIGRATION_CONSOLE_CLI)
             .addArgs([cleanupTlhScript])
         )
     )
--- a/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/createSnapshot.ts
+++ b/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/createSnapshot.ts
@@ -131,7 +131,7 @@ export const CreateSnapshot = WorkflowBuilder.create({
             .addEnvVar("JDK_JAVA_OPTIONS",
                 expr.dig(expr.deserializeRecord(b.inputs.createSnapshotConfig), ["jvmArgs"], "")
             )
-            .addResources(DEFAULT_RESOURCES.MIGRATION_CONSOLE_CLI)
+            .addResources(DEFAULT_RESOURCES.JAVA_MIGRATION_CONSOLE_CLI)
             .addArgs([
                 expr.literal("---INLINE-JSON"),
                 expr.asString(expr.serialize(
--- a/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
+++ b/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
@@ -84,7 +84,7 @@ export const FullMigration = WorkflowBuilder.create({
         .addContainer(cb => cb
             .addImageInfo(cb.inputs.imageMigrationConsoleLocation, cb.inputs.imageMigrationConsolePullPolicy)
             .addCommand(["sh", "-c"])
-            .addResources(DEFAULT_RESOURCES.MIGRATION_CONSOLE_CLI)
+            .addResources(DEFAULT_RESOURCES.SHELL_MIGRATION_CONSOLE_CLI)
             .addArgs(["echo runReplayerForTarget"])))
 
 
--- a/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/metadataMigration.ts
+++ b/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/metadataMigration.ts
@@ -149,7 +149,7 @@ export const MetadataMigration = WorkflowBuilder.create({
                 expr.dig(expr.deserializeRecord(b.inputs.metadataMigrationConfig), ["jvmArgs"], "")
             )
             .addEnvVarsFromRecord(getTargetHttpAuthCreds(getHttpAuthSecretName(b.inputs.targetConfig)))
-            .addResources(DEFAULT_RESOURCES.MIGRATION_CONSOLE_CLI)
+            .addResources(DEFAULT_RESOURCES.JAVA_MIGRATION_CONSOLE_CLI)
             .addCommand(["/root/metadataMigration/bin/MetadataMigration"])
             .addArgs([
                 b.inputs.commandMode,
--- a/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/migrationConsole.ts
+++ b/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/migrationConsole.ts
@@ -145,7 +145,7 @@ export const MigrationConsole = WorkflowBuilder.create({
                 expr.dig(expr.deserializeRecord(c.inputs.configContents), ["source_cluster","authConfig","basic","secretName"], "")))
             .addEnvVarsFromRecord(getTargetHttpAuthCreds(
                 expr.dig(expr.deserializeRecord(c.inputs.configContents), ["target_cluster","authConfig","basic","secretName"], "")))
-            .addResources(DEFAULT_RESOURCES.MIGRATION_CONSOLE_CLI)
+            .addResources(DEFAULT_RESOURCES.JAVA_MIGRATION_CONSOLE_CLI)
             .addArgs([
                 expr.fillTemplate(SCRIPT_ARGS_FILL_CONFIG_AND_RUN_TEMPLATE, {
                     "FILE_CONTENTS": expr.toBase64(expr.asString(c.inputs.configContents)),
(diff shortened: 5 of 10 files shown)


DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,355 @@
...
+
+    // Outage trigger
+    private static final String OUTAGE_TRIGGER_LOG_LINE = "Reindexing completed for Index";
+
+    @TempDir
+    Path tempRootDir;
+
+    @Test
+    @SneakyThrows
+    void shouldPreserveMigratedDocsButFailFinalizationWhenCoordinatorUnavailableStartsPostBackfill() {
```
gregschohn: As per the jira task, this is misnamed.  That task "verifies RFS can recover gracefully when the coordinator becomes unavailable".  It should be in a failing (need to be skipped state) for now, but the end game is to test that while it fails now, once we get proper retries and time values, that it will work.  

Think about what we're going to change and think about how this test will help us develop the fix.  I would name this something like "allDocsMigratedButCoordinatorRestartsAtCompletion".

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,355 @@
...
+            var preOutageState = readCoordinatorState(osCoordinatorContainer.getUrl());
+            Assertions.assertTrue(countIncompleteWorkItems(preOutageState) > 0,
+                "No incomplete work items before outage injection: " + preOutageState);
+
+            // Run RFS for last shard
+            var result = runFinalWorkerWithMidWorkerOutage(
+                tempDirSnapshot, targetProxy.getProxyUriAsString(), coordinatorProxy, workerArgs);
+
+            var finalDocs = getTargetDocCount(osTargetContainer.getUrl(), "geonames", false);
+            var coordStateAfter = readCoordinatorState(osCoordinatorContainer.getUrl());
```
gregschohn: It would be good to get a verification on this - do you have any shards left?  The mutations/side effects in the coordinator cluster are the ones that you care about.  You should be checking them (w/ refreshes) - and providing some buffer time of a few seconds, about an order of magnitude more than where you expect your systems to be performing.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,355 @@
...
+        new CreateSnapshot(snapshotArgs, testSnapshotContext.createSnapshotCreateContext()).run();
+        esSource.copySnapshotData(snapshotDir.toString());
+
+        var expectedDocs = getTargetDocCount(esSource.getUrl(), "geonames", true);
+        Assertions.assertTrue(expectedDocs > 0, "Expected source doc count to be positive");
+        return expectedDocs;
+    }
+
+    @SneakyThrows
+    private void runBackfillUntilOnePending(
```
gregschohn: As per above, you can delete this method

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,355 @@
...
+ *   - retry-exhaustion or failure signature in RFS output
+ */
+@Tag("isolatedTest")
+@Slf4j
+public class RfsOpenSearchCoordinatorOutageTest extends SourceTestBase {
+
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+
+    // Data config
+    private static final int SHARDS = 3;
```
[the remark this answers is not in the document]
jugal-chauhan: Changes have been made as discussed offline.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
jugal-chauhan: Done

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
jugal-chauhan: I'll add a minimal validation to ensure atleast one entry does not have a `completedAt` field

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
jugal-chauhan: Done

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,477 @@
...
+            var expectedDocs = getDocCountFromCluster(esSourceContainer.getUrl(), INDEX_NAME, true);
+            Assertions.assertEquals(TOTAL_DOCS, expectedDocs, "Expected source doc count to match configured TOTAL_DOCS");
+
+            // === ACTION : Start outage scheduler
+            var testStartNanos = System.nanoTime();
+            var scenarioTimeoutSeconds = TOTAL_TIMEOUT_SECONDS + (reEnableCoordinator ? reEnableAfterSeconds : 0);
+            var totalDeadlineNanos = testStartNanos + TimeUnit.SECONDS.toNanos(scenarioTimeoutSeconds);
+            var outageInjected = new AtomicBoolean(false);
+            var coordinatorReEnabled = new AtomicBoolean(false);
+            var outageThreadException = new AtomicReference<Throwable>();
```
gregschohn: Some of the assertion checks above aren't strictly necessary or don't need to be there.  They might be able to be earlier... However, there could be some value in verifying that the timelines match what you expected.  

If you change these to AtomicLongs, you can store the seconds since epoch & then do assertions that a<b; b<c; etc.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,477 @@
...
+                    targetProxy.getProxyUriAsString(),
+                    coordinatorProxy.getProxyUriAsString(),
+                    testContext);
+            } catch (Exception e) {
+                thrownException.set(e);
+                log.atInfo().setCause(e).setMessage("RFS threw exception (expected in outage scenario)").log();
+            }
+
+            // === ACTION : Wait for outage scheduler thread to finish after RFS exits
+            outageThread.join(TimeUnit.SECONDS.toMillis(reEnableCoordinator ? reEnableAfterSeconds + 10 : 5));
```
gregschohn: your test is done once RFS throws or doesn't, right?  What are you waiting for?
Can you check the state on the coordinator proxy when RFS stops running, do some assertions, and then be done with the test?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/RfsOpenSearchCoordinatorOutageTest.java
```diff
@@ -0,0 +1,477 @@
...
+
+        try (var workCoordinator = coordinatorFactory.get(
+            new CoordinateWorkHttpClient(coordinatorConnectionContext),
+            TOLERABLE_CLIENT_SERVER_CLOCK_DIFFERENCE_SECONDS,
+            UUID.randomUUID().toString(),
+            Clock.systemUTC(),
+            workItemRef::set
+        )) {
+            var progressCursor = new AtomicReference<WorkItemCursor>();
+            RfsMigrateDocuments.run(
```
gregschohn: I was expecting to see this use SourceTestBase.  Why not just tweak SourceTestBase to take the coordinator cluster connection info & leverage that like the other tests do?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SourceTestBase.java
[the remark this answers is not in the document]
jugal-chauhan: Added the AC

PR #2229, opened 2026-02-14
Fix 8 traffic replayer bugs: memory leak, context leaks, metrics inflation, and more

### LINKED ISSUE #1: Create pexpect core library
### Task
Set up core Python library for interacting w/ the operating system via command-line.  Practically, this means taking the code and unit tests from the Vingilot tool, upgrading them to Python3, and putting them in the new Migrations repo so its available for future usage.

### Acceptance Criteria
* pexpect primitives in Migrations Repo
* Unit tests for the primatives
* Upgraded to Python3 w/ type hints

AndreKurait:
## Summary

Deep code review of the traffic replayer uncovered 10 bugs across 8 source files. Each fix is accompanied by a targeted test that reproduces the original behavior and validates the correction. The commit history follows a test-driven approach: first a commit adding 8 reproduction tests, then one commit per fix.

All fixes are backward compatible. No API changes, no configuration changes, no behavioral changes visible to users except corrected metrics and eliminated resource leaks.

> **Overlap with PR #2232:** Fixes #1 (cache invalidation wrong key type) and #8 (ReadSegment adding empty bytes) are also addressed by [PR #2232](https://github.com/opensearch-project/opensearch-migrations/pull/2232), which takes a more comprehensive approach to #1 by keying on `(nodeId, connectionId)` via `TrafficChannelKeyFormatter` (also fixing `ChannelContextManager`) and adds higher-level regression tests (cross-node collision, segmented-read, stale session). When rebased on main after #2232 merges, the corresponding commits in this PR can be dropped.

## Impact Overview

| Fix | File(s) | Real-World Impact | Severity | Likelihood | Status |
|-----|---------|-------------------|----------|------------|--------|
| [Cache invalidation wrong key type](#1-clientconnectionpool-cache-invalidation-wrong-key-type) | `ClientConnectionPool.java` | Unbounded memory leak — closed connections never evicted from cache | **High** | Every replay session | ⚠️ Superseded by #2232 |
| [NPE in retry logic when source response is null](#2-npe-in-opensearchdefaultretry-when-source-response-is-null) | `OpenSearchDefaultRetry.java` | Crash — NPE kills the replay loop when source traffic has no captured response | **High** | Requires `_bulk` + target errors + missing source response |  |
| [CLOSED_PREMATURELY context leak](#3-closed_prematurely-context-leak) | `TrafficReplayerCore.java` | Context/span leak — traffic stream contexts never closed for prematurely closed streams | **Medium** | Every premature close (connection resets, incomplete HTTP) |  |
| [WriteSegment missing timestamp](#4-writesegment-missing-lastpackettimestamp) | `CapturedTrafficToHttpTransactionAccumulator.java` | NPE — null `lastPacketTimestamp` used in `Duration.between()` for segmented responses | **Medium** | Segmented responses reaching tuple logging |  |
| [Metrics double-counting on retries](#5-metrics-double-counting-on-retries) | `TrafficReplayerCore.java` | Inflated metrics — success/exception counts multiplied by retry attempts | **Low-Medium** | Every retry attempt |  |
| [Double-close of context spans](#6-double-close-of-context-spans) | `NettyPacketToHttpConsumer.java` | Inflated metrics — `targetTransactionCount` double-counted per request | **Low-Medium** | Every request (normal path) |  |
| [kafkaRecordsReadyToCommit never set](#7-kafkarecordsreadytocommit-never-set-to-true) | `TrackingKafkaConsumer.java` | Delayed Kafka commits — offsets committed on next keepAlive interval instead of immediately | **Low** | Every Kafka-based replay |  |
| [ReadSegment adding empty bytes](#8-readsegment-adding-empty-bytes) | `CapturedTrafficToHttpTransactionAccumulator.java` | No functional impact — zero-length byte arrays added to request packet list | **Negligible** | Every ReadSegment observation | ⚠️ Superseded by #2232 |
| [Response.toString() uses REQUEST type](#9-responsetostring-uses-request-type) | `HttpMessageAndTimestamp.java` | Cosmetic — response objects formatted as requests in logs | **Negligible** | Every response log |  |
| [Tight retry loop when referenceStartTime is in the past](#10-tight-retry-loop-when-referencestarttime-is-in-the-past) | `RequestSenderOrchestrator.java` | Event loop starvation — retries execute immediately in a tight loop, blocking all other work | **High** | Retries with `--speedup-factor` or `--lookahead-time-window` |  |

---

### 1. ClientConnectionPool cache invalidation wrong key type

> ⚠️ **Superseded by [PR #2232](https://github.com/opensearch-project/opensearch-migrations/pull/2232)**, which takes a more comprehensive approach: keys the cache on `(nodeId, connectionId)` via `TrafficChannelKeyFormatter` instead of just `connectionId`, also fixing `ChannelContextManager` to prevent cross-node connection ID collisions. Includes regression tests for cross-node collision and stale session scenarios.

**File:** `ClientConnectionPool.java`
**Impact: High — unbounded memory leak**

`closeConnection()` called `connectionId2ChannelCache.invalidate(connId)` where `connId` is a `String`, but the cache key is a `Key` object (composite of `connId` + `sessionNumber`). Guava `Cache.invalidate(Object)` does an equality check — `String` vs `Key` never matches, so the entry silently remains. The cache has **no max size and no expiration policy** (`CacheBuilder.newBuilder().build()`), so entries accumulate without bound for the lifetime of the replayer process.

```diff
- connectionId2ChannelCache.invalidate(connId);
+ connectionId2ChannelCache.invalidate(getKey(connId, sessionNumber));
```

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every single replay session. Every call to `closeConnection()` fails to evict the cache entry because `String.equals(Key)` is always `false`. This is not theoretical — it happens on every connection close.

**What happens:** The `connectionId2ChannelCache` grows monotonically. Each entry holds a `ConnectionReplaySession` which contains a Netty `EventLoop` reference and channel state. Over a long replay (millions of connections), this causes steadily increasing heap usage.

**Exception:** No exception is thrown. The leak is completely silent. The replayer continues to function but consumes increasing memory. Eventually, the JVM runs out of heap and throws `java.lang.OutOfMemoryError: Java heap space`, or GC pressure causes severe latency degradation before that point.

**Likelihood:** **Guaranteed** — affects every replay session that opens and closes connections (i.e., all of them).
</details>

---

### 2. NPE in OpenSearchDefaultRetry when source response is null

**File:** `OpenSearchDefaultRetry.java`
**Impact: High — crashes the replay loop**

`OpenSearchDefaultRetry.shouldRetry()` called `rrp.getResponseData().asByteBuf()` without null-checking `getResponseData()`. This causes a `NullPointerException` when:
1. A `_bulk` request is replayed to the target
2. The target returns HTTP 200 but with `"errors": true` in the bulk response body
3. The retry logic tries to compare against the **source** captured response
4. But the source traffic had **no response captured** (e.g., connection closed before response) → `getResponseData()` returns `null` → NPE

The parent class `DefaultRetry.shouldRetry()` already handles this correctly using `Optional.ofNullable(rrp.getResponseData())`. The `OpenSearchDefaultRetry` override for bulk requests skipped that null-safety.

```diff
- bulkResponseHadNoErrors(rrp.getResponseData().asByteBuf()) ?
-     RequestSenderOrchestrator.RetryDirective.RETRY :
-     RequestSenderOrchestrator.RetryDirective.DONE,
+ Optional.ofNullable(rrp.getResponseData())
+     .map(sourceResponse -> bulkResponseHadNoErrors(sourceResponse.asByteBuf()) ?
+         RequestSenderOrchestrator.RetryDirective.RETRY :
+         RequestSenderOrchestrator.RetryDirective.DONE)
+     .orElse(RequestSenderOrchestrator.RetryDirective.DONE),
```

When the source response is null, return `DONE` (don't retry) since there's no source response to compare against.

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** All four conditions must be met simultaneously:
1. Source traffic contains a `_bulk` request (common in OpenSearch/Elasticsearch workloads)
2. The target cluster returns HTTP 200 with `"errors": true` (e.g., mapping conflicts, version conflicts, or shard failures on some documents in the bulk)
3. The retry logic reconstructs the source request-response pair
4. The source capture has no response for this request (the capture proxy recorded the request but the connection was reset/closed before the response arrived)

Condition 4 is the key gate. This happens when the source cluster's connection to the client was interrupted mid-flight — e.g., network timeout, load balancer cut, or the source cluster crashed during the response. The capture proxy records the request but never sees a response.

**What happens:** `NullPointerException` at `rrp.getResponseData().asByteBuf()`. This exception propagates up through the retry logic and kills the replay processing for that traffic stream. Depending on error handling upstream, it may crash the entire replay loop or just drop that request.

**Exception:**
```
java.lang.NullPointerException: Cannot invoke "org.opensearch.migrations.replay.HttpMessageAndTimestamp.asByteBuf()" because the return value of "org.opensearch.migrations.replay.IRequestResponsePacketPair.getResponseData()" is null
    at org.opensearch.migrations.replay.http.retries.OpenSearchDefaultRetry.lambda$shouldRetry$2(OpenSearchDefaultRetry.java:125)
```

**Likelihood:** **Moderate** — requires the intersection of bulk requests, target-side errors, and missing source responses. More likely in large-scale migrations where network interruptions during capture are not uncommon.
</details>

---

### 3. CLOSED_PREMATURELY context leak

**File:** `TrafficReplayerCore.java`
**Impact: Medium — resource/span leak**

`commitTrafficStreams()` guarded both context closing AND commit sending behind `if (shouldCommit && ...)`. When `shouldCommit` is false (e.g., `CLOSED_PREMATURELY` status), the entire loop was skipped — contexts were never closed. This leaks traffic stream contexts (spans/traces) that are never cleaned up, affecting observability and potentially leaking memory over time.

**Fix:** Always close contexts; only skip the `commitTrafficStream()` call when `shouldCommit` is false.

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every time a traffic stream is closed prematurely. This happens when:
- The `CapturedTrafficToHttpTransactionAccumulator.close()` method is called while there are still live (incomplete) streams — e.g., a connection that sent a request but never received a complete response
- Connection resets mid-transaction (client or server side)
- Incomplete HTTP messages in the captured traffic (truncated capture)
- The replayer shuts down while streams are still in-flight

The accumulator's `close()` method iterates all `liveStreams` and fires them with `CLOSED_PREMATURELY` status.

**What happens:** The `ITrafficStreamKey.getTrafficStreamsContext()` (an OpenTelemetry span/scope) is never closed. This means:
- The span is never ended, so it never appears in traces (or appears as an orphaned span)
- Any resources held by the context (memory, file handles for span exporters) are leaked
- The `trafficStreamLifetimeCount` metric is never incremented for these streams, making monitoring inaccurate

**Exception:** No exception is thrown. The leak is silent. Over time, with many premature closes, this accumulates leaked spans and associated memory.

**Likelihood:** **Common** — premature closes happen regularly in real traffic captures. Any connection that was in-flight when the capture window ended, or any connection reset during capture, produces a `CLOSED_PREMATURELY` stream.
</details>

---

### 4. WriteSegment missing lastPacketTimestamp

**File:** `CapturedTrafficToHttpTransactionAccumulator.java`
**Impact: Medium — potential NPE for segmented responses**

The `WriteSegment` handler called `addSegment()` but never called `setLastPacketTimestamp()`. The `Write` handler (non-segmented) calls `addResponseData()` which sets the timestamp. This leaves `responseData.getLastPacketTimestamp()` as `null` for responses built from WriteSegments. Downstream, `ParsedHttpMessagesAsDicts` computes `Duration.between(requestData.getLastPacketTimestamp(), responseData.getLastPacketTimestamp())` — a null argument causes NPE.

```diff
  rrPair.responseData.addSegment(observation.getWriteSegment().getData().toByteArray());
+ rrPair.responseData.setLastPacketTimestamp(timestamp);
```

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** When the captured traffic contains segmented write observations (`WriteSegment` instead of `Write`). This happens when:
- The capture proxy splits large HTTP responses into multiple segments (common for large response bodies)
- The response data is delivered incrementally through the capture pipeline

The null timestamp only causes an NPE when the response reaches `ParsedHttpMessagesAsDicts.getSourceResponseOp()`, which computes `Duration.between(requestData.getLastPacketTimestamp(), responseData.getLastPacketTimestamp())`. This is called during tuple logging/output.

**What happens:** `Duration.between()` throws a `NullPointerException` because `responseData.getLastPacketTimestamp()` is `null`. This crashes the tuple handling for that specific request-response pair.

**Exception:**
```
java.lang.NullPointerException: Cannot invoke "java.time.Instant.getEpochSecond()" because "endExclusive" is null
    at java.base/java.time.Duration.between(Duration.java:...)
    at org.opensearch.migrations.replay.ParsedHttpMessagesAsDicts.getSourceResponseOp(ParsedHttpMessagesAsDicts.java:97)
```

**Likelihood:** **Moderate** — depends on whether the capture proxy produces `WriteSegment` observations. Large responses (e.g., scroll queries, large search results) are more likely to be segmented. The NPE only manifests when the tuple is logged/output, not during replay itself.
</details>

---

### 5. Metrics double-counting on retries

**File:** `TrafficReplayerCore.java`
**Impact: Low-Medium — inflated monitoring metrics**

`perResponseConsumer` incremented `exceptionRequestCount` or `successfulRequestCount` and was called for **every retry attempt**, not just the final outcome. A request that fails twice then succeeds would be counted as 2 exceptions + 1 success = 3 total, instead of just 1 success. This inflates metrics on monitoring dashboards but does not affect replay behavior.

**Fix:** Moved counting from `perResponseConsumer` (called per retry) to a new `countFinalOutcome()` method called from `handleCompletedTransaction` (called once per request). `perResponseConsumer` now only logs.

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every time a request is retried. The replayer retries requests when:
- The target returns a 5xx error (server error)
- The target returns a 429 (too many requests / throttled)
- For `_bulk` requests: the target returns 200 but with `"errors": true` and the source had no errors
- Connection failures to the target

With exponential backoff, a request might be retried 2-5 times before succeeding or giving up.

**What happens:** For a request that fails twice then succeeds:
- `exceptionRequestCount` is incremented 2 times (once per failed retry)
- `successfulRequestCount` is incremented 1 time (for the final success)
- Total counted: 3 responses for 1 actual request

This makes monitoring dashboards show inflated error rates and inflated total request counts. Operators may think the migration is experiencing more errors than it actually is.

**Exception:** No exception. Purely a metrics accuracy issue.

**Likelihood:** **Common** — retries happen regularly during migrations, especially when the target cluster is under load or when there are mapping differences causing bulk errors.
</details>

---

### 6. Double-close of context spans

**File:** `NettyPacketToHttpConsumer.java`
**Impact: Low-Medium — inflated `targetTransactionCount` metric**

In `finalizeRequest()`, when `channel != null`, `deactivateChannel()` closes both `getCurrentRequestSpan()` and `getParentContext()` in its `finally` block. The outer `finally` block then closed them again. The span close is idempotent (no crash), but the `targetTransactionCount` metric was incremented twice per request.

**Fix:** Restructured to if/else — `channel == null` closes spans directly; `channel != null` delegates to `deactivateChannel()` which handles closing in its own `finally`.

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every single request where `channel != null` — which is the normal, successful path. When a request is sent to the target, a Netty channel is established, and `finalizeRequest()` is called with a non-null channel. This is the overwhelmingly common case.

The only time `channel == null` is when the connection to the target failed before a channel was established (e.g., DNS resolution failure, connection refused).

**What happens:** `deactivateChannel()` closes the request span and parent context in its `finally` block. Then the outer code in `finalizeRequest()` closes them again. Since `close()` is idempotent on the span objects, no exception is thrown. However, `sendMeterEventsForEnd()` is called twice, which increments `targetTransactionCount` twice for every request.

This means the `targetTransactionCount` metric shows exactly 2x the actual number of requests sent to the target. Operators monitoring this metric would see double the expected throughput.

**Exception:** No exception. The double-close is silently absorbed by the idempotent close implementation.

**Likelihood:** **Guaranteed** — affects every successful request (the normal path).
</details>

---

### 7. kafkaRecordsReadyToCommit never set to true

**File:** `TrackingKafkaConsumer.java`
**Impact: Low — delayed Kafka offset commits**

The `kafkaRecordsReadyToCommit` flag was only set in `onPartitionsRevoked()`, never when `commitKafkaKey()` added entries to `nextSetOfCommitsMap`. This flag controls `getNextRequiredTouch()` — when `true`, it returns `Instant.now()` (immediate commit); when `false`, it returns `lastTouchTime + keepAliveInterval`. Since the flag was always `false` during normal operation, commits were delayed by the keepAlive interval instead of happening immediately. Commits still happened eventually, but the delay increases the reprocessing window after a crash.

```diff
  nextSetOfCommitsMap.put(k, v);
+ kafkaRecordsReadyToCommit.set(true);
```

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every Kafka-based replay session. When the replayer finishes processing a traffic stream and calls `commitKafkaKey()`, the offset is added to `nextSetOfCommitsMap` but `kafkaRecordsReadyToCommit` stays `false`.

**What happens:** `getNextRequiredTouch()` returns `lastTouchTime + keepAliveInterval` (default 30 seconds) instead of `Instant.now()`. This means the Kafka consumer's `touch()` method — which actually performs the commit — won't be called until the next keepAlive interval expires.

The practical impact: if the replayer crashes within that 30-second window, the offsets haven't been committed yet, so those traffic streams will be reprocessed on restart. This increases the "at-least-once" reprocessing window from near-zero to up to 30 seconds of work.

**Exception:** No exception. Commits still happen, just delayed.

**Likelihood:** **Guaranteed** for Kafka-based replays — the flag is never set during normal operation. The impact severity depends on crash frequency and the keepAlive interval.
</details>

---

### 8. ReadSegment adding empty bytes

> ⚠️ **Also addressed by [PR #2232](https://github.com/opensearch-project/opensearch-migrations/pull/2232)**, which removes the same erroneous line and adds a higher-level regression test (`testReadSegmentRequestAvoidsUnexpectedEmptyPacket`).

**File:** `CapturedTrafficToHttpTransactionAccumulator.java`
**Impact: Negligible — no functional impact**

The `ReadSegment` handler called `observation.getRead().getData().toByteArray()` — but `getRead()` on a ReadSegment observation returns a default empty `ReadObservation` (protobuf default). This added a zero-length byte array to the request packet list before the actual segment data. The empty arrays contribute nothing to the assembled HTTP request content, so there is no functional impact. Minor unnecessary memory overhead from extra list entries.

```diff
- rrPair.addRequestData(timestamp, observation.getRead().getData().toByteArray());
  rrPair.requestData.addSegment(observation.getReadSegment().getData().toByteArray());
```

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every time a `ReadSegment` observation is processed. This happens when the capture proxy splits large HTTP requests into multiple segments.

**What happens:** An extra zero-length `byte[]` is added to `packetBytes` for each `ReadSegment`. When the request is later assembled, these empty arrays contribute nothing (concatenating empty bytes is a no-op). The HTTP request sent to the target is identical with or without the bug.

**Exception:** No exception. No functional impact. The only effect is slightly more list entries in memory (one extra empty `byte[]` per segment), which is negligible.

**Likelihood:** **Guaranteed** for segmented reads, but the impact is truly negligible.
</details>

---

### 9. Response.toString() uses REQUEST type

**File:** `HttpMessageAndTimestamp.java`
**Impact: Negligible — cosmetic logging issue**

`Response.toString()` called `super.format(Optional.of(HttpByteBufFormatter.HttpMessageType.REQUEST))` instead of `RESPONSE`. This is a copy-paste error that causes response objects to be formatted as requests in log output and debugging. No functional impact on replay behavior.

```diff
- return super.format(Optional.of(HttpByteBufFormatter.HttpMessageType.REQUEST));
+ return super.format(Optional.of(HttpByteBufFormatter.HttpMessageType.RESPONSE));
```

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** Every time a `Response` object's `toString()` is called — typically during debug/trace logging, or when an exception message includes the response.

**What happens:** The HTTP response bytes (e.g., `HTTP/1.1 200 OK\r\n...`) are parsed as if they were an HTTP request. The Netty HTTP decoder tries to interpret `HTTP/1.1` as the HTTP method, `200` as the URI, and `OK` as the HTTP version. This produces garbled or confusing log output, but has zero impact on replay behavior since `toString()` is only used for logging/debugging.

**Exception:** No exception. The formatter handles parse failures gracefully (falls back to raw bytes). The output is just confusing to read.

**Likelihood:** **Guaranteed** when response logging is enabled, but the impact is purely cosmetic.
</details>

---

### 10. Tight retry loop when referenceStartTime is in the past

**File:** `RequestSenderOrchestrator.java`
**Impact: High — event loop starvation**

When requests fail and need retries, the retry time was computed as `referenceStartTime + retryDelay`. With `--speedup-factor` and `--lookahead-time-window`, `referenceStartTime` can be far in the past (it represents the original timestamp from the captured traffic, adjusted by the time shifter). If `referenceStartTime` is, say, 10 minutes in the past and `retryDelay` is 1 second, the computed retry time is still ~10 minutes in the past. `NettyFutureBinders` clamps negative delays to 0 via `Math.max(0, delay.toMillis())`, causing immediate execution — creating a tight retry loop that monopolizes the event loop thread.

```diff
- var newStartTime = referenceStartTime.plus(nextRetryDelay);
+ var computedStartTime = referenceStartTime.plus(nextRetryDelay);
+ // Ensure retry is not scheduled in the past to prevent tight retry loops
+ // that monopolize event loop threads when referenceStartTime is far in the past
+ var now = now();
+ var newStartTime = computedStartTime.isBefore(now)
+     ? now.plus(nextRetryDelay)
+     : computedStartTime;
```

<details>
<summary><b>🔍 Real-Life Scenario</b></summary>

**When it triggers:** When all of these conditions are met:
1. The replayer is using `--speedup-factor` > 1 (replaying faster than real-time) or `--lookahead-time-window` (scheduling requests ahead)
2. A request fails and needs to be retried
3. The `referenceStartTime` (derived from the original captured traffic timestamp) is in the past relative to wall clock time

This is common during catch-up scenarios: the replayer is processing a backlog of captured traffic and scheduling requests based on their original timestamps (adjusted by speedup factor). The original timestamps are in the past, so `referenceStartTime` is in the past.

**What happens:** The retry loop computes `newStartTime = referenceStartTime + retryDelay`. Since `referenceStartTime` is far in the past, `newStartTime` is also in the past. The scheduling delay `Duration.between(now(), newStartTime)` is negative, which `NettyFutureBinders` clamps to 0ms. The retry executes immediately. If it fails again, the next retry also executes immediately (the exponential backoff is applied to `referenceStartTime` which is still in the past). This creates a tight loop:

```
retry 1: referenceStartTime + 1s = still in past → execute immediately
retry 2: referenceStartTime + 2s = still in past → execute immediately  
retry 3: referenceStartTime + 4s = still in past → execute immediately
...
```

The event loop thread is monopolized by this tight retry loop, blocking all other scheduled work (other requests, Kafka consumer polling, keepAlive). The replayer appears to hang.

**Exception:** No exception is thrown. The replayer doesn't crash — it just becomes unresponsive because the event loop is stuck in a tight retry loop. Kafka consumer stops polling, which triggers a rebalance after the session timeout.

**Likelihood:** **High** — any replay with `--speedup-factor` > 1 that encounters retryable errors (5xx, 429, bulk errors) will hit this. This is a common production scenario during large-scale migrations.
</details>

---

## How to Run Tests

```bash
./gradlew :TrafficCapture:trafficReplayer:test --tests "org.opensearch.migrations.replay.bugfixes.*"
```

## Commit Structure

The first commit adds 8 reproduction tests (all initially asserting on buggy behavior). Each subsequent commit fixes one bug and updates its test to assert on correct behavior. This allows reviewers to verify each fix independently by checking out individual commits.

> **Note:** The diff includes changes from upstream merges (ArgNameConstants consolidation, EKS coordinator support). These are not part of this PR's bug fixes.

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/RequestSenderOrchestrator.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/RequestSenderOrchestrator.java
@@ -399,7 +399,13 @@ public class RequestSenderOrchestrator {
                     return TextTrackedFuture.failedFuture(t, () -> "failed future");
                 }
                 if (dtr.directive == RetryDirective.RETRY) {
-                    var newStartTime = referenceStartTime.plus(nextRetryDelay);
+                    var computedStartTime = referenceStartTime.plus(nextRetryDelay);
+                    // Ensure retry is not scheduled in the past to prevent tight retry loops
+                    // that monopolize event loop threads when referenceStartTime is far in the past
+                    var now = now();
+                    var newStartTime = computedStartTime.isBefore(now)
+                        ? now.plus(nextRetryDelay)
+                        : computedStartTime;
                     log.atInfo().setMessage("Making request scheduled at {}").addArgument(newStartTime).log();
                     var schedulingDelay = Duration.between(now(), newStartTime);
                     return NettyFutureBinders.bindNettyScheduleToCompletableFuture(


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
```diff
@@ -432,16 +432,13 @@ public TrackedFuture<String, AggregatedRawResponse> finalizeRequest() {
...
         }, () -> "Waiting for previous consumes to set the future")
         .map(f -> f.whenComplete((v, t) -> {
-            try {
-                if (channel == null) {
-                    log.atTrace().setMessage(
-                        "finalizeRequest().whenComplete has no channel present that needs to be to deactivated.").log();
-                } else {
-                    deactivateChannel();
-                }
-            } finally {
```
gregschohn: This seems dangerous - if there was a runtime exception, we'd leak the spans that we would have otherwise closed.  Unless it's the case that a Runtime exception in the above block would definitely cause the process to exit, we should keep the defensive finally.  (an exception to push along to the spans would be great, but is not critical)

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
```diff
@@ -101,7 +101,7 @@ public void closeConnection(IReplayContexts.IChannelKeyContext ctx, int sessionN
         var connectionReplaySession = connectionId2ChannelCache.getIfPresent(getKey(connId, sessionNumber));
         if (connectionReplaySession != null) {
             closeClientConnectionChannel(connectionReplaySession);
-            connectionId2ChannelCache.invalidate(connId);
+            connectionId2ChannelCache.invalidate(getKey(connId, sessionNumber));
```
gregschohn: This last line here also needs to be updated
```
    public void closeConnection(IReplayContexts.IChannelKeyContext ctx, int sessionNumber) {
        var connId = ctx.getConnectionId();
        log.atTrace().setMessage("closing connection for {}").addArgument(connId).log();
        var connectionReplaySession = connectionId2ChannelCache.getIfPresent(getKey(connId, sessionNumber));
        if (connectionReplaySession != null) {
            closeClientConnectionChannel(connectionReplaySession);
            connectionId2ChannelCache.invalidate(connId);
```

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayerCore.java
```diff
@@ -236,10 +253,12 @@ private void commitTrafficStreams(
 
         @SneakyThrows
         private void commitTrafficStreams(boolean shouldCommit, List<ITrafficStreamKey> trafficStreamKeysBeingHeld) {
-            if (shouldCommit && trafficStreamKeysBeingHeld != null) {
+            if (trafficStreamKeysBeingHeld != null) {
```
gregschohn: I'm struggling to get my head around shouldCommit.  I suspect that this might have been here for some corner cases (non-happy path) where some requests couldn't go through & I was being "conservative" to NOT commit them.  Of course, having just one message NOT commit, that can create significant downstream problems for future readers of the Kafka topic.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
AndreKurait: This is exactly what this is updating.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/NettyPacketToHttpConsumer.java
AndreKurait: Agree on the longer-term fix — spans should be idempotent on close so double-counting is structurally impossible. I'll open a Jira story under MIGRATIONS-2875 (Replayer Improvements) to track that.

For this PR, the current restructuring of `finalizeRequest()` into if/else eliminates the double-close path, so it's correct as-is. Happy to also pull the closes out of `deactivateChannel` and rename it if you feel strongly, but figured that's better as a follow-up alongside the idempotent-close work.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayerCore.java
```diff
@@ -236,10 +253,12 @@ private void commitTrafficStreams(
...
                 for (var tsk : trafficStreamKeysBeingHeld) {
                     tsk.getTrafficStreamsContext().close();
-                    trafficCaptureSource.commitTrafficStream(tsk);
+                    if (shouldCommit) {
+                        trafficCaptureSource.commitTrafficStream(tsk);
+                    }
```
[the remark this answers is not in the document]
AndreKurait: These are orthogonal operations — `close()` is resource cleanup (end span, release context), `commitTrafficStream()` is Kafka offset advancement. They need to be independently controlled.

The concrete case: during graceful shutdown, `CapturedTrafficToHttpTransactionAccumulator.close()` fires `CLOSED_PREMATURELY` for all in-flight streams. We must close those contexts (otherwise spans leak and metrics are wrong), but we must NOT commit (the source wasn't fully consumed — Kafka needs to re-deliver on restart).

If they're coupled, you're forced to choose between resource leaks and data loss. The decoupled design — always close, conditionally commit based on source completeness — avoids both.

PR #2234, opened 2026-02-16
Fix Kyverno reliability: namespace isolation, safe uninstall, and CI verification

AndreKurait:
## Motivation

Kyverno uses [admission webhooks](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/) with `failurePolicy=Fail` to enforce policies. This creates two reliability risks:

1. **Self-deadlock on uninstall**: `helm uninstall kyverno` triggers pre-delete hooks that need to create pods, but those pod creations are intercepted by Kyverno's own webhooks — which are being torn down. The uninstall hangs indefinitely. See [Kyverno webhook troubleshooting](https://kyverno.io/docs/troubleshooting/webhook/).

2. **Self-deadlock on crash recovery**: If Kyverno's admission controller crashes, Kubernetes tries to restart it, but the pod creation is blocked by the still-registered webhook (`failurePolicy=Fail`). If Kyverno runs in the same namespace as its webhook scope, it can never recover. The fix is [`excludeKyvernoNamespace`](https://kyverno.io/docs/installation/customization/#resource-filters) — Kyverno skips its own namespace so it can always restart.

### Namespace isolation (`values.yaml`, `installJob.yaml`)
- Move Kyverno from `ma` namespace to dedicated `kyverno-ma` namespace
- Set [`excludeKyvernoNamespace: true`](https://kyverno.io/docs/installation/customization/#resource-filters) — Kyverno's webhooks skip its own namespace, breaking the self-deadlock cycle
- Patch `kyverno-ma` namespace with an [ownerReference](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/) pointing to `ma` namespace — when `ma` is deleted, Kubernetes garbage-collects `kyverno-ma` automatically

### Uninstall reliability (`_uninstallHelper.tpl`, `uninstallJob.yaml`)
- **Two-phase uninstall strategy**: Pre-delete hook uninstalls all charts except fluent-bit (kept for logging) and kyverno (uninstalled last). Post-delete hook cleans up remaining charts (fluent-bit).
- **Hardcoded kyverno/fluent-bit ordering** — kyverno is always uninstalled last, fluent-bit is deferred to post-delete. No user-configurable flags.
- Delete webhook configurations *before* `helm uninstall kyverno` to prevent the chicken-and-egg deadlock
- Wrap `helm uninstall kyverno` with 120s timeout + force-cleanup fallback
- Clean up Kyverno CRDs after uninstall (cascade-deletes all ClusterPolicy/Policy resources)
- Discovery-based cleanup — kyverno cleanup runs regardless of `conditionalPackageInstalls.kyverno` value, so previously-installed kyverno is always cleaned up
- Comprehensive documentation in `_uninstallHelper.tpl` explaining the two-phase strategy, kyverno ordering rationale, and ownership model

### Setup reliability (`kyvernoMountLocalAwsCreds.yaml`, `kyvernoZeroResourceRequests.yaml`)
- Wait for CRD establishment + admission controller readiness before applying policies (prevents race condition)
- Fix namespace reference for admission controller wait to use `kyverno-ma` instead of release namespace
- Increase `backoffLimit` 5→10, add `set -e` for fail-fast

### CI / test cleanup (`k8s_service.py`, `test_runner.py`, `k8sLocalDeployment.groovy`)
- `cleanup_deployment()` now runs `helm uninstall` before namespace deletion, surfacing uninstall issues in CI
- Delete kyverno webhooks before namespace deletion to prevent API blocking
- Delete `kyverno-ma` namespace explicitly in cleanup (ownerReference cascade may not complete in time with force-delete)
- Jenkins cleanup stage deletes kyverno webhooks and `kyverno-ma` namespace before `ma`

## Coverage

| Issue | Severity | Status |
|-------|----------|--------|
| Kyverno self-deadlock on crash recovery | 🔴 Critical | ✅ Fixed (namespace isolation + `excludeKyvernoNamespace`) |
| Webhook deadlock on uninstall | 🔴 Critical | ✅ Fixed (pre-delete webhook cleanup) |
| No timeout on `helm uninstall kyverno` | 🟡 Medium | ✅ Fixed (120s timeout + force-cleanup) |
| Policy apply race condition | 🟡 Medium | ✅ Fixed (CRD + admission controller wait) |
| CRD/ClusterPolicy cleanup | 🟢 Low | ✅ Fixed (CRD deletion cascades to policies) |
| Kyverno namespace lifecycle | 🟢 Low | ✅ Fixed (ownerReference cascade deletion) |

### DIFF
--- a/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/uninstallJob.yaml
+++ b/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/uninstallJob.yaml
@@ -36,7 +36,7 @@ spec:
               {{ include "migration.helmUninstallFunctions" . | nindent 14 }}
               # Phase 1: Uninstall all charts except fluent-bit (kept for logging) and kyverno (uninstalled last).
               # See _uninstallHelper.tpl for the full uninstall strategy documentation.
-              uninstall_charts fluent-bit
+              uninstall_all_charts_except fluent-bit
 
 ---
 # Phase 2: Post-delete job to clean up charts that were intentionally kept alive during Phase 1.
@@ -80,7 +80,7 @@ spec:
               {{ include "migration.helmUninstallFunctions" . | nindent 14 }}
               # Phase 2: Uninstall remaining charts (fluent-bit etc.) that were kept active during Phase 1.
               # Charts already removed in Phase 1 are safely skipped (discovery-based, ownership-checked).
-              uninstall_charts
+              uninstall_all_charts_except
 
               # Final cleanup: delete the release namespace itself.
               # --wait=false returns immediately (marks for deletion); the namespace
--- a/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/helpers/_uninstallHelper.tpl
+++ b/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/helpers/_uninstallHelper.tpl
@@ -2,21 +2,21 @@
   Uninstall Helper — Helm Release Cleanup Functions
   ==================================================
 
-  This template defines `uninstall_charts`, the core function used by both the
-  pre-delete and post-delete uninstall jobs.
+  This template defines `uninstall_all_charts_except`, the core function used by
+  both the pre-delete and post-delete uninstall jobs.
 
   ## Uninstall Strategy (Two-Phase)
 
   Phase 1 — Pre-delete hook (uninstallJob.yaml, hook-weight: -10):
     Runs BEFORE Helm deletes the umbrella chart's own resources.
-    Calls: uninstall_charts fluent-bit
+    Calls: uninstall_all_charts_except fluent-bit
     Effect: Uninstalls all managed child charts IN PARALLEL, except:
       - fluent-bit: kept alive so logs are captured during teardown
       - kyverno: always uninstalled LAST and SYNCHRONOUSLY (see below)
 
   Phase 2 — Post-delete hook (uninstallJob.yaml, hook-weight: 10):
     Runs AFTER Helm deletes the umbrella chart's own resources.
-    Calls: uninstall_charts
+    Calls: uninstall_all_charts_except
     Effect: Uninstalls any remaining charts (primarily fluent-bit).
     The function re-discovers releases dynamically, so charts already removed
     in Phase 1 are simply not found — no double-uninstall occurs.
@@ -44,7 +44,7 @@
 */}}
 {{- define "migration.helmUninstallFunctions" -}}
 {{- $kyvernoChart := index .Values.charts "kyverno" -}}
-uninstall_charts() {
+uninstall_all_charts_except() {
   # Args: space-separated release names to skip in the parallel uninstall phase
   local RELEASES_TO_SKIP=("$@")
   # Kyverno is always handled separately at the end (see header comment)
--- a/libraries/testAutomation/testAutomation/test_runner.py
+++ b/libraries/testAutomation/testAutomation/test_runner.py
@@ -201,6 +201,9 @@ class TestRunner:
             helm_uninstall_error = e
         self.cleanup_clusters()
         self.k8s_service.delete_namespace()
+        # Assert both namespaces are actually gone
+        self.k8s_service.wait_for_namespace_deleted("kyverno-ma", timeout_seconds=60)
+        self.k8s_service.wait_for_namespace_deleted(self.k8s_service.namespace, timeout_seconds=120)
         if helm_uninstall_error:
             raise HelmCommandFailed(
                 f"Helm uninstall of '{MA_RELEASE_NAME}' release failed cleanly. "
--- a/vars/k8sLocalDeployment.groovy
+++ b/vars/k8sLocalDeployment.groovy
@@ -106,19 +106,16 @@ def call(Map config = [:]) {
                         script {
                             sh "kubectl config use-context minikube"
                             sh """
-                                # Delete kyverno webhooks first — they can block namespace deletion
+                                # Attempt clean helm uninstall first (triggers hook-based cleanup)
+                                helm uninstall ma -n ma || true
+
+                                # Delete kyverno webhooks in case helm uninstall didn't run cleanly
                                 kubectl delete mutatingwebhookconfigurations -l app.kubernetes.io/instance=kyverno --ignore-not-found || true
                                 kubectl delete validatingwebhookconfigurations -l app.kubernetes.io/instance=kyverno --ignore-not-found || true
 
-                                # Delete kyverno's separate namespace (if it exists from a previous run)
+                                # Force-delete namespaces if they still exist
                                 kubectl delete namespace kyverno-ma --ignore-not-found --grace-period=0 || true
-
-                                timeout 60 kubectl delete namespace ma --ignore-not-found || {
-                                    echo "Namespace delete timed out, forcing cleanup..."
-                                    kubectl get all -A
-                                    kubectl api-resources --verbs=list --namespaced -o name | xargs -n1 -I{} kubectl get {} -n ma -o name 2>/dev/null | xargs -r -n1 kubectl patch -n ma --type=merge -p '{"metadata":{"finalizers":null}}'
-                                    kubectl delete namespace ma --ignore-not-found --force --grace-period=0 || true
-                                }
+                                kubectl delete namespace ma --ignore-not-found --grace-period=0 --force || true
                             """
                         }
                     }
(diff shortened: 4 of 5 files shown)


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/installJob.yaml
```diff
@@ -77,22 +77,27 @@ spec:
...
-              echo "Kyverno installed and ready"
+              # Install charts marked installFirst synchronously before all others
+      {{- range $name, $chart := .Values.charts }}
+        {{- if and (index $.Values.conditionalPackageInstalls $name) $chart.installFirst }}
+              echo "Installing {{ $name }} first (must be ready before other charts)..."
+              {{- include "migration.installChart" (dict "name" $name "chart" $chart "root" $ "forceWait" true) | nindent 14 }}
+              echo "{{ $name }} installed and ready"
+
+              # Set ownerReference on namespace so it gets cascade-deleted with the release namespace
+              {{- if and $chart.namespaceOwnership (ne ($chart.namespace | default $.Release.Namespace) $.Release.Namespace) }}
```
gregschohn: Will everything work ok if kyverno is in the same namespace?  If so, we shouldn't need a separate namespace.  If yes, we should show an error rather than silently proceed.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/uninstallJob.yaml
```diff
@@ -34,8 +34,8 @@ spec:
               set -e
 
               {{ include "migration.helmUninstallFunctions" . | nindent 14 }}
-              # Skip fluent-bit (logging); kyverno handled at end of uninstall_charts
-              uninstall_charts fluent-bit
+              # Skip charts deferred to post-delete phase
+              uninstall_charts {{ range $name, $chart := .Values.charts }}{{ if eq (toString ($chart.uninstallPhase | default "pre-delete")) "post-delete" }}{{ $name }} {{ end }}{{ end }}
```
gregschohn: See below - it's a lot more readable and logical to just spell these out.  The reason that we want to hold on fluent bit isn't because of configuration but to maximize the likelihood of getting every log through to the final sinks.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/helpers/_uninstallHelper.tpl
```diff
@@ -2,8 +2,26 @@
...
+        break
+      fi
+    done
+    if [ "$ALREADY_SKIPPED" = "false" ]; then
+      UNINSTALL_LAST+=("{{ $name }}")
+      RELEASES_TO_SKIP+=("{{ $name }}")
+    fi
+  }
+  {{- end }}
+{{- end }}
```
gregschohn: This is too obtuse for me to understand

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -621,9 +622,17 @@ charts:
         enabled: false
 
   kyverno:
+    installFirst: true           # Install before all other charts (webhooks must be ready)
+    uninstallLast: true          # Uninstall after all other charts, synchronously
```
gregschohn: same concern - it seems like this would be catastrophic if overridden incorrectly.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -651,4 +660,4 @@ charts:
           requests: null
           limits: null
       config:
-        excludeKyvernoNamespace: false
+        excludeKyvernoNamespace: true
```
gregschohn: Where is this masking filter applied?  I would assume that everything going through the namespace would be getting passed through kyverno - and kyverno would be applying some kind of predicate on each resource. Is this a 'global' predicate to glob onto all of the other predicates?

gregschohn (commented):
I have some comments.  I need to reserve some others because I can't get my head around the uninstall logic.  Please help document that part of the code better.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/helpers/_uninstallHelper.tpl
AndreKurait: Yeah, that was over-engineered. Replaced the whole generic `uninstallLast` loop with a simple hardcoded `RELEASES_TO_SKIP+=("kyverno")` — back to the original approach but with the new cleanup improvements (webhook deletion, timeout with fallback, CRD cleanup).

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
AndreKurait: Agreed — removed both flags. Install-first and uninstall-last for kyverno are now hardcoded in the templates. Kept `preUninstallCleanup`, `postUninstallCleanup`, and `uninstallTimeout` in values since those are operational parameters rather than ordering logic.

AndreKurait:
Added comprehensive documentation to the uninstall logic in `fbc3a4b75`:

**`_uninstallHelper.tpl`** now has a header comment block explaining:
- The two-phase uninstall strategy (pre-delete vs post-delete hooks)
- Why kyverno must be uninstalled last (webhook blocking)
- The kyverno cleanup sequence (webhooks → helm uninstall → force-clean → CRDs)
- The ownership model (`global.managedBy` tagging)

Also fixed a bug: the kyverno cleanup section was guarded by `conditionalPackageInstalls.kyverno`, which would skip cleanup if kyverno was previously installed then disabled. Removed the guard — the cleanup is discovery-based so it's a safe no-op when kyverno isn't present.

Cleaned up the empty re-trigger commit from the branch too.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
gregschohn: For my own knowledge, here's what Claude tells me kyverno's global resource exclusion list looks like...
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: kyverno
  namespace: kyverno
data:
  excludeGroupRole: "system:serviceaccounts:kube-system,system:nodes,system:kube-scheduler"
  resourceFilters: >
    [Event,*,*]
    [*,kube-system,*]
    [*,kube-public,*]
    [*,kube-node-lease,*]
    [*,kyverno,*]
    [Node,*,*]
    [Node/*,*,*]
    [APIService,*,*]
    [APIService/*,*,*]
    [TokenReview,*,*]
    [SubjectAccessReview,*,*]
    [SelfSubjectAccessReview,*,*]
    [Binding,*,*]
    [Pod/binding,*,*]
    [ReplicaSet,*,*]
    [ReplicaSet/*,*,*]
    [AdmissionReport,*,*]
    [AdmissionReport/*,*,*]
    [ClusterAdmissionReport,*,*]
    [ClusterAdmissionReport/*,*,*]
    [BackgroundScanReport,*,*]
    [BackgroundScanReport/*,*,*]
    [ClusterBackgroundScanReport,*,*]
    [ClusterBackgroundScanReport/*,*,*]
    ```

File: vars/k8sLocalDeployment.groovy [hunk 6]
```diff
@@ -106,6 +106,13 @@ def call(Map config = [:]) {
                         script {
                             sh "kubectl config use-context minikube"
                             sh """
+                                # Delete kyverno webhooks first — they can block namespace deletion
+                                kubectl delete mutatingwebhookconfigurations -l app.kubernetes.io/instance=kyverno --ignore-not-found || true
+                                kubectl delete validatingwebhookconfigurations -l app.kubernetes.io/instance=kyverno --ignore-not-found || true
+
+                                # Delete kyverno's separate namespace (if it exists from a previous run)
+                                kubectl delete namespace kyverno-ma --ignore-not-found --grace-period=0 || true
```
gregschohn: Is this the right order?  Is this the order that helm would do?  If it is, then this can be ok for now.  I'd like to see this do a helm uninstall instead.

libraries/testAutomation/testAutomation/test_runner.py
```diff
@@ -193,8 +193,19 @@ def cleanup_clusters(self) -> None:
             logger.warning(f"Failed to cleanup labeled Kubernetes resources: {e}")
 
     def cleanup_deployment(self) -> None:
+        helm_uninstall_error = None
+        try:
+            self.k8s_service.helm_uninstall(release_name=MA_RELEASE_NAME)
```
gregschohn: can we assert that the namespace(s) are both gone?

PR #2261, opened 2026-02-18
Add 15s default timeout to console clusters curl command

AndreKurait:
## Description

The `console clusters curl` command previously had no timeout, causing requests to hang indefinitely when a cluster is unreachable or slow to respond. This adds a `--timeout` CLI option that defaults to **15 seconds**, and provides a user-friendly error message when a timeout occurs.

Users who need longer timeouts for specific requests can override with `--timeout <seconds>`, e.g.:

```bash
console clusters curl target_cluster '/_snapshot/repo/_status' --timeout 60
```

### Before
- No timeout: requests hang forever if a cluster is unreachable
- Timeout errors (if manually set) showed raw `HTTPSConnectionPool` internals

### After
- 15s default timeout
- Clean error message on timeout:
  ```
  Error: Request timed out after 15s. Use --timeout to specify a longer timeout.
  ```

## Changes

- **`cli.py`** — Added `--timeout` click option (default 15s), passed through to `call_api`
- **`middleware/clusters.py`** — Added `requests.exceptions.Timeout` handler with user-friendly message
- **`tests/test_cli.py`** — Updated 4 existing test assertions for new default; added CLI-level timeout error test
- **`tests/middleware/test_clusters.py`** — Added middleware-level timeout error test

## Testing

- Unit tests verify the 15s default is passed through correctly
- Unit tests verify timeout errors produce clean, actionable messages (no stacktraces, no `HTTPSConnectionPool` internals)
- All tests pass locally (Python 3.11) and in CI

### DIFF
--- a/migrationConsole/lib/console_link/console_link/middleware/clusters.py
+++ b/migrationConsole/lib/console_link/console_link/middleware/clusters.py
@@ -1,3 +1,5 @@
+import requests.exceptions
+
 from console_link.models.cluster import Cluster, HttpMethod
 from console_link.models.snapshot import Snapshot
 from dataclasses import dataclass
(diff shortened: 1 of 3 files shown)


migrationConsole/lib/console_link/console_link/cli.py
```diff
@@ -247,10 +247,12 @@ def parse_headers(header: str) -> Dict:
 @click.option('-H', '--header', multiple=True, help='Pass custom header(s) to the server.')
 @click.option('-d', '--data', help='Send specified data in a POST request.')
 @click.option('--json', 'json_data', help='Send data as JSON.')
+@click.option('--timeout', type=int, default=15, show_default=True,
```
gregschohn: My only question here is whether this should have a default of not?  A lot of cluster calls can take much longer than 15s. OTOH - 15s isn't too long, you'll get the message, then you can type 9999999.  As a user, having no timeout (default<0... which is coerced to None) makes more sense.  

The rest of the PR looks good though - we definitely should be exposing this flag

PR #2298, opened 2026-02-25
Disable configuration cache globally via gradle.properties

AndreKurait:
Sets `org.gradle.configuration-cache=false` in `gradle.properties` to disable configuration cache globally, and removes all `--no-configuration-cache` flags that were previously needed as workarounds.

### Changes
- `gradle.properties`: `org.gradle.configuration-cache=true` → `false`
- Removed `--no-configuration-cache` from:
  - `vars/k8sLocalDeployment.groovy`
  - `vars/eksBYOSIntegPipeline.groovy`
  - `.github/workflows/release-drafter.yml`
  - `deployment/k8s/charts/components/buildImages/templates/build-images-job.yaml`
  - `deployment/k8s/aws/aws-bootstrap.sh`

Signed-off-by: Andre Kurait <andrekurait@gmail.com>

### DIFF
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -937,7 +937,7 @@ if [[ "$build_images" == "true" ]]; then
     | docker login --username AWS --password-stdin "$ecr_domain" \
     || { echo "ECR login failed"; exit 1; }
 
-  "$base_dir/gradlew" -p "$base_dir" :buildImages:${BUILD_TARGET} -PregistryEndpoint="$MIGRATIONS_ECR_REGISTRY" -x test --no-configuration-cache || exit
+  "$base_dir/gradlew" -p "$base_dir" :buildImages:${BUILD_TARGET} -PregistryEndpoint="$MIGRATIONS_ECR_REGISTRY" -x test || exit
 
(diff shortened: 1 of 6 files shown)


deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -945,7 +937,7 @@ if [[ "$build_images" == "true" ]]; then
     | docker login --username AWS --password-stdin "$ecr_domain" \
     || { echo "ECR login failed"; exit 1; }
 
-  "$base_dir/gradlew" -p "$base_dir" :buildImages:${BUILD_TARGET} -PregistryEndpoint="$MIGRATIONS_ECR_REGISTRY" -x test || exit
+  "$base_dir/gradlew" -p "$base_dir" :buildImages:${BUILD_TARGET} -PregistryEndpoint="$MIGRATIONS_ECR_REGISTRY" -x test --no-configuration-cache || exit
```
gregschohn: Can we set this as a project wide setting?  If we're saying that we can't properly build an image with the configuration cache, we shouldn't require a user to run gradle with a flag to disable it - it should default to off

AndreKurait: updated, thanks

PR #2374, opened 2026-03-09
Update Jenkins workflow concurrency to per-job per-PR

AndreKurait:
## Description

The current top-level concurrency group limits the entire Jenkins workflow to one run per branch, which means:
- On main: pushes don't cancel each other (good), but they're serialized into one concurrent run
- On PRs: a new push cancels **all** running Jenkins tests, not just the same test

### Changes

Removed the top-level concurrency group and added per-job concurrency:

| Scenario | Behavior |
|----------|----------|
| **Main pushes** | Unlimited concurrency — each run gets a unique group via `github.run_id`, so runs never block or cancel each other |
| **PR pushes** | 1 concurrent run per test per PR — grouped by `PR number + job name`, with `cancel-in-progress: true` so a new push to the same PR cancels only the same test's previous run |

This means on a PR, pushing a new commit will cancel the old `full-es68-e2e-aws-test` run for that PR but won't affect the `elasticsearch-5x-k8s-local-test` or `eks-integ-test` runs.

### DIFF
--- a/.github/workflows/jenkins_tests.yml
+++ b/.github/workflows/jenkins_tests.yml
@@ -7,10 +7,6 @@ on:
   pull_request_target:
     types: [opened, synchronize, reopened, labeled]
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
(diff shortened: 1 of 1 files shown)


AndreKurait:
> Per job concurrency means GHA cancels the poller, not the Jenkins job. Do we want to address this in the same PR ?

That needs to be fixed in the Action instead

PR #2390, opened 2026-03-12
Optimize k8s local integration test execution time

AndreKurait:
## Description

Reduces k8s local integration test overhead by eliminating redundant helm install/uninstall cycles between version combinations and tightening various polling intervals.

### Changes

**Lightweight cleanup between version combinations (~10 min saved)**
- Reuse the MA helm installation (argo, kyverno, etcd, migration-console) across version combos instead of full namespace delete + reinstall per version
- Only delete Argo Workflow instances and test clusters between versions, preserving WorkflowTemplates
- Eliminates ~2.5 min of helm install + pod startup overhead per version transition (×4 transitions = ~10 min)

**Faster polling and probes**
- Reduce migration workflow monitor polling from 60s to 5s with exponential backoff (factor 2, cap 15s)
- Reduce pod health check polling from 3s to 2s
- Reduce test cluster readiness probe period and successThreshold
- Reduce namespace deletion polling from 3s to 1s
- Reduce kyverno uninstall timeout from 120s to 30s

**Disable unnecessary components for local testing**
- Skip kube-prometheus-stack, otel-collector, jaeger, fluent-bit chart installations

**Bug fixes found during testing**
- Fix Argo retry backoff factor: use integer `2` instead of float `1.5` (Argo requires int)
- Fix `create_document` to expect 201 status code
- Fix duplicate cluster readiness wait in workflow template
- Increase OSB backfill test timeout from 600s to 1000s to match base class default

### Testing

Validated on CI builds 10-15 across all 3 target versions (OS_1.3, OS_2.19, OS_3.1). All non-OSB tests pass. OSB (OpenSearchBenchmarkBackfill) tests have intermittent timeouts on resource-constrained minikube, addressed by increasing timeout to 1000s.

### Performance

Per-version transition overhead: **~2.5 min → ~0s** (lightweight cleanup vs full reinstall)
Total setup overhead across 5 versions: **~13 min → ~3 min**

### DIFF
--- a/migrationConsole/lib/integ_test/testWorkflows/clusterWorkflows.yaml
+++ b/migrationConsole/lib/integ_test/testWorkflows/clusterWorkflows.yaml
@@ -430,7 +430,6 @@ spec:
           - |
             # Add Elasticsearch Helm repository
             helm repo add elastic https://helm.elastic.co
-            helm repo update
             
             # Create Elasticsearch configuration for ES 5.6.16
             cat > /tmp/elasticsearch.yml << EOF
@@ -531,7 +530,6 @@ spec:
           - |
             # Add Elasticsearch Helm repository
             helm repo add elastic https://helm.elastic.co
-            helm repo update
             
             # Create Elasticsearch configuration for ES 6.8.23 OSS (minimal config)
             cat > /tmp/elasticsearch.yml << EOF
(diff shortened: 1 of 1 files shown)


AndreKurait:
Recreating PR from fork

PR #2460, opened 2026-03-19
Add Argo artifact output support for large workflow outputs

AndreKurait:
## Problem

Workflow output parameters (e.g. `statusOutput`, `monitorResult`) are stored inline in the Workflow CR in etcd via `outputs.parameters[].valueFrom.path`. This causes two problems:

1. **etcd size limit** — Large outputs (metadata evaluation results, monitor status) can approach ~2MB. etcd has a 1MB object limit, causing workflow failures on real-world migrations with many indices.
2. **Node eviction** — If an EKS Auto Mode node is reclaimed, the pod disappears before the Argo wait container reads the output file from the container filesystem.

## Solution

Use Argo's `outputs.artifacts` instead of `outputs.parameters` for large outputs. Artifacts are uploaded to S3; only a small S3 key reference is stored in the Workflow CR. The Python CLI retrieves artifact content transparently via the Argo Server artifact API.

## Customer UX

**No changes to the user-facing CLI commands.** The artifact storage is transparent — users interact with the same commands as before:

```bash
workflow submit    # unchanged
workflow status    # artifact outputs fetched lazily from S3 via Argo API
workflow approve   # unchanged
workflow output    # large outputs stream from S3 instead of inline in etcd
```

### What changes for the user

| Aspect | Before | After |
|--------|--------|-------|
| `workflow status` output | Identical | Identical — artifact content fetched transparently |
| `workflow output` | Reads inline parameters from Workflow CR | Reads from S3 via Argo artifact API (same display) |
| Large migrations (many indices) | ❌ Fails with etcd size limit | ✅ Works — only S3 key stored in CR |
| Node eviction during output save | ❌ Output lost, step marked failed | ✅ Output persisted to S3 before pod exits |
| Workflow logs | Not archived | ✅ Archived to S3 (`archiveLogs: true`) |

### Verified on EKS with managed AOS clusters

Tested end-to-end on account `969644818851` (us-east-1):
- Source: Elasticsearch 7.10 → Target: OpenSearch 2.19
- Workflow ran through: snapshot creation → metadata evaluate → suspended at approval
- Metadata evaluation artifact successfully stored in S3 and retrieved via `workflow status`

### Bug found and fixed during testing

The Argo artifact repository Helm configuration passed an empty string for `endpoint` on EKS. Fixed by setting `endpoint` dynamically via `--set` only when LocalStack is configured; EKS uses `useSDKCreds: true` with no endpoint.

### Layer 1: Helm — Argo Artifact Repository
- Configure `artifactRepository.s3` in Argo Helm chart values with bucket, key format, and log archiving
- `values.yaml`: secure defaults (`useStaticCredentials: false`, `insecure: false`) — no credentials in base config
- `valuesForLocalK8s.yaml`: static credentials via `argo-artifact-s3-creds` Secret, `insecure: true`
- `valuesEks.yaml`: only `useSDKCreds: true` (all other settings inherit from secure base)
- Dynamically set `bucket` and `endpoint` via `--set` in the install job
- Create `argo-artifact-s3-creds` Secret only for LocalStack environments

### Layer 2: TypeScript Builders — Artifact Output Support
- Add `OutputArtifactDef` / `OutputArtifactsRecord` types to `parameterSchemas.ts` with detailed JSDoc
- Add `addArtifactOutput()` to `ContainerBuilder` with compile-time duplicate-name blocking via `UniqueNameConstraintAtDeclaration` and `ArtifactScope` type parameter
- Add `UniqueNameConstraintAtDeclaration` to `addPathOutput()` as well
- Render `outputs.artifacts` in `argoResourceRenderer.ts`

### Layer 3: Migration Templates
- Convert `statusOutput` to artifact output in `migrationConsole.ts`
- Convert `monitorResult` to artifact output in `testMigrationWithWorkflowCli.ts`
- `overriddenPhase` stays as a parameter (tiny string used in Argo expression plumbing)

### Layer 4: Python CLI — Lazy Artifact Resolution
- `WorkflowService.create_artifact_resolver()` returns a closure with connection context baked in
- `display_workflow_tree()` accepts the resolver and invokes it lazily — only when a rendered node has an `ArtifactRef`
- `get_step_status_output()` returns `str | ArtifactRef | None` — parameters checked first, artifacts as fallback
- `WorkflowService.get_artifact_content()` fetches via Argo Server REST API
- For the ~90% of nodes without artifacts, no API call happens

### Formatting only (no semantic changes)
- `userSchemas.ts`: consistent 4-space indent, missing semicolons
- `argoSchemas.ts`: import deduplication, indent fixes

## Testing
- E2e integ tests for artifact outputs (`containerOutputs.integ.test.ts`) — submits workflows with artifact outputs to a real Argo cluster
- E2e integ test for WaitForExistingResource retry (`rendererToArgo.integ.test.ts`)
- 10 Python unit tests for artifact fetching, resolution, and display (`test_artifact_output.py`)
- All existing tests pass (112 TS unit tests, Python tests)
- End-to-end verified on EKS with managed AOS domains (ES 7.10 → OS 2.19)


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/installJob.yaml
```diff
@@ -226,6 +226,15 @@ spec:
           {{- end }}
                   --timeout {{ default "300" $chart.timeout }}s \
                   --set global.managedBy="{{ $.Release.Name }}" \
+          {{- if eq $name "argo-workflows" }}
+                  --set artifactRepository.s3.bucket="{{ printf "migrations-default-%s-%s-%s" (toString $.Values.aws.account) $.Values.stageName $.Values.aws.region }}" \
+                  --set artifactRepository.s3.region="{{ $.Values.aws.region }}" \
+            {{- if $.Values.defaultBucketConfiguration.useLocalStack }}
+                  --set artifactRepository.s3.endpoint="{{ $.Values.defaultBucketConfiguration.endpoint }}.{{ $.Release.Namespace }}.svc.cluster.local:4566" \
+            {{- else }}
+                  --set artifactRepository.s3.endpoint="s3.{{ $.Values.aws.region }}.amazonaws.com" \
```
gregschohn: do we have another variable for s3 endpoint already in our helm values?  I think not, so this is fine, but we should do a sweep to verify.  I recall discussion about for what other cases a user might want to set the endpoint here.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/argoArtifactRepository.yaml
```diff
@@ -0,0 +1,14 @@
+{{- if and (get .Values.conditionalPackageInstalls "argo-workflows") .Values.defaultBucketConfiguration.useLocalStack }}
```
gregschohn: Would it make sense to make this ONLY conditional for localstack rather than both?  We'd use the same creds for every app, right?

migrationConsole/lib/console_link/console_link/workflow/tree_utils.py
```diff
@@ -442,6 +469,12 @@ def add_nodes_to_tree(nodes, parent_tree):
             # Get statusOutput for all nodes that might have it
             status_output = get_step_status_output(workflow_data, node['id']) if workflow_data else ""
 
+            # Resolve artifact references
+            if isinstance(status_output, ArtifactRef) and artifact_resolver:
+                status_output = artifact_resolver(status_output) or ""
+            elif isinstance(status_output, ArtifactRef):
+                status_output = ""
```
gregschohn: why do we need to call an artifact resolver to render the workflow tree?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: Just checking, were there any material changes in this file?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/installJob.yaml
```diff
@@ -226,6 +226,15 @@ spec:
           {{- end }}
                   --timeout {{ default "300" $chart.timeout }}s \
                   --set global.managedBy="{{ $.Release.Name }}" \
+          {{- if eq $name "argo-workflows" }}
+                  --set artifactRepository.s3.bucket="{{ printf "migrations-default-%s-%s-%s" (toString $.Values.aws.account) $.Values.stageName $.Values.aws.region }}" \
```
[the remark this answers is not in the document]
gregschohn: What happens if this bucket is already taken?
Do we delete this bucket when we're deleting helm?

PR #2462, opened 2026-03-19
Add CRD-based workflow lifecycle: reset, resubmit, approval gates, and Kubernetes-native teardown

AndreKurait:
## Description

Introduces CRD-based lifecycle management for every migration component. Each resource (Kafka, proxy, snapshot, backfill, replayer) gets a custom resource that tracks its state and owns its Kubernetes objects. Teardown is deletion-based — deleting a CRD cascades through owned resources via Kubernetes garbage collection. This enables surgical reset of individual components and iterative resubmit without downtime.

📄 **Design doc**: [`docs/WorkflowCrdDesign.md`](docs/WorkflowCrdDesign.md) — covers the ownership graph, proxy lifecycle state transitions (Capturing → Non-Capture → Deleted), ordered teardown guarantees, and implementation details.

### Core Paradigm Shift

**Before:** Resources were owned by the Argo workflow (`setOwnerReference: true`). Deleting the workflow killed everything. Approval gates were Argo suspend steps. There was no way to reset a single component.

**After:** Resources are owned by migration CRDs (`setOwnerReference: false` on workflow). Each CRD tracks lifecycle state (Created → Ready → Teardown). The workflow watches for CRD deletion to know when to exit. Approval gates are CRD status patches. You can reset any component independently.

### New CLI Commands

| Command | What it does |
|---------|-------------|
| `workflow reset` | List all migration resources, their status, and dependencies |
| `workflow reset <name>` | Delete a specific resource (blocked if dependents exist) |
| `workflow reset --cascade <name>` | Delete a resource and all its dependents |
| `workflow reset --all` | Delete all migration CRDs + stop/delete workflows |
| `workflow resubmit` | Stop current workflow + delete + submit new one (preserves existing resources) |
| `workflow approve <pattern>` | Approve pending migration gates via CRD status patching (supports globs) |
| `workflow proxy disable-capture [name]` | Switch proxy to non-capture pass-through mode (resubmits without modifying saved config) |
| `workflow proxy enable-capture [name]` | Restore traffic capture on a proxy |

### Dynamic Dependency Graph

Each CRD instance declares its dependencies in `spec.dependsOn[]`. The CLI reads these from live CRDs — **no hardcoded dependency order in Python**.

```yaml
apiVersion: migrations.opensearch.org/v1alpha1
kind: SnapshotMigration
metadata:
  name: source-target-snap1
spec:
  dependsOn: ["source-snap1"]   # depends on this DataSnapshot
```

With 3 parallel migrations, resetting `snap-a` only affects `mig-a`, not `mig-b`:
```
kafka1
├── proxy1 (depends on: kafka1)
│   ├── snap-a (depends on: proxy1)
│   │   └── mig-a (depends on: snap-a)
│   └── snap-b (depends on: proxy1)
│       └── mig-b (depends on: snap-b)
└── replay1 (depends on: proxy1)
```

### Concurrent DAG Deletion

`--cascade` deletes resources as a concurrent DAG — each resource starts deleting the moment all its children are gone. Independent branches proceed without waiting for each other:

```mermaid
graph BT
    miga[mig-a] -->|depends on| snapa[snap-a]
    migb[mig-b] -->|depends on| snapb[snap-b]
    snapa -->|depends on| proxy1
    snapb -->|depends on| proxy1
    replay1 -->|depends on| proxy1

    style miga fill:#e74c3c,color:#fff
    style migb fill:#e74c3c,color:#fff
    style replay1 fill:#e74c3c,color:#fff
    style snapa fill:#f39c12,color:#fff
    style snapb fill:#f39c12,color:#fff
    style proxy1 fill:#2ecc71,color:#fff
```

Deletion proceeds bottom-up, with maximum concurrency per wave:
| Wave | Deletes | Why |
|------|---------|-----|
| 🔴 1 | `mig-a`, `mig-b`, `replay1` | Leaves — no children, all start immediately |
| 🟠 2 | `snap-a` (after `mig-a`), `snap-b` (after `mig-b`) | Each starts as soon as its child is gone — `snap-a` doesn't wait for `mig-b` |
| 🟢 3 | `proxy1` | Starts only after all children (`snap-a`, `snap-b`, `replay1`) are gone |

### Dependency-Aware Deletion

Targeted reset is blocked when live dependents exist:
```
$ workflow reset snap-a
Cannot delete — dependent resources exist:
  Snapshot Migration: mig-a

Use --cascade to delete them too.
```

### Iterative Development Flow

```bash
workflow submit                        # creates everything
# backfill needs config change...
workflow reset snapshotmigration/*     # kill just backfill
workflow configure edit                # tweak config
workflow resubmit                      # proxy + Kafka stay alive, backfill restarts
```

### Full Reset Flow

```bash
workflow reset --all                   # deletes all CRDs + workflows
workflow submit                        # fresh start
```

### Teardown Model

```mermaid
sequenceDiagram
    participant CLI as CLI (workflow reset)
    participant CRD as Migration CRD
    participant K8s as Kubernetes GC
    participant WF as Argo Workflow

    CLI->>CRD: kubectl delete (foreground)
    K8s->>K8s: Cascade-delete owned Deployments/Pods
    K8s->>CRD: All dependents gone → CRD deleted
    WF->>WF: waitForCrdDeletion resolves
    CLI->>WF: patch spec.shutdown=Stop + delete
```

### Ownership Graph

CRDs are **not** owned by the workflow (`setOwnerReference: false`), so deleting the workflow preserves existing resources. This enables iterative resubmit.

Ownership is chained for ordered shutdown: **RFS pods die before Coordinator pods** — zero spurious error logs.

### Skip-if-Ready on Resubmit

When a new workflow starts and finds existing CRDs in `Ready` state (from a previous run), it skips deployment and goes straight to waiting. Proxy and Kafka are reused without redeployment.

### Convention-Based Reset

`workflow reset --all` is workflow-name-agnostic. It:
1. Deletes all migration CRDs (foreground cascading handles ordering via ownerReferences)
2. Discovers and stops/deletes all Argo workflows in the namespace via k8s API

No `--argo-server`, `--workflow-name`, or `--token` flags needed.

### CRDs (`migrations.opensearch.org/v1alpha1`)

| CRD | Tracks | Owns | dependsOn |
|-----|--------|------|-----------|
| `KafkaCluster` | Kafka lifecycle | Strimzi Kafka, NodePool, Topics | (none — root) |
| `CapturedTraffic` | Capture proxy | Proxy Deployment + Service | kafka cluster |
| `DataSnapshot` | Snapshot state | (none, immutable) | proxy names |
| `SnapshotMigration` | Backfill + metadata | Coordinator StatefulSet, Secret, Service → RFS Deployment | data snapshot |
| `TrafficReplay` | Replayer | Replayer Deployment | proxy name |
| `ApprovalGate` | Human checkpoints | (none) | (none) |

### New Workflow Templates

| Template | Purpose |
|----------|---------|
| `createKafkaCluster`, `createSnapshotMigration`, `createCapturedTraffic`, `createTrafficReplay` | Create CRDs with initial phase |
| `getResourceUid` | Read CRD UID + phase for ownerReference injection and skip-if-Ready |
| `waitForCrdDeletion` | `kubectl wait --for=delete` — blocks until CRD gone (1yr timeout, retry-on-failure) |
| `deleteCrd` | Self-delete CRD on natural completion |
| `patchKafkaClusterReady`, `patchCapturedTrafficReady`, `patchTrafficReplayReady`, `patchSnapshotMigrationReady` | Signal component readiness |
| `createApprovalGate` / `waitForApproval` | CRD-based approval gates (replaces Argo suspend steps) |
| `proxyKafkaGoneTeardown` | Dual-mode proxy teardown: Kafka deleted → redeploy in non-capture mode |
| `migrateAndSelfTeardown` | Run migration → patch Ready → self-delete CRD |

### What Changed

| Area | Changes |
|------|---------|
| **Helm/k8s** | 3 new CRDs (ApprovalGate, KafkaCluster, TrafficReplay) with `dependsOn` schema; expanded RBAC for new types; CloudWatch dashboard refactored from Job to inline template in installJob.yaml; ACK controller `waitForInstallation` + `deletionPolicy` |
| **CLI (new)** | `reset.py` (384 lines — concurrent DAG deletion), `resubmit.py`, `suspend_steps.py` (Argo lifecycle utils), `crd_utils.py` |
| **CLI (rewritten)** | `approve.py` — from Argo suspend-step resumption to ApprovalGate CRD patching |
| **CLI (removed)** | `stop.py` command, `WorkflowService.stop_workflow()` |
| **CLI (improved)** | `submit.py` catches AlreadyExists and suggests resubmit; `script_runner.py` captures stderr for better error messages; `config.py` auto-generates raw_yaml from data dict |
| **Workflow templates** | `fullMigration.ts` — major rewrite adding CRD lifecycle for all components; `resourceManagement.ts` — 256 new lines of CRD CRUD, approval gates, teardown signals; all resource manifests use ownerReferences to CRDs |
| **Config processor** | `migrationInitializer.ts` generates KafkaCluster and TrafficReplay CRDs with dependsOn |
| **Integration tests** | `test_runner.py` runs `workflow reset --all` before cleanup; `k8s_service.py` cleans up ACK dashboard CRs; reduced resource requests in test configs; `monitor.sh` creates output file upfront for retry safety |
| **CI** | `k8sLocalDeployment.groovy` — cleanup timeout increased to 10min, added `--timeout` flags to kubectl commands |

### Testing
- **TypeScript**: 13 test suites, 129 tests, 13 snapshots — all passing
- **Python unit tests**: `test_reset.py` (188 lines) — CRD listing, dependency resolution, foreground deletion, CLI list/delete/cascade/reset-all modes
- **Python integration tests**: `test_reset_integration.py` (542 lines) — e2e with real k8s CRDs (k3s/minikube), approval gate lifecycle, autocomplete, submit error paths
- **Updated tests**: Rewrote approve tests for CRD model, removed stop command tests, removed WorkflowService.stop_workflow tests
- **Dependency isolation**: Parallel migrations only affect their own dependency chain


migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
```diff
@@ -0,0 +1,188 @@
...
+        click.echo("Timed out waiting for workflow completion.", err=True)
+
+    if delete_workflow(workflow_name, namespace, argo_server, token, insecure):
+        click.echo(f"  ✓ Deleted workflow '{workflow_name}'")
+    else:
+        click.echo(f"  ✗ Failed to delete workflow '{workflow_name}'", err=True)
+
+
+@click.command(name="reset")
+@click.argument('path', required=False, default=None)
```
gregschohn: I didn't think that we wanted to support granular resets yet.  I personally think that we SHOULD, but if we do, we should have autocomplete here (for me that is useful documentation).

gregschohn (commented):
I'm trying to get my head around the semantics of a reset.  What state should a reset leave the user in?  When should other commands be available w/r/t a reset?

We already have a need to do partial resets thanks to the capture proxy.  The mental model in my head is to have reset always stop the workflow.  The user should need to start it again (or the user could specify that the workflow should be automatically restarted).

DURING the reset, a user shouldn't be able to deploy another workflow - they should be forced to wait until the reset has completed.  The only easy path that I'm seeing to get clean semantics is to hijack the migration-workflow with a reset workflow that is run immediately after the existing one is deleted (if it was running).  Following that deletion, we'd kick off a new workflow that should wait for any cleanup to happen - I'm assuming that we need a period for teardown.

Maybe I'm missing it, but if this is implementing another model, can you please document it in the reset developer & user documentation.

gregschohn (commented):
I’m having an issue trying to reconcile why a user would need to delete a snapshot resource when an upstream proxy is deleted.  That makes sense for RFS (pre-delta snapshot) and the replayer, but since we support incremental snapshots, why not just update it.  I get the idea and like the completeness, but it feels like users’ needs are probably going to be subtle here.

Have you thought about this one at all yet?

The new VAP work has a similar graph for when and what to wait on for dependencies - snapshots will wait for proxy’s (actually ‘capturedtraffic’, which has an unused level of indirection from just the proxy) to be ready before it updates… and it will update IIRC if the proxy was updated.  The issue here is that we might have some resources that are really valuable to keep around for updates.  The current resource model seems to make that impossible.

Even if we’re only starting to do these smarter updates, they’re REALLY valuable for customers.  I fear that w/out support for them, they may resort/revert to more dangerous operations.

migrationConsole/lib/console_link/console_link/workflow/models/config.py
```diff
@@ -15,7 +16,16 @@ def __init__(self, data: Dict[str, Any] = None, raw_yaml: Optional[str] = None):
...
+        if raw_yaml is not None:
+            self.raw_yaml = raw_yaml
+        elif self.data:
+            yaml = YAML()
+            yaml.preserve_quotes = True
+            stream = StringIO()
+            yaml.dump(self.data, stream)
+            self.raw_yaml = stream.getvalue()
+        else:
+            self.raw_yaml = ""
```
gregschohn, line 28: why are we doing this again?  Why would you ever rewrite what the user sent in?

migrationConsole/lib/console_link/console_link/workflow/services/script_runner.py
```diff
@@ -93,9 +93,9 @@ def run(
                 if e.stderr:
                     print(f"stderr: {e.stderr}", file=sys.stderr)
             else:
-                logger.error(f"Script failed with exit code {e.returncode}")
+                logger.debug(f"Script failed with exit code {e.returncode}")
                 if e.stderr:
-                    logger.error(f"stderr: {e.stderr}")
+                    logger.debug(f"stderr: {e.stderr}")
```
gregschohn: why are these getting quieter?

gregschohn:
Suggestion: both `approve.py` and `reset.py` now independently implement the same CRD-listing and TTL-cached-autocompletion patterns. Consider folding them into `crd_utils.py` to eliminate ~40 lines of duplicated boilerplate.

**What's duplicated:**
1. `list_approval_gates()` in approve.py is a subset of `_list_migration_resources()` in reset.py — both call `list_namespaced_custom_object` and extract name/phase
2. Both files have their own cache-file-based autocompletion: `_get_cache_file` + `_get_cached_pending_names` (approve) vs `_get_reset_cache_file` + `_get_cached_resource_names` (reset) — same TTL check, same JSON read/write, same tempdir path

**Suggested refactoring — add two functions to crd_utils.py:**

```python
# crd_utils.py — add these alongside existing has_glob/match_names

RESETTABLE_PLURALS = [
    'kafkaclusters', 'capturedtraffics', 'datasnapshots',
    'snapshotmigrations', 'trafficreplays', 'approvalgates',
]

DISPLAY_NAMES = {
    'kafkaclusters': 'Kafka Cluster',
    'capturedtraffics': 'Capture Proxy',
    'datasnapshots': 'Data Snapshot',
    'snapshotmigrations': 'Snapshot Migration',
    'trafficreplays': 'Traffic Replay',
    'approvalgates': 'Approval Gate',
}


def list_migration_resources(namespace, plurals=None):
    """List CRD instances. Returns list of (plural, name, phase, deps)."""
    custom = client.CustomObjectsApi()
    plurals = plurals or RESETTABLE_PLURALS
    results = []
    for plural in plurals:
        try:
            items = custom.list_namespaced_custom_object(
                group=CRD_GROUP, version=CRD_VERSION,
                namespace=namespace, plural=plural
            ).get('items', [])
            for item in items:
                name = item['metadata']['name']
                phase = item.get('status', {}).get('phase', 'Unknown')
                deps = item.get('spec', {}).get('dependsOn', []) or []
                results.append((plural, name, phase, deps))
        except ApiException:
            pass
    return results


def cached_crd_completions(namespace, cache_key, fetch_names, ttl=10):
    """Generic TTL-cached name fetcher for shell autocompletion."""
    cache_dir = Path(tempfile.gettempdir()) / "workflow_completions"
    cache_dir.mkdir(exist_ok=True)
    cache_file = cache_dir / f"{cache_key}_{namespace}.json"

    if cache_file.exists() and (time.time() - cache_file.stat().st_mtime) < ttl:
        try:
            return json.loads(cache_file.read_text()).get('names', [])
        except Exception:
            pass

    try:
        load_k8s_config()
        names = fetch_names(namespace)
        cache_file.write_text(json.dumps({'names': names}))
        return names
    except Exception:
        return []
```

**Then approve.py simplifies to:**

```python
# Remove: list_approval_gates, _get_cache_file, _get_cached_pending_names
# Remove imports: json, tempfile, time
# Add imports from crd_utils: cached_crd_completions, list_migration_resources

def _pending_gate_names(namespace):
    return [n for _, n, phase, _ in list_migration_resources(namespace, ['approvalgates'])
            if phase == 'Pending']

def get_approval_task_name_completions(ctx, _, incomplete):
    ns = ctx.params.get('namespace', 'ma')
    return [CompletionItem(n)
            for n in cached_crd_completions(ns, 'approval_gates', _pending_gate_names)
            if n.startswith(incomplete)]
```

And in `approve_command`, replace:
```python
gates = list_approval_gates(namespace)
pending = [(name, phase) for name, phase in gates if phase == 'Pending']
```
with:
```python
pending_names = _pending_gate_names(namespace)
```

**And reset.py simplifies to:**

```python
# Remove: _list_migration_resources, RESETTABLE_PLURALS, DISPLAY_NAMES,
#         _get_reset_cache_file, _get_cached_resource_names
# Remove imports: json, tempfile
# Add imports from crd_utils: DISPLAY_NAMES, cached_crd_completions, list_migration_resources

def _resettable_names(namespace):
    return [n for _, n, phase, _ in list_migration_resources(namespace)
            if phase != 'Teardown']

def _get_resource_completions(ctx, _, incomplete):
    ns = ctx.params.get('namespace', 'ma')
    return [n for n in cached_crd_completions(ns, 'reset_resources', _resettable_names)
            if n.startswith(incomplete)]
```

Replace all `_list_migration_resources(namespace)` calls with `list_migration_resources(namespace)`.

This keeps the `plurals` parameter so approve can scope to just `['approvalgates']` while reset queries all types, and the `cache_key` parameter keeps cache files separate. The `approve_gate` function stays in approve.py since it's specific to status patching.

migrationConsole/lib/integ_test/integ_test/ma_workflow_test.py
```diff
@@ -26,7 +27,6 @@ def setup_and_teardown(request, keep_workflows, test_case: MATestBase):
         if request.node.rep_call and request.node.rep_call.failed:
             logger.info(f"Test failed - printing workflow details for {test_case.workflow_name}")
             test_case.argo_service.print_workflow_details(workflow_name=test_case.workflow_name)
-            # Save detailed diagnostics to logs directory (archived by Jenkins)
```
gregschohn, line 29: not a bad comment.  We should keep it around

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -98,13 +98,38 @@ export const FullMigration = WorkflowBuilder.create({
         .addRequiredInput("version", typeToken<string>())
         .addOptionalInput("groupName_view", c => "Kafka Cluster")
         .addOptionalInput("sortOrder_view", c => 999)
+        .addInputsFromRecord(makeRequiredImageParametersForKeys(["MigrationConsole"]))
 
         .addSteps(b => b
+            .addStep("createCrd", ResourceManagement, "createKafkaCluster", c =>
```
gregschohn: nit, because everybody does it, but you mean CR or custom resource, NOT custom resource definition.  Since we also DO need to maintain those crds too, the lack of precision will become problematic.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -113,6 +138,60 @@ export const FullMigration = WorkflowBuilder.create({
 
     // ── Section 2: Proxies ───────────────────────────────────────────────
 
+    // Helper: when Kafka is deleted, redeploy proxy in non-capture mode, then wait for CapturedTraffic deletion
+    .addTemplate("proxyKafkaGoneTeardown", t => t
```
gregschohn: There's a neat idea here about having the workflow run in listener mode to trigger a state change, but I think we should defer that as a feature for the post 3.0 release

gregschohn (commented):
There are a lot of changes to make here that are required for this code to behave as expected with the new checksum-dependency tracking system.

I've pulled this branch (as of 7AM Eastern) and am working through the necessary changes.  I'll push a new branch later today.

AndreKurait:
Superseded by newer PRs that cover this

PR #2486, opened 2026-03-20
Classify bulk item errors and add configurable non-retryable exception types in replay retry handler

AndreKurait:
Bulk responses with `"errors": true` were previously retried based on the source response. This PR parses individual item errors to distinguish retryable from non-retryable errors using a streaming JSON parser, and makes the non-retryable error set configurable via CLI.

**Retry decision for bulk 200 responses:**
- `"errors": false` → done (short-circuits without reading items)
- All item errors non-retryable (e.g. `version_conflict`, `mapper_parsing`) → done
- Any retryable item error (e.g. `unavailable_shards`, `cluster_block`, unknown types) → retry
- `error` field without `type` or as a plain string → retry (fail-open)
- Unparseable body (non-JSON, malformed) → fall back to superclass status code comparison

**Other status codes (unchanged):**
- 429/5xx → retry
- Non-200 non-bulk → superclass handles via source vs target status comparison

**Streaming parser:** Uses Jackson's non-blocking `ByteBufferFeeder` to stream-parse response chunks through the Netty pipeline. Short-circuits on `"errors": false` or first retryable error without buffering the full body. Tests verify the analyzer resolves before all chunks are delivered.

**Configurable non-retryable exceptions:** New `--non-retryable-doc-exception-types` CLI arg allows overriding the default set of non-retryable error types. When omitted, uses the built-in defaults. When provided, replaces the defaults entirely (absolute, not additive). Default error types are defined in a shared `BulkDocErrorTypes` class in `coreUtilities` so both the replayer and RFS can reference the same vocabulary.

**Distinction from RFS `--allowed-doc-exception-types`:**
- Replayer `--non-retryable-doc-exception-types`: don't retry, still counts as failure
- RFS `--allowed-doc-exception-types`: don't retry, treats as success
- Different defaults: replayer defaults to a populated set of known client errors; RFS defaults to empty (opt-in)

### DIFF
--- a/TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/http/retries/OpenSearchDefaultRetryE2ETest.java
+++ b/TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/http/retries/OpenSearchDefaultRetryE2ETest.java
@@ -11,6 +11,7 @@ import org.opensearch.migrations.replay.TrafficReplayerTopLevel;
 import org.opensearch.migrations.replay.TransformedTargetRequestAndResponseList;
 import org.opensearch.migrations.replay.datahandlers.NettyPacketToHttpConsumer;
 import org.opensearch.migrations.replay.datatypes.ByteBufList;
+import org.opensearch.migrations.replay.datatypes.ByteBufListProducer;
 import org.opensearch.migrations.replay.datatypes.HttpRequestTransformationStatus;
 import org.opensearch.migrations.replay.datatypes.TransformedOutputAndResult;
 import org.opensearch.migrations.testutils.SimpleHttpResponse;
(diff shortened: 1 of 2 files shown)


coreUtilities/src/main/java/org/opensearch/migrations/BulkDocErrorTypes.java
```diff
@@ -0,0 +1,35 @@
...
+
+import lombok.experimental.UtilityClass;
+
+/**
+ * Known OpenSearch bulk item error types classified by retry behavior.
+ * Shared between the traffic replayer (non-retryable classification) and
+ * RFS (document exception allowlist) so both use a consistent vocabulary.
+ */
+@UtilityClass
+public class BulkDocErrorTypes {
```
gregschohn, line 13: nit: this isn't the right package for this.  coreUtilities, IIRC, should be for necessities that would be required for just about any java project.  These are all very specific to OpenSearch/Elasticsearch HTTP traffic.

I think you'd be better off to move the OS specific stuff out of Java and ONLY track it in the orchestration layer (see below) when you can.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -392,6 +392,15 @@ export const USER_REPLAYER_PROCESS_OPTIONS = z.object({
         .describe("Maximum number of HTTP requests that can be in-flight simultaneously to the target cluster. Limits concurrency to prevent overwhelming the target."),
     numClientThreads: z.number().default(0).optional()
         .describe("Number of threads used to send replayed requests to the target. 0 uses the Netty event loop (typically number of available processors)."),
+    nonRetryableDocExceptionTypes: z.array(z.string()).optional()
+        .describe("List of document-level exception types that should not be retried during bulk replay. " +
```
gregschohn, line 396: Is there a reason that you're encoding the default list in here rather than setting the default directly here in the schema?  I think that more often, we've had two defaults in favor of being more explicit.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/http/retries/OpenSearchDefaultRetry.java
```diff
@@ -32,93 +30,201 @@
...
 
         @Override
         public void channelRead(@NonNull ChannelHandlerContext ctx, @NonNull Object msg) throws Exception {
-            if (msg instanceof HttpResponse && errorField == null) {
-                var statusCode = ((HttpResponse) msg).status().code();
-                if (statusCode != 200) {
-                    log.atDebug().setMessage("Bulk response status {} is not 200, treating as error")
-                        .addArgument(statusCode).log();
-                    errorField = true;
-                }
```
gregschohn, line 63: this was a redundant check from below

Issue #2487, opened 2026-03-21
[BUG] MetadataMigration error from es 8.17.0 to  opensearch 3.4 cluster


haojunq:
### What is the bug?
_A clear and concise description of the bug._

while I create snapshot from es 8.17.0 and migrate into  opensearch 3.4  and run  MetadataMigration:run commands
` ./gradlew MetadataMigration:run --args='migrate --snapshot-name snapshot_1 --file-system-repo-path /home/snapshot-es8 --source-version "ES_8.17" --target-host https://x.x.x.x:9200 --target-username admin --target-password "xxx" --target-insecure'`

, it return errors likes:

```
elasticsearch8-index failed on target cluster: Create object failed for elasticsearch8-index
{"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Version id 8521000 must contain OpenSearch mask"}],"type":"illegal_argument_exception","reason":"Version id 8521000 must contain OpenSearch mask"},"status":400}
```

i don't know why the versionid is 8521000 , but not 8170000,   both two version id are not defined in opensearch 3.4  Version.java
<img width="885" height="205" alt="Image" src="https://github.com/user-attachments/assets/db15e722-358a-44e1-a692-a48a1f555733" />

BTW:  i found an issue from elasticsearch repo , `IndexVersion (and similarly TransportVersion) are now decoupled from Version. The id now no longer maps directly to version ids.`
https://github.com/elastic/elasticsearch/issues/115796

```
{
  "snapshots" : [
    {
      "snapshot" : "snapshot_1",
      "uuid" : "PVNJBcsWRc-_VaW57MfuTQ",
      "repository" : "migration_repo",
      "version_id" : 8521000,
      "version" : "8.17.0",
      xxx
}
```

1. Does MetadataMigration can read version string from snapshot metadata but not versionid?
2. Even if the version_id is correct,  but opensearch 3.0 has removed the ES versions defined in the LegacyESVersion.java fle, Does MetadataMigration commands return error in all opensearch 3.x versions?

### What are your migration environments?

Source Cluster: Elasticsearch 8.17.0
TargetCluster: OpenSearch 3.4.0

### How can one reproduce the bug?
1. create an snapshot on Elasticsearch 8.17.0
PUT _snapshot/migration_repo/snapshot_1?wait_for_completion=false
  "indices": "*", 
  "ignore_unavailable": true,
  "include_global_state": true,
  "compress": false 
}'

[...]

PR #2499, opened 2026-03-24
Strip index.version.created setting during metadata transformation

sumobrian:
## Description

Fixes https://github.com/opensearch-project/opensearch-migrations/issues/2487

### Problem

When migrating from **Elasticsearch 8.17.0** to **OpenSearch 3.4**, MetadataMigration fails with:

```
Version id 8521000 must contain OpenSearch mask
```

**Root cause:** ES 8.x decoupled `IndexVersion` from the release version (see elastic/elasticsearch#115796). ES 8.17.0 writes `version_id: 8521000` (an internal IndexVersion) into snapshot index settings as `index.version.created`, instead of the release-derived `8170000`. When this internal setting is forwarded unchanged to OpenSearch during metadata migration, OpenSearch 3.x rejects it because it doesn't recognize the version ID.

### Fix

Added `TransformFunctions.removeVersionCreatedSetting()` which strips the internal `settings.version` node (containing `created` and `upgraded` fields) from index settings during transformation. This is safe because:
- `index.version.created` is a private/internal setting that records which version created the index
- The target cluster assigns its own version metadata when an index is created

The fix is wired into all transformer pipelines:
- `Transformer_ES_7_10_OS_2_11` (also used for ES 8.x and OS sources)
- `Transformer_ES_6_8_to_OS_2_11` (inherited by ES 5.x transformers)
- `Transformer_ES_2_4_to_OS_2_19`

### Testing

- 4 new unit tests for `removeVersionCreatedSetting()`
- Updated existing `Transformer_ES_7_10_OS_2_11Test` expected outputs to reflect stripped `version.created`
- All 13 transformer tests pass


AndreKurait:
Addressing in https://github.com/opensearch-project/opensearch-migrations/pull/2502 with tests

PR #2502, opened 2026-03-24
Fix 'Version id must contain OpenSearch mask' error when migrating ES 8.x to OS 3.4+

AndreKurait:
## Description

Fixes index creation failures when migrating from Elasticsearch 8.17+ to OpenSearch 3.4+.

### Problem

ES 8.17+ writes decoupled IndexVersion IDs (e.g. `8521000`) into `index.version.created` instead of release-derived version IDs. On OpenSearch 3.4+, sending this value triggers:

```
{"type":"illegal_argument_exception","reason":"Version id 8521000 must contain OpenSearch mask"}
```

This differs from OS 3.0 which rejects `version.created` as a private setting (already handled by the retry loop). OS 3.4+ actually attempts to parse the version ID and fails because it lacks the MASK bit (`0x08000000`).

### Fix

Added a `VERSION_MASK_ERROR` regex pattern in `InvalidResponse` to recognize `"Version id N must contain OpenSearch mask"` and map it to `index.version.created`. This lets the existing retry loop in `IndexCreator_OS_2_11` strip the setting and retry index creation successfully.

### Testing

- Unit test in `InvalidResponseTest` verifying the new pattern extracts `index.version.created`
- E2e test (`VersionCreatedSettingTest`) migrating ES 8.17 and ES 8.19 → OS 3.4.0 via snapshot
  - Fails without the fix (exit code 1)
  - Passes with the fix
- Also adds `OS_V3_4_0` container version

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/InvalidResponse.java
@@ -18,6 +18,7 @@ import lombok.extern.slf4j.Slf4j;
 public class InvalidResponse extends RfsException {
     private static final Pattern UNKNOWN_SETTING = Pattern.compile("unknown setting \\[([a-zA-Z0-9_.-]+)\\].+");
     private static final Pattern PRIVATE_SETTING = Pattern.compile(".*private index setting \\[([a-zA-Z0-9_.-]+)\\] can not be set explicitly.*");
+    private static final Pattern VERSION_MASK_ERROR = Pattern.compile("Version id \\d+ must contain OpenSearch mask");
     private static final Pattern UNSUPPORTED_MAPPING_PARAM = Pattern.compile("unsupported parameters:\\s+(.+)");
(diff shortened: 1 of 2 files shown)


RFS/src/testFixtures/java/org/opensearch/migrations/bulkload/framework/SearchClusterContainer.java
```diff
@@ -182,6 +182,7 @@ private static Map<String, String> overrideAndRemoveEnv(
     public static final ContainerVersion OS_V2_18_0 = OpenSearchVersion.fromTag("2.18.0");
     public static final ContainerVersion OS_V2_19_4 = OpenSearchVersion.fromTag("2.19.4");
     public static final ContainerVersion OS_V3_0_0 = OpenSearchVersion.fromTag("3.0.0");
+    public static final ContainerVersion OS_V3_4_0 = OpenSearchVersion.fromTag("3.4.0");
```
jugal-chauhan: Thanks for this

jugal-chauhan (approved):
Approving this PR as it is a good addition to mask the `index.version.created` field. 
Looking forward to #2505 supporting a new target version OS 3.4

PR #2509, opened 2026-03-24
Remove redundant virtualenv activation in migration console Dockerfile

AndreKurait:
## Description

The migration console Dockerfile sets up Python virtualenv activation via two paths in `~/.bashrc`:

1. **`/etc/profile.d/venv.sh`** — sourced from `~/.bashrc`, which in turn sources `/.venv/bin/activate`
2. **Direct `source /.venv/bin/activate`** — also appended to `~/.bashrc`

This causes the virtualenv to be activated twice every time an interactive shell starts. The second activation is redundant since `/etc/profile.d/venv.sh` already handles it.

## Changes

Removed the redundant `source /.venv/bin/activate` line from the `~/.bashrc` setup in the Dockerfile. The single activation path through `/etc/profile.d/venv.sh` is preserved.

## Testing

- No behavioral change — the virtualenv is still activated on every interactive shell session
- The only difference is it activates once instead of twice

## Diff

```diff
-    echo 'source /.venv/bin/activate' >> ~/.bashrc && \
```


jugal-chauhan (approved):
Thanks for the quick change!

PR #2525, opened 2026-03-25
Add resource requests/limits and do-not-disrupt to buildkit pods

AndreKurait:
## Overview

Buildkit pods created by `docker buildx` with the kubernetes driver were running as `BestEffort` QoS — no resource requests or limits. This meant they could be scheduled on undersized nodes and evicted under memory pressure during large multi-arch image builds.

### Resource requests/limits for buildkit pods (`setupK8sBuilders.sh`)
- Adds `--driver-opt` resource requests/limits to the `docker buildx create` calls for the kubernetes driver (EKS only, not minikube):
  - Requests: 4 cpu, 8Gi memory
  - Limits: 8 cpu, 16Gi memory
- All values are overridable via `BUILDKIT_REQUESTS_CPU`, `BUILDKIT_LIMITS_MEMORY`, etc. env vars
- Adds `karpenter.sh/do-not-disrupt` annotation to prevent Karpenter from evicting build nodes mid-build

### Build nodepool improvements (`build-nodepool.yaml`)
- Bumps nodepool limits from 32 cpu / 64Gi → 64 cpu / 128Gi to accommodate resource requests for both arch builders
- Changes disruption policy from `WhenEmptyOrUnderutilized` → `WhenEmpty` so build nodes are only reclaimed when they have zero pods

## Testing
- Minikube path is unaffected — resource opts are only in the EKS (kubernetes driver) branch
- `localTesting.sh` continues to zero out resources via `BUILDKIT_HELM_ARGS` for the static buildkitd pods

### DIFF
--- a/buildImages/setupK8sBuilders.sh
+++ b/buildImages/setupK8sBuilders.sh
@@ -38,13 +38,13 @@ if [[ "$CONTEXT" =~ (eks:|gke_|aks-|migration-eks-) ]]; then
   
   # Resource requests/limits for buildkit pods — ensures pods get scheduled on
   # appropriately sized nodes and aren't evicted during large builds.
-  # do-not-disrupt prevents Karpenter from evicting build nodes mid-build.
+  # Note: Karpenter disruption protection is handled by the NodePool's WhenEmpty
+  # consolidation policy — the kubernetes driver doesn't support pod annotations.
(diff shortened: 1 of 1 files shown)


gregschohn (commented):
If you can get the eks integ test to build cleanly in Jenkins, then this LGTM,.

PR #2540, opened 2026-03-26
Fix migration console pod scheduling to dedicated Karpenter NodePool

AndreKurait:
## Problem

The migration console pod lands on `general-purpose` instead of `dedicated-small-instance-migration-console-pool`, even though the dedicated pool exists and the pod has the correct toleration.

The `requiredDuringSchedulingIgnoredDuringExecution` node affinity allows both `dedicated-small-instance-migration-console-pool` and `general-purpose`. Since `general-purpose` already has 3 ready nodes, the scheduler places the pod there immediately — the `preferredDuringScheduling` weight of 100 for the dedicated pool is just a hint and gets ignored when a valid node already exists.

```
kubectl get nodepools
NAME                                              NODES   READY
dedicated-small-instance-migration-console-pool   0       True    # never gets a node
general-purpose                                   3       True    # pod lands here
```

## Fix

Remove `general-purpose` from the `required` node affinity so the pod can **only** schedule on `dedicated-small-instance-migration-console-pool`. This forces Karpenter to provision a `t`-family on-demand node in the dedicated pool. The `preferred` rules are also removed since they are redundant when there is only one allowed pool.

The taint (`migration-console=dedicated:NoSchedule`) keeps other pods off the dedicated node, and the hard affinity constraint ensures the migration console actually lands there.

## Testing

- Verified the rendered template produces valid YAML with the correct affinity/toleration structure
- The dedicated pool's `consolidationPolicy: WhenEmpty` with `consolidateAfter: 5m` ensures the node stays stable while the console is running


jugal-chauhan (approved):
LGTM

PR #2545, opened 2026-03-26
Assemble self-contained aws-bootstrap.sh for release

AndreKurait:
## Problem

`aws-bootstrap.sh` calls external scripts (`mirrorToEcr.sh`, `generatePrivateEcrValues.sh`, `privateEcrManifest.sh`) that live in the repo tree. But the user install path downloads only `aws-bootstrap.sh` as a standalone file:

```bash
curl -sL -o aws-bootstrap.sh "https://github.com/.../releases/latest/download/aws-bootstrap.sh"
```

This means image mirroring (which runs by default) fails for standalone users because the helper scripts aren't available.

## Fix

Refactor the helper scripts so their core logic is in functions with `BASH_SOURCE` main guards — they work both standalone and when sourced.

`aws-bootstrap.sh` now sources these scripts at runtime. For release, a new `assemble-bootstrap.sh` inlines all dependencies into a single self-contained file.

### How it works

**Development (repo checkout):**
```
aws-bootstrap.sh → sources mirrorToEcr.sh, generatePrivateEcrValues.sh, privateEcrManifest.sh
```

**Release (standalone download):**
```
assemble-bootstrap.sh → produces dist/aws-bootstrap.sh with everything inlined
release-drafter.yml → publishes dist/aws-bootstrap.sh
```

### Files changed
- `mirrorToEcr.sh`: Extract `mirror_images_to_ecr()`, `mirror_charts_to_ecr()`, `copy_image()`, `ensure_crane()` functions
- `generatePrivateEcrValues.sh`: Extract `generate_private_ecr_values()` function
- `aws-bootstrap.sh`: Source helpers and call functions instead of external scripts
- **New**: `assemble-bootstrap.sh` — concatenates function definitions + main script
- `release-drafter.yml`: Run assembly, publish `dist/aws-bootstrap.sh`


gregschohn (commented):
Thanks for getting to this... I got totally distracted with this when I was putting those other scripts together.

An alternative could be to look for the directory of the file that we're trying to run and downloading each on demand - or packaging a tarball of those alongside the script & having the bootstrap do that download.  

My two concerns are that the script could be really long & it won't be exactly what developers are using.  I think that we'd probably move from using separate processes to running within the same shell, which will also have an impact on variable scopes.

deployment/k8s/aws/assemble-bootstrap.sh
```diff
@@ -0,0 +1,84 @@
...
+    extract_body "$abs_path"
+    echo "# --- END inlined: $basename ---"
+    echo ""
+  done
+
+  # Emit the main bootstrap script, removing the source block
+  sed -E \
+    -e '1{/^#!/d}' \
+    -e '/^# --- source helper scripts/,/^fi$/d' \
+    "$BOOTSTRAP"
```
jugal-chauhan: This seems fragile. Added a comment below to consider adding a ending block and then modifying the logic here to remove everything between these 2 comment-lines

PR #2568, opened 2026-03-29
Added support for nested facets

akshay2000:
### Description
Implements Solr's `facet` sub-facet nesting for all facet types (terms, range, query) by recursively converting nested `facet` maps to OpenSearch `aggs` on the request side, and recursively unwinding nested aggregation results back to Solr's format on the response side.

### Testing
- Unit tests
- E2E tests

### Check List


niravpi:
Looks good overall, also good test coverage across facet types. 
Couple of nits: 
- the response-side tests only exercise nested aggs inside terms buckets — it'd be good to add a case with date_histogram buckets (which have key_as_string) to verify BUCKET_META_KEYS filtering works there too.
-  Similarly, a mixed nesting chain like query → terms → range would cover the filter + bucket combination, thoughts ?
Not blocking PR on this as it can be added as UT coverage enhancement if needed

PR #2569, opened 2026-03-30
Add explicit index template mappings and reporting config wiring for OpenSearchMetricsSink

Shivani-techno:
### Description
Adds explicit index template creation to OpenSearchMetricsSink so validation metrics
indices have consistent, optimized field mappings instead of relying on dynamic mapping.

Changes:
- OpenSearchMetricsSink: Creates composable index template at startup with explicit
  types (keyword for aggregation fields, date for timestamp, nested for comparisons,
  text for request bodies with no length limit)
- ReportingConfig: New YAML config parser for reporting framework settings
- ShimMain: Added --reporting-config CLI parameter to enable validation reporting
- ShimProxy/MultiTargetRoutingHandler: Wired MetricsReceiver into the request pipeline
  to collect and publish validation metrics after each request
- docker-compose.validation.yml: Added reporting config mount for shim-solr-primary
- reporting-config.yaml: Sample configuration with placeholder credentials


akshay2000:
Please fix the DCO failure.

akshay2000 (changes requested):

akshay2000 (changes requested):

akshay2000 (changes requested):

akshay2000 (approved):

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/reporting/OpenSearchMetricsSink.java
```diff
@@ -0,0 +1,374 @@
...
+            }
+            """, indexPrefix);
+    }
+
+    @Override
+    public void submit(ValidationDocument document) {
+        try {
+            synchronized (buffer) {
+                buffer.add(document);
+                if (buffer.size() >= bulkSize) {
```
AndreKurait: Depending on number of documents without consideration for their size is troublesome. E.g. if each doc was 20MB this would be a 2GB request which opensearch would reject

AndreKurait (changes requested):
Please take a look at the tuple interface in the replayer. See https://github.com/opensearch-project/opensearch-migrations/pull/2605 which can be leveraged for high performance validation here.

AndreKurait (commented):
This comment serves so i can use "Changes since reivew" post force-push's

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/reporting/OpenSearchMetricsSink.java
```diff
@@ -0,0 +1,388 @@
...
+        } catch (JsonProcessingException e) {
+            log.error("Failed to serialize ValidationDocument batch", e);
+        }
+    }
+
+    /**
+     * Checks for partial failures in bulk response.
+     * Even if HTTP status is 200, individual documents may have failed.
+     */
+    void checkPartialFailures(String responseBody, int totalDocs) {
```
AndreKurait: We already have exisitng logic in RFS which does this as well as a bisect on failure to only retry the failed docs, could be useful here

jugal-chauhan:
Checking in, are there more changes expected here ? Should we move this PR into draft until these changes are made ?

akshay2000:
Closing this PR because we have a different roadmap.

PR #2583, opened 2026-03-30
Detect already-existing items and surface non-zero ex…

Davidding4718:
### Description
Explicitly detect metadata items that already exist on the target cluster during migrate/evaluate runs, instead of silently skipping them.

When a migration is re-run after a partial failure, previously migrated items would be silently skipped, creating stale-schema risk if the source schema changed between runs. This change surfaces those items with a non-zero exit code, forcing operators to make a deliberate decision.

The `--allow-existing` flag restores the legacy silent-skip behavior for operators who have verified their target state.

### Issues Resolved
https://opensearch.atlassian.net/browse/MIGRATIONS-3015

### Testing
- Unit tests: MigrateTest, EvaluateTest, ItemsTest cover exit-code behavior, CLI/JSON output formatting, and `--allow-existing` flag
- Property-based tests (jqwik, 100 iterations each): verify exit-code correctness, CLI output completeness, and JSON field presence across random item combinations
- E2E tests: partial-failure re-run scenario exits non-zero due to `ALREADY_EXISTS;` same scenario with `--allow-existing` exits 0

### Check List

### DIFF
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexRunner.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/worker/IndexRunner.java
@@ -49,15 +49,7 @@ public class IndexRunner {
                         .build());
             } else {
                 var rawResults = createIndex(index.getName(), mode, context);
-                if (allowExistingIndices) {
-                    creationResults = rawResults.stream()
-                        .map(r -> r.getFailureType() == CreationFailureType.ALREADY_EXISTS
-                            ? CreationResult.builder().name(r.getName()).build()
-                            : r)
-                        .collect(Collectors.toList());
-                } else {
-                    creationResults = rawResults;
-                }
+                creationResults = applyAllowExistingIndices(rawResults);
             }
 
             creationResults.forEach(results::index);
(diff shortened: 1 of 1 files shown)


MetadataMigration/src/main/java/org/opensearch/migrations/cli/Items.java
```diff
@@ -143,6 +193,11 @@ public JsonNode asJsonOutput() {
 
         root.put("dryRun", dryRun);
 
+        int alreadyExistsCount = getAlreadyExistsCount();
+        if (alreadyExistsCount > 0) {
+            root.put("alreadyExistsCount", alreadyExistsCount);
```
AndreKurait: We have several defined error types, why is this one bubbled up to the top level?

MetadataMigration/src/test/java/org/opensearch/migrations/EndToEndTest.java
```diff
@@ -293,6 +296,116 @@ private List<String> getNames(List<CreationResult> items) {
...
+            assertThat(targetOperations.get("/" + indexA).getKey(), equalTo(200));
+            assertThat(targetOperations.get("/" + indexB).getKey(), equalTo(404));
+
+            // Re-run migration for both indices — indexA already exists → ALREADY_EXISTS
+            var run2Args = new MigrateOrEvaluateArgs();
+            run2Args.sourceArgs.host = sourceCluster.getUrl();
+            run2Args.targetArgs.host = targetCluster.getUrl();
+            // No allowlist — migrate all indices
+
+            MigrationItemResult run2Result = executeMigration(run2Args, MetadataCommands.MIGRATE);
```
AndreKurait: Let's run EVALUATE first to verify it fails too

MetadataMigration/src/test/java/org/opensearch/migrations/EndToEndTest.java
```diff
@@ -293,6 +296,116 @@ private List<String> getNames(List<CreationResult> items) {
         return items.stream().map(CreationResult::getName).collect(Collectors.toList());
     }
 
+    @Test
+    @Tag("isolatedTest")
+    @SneakyThrows
+    void partialFailureRerun_exitsNonZeroDueToAlreadyExists() {
```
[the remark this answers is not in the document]
AndreKurait: @Davidding4718 has this been addressed?

PR #2605, opened 2026-04-01
Replay tuples to S3: pluggable TupleSink, per-thread writing, direct S3 streaming

AndreKurait:
## Summary

Replace the Log4J2 async appender tuple path with a pluggable `TupleSink` interface and per-thread writing via `ThreadLocalTupleWriter`. Each Netty event loop thread gets its own sink instance via `FastThreadLocal` — no cross-thread synchronization needed. Kafka offset commits are gated on write confirmation (`fsync` for file sinks, S3 upload completion for direct S3).

## New `tupleSink` Gradle module (`TrafficCapture/tupleSink`)

- `TupleSink` — pluggable interface for tuple destinations (single-threaded per instance)
- `GzipJsonLinesSink` — gzip-compressed JSON lines to local files with rotation by size/age, `syncFlush` + `fsync` per batch for durability while preserving the deflate dictionary
- `S3TupleSink` — streams gzip-compressed tuples directly to S3 via `BlockingOutputStreamAsyncRequestBody` (no in-memory buffering of full objects)
- `CallbackTupleSink` — test helper that completes futures immediately

## Per-thread writing (`trafficReplayer`)

- `ThreadLocalTupleWriter` — uses Netty `FastThreadLocal` to assign each event loop its own `TupleSink`
- `ParsedHttpMessagesAsDicts.toTupleMap()` — builds the structured `Map<String, Object>` consumed by `TupleSink`
- `TrafficReplayerCore.packageAndWriteTuple()` — async write path that defers Kafka commit until write future completes
- `TrafficReplayerTopLevel` — wiring for both new async path (`ThreadLocalTupleWriter`) and legacy sync path
- CLI flags: `--tuple-s3-bucket`, `--tuple-s3-region`, `--tuple-s3-prefix`, `--tuple-s3-endpoint`, `--tuple-max-buffer-seconds`, `--tuple-max-file-size-mb`, `--tuple-max-per-file`

## Helm / Kubernetes

- S3 artifacts bucket env vars injected into migration console when `defaultBucketConfiguration.create` is enabled
- Migration console init container generates unified workflow schema from live Strimzi OpenAPI

## Schema / orchestration

- `userSchemas.ts` — added `tupleS3Bucket`, `tupleS3Region`, `tupleS3Prefix`, `tupleMaxPerFile`, `tupleMaxBufferSeconds`, `tupleMaxFileSizeMb` to replayer process options

## Tests

- `GzipJsonLinesSinkTest` — rotation by size/age, periodic flush, future lifecycle
- `TupleWriteBlockingBehaviorTest` — verifies offset commits are gated on tuple write future completion
- Existing e2e tests updated to use `ThreadLocalTupleWriter` with `CallbackTupleSink`

## Architecture

```
Netty EL 0  →  TupleSink 0  →  tuples-0-{ts}-{seq}.log.gz  (GzipJsonLinesSink)
Netty EL 1  →  TupleSink 1  →  s3://{bucket}/{prefix}/...   (S3TupleSink)
...
Netty EL N  →  TupleSink N
```

Each event loop thread writes to its own sink via `FastThreadLocal`. For `GzipJsonLinesSink`, `syncFlush` pushes compressed data through without resetting the deflate dictionary (~39x compression). For `S3TupleSink`, tuples are gzip-compressed and streamed directly to S3 as they arrive via `BlockingOutputStreamAsyncRequestBody` — the CRT client handles multipart upload automatically.

## Offset Commit Gating

```
handleCompletedTransaction()
  → packageAndWriteTuple() → CompletableFuture<Void>
       → .whenComplete(() -> commitTrafficStreams(...))
```

Commits are deferred until the write future completes, ensuring no data loss if the process crashes after commit.


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -186,6 +189,14 @@ public static class Parameters {
...
         private TupleTransformationParams tupleTransformationParams = new TupleTransformationParams();
 
+        @Parameter(
+            required = false,
+            names = { "--enable-sync-tuples", "--enableSyncTuples" },
+            arity = 0,
+            description = "Enable the new synchronous tuple writing infrastructure (Disruptor + GzipJsonLinesSink). "
+                + "When not set, tuples are written via the legacy Log4J2 async appender path.")
+        boolean enableSyncTuples;
+
```
jugal-chauhan: I don't see `replayer.ts` being updated. Shouldn't we add  `--enable-sync-tuples` to the config dict ? So that all EKS deployments use this route

jugal-chauhan (commented):
Thanks for making these changes. I have some comments on whether do we add more tests and observability/metrics to the new path ?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayerCore.java
```diff
@@ -191,25 +196,39 @@ Void handleCompletedTransaction(
...
+                                tupleHandlingContext,
+                                resultTupleConsumer,
+                                rrPair,
+                                summary,
+                                (Exception) t
+                            );
+                        }
                     }
                     // Count the final outcome once per request (not per retry)
                     countFinalOutcome(summary, t);
```
AndreKurait: Note: Move this in the callback

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/s3/tupleConfigMap.yaml
```diff
@@ -0,0 +1,13 @@
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: migrations-default-s3-tuple-config
+  labels:
+    workflows.argoproj.io/configmap-type: Parameter
+data:
+  TUPLE_S3_BUCKET: {{ printf "\"migrations-default-%s-%s-%s\"" (toString .Values.aws.account) .Values.stageName .Values.aws.region }}
```
gregschohn: We have something like this for the snapshot bucket....
```
  {{- if .Values.defaultBucketConfiguration.useLocalStack }}
  ENDPOINT: "localstack://{{ .Values.defaultBucketConfiguration.endpoint }}.{{ .Release.Namespace }}.svc.cluster.local:4566"
  ENDPOINT_HTTP: "http://{{ .Values.defaultBucketConfiguration.endpoint }}.{{ .Release.Namespace }}.svc.cluster.local:4566"
  {{- else }}
```

Where do we create this bucket?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -9,7 +9,9 @@ import {
 
 export type ContainerVolumePair = {
     readonly container: Container,
-    readonly volumes: Volume[]
+    readonly volumes: Volume[],
+    readonly sidecars: Container[],
+    readonly initContainers: Container[]
```
gregschohn: I always forget, what's the difference between sidecars and init containers?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -45,7 +47,22 @@ export function setupTestCredsForContainer(
...
+        initContainers: def.initContainers.map(ic => {
+            const {volumeMounts: icMounts, ...restOfIc} = ic as any;
+            return {
+                ...restOfIc,
+                volumeMounts: [
+                    ...(icMounts ?? []),
+                    {
+                        name: TEST_CREDS_VOLUME_NAME,
+                        mountPath: "/config/credentials",
+                        readOnly: true
```
gregschohn: this seems brittle

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -125,6 +142,136 @@ export function setupLog4jConfigForContainer(
...
+ *
+ * Uses a hostPath volume for the mount point to avoid SELinux relabeling failures.
+ * Creates per-pod subdirectories (using POD_NAME) inside the hostPath.
+ */
+export function setupS3MountpointVolumeForContainer(
+    s3Bucket: BaseExpression<string>,
+    s3Region: BaseExpression<string>,
+    s3Prefix: BaseExpression<string>,
+    mountPath: string,
+    useLocalStack: BaseExpression<boolean>,
```
gregschohn: I'd prefer to see the endpoint passed in directly - we might have users that are using non-conventional S3 or MinIO endpoints.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -125,6 +142,136 @@ export function setupLog4jConfigForContainer(
...
+    mountpointS3Image: BaseExpression<string>,
+    def: ContainerVolumePair): ContainerVolumePair {
+
+    const SHARED_HOSTPATH_VOLUME = "s3-tuple-mnt";
```
gregschohn: this should be passed as an argument

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -125,6 +142,136 @@ export function setupLog4jConfigForContainer(
...
+        initContainers: [
+            ...def.initContainers,
+            {
+                name: "mount-s3-tuples",
+                restartPolicy: "Always",
+                image: makeStringTypeProxy(mountpointS3Image),
+                command: ["sh", "-c"],
+                args: [
+                    "export PATH=/mountpoint-s3/bin:$PATH; " +
+                    "P=/mnt/.pods/${POD_NAME}; " +
```
gregschohn: How can I as a developer find out what these settings need to be?
I'm looking at https://github.com/awslabs/mountpoint-s3/blob/main/doc/CONFIGURATION.md and that's a great guide, but requires devs to read the whole thing.  Is there a single place that can show me all of the settings?  Can you document what each one of these is for.
Some of these values should be parameterized (probably at the TS level, not argo - if we're RO/RW, we would know that at compile time).

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -9,7 +9,9 @@ import {
 
 export type ContainerVolumePair = {
```
gregschohn, line 10: rename this.  Can you make sidecars, initContainers, and volumes default to []?

gregschohn: Actually - I'm confused.  Is this meant to no represent a pod?  That's significantly different than a container and its volumes.  It feels like this abstraction is degrading and it's bleeding into other types that don't need to be degraded.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts
```diff
@@ -98,10 +118,18 @@ function getReplayerDeploymentManifest
...
     };
-    const finalContainerDefinition =
-        setupLog4jConfigForContainer(args.useCustomLogging, args.loggingConfigMap,
-            {container: baseContainerDefinition, volumes: []},
-            args.jvmArgs);
+    const withLog4j = setupLog4jConfigForContainer(args.useCustomLogging, args.loggingConfigMap,
+        {container: baseContainerDefinition, volumes: [], sidecars: [], initContainers: []},
+        args.jvmArgs);
+    const withS3 = setupS3MountpointVolumeForContainer(
+        args.tupleS3Bucket,
```
gregschohn: these would probably be better to pass around as an aggregate object from the user schema and only have to deal with one value rather than multiple tuple-s3 config values

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts
```diff
@@ -128,15 +156,39 @@ function getReplayerDeploymentManifest
...
+                    containers: [finalContainerDefinition.container, ...finalContainerDefinition.sidecars],
+                    volumes: [...finalContainerDefinition.volumes]
                 },
             },
         }
     }
 }
 
 
+function defaultTupleS3Config(s3TupleConfigMap: AllowLiteralOrExpression<string>) {
```
gregschohn: I'd prefer to see us move completely to the workflow config for this sort of stuff rather than having any defaults.  Defaults are more for us to support and more to explain to customers

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts
```diff
@@ -183,28 +244,47 @@ export const Replayer = WorkflowBuilder.create({
...
+
+        .addInputsFromRecord(makeRequiredImageParametersForKeys(["TrafficReplayer", "MountpointS3"]))
 
-        .addInputsFromRecord(makeRequiredImageParametersForKeys(["TrafficReplayer"]))
+        .addSteps(b => {
+            // Resolve effective values: user override wins, else ConfigMap default
+            const userBucket = expr.asString(expr.dig(expr.deserializeRecord(b.inputs.replayerOptions), ["tupleS3Bucket"], ""));
+            const userRegion = expr.asString(expr.dig(expr.deserializeRecord(b.inputs.replayerOptions), ["tupleS3Region"], ""));
+            const userPrefix = expr.asString(expr.dig(expr.deserializeRecord(b.inputs.replayerOptions), ["tupleS3Prefix"], ""));
+            const effectiveBucket = expr.cast(expr.ternary(expr.isEmpty(userBucket), b.inputs.defaultTupleS3Bucket, userBucket)).to<string>();
```
gregschohn: Another option on defaulting would be to do this during initialization - when the s3 configs weren't set, pull them in from the default configmap.  That would simplify this and keep the workflow size down.  That makes everything easier to test and keeps the workflow deterministic.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -280,6 +280,14 @@ export const USER_REPLAYER_WORKFLOW_OPTIONS = z.object({
...
             "By default, limits equal requests, giving the pod 'Guaranteed' QoS (least likely to be evicted). " +
             "Setting requests lower than limits results in 'Burstable' QoS, allowing the pod to use less resources when idle but burst up to the limit."),
+    tupleS3Bucket: z.string().default("").optional()
+        .describe("S3 bucket name for tuple output via Mountpoint S3. When set, a mount-s3 sidecar mounts the bucket at the tupleOutputDir path. Leave empty to write tuples to local storage only."),
+    tupleS3Region: z.string().default("").optional()
+        .describe("AWS region for the tuple S3 bucket. Required when tupleS3Bucket is set."),
+    tupleS3Prefix: z.string().default("tuples/").optional()
+        .describe("S3 key prefix within the tuple bucket. Defaults to 'tuples/'."),
+    useLocalStack: z.boolean().default(false).optional()
+        .describe("Use LocalStack S3 endpoint instead of real AWS S3. Only for local testing."),
```
gregschohn: These seem like they need to be grouped together so that you can easily explain the semantics.  If prefix is set, but nothing else is, will we use the default tuple bucket from the config map?  What if I remove the prefix, do we still use it?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -316,6 +324,12 @@ export const USER_REPLAYER_PROCESS_OPTIONS = z.object({
         .describe("Base64-encoded tuple transformer configuration." + TUPLE_TRANSFORMER_SUFFIX),
     tupleTransformerConfigFile: z.string().optional()
         .describe("Path to a JSON file containing tuple transformer configuration." + TUPLE_TRANSFORMER_SUFFIX + EXPERT_FILE_SUFFIX),
+    tupleOutputDir: z.string().default("./tuples").optional()
```
gregschohn: how would a user know what the file path would be for a path that they don't control if it's an s3 mountpoint?  If a user would put in a local path here, they'd likely blow out their disk and cause the replayer to crash, right?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpRequestConvertHandler.java
```diff
@@ -31,7 +31,9 @@ public NettyDecodedHttpRequestConvertHandler(
     @Override
     public void channelRead(@NonNull ChannelHandlerContext ctx, @NonNull Object msg) throws Exception {
         if (msg instanceof HttpRequest) {
-            httpTransactionContext.onHeaderParse();
+            if (httpTransactionContext != null) {
```
gregschohn: Why would this ever be set to null?  If we didn't need the null checks until now, what has changed?
IIRC, we have often just set up dummy contexts and we typically have test context libraries just for this.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/sink/ThreadLocalTupleWriter.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.function.IntFunction;
+
+import org.opensearch.migrations.replay.ParsedHttpMessagesAsDicts;
+import org.opensearch.migrations.replay.SourceTargetCaptureTuple;
+
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Per-thread tuple writer that eliminates the need for a shared Disruptor.
```
gregschohn: drop 'shared disruptor'

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/sink/ThreadLocalTupleWriter.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>Each calling thread gets its own {@link TupleSink} instance (typically a
+ * {@link GzipJsonLinesSink}). Since each Netty event loop is single-threaded,
+ * the per-thread sink requires no synchronization — both the gzip write and the
+ * flush+fsync happen on the calling thread.</p>
```
gregschohn: I'd like to see these files more composable for different policies, etc - but that can be for another day when we have more options.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/sink/ThreadLocalTupleWriter.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>The fsync cost (~1ms for small writes to Mountpoint S3) is comparable to the
```
gregschohn: this begs the question - what about for 'large writes' - like if we have 100MB buffered in a file?  There's no disruptor in mainline here, so it's worth scrubbing a reference to it

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/sink/ThreadLocalTupleWriter.java
```diff
@@ -0,0 +1,76 @@
...
+ * Disruptor publish latency it replaces, and keeps the single-threaded invariant
+ * that {@link GzipJsonLinesSink} depends on.</p>
+ */
+@Slf4j
+public class ThreadLocalTupleWriter implements AutoCloseable {
+    private final ConcurrentHashMap<Long, TupleSink> sinks = new ConcurrentHashMap<>();
+    private final AtomicInteger threadIndexCounter = new AtomicInteger();
```
gregschohn: The usage of the word 'thread' is confusing/misleading me.  Is a thread index just an index within this instance?  Why does this class need to care about its threading model?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
```diff
@@ -129,6 +129,23 @@ public ParsedHttpMessagesAsDicts(
         fillStatusCodeMetrics(context, sourceResponseOp, targetResponseOps4);
     }
 
+    /**
+     * Build the structured tuple map used by {@link org.opensearch.migrations.replay.sink.TupleSink} implementations.
+     */
+    public Map<String, Object> toTupleMap(SourceTargetCaptureTuple tuple) {
```
gregschohn, line 135: What's different from how we used to do this before?  Are we making an extra round trip on serialization?  If it's just for adding one more level hierarchy to dictionaries that were already parsed, that's probably fine/not that impactful

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ParsedHttpMessagesAsDicts.java
```diff
@@ -171,17 +188,16 @@ private static Map<String, Object> convertRequest(
         @NonNull List<byte[]> data
     ) {
         return makeSafeMap(context, () -> {
-            try (var transformationCtx = context.getLogicalEnclosingScope().createTransformationContext();
-                var messageHolder = RefSafeHolder.create(
+            try (var messageHolder = RefSafeHolder.create(
```
gregschohn: Why don't we care about these metrics any more?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -186,6 +188,14 @@ public static class Parameters {
         @ParametersDelegate
         private TupleTransformationParams tupleTransformationParams = new TupleTransformationParams();
 
+        @Parameter(
+            required = false,
+            names = { "--enable-sync-tuples", "--enableSyncTuples" },
```
gregschohn: why does this use the word 'sink'?  Is this necessary if the other values are present (like output dir)?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -289,6 +299,28 @@ public static class Parameters {
...
+            arity = 1,
+            description = "Maximum seconds before rotating/committing a tuple file")
+        int tupleMaxLagSeconds = 60;
+
+        @Parameter(
+            required = false,
+            names = { "--tuple-max-file-size-mb", "--tupleMaxFileSizeMb" },
+            arity = 1,
+            description = "Maximum uncompressed size in MB before rotating a tuple file")
+        int tupleMaxFileSizeMb = 256;
```
gregschohn, line 372: is this a max or is it a trigger size?  Are we guaranteed to get < 256MB uncompressed?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -289,6 +299,28 @@ public static class Parameters {
...
+            arity = 1,
+            description = "Directory for compressed tuple output files (local path or Mountpoint S3 mount)")
+        String tupleOutputDir = "./tuples";
+
+        @Parameter(
+            required = false,
+            names = { "--tuple-max-lag-seconds", "--tupleMaxLagSeconds" },
+            arity = 1,
+            description = "Maximum seconds before rotating/committing a tuple file")
+        int tupleMaxLagSeconds = 60;
```
gregschohn: I'd really like to see a mode where we write just one file tuple per file too.  I'm worried about addressability for users.  In that mode, I'd hope to see us be able to take advantage of directory paths to help us out - like PROXY_NODE_ID/CONNECTION_ID/INDEX.json

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/sink/ThreadLocalTupleWriter.java
```diff
@@ -0,0 +1,76 @@
...
+        var future = new CompletableFuture<Void>();
+        var map = parsed.toTupleMap(tuple);
+        sink.accept(map, future);
+        sink.onEndOfBatch();
+        return future;
+    }
+
+    @Override
+    public void close() {
+        sinks.values().forEach(sink -> {
```
gregschohn: this doesn't seem like it is threadsafe.  You're on a single thread and calling close on lots of objects that assume that they're being used only from one thread.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -571,17 +603,35 @@ private static void runReplayMode(Parameters params) throws Exception {
...
+                    blockingTrafficSource,
+                    timeShifter,
+                    tupleWriter,
+                    Duration.ofMillis(params.quiescentPeriodMs)
+                );
+            } else {
+                var resultsToLogsConsumer = new ResultsToLogsConsumer(null, null,
+                        () -> transformationLoader.getTransformerFactoryLoader(tupleTransformerConfig));
+                var tupleWriter = new TupleParserChainConsumer(resultsToLogsConsumer);
+                tr.setupRunAndWaitForReplayWithShutdownChecks(
```
gregschohn: isn't this the same invocation as the previous branch?  Refactor to shrink the size of the branch

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -596,6 +646,21 @@ private static void runReplayMode(Parameters params) throws Exception {
         }
     }
 
+    /**
+     * When running with a Mountpoint S3 sidecar, each pod's FUSE mount lives at
+     * {@code <hostPath>/.pods/<POD_NAME>/s3}. Append that suffix so the replayer
+     * writes through the FUSE mount instead of to the bare hostPath root.
+     */
+    static Path resolveTupleOutputDir(String tupleOutputDir) {
+        var podName = System.getenv("POD_NAME");
```
gregschohn: The replayer shouldn't know anything about k8s - this should be passed into it.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayerCore.java
```diff
@@ -191,25 +196,39 @@ Void handleCompletedTransaction(
...
+                                if (writeErr != null) {
+                                    log.atError().setCause(writeErr)
+                                        .setMessage("Tuple write failed for {} (streams={})")
+                                        .addArgument(context)
+                                        .addArgument(rrPair.trafficStreamKeysBeingHeld).log();
+                                }
+                                commitTrafficStreams(rrPair.completionStatus, rrPair.trafficStreamKeysBeingHeld);
+                            });
+                        } else {
+                            packageAndWriteResponse(
```
gregschohn, line 217: Can we make a default tuple writer that does the legacy stuff?  I would have expected the existing tupleconsumer to be used as a starting point - and then add commitTrafficStreams as a (possibly indirect) responsibility.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayerTopLevel.java
```diff
@@ -162,7 +163,47 @@ public void setupRunAndWaitForReplayToFinish(
...
+            trafficSource, timeShifter, null, resultTupleConsumer, null, quiescentDuration);
+    }
+
+    public void setupRunAndWaitForReplayToFinish(
+        Duration observedPacketConnectionTimeout,
+        Duration targetServerResponseTimeout,
+        BlockingTrafficSource trafficSource,
+        TimeShifter timeShifter,
+        ThreadLocalTupleWriter tupleWriter,
+        Consumer<SourceTargetCaptureTuple> tupleObserver,
```
gregschohn, line 207: I'm still not sure why these aren't merged together into a single responsibility.  There's a lot for the replayer to manage.  If we can group responsibilities together, it makes this outer layer a lot simpler to read and reason with.

Also, why do you need an observer AND a writer?  It suspect that this patch could be smaller and more futureproof if we replaced the concrete class ThreadLocalTupleWriter with an interface with the necessary responsibilities.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/e2etests/TupleWriteBlockingBehaviorTest.java
```diff
@@ -0,0 +1,259 @@
...
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Timestamp;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Verifies that the ThreadLocalTupleWriter architecture correctly:
+ * <ol>
+ *   <li>Blocks Kafka commits until tuple futures are completed (durability confirmed)</li>
```
gregschohn: nit: commits are a general concept - if we switched out kafka for kinesis or something else, this test should still work.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/e2etests/TupleWriteBlockingBehaviorTest.java
```diff
@@ -0,0 +1,259 @@
...
+        final CountDownLatch allAccepted;
+
+        LatchedTupleSink(int expectedCount) {
+            this.allAccepted = new CountDownLatch(expectedCount);
+        }
+
+        @Override
+        public void accept(Map<String, Object> tupleMap, CompletableFuture<Void> future) {
+            heldFutures.add(future);
+            acceptCount.incrementAndGet();
```
gregschohn: countdownlatch has a getCount() getter.  You don't need this volatile value too.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/e2etests/TupleWriteBlockingBehaviorTest.java
```diff
@@ -0,0 +1,259 @@
...
+    /**
+     * Verifies that subsequent requests on the same connection are NOT blocked
+     * while tuple writes are pending.
+     *
+     * Strategy: use a LatchedTupleSink. If request processing were blocked on tuple
+     * completion, we'd never see all NUM_REQUESTS accepted. The fact that all tuples
+     * arrive (while none are completed) proves requests flow independently.
+     */
+    @Test
+    public void tupleWriteDoesNotBlockNextRequest() throws Throwable {
```
gregschohn, line 203: Thanks - this is a nice test.  Took me a while to process it, but it's nice that we can do such a test in 40 lines of code.

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/sink/GzipJsonLinesSinkTest.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        // Each onEndOfBatch rotates, so we get 2 data files (close() produces an empty file that's also valid)
+        var gzFiles = Files.list(tempDir).filter(p -> p.toString().endsWith(".log.gz")).sorted().toList();
+        assertTrue(gzFiles.size() >= 2, "Expected at least 2 files from rotation, got " + gzFiles.size());
+
+        int totalLines = 0;
+        for (var file : gzFiles) {
+            try (var reader = new BufferedReader(
+                new InputStreamReader(new GZIPInputStream(new FileInputStream(file.toFile()))))) {
```
gregschohn: these should be broken up to be exception-safe

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/sink/GzipJsonLinesSinkTest.java
```diff
@@ -0,0 +1,156 @@
...
+                String line;
+                while ((line = reader.readLine()) != null && !line.isEmpty()) {
+                    var parsed = MAPPER.readValue(line, Map.class);
+                    assertTrue(((String) parsed.get("connectionId")).startsWith("conn"));
```
gregschohn: it would be nice to also dump the connectionId into a set and check the size of the set at the end

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/TupleSink.java
```diff
@@ -0,0 +1,26 @@
...
+ * on commit; streaming sinks complete per-item as acks arrive.
+ *
+ * <p>Each instance is single-threaded — one per Netty event loop thread.</p>
+ */
+public interface TupleSink extends AutoCloseable {
+    /** Accept one tuple. The sink owns the future and must complete it when durability is confirmed. */
+    void accept(Map<String, Object> tupleMap, CompletableFuture<Void> future);
+
+    /** Hint that no more items are immediately available. Batch sinks should flush buffers. */
+    void onEndOfBatch();
```
gregschohn: This call doesn't make much sense for the replayer.  It looks like it's called after every message.  How do you know what's the end of a batch or not.  Does it matter anyway to your interfaces, especially if they're running synchonously?

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/TupleSink.java
```diff
@@ -0,0 +1,26 @@
...
+
+    /** Periodic callback even when no events arrive. Batch sinks should check time-based thresholds. */
+    void onIdle();
```
gregschohn: who calls this?

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/sink/GzipJsonLinesSinkTest.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    @Test
+    void onIdleFlushesUncommittedTuples() throws Exception {
+        try (var sink = new GzipJsonLinesSink(tempDir, 10 * 1024 * 1024, Duration.ofMinutes(10))) {
+            var f1 = new CompletableFuture<Void>();
+            sink.accept(makeTuple("conn1.0"), f1);
+            assertFalse(f1.isDone());
+
+            sink.onIdle();
```
gregschohn: onIdle seems weird to me.  Who is responsible for calling this?  Should this be something more like, periodicFlush or checkForFlush?

TrafficCapture/trafficReplayer/build.gradle
```diff
@@ -39,10 +40,10 @@ dependencies {
     implementation libs.log4j.slf4j2.impl
     implementation libs.slf4j.api
     implementation libs.disruptor
+    implementation libs.commons.compress
```
gregschohn: what do we use this for?

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/GzipJsonLinesSink.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    @Override
+    public void onEndOfBatch() {
+        if (pendingFutures.isEmpty()) {
+            return;
+        }
+        // Mountpoint S3 has write-once-then-close semantics: once fsync completes
+        // the S3 upload, further writes to the same file fail. Always rotate
+        // (finish + close + open new) so each fsync is the final write to its file.
```
gregschohn: this Gzip implementation is very strongly coupled to the files that it's writing to.

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/GzipJsonLinesSink.java
```diff
@@ -0,0 +1,149 @@
...
+            return;
+        }
+        completeAll();
+        openNewFile();
+    }
+
+    private void openNewFile() {
+        var timestamp = TIMESTAMP_FORMAT.format(Instant.now());
+        var seq = sequenceCounter.getAndIncrement();
+        var filename = String.format("tuples-%d-%s-%d.log.gz", threadIndex, timestamp, seq);
```
gregschohn, line 143: we should include some id for a replayerId - even if we only run one at a time, we run do restart them over time.

gregschohn, line 143: This directory is going to get very big - potentially millions of items, we should use some directory sharding.  I'd recommend just formatting the timestamp and doing it by day/hour.  That means you'd have to move the directory creation into this method.

gregschohn (commented):
I have two major comments.  
1) I'd like to see a mode where a user can write one tuple per file (another flag for max tuples per file=1 would work - maybe we can simulate that today w/ maxSize=1). 
2) I think it's worth keeping the S3 mount stuff that can be used for mounting RO FSs, which can be used as a pre-req for RFS - though maybe in a separate PR.  I'd want to see the replayer invoke S3 directly.  Ideally, this would be handled how we handle signing or better, with a library, ideally hot-loaded as a ServiceProvider.  Doing it directly will be simpler and easier to control/maintain.  Right now, there's a lot of complexity w/out any gain (let me know if I'm missing something).

I have significant reservations about writing data through the filesystem through the service - especially for something that's as high-load as this.  We don't have performance testing & it's not possible that doing writes to S3 through an intermediary would be _more_ efficient than if we were able to manage it directly.

The github page for mountpoint-s3 says
> Mountpoint for Amazon S3 is optimized for applications that need high read throughput to large objects, potentially from many clients at once, and to write new objects sequentially from a single client at a time.

We're not in that class - we're writing from multiple clients (soon) with multiple threads and not using it to read at all.  Giving devs a quasi-file interface (as it doesn't support posix operations) sets us up for future traps - how often will every dev and agent remember the rules of how FS ops are translated into S3 - how do they figure out that mapping (like how do i save these files to glacier?).  There are other concerns that I see within the documentation too - but even if it were perfect, it doesn't seem like the right fit for us here.

We need to take on more complexity and resources to setup the sidecars (including kernel interops and context switches).  At best, we'll have to monitor the performance of another system (the sidecar).  At worst, we'll be trying to figure out how to get it to do behave in ways that we need it to to meet our performance.

The only case where I could see us wanting to mount S3 would be if we had a legacy system that required a FS interface (like lucene).  If we need to make changes anyway, making the changes as directly as possible gives us the most control.

As the code is constructed right now, it has a lot of assumptions threaded through it - instead of using files to make the system more flexible, the overall change makes the modular boundaries more coupled.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -298,25 +298,6 @@ charts:
   strimzi-kafka-operator:
     version: 0.50.0
     repository: "https://strimzi.io/charts/"
-    values:
```
gregschohn, line 301: this seems reasonable to not include and to use defaults (I'd need to check git blame to see why they were there).  Is there a reason that these are being excised in this PR though?

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/migrationConsole.yaml
```diff
@@ -211,6 +211,31 @@ spec:
               value: "/root/schema/workflowMigration.schema.json"
             - name: MIGRATION_SAMPLE_CONFIG_PATH
               value: "/root/schema/sample.yaml"
+            {{- if .Values.defaultBucketConfiguration.create }}
```
gregschohn, line 214: should this also key off of conditionalPackageInstalls.aws-mountpoint-s3-csi-driver?

migrationConsole/Dockerfile
```diff
@@ -80,6 +80,16 @@ RUN dnf install -y  \
         && \
         dnf clean all
 
+# Install mount-s3 for direct S3 bucket mounting (requires SYS_ADMIN capability at runtime)
+RUN MOUNT_S3_VERSION="1.22.2" && \
```
gregschohn, line 84: Is there not an image with this on it already?  We're mounting it through a sidecar, right?

migrationConsole/start-console.sh
```diff
@@ -5,6 +5,23 @@ if [[ -f /root/loadServicesFromParameterStore.sh ]]; then
   /root/loadServicesFromParameterStore.sh
 fi
 
+# Mount S3 bucket if configured via environment variables
```
gregschohn, line 8: Follow-up from above - can we mount this as a sidecar?  That seems cleaner/more modular in a couple different dimensions

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -415,6 +415,10 @@ export const USER_REPLAYER_PROCESS_OPTIONS = z.object({
         .describe("Base64-encoded tuple transformer configuration." + TUPLE_TRANSFORMER_SUFFIX),
     tupleTransformerConfigFile: z.string().optional()
         .describe("Path to a JSON file containing tuple transformer configuration." + TUPLE_TRANSFORMER_SUFFIX + EXPERT_FILE_SUFFIX),
+    tupleMaxLagSeconds: z.number().default(60).optional()
```
gregschohn: let's pick a different word than lag since it's already used in a number of different ways.  tupleMaxBufferSeconds?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyDecodedHttpResponseConvertHandler.java
```diff
@@ -28,7 +28,10 @@ public NettyDecodedHttpResponseConvertHandler(
     @Override
     public void channelRead(@NonNull ChannelHandlerContext ctx, @NonNull Object msg) throws Exception {
         if (msg instanceof HttpResponse) {
-            httpTransactionContext.onHeaderParse();
+            // Null when called from ParsedHttpMessagesAsDicts (parse-only, no transformation context)
+            if (httpTransactionContext != null) {
```
gregschohn: same question as above/before about why would we NOT pass a context in here?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/datahandlers/http/NettyJsonBodyAccumulateHandler.java
```diff
@@ -112,7 +112,7 @@ public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception
                     } else {
                         payload.put(JsonKeysForHttpMessage.INLINED_JSON_BODY_DOCUMENT_KEY, parsedJsonObjects.get(0));
                     }
-                    if (!jsonAccumulator.hasPartialValues()) {
+                    if (!jsonAccumulator.hasPartialValues() && context != null) {
```
gregschohn: same questions

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/sink/ThreadLocalTupleWriter.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Per-thread sink registry that assigns each calling thread its own {@link TupleSink}.
+ *
+ * <p>Since each Netty event loop is single-threaded, the per-thread sink requires
+ * no synchronization.</p>
+ */
+@Slf4j
+public class ThreadLocalTupleWriter implements AutoCloseable {
+    private final ConcurrentHashMap<Long, TupleSink> sinks = new ConcurrentHashMap<>();
```
gregschohn: You should use FastThreadLocal here instead - here's a full class replacement
```
public class ThreadLocalTupleWriter implements AutoCloseable {
    private final List<TupleSink> sinks = new CopyOnWriteArrayList<>();
    private final AtomicInteger sinkIndexCounter = new AtomicInteger();
    private final IntFunction<TupleSink> sinkFactory;

    private final FastThreadLocal<TupleSink> threadLocalSink = new FastThreadLocal<>() {
        @Override
        protected TupleSink initialValue() {
            TupleSink sink = sinkFactory.apply(sinkIndexCounter.getAndIncrement());
            sinks.add(sink);   // register for close()
            return sink;
        }
    };

    public ThreadLocalTupleWriter(IntFunction<TupleSink> sinkFactory) {
        this.sinkFactory = sinkFactory;
    }

    public CompletableFuture<Void> writeTuple(SourceTargetCaptureTuple tuple, ParsedHttpMessagesAsDicts parsed) {
        var future = new CompletableFuture<Void>();
        threadLocalSink.get().accept(parsed.toTupleMap(tuple), future);
        return future;
    }

    @Override
    public void close() {
        sinks.forEach(sink -> {
            try { sink.close(); }
            catch (Exception e) { log.atError().setCause(e).setMessage("Error closing TupleSink").log(); }
        });
        sinks.clear();
        threadLocalSink.remove();
    }
}
```

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -659,6 +711,37 @@ private static void runReplayMode(Parameters params) throws Exception {
         }
     }
 
+    private static ThreadLocalTupleWriter createS3TupleWriterIfConfigured(Parameters params) {
+        if (params.tupleS3Bucket == null || params.tupleS3Bucket.isEmpty()) {
+            return null;
+        }
+        log.info("S3 tuple writing enabled — bucket={}, region={}, prefix={}",
+            params.tupleS3Bucket, params.tupleS3Region, params.tupleS3Prefix);
+        var s3ClientBuilder = S3AsyncClient.crtBuilder()
```
gregschohn, line 738: Eventually, we should make the tuple writer hot-loadable so that we can write tuples to S3, FS, other block storage, etc without needing a ton of jars laying around.

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/GzipJsonLinesSink.java
```diff
@@ -0,0 +1,169 @@
...
+import java.util.Map;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.zip.GZIPOutputStream;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.extern.slf4j.Slf4j;
+
+/**
+ * Writes tuples as gzip-compressed JSON lines to files compatible with Mountpoint S3.
```
gregschohn: update the comment, it isn't relevant ONLY for s3 mounts - would be relevant for any FS output, right

This isn't used in the current implementation, right?

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/S3TupleSink.java
```diff
@@ -0,0 +1,198 @@
...
+        this.maxFileSizeBytes = maxFileSizeBytes;
+        this.maxFileAge = maxFileAge;
+        this.maxTuplesPerFile = maxTuplesPerFile;
+        openNewBuffer();
+    }
+
+    @Override
+    public void accept(Map<String, Object> tupleMap, CompletableFuture<Void> future) {
+        try {
+            byte[] json = mapper.writeValueAsBytes(tupleMap);
```
gregschohn, line 88: let's do this using the streaming interface instead

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/S3TupleSink.java
```diff
@@ -0,0 +1,198 @@
...
+            rotate();
+        }
+        gzipOut = null;
+        byteBuffer = null;
+    }
+
+    private boolean shouldRotate() {
+        return uncompressedBytes >= maxFileSizeBytes
+            || (maxTuplesPerFile > 0 && tupleCount >= maxTuplesPerFile)
+            || Duration.between(fileOpenedAt, Instant.now()).compareTo(maxFileAge) >= 0;
```
gregschohn: i'm fine if we don't say maxBytes but change the nomenclature to rotateFilesGreaterThanMB or something like that

migrationConsole/start-console.sh
AndreKurait: We can, i experimented with that initially though ultimately went the binary approach to simplify the resources/requests 
image

gregschohn (approved):

PR #2670, opened 2026-04-09
[Fix] Treat expected=0 version pairs as skipped, not failed in test runner

jugal-chauhan:
### Description
Jenkins runs for `OS_1.3 → OS_2.19` were failing with "Tests failed (or no tests executed)" even though no tests were supposed to run. (Link to view failed Jenkins job is [here](https://migrations.ci.opensearch.org/job/main/job/main-k8s-matrix-test/74/))

- OS_1.3 is a valid source version but has no compatible tests against OS_2.19. 
- `conftest.py` was correctly reporting expected=0, passed=0, failed=0, but `test_runner.py` treated passed=0 as a failure 
unconditionally
- Additionally, conftest.py was incrementing failed and recording a "failed" result when no compatible tests were found, which compounded the issue

The bug had no test coverage. There was no test case for the expected=0 scenario in PR #2656 , which is the reason we did not catch this earlier.

#### Fix

1. Added a test test_zero_expected_tests_succeeds to reproduce the failure. It broke immediately, confirming the bug.
2. `test_runner.py` : skip the pass/fail check entirely when `expected==0 and failed==0`; also skip the "no tests executed" guard when all version pairs legitimately have `expected==0`.
3. `conftest.py` : record empty version pairs as "skipped" instead of "failed", and stop incrementing the failed counter.

### Issues Resolved
N/A

### Testing
- Added test_zero_expected_tests_succeeds = was failing before fix, passes after
- All 15 existing tests continue to pass
- Waiting on K8s-matirx-test to pass : https://migrations.ci.opensearch.org/job/pr-checks/job/pr-k8s-matrix-test/29/

### Check List

### DIFF
--- a/libraries/testAutomation/testAutomation/test_runner.py
+++ b/libraries/testAutomation/testAutomation/test_runner.py
@@ -297,22 +297,27 @@ class TestRunner:
                                              reuse_clusters=reuse_clusters,
                                              test_reports_dir=test_reports_dir)
                 test_reports.append(test_report)
-                tests_failed = test_report.summary.failed > 0 or test_report.summary.passed == 0
 
                 expected = test_report.summary.expected
-                if expected is not None and test_report.summary.passed != expected:
-                    logger.warning(f"Expected {test_report.summary.expected} tests but only "
-                                   f"{test_report.summary.passed} passed "
-                                   f"({test_report.summary.failed} failed)")
-                    tests_failed = True
-
-                if tests_failed:
-                    logger.warning(f"Tests failed (or no tests executed) for migrations "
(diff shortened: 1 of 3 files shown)


AndreKurait (approved):
For thought, this seems very convoluted in addition with https://github.com/opensearch-project/opensearch-migrations/pull/2656, I don't feel confident that we would catch the bug we initially had that no tests were running and we passed.

jugal-chauhan:
> For thought, this seems very convoluted in addition with #2656, I don't feel confident that we would catch the bug we initially had that no tests were running and we passed.

You're right it's convoluted. But we went from zero tests to 15, and the original bug is now directly tested!

PR #2671, opened 2026-04-09
Fix websocket drop failure on jenkins by replacing streaming with a background task with polling

jugal-chauhan:
### Description

All EKS integration pipelines (`eks-integ-test`, `eks-aoss-*`, `eks-byos-integ-test`) run pytest inside the migration console pod via a single long-lived Kubernetes WebSocket connection (`stream.stream(connect_get_namespaced_pod_exec)`). This connection is held open for the entire pytest run (10 minutes for AOSS tests, up to 15 hours for BYOS).

Kubernetes WebSocket connections can drop due to EKS API server rotation, network path instability, or transient infrastructure events. When this happens mid-test, Jenkins sees a `Connection to remote host was lost` error and marks the build as failed, even though the tests inside the pod may still be running successfully. This was causing repeated false failures on the AOSS pipelines this week.

#### Solution

Replace the single long-lived WebSocket with a fire-and-forget + poll pattern:

1. **`exec_background_cmd()`** — launches pytest in the background inside the pod via `nohup bash -c '... | tee {log_file}; echo ${PIPESTATUS[0]} > {exit_code_file}' &`. Returns immediately after a short exec (~1s).

2. **`poll_cmd_completion()`** — polls every 30s using short-lived execs. Tails the log file for Jenkins console visibility (using `wc -c` offset tracking to avoid reprinting), and checks for the exit code file to detect completion. Each poll exec is ~1s — immune to WebSocket drops regardless of test duration.

Jenkins console output looks nearly identical to before (periodic log chunks instead of real-time lines). Test pass/fail is still driven entirely by the test report, same as before.

#### Changes
- `k8s_service.py`: added `exec_background_cmd()` and `poll_cmd_completion()`
- `test_runner.py`: `run_tests()` now uses background exec + poll instead of a single streaming exec. Added a diagnostic warning if pytest exits non-zero but the test report shows no failures (catches cases where pytest crashes before writing results).

### Issues Resolved
N/A

### Testing
- On all EKS pipelines on `pr-checks/`

### Check List


libraries/testAutomation/testAutomation/k8s_service.py
```diff
@@ -116,6 +116,72 @@ def get_migration_console_pod_id(self) -> str:
...
+    def exec_background_cmd(self, command_list: List, log_file: str, exit_code_file: str) -> None:
+        """Launch a command in the background inside the migration console pod.
+
+        The command's stdout/stderr are tee'd to log_file so Jenkins can tail them.
+        On completion, the exit code is written to exit_code_file.
+        """
+        inner_cmd = " ".join(command_list)
+        wrapper = (
+            f"nohup bash -c '"
+            f"{inner_cmd} 2>&1 | tee {log_file}; "
```
AndreKurait: nit: I don't quite understand what we're doing with tee since we're only consuming the output via the file and not stdout

libraries/testAutomation/testAutomation/k8s_service.py
```diff
@@ -116,6 +116,72 @@ def get_migration_console_pod_id(self) -> str:
...
+        """
+        last_offset = 0
+        while True:
+            time.sleep(poll_interval)
+
+            # Tail new log lines for Jenkins console visibility
+            try:
+                tail_cmd = ["sh", "-c",
+                            f"wc -c < {log_file} 2>/dev/null && "
+                            f"tail -n {tail_lines} {log_file} 2>/dev/null"]
```
AndreKurait: I'm very confused how this works, with tail lines being a fixed value (e.g. 50) What happens if there are more than 50 lines since calling this last.

I'd prefer an approach where we do `tail -n +1 filename` which prints everything after the first line. e.g.

```
#!/usr/bin/env bash
file="$1"
start=1

while true; do
  new=$(tail -n +"$start" "$file")
  if [ -n "$new" ]; then
    printf '%s\n' "$new"
    n=$(printf '%s\n' "$new" | wc -l)
    start=$((start + n))
  fi
  sleep 2
done
```

libraries/testAutomation/testAutomation/k8s_service.py
jugal-chauhan: We do not need the Tee here

jugal-chauhan: It has been removed in the latest commit. I replaced it with a simple `sh -c` and  `> {log_file} 2>&1; echo $?`. This also eliminated the bash/PIPESTATUS dependency

libraries/testAutomation/testAutomation/k8s_service.py
jugal-chauhan: Thanks for pointing this out. I've modified the logic to follow the suggested approach here. The poller now tracks `next_line` and uses `tail -n +{next_line}` each poll cycle, so every line is printed exactly once regardless of output volume between polls

libraries/testAutomation/testAutomation/k8s_service.py
```diff
@@ -116,6 +116,78 @@ def get_migration_console_pod_id(self) -> str:
...
+
+            # Print new log lines since last poll
+            try:
+                resp = self.exec_migration_console_cmd(
+                    command_list=["tail", "-n", f"+{next_line}", log_file],
+                    unbuffered=False)
+                if resp and resp.strip():
+                    new_lines = resp.strip()
+                    print(new_lines)
+                    next_line += new_lines.count("\n") + 1
```
AndreKurait: This breaks on a partial line if the file ends in a partial line

AndreKurait: imagine a file that is just `a a ...` on the same line, apologies, this was kind of my bad suggestion for using tail.

AndreKurait: Using -c here instead of -n to make the boundary on characters should help

libraries/testAutomation/testAutomation/k8s_service.py
```diff
@@ -116,6 +116,77 @@ def get_migration_console_pod_id(self) -> str:
...
+            # Print new log bytes since last poll (no filtering/rewrapping for Jenkins parity)
+            try:
+                resp = self.exec_migration_console_cmd(
+                    command_list=["tail", "-c", f"+{next_byte + 1}", log_file],
+                    unbuffered=False)
+                if resp:
+                    print(resp, end="", flush=True)
+                    next_byte += len(resp.encode("utf-8"))
+            except Exception as e:
+                logger.debug(f"Log tail failed (may be transient): {e}")
```
AndreKurait: nit: I'd probably put this at Info level at lease

AndreKurait (approved):

libraries/testAutomation/testAutomation/k8s_service.py
jugal-chauhan: Addressed

PR #2672, opened 2026-04-09
Expand migration-console nodepool and exclude daemonsets from its node

AndreKurait:
Increase the dedicated migration-console Karpenter NodePool memory limit from 4Gi to 8Gi to allow larger `t`-category instances (e.g. `t3.large`).

Exclude fluent-bit and otel-collector daemonsets from the migration-console node via `nodeAffinity` — they serve no purpose there and waste memory.

Reduce migration-console pod memory from 2.5Gi to 2Gi since daemonset overhead is no longer a concern on this node.


gregschohn (commented):
Can you add a 'testing' spot to the PR description to confirm that you've tested this - then I'll approve

gregschohn (approved):
YOLO

PR #2677, opened 2026-04-10
Added Facet Comparison Support for KV Based Buckets

akshay2000:
## Add JSON Facets API support to FacetComparator

Extends `compareFacets()` to handle the JSON Facets API format alongside the existing flat format. Both are normalised to a common bucket map before shared comparison logic. No changes to callers.

Previously only supported flat facets:
```json
{"facet_counts": {"facet_fields": {"category": ["books", 10, "music", 5]}}}
```

Now also supports JSON facets:
```json
{"facets": {"count": 200000, "cat": {"buckets": [{"val": "books", "count": 100}, {"val": "music", "count": 50}]}}}
```

### Testing
Unit tests

### Check List

### DIFF
--- a/TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/reporting/FacetComparator.java
+++ b/TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/reporting/FacetComparator.java
@@ -2,6 +2,7 @@ package org.opensearch.migrations.transform.shim.reporting;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


niravpi:
Overall it looks good, just one follow up for my understanding.
compareFacets() now runs both compareFlatFacets() and compareJsonFacets() on every call. If a response has both facet_counts (flat) and facets (JSON) which can happen if someone uses both APIs in the same query, would we get duplicate comparison entries for the same logical facet? Or is that not a realistic scenario since the shim would only
produce one format?

akshay2000:
> Or is that not a realistic scenario since the shim would only
produce one format?

As of now shim only supports JSON facets. But regardless of that, if the user mentions same facets in two different APIs (and I don't know if both are supported by Solr in the same query) it will produce buckets in two different formats. And if that happens, we - as a reporting framework - should not assume the responsibility of de-duping those. So, yes the report will and should contain the duplicate keys.

By the way we don't need two different APIs for this. Nothing is stopping user from creating the same buckets under two different facets.

niravpi (approved):

AndreKurait (approved):

PR #2688, opened 2026-04-12
feat(solr-transforms): Add Input Validation for Supported/Unsupported Params, Null/Empty Handling & Fix URL + Decoding

niravpi:
### Summary

   Adds request-level input validation that detects unsupported Solr params, validates param types (integer, boolean, json), rejects null/empty values, and enforces pattern rules — all before endpoint transforms run. Also fixes the `+` as space
   URL decoding bug at the root cause — the GraalVM URLSearchParams polyfill now correctly decodes `+` as space per the WHATWG URL spec, eliminating the need for workarounds in application code.

   Additionally fixes a pre-existing bug where the response transform in dual-target mode received the post-transform request (with URI rewritten to `/_search`), preventing response transforms from detecting original Solr params like
   `cursorMark`. The response transform now receives a snapshot of the request before the request transform runs.

   ### Files Changed (21)

   **New:**
   - `features/validation.ts` — validation transform + types
   - `features/validation.test.ts` — 34 unit tests
   - `context.test.ts` — 11 unit tests for URL decoding, endpoint detection, collection extraction

   **Modified (TS):**
   - `context.ts`, `registry.ts`, `cursor-pagination.ts`, `cursor-pagination.test.ts`
   - `query-q.ts`, `field-list.ts`, `sort.ts`, `highlighting.ts`, `json-facets.ts`, `select-uri.ts`
   - `cases.testcase.ts`, `test-types.ts`

   **Modified (Java):**
   - `SolrTransformerProvider.java` — `+` decoding fix in URLSearchParams polyfill
   - `MultiTargetRoutingHandler.java` — response transform receives pre-transform request snapshot
   - `TestCaseDefinition.java`, `ShimTestFixture.java`, `TransformationShimE2ETest.java`
   - `URLSearchParamsPolyfillTest.java` — updated reference from `ShimMain` to `SolrTransformerProvider`

   ### Changes

   **Input Validation (`validation.ts`, `registry.ts`)**
   - Features self-declare supported params via `params[]`, `paramPrefixes[]`, and `paramRules[]`
   - Unsupported param detection with fail-fast error responses
   - Four param type validators: `integer`, `boolean`, `json`, `rejectPattern`
   - Pre-compiled regex patterns for reject rules

   **URLSearchParams Polyfill Fix (`SolrTransformerProvider.java`, `context.ts`)**
   - Root cause fix: polyfill now calls `.replace(/\+/g, ' ')` before `decodeURIComponent` per WHATWG URL spec
   - `+` decodes as space, `%2B` decodes as literal `+` — matches browser-native behavior
   - Removed the `replaceAll('+', '%20')` workaround from `parseParams()` in `context.ts` — no custom decoding logic needed
   - Removed per-feature `replaceAll('+', ' ')` workaround in cursor-pagination
   - Any code using `new URLSearchParams()` now gets correct `+` decoding automatically

   **Response Transform Request Context Fix (`MultiTargetRoutingHandler.java`)**
   - Response transform now receives a snapshot of the request taken after cursor rewrite but before the request transform rewrites the URI to `/_search`
   - Fixes cursor pagination in dual-target mode — response transform can now detect `cursorMark` and produce `nextCursorMark`
   - Generic fix that benefits all response transforms, not just cursor — any response transform can now inspect original Solr params
   - Cursor pagination now works correctly on all 3 sandbox modes (single 18080, dual-solr 18083, dual-os 18084)

   **Bug Fixes**
   - Removed duplicate `jsonFacets.request` in `registry.ts`
   - Integer validator rejects `10abc` (was silently accepted by `parseInt`)

   **E2E Error-Path Testing**
   - Java harness extended with `httpGetRaw()`, `expectedStatusCode`, `expectedErrorContains`
   - 13 validation e2e test cases covering unsupported params, invalid types, local params, empty values
   - E2E test for `+` decoding in sort params (`url-plus-decoding-in-sort`)
   - Java integration test for polyfill `+` decoding and `%2B` literal plus

   ### Testing

   | Layer | Result |
   |-------|--------|
   | Unit tests (vitest) | 23 files, 526 tests ✅ |
   | E2E integration (Solr 8 + 9) | All passed ✅ |
   | SonarQube | 0 issues in changed files ✅ |
   | Sandbox single mode (167 queries) | 88/167 pass, 79 correctly rejected ✅ |
   | Sandbox dual mode (167 queries) | 167/167 pass ✅ |

   ### Sandbox Results

   **Mainline baseline: 112/167 pass** — but ~30 of those return silently wrong results (filters ignored, facets missing, stats dropped).

   **Feature branch: 88/167 pass** — all 88 return correct results, zero silent failures. The `+` URL decoding fix resolved 28 additional failures (sort, cursor, boolean queries) compared to pre-fix.

   The 79 rejections break down as:
   - `fq` filter queries (30) — not yet implemented
   - Legacy `facet.*` API (18) — only `json.facet` supported
   - `defType`/dismax/edismax (5)
   - Wildcard/fuzzy/proximity (6) — parser gap
   - `stats`, `group`, `spellcheck`, `mlt`, `bf` (8)
   - Non-select endpoints (6) — stream, get, schema, admin
   - ISO date colons in `q` (2) — parser gap
   - Local params `{!...}` (2)
   - Function sort, nested facet `avg()` (2)

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/cases.testcase.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/cases.testcase.ts
@@ -646,4 +646,144 @@ export const testCases: TestCase[] = [
       { path: '$.highlighting[*][*][*]', rule: 'regex', expected: '.*<em>.*</em>.*', reason: 'Fragment text may differ between Solr and OpenSearch highlighters' },
     ],
   }),
+
+  // ───────────────────────────────────────────────────────────
+  // Validation tests — error-path (expectedStatusCode)
+  // ───────────────────────────────────────────────────────────
+
+  solrTest('validation-unsupported-param-fq', {
+    description: 'Unsupported param fq should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&fq=title:test&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-unsupported-param-facet', {
+    description: 'Unsupported legacy facet params should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&facet=true&facet.field=title&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-unsupported-param-defType', {
+    description: 'Unsupported defType param should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=test&defType=dismax&qf=title&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-invalid-rows-non-numeric', {
+    description: 'Non-numeric rows should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&rows=abc&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-invalid-start-non-numeric', {
+    description: 'Non-numeric start should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&start=xyz&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-invalid-hl-boolean', {
+    description: 'Invalid boolean value for hl should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&hl=yes&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-local-params-in-q', {
+    description: 'Local params syntax in q should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=' + encodeURIComponent('{!dismax qf=title}hello') + '&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-invalid-sort-no-direction', {
+    description: 'Sort without direction should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&sort=price&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('url-plus-decoding-in-sort', {
+    description: 'Sort with + as space (e.g. sort=price+asc) should decode correctly and return sorted results',
+    documents: [
+      { id: '1', title: 'alpha', price: 30 },
+      { id: '2', title: 'beta', price: 10 },
+      { id: '3', title: 'gamma', price: 20 },
+    ],
+    requestPath: '/solr/testcollection/select?q=*:*&sort=price+asc&wt=json',
+    solrSchema: {
+      fields: {
+        title: { type: 'text_general' },
+        price: { type: 'pfloat' },
+      },
+    },
+    opensearchMapping: {
+      properties: {
+        id: { type: 'keyword' },
+        title: { type: 'text' },
+        price: { type: 'float' },
+      },
+    },
+    assertionRules: SOLR_INTERNAL_RULES,
+  }),
+
+  solrTest('validation-invalid-json-facet', {
+    description: 'Malformed json.facet should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&json.facet=' + encodeURIComponent('{bad json}') + '&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-local-params-in-fl', {
+    description: 'Local params syntax in fl should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&fl=' + encodeURIComponent('id,{!func}div(price,2)') + '&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-empty-rows', {
+    description: 'Empty rows value should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&rows=&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-empty-start', {
+    description: 'Empty start value should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    requestPath: '/solr/testcollection/select?q=*:*&start=&wt=json',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
+  }),
+
+  solrTest('validation-valid-request-passes', {
+    description: 'Valid request with supported params should pass through normally',
+    documents: [{ id: '1', title: 'test document' }],
+    requestPath: '/solr/testcollection/select?q=*:*&rows=10&start=0&fl=id,title&wt=json',
+    solrSchema: {
+      fields: {
+        title: { type: 'text_general' },
+      },
+    },
+    opensearchMapping: {
+      properties: {
+        title: { type: 'text' },
+      },
+    },
+  }),
 ];
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/query-q.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/query-q.ts
@@ -10,7 +10,16 @@
  */
 import type { MicroTransform } from '../pipeline';
 import type { RequestContext, JavaMap } from '../context';
+import type { ParamRule } from './validation';
 import { translateQ } from '../query-engine/orchestrator/translateQ';
+
+/** Solr query params this feature handles. */
+export const params = ['q', 'rows', 'start', 'q.op'];
+export const paramRules: ParamRule[] = [
+  { name: 'rows', type: 'integer' },
+  { name: 'start', type: 'integer' },
+  { name: 'q', type: 'rejectPattern', pattern: String.raw`^\{!`, reason: 'Local params ({!...}) syntax in q is not supported' },
+];
 
 export function parseSolrQuery(q: string): JavaMap {
   if (!q || q === '*:*') return new Map([['match_all', new Map()]]);
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/sort.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/sort.ts
@@ -11,7 +11,13 @@
  */
 import type { MicroTransform } from '../pipeline';
 import type { RequestContext } from '../context';
+import type { ParamRule } from './validation';
 
+/** Solr query params this feature handles. */
+export const params = ['sort'];
+export const paramRules: ParamRule[] = [
+  { name: 'sort', type: 'rejectPattern', pattern: String.raw`\{!`, reason: 'Local params ({!...}) syntax in sort is not supported' },
+];
 function parseSortClause(clause: string): Map<string, any> {
   const parts = clause.trim().split(/\s+/);
   if (parts.length < 2 || !parts[0]) {
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
@@ -1,0 +1,119 @@
+/**
+ * Validation — fail fast on null/empty inputs and unsupported Solr features.
+ *
+ * Runs as a global request transform (before endpoint-specific transforms).
+ * All validation rules are generic and injected by registry.ts via initValidation().
+ *
+ * Features declare:
+ *   - params / paramPrefixes — for supported param discovery
+ *   - paramRules — for param type validation (integer, boolean, json)
+ */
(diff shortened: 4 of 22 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
```diff
@@ -20,9 +21,40 @@ import * as highlighting from './features/highlighting';
...
+
+/**
+ * All feature modules — used to auto-aggregate supported params and rules.
+ * hitsToDocs and aggsToFacets are response-only transforms with no request
+ * params, so they don't need to be listed here for validation discovery.
+ */
+const FEATURE_MODULES = [
+  selectUri, queryQ, cursorPagination, fieldList,
+  sort, jsonFacets, highlighting,
+] as Array<{ params?: string[]; paramPrefixes?: string[]; paramRules?: ParamRule[] }>;
```
nagarajg17: Can we have a struct or interface for this to group it together? `params?: string[]; paramPrefixes?: string[]; paramRules?: ParamRule[]`

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
```diff
@@ -20,9 +21,40 @@ import * as highlighting from './features/highlighting';
...
+import * as validation from './features/validation';
+import { initValidation } from './features/validation';
+import type { ParamRule } from './features/validation';
+
+/**
+ * All feature modules — used to auto-aggregate supported params and rules.
+ * hitsToDocs and aggsToFacets are response-only transforms with no request
+ * params, so they don't need to be listed here for validation discovery.
+ */
+const FEATURE_MODULES = [
```
nagarajg17: Any possibility to read from registry directly? because all transformations are registered there and is source of truth for that

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
```diff
@@ -20,9 +21,40 @@ import * as highlighting from './features/highlighting';
...
+  selectUri, queryQ, cursorPagination, fieldList,
+  sort, jsonFacets, highlighting,
+] as Array<{ params?: string[]; paramPrefixes?: string[]; paramRules?: ParamRule[] }>;
+
+/** Common Solr params that don't need translation but are valid. */
+const COMMON_PARAMS = ['wt', 'indent', 'echoParams', 'df'];
```
nagarajg17, line 52: not sure if we are supporting indent, echoParams now

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
```diff
@@ -33,7 +65,6 @@ export const requestRegistry: TransformRegistry<RequestContext> = {
       fieldList.request, // fl=... → _source
       highlighting.request, // hl=true → highlight block
       sort.request, // sort=... → sort DSL
-      jsonFacets.request, // json.facet → aggs
```
nagarajg17, line 36: nit: Intended removal?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export interface ParamRule {
+  name: string;
+  type: ParamType;
+  /** For 'rejectPattern': regex pattern that should NOT match. */
+  pattern?: string;
+  /** For 'rejectPattern': human-readable reason shown in error. */
+  reason?: string;
+}
+
```
nagarajg17, line 23: nit: Can we add comment and ex for each of these params, prefixes, rules

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
```diff
@@ -0,0 +1,106 @@
...
+export type ParamType = 'integer' | 'boolean' | 'json' | 'rejectPattern';
+
+export interface ParamRule {
+  name: string;
+  type: ParamType;
+  /** For 'rejectPattern': regex pattern that should NOT match. */
+  pattern?: string;
+  /** For 'rejectPattern': human-readable reason shown in error. */
+  reason?: string;
```
nagarajg17: These 2 seems to mainly for unsupported features and above 2 for data checks (null/empty). Should we divide into 2 interfaces? Both seems disjoint. Typescript allows variable to have multiple types  `int|string`

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
```diff
@@ -0,0 +1,106 @@
...
+    // Param type checks (auto-discovered from features)
+    for (const rule of paramRules) {
+      const val = ctx.params.get(rule.name);
+      if (val != null) {
+        const error = TYPE_VALIDATORS[rule.type](val, rule.name, rule);
+        if (error) throw new Error(`Validation failed: ${error}`);
+      }
+    }
+
+    // Unsupported param detection
```
nagarajg17: Should this come before paramsRules check?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/query-q.ts
```diff
@@ -10,8 +10,17 @@
...
+import type { ParamRule } from './validation';
 import { translateQ } from '../query-engine/orchestrator/translateQ';
 
+/** Solr query params this feature handles. */
+export const params = ['q', 'rows', 'start', 'q.op'];
+export const paramRules: ParamRule[] = [
+  { name: 'rows', type: 'integer' },
+  { name: 'start', type: 'integer' },
+  { name: 'q', type: 'rejectPattern', pattern: String.raw`^\{!`, reason: 'Local params ({!...}) syntax in q is not supported' },
+];
```
nagarajg17, line 22: - For AST which parser generates via PEG we would need validation for that
- Do we plan to apply regex on q param or AST? because if we apply on q param, AST generation done before that is wasteful operation
- For q param having various types of inputs? how to validate that
- Other unsupported features like function queries, fuzzy etc, will that be in follow up PR?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/sort.ts
```diff
@@ -11,7 +11,13 @@
  */
 import type { MicroTransform } from '../pipeline';
 import type { RequestContext } from '../context';
+import type { ParamRule } from './validation';
 
+/** Solr query params this feature handles. */
+export const params = ['sort'];
+export const paramRules: ParamRule[] = [
+  { name: 'sort', type: 'rejectPattern', pattern: String.raw`\{!`, reason: 'Local params ({!...}) syntax in sort is not supported' },
```
nagarajg17, line 19: nit: Local params seems to repeated in all files. Do we need to maintain common list or atleast ref in different files from common list

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/netty/MultiTargetRoutingHandler.java
```diff
@@ -225,8 +225,10 @@ private DispatchResult dispatchAll(
                 ? (TargetDispatchContext) requestCtx.createTargetDispatchContext(name)
                 : null;
 
-            // Response transform needs per-target URI (with rewritten cursorMark)
-            Map<String, Object> responseRequestMap = hasCursorMark ? targetRequestMap : requestMap;
+            // Snapshot the request before the request transform rewrites the URI (e.g., to /_search).
+            // The response transform needs the original Solr URI to detect params like cursorMark, sort, etc.
+            Map<String, Object> responseRequestMap = target.requestTransform() != null
```
nagarajg17, line 261: Is this for `+` issue fix? Do we need this changes?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/cases.testcase.ts
```diff
@@ -646,4 +646,144 @@ export const testCases: TestCase[] = [
       { path: '$.highlighting[*][*][*]', rule: 'regex', expected: '.*<em>.*</em>.*', reason: 'Fragment text may differ between Solr and OpenSearch highlighters' },
     ],
   }),
+
+  // ───────────────────────────────────────────────────────────
+  // Validation tests — error-path (expectedStatusCode)
+  // ───────────────────────────────────────────────────────────
+
+  solrTest('validation-unsupported-param-fq', {
```
nagarajg17, line 654: nit: We can separate out these new tests into separate file

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
niravpi: Ack. Extracted a `FeatureModule` interface and replaced the inline `as Array<{...}>` cast with `FeatureModule[]`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
niravpi: The registry holds `MicroTransform` objects (`{ name, match, apply }`) — these don't carry param metadata. The `params`, `paramPrefixes`, and `paramRules` are module-level exports, not part of the transform interface. Merging them into `MicroTransform` would couple validation concerns into the transform contract. The `FEATURE_MODULES` array lives right next to the registry in the same file, so the risk of them going out of sync is low. If we find this becoming a maintenance issue as more features are added, we can revisit.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
niravpi: Yes, intentional. There was a duplicate `jsonFacets.request` entry in the request pipeline this PR removes it. Mentioned in the PR description under Bug Fixes.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
niravpi: Ack. Added JSDoc block with `@example` for all 4 types (integer, boolean, json, rejectPattern) and per-field documentation with examples.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
niravpi: Considered this, but the interface is only 4 fields with 2 optional. A discriminated union (`DataRule | RejectRule`) would add more boilerplate in every feature file for minimal type safety gain. The validation logic already switches on `type` and TypeScript's control flow narrowing handles it correctly. The JSDoc additions now clearly document which fields belong to which type. If `ParamRule` grows more types in the future, we can revisit splitting it.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
niravpi: Good catch. Done, reordered so unsupported param detection runs first (fail fast on unknown params), then paramRules type checks run on known params.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/query-q.ts
niravpi: The `rejectPattern` on `q` only catches local params `{!...}`  this is a syntax-level check that runs before parsing. It's intentionally separate from AST validation. The AST parser handles query structure validation (field queries, boolean operators, ranges, etc.) and already rejects unsupported constructs like wildcards and fuzzy queries at the transform stage. Unsupported query types like function queries can be addressed in feature specific PRs as the parser is extended they'll be rejected at the AST level, not via regex.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/sort.ts
niravpi: Each feature declares its own `rejectPattern` because the error reason is param-specific (`"in sort"`, `"in q"`, `"in fl"`). Could extract a helper like `rejectLocalParams('sort')` but the duplication is minimal (3 occurrences) and each is a single line. Can refactor if it grows.

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/netty/MultiTargetRoutingHandler.java
niravpi: No, this is a separate bug fix. In dual-target mode, the response transform was receiving the post-transform request (URI already rewritten to `/_search`), so it couldn't detect original Solr params like `cursorMark` or `sort`. This change snapshots the request before the request transform runs, so the response transform sees the original Solr URI. This fixes cursor pagination in dual-target mode and benefits any future response transform that needs to inspect original Solr params.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/cases.testcase.ts
niravpi: Ack, Can be done as a follow-up cleanup if the file gets too large.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
niravpi: These params are allowlisted so validation doesn't reject them as unsupported — updated the comment to be explicit about each one: 
- `df` — actively supported, the query parser uses it as the default field for bare terms
- `wt` — only `json` is returned; other formats (xml, csv) are not translated. Clients commonly send `wt=json` and rejecting it would break existing queries
- `indent` — pretty-print flag, ignored by the shim
- `echoParams` — param echo in response header, not implemented

`indent` and `echoParams` don't affect query correctness they're cosmetic params that Solr clients include out of habit. Rejecting them would force clients to strip these params for no functional benefit.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/query-q.ts
nagarajg17: As we discussed offline, lets give a thought on AST or q param validation in a separate PR

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/validation.ts
niravpi: Ass discussed, updated to have two interface

nagarajg17 (approved):

jugal-chauhan (approved):
LGTM! Thanks for bringing in these changes

PR #2689, opened 2026-04-13
Wired Up Reporting to Shim

akshay2000:
### Description
Wires the existing `MetricsReceiver` and `ReportingSink` into `MultiTargetRoutingHandler` so every dual-target request automatically produces a `ValidationDocument` with hit-count drift, latency delta, and optional request/response bodies. Reporting is enabled via a `--reportingConfig` JSON CLI flag and is wired into the dev sandbox dual-target shims. All new constructor params default to null for backward compatibility.

### Testing
- Unit tests

### Check List


TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/ShimMain.java
```diff
@@ -317,6 +367,44 @@ static Map<String, Object> extractBindings(String configStr) {
...
+            if (!config.containsKey("outputDir") || !(config.get("outputDir") instanceof String)) {
+                throw new ParameterException(
+                    "--reportingConfig requires \"outputDir\" key with a string value");
+            }
+            return config;
+        } catch (com.fasterxml.jackson.core.JsonProcessingException e) {
+            throw new ParameterException("Invalid --reportingConfig JSON: " + e.getMessage());
+        }
+    }
+
```
niravpi: The config format `[{"FileSystemReportingSink": {...}}]` implies pluggability, but `parseReportingConfig` silently ignores the provider key and `createReportingSink` always creates a `FileSystemReportingSink`.
If a user passes `[{"AnyOtherValue": {"outputDir": "/tmp/reports"}}]`, it would silently create a filesystem sink instead of erroring out.

Is this something intentional ?

akshay2000: That was a leftover from an older config format that had only one thing. Corrected.

niravpi (approved):

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/netty/MultiTargetRoutingHandler.java
```diff
@@ -228,18 +259,37 @@ private DispatchResult dispatchAll(
             // Response transform needs per-target URI (with rewritten cursorMark)
             Map<String, Object> responseRequestMap = hasCursorMark ? targetRequestMap : requestMap;
             futures.put(name, dispatchToTarget(target, targetRequestMap, responseRequestMap, dispatchCtx));
+
+            collectTransformData(target, name, targetRequestMap,
```
AndreKurait, line 263: How is metricsMap not empty at this point? I would expect the responses are async and we have to wait on futures before the responses are populated

akshay2000: The metrics are coming from the request translations. For example, if a request translation notices some limitation, (like offsets not being supported by OpenSearch inside aggregation buckets) the transform will emit a metric. So the metrics will be available as soon as the request transform is run and we come back from GraalVM.

AndreKurait: The pr overview here says
```
so every dual-target request automatically produces a ValidationDocument 
with hit-count drift, latency delta, and optional request/response bodies.
``` 

Is this accurate?

akshay2000: That is correct. (Obviously the user must pass `--reportingConfig` to enable this.) The list is not exhaustive. Do you see an accuracy issue here?

AndreKurait (approved):

AndreKurait (approved):

PR #2717, opened 2026-04-14
Default capture proxy to TLS with self-signed cert

jugal-chauhan:
### Description
Follows the same secure-by-default pattern established for Kafka in #2412. When a user configures a capture proxy without specifying a TLS block, the config transformer now injects a self-signed cert-manager TLS configuration using the migrations-ca ClusterIssuer. Users can also explicitly opt out with `tls: {mode: "plaintext"}`.

##### Schema (userSchemas.ts):

- Added plaintext mode to PROXY_TLS_CONFIG discriminated union. This will now be the only way to run the proxy without TLS

##### Config transformer (migrationConfigTransformer.ts):

- `defaultProxyTlsConfig()` will generate a certManager TLS config with `*.svc.cluster.local` + `proxy name as SANs`
- During normalization, absent TLS (with no legacy sslConfigFile) gets the default injected
- `plaintext` mode is stripped before passing to Argo, preserving existing HTTP behavior

##### Helm (values.yaml, issuerConfigmap.yaml, migrationConsole.yaml):

- Issuer name/kind exposed as configurable values with `migrations-ca` / `ClusterIssuer` defaults
- ConfigMap + env var injection into migration-console pod so the transformer can read the issuer config at runtime
- Hardcoded fallback in the transformer ensures it works in tests and local dev without the configmap

### Impact of these changes

- Existing proxy configs without tls will now get TLS. This is intentional (secure-by-default)
- Anyone currently relying on plaintext proxy behavior must add `tls: {mode: "plaintext"}` to preserve the old behavior
- `sslConfigFile` is respected. If the legacy `sslConfigFile` option is set, no default TLS is injected (they're mutually exclusive per the existing schema validation)
- Namespace in SANs: Uses `*.svc.cluster.local` rather than `*.ma.svc.cluster.local` to avoid hardcoding the namespace
- Requires cert-manager: The default TLS injection assumes cert-manager is installed and the migrations-ca ClusterIssuer exists (created by the selfSignedClusterIssuer helm hook). Deployments without cert-manager that use proxies will need `tls: {mode: "plaintext"}`

### Issues Resolved
[MIGRATIONS-3029](https://opensearch.atlassian.net/browse/MIGRATIONS-3029)

### Testing
- Updated existing proxy endpoint test (now expects `https://` and `allowInsecure: true`)
- Added `traffic-plaintext-tls.json` valid fixture for the plaintext opt-out path
- Refreshed argo schema snapshots

### Check List


orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
```diff
@@ -197,6 +197,22 @@ function normalizeKafkaClusterConfig(
...
 
+function defaultProxyTlsConfig(proxyName: string) {
+    return {
+        mode: "certManager" as const,
+        issuerRef: {
+            name: process.env.PROXY_DEFAULT_ISSUER_NAME || "migrations-ca",
+            kind: (process.env.PROXY_DEFAULT_ISSUER_KIND || "ClusterIssuer") as "ClusterIssuer" | "Issuer",
+        },
+        dnsNames: [
+            "*.svc.cluster.local",
```
AndreKurait: I don't think this works, we need the namespace in the path here since IIRC * only goes one level in SAN

jugal-chauhan: We can use 
```
dnsNames: [                                                                                                  
    proxyName,                                                                                               
    `${proxyName}.ma`,                                                                                       
    `${proxyName}.ma.svc.cluster.local`,                                                                     
],   
   ```

The alternative is to just use the short name and let clients connect with `allow_insecure: true` (which the transformer already sets when TLS is present). Since this is a self signed CA, clients won't trust it regardless without either `allow_insecure` or explicitly trusting the CA

Comment (jugal-chauhan) [reply orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts]: Can you suggest an approach in which we do not have to hardcode `ma` namespace ?

Comment (AndreKurait) [reply orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts]: Actually, we need to also include the NLB domain name (or trust all) in AWS

Comment (jugal-chauhan) [reply orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts]: Included in the latest commit

**Comment by jugal-chauhan** [comment]
This PR will fail on cdc tests until #2721

File: orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts [hunk 2]
```diff
@@ -197,6 +197,28 @@ function normalizeKafkaClusterConfig(
...
+        `${proxyName}.ma.svc.cluster.local`,
+    ];
+    if (awsRegion) {
+        dnsNames.push(`*.elb.${awsRegion}.amazonaws.com`);
+    }
+    return {
+        mode: "certManager" as const,
+        issuerRef: {
+            name: process.env.PROXY_DEFAULT_ISSUER_NAME || "migrations-ca",
+            kind: (process.env.PROXY_DEFAULT_ISSUER_KIND || "ClusterIssuer") as "ClusterIssuer" | "Issuer",
```
AndreKurait: Do these need to be configurable through an environment variable?

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
```diff
@@ -206,9 +228,32 @@ function normalizeTrafficConfig(traffic: InputConfig["traffic"]): InputConfig["t
...
     for (const [key, proxy] of Object.entries(traffic.proxies ?? {})) {
-        normalizedProxies[key] = proxy.kafkaTopic === ""
+        let normalized = proxy.kafkaTopic === ""
             ? (({kafkaTopic, ...rest}) => rest)(proxy)
             : proxy;
+
+        // Secure-by-default: inject self-signed TLS when no TLS config is specified
+        // and no legacy sslConfigFile is in use. Users can opt out with tls.mode: "plaintext".
+        if (!normalized.proxyConfig?.tls && !normalized.proxyConfig?.sslConfigFile) {
+            console.error(`TLS was auto-configured for '${key}' (secure-by-default). Use tls.mode: "plaintext" to opt out.`);
```
AndreKurait: Does console.error work here? I thought we had a different syntax to throw a specific zod error with a message

AndreKurait: Maybe for this we need to set a zod message instead of console.error here, Regardless, this should be info instead

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
jugal-chauhan: The fallbacks exist so the happy path works without configuration. Removing them creates a hard dependency on the ConfigMap existing

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
jugal-chauhan: Zod messages are only for validation errors. This is a successful normalization (accepting valid input and enhancing it). There's no Zod construct for "this is valid, but here's an FYI."

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
jugal-chauhan: I have dropped the configmap. THe defaults are now inline in `orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts`

AndreKurait (approved):

PR #2722, opened 2026-04-15
Limit crane copy parallelism to 4 concurrent jobs in ECR mirroring

AndreKurait:
## Summary

The `mirror_images_to_ecr` function in `mirrorToEcr.sh` previously launched all `crane copy` operations as background jobs simultaneously with no concurrency limit. This was observed to cause OOM (out of memory) failures when running in AWS CloudShell, and could also overwhelm ECR API rate limits when mirroring many images.

## Changes

Adds a semaphore pattern that caps concurrent `crane copy` operations at 4. Before launching each background job, the loop checks the number of running background processes (`jobs -rp`) and blocks with `wait -n` if the limit is reached.

## Test Plan

- Manual testing of the mirroring script with a representative image list
- Verified the throttling logic works correctly with bash job control


jugal-chauhan (approved):
LGTM

PR #2724, opened 2026-04-15
Add wildcard and fuzzy query translation rules.

Shivani-techno:
### Description
Adds wildcard and fuzzy query translation support to the Solr-to-OpenSearch query engine.

Previously, `fieldRule` threw errors when encountering wildcard (`*`, `?`) or fuzzy (`~`, `~N`) patterns in field:value queries. This PR replaces those errors with proper OpenSearch query DSL translation:

- `title:te?t` → `{"wildcard": {"title": {"value": "te?t"}}}`
- `title:test*` → `{"wildcard": {"title": {"value": "test*"}}}`
- `title:roam~` → `{"fuzzy": {"title": {"value": "roam"}}}`
- `title:roam~2` → `{"fuzzy": {"title": {"value": "roam", "fuzziness": 2}}}`

No parser or AST changes needed — wildcards and fuzzy suffixes are already parsed as part of `FieldNode.value` by the PEG grammar. The change is entirely within `fieldRule.ts`.

### Issues Resolved

Resolves the TODO items in `fieldRule.ts`:
- "Add support for wildcard queries"
- "Add support for fuzzy queries"

### Testing

- 21 unit tests in `fieldRule.test.ts` (5 wildcard, 4 fuzzy, 3 non-fuzzy tilde edge cases, 2 boost compatibility, plus existing match/exists/leaf tests)
- All 483 tests in the SolrTransformations test suite pass
- Output uses expanded Map form compatible with `boostRule` for boost injection

### Check List


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -1,69 +1,66 @@
...
-    console.error(msg);
-    throw new Error(msg);
+  // Fuzzy search: field:roam~ or field:roam~2 → fuzzy query
+  // Check fuzzy before wildcard since ~ is more specific
+  const fuzzyMatch = FUZZY_PATTERN.exec(value);
+  if (fuzzyMatch) {
+    const term = fuzzyMatch[1];
+    const distance = fuzzyMatch[2];
+    const fuzzyParams = new Map<string, any>([['value', term]]);
+    if (distance) {
```
niravpi: Nit: `if (distance !== '')` reads better than `if (distance)`. It works today because `"0"` is truthy in JS, but the intent ("was a digit present?") is clearer with an explicit empty-string check.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -1,69 +1,66 @@
...
+      fuzzyParams.set('fuzziness', parseInt(distance, 10));
```
niravpi: AFAIU from the docs, OpenSearch only accepts fuzziness 0, 1, 2, or `"AUTO"`. A query like `roam~9` would produce `{"fuzziness": 9}` which OpenSearch rejects at query time. Should we consider either validating here (Solr also only allows 0-2) or adding a comment that we intentionally let OpenSearch handle the rejection.

niravpi:
Unit tests cover `fieldRule` in isolation, but there's no test verifying the full round-trip: PEG parser → AST → transformer → DSL. The parser producing correct `FieldNode.value` for wildcard/fuzzy patterns is assumed but unverified. #2688 established the e2e framework in `cases.testcase.ts` adding a few positive cases (`title:te?t`, `title:test*`, `title:roam~`, `title:roam~2`) would close this gap and verify against real Solr.

niravpi:
#2688's sandbox results categorize "Wildcard/fuzzy/proximity (6)" as expected rejections (part of the 79 rejection count). Once this PR merges, those queries will start succeeding, can you please try to run the sandbox validation once and see if Wildcard/fuzzy query succeed ?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
Shivani-techno: Done,  updated to if (distance !== '').

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
Shivani-techno: Done, added validation to clamp fuzziness to max 2 using Math.min(fuzziness, 2), matching both Solr and OpenSearch limits. Also added a test case for roam~9 verifying it produces fuzziness: 2.

niravpi:
Looks like integration test comment is pending can you please respond back to the same?
Also did we tried validating via sandbox validation ?

niravpi:
Based on the code seems like the fuzzy logic in `fieldRule` will not be used as  `~` is not in the PEG grammar's `valueChars` (`$[a-zA-Z0-9._*#$@?/]+`), so `title:roam~2` fails at the parser before `fieldRule` ever sees it. Unit tests pass because they construct `FieldNode` manually, bypassing the parser.
Can you please help to confirm this behaviour once ?

Shivani-techno:
> Unit tests cover `fieldRule` in isolation, but there's no test verifying the full round-trip: PEG parser → AST → transformer → DSL. The parser producing correct `FieldNode.value` for wildcard/fuzzy patterns is assumed but unverified. [#2688](https://github.com/opensearch-project/opensearch-migrations/pull/2688) established the e2e framework in `cases.testcase.ts` adding a few positive cases (`title:te?t`, `title:test*`, `title:roam~`, `title:roam~2`) would close this gap and verify against real Solr.

Added 4 e2e test cases in wildcard-fuzzy-query.testcase.ts following the framework from #2688: title:te?t, title:search*, title:roam~, and title:roam~2. These verify the full PEG parser → AST → transformer → DSL round-trip. Also found and fixed a gap in the PEG grammar — ~ was missing from valueChars, which caused fuzzy patterns to fail at parse time.

Shivani-techno:
> [#2688](https://github.com/opensearch-project/opensearch-migrations/pull/2688)'s sandbox results categorize "Wildcard/fuzzy/proximity (6)" as expected rejections (part of the 79 rejection count). Once this PR merges, those queries will start succeeding, can you please try to run the sandbox validation once and see if Wildcard/fuzzy query succeed ?

Ran sandbox validation — wildcard and fuzzy queries now succeed through the shim.

Shivani-techno:
> Looks like integration test comment is pending can you please respond back to the same? Also did we tried validating via sandbox validation ?

Yes, both done.

Shivani-techno:
> Based on the code seems like the fuzzy logic in `fieldRule` will not be used as `~` is not in the PEG grammar's `valueChars` (`$[a-zA-Z0-9._*#$@?/]+`), so `title:roam~2` fails at the parser before `fieldRule` ever sees it. Unit tests pass because they construct `FieldNode` manually, bypassing the parser. Can you please help to confirm this behaviour once ?

Fixed in the latest commit by adding ~ to the grammar's valueChars rule: $[a-zA-Z0-9._*#$@?/~]+. Confirmed working end-to-end via sandbox — fuzzy queries now parse correctly and produce the expected DSL.

Shivani-techno:
> > Based on the code seems like the fuzzy logic in `fieldRule` will not be used as `~` is not in the PEG grammar's `valueChars` (`$[a-zA-Z0-9._*#$@?/]+`), so `title:roam~2` fails at the parser before `fieldRule` ever sees it. Unit tests pass because they construct `FieldNode` manually, bypassing the parser. Can you please help to confirm this behaviour once ?

Fixed in the latest commit by adding ~ to the grammar's valueChars rule: `$[a-zA-Z0-9._*#$@?/~]+`. Confirmed working end-to-end via sandbox — fuzzy queries now parse correctly and produce the expected DSL.

The double backticks (``) around the regex prevent GitHub from interpreting # and $ as special characters.

niravpi (approved):

PR #2737, opened 2026-04-15
feat(solr-transforms): Add single document update transform for /upda…

niravpi:
### Description

   Adds single document update support to the Solr-to-OpenSearch translation shim. Translates Solr's `/update/json/docs` endpoint to OpenSearch's Index API (`PUT /_doc/{id}`), with both request and response transforms so the client sees a Solr-compatible response.

   **Request transform:**
   - `POST /solr/{collection}/update/json/docs` → `PUT /{collection}/_doc/{id}`
   - `commit=true` → `?refresh=true` (both make doc immediately searchable)
   - Body passed through unchanged — no field-level translation needed

   **Response transform:**
   - OpenSearch: `{"_index":"mycore","_id":"1","result":"created","_version":1,...}`
   - Solr: `{"responseHeader":{"status":0,"QTime":0}}`

  
   ### Files Changed (4)

   | File | Change |
   |---|---|
   | `features/update-doc.ts` | New — request transform (URI rewrite, method, commit) + response transform (OpenSearch → Solr format) |
   | `features/update-doc.test.ts` | New — 24 unit tests |
   | `update.testcase.ts` | New — 8 e2e test cases (Solr 8/9 + OpenSearch containers) |
   | `registry.ts` | Modified — wired `update` endpoint in request and response registries |

   ### Testing

   | Layer | Result |
   |---|---|
   | Unit tests (vitest) | 24 files, 550 tests ✅ |
   | Java + TS via Gradle | BUILD SUCCESSFUL ✅ |
   | E2E integration (Solr 8 + 9) | All passed ✅ |
   | Sandbox (200K dataset, 167 queries) | 88/167 pass — same as baseline, zero regressions ✅ |
   | Manual sandbox validation | Update, overwrite, error paths all verified ✅ |

   **Unit test coverage (24 tests):**
   - Match guard: `/update/json/docs` matches, `/update` and `/select` don't (4)
   - Happy path: URI rewrite, method, special chars in id, numeric id (3)
   - Commit handling: `commit=true`, absent, `false`, `yes`, `1` (5)
   - Fail-fast: empty body, missing id, empty id, null id, array body, `commitWithin`, `commitWithin=0` (7)
   - Response transform: match guard, created → 0, updated → 0, noop → 1, clears OS fields (5)

   **E2E test cases (8):**
   - Basic single doc, multiple field types, without commit, special chars in id, overwrite existing (5 happy path)
   - Missing id, empty body, commitWithin (3 error path)

   ### Check List

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
@@ -1,0 +1,93 @@
+/**
+ * Single document update — /solr/{collection}/update/json/docs → /{collection}/_doc/{id}
+ *
+ * Translates Solr's JSON document ingestion shortcut to OpenSearch's Index API.
+ *
+ * Solr:       POST /solr/{collection}/update/json/docs  {"id":"1","title":"hello"}
+ * OpenSearch:  PUT /{collection}/_doc/1                  {"id":"1","title":"hello"}
+ *
+ * Commit handling:
+ *   - commit=true → ?refresh=true (both make doc immediately searchable)
+ *   - commitWithin=N → ?refresh=true (OpenSearch has no timed refresh equivalent,
+ *     so we refresh immediately. This is more aggressive than Solr's batched commit
+ *     but satisfies the "searchable within N ms" contract. See LIMITATIONS.md.)
+ *
+ * Fail-fast on:
+ *   - Missing or empty body
+ *   - Array body (bulk not supported yet)
+ *   - Missing "id" field in document
+ *
+ * The body is passed through unchanged — no field-level translation needed.
+ */
+import type { MicroTransform } from '../pipeline';
+import type { RequestContext, ResponseContext } from '../context';
+
+/** Detect GraalVM Java ArrayList — exposes numeric keys + length, unlike a Map. */
+function isJavaList(obj: any): boolean {
+  return typeof obj.get === 'function' && obj.get('0') !== undefined && obj.get('length') !== undefined;
+}
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'update-doc',
+  match: (ctx) => /\/update\/json\/docs/.test(ctx.msg.get('URI') || ''),
+  apply: (ctx) => {
+    const body = ctx.body;
+
+    if (!body || body.size === 0) {
+      throw new Error('[update-doc] Request body is empty — expected a JSON document');
+    }
+
+    if (isJavaList(body)) {
+      throw new Error('[update-doc] Array/bulk updates not supported yet — send one document at a time');
+    }
+
+    const id = body.get('id');
+    if (id == null || id === '') {
+      throw new Error('[update-doc] Document must have an "id" field');
+    }
+
+    // Build target URI with structured params
+    const targetParams = new URLSearchParams();
+    if (ctx.params.get('commit') === 'true' || ctx.params.has('commitWithin')) {
+      targetParams.set('refresh', 'true');
+    }
+
+    const query = targetParams.toString();
+    const suffix = query ? '?' + query : '';
+    const uri = `/${ctx.collection}/_doc/${encodeURIComponent(String(id))}${suffix}`;
+
+    ctx.msg.set('URI', uri);
+    ctx.msg.set('method', 'PUT');
+  },
+};
+
+/**
+ * Response transform — convert OpenSearch _doc response to Solr update response.
+ *
+ * OpenSearch returns: {"_index":"mycore","_id":"1","result":"created","_version":1,...}
+ * Solr returns:       {"responseHeader":{"status":0,"QTime":N}}
+ */
+export const response: MicroTransform<ResponseContext> = {
+  name: 'update-doc-response',
+  match: (ctx) => ctx.responseBody.has('result') && ctx.responseBody.has('_id'),
+  apply: (ctx) => {
+    const result = ctx.responseBody.get('result');
+    const status = (result === 'created' || result === 'updated') ? 0 : 1;
+
+    // Clear OpenSearch-specific fields
+    const keys = Array.from(ctx.responseBody.keys());
+    for (const key of keys) {
+      ctx.responseBody.delete(key);
+    }
+
+    // Write Solr-format response.
+    // QTime is 0 because OpenSearch's _doc response doesn't include processing time.
+    ctx.responseBody.set(
+      'responseHeader',
+      new Map<string, unknown>([
+        ['status', status],
+        ['QTime', 0],
+      ]),
+    );
+  },
+};
--- a/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
+++ b/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
@@ -15,6 +15,7 @@ the root cause, and provides a workaround where one exists.
 | [CURSOR-UNIQUEKEY](#cursor-uniquekey) | Cursor pagination assumes `id` as Solr's uniqueKey field |
 | [CURSOR-REPLAY](#cursor-replay) | Traffic replay with cursorMark not supported |
 | [SOLRCONFIG-REPLAYER](#solrconfig-replayer) | Traffic replayer requires manual solrConfig in bindingsObject |
+| [COMMITWITHIN](#commitwithin) | `commitWithin=N` translated to immediate refresh |
 
 ---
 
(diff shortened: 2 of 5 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
```diff
@@ -0,0 +1,88 @@
...
+ * Single document update — /solr/{collection}/update/json/docs → /{collection}/_doc/{id}
+ *
+ * Translates Solr's JSON document ingestion shortcut to OpenSearch's Index API.
+ *
+ * Solr:       POST /solr/{collection}/update/json/docs  {"id":"1","title":"hello"}
+ * OpenSearch:  PUT /{collection}/_doc/1                  {"id":"1","title":"hello"}
+ *
+ * Commit handling:
+ *   - commit=true → ?refresh=true (both make doc immediately searchable)
+ *   - commitWithin=N is NOT supported — OpenSearch has no equivalent timed refresh.
```
akshay2000: We should call this out in limitations.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
```diff
@@ -0,0 +1,88 @@
...
+ *     Silently dropping it would violate the user's commit timing expectations.
+ *
+ * Fail-fast on:
+ *   - Missing or empty body
+ *   - Array body (bulk not supported yet)
+ *   - Missing "id" field in document
+ *   - commitWithin param (no OpenSearch equivalent)
```
akshay2000: Interesting. Wouldn't setting `refresh=true` be better than failing here? The parameter `commitWithin=N` says commit no later than `N` milliseconds and `refresh=true` does satisfy that. I do understand the performance implications though.

Personally, I would prefer if we commit fast and call out a limitation. Eventually (soon enough) we would also emit a metric counter when we run into such requests so that the customer can make an informed decision.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
```diff
@@ -0,0 +1,88 @@
...
+  match: (ctx) => /\/update\/json\/docs/.test(ctx.msg.get('URI') || ''),
+  apply: (ctx) => {
+    const body = ctx.body;
+
+    if (!body || body.size === 0) {
+      throw new Error('[update-doc] Request body is empty — expected a JSON document');
+    }
+
+    // Array/bulk not supported — GraalVM Java ArrayList exposes numeric keys + length
+    if (typeof body.get === 'function' && body.get('0') !== undefined && body.get('length') !== undefined) {
```
akshay2000: I don't know if this will be a common enough use case, but `isJavaList()` is much more readable.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    if (ctx.params.get('commitWithin')) {
+      throw new Error('[update-doc] commitWithin is not supported — OpenSearch has no equivalent timed refresh');
+    }
+
+    // URI rewrite
+    let uri = `/${ctx.collection}/_doc/${encodeURIComponent(String(id))}`;
+    if (ctx.params.get('commit') === 'true') {
+      uri += '?refresh=true';
```
akshay2000: Isn't there a more structured way to write URL params than string concatenation? I thought we had some params structure.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
```diff
@@ -0,0 +1,88 @@
...
+    ctx.msg.set('URI', uri);
+    ctx.msg.set('method', 'PUT');
+  },
+};
+
+/**
+ * Response transform — convert OpenSearch _doc response to Solr update response.
+ *
+ * OpenSearch returns: {"_index":"mycore","_id":"1","result":"created","_version":1,...}
+ * Solr returns:       {"responseHeader":{"status":0,"QTime":N}}
```
akshay2000, line 68: Does Solr actually return `"responseHeader"` key as a part of the body? Kinda funny!

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
```diff
@@ -0,0 +1,88 @@
...
+    for (const key of keys) {
+      ctx.responseBody.delete(key);
+    }
+
+    // Write Solr-format response
+    ctx.responseBody.set(
+      'responseHeader',
+      new Map<string, unknown>([
+        ['status', status],
+        ['QTime', 0],
```
akshay2000, line 89: Why is `QTime` being set to 0? Is it because OS doesn't give you this value?

akshay2000 (changes requested):

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: Added to `docs/LIMITATIONS.md` under the `COMMITWITHIN` section. Documents that `commitWithin=N` is translated to immediate refresh (`?refresh=true`) since OpenSearch has no timed refresh equivalent, with notes on the performance implications.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: Actually I was kind of 50-50 on this decision during implementation and went ahead with limitation but I think it does make sense to call out it in a limitation and avoid failure so changed to translate `commitWithin=N` to `?refresh=true` instead of throwing. While it's more aggressive than Solr's batched commit, it satisfies the "searchable within N ms" contract. Added to `LIMITATIONS.md` with a note about the performance difference (per-request refresh vs Solr's batched commit). Future work: emit a metric counter when `commitWithin` is encountered so customers can track usage

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: Good call — extracted into a named `isJavaList()` helper function.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: I think thats fair point, updated it to use `URLSearchParams` for building the query string

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: Yes! Solr wraps all responses (select, update, admin) in a JSON envelope with `responseHeader` containing `status` and `QTime`. It's their standard response format even error responses include it. Its kind of strange but to keep the parity we replicate this so Solr clients parsing the response don't break.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: Correct, OpenSearch's `_doc` response doesn't include processing time. We set `QTime: 0` as a placeholder since the field is expected by Solr clients. Added a code comment explaining this. The e2e tests use an `ignore` assertion rule on `QTime` since the actual value differs between Solr and the proxy.

akshay2000: If there are more cases like this, I wonder if we should standardize on a central constant value. Maybe -1 to make it super clear.

akshay2000 (approved):

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-doc.ts
niravpi: Yes I think -1 would make it more clear, will check if there are any other such attribute and we can raise separate enhancement PR for the same

PR #2741, opened 2026-04-16
Solr backfill: multi-shard, S3 subpaths, successive backups, two-level layout, and loud empty-run errors

AndreKurait:
## Summary

Makes the Solr-to-OpenSearch backfill pipeline work in realistic setups: multi-shard collections, multi-collection snapshots, S3 buckets with subpaths, successive backups to the same location, and non-trivial schema definitions. Turns several silent-failure modes into loud errors.

## Background: Solr backup layout

Several bugs stem from misunderstanding Solr's on-disk backup structure. A single `BACKUP` API call for collection `myCol` produces:

```
<location>/<backupName>/myCol/
    zk_backup_0/configs/<configName>/managed-schema.xml
    shard_backup_metadata/md_shard1_0.json
    backup_0.properties
    index/...
```

Successive backups to the same location bump the suffix: `zk_backup_1`, `md_shard1_1.json`, `backup_1.properties`, etc. Code that hardcoded `zk_backup_0` or counted shards before downloading `shard_backup_metadata` silently restored stale or partial data.

Solr 8 also enforces **one collection per backup name** — `BackupCmd.createAndValidateBackupPath()` rejects a second collection backing up under the same name. Multi-collection snapshots need a separate `BACKUP` per collection with distinct names.

## Problems solved

1. **Silent empty runs** — metadata migrate/evaluate returned exit 0 when 0 items were found, masking misconfigured filters and missing snapshots.
2. **Wrong copyField target type** — `SolrSchemaConverter` hardcoded `text` for all copyField destinations instead of resolving the dynamic-field pattern the target matched (e.g. `title_str` → `*_str` → `strings` → `keyword`).
3. **S3 subpath ignored on write** — `SolrSnapshotCreator` hardcoded `location=/` when writing to S3, so `s3://bucket/my-prefix` backups landed at the bucket root.
4. **S3 subpath ignored on read** — `RfsMigrateDocuments` and `ClusterReaderExtractor` built S3 URIs without the key prefix, so reads failed on any non-root path.
5. **Only 1 of N shards migrated** — `SolrBackupIndexMetadataFactory.fromRepo()` counted shards before `shard_backup_metadata/` was downloaded from S3, always finding exactly 1.
6. **Stale zk_backup_0 assumption** — schema parsing and shard-metadata discovery pointed at `zk_backup_0` even after successive backups produced newer revisions.
7. **Wasted S3 downloads on restart** — RFS re-downloaded every shard from S3 on pod restart even when the work coordinator showed all work complete.
8. **BACKUP failed with S3 subpath locations** — some S3-compatible stores reject Solr's BACKUP when the parent prefix has no directory marker object.
9. **Multi-collection backups failed after the first collection** — all collections used the same backup name, which Solr 8 rejects.
10. **Downstream reader couldn't navigate two-level layout** — the reader expected `<collection>/zk_backup_N/` directly but the actual layout nests as `<collection>/<collection>/zk_backup_N/`.

### Error detection: `--succeed-on-empty`

`Items.java` gains `succeedOnEmpty`, `isEmpty()`, and `getEmptyItemsErrors()` which feed `getAllErrors()`. New `--succeed-on-empty` CLI flag (default `true` for backwards compat; set to `false` to fail on zero items). Covered by `ItemsTest`.

### Schema: dynamic-field-aware copyField resolution

`SolrSchemaConverter.processCopyFields` walks the dynamic-field list to find the first pattern matching each copyField destination, then maps the resolved Solr type to the correct OpenSearch type.

### S3 path support (write and read)

`SolrSnapshotCreator.extractS3Path()` parses `s3://bucket/foo/bar` into the BACKUP API's `location` parameter. `RfsMigrateDocuments` and `ClusterReaderExtractor` include the S3 key prefix when constructing read paths. `S3Repo.listSubDirectories()` enables enumeration of `zk_backup_N` entries without mirroring the entire prefix.

### S3 directory marker for subpath locations

`SolrBackupStrategy.ensureS3LocationExists()` writes a zero-byte `application/x-directory` marker before the BACKUP call when the target is a non-root S3 prefix. Failures are logged and swallowed so backup proceeds against providers that don't require the marker.

### Per-collection backup namespacing

`SolrSnapshotCreator` derives a per-collection backup name: `<snapshotName>-<collection>`. This satisfies Solr 8's one-collection-per-backup-name constraint. Resulting layout:

```
<location>/<snapshotName>-<coll1>/<coll1>/...
<location>/<snapshotName>-<coll2>/<coll2>/...
```

### Two-level layout support in downstream reader

`ClusterReaderExtractor` and `SolrBackupLayout.findLatestZkBackup()` now handle the nested `<collection>/<collection>/zk_backup_N/` structure by checking one level deeper when no `zk_backup_N` is found directly.

### Multi-shard discovery and work-coordinator early exit

`SolrBackupIndexMetadataFactory.fromRepo()` accepts a `collectionPreparer` callback invoked before counting shards — wired to the S3 download path so all shards are discovered. `RfsMigrateDocuments` checks the work coordinator for pending items before any S3 download; if complete, the pod exits immediately.

### `SolrBackupLayout` utility

Centralizes `findLatestZkBackup(Path)`, `findLatestZkBackupName(List<String>)`, and `findLatestShardMetadataFiles(Path)`. Used by `SolrSchemaXmlParser`, `SolrBackupSource`, `ClusterReaderExtractor`, and `RfsMigrateDocuments`.

### `SolrBackupStrategy` (CreateSnapshot)

New strategy class for `CreateSnapshot` that handles both SolrCloud (Collections API) and standalone Solr. Encapsulates directory-marker creation and per-collection naming.

## Testing

**E2E testcontainers tests** (Solr 8.11 + LocalStack, tagged `isolatedTest`):
- `TestCreateSnapshotSolrS3` — 9 tests covering `ensureS3LocationExists` (bucket root skip, subpath create, nested subpath, idempotency, bad-endpoint swallow), S3 client building, directory marker verification, full `CreateSnapshot.run()` flow (multi-collection, multi-shard, auto-discovery)
- `SolrS3MetadataMigrationTest` — full metadata evaluate against S3-backed Solr snapshot, pins the two-level layout fix

**Unit tests**: `SolrSchemaConverterTest` (+4), `SolrSnapshotCreatorTest` (+5), `SolrBackupLayoutTest` (+13), `ItemsTest` (+2)

**Python integ**: `solr_tests.py` updated to use collection-scoped `_count` API for verification. `default_operations.get_all_index_details` no longer crashes when `index_prefix_ignore_list` is `None`.

## Migration notes

- `--succeed-on-empty` is new; default `true` preserves existing behavior. Set `false` in CI to catch misconfigurations.
- Existing backups with only `zk_backup_0` continue to work — the layout logic is a superset.
- Multi-collection backups now use `<snapshotName>-<collection>` naming. There is no upgrade path for half-written snapshots from earlier revisions — re-take the snapshot.


DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -32,6 +32,7 @@
 import org.opensearch.migrations.bulkload.models.IndexMetadata;
 import org.opensearch.migrations.bulkload.pipeline.DocumentMigrationBootstrap;
 import org.opensearch.migrations.bulkload.solr.SolrBackupIndexMetadataFactory;
+import org.opensearch.migrations.bulkload.solr.SolrBackupLayout;
```
oeyh, line 39: Nonblocking but this class is growing larger, maybe we can extract some of the Solr part to a helper class for better testing.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -968,6 +912,113 @@ private static void runSolrBackupMigration(
...
+
+                // Discover collection names (cheap: single S3 list-directories call or local dir scan).
+                // The schema XMLs themselves are fetched lazily by the collectionPreparer below -- only
+                // when ShardWorkPreparer actually needs to iterate shards for an uncompleted work item.
+                // If work-coordination already has everything marked complete, ShardWorkPreparer short-
+                // circuits via onAlreadyCompleted and no schema downloads happen at all.
+                var schemas = new java.util.LinkedHashMap<String, JsonNode>();
+                final List<String> collections;
+                if (s3Repo != null) {
+                    collections = new java.util.ArrayList<>(s3Repo.listTopLevelDirectories());
```
oeyh: I don't know if we enforce the class reference styles. There are a number of fully qualified inline references (e.g. `java.util.ArrayList`) in this and other files, some are not necessary because there're already proper import statements.

oeyh (commented):
Looks good! The change is large but the PR description is very helpful. A couple minor things:

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
AndreKurait: Thanks! I've resolved this in https://github.com/opensearch-project/opensearch-migrations/pull/2741/commits/61f611c52c4959b2e2457b58cf787f4c5ecbaf02

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
AndreKurait: Good point, let me see where we can align the interfaces

oeyh (approved):

PR #2742, opened 2026-04-16
feat(solr-transforms): Add JSON request body format support for Solr to OpenSearch migration

niravpi:
### Description

Adds support for Solr's JSON request body format on the `/select` endpoint. Solr accepts queries as either URL params or a POST with JSON body using different key names. This PR adds a pre-processing transform that normalizes JSON body keys into URL params so all downstream transforms work unchanged.

**Solr JSON body format:**

    POST /solr/{collection}/select
    {"query": "title:java", "limit": 10, "offset": 0, "sort": "price asc", "fields": "id,title"}

**Key mapping ([Solr JSON Request API spec](https://solr.apache.org/guide/solr/latest/query-guide/json-request-api.html#supported-properties-and-syntax)):**

| JSON body key | URL param | Notes |
|---|---|---|
| `query` | `q` | |
| `limit` | `rows` | |
| `offset` | `start` | |
| `sort` | `sort` | |
| `filter` | `fq` | String or array |
| `fields` | `fl` | String or array (joined with comma) |
| `facet` | `json.facet` | Moved in body for downstream json-facets transform |
| `params` | merged | Arbitrary key-value pairs merged into ctx.params |

JSON body keys take precedence over URL params (verified against Solr 8/9).

**Not yet supported (documented in LIMITATIONS.md):**
- `queries` — named sub-queries for JSON Query DSL (depends on local params `{!v=$...}`)
- `json.<param_name>` — URL parameter prefix passthrough
- `json=` — entire JSON body as single URL param

### Files Changed (6)

| File | Change |
|---|---|
| `features/json-request.ts` | New — pre-processor that maps JSON body keys to URL params |
| `features/json-request.test.ts` | New — 17 unit tests |
| `json-request.testcase.ts` | New — 6 e2e test cases (Solr 8/9 + OpenSearch) |
| `registry.ts` | Modified — wired `json-request` as first transform in select pipeline |
| `features/select-uri.ts` | Modified — fix duplicate Content-Type header on POST requests |
| `docs/LIMITATIONS.md` | Modified — added JSON-QUERIES and JSON-PARAM-PREFIX limitations |

### Pipeline Order

    global:  validation
    select:  jsonRequest → solrconfigDefaults → selectUri → queryQ → cursorPagination → jsonFacets → fieldList → highlighting → sort

`json-request` runs first so JSON body keys are available as URL params before defaults fill in gaps and before other transforms consume them.

### Testing

| Layer | Result |
|---|---|
| Unit tests (vitest) | 28 files, 638 tests ✅ |
| E2E integration (Solr 8 + 9) | 6/6 passed ✅ |

### Check List

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
@@ -1,0 +1,83 @@
+/**
+ * JSON request body — normalize Solr JSON body format into URL params.
+ *
+ * Solr accepts queries as either URL params or a JSON request body:
+ *   GET  /select?q=title:java&rows=10
+ *   POST /select  {"query":"title:java","limit":10}
+ *
+ * The JSON body uses different key names than URL params:
+ *   query  → q        limit  → rows      offset → start
+ *   sort   → sort     filter → fq        fields → fl
+ *   params → merged into ctx.params (arbitrary key-value pairs)
+ *
+ * This pre-processor runs first in the select pipeline. It reads the
+ * incoming JSON body, maps keys to URL param equivalents, merges them
+ * into ctx.params, and clears the body so downstream transforms can
+ * write the OpenSearch request body fresh.
+ *
+ * JSON body keys take precedence over URL params (verified against Solr 8/9).
+ */
+import type { MicroTransform } from '../pipeline';
+import type { RequestContext, JavaMap } from '../context';
+
+/**
+ * Solr JSON body key → URL param name.
+ *
+ * Complete mapping per Solr's JSON Request API:
+ * https://solr.apache.org/guide/solr/latest/query-guide/json-request-api.html#supported-properties-and-syntax
+ *
+ * Not yet supported:
+ *   - json.<param_name> generic passthrough (URL-param-to-JSON bridge, opposite direction)
+ *   - queries (JSON Query DSL additional named queries, depends on local params {!v=$...})
+ */
+const KEY_MAP: Record<string, string> = {
+  query: 'q',
+  limit: 'rows',
+  offset: 'start',
+  sort: 'sort',
+  filter: 'fq',
+  fields: 'fl',
+};
+
+/** Map JSON body keys to URL params. JSON body takes precedence over URL params (verified against Solr 8/9). */
+function mapBodyKeysToParams(body: JavaMap, params: URLSearchParams): void {
+  for (const [jsonKey, paramKey] of Object.entries(KEY_MAP)) {
+    if (!body.has(jsonKey)) continue;
+    const val = body.get(jsonKey);
+    params.set(paramKey, Array.isArray(val) ? val.join(',') : String(val));
+  }
+}
+
+/** Merge the "params" object from JSON body into URL params. */
+function mergeParamsObject(body: JavaMap, params: URLSearchParams): void {
+  const paramsObj = body.get('params');
+  if (!paramsObj || typeof paramsObj.entries !== 'function') return;
+  for (const [key, val] of paramsObj.entries()) {
+    if (!params.has(key)) params.set(key, String(val));
+  }
+}
+
+/** Move "facet" to "json.facet" key for downstream json-facets transform. */
+function moveFacetKey(body: JavaMap, params: URLSearchParams): void {
+  if (body.has('facet') && !params.has('json.facet')) {
+    body.set('json.facet', body.get('facet'));
+  }
+}
+
+/** Remove all JSON body keys that were mapped to params. */
+function clearMappedKeys(body: JavaMap): void {
+  for (const jsonKey of Object.keys(KEY_MAP)) body.delete(jsonKey);
+  body.delete('params');
+  body.delete('facet');
+}
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'json-request',
+  match: (ctx) => ctx.body.size > 0 && ctx.body.has('query'),
+  apply: (ctx) => {
+    mapBodyKeysToParams(ctx.body, ctx.params);
+    mergeParamsObject(ctx.body, ctx.params);
+    moveFacetKey(ctx.body, ctx.params);
+    clearMappedKeys(ctx.body);
+  },
+};
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/select-uri.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/select-uri.ts
@@ -19,6 +19,9 @@ export const request: MicroTransform<RequestContext> = {
     ctx.msg.set('method', 'POST');
     const headers = ctx.msg.get('headers');
     if (headers) {
+      // Remove original Content-Type (may differ in casing) to avoid duplicate headers.
+      // The Java proxy uses headers.add() which would create two Content-Type entries.
+      headers.delete('Content-Type');
       headers.set('content-type', 'application/json');
     }
   },
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/json-request.testcase.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/json-request.testcase.ts
@@ -1,0 +1,135 @@
+/**
+ * E2E test cases for JSON request body format on /select.
+ *
+ * Solr accepts queries as POST with JSON body using different key names
+ * (query, limit, offset, etc.). The json-request transform normalizes
+ * these to URL params before the rest of the pipeline runs.
+ */
+import { solrTest, SOLR_INTERNAL_RULES } from '../test-types';
+import type { TestCase } from '../test-types';
+
+export const testCases: TestCase[] = [
+  solrTest('json-request-basic-query', {
+    description: 'JSON body with query key should work like q= URL param',
+    documents: [
+      { id: '1', title: 'hello world' },
+      { id: '2', title: 'goodbye world' },
+    ],
+    method: 'POST',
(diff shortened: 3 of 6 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
```diff
@@ -0,0 +1,83 @@
...
+  for (const [key, val] of paramsObj.entries()) {
+    if (!params.has(key)) params.set(key, String(val));
+  }
+}
+
+/** Move "facet" to "json.facet" key for downstream json-facets transform. */
+function moveFacetKey(body: JavaMap, params: URLSearchParams): void {
+  if (body.has('facet') && !params.has('json.facet')) {
+    body.set('json.facet', body.get('facet'));
+  }
```
nagarajg17, line 64: what happens if `params.has('json.facet')` => True

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+
+/** Remove all JSON body keys that were mapped to params. */
+function clearMappedKeys(body: JavaMap): void {
+  for (const jsonKey of Object.keys(KEY_MAP)) body.delete(jsonKey);
```
nagarajg17, line 69: Won't error if key is not found?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/select-uri.ts
```diff
@@ -19,6 +19,9 @@ export const request: MicroTransform<RequestContext> = {
     ctx.msg.set('method', 'POST');
     const headers = ctx.msg.get('headers');
     if (headers) {
+      // Remove original Content-Type (may differ in casing) to avoid duplicate headers.
+      // The Java proxy uses headers.add() which would create two Content-Type entries.
+      headers.delete('Content-Type');
```
nagarajg17, line 24: Below we are setting content-type header. Won't it override if key was already present?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
```diff
@@ -0,0 +1,83 @@
...
+  for (const [jsonKey, paramKey] of Object.entries(KEY_MAP)) {
+    if (!body.has(jsonKey) || params.has(paramKey)) continue;
+    const val = body.get(jsonKey);
+    params.set(paramKey, Array.isArray(val) ? val.join(',') : String(val));
+  }
+}
+
+/** Merge the "params" object from JSON body into URL params. */
+function mergeParamsObject(body: JavaMap, params: URLSearchParams): void {
+  const paramsObj = body.get('params');
```
nagarajg17, line 53: Will Solr throw error if we place like say limit field in params?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/json-request.testcase.ts
```diff
@@ -0,0 +1,97 @@
...
+ * E2E test cases for JSON request body format on /select.
+ *
+ * Solr accepts queries as POST with JSON body using different key names
+ * (query, limit, offset, etc.). The json-request transform normalizes
+ * these to URL params before the rest of the pipeline runs.
+ */
+import { solrTest, SOLR_INTERNAL_RULES } from '../test-types';
+import type { TestCase } from '../test-types';
+
+export const testCases: TestCase[] = [
```
nagarajg17, line 11: Can we add integ for
1. Testing precedence
2. Params

nagarajg17:
Solr supports json-prefixed query parameters - `curl http://localhost:8983/solr/techproducts/query -d 'json={"query":"memory"}'`
https://solr.apache.org/guide/solr/latest/query-guide/json-request-api.html

niravpi:
> Solr supports json-prefixed query parameters - `curl http://localhost:8983/solr/techproducts/query -d 'json={"query":"memory"}'` https://solr.apache.org/guide/solr/latest/query-guide/json-request-api.html

 Actually the doc you shared is around both `json={"query":"..."}` (entire JSON as URL param) and `json.<param_name>` (individual property overrides). Both are documented in LIMITATIONS.md under `JSON-PARAM-PREFIX` as not supported. They're URL-to-JSON-body bridges the opposite direction from what this transform handles. Our transform normalizes JSON request body → URL params, not the other way around.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
niravpi: No, `Map.delete()` returns `false` silently on non-existent keys per the JS spec. Same for Java Maps via GraalVM. So `body.delete('filter')` when `filter` wasn't in the body is a safe no-op.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/select-uri.ts
niravpi: The issue was case sensitivity. The original POST request has `Content-Type` (capital C-T) and we set `content-type` (lowercase). The Java proxy's header Map treats these as two different keys, so both end up in the outgoing request OpenSearch rejects with "only one Content-Type header should be provided". The `delete('Content-Type')` removes the original casing, then `set('content-type', ...)` adds the lowercase version one header. For GET requests (no original Content-Type), the delete is a no-op.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
niravpi: Solr won't error it ignores unrecognized params. In our transform, `params` values are merged into `ctx.params` as-is. If someone puts `rows` (the URL param name) in `params`, it works normally. If they put `limit` (the JSON body key name), it gets merged into `ctx.params` as `limit` which our validation would reject as an unsupported param since `limit` isn't a recognized URL param name.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/json-request.testcase.ts
niravpi: Thanks for bringing this up. 
Added both:
  1. `json-request-url-params-precedence` sends `rows=1` in URL + `limit=99` in JSON body, verifies JSON body wins. We also corrected the precedence logic based on manual testing against Solr 8/9 — JSON body takes precedence over URL params
  2. `json-request-with-params-object` sends `params: { rows: '1' }` in JSON body, verifies it's merged correctly.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
niravpi: When `json.facet` is already present as a URL param, the body `facet` is dropped and the URL param version is used. Solr actually merges both sources (returns facets from both URL and body), but our downstream `json-facets.ts` uses an either/or approach it reads from body first, falls back to params, but doesn't merge both. So if we moved body `facet` to `body.json.facet`, the downstream would use body and silently drop the URL param facets. Current behavior (URL param wins) is the safer choice given this downstream limitation. The proper fix is a `json-facets.ts` change to merge both sources can be address in a separate  PR.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/select-uri.ts
nagarajg17: Why exactly it fails with this PR changes?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
nagarajg17: Can we add a TODO or follow up task in the tracker. Mostly Akshay can work on this

niravpi: ack will do it

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/select-uri.ts
niravpi: Before this PR, `/select` only received **GET requests** no `Content-Type` header present, so `select-uri` setting `content-type` (lowercase) was the only one. No conflict. This PR adds **POST with JSON body** support. Those requests arrive with `Content-Type: application/json` (uppercase). The `json-request` transform extracts body keys into URL params and clears the body, then `select-uri` converts to GET but the original `Content-Type` is still in the header map. When `select-uri` sets `content-type: application/x-www-form-urlencoded` (lowercase), the Java proxy's **case-sensitive** header map keeps both → OpenSearch rejects with 400. The `delete('Content-Type')` removes the stale original before setting the new one. For GET requests it's a no-op.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-request.ts
nagarajg17: Confirm me once added. Rest looks good

niravpi: Done

nagarajg17 (approved):

PR #2747, opened 2026-04-16
feat(solr-transforms): Add delete-by-id and /update command dispatcher

niravpi:
### Description

Adds a command dispatcher for Solr's `/update` endpoint and single document delete-by-id support. The dispatcher parses the JSON body, identifies the command type, and delegates to the appropriate handler. Also refactors the response transform from `update-doc.ts` into a generic handler that supports create, update, and delete responses.

**Supported commands on `/update`:**

| Solr command | OpenSearch translation | Status |
|---|---|---|
| `{"delete": {"id": "1"}}` | `DELETE /{collection}/_doc/1` | ✅ New |
| `{"add": {"doc": {...}}}` | `PUT /{collection}/_doc/{id}` (delegates to update-doc) | ✅ New |
| `{"delete": {"query": "..."}}` | — | ❌ Fail fast |
| `{"commit": {}}` | — | ❌ Fail fast |
| Mixed commands | — | ❌ Fail fast |
| Array/bulk operations | — | ❌ Fail fast |
| XML content type | — | ❌ Fail fast |

**Architecture:**

```
POST /solr/{collection}/update
        │
        ▼
  update-command.ts (dispatcher)
        │
        ├── "delete" with "id"  →  delete-doc.ts  →  DELETE /_doc/{id}
        ├── "add" with "doc"    →  update-doc.ts   →  PUT /_doc/{id}
        └── unknown/mixed       →  fail fast
```

```
Pipeline order:
  update: [
    updateDoc.request,      // /update/json/docs (direct path, unchanged)
    updateCommand.request,  // /update (command dispatcher)
  ]
  response: [
    updateCommand.response, // generic _doc response → Solr format
  ]
```

**Response transform refactoring:**
- Moved from `update-doc.ts` to `update-command.ts` since it's generic for all `_doc` API operations
- Now handles `created`, `updated`, `deleted`, and `not_found` results (previously only `created`/`updated`)
- `not_found` maps to status 0 (matching Solr's idempotent delete behavior)

### Files Changed (9)

| File | Change |
|---|---|
| `features/update-command.ts` | New — command dispatcher + generic response transform |
| `features/update-command.test.ts` | New — 30 unit tests |
| `features/delete-doc.ts` | New — delete-by-id handler |
| `features/delete-doc.test.ts` | New — 7 unit tests |
| `update-command.testcase.ts` | New — 10 e2e test cases (5 happy + 5 error) |
| `features/update-doc.ts` | Modified — removed response transform (moved to update-command) |
| `features/update-doc.test.ts` | Modified — removed response transform tests (moved to update-command) |
| `registry.ts` | Modified — wired update-command into update endpoint pipeline |
| `docs/LIMITATIONS.md` | Modified — added UPDATE-COMMANDS section |

### Testing

| Layer | Result |
|---|---|
| Unit tests (vitest) | 26 files, 581 tests ✅ |
| E2E integration (Solr 8 + 9) | 10/10 passed ✅ |
| Sandbox (200K dataset, 167 queries) | 88/167 — same baseline, zero regressions ✅ |
| Manual sandbox validation | 14/14 tests passed ✅ |

### Check List

### DIFF
--- a/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
+++ b/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
@@ -21,6 +21,7 @@ the root cause, and provides a workaround where one exists.
 | [FQ-CACHING](#fq-caching)                       | Filter query caching granularity differs between Solr and OpenSearch |
 | [JSON-QUERIES](#json-queries)                   | JSON Request API `queries` key not supported |
 | [JSON-PARAM-PREFIX](#json-param-prefix)         | JSON Request API `json.<param>` prefix passthrough not supported |
+| [UPDATE-COMMANDS](#update-commands)             | Only `add` and `delete-by-id` commands supported; JSON only |
 
 ---
 
@@ -461,3 +462,35 @@ parameters with the `json.` prefix are treated as unsupported params and
 rejected by validation.
 
 **Reference:** https://solr.apache.org/guide/solr/latest/query-guide/json-request-api.html#json-parameter-merging
+
+---
+
+## UPDATE-COMMANDS
+
+**Feature:** Solr `/update` endpoint command support
+
+**Solr behaviour:**
+The `/update` endpoint accepts JSON or XML bodies with multiple command types:
+`add`, `delete` (by id or query), `commit`, `optimize`, `rollback`. Commands
+can be mixed in a single request and can use arrays for bulk operations.
+Both JSON and XML content types are supported.
+
+**Current support:**
+
+| Command | Status |
+|---|---|
+| `add` (single doc) | ✅ Supported — `{"add":{"doc":{...}}}` |
+| `add` with `boost` | ❌ Not supported — fails fast (OpenSearch has no document-level boost) |
+| `add` with `overwrite: false` | ❌ Not supported — fails fast (OpenSearch always overwrites) |
+| `delete` by id | ✅ Supported — `{"delete":{"id":"..."}}` |
+| `delete` by query | ❌ Not supported — fails fast |
+| `commit` | ❌ Not supported — fails fast |
+| `optimize` / `rollback` | ❌ Not supported — fails fast |
+| Mixed commands | ❌ Not supported — fails fast |
+| Array/bulk operations | ❌ Not supported — fails fast |
+| XML content type | ❌ Not supported — fails fast (JSON only) |
+
+**Content type:**
+Only `application/json` request bodies are supported. XML bodies
+(`text/xml`, `application/xml`) cannot be parsed by the shim and will
+fail with an error. Clients using XML format must switch to JSON.
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
@@ -1,0 +1,36 @@
+/**
+ * Single document delete by ID — DELETE /{collection}/_doc/{id}
+ *
+ * Called by update-router.ts after extracting the delete command data.
+ * Expects ctx.body to be the normalized delete data: {"id": "..."}
+ *
+ * Solr:       POST /solr/{collection}/update  {"delete":{"id":"1"}}
+ * OpenSearch:  DELETE /{collection}/_doc/1
+ */
+import type { MicroTransform } from '../pipeline';
+import type { RequestContext } from '../context';
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'delete-doc',
+  apply: (ctx) => {
+    if (!ctx.collection) {
+      throw new Error('[delete-doc] delete: collection could not be determined from URI');
+    }
+
+    const rawId = ctx.body.get('id');
+    const id = rawId != null ? String(rawId).trim() : '';
+    if (!id) {
+      throw new Error('[delete-doc] delete: document must have a non-empty "id" field');
+    }
+
+    const targetParams = new URLSearchParams();
+    if (ctx.params.get('commit') === 'true' || ctx.params.has('commitWithin')) {
+      targetParams.set('refresh', 'true');
+    }
+
+    const query = targetParams.toString();
+    const suffix = query ? '?' + query : '';
+    ctx.msg.set('URI', `/${ctx.collection}/_doc/${encodeURIComponent(id)}${suffix}`);
+    ctx.msg.set('method', 'DELETE');
+  },
+};
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
@@ -1,0 +1,390 @@
+import { describe, it, expect } from 'vitest';
+import { request, response } from './update-router';
+import type { RequestContext, ResponseContext, JavaMap } from '../context';
+
+function buildCtx(uri: string, body: Map<string, any>): RequestContext {
+  const msg = new Map<string, any>([
+    ['URI', uri],
+    ['method', 'POST'],
+    ['headers', new Map()],
+  ]) as unknown as JavaMap;
+  return {
+    msg,
+    endpoint: 'update',
+    collection: /\/solr\/([^/]+)\//.exec(uri)?.[1],
+    params: new URLSearchParams(),
+    body: body as unknown as JavaMap,
+  };
+}
+
+function deleteBody(id: any): Map<string, any> {
+  return new Map([['delete', new Map([['id', id]])]]);
+}
+
+function deleteByQueryBody(query: string): Map<string, any> {
+  return new Map([['delete', new Map([['query', query]])]]);
+}
+
+function addBody(doc: Record<string, any>): Map<string, any> {
+  return new Map([['add', new Map([['doc', new Map(Object.entries(doc))]])]]);
+}
+
+describe('update-router request', () => {
+  // --- routing ---
+
+  it('routes /update/json/docs to update-doc handler', () => {
+    const body = new Map<string, any>([['id', '1'], ['title', 'hello']]);
+    const msg = new Map<string, any>([
+      ['URI', '/solr/mycore/update/json/docs'],
+      ['method', 'POST'],
+      ['headers', new Map()],
+    ]) as unknown as JavaMap;
+    const ctx: RequestContext = {
+      msg, endpoint: 'update', collection: 'mycore',
+      params: new URLSearchParams(), body: body as unknown as JavaMap,
+    };
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/1');
+    expect(ctx.msg.get('method')).toBe('PUT');
+  });
+
+  it('preserves body fields when routing /update/json/docs', () => {
+    const body = new Map<string, any>([['id', '1'], ['title', 'hello'], ['price', 9.99]]);
+    const msg = new Map<string, any>([
+      ['URI', '/solr/mycore/update/json/docs'],
+      ['method', 'POST'],
+      ['headers', new Map()],
+    ]) as unknown as JavaMap;
+    const ctx: RequestContext = {
+      msg, endpoint: 'update', collection: 'mycore',
+      params: new URLSearchParams(), body: body as unknown as JavaMap,
+    };
+    request.apply(ctx);
+    expect(ctx.body.get('title')).toBe('hello');
+    expect(ctx.body.get('price')).toBe(9.99);
+  });
+
+  it('does not match /foo/update/json/docs123 as json docs path', () => {
+    const body = new Map<string, any>([['id', '1']]);
+    const msg = new Map<string, any>([
+      ['URI', '/solr/mycore/update/json/docs123'],
+      ['method', 'POST'],
+      ['headers', new Map()],
+    ]) as unknown as JavaMap;
+    const ctx: RequestContext = {
+      msg, endpoint: 'update', collection: 'mycore',
+      params: new URLSearchParams(), body: body as unknown as JavaMap,
+    };
+    // Should go through dispatchCommand, not json docs path — will throw since body has no command
+    expect(() => request.apply(ctx)).toThrow('no recognized command');
+  });
+
+  it('throws on XML content type', () => {
+    const body = new Map<string, any>([['delete', new Map([['id', '1']])]]);
+    const msg = new Map<string, any>([
+      ['URI', '/solr/mycore/update'],
+      ['method', 'POST'],
+      ['headers', new Map([['Content-Type', 'text/xml']])],
+    ]) as unknown as JavaMap;
+    const ctx: RequestContext = {
+      msg, endpoint: 'update', collection: 'mycore',
+      params: new URLSearchParams(), body: body as unknown as JavaMap,
+    };
+    expect(() => request.apply(ctx)).toThrow('only JSON content type is supported');
+  });
+
+  it('throws on non-POST method', () => {
+    const body = new Map<string, any>([['delete', new Map([['id', '1']])]]);
+    const msg = new Map<string, any>([
+      ['URI', '/solr/mycore/update'],
+      ['method', 'GET'],
+      ['headers', new Map()],
+    ]) as unknown as JavaMap;
+    const ctx: RequestContext = {
+      msg, endpoint: 'update', collection: 'mycore',
+      params: new URLSearchParams(), body: body as unknown as JavaMap,
+    };
+    expect(() => request.apply(ctx)).toThrow('only POST method is accepted');
+  });
+
+  // --- delete: happy path ---
+
+  it('dispatches delete to DELETE /_doc/{id}', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody('1'));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/1');
+    expect(ctx.msg.get('method')).toBe('DELETE');
+  });
+
+  it('handles delete with special chars in id', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody('doc/with spaces'));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/doc%2Fwith%20spaces');
+  });
+
+  it('handles delete with numeric id', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody(42));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/42');
+  });
+
+  // --- delete: error cases ---
+
+  it('throws on delete with empty id', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody(''));
+    expect(() => request.apply(ctx)).toThrow('non-empty "id"');
+  });
+
+  it('throws on delete with null id', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody(null));
+    expect(() => request.apply(ctx)).toThrow('non-empty "id"');
+  });
+
+  it('throws on delete with whitespace-only id', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody('   '));
+    expect(() => request.apply(ctx)).toThrow('non-empty "id"');
+  });
+
+  it('throws on delete with unsupported _route_ field', () => {
+    const body = new Map([['delete', new Map<string, any>([['id', '1'], ['_route_', 'shard1']])]]);
+    const ctx = buildCtx('/solr/mycore/update', body);
+    expect(() => request.apply(ctx)).toThrow("unsupported field '_route_'");
+  });
+
+  it('throws on delete with unsupported version field', () => {
+    const body = new Map([['delete', new Map<string, any>([['id', '1'], ['version', 123]])]]);
+    const ctx = buildCtx('/solr/mycore/update', body);
+    expect(() => request.apply(ctx)).toThrow("unsupported field 'version'");
+  });
+
+  it('handles delete with id containing reserved chars (/, ?, #)', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteBody('doc/1?q=x#frag'));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/doc%2F1%3Fq%3Dx%23frag');
+  });
+
+  it('handles delete with very long id', () => {
+    const longId = 'a'.repeat(500);
+    const ctx = buildCtx('/solr/mycore/update', deleteBody(longId));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe(`/mycore/_doc/${longId}`);
+  });
+
+  it('throws on delete-by-query', () => {
+    const ctx = buildCtx('/solr/mycore/update', deleteByQueryBody('title:old'));
+    expect(() => request.apply(ctx)).toThrow('delete-by-query is not supported');
+  });
+
+  it('throws on invalid delete format (not a Map)', () => {
+    const ctx = buildCtx('/solr/mycore/update', new Map([['delete', 'invalid']]));
+    expect(() => request.apply(ctx)).toThrow('invalid command format');
+  });
+
+  // --- add: happy path ---
+
+  it('dispatches add to PUT /_doc/{id}', () => {
+    const ctx = buildCtx('/solr/mycore/update', addBody({ id: '1', title: 'hello' }));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/1');
+    expect(ctx.msg.get('method')).toBe('PUT');
+  });
+
+  it('handles add with multiple fields', () => {
+    const ctx = buildCtx('/solr/mycore/update', addBody({ id: '1', title: 'hi', price: 9.99 }));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/1');
+  });
+
+  it('handles add with numeric id', () => {
+    const ctx = buildCtx('/solr/mycore/update', addBody({ id: 42, title: 'test' }));
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/42');
+  });
+
+  // --- add: error cases ---
+
+  it('throws on add without doc field', () => {
+    const body = new Map<string, any>([['add', new Map([['id', '1']])]]);
+    const ctx = buildCtx('/solr/mycore/update', body);
+    expect(() => request.apply(ctx)).toThrow('must contain a "doc" field');
+  });
+
+  it('throws on invalid add format (not a Map)', () => {
+    const body = new Map<string, any>([['add', 'invalid']]);
+    const ctx = buildCtx('/solr/mycore/update', body);
+    expect(() => request.apply(ctx)).toThrow('invalid command format');
+  });
+
+  it('throws on add with doc missing id', () => {
+    const ctx = buildCtx('/solr/mycore/update', addBody({ title: 'no id' }));
+    expect(() => request.apply(ctx)).toThrow('must have an "id" field');
+  });
+
+  it('throws on add with boost', () => {
+    const body = new Map<string, any>([
+      ['add', new Map<string, any>([['doc', new Map([['id', '1']])], ['boost', 2]])],
+    ]);
+    const ctx = buildCtx('/solr/mycore/update', body);
+    expect(() => request.apply(ctx)).toThrow('document-level boost is not supported');
+  });
+
+  it('throws on add with overwrite=false', () => {
+    const body = new Map<string, any>([
+      ['add', new Map<string, any>([['doc', new Map([['id', '1']])], ['overwrite', false]])],
+    ]);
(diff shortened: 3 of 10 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
k-rooot, line 1: Any thoughts on implementing the update functionalities as a single update-router transform that inspects the request body and dispatches to sub-handlers:

update-router.ts (match: /update/ endpoint)
  ├── detects body shape → routes to:
  ├── update-doc.ts        — single doc add (existing, /update/json/docs)
  ├── update-delete.ts     — delete by id / delete by query
  ├── update-bulk.ts       — array body → _bulk API
  ├── update-commit.ts     — commit/optimize commands
  └── update-atomic.ts     — partial updates with modifiers

The router would be a single micro transform registered on the update endpoint that inspects the body keys (delete, commit, optimize, or array/object doc) and calls the appropriate handler. This keeps the pipeline registration clean, one entry point for `/update`.

Let me know what you think.

niravpi: Thanks for the suggestion! Refactored to implement exactly this pattern, a single `update-router.ts` that serves as the sole entry point for all `/update/*` paths.

   **Architecture:**

   update-router.ts (single entry point for /update/*)
   ├── /update/json/docs             → update-doc.ts (body IS the document)
   ├── /update {"add":{"doc":{…}}}   → update-doc.ts (unwraps add.doc)
   ├── /update {"delete":{"id":"…"}} → delete-doc.ts
   └── unsupported commands          → fail fast with clear error

   **Registry now has a single entry:**
   ```ts
   update: [
     updateRouter.request,   // single entry point: routes /update/* by path and body shape
   ]

   The router detects the path (/update/json/docs vs /update) and body shape (command type), then delegates to the appropriate handler via direct .apply() call.

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
```diff
@@ -368,3 +369,35 @@ internal heuristics.
...
+`add`, `delete` (by id or query), `commit`, `optimize`, `rollback`. Commands
+can be mixed in a single request and can use arrays for bulk operations.
+Both JSON and XML content types are supported.
+
+**Current support:**
+
+| Command | Status |
+|---|---|
+| `add` (single doc) | ✅ Supported — `{"add":{"doc":{...}}}` |
+| `add` with `boost` | ❌ Not supported — boost is silently ignored (OpenSearch has no document-level boost) |
```
k-rooot: Instead of silently dropping the boost, we should fail the request with a clear reason such as “unsupported in OpenSearch” or “translation unavailable.” This aligns with the fail-fast approach we are adopting for query translation.

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
```diff
@@ -368,3 +369,35 @@ internal heuristics.
...
+| `add` with `overwrite: false` | ❌ Not supported — always overwrites |
+| `delete` by id | ✅ Supported — `{"delete":{"id":"..."}}` |
```
k-rooot: Will these two fail silently or with appropriate messaging?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+  return /\/update\/json\/docs/.test(uri);
+}
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'update-router',
+  apply: (ctx) => {
+    const uri = ctx.msg.get('URI') || '';
+
+    // /update/json/docs — body IS the document, delegate directly
+    if (isJsonDocsPath(uri)) {
```
k-rooot, line 96: This path skips all validation performed in dispatchCommand.
* /update validates empty body, unsupported formats
* /update/json/docs blindly forwards

This can lead to subtle bugs (e.g., empty body or array payloads slipping through).

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const command = identifyCommand(body);
+  if (!command) {
+    throw new Error('[update-router] No recognized command in request body');
+  }
+
+  const commandData = body.get(command);
+
+  // Array/bulk not supported — single command only
+  if (Array.isArray(commandData) || commandData?.get?.('0') !== undefined) {
```
k-rooot: This check is fragile for two reasons:
* get('0') depends on Map semantics and will not work with plain objects
* It does not correctly detect all array cases (especially nested structures like {"add":[{"doc":...}]})

Any thoughts to normalise first, then rely on Array.isArray() only.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+    throw new Error('[update-router] Request body is empty or not JSON — only JSON content type is supported');
+  }
+
+  if (hasMultipleCommands(body)) {
+    throw new Error('[update-router] Mixed commands in a single request are not supported yet');
+  }
+
+  const command = identifyCommand(body);
+  if (!command) {
+    throw new Error('[update-router] No recognized command in request body');
```
k-rooot: 1. The error gives no visibility into what was actually received. Can we include the keys from the request body in the error message. This will make debugging significantly easier.

2. Across the file, error messages vary in tone and structure:* "Array/bulk operations are not supported yet"* "Add command must contain a \"doc\" field"

Can we standardize the format?
```
[update-router] <operation>: <reason>
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+
+function handleDelete(ctx: RequestContext, data: any): void {
+  if (!data || typeof data.has !== 'function') {
+    throw new Error('[update-router] Invalid delete command format');
+  }
+  if (data.has('query')) {
+    throw new Error('[update-router] Delete-by-query is not supported yet');
+  }
+  ctx.body = data;
+  deleteDocRequest.apply(ctx);
```
k-rooot, line 159: This simply forwards the entire delete payload downstream. However, Solr delete supports additional fields:

```
{"delete":{"id":"1","_route_":"x","version":123}}
```

Right now, _route_ is silently ignored and versioning is lost

Even if downstream doesn’t support these yet, silently dropping them is risky. Can we explicitly extract and validate supported fields (id), and fail clearly for unsupported ones

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+ * Generic for all _doc API operations (create, update, delete).
+ * OpenSearch returns: {"_index":"mycore","_id":"1","result":"created|updated|deleted",...}
+ * Solr returns:       {"responseHeader":{"status":0,"QTime":N}}
+ */
+export const response: MicroTransform<ResponseContext> = {
+  name: 'update-response',
+  match: (ctx) => ctx.responseBody.has('result') && ctx.responseBody.has('_id'),
+  apply: (ctx) => {
+    const result = ctx.responseBody.get('result');
+    const status = (result === 'created' || result === 'updated' || result === 'deleted' || result === 'not_found') ? 0 : 1;
```
k-rooot, line 201: Treating not_found as success is questionable. While Solr can treat delete of a missing doc as success, that behaviour is context-dependent and should be explicitly defined not inferred.

Can we consider HTTP status code (if available in context) and document how not_found should be treated.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+    for (const key of keys) {
+      ctx.responseBody.delete(key);
+    }
+
+    // QTime is 0 because OpenSearch's _doc response doesn't include processing time.
+    ctx.responseBody.set(
+      'responseHeader',
+      new Map<string, unknown>([
+        ['status', status],
+        ['QTime', 0],
```
k-rooot, line 213: This is misleading. Even if OpenSearch doesn’t return timing, hardcoding 0 can confuse monitoring and debugging.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+  for (const cmd of KNOWN_COMMANDS) {
+    if (body.has(cmd)) count++;
+    if (count > 1) return true;
+  }
+  return false;
+}
+
+/** Check if the path is the /update/json/docs shortcut. */
+function isJsonDocsPath(uri: string): boolean {
+  return /\/update\/json\/docs/.test(uri);
```
k-rooot: This regex will match unintended paths like `/foo/update/json/docs123`

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+ *   /update  mixed commands             → multiple commands in one request
+ *   /update  XML body                   → only JSON supported
+ *
+ * Future handlers are added as new cases in dispatchCommand() —
+ * no pipeline or registry changes needed.
+ *
+ * Response transform is also here — generic for all _doc API responses
+ * (create, update, delete). Converts OpenSearch format to Solr format.
+ */
+import type { MicroTransform } from '../pipeline';
```
k-rooot, line 29: 1. There is no explicit check for JSON content type. Given that XML is explicitly unsupported (as per comments), requests with XML payloads will fail in unpredictable ways. Can we validate Content-Type early and fail clearly.

2. Right now the router follows a simple:
```
detect → dispatch
```

This works for P0, but future features (bulk, atomic updates) will require:
```
parse → normalize → dispatch
```

You’re very close to needing this shift, so it’s worth keeping in mind as you extend functionality.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  const commandData = body.get(command);
+
+  // Array/bulk not supported — single command only
+  if (Array.isArray(commandData) || commandData?.get?.('0') !== undefined) {
+    throw new Error('[update-router] Array/bulk operations are not supported yet — send one command at a time');
+  }
+
+  switch (command) {
```
k-rooot: This works, but as more commands are added (commit, optimize, bulk), this will grow. Can we replace with a handler registry map for better scalability.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
```diff
@@ -99,7 +99,7 @@ export const responseRegistry: TransformRegistry<ResponseContext> = {
       responseHeader.response, // synthesize responseHeader
     ],
     update: [
-      updateDoc.response, // OpenSearch _doc response → Solr update response format
+      updateRouter.response,  // generic _doc response → Solr format (create/update/delete)
```
k-rooot, line 104: Any reason why we need to register update in two places.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/update-router.testcase.ts
```diff
@@ -0,0 +1,161 @@
+/**
+ * E2E test cases for /update command dispatcher — delete-by-id and add-via-update.
```
k-rooot, line 2: 1. The below tests validate response shape but not actual side effects. 

`update-cmd-delete-by-id`, `update-cmd-add-single-doc` -> Right now, these tests only assert the response format, not whether the document was actually deleted or indexed. 

This is a critical gap for an E2E test. eg: A bug in deleteDocRequest could silently fail, and this test would still pass.

2. Missing coverage for `/update/json/docs`

3. Can we add a test for invalid body types (non-JSON) and unknown commands

4. Delete tests don’t cover edge cases beyond happy path -> numeric IDs, long ids, IDs with reserved characters (e.g. /, ?, #)

5. Can we add test for missing doc in add command

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/update-router.testcase.ts
```diff
@@ -0,0 +1,161 @@
...
+  // --- error paths ---
+
+  solrTest('update-cmd-error-delete-by-query', {
+    description: 'Delete-by-query should return 500',
+    documents: [{ id: '1', title: 'test' }],
+    method: 'POST',
+    requestBody: JSON.stringify({ delete: { query: 'title:test' } }),
+    requestPath: '/solr/testcollection/update',
+    expectedStatusCode: 500,
+    expectedErrorContains: 'Request transform failed',
```
k-rooot, line 109: This pattern of generic messaging validation is repeated across all error tests. is this intentional? Should  we validate which failure occurred?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
```diff
@@ -0,0 +1,247 @@
...
+}
+
+function addBody(doc: Record<string, any>): Map<string, any> {
+  return new Map([['add', new Map([['doc', new Map(Object.entries(doc))]])]]);
+}
+
+describe('update-router request', () => {
+  // --- routing ---
+
+  it('routes /update/json/docs to update-doc handler', () => {
```
k-rooot, line 35: Should we verify that the body is preserved correctly and that fields are passed through as expected

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
```diff
@@ -0,0 +1,247 @@
...
+    const ctx: RequestContext = {
+      msg, endpoint: 'update', collection: 'mycore',
+      params: new URLSearchParams(), body: body as unknown as JavaMap,
+    };
+    request.apply(ctx);
+    expect(ctx.msg.get('URI')).toBe('/mycore/_doc/1');
+    expect(ctx.msg.get('method')).toBe('PUT');
+  });
+
+  it('routes /update with delete command to delete-doc handler', () => {
```
k-rooot, line 51: Seems like a duplicate of `dispatches delete to DELETE /_doc/{id}`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
```diff
@@ -0,0 +1,247 @@
...
+  });
+
+  it('throws on unsupported optimize command', () => {
+    const ctx = buildCtx('/solr/mycore/update', new Map([['optimize', new Map()]]));
+    expect(() => request.apply(ctx)).toThrow("'optimize' is not supported yet");
+  });
+
+  it('throws on array of deletes', () => {
+    // Simulate Java ArrayList with numeric keys + length
+    const deleteArray = new Map<string, any>([['0', new Map([['id', '1']])], ['1', new Map([['id', '2']])], ['length', 2]]);
```
k-rooot: This is a clever hack, but tightly coupled to current implementation (get('0'))  and not representative of real JS arrays

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
```diff
@@ -0,0 +1,247 @@
+import { describe, it, expect } from 'vitest';
+import { request, response } from './update-router';
+import type { RequestContext, ResponseContext, JavaMap } from '../context';
+
+/** Build a RequestContext with a nested Map body (auto-converts objects to Maps). */
```
k-rooot: Can we add test coverage for 
1. unknown command payload
2. malformed add.doc
3. /update with no command but valid JSON
4. response transform doesn’t test partial/malformed responses
5. unexpected result values
6. response cleanup completeness

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
```diff
@@ -0,0 +1,24 @@
...
+ * OpenSearch:  DELETE /{collection}/_doc/1
+ */
+import type { MicroTransform } from '../pipeline';
+import type { RequestContext } from '../context';
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'delete-doc',
+  apply: (ctx) => {
+    const id = ctx.body.get('id');
+    if (id == null || id === '') {
```
k-rooot: This allows `id = '   '` which becomes `/_doc/%20%20%20`. Should we trim?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
```diff
@@ -0,0 +1,24 @@
...
+      throw new Error('[delete-doc] Delete command must have a non-empty "id" field');
+    }
+
+    ctx.msg.set('URI', `/${ctx.collection}/_doc/${encodeURIComponent(String(id))}`);
```
k-rooot: If collection is undefined `/undefined/_doc/1`. Should we add a validation here?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
```diff
@@ -0,0 +1,24 @@
...
+    ctx.msg.set('method', 'DELETE');
```
k-rooot, line 34: 1. As per Solr behaviour, `{"delete":{"id":"1","_route_":"x","version":123}}`

_route_ may be required for correct shard routing and versioning may matter for concurrency. Even if not supported, silently ignoring is risky.

2. is it intentional to overwrite method unconditionally?

k-rooot (changes requested):

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
niravpi: Fixed `handleAdd` now throws `"[update-router] add: document-level boost is not supported — OpenSearch has no equivalent"`. Updated LIMITATIONS.md. Added unit test + e2e test.

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
niravpi: Fixed  throws `"[update-router] add: overwrite=false is not supported — OpenSearch always overwrites"` when `overwrite` is explicitly `false`. `overwrite=true` passes through. Updated LIMITATIONS.md. Added unit test + e2e test.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Fixed, added empty body validation in the router before delegating to `/update/json/docs`. The router now throws `"[update-router] json-docs: request body is empty"` for empty/missing bodies, acting as a gatekeeper. `update-doc.ts` still has its own validation as a defense-in-depth layer.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Replaced `commandData?.get?.('0')` with `isArrayLike()` helper using `Symbol.iterator` check. Excludes strings and Maps. Works for both JS arrays and GraalVM Java ArrayLists.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: >The error gives no visibility into what was actually received. Can we include the keys from the request body in the error message. This will make debugging significantly easier.

Added `bodyKeys()` helper. Unrecognized command error now includes actual keys: `"[update-router] dispatch: no recognized command in request body (keys: foo, baz)"`.

>Across the file, error messages vary in tone and structure:* "Array/bulk operations are not supported yet"* "Add command must contain a "doc" field"

All error messages now follow [update-router] <operation>: <reason> pattern consistently. Examples: [update-router] delete: delete-by-query is not supported yet, [update-router] dispatch: request body is empty or not JSON, [update-router] add: document-level boost is not supported.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Ack, handleDelete now explicitly validates that only supported fields are present. Added a SUPPORTED_DELETE_FIELDS Set containing ['id']. The handler iterates all keys in the delete payload and fails fast on any unsupported field:  [update-router] delete: unsupported field '_route_' only 'id' is supported
This ensures Solr-specific fields like _route_, version, and commitWithin are never silently dropped. As we add support for more fields, we simply add them to the Set.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: In Solr, deleting a non-existent doc returns `{"responseHeader":{"status":0}}`. This is Solr's documented idempotent delete behavior.

Regarding HTTP status code: OpenSearch returns HTTP 404 with `result: "not_found"` for non-existent docs, vs HTTP 200 with `result: "deleted"` for existing docs. The status code is available in the response map via `ctx.response.get('code')`, though we currently use the `result` field for simplicity since it provides the same information.
Added detailed code comment documenting the explicit result mapping:
  > created/updated/deleted → status 0 (success)
  > not_found → status 0 (Solr treats delete of missing doc as success — verified against Solr 8/9)
  > anything else → status 1 (error)

This is an explicit, documented decision matching Solr's actual behavior, not an inferred default.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Agree that hardcoding 0 is not ideal. However, OpenSearch's _doc API response doesn't include any processing time field — unlike the _search API which returns took. We have three options:
   1. `QTime: 0` (current) — technically inaccurate but Solr clients expect a number
   2. `QTime: -1` — signals "not available" but may break clients that assume non-negative
   3. Omit `QTime` — but Solr always includes it, and some clients may fail on missing field
We went with 0 to match the pattern used by the existing response-header.ts for select queries (which also sets QTime: 0). Added a code comment making this explicit.

Open to changing to -1 if the team prefers a clearer "not available" signal, but that would need to be consistent across all response transforms (select also uses 0 today). We can take that up in follow up CR if needed

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Fixed, tightened regex from /\/update\/json\/docs/ to /\/update\/json\/docs(\?|$)/. Now only matches paths ending with /docs or /docs?... (query string). Paths like /update/json/docs123 no longer match. Added
unit test verifying this.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Added Content-Type validation at the top of the router, before any routing logic. Non-JSON content types now fail fast: [update-router] dispatch: only JSON content type is supported, got 'text/xml'

 Regarding the ```parse → normalize → dispatch pattern``` agreed this will be needed for bulk and atomic updates. The current detect → dispatch is sufficient for single-command operations, but the handler registry
   map we introduced (COMMAND_HANDLERS) is designed with this evolution in mind. When we add bulk support, the normalization step would sit between command identification and handler dispatch extracting
   individual operations from arrays before routing each to the appropriate handler. No architectural changes needed, just an additional step in the pipeline.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Done, replaced the switch statement with a COMMAND_HANDLERS registry map

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/registry.ts
niravpi: The pipeline has two separate registries requestRegistry for transforming incoming requests and responseRegistry for transforming outgoing responses. They have different type signatures (RequestContext vs ResponseContext) and can't be combined.

   This is the same pattern used by select, which has 8 request transforms and 5 response transforms registered separately. Every endpoint that transforms both directions needs entries in both registries.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/update-router.testcase.ts
niravpi: Added side-effect verification tests using requestSequence — these perform the write operation, then query both Solr and OpenSearch via the shim to verify the actual data:

   - update-cmd-add-then-query: adds a doc via /update, then queries to verify it exists and matches Solr
   - update-cmd-delete-then-query: seeds a doc, deletes via /update, then queries to verify it's gone in both backends

Also found and fixed a bug during this delete-doc.ts wasn't translating ```commit=true → ?refresh=true```, so deleted docs weren't immediately unsearchable in OpenSearch. So thanks for pointing this out.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
niravpi: Intentional, the Java shim wraps all transform errors into a generic response: Primary target 'opensearch' failed: Request transform failed for target 'opensearch'

   If we want more specific error matching in e2e, that would require changing the Java shim's error response format to include the original error message which can be taken as a separate enhancement PR.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
niravpi: Ack, removed the duplicate. Replaced it with a body preservation test that verifies document fields (`title`, `price`) are correctly passed through when routing `/update/json/docs` to the update-doc handler.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
niravpi: Ack, changed to use a real JS array instead of simulating with Map numeric keys

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.test.ts
niravpi: Ack, Added all

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
niravpi: Ack, added String(rawId).trim() in delete-doc.ts. Whitespace-only ids now fail fast with "non-empty id" error. Added unit test for '   ' id.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
niravpi: Ack, added if (!ctx.collection) throw guard at the top of delete-doc.ts

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-doc.ts
niravpi: Two parts:

   1. `_route_`/`version` :  Fixed. handleDelete now validates only supported fields using SUPPORTED_DELETE_FIELDS Set. Any unsupported field fails fast: "[update-router] delete: unsupported field '_route_' — only 'id' is supported".
   2. Method overwrite : Yes, intentional. Solr's /update only accepts POST. We convert to the OpenSearch equivalent: DELETE for delete operations, PUT for add operations. Added POST method validation in the router non-POST requests now fail fast: "[update-router] dispatch: only POST method is accepted for /update, got 'GET'".

k-rooot (approved):

PR #2749, opened 2026-04-17
BugFix for prefix sequence queries

nagarajg17:
### Description
When we have prefix terms like `+`, `-` ex: `+jakarta +lucene` => it means to search for documents that must contain "jakarta" and  "lucene,". When we don't have prefix terms ex: `jakarta lucene` => we search for docs that contain either jakarta or lucene or both. By default space works like OR operator but when we've prefix operators it is implicit AND. Fixing bug to consider as implicit AND when we've prefix operators

Sample -
| Query | Before | After |
|-------|--------|-------|
| `+A +B` | A OR B | A AND B |
| `+A -B` | A OR (NOT B) | A AND NOT B |

### Testing
UTs and Integ

### Check List
-  [x] New functionality includes testing

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/boolNode.parser.test.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/boolNode.parser.test.ts
@@ -241,15 +241,13 @@ describe('parseSolrQuery – BoolNode', () => {
   it('parses mixed + and - prefixes', () => {
     const { ast, errors } = parseSolrQuery('+title:java -status:draft', emptyParams);
     expect(errors).toEqual([]);
-    // Two adjacent terms → implicit OR
+    // Consecutive prefix operators are merged: +A -B → { and: [A], not: [B] }
+    // This matches Solr semantics where +/- prefixed terms form an AND relationship
     expect(ast).toEqual({
       type: 'bool',
-      and: [],
-      or: [
-        { type: 'bool', and: [{ type: 'field', field: 'title', value: 'java' }], or: [], not: [] },
-        { type: 'bool', and: [], or: [], not: [{ type: 'field', field: 'status', value: 'draft' }] },
-      ],
-      not: [],
+      and: [{ type: 'field', field: 'title', value: 'java' }],
+      or: [],
+      not: [{ type: 'field', field: 'status', value: 'draft' }],
     });
   });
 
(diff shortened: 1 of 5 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/solr.pegjs
```diff
@@ -71,11 +71,33 @@ unaryExpr
...
 
+// Prefix sequence: consecutive +/- prefixed terms form an AND relationship.
+// In Solr, `+A +B` means "A required AND B required", not "(+A) OR (+B)".
+// This rule directly builds a BoolNode with and/not arrays from the prefix operators.
+//
+// Examples:
+//   `+A +B` → BoolNode { and: [A, B], or: [], not: [] }
+//   `-A -B` → BoolNode { and: [], or: [], not: [A, B] }
+//   `+A -B` → BoolNode { and: [A], or: [], not: [B] }
+prefixSequence
```
k-rooot, line 87: 1. This only triggers when all adjacent terms are prefixed. Mixed sequences fall back to orExpr implicit OR. 
```
+A B
A +B
+A B C
```

New parsing:
```
(+A) OR B
A OR (+B)
```

Expected parsing
```
A AND B
A AND B
A AND B AND C
```

2. Do we need to handle for interleaved sequences like `+A +B C +D`

Current behaviour
```
+A +B → prefixSequence
C +D → broken into OR
```

can become 
```
A AND B AND C AND D
```

This reinforces that grammar alone won’t fully solve this and need AST normalization.

3. We may also need to handle an edge case for prefix + explicit AND scenarios
```
+A AND +B
```

will get convered to 
```
+A → unaryExpr
AND
+B → unaryExp
```

This can be changed to `AND: [A, B]`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/solr.pegjs
```diff
@@ -71,11 +71,33 @@ unaryExpr
...
+  = head:prefixWithType tail:(__ prefixWithType)+ {
+      const all = [head, ...tail.map(t => t[1])];
+      const and = all.filter(p => p.op === '+').map(p => p.expr);
+      const not = all.filter(p => p.op === '-').map(p => p.expr);
+      return { type: 'bool', and, or: [], not };
+    }
+  / prefixExpr
```
k-rooot, line 94: For `prefixWithType` and `prefixExpr`, were you able to verify whether we can parse parse grouped or nested expressions?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/solr.pegjs
```diff
@@ -71,11 +71,33 @@ unaryExpr
...
+// Examples:
+//   `+A +B` → BoolNode { and: [A, B], or: [], not: [] }
+//   `-A -B` → BoolNode { and: [], or: [], not: [A, B] }
+//   `+A -B` → BoolNode { and: [A], or: [], not: [B] }
+prefixSequence
+  = head:prefixWithType tail:(__ prefixWithType)+ {
+      const all = [head, ...tail.map(t => t[1])];
+      const and = all.filter(p => p.op === '+').map(p => p.expr);
+      const not = all.filter(p => p.op === '-').map(p => p.expr);
+      return { type: 'bool', and, or: [], not };
```
k-rooot, line 92: You are bypassing `implicit: true` flag used elsewhere. Should we return the following instead if the downstream rely on this detail 

```
return { type: 'bool', and, or: [], not, implicit: false };
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/boolNode.parser.test.ts
```diff
@@ -311,6 +309,69 @@ describe('parseSolrQuery – BoolNode', () => {
       not: [{ type: 'bare', value: 'draft', isPhrase: false, defaultField: 'status' }],
     });
   });
+
+  // ─── Prefix sequence merging ───────────────────────────────────────────
```
k-rooot, line 313: Can we add test coverage for 
1. Fallback path - test to ensure single prefix still behaves correctly after this routing change
2. Verify sequence only applies to consecutive prefixes to confirm that non consecutive prefixes do not get merged.
3. Precedence validation specific to prefixSequence
4. Explicit OR override check for prefixed terms

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/solr.pegjs
nagarajg17: The implicit flag only matters when or is non-empty — applyDefaultOperator checks boolNode.implicit && node.or.length > 0. Since prefixSequence always returns or: [], the flag has no effect regardless of its value. But yes added it for avoiding confusion

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/solr.pegjs
nagarajg17: 1. That is expected behaviour. `+jakarta lucene` => to search for documents that must contain "jakarta" and that may or may not contain "lucene" https://solr.apache.org/guide/solr/latest/query-guide/standard-query-parser.html#the-boolean-operator
2.  Nice edge case. Above answer might explain this as well
3. Correct. Current AST output adds one layer of nesting which can be optimized. We've a separate task in tracker for it `Bool query optimizations via flatenning` with P2 priority. We can do that generically in that task. 1 layer of nesting performance different is quite negligible. We can take a call in that task if we want to do that because it adds lot of conditional statements and code complexity for very marrginal or negligible gain
```
// Something like this we need to do

const flattenInto = (nodes: ASTNode[], targetClause: 'and' | 'or' | 'not'): ASTNode[] =>
    nodes.flatMap(n => {
      if (n.type !== 'bool') return [n];
      const hasAnd = n.and.length > 0, hasOr = n.or.length > 0, hasNot = n.not.length > 0;
      const clauseCount = [hasAnd, hasOr, hasNot].filter(Boolean).length;
      if (clauseCount !== 1) return [n]; // mixed or empty — don't flatten
      if (targetClause === 'and' && hasAnd) return n.and;
      if (targetClause === 'or'  && hasOr)  return n.or;
      if (targetClause === 'not' && hasNot) return n.not;
      return [n];
    });
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/solr.pegjs
nagarajg17: Yes — both prefixWithType and prefixExpr delegate to primary, and primary's first alternative is group which handles any parenthesized sub-expression including nested booleans. So +(A OR B), -(A AND B), and arbitrarily deep nesting all work correctly. Added tests for same

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/boolNode.parser.test.ts
nagarajg17: Added

k-rooot (approved):

PR #2771, opened 2026-04-21
Reconstruct _source for disabled and partial-source indices during migration

AndreKurait:
## Summary

Adds support for migrating Elasticsearch indices where `_source` is disabled (`_source: {enabled: false}`) or filtered (`_source.includes`/`excludes`). When `_source` bytes aren't usable, documents are reconstructed from the Lucene segment — stored fields, doc_values, points, inverted-index terms, and (on ES 7+/OpenSearch) `_recovery_source`.

Gated behind `--enable-sourceless-migrations`. Without the flag, metadata migration fails fast with a clear error listing the offending indices. Indices with full `_source` keep the normal fast path unchanged.

## Background: `_source` variants

```
Full source (default):    _source holds the complete original JSON
Disabled source:          _source: { enabled: false }     → no JSON at all
Partial source:           _source: { includes: [...] }    → only listed fields
                          _source: { excludes: [...] }    → listed fields dropped
```

Normal migration reads `_source` verbatim. Disabled source has nothing to read; partial source is missing fields. Both need reconstruction from the Lucene segment.

## Problems solved

1. **Sourceless indices silently dropped** — documents from `_source: disabled` indices produced empty bytes and were dropped with a warning. Now reconstructed from the segment.
2. **Partial source lost excluded fields** — `isSourceEnabled()` only checked the `enabled` flag, so `includes`/`excludes` indices migrated with fields missing. Now `needsSourceReconstruction()` covers both cases and `mergeWithDocValues()` fills in the gaps.
3. **No fail-fast for users** — running migration against sourceless indices produced confusing partial results. Metadata migration now rejects them upfront unless the flag is set.

### Source detection (`IndexMetadata`)
`isSourceEnabled()` now walks all mapping formats (ES 1.x typed → 7.x+ typeless), with `getSourceNode()` unwrapping the array-wrapped shape ES 5.x/6.x emits. New `isSourcePartial()` detects `includes`/`excludes`. `needsSourceReconstruction()` combines both, consumed by `DocumentMigrationBootstrap` and `MigratorEvaluatorBase`.

### Reconstruction pipeline (`LuceneReader`)
`resolveSourceBytes()` routes through three paths: source present + no mapping context → passthrough (unchanged); source present + context → `mergeWithDocValues`; no source + context → full `reconstructSource`. `FieldMappingContext` is built per-index and cached.

### `SourceReconstructor` recovery sources (in priority order)
- `_recovery_source` — ES 7+ / OpenSearch fast path (soft-deletes default-on)
- Stored fields and doc_values — all versions
- Points — numeric, date, IP
- Trie-encoded terms — ES 1.x–5.x numerics/IPs with no store or doc_values
- Inverted-index term positions via `SegmentTermIndex` — lossy text body recovery with word order preserved
- `mergeWithDocValues()` — partial-source merge that only adds missing fields

### CLI + orchestration
`--enable-sourceless-migrations` on `MigrateOrEvaluateArgs` and `RfsMigrateDocuments`, plumbed through the Argo workflow schemas. Defaults to off.

## Testing

- **`SourceReconstructorTest`** — 51 unit tests (925 lines) covering every field type (keyword, text, long, int, double, float, scaled_float, boolean, date, date_nanos, ip, geo_point, binary) × every recovery path (stored, doc_values, points, trie terms, positions) plus edge cases: IP normalization, boolean encoding, scaled_float factor, merge semantics.
- **`NoStoredSourceMigrationTest`** — 4 parameterized matrix tests × 9 source versions (ES 1.7, 2.4, 5.6, 6.8, 7.10, 8.17, OS 1.3, 2.19) → OS 3.5, i.e. 36 CI cases. Covers 15+ field types × 4 storage permutations (dv+store, dv-only, store-only, neither), partial-source merge with doc_values and stored fields, and the unrecoverable-document skip path.
- **`SourcelessMigrationTest`** — E2E pipeline against a realistic email-archive index (10 field types, 3 documents). Validates field fidelity end-to-end and the fail-fast behavior when the flag is omitted.

## Compatibility

Default behavior is unchanged — the flag is off by default. Indices with full `_source` never enter the reconstruction path.


AndreKurait:
Large BYOS run https://migrations.ci.opensearch.org/job/pr-checks/job/pr-eks-byos-integ-test/101/ finished within margin of error (vs baseline main commit at https://migrations.ci.opensearch.org/job/pr-checks/job/pr-eks-byos-integ-test/106/

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
jugal-chauhan: This function is currently skipping ALL fields containing `.`. This is fine for `title.keyword` which is a property. But it also drops `address.city` which is an object subfield (The `FieldMappingContext` correctly parses `properties` and not `fields` which can be useful here)

Suggestion : 
```
private static boolean shouldSkipField(String fieldName, FieldMappingContext mappingContext) {
    if (fieldName.startsWith("_")) return true;
    if (!fieldName.contains(".")) return false;
    // Dotted field: skip only if it's NOT a known field in the mapping
    return mappingContext == null || mappingContext.getFieldInfo(fieldName) == null;
}
```

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -33,125 +33,119 @@ private static boolean shouldSkipField(String fieldName) {
...
+    private static void populateFromSegment(Map<String, Object> target, LuceneLeafReader reader, int docId,
+            LuceneDocument document, FieldMappingContext mappingContext, SegmentTermIndex termIndex) throws IOException {
+        // 1. Stored fields (exact values when present)
+        for (var field : document.getFields()) {
+            String fieldName = field.name();
+            if (shouldSkipField(fieldName) || target.containsKey(fieldName)) {
+                continue;
+            }
+            FieldMappingInfo mappingInfo = mappingContext != null ? mappingContext.getFieldInfo(fieldName) : null;
+            Object value = getStoredFieldValue(field, mappingInfo);
```
jugal-chauhan, line 209: In all 3 loops present  under `populateFromSegment()`. the return value of `mappingContext.getFieldInfo(fieldName)` should include `address.city` but `shouldSkipField` is filtering it out

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -33,125 +33,119 @@ private static boolean shouldSkipField(String fieldName) {
...
+     * same stored → doc_values → points/terms recovery chain.
+     */
+    private static void populateFromSegment(Map<String, Object> target, LuceneLeafReader reader, int docId,
+            LuceneDocument document, FieldMappingContext mappingContext, SegmentTermIndex termIndex) throws IOException {
+        // 1. Stored fields (exact values when present)
+        for (var field : document.getFields()) {
+            String fieldName = field.name();
+            if (shouldSkipField(fieldName) || target.containsKey(fieldName)) {
+                continue;
+            }
```
jugal-chauhan: Can we use `target.put` here ? Already implemented in [SolrConfigProvider.java](https://github.com/opensearch-project/opensearch-migrations/blob/7a3d6fb16ed8cf9a6f6bd61c6a8ebd93187292a3/TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrConfigProvider.java#L93)

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -33,125 +33,119 @@ private static boolean shouldSkipField(String fieldName) {
...
+                target.put(fieldName, value);
+            }
+        }
+
+        // 2. Doc values (fast, lossless for typed fields)
+        for (DocValueFieldInfo fieldInfo : reader.getDocValueFields()) {
+            String fieldName = fieldInfo.name();
+            if (shouldSkipField(fieldName) || target.containsKey(fieldName)) {
+                continue;
+            }
```
jugal-chauhan: Same concern here. I am concerned about `containsKey` doing a single match on the first field, and skipping over the remaining ones

jugal-chauhan (commented):
Thanks for brining in these changes! I like how the design has been implemented for this option `enableSourcelessMigrations`. I have left a couple of comments for the stored fields path and on mapping context. I have not yet looked at why these junit tests are failing

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/DocumentMigrationBootstrap.java
```diff
@@ -279,11 +296,34 @@ private DocumentSource createDocumentSource() {
...
+                            name, !meta.isSourceEnabled(), meta.isSourcePartial());
+                        return new FieldMappingContext(meta.getMappings());
+                    } catch (Exception e) {
+                        throw new RuntimeException(
+                            "Failed to read metadata for index " + name +
+                            " — cannot determine if source reconstruction is needed", e);
+                    }
+                });
+            });
+        }
```
jugal-chauhan, line 332: The `sourcelessMappingContextProvider` lambda uses `ConcurrentHashMap.computeIfAbsent` to cache `FieldMappingContext` per index name. 

- For sourceless indices this works perfectly : `FieldMappingContext` is non-null and gets cached. 
- But for normal indices (`_source` enabled), the lambda returns `null` to signal `"no reconstruction needed"`

Issue is that `ConcurrentHashMap.computeIfAbsent` does not store `null` returns raising `NullPointerException`. This means `indexMetadataFactory.fromRepo()` which reads and parses snapshot metadata from disk is being re-executed on every shard of every normal index instead of once per index.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/DocumentMigrationBootstrap.java
```diff
@@ -279,11 +296,34 @@ private DocumentSource createDocumentSource() {
...
+        if (enableSourcelessMigrations && indexMetadataFactory != null) {
+            log.info("Sourceless migrations enabled — will reconstruct _source from stored fields/doc_values");
+            Map<String, FieldMappingContext> cache = new ConcurrentHashMap<>();
+            builder.sourcelessMappingContextProvider(indexName -> {
+                return cache.computeIfAbsent(indexName, name -> {
+                    try {
+                        var meta = indexMetadataFactory.fromRepo(snapshotName, name);
+                        if (!meta.needsSourceReconstruction()) {
+                            log.debug("Index {} has full _source enabled, no reconstruction needed", name);
+                            return null;
```
jugal-chauhan: Suggestion using sentinel : 
```
private static final FieldMappingContext NO_RECONSTRUCTION_NEEDED = new FieldMappingContext(null);

var result = cache.computeIfAbsent(indexName, name -> {
    var meta = indexMetadataFactory.fromRepo(snapshotName, name);
    if (!meta.needsSourceReconstruction()) {
        return NO_RECONSTRUCTION_NEEDED;
    }
    return new FieldMappingContext(meta.getMappings());
});
return result == NO_RECONSTRUCTION_NEEDED ? null : result;
```

jugal-chauhan (approved):
The overall design of this feature is really well implemented by the two-flag approach (`--enable-sourceless-migrations` for the reconstruction pipeline, `--use-recovery-source` as an opt-in for the transient ES 7+ field) along with the fail-fast validation during metadata migration that lists exactly which indices are affected. 

I have left a comment below for the happy-path where `_source` is enabled. Happy to see this addressed in a follow up PR as well, as it is not related to correctness.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/DocumentMigrationBootstrap.java
AndreKurait: Merged in a simplifying change in https://github.com/opensearch-project/opensearch-migrations/pull/2771/changes/6059a1e3ab8dd614c61cb169b10adbd688242ad6

jugal-chauhan: Thanks! Lets merge in this PR after GHA passes

PR #2775, opened 2026-04-21
orchestrationSpecs: add serializeSnapshotCreation override, clarify indexAllowlist descriptions

AndreKurait:
## Summary

Turns snapshot serialization into an explicit, overridable config on the source cluster, instead of something implicitly derived from the source version. Also sharpens the documentation on the three `indexAllowlist` fields (CreateSnapshot / Metadata / RFS) so users can tell which one filters server-side vs client-side.

## Background: how snapshot serialization is gated

A `ConfigMap` named `concurrency-config` backs an Argo semaphore attached at the **`snapshotworkflow` template level** — so the lock covers the entire workflow invocation, not just the create step. That means one holder at a time owns:

```
runCreateSnapshot  →  getConsoleConfig  →  waitForCompletion  →  patchDataSnapshotCompleted
```

Including the polling wait. This matters for legacy repos that can't tolerate *any* concurrent snapshot op, not just concurrent starts.

The semaphore key granularity was hardcoded by `isLegacyVersion(sourceVersion)`:

```
serialize=true   →  snapshot-legacy-<sourceName>              (one key shared per source)
serialize=false  →  snapshot-modern-<sourceName>-<snapshotName> (one key per snapshot)
```

Version-based default is usually right but wrong for modern clusters whose repo or indices don't support concurrent snapshots.

## Problems Solved

1. **No opt-in to serialization for modern sources** — e.g. OS 2.x with UltraWarm indices (one index per snapshot, no concurrency) was stuck in the parallel bucket.
2. **No opt-out for legacy sources** — operators who know their ES 6/7 repo tolerates concurrent snapshots were stuck sharing one key.
3. **CreateSnapshot `indexAllowlist` was mis-documented** — the description claimed `regex:` support. It doesn't — entries are comma-joined and sent verbatim as the `_snapshot` API's `indices` field (native multi-index syntax, wildcards, `-` exclusions; no regex). Only RFS/Metadata allowlists support `regex:`.
4. **No framing of where each allowlist applies** — three fields share the same name with no hint whether filtering happens server-side at snapshot time (irreversible) or client-side on snapshot contents (re-narrowable later).

### New optional flag: `sourceClusters.<name>.snapshotInfo.serializeSnapshotCreation`

Boolean, optional. Omitted = legacy defaults to `true`, modern to `false` (prior behavior preserved). Setting it explicitly overrides in either direction.

### Shared resolver keeps config and workflow in lockstep

`semaphoreUtils.ts` exposes `resolveSerializeSnapshotCreation(version, override)`, and `generateSemaphoreKey()` now takes the resolved boolean rather than the version string. Both the `concurrency-config` builder (`MigrationInitializer`) and the workflow template reference (`MigrationConfigTransformer`) call the resolver, so they can't disagree about what key a workflow will request.

Existing key prefixes (`snapshot-legacy-…`, `snapshot-modern-…`) are unchanged to avoid churning the 7 workflow-template snapshot fixtures.

### `indexAllowlist` descriptions clarified

All three now say explicitly *where* filtering happens and whether `regex:` is supported:

- **CreateSnapshot** — server-side at snapshot time; native `_snapshot` syntax (wildcards, `-` exclusions); no regex. Excluded indices can't be recovered downstream.
- **Metadata** — client-side on snapshot contents; supports `regex:`.
- **RFS** — client-side on snapshot contents; supports `regex:`.

## Testing

- 2 new unit tests in `semaphore.test.ts` covering the override in both directions (modern forced to serialize → one shared key; legacy forced parallel → distinct keys per snapshot).
- Updated `argoSchemaSnapshot.test.ts.snap` for the new schema field + rewritten descriptions.
- Full orchestrationSpecs workspace: 206 passing (+2 skipped) across 4 packages. No workflow-template fixture diffs.

## Migration notes

Fully backward-compatible. To force serialization on a modern source:

```yaml
sourceClusters:
  mySource:
    version: "OS 2.11"
    snapshotInfo:
      serializeSnapshotCreation: true
      snapshots: { ... }
```

### DIFF
--- a/orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
+++ b/orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
@@ -15,7 +15,7 @@ import {StreamSchemaTransformer} from './streamSchemaTransformer';
 import { z } from 'zod';
 import {promises as dns} from "dns";
 import {createHash} from "crypto";
-import { generateSemaphoreKey } from './semaphoreUtils';
+import { generateSemaphoreKey, resolveSerializeSnapshotCreation } from './semaphoreUtils';
 import {validateInputAgainstUnifiedSchema} from "./unifiedSchemaValidator";
 
 type InputConfig = z.infer<typeof OVERALL_MIGRATION_CONFIG>;
@@ -598,7 +598,12 @@ export class MigrationConfigTransformer extends StreamSchemaTransformer<
                 const proxyDeps = proxyNamesBySource.get(sourceName);
 
                 const { snapshotPrefix: _sp, ...createSnapshotOpts } = snapshotDef.config.createSnapshotConfig;
-                const semaphore = this.generateSemaphoreConfig(sourceCluster.version, sourceName, snapshotName);
+                const semaphore = this.generateSemaphoreConfig(
+                    sourceCluster.version,
+                    sourceName,
+                    snapshotName,
+                    snapshotInfo?.serializeSnapshotCreation
+                );
                 createConfigs.push({
                     label: snapshotName,
                     snapshotPrefix: snapshotDef.config.createSnapshotConfig.snapshotPrefix || snapshotName,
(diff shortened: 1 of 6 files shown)


oeyh (approved):
LGTM!

PR #2779, opened 2026-04-22
Enable TLS by default for Argo Workflows server

oeyh:
### Description

Enable TLS on the Argo Workflows server by default using the built-in `--secure` mode. The server generates a self-signed certificate at startup and serves HTTPS. All migration console CLI commands now default to `https://` with TLS verification disabled, since all Argo server communication is internal (migration console to Argo server within the cluster) and server identity verification is not required.

Changes:
- Set `server.secure: true` in the Argo Helm chart values (the chart handles both the `--secure` flag and HTTPS health probe scheme)
- Extract shared `DEFAULT_ARGO_SERVER_URL` constant into `argo_utils.py` to eliminate URL duplication across 5 CLI command files
- Change default URL scheme from `http://` to `https://` in all workflow CLI commands
- Default `--insecure` flag to `True` for Argo server calls
- Fix insecure default in autocomplete_k8s_labels.py
- Update unit tests to reflect new defaults

### Issues Resolved

[MIGRATIONS-3040](https://opensearch.atlassian.net/browse/MIGRATIONS-3040)

### Testing

**Unit tests**: 44 tests updated and passing — URL scheme changed from `http://` to `https://`, `--insecure` default flipped to `True`.

**EKS integration tests** (deployed to dev account, us-east-1):

- `curl -sk https://localhost:2746/api/v1/info` returns valid JSON; `curl http://` rejected with "Client sent an HTTP request to an HTTPS server"
- Deployment args contain `--secure=true` with no conflicting `--secure=false`
- Helm chart automatically sets readiness/liveness probe `scheme: HTTPS` when `server.secure: true`; health endpoint returns 200
- `workflow status` from migration console pod connects to `https://<argo-server>:2746` with `insecure=True`
- `workflow submit` and `workflow status` communicate with Argo API over HTTPS successfully
- `curl https://argo-server:2746` without `-k` fails with exit code 60 (cert verification failed), confirming real TLS with self-signed cert

### Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ TrafficCapture/dockerSolution/src/main/docker/migrationConsole/staging
 
 # Directory used to temporarily store snapshots during testing
 DocumentsFromSnapshotMigration/docker/snapshots/
-transformation/standardJavascriptTransforms/coverage/
+**/coverage/
 transformation/standardJavascriptTransforms/node_modules/
 orchestrationSpecs/k8sResources/
 orchestrationSpecs/dist/
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ TrafficCapture/dockerSolution/src/main/docker/migrationConsole/staging
 
 # Directory used to temporarily store snapshots during testing
 DocumentsFromSnapshotMigration/docker/snapshots/
-**/coverage/
+transformation/standardJavascriptTransforms/coverage/
 transformation/standardJavascriptTransforms/node_modules/
 orchestrationSpecs/k8sResources/
 orchestrationSpecs/dist/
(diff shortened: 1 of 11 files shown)


jugal-chauhan (commented):
Thanks for brining in these changes!

Does these files need to update ?
- `forwardAllServicePorts.sh` 
- `autocomplete_k8s_labels.py` has insecure default still False

oeyh:
@jugal-chauhan Good catch on `autocomplete_k8s_labels.py`, fixed it in [9aeacea](https://github.com/opensearch-project/opensearch-migrations/pull/2779/commits/9aeaceafb325a56cbc1945c12b907d5535cebb15).
`forwardAllServicePorts.sh` should be good as the port (2746) is not changing.

jugal-chauhan (approved):
LGTM!

PR #2757, opened 2026-04-20
Adding translator for query-fields param

nagarajg17:
### Description
Adding translator for [query-fields](https://solr.apache.org/guide/solr/latest/query-guide/dismax-query-parser.html#qf-query-fields-parameter) param. Dismax and eDismax supports query fields param

### Check List
- [ ] New functionality includes testing
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2781, opened 2026-04-22
feat(solr-transforms): Add mm (minimum should match) parameter support

niravpi:
### Summary

   Adds support for Solr's `mm` (minimum should match) parameter, translating it to OpenSearch's `minimum_should_match` on the `bool` query. The `mm` value is a direct passthrough since both Solr and OpenSearch share identical syntax.

   Only applies when `defType=dismax` or `defType=edismax`, matching Solr's behavior where the standard (lucene) query parser ignores `mm`.

   ### Changes

   | File | Description |
   |---|---|
   | `features/minimum-match.ts` | Request transform: reads `mm`, sets `minimum_should_match` on the bool query. Gated on `defType`. Includes `paramRules` for character validation. |
   | `features/minimum-match.test.ts` | 20 unit tests: match guard, apply with all mm formats (int, -int, %, -%, conditional, multi-conditional), no-op cases, paramRules validation |
   | `integration-tests/minimum-match.testcase.ts` | 12 E2E tests: 5 dismax + 5 edismax happy-path, `mm.autoRelax` rejection, invalid value rejection |
   | `registry.ts` | Register transform in `FEATURE_MODULES` and `requestRegistry` (after `queryQ`) |
   | `docs/LIMITATIONS.md` | Document 3 limitations: MM-AUTORELAX, MM-EDISMAX-OPERATOR-DEFAULT, MM-STOPWORD-MULTIFIELD |

   ### Supported mm syntax (all 6 Solr formats)

   | Format | Example | Behavior |
   |---|---|---|
   | Positive integer | `mm=3` | At least 3 clauses must match |
   | Negative integer | `mm=-2` | Total clauses minus 2 must match |
   | Percentage | `mm=75%` | 75% of clauses must match (rounded down) |
   | Negative percentage | `mm=-25%` | 25% of clauses can be missing |
   | Conditional | `mm=3<90%` | If ≤3 clauses all required, else 90% |
   | Multiple conditionals | `mm=2<-25% 9<-3` | Tiered conditions |

   ### Known limitations (documented in LIMITATIONS.md)

   - **MM-AUTORELAX**: `mm.autoRelax` is rejected (fail-fast). No OpenSearch equivalent — requires per-field stopword analysis visibility.
   - **MM-EDISMAX-OPERATOR-DEFAULT**: When `mm` is omitted and query contains explicit operators under `q.op=AND`, the default `mm` inference may differ from eDisMax. Workaround: set `mm` explicitly.
   - **MM-STOPWORD-MULTIFIELD**: Solr applies `mm` to the max optional clause count across fields with different analyzers. OpenSearch applies `minimum_should_match` at the `bool` query level. Edge case with divergent stopword lists.

   ### Testing

   | Layer | Result |
   |---|---|
   | Unit tests (vitest) | 719/719 passed (20 new) |
   | E2E integration (Solr 8 + 9) | All 12 mm tests passed on both versions |
   ### References

   - [Solr DisMax mm parameter](https://solr.apache.org/guide/solr/latest/query-guide/dismax-query-parser.html#mm-minimum-should-match-parameter)
   - [Solr eDisMax mm parameter](https://solr.apache.org/guide/solr/latest/query-guide/edismax-query-parser.html)
   - [OpenSearch minimum_should_match](https://opensearch.org/docs/latest/query-dsl/minimum-should-match/)

### DIFF
--- a/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
+++ b/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
@@ -22,6 +22,10 @@ the root cause, and provides a workaround where one exists.
 | [JSON-QUERIES](#json-queries)                   | JSON Request API `queries` key not supported |
 | [JSON-PARAM-PREFIX](#json-param-prefix)         | JSON Request API `json.<param>` prefix passthrough not supported |
 | [UPDATE-COMMANDS](#update-commands)             | Only `add` and `delete-by-id` commands supported; JSON only |
+| [MM-AUTORELAX](#mm-autorelax)                   | eDisMax `mm.autoRelax` parameter not supported |
+| [MM-EDISMAX-OPERATOR-DEFAULT](#mm-edismax-operator-default) | eDisMax `mm` default with explicit operators not fully replicated |
+| [MM-EDISMAX-MIXED-OPERATORS](#mm-edismax-mixed-operators) | eDisMax `mm` with mixed explicit and implicit operators applies to wrong scope |
+| [MM-STOPWORD-MULTIFIELD](#mm-stopword-multifield) | `mm` with per-field stopword removal differs from Solr |
 
 ---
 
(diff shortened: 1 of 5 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
```diff
@@ -0,0 +1,41 @@
...
+export const paramRules: ParamRule[] = [
+  {
+    name: 'mm',
+    type: 'rejectPattern',
+    pattern: String.raw`[^-\d%<\s]`,
+    reason: 'Invalid mm syntax — only digits, -, %, <, and spaces are allowed (e.g., "2", "75%", "-1", "3<90%", "2<-25% 9<-3")',
+  },
+];
+
+const DISMAX_TYPES = new Set(['dismax', 'edismax']);
```
nagarajg17, line 27: nit: QUERY_PARSER_Types?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
```diff
@@ -0,0 +1,41 @@
...
+  name: 'minimum-match',
+  match: (ctx) => ctx.params.has('mm') && DISMAX_TYPES.has(ctx.params.get('defType') ?? ''),
+  apply: (ctx) => {
+    const query = ctx.body.get('query');
+    if (!query || typeof query.get !== 'function') return;
+
+    const boolClause = query.get('bool');
+    if (!boolClause || typeof boolClause.set !== 'function') return;
+
+    boolClause.set('minimum_should_match', ctx.params.get('mm')!);
```
nagarajg17: 1. What if existing  bool query has this param set already? it will override. See boolRule.ts for context
2. nit: `!` might be redundant as we check in match clause for same

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/integration-tests/minimum-match.testcase.ts
```diff
@@ -0,0 +1,105 @@
...
+    }),
+
+    solrTest(`${defType}-mm-single-term`, {
+      description: 'mm with single term — always requires that term',
+      documents: docs,
+      requestPath: mmPath('java', defType, '100%'),
+      solrSchema: schema,
+      opensearchMapping: mapping,
+    }),
+  ];
```
nagarajg17, line 97: Could we add test case for `3<90%` and `2<-25% 9<-3` formats?

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
```diff
@@ -465,6 +468,7 @@ rejected by validation.
 
 ---
 
+<<<<<<< HEAD
```
nagarajg17: nit: Merge conflict?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'minimum-match',
+  match: (ctx) => ctx.params.has('mm') && DISMAX_TYPES.has(ctx.params.get('defType') ?? ''),
+  apply: (ctx) => {
+    const query = ctx.body.get('query');
+    if (!query || typeof query.get !== 'function') return;
+
+    const boolClause = query.get('bool');
+    if (!boolClause || typeof boolClause.set !== 'function') return;
```
nagarajg17, line 37: `minimum_should_matc` looks like non bool type also support. Check once https://docs.opensearch.org/latest/query-dsl/minimum-should-match/

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
niravpi: Kept as `DISMAX_TYPES`  as IMO `QUERY_PARSER_TYPES` would be too generic since it could imply lucene/standard as well, which don't support `mm`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/integration-tests/minimum-match.testcase.ts
niravpi: Ack, added it

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
niravpi: Ohh might have missed it, thanks for highlighting it. Fixed in latest revision

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
niravpi: >What if existing bool query has this param set already? it will override. See boolRule.ts for context

The `boolRule.ts` sets `minimum_should_match: 1` only on nested bools where `must` + `should` coexist this is a correctness fix for standard parser queries where OpenSearch would otherwise make `should` purely optional. In DisMax/eDisMax, the top-level query is a flat `should` only bool (bare terms become `dis_max` clauses, no `must`), so `boolRule` never sets `minimum_should_match` at that level — no conflict. If they do overlap (e.g., explicit `+` operators in DisMax producing `must` + `should`), the user's explicit `mm` should take precedence over the structural default.

>nit: ! might be redundant as we check in match clause for same

Ack, Updated it

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
niravpi: Checked. OpenSearch's `minimum_should_match` on `match` queries controls how many analyzed tokens must match within a single field that's a different semantic than Solr's `mm`, which controls how many query terms must match across the outer `bool` query. In DisMax, each bare term becomes a `dis_max` across `qf` fields, and `mm` applies to how many of those term-level clauses must match. Setting it on the `bool` is the correct mapping.
For single-term queries, no `bool` is produced (just a `dis_max`), so our transform no-ops which is correct since Solr clamps `mm` to 1 with a single clause anyway.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
nagarajg17: Dismax is default should but edismax allows standard query parser also which makes it tricky. edismax will support query like `A AND (B OR C)` but dismax won't. So edismax can have `minimum_should_match` in bool query. Here if `mm` param is there it would apply to `B OR C` clause only or if it ignores we need to check, applying on `A AND  (B OR C)` will produce incorrect results. Another variation q=`A AND B C D`, mm may apply to only C, D

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
nagarajg17: There are conversions where we don't use bool clause like ex:
`q=opensearch software & qf=name^3 description^1.5 category`,
it will translate to 
```
"query": {
    "multi_match": {
      "query": "opensearch software",
      "fields": ["name^3", "description^1.5", "category"],
      "type": "best_fields"
    }
  }
```

if we've mm param in this like `q=opensearch software & qf=name^3 description^1.5 category & mm=50%`, then translation would be
```
"query": {
    "multi_match": {
      "query": "opensearch software",
      "fields": ["name^3", "description^1.5", "category"],
      "type": "best_fields".
       "minimum_should_match": "50%"
    }
  }
```
Check this PR and doc

https://github.com/opensearch-project/opensearch-migrations/pull/2757

https://docs.opensearch.org/latest/query-dsl/full-text/multi-match/

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
niravpi: Good catch. I verified the actual DSL output for `q=A AND B C D&defType=edismax&qf=title` — our parser produces `bool { should: [bool{must:[A,B]}, C, D] }` rather than Solr's flat `bool { must: [A,B], should: [C,D] }`. This means `mm` applies to all 3 top-level `should` entries instead of just the optional C and D.
This is a pre-existing parser limitation the PEG grammar nests `A AND B` as a sub-expression rather than hoisting `and` clauses to the parent level (`boolRule.ts` has a TODO for AST flattening). The `mm` transform itself correctly sets`minimum_should_match` on the top-level `bool`; the issue is upstream in how the parser structures mixed operators.
For pure DisMax (no explicit operators), all terms are flat `should` clauses and `mm` works correctly. Documented this as `MM-EDISMAX-MIXED-OPERATORS` in LIMITATIONS.md. The AST flattening fix would be a separate PR since it affects all bool queries, not just `mm`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/minimum-match.ts
niravpi: Good point, but our query structure already handles this correctly. For multi-term queries like `q=opensearch software&qf=name^3 description^1.5 category`, the parser produces `BoolNode { or: [BareNode("opensearch"), BareNode("software")] }`, which transforms to:
 

   > {
   >   "bool": {
   >     "should": [
   >       { "multi_match": { "query": "opensearch", "fields": ["name^3", "description^1.5", "category"] } },
   >       { "multi_match": { "query": "software", "fields": ["name^3", "description^1.5", "category"] } }
   >     ]
   >   }
   > }
   >

Each term gets its own `multi_match`, wrapped in a `bool.should`. Our `mm` sets `minimum_should_match` on this `bool`, controlling how many term-level clauses must match which is the correct Solr semantics.
  
 A single `multi_match` with all terms combined (where `minimum_should_match` would control token-level matching) is not what we produce. Single-term queries do produce a standalone `multi_match`, but `mm` is meaningless with 1 clause Solr clamps it to 1 regardless.

nagarajg17 (approved):

PR #2782, opened 2026-04-22
Update default to 3-broker Kafka with anti-affinity for EKS resiliency

jugal-chauhan:
### Description
Workflow-managed Kafka clusters now default to a 3-broker, RF=3 / minISR=2 configuration with soft pod anti-affinity, so a single EKS node rotation (or any single broker going down) should not fail the migration.

### What changed

**`DEFAULT_AUTO_CREATE_KAFKA` in `userSchemas.ts`**
- `nodePoolSpecOverrides.replicas`: 1 → **3** (combined-mode KRaft pool)
- `offsets.topic.replication.factor`: 1 → **3**
- `transaction.state.log.replication.factor`: 1 → **3**
- `default.replication.factor`: 1 → **3**
- `min.insync.replicas`: 1 → **2**
- `transaction.state.log.min.isr`: 1 → **2**
- `topicSpecOverrides.replicas`: 1 → **3**
- `storage.size`: 1Gi → **2Gi** (smoke-test size; real workloads should override)
- Added **soft** pod anti-affinity on `kubernetes.io/hostname` so brokers
  prefer distinct nodes but still schedule on single-node dev clusters

With RF=3 + minISR=2, Strimzi's auto-generated PDB (`maxUnavailable: 1`) keeps the cluster writable through rolling restarts and single-broker failures.

**Documentation in the schema defaults**
Added concise inline comments explaining the RF/minISR choice, the ISR definition, the 2Gi storage rationale, and why anti-affinity is soft instead of required.

**Dead code cleanup (follow-up to #2842)**
- Removed the orphaned `cleanup_cdc_resources()` function in `cdc_base.py` and its `_kubectl_delete` helper. #2842 already routed CDC test teardown through `workflow reset --all --include-proxies --delete-storage`; this removes the stale call-target.

### Issues Resolved
[MIGRATIONS-3030](https://opensearch.atlassian.net/browse/MIGRATIONS-3030)

### Testing

 `npm test` passes in `orchestrationSpecs/` — 50 schema tests + 13 transformer validation tests, all green.
- Snapshot tests (`argoSchemaSnapshot`, `unifiedSchema`) updated for the new defaults and the `template.pod.affinity` fixture shape.
- `migrationConfigTransformer.validation.test.ts` asserts the new merged config including the anti-affinity block.

### Check List


migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
```diff
@@ -210,8 +210,9 @@ def cleanup_cdc_resources(namespace: str):
     _kubectl_delete("deployment", f"capture-proxy-{TARGET_LABEL}-replayer", namespace)
     _kubectl_delete("service", "capture-proxy", namespace)
 
-    # 4. Kafka PVC
-    _kubectl_delete("pvc", "data-default-dual-role-0", namespace)
+    # 4. Kafka PVCs (one per broker replica)
+    for i in range(3):
```
gregschohn: are you setting this somewhere else?  How do you know it's 3?  Do you need to query this value from elsewhere?  It might be cleaner to use selector patterns...

BUT - we should be using reset for this to tear this down.  Why would reset not be the right tool here?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -237,22 +237,46 @@ const DEFAULT_AUTO_CREATE_KAFKA = {
...
         }
     },
     nodePoolSpecOverrides: {
-        replicas: 1,
+        replicas: 3,
         roles: ["controller", "broker"],
         storage: {
             type: "persistent-claim",
-            size: "1Gi",
+            size: "2Gi",
```
gregschohn, line 264: why did this need to go to 2GB?  Both seem way too small for most everybody anyway.  Was this a side effect to more replicas?

orchestrationSpecs/packages/config-processor/scripts/samples/proxyWithoutTls.wf.yaml
```diff
@@ -10,28 +10,7 @@ kafkaClusterConfiguration:
       clusterSpecOverrides:
         kafka:
           config:
-            auto.create.topics.enable: false
```
gregschohn, line 13: On this one - I don't see this PR flipping this to false.  Even if it was already false, should we explicitly put it into the default block?

orchestrationSpecs/packages/config-processor/scripts/samples/proxyWithoutTls.wf.yaml
```diff
@@ -10,28 +10,7 @@ kafkaClusterConfiguration:
...
-        storage:
-          type: persistent-claim
-          size: 1Gi
-          deleteClaim: true
-      topicSpecOverrides:
-        partitions: 1
-        replicas: 1
-        config:
-          retention.ms: 604800000
-          segment.bytes: 1073741824
```
gregschohn, line 34: These values are very useful for testing locally.  I use these files constantly.  We can put comments in like `# override with non-reliable settings for proof-of-concept` testing.  Notice that these files are NOT packaged to the migration console

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -237,22 +237,46 @@ const DEFAULT_AUTO_CREATE_KAFKA = {
...
-                "offsets.topic.replication.factor": 1,
-                "transaction.state.log.replication.factor": 1,
-                "transaction.state.log.min.isr": 1,
-                "default.replication.factor": 1,
-                "min.insync.replicas": 1,
+                "offsets.topic.replication.factor": 3,
+                "transaction.state.log.replication.factor": 3,
+                "transaction.state.log.min.isr": 2,
+                "default.replication.factor": 3,
+                "min.insync.replicas": 2,
```
gregschohn, line 251: can you add comments about what these are and why you chose them.  I always have to look up isr.  
I want to make sure that future devs understand the intent here (they might not look at git history/blame)

orchestrationSpecs/packages/schemas/tests/fixtures/strimzi/minimal-openapi.json
```diff
@@ -182,6 +182,22 @@
...
+                "properties": {
+                  "affinity": {
+                    "type": "object",
+                    "additionalProperties": true
+                  }
+                },
+                "additionalProperties": true
+              }
+            },
+            "additionalProperties": true
```
gregschohn, line 200: how did you have to add these lines?

orchestrationSpecs/KAFKA_IMPLEMENTATION_PLAN.md
gregschohn, line 1: please delete this whole file - plans are only relevant during build out - once they're done, they can be deleted and fold into the git commit log.

gregschohn (commented):
This looks mostly good.  I had some questions about 2GB storage and would rather not change the test sample files (the extra time & resources are a drag on development efficiency).

Have you done a test yet where EKS auto mode is scaling nodes up and moving pods to new instances?  I'd be interested to see a summary of pod::node movements over time.  I'd also hope that our test time goes down now that the system is much less sensitive to a single kafka node going down.  Those can be follow up tasks though

migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
jugal-chauhan: Yes you are right. Workflow reset in cleanup was added in #2842 . I'll remove the stale commands and constants from this file in the following commit.

orchestrationSpecs/packages/config-processor/scripts/samples/proxyWithoutTls.wf.yaml
jugal-chauhan: This PR does not change `auto.create.topics.enable` - It was, and remains, `false` by default. The explicit `false` lives in `DEFAULT_AUTO_CREATE_KAFKA` in [userSchemas.ts](https://github.com/opensearch-project/opensearch-migrations/pull/2782/changes#diff-45cf34dffcb6b9ea5add142eb7e2fc50431dfc17415bf7ad921d5db7b01a8a38R239)

orchestrationSpecs/KAFKA_IMPLEMENTATION_PLAN.md
jugal-chauhan: Removed in followup commits

orchestrationSpecs/packages/config-processor/scripts/samples/proxyWithoutTls.wf.yaml
jugal-chauhan: I put those back in [2d83cb415](https://github.com/opensearch-project/opensearch-migrations/pull/2782/changes/2d83cb4158b5697d7fadba8e22cafc2e5e221b68) and added a 2-line banner at the top of the autoCreate block in all three `proxyWithoutTls*.wf.yaml` files

orchestrationSpecs/packages/schemas/src/userSchemas.ts
jugal-chauhan: Done

orchestrationSpecs/packages/schemas/tests/fixtures/strimzi/minimal-openapi.json
jugal-chauhan: The fixture's `KafkaNodePoolSpec` has `additionalProperties: false`, so once `DEFAULT_AUTO_CREATE_KAFKA` gained a `spec.template.pod.affinity.podAntiAffinity` subtree, the snapshot test started diffing. The generated merged schema surfaced a property the fixture hadn't declared.

Without the fixture change, `argoSchemaSnapshot.test.ts` ("argo user matches expected") fails with a snapshot diff showing the template block the generator now emits but the old snapshot doesn't have. (I verified this by reverting the fixture to origin/main locally and re-running the test, only that one test fails, and with a snapshot diff, not an AJV validation error)

gregschohn (approved):

PR #2783, opened 2026-04-22
Remove insecure authHeaderValue replayer option and add auth conflict validation

jugal-chauhan:
### Problem

The replayer's `authHeaderValue` field in the Zod schema had `.default("")`, causing an empty string to be serialized into the replayer's INLINE-JSON config. When the target cluster used SigV4 auth (auto-derived at deploy time as `sigv4AuthHeaderServiceRegion`), the Java replayer received two auth options and rejected the pod at startup with:

> Cannot specify more than one auth option

This caused `CrashLoopBackOff` on all CDC integration tests

### Solution

Remove `authHeaderValue` entirely. The field was insecure (plaintext credentials in config rather than via a Kubernetes Secret) and predated the `--target-username`/`--target-password` CLI params which load credentials from the environment securely. Users needing custom request auth can use request transformations.

Additionally, add Zod cross-cutting validation to catch `removeAuthHeader: true` conflicting with a target that has SigV4 or basic auth configured — preventing the same class of dual-auth startup crash at config time rather than at pod startup.

### Changes

**Schema & Validation** (`orchestrationSpecs/`)
- Removed `authHeaderValue` from `USER_REPLAYER_PROCESS_OPTIONS`
- Added `OVERALL_MIGRATION_CONFIG` validation: `removeAuthHeader` + sigv4/basic target → config error
- Updated schema snapshots
- Added invalid fixture: `replayer-remove-auth-conflicts-with-target-sigv4.json`

**CRD & Admission Policies** (`deployment/k8s/`)
- Removed `authHeaderValue` from TrafficReplay CRD spec
- Removed all `authHeaderValue` references from ValidatingAdmissionPolicy gated-field rules

**Java CLI** (`TrafficCapture/trafficReplayer/`)
- Removed `--auth-header-value` / `--authHeaderValue` CLI parameter and `AUTH_HEADER_VALUE_ARG` constant
- Simplified `buildAuthTransformerFactory()` — basic auth now goes directly through `targetUsername`/`targetPassword`
- `StaticAuthTransformerFactory` retained (still used by the basic auth path)

**Docker & Docs**
- `docker-compose.yml`: Replaced `--auth-header-value Basic\\...` with `--target-username admin --target-password 'myStrongPassword123!'`
- `trafficReplayer/README.md`: Updated auth documentation to reference current options
- `docs/reconfiguringWorkflows.md`: Removed `authHeaderValue` row from change-restriction table

### Issues Resolved
N/A

### Testing
via Jenkins CDC tests

### Check List

### DIFF
--- a/orchestrationSpecs/packages/schemas/src/userSchemas.ts
+++ b/orchestrationSpecs/packages/schemas/src/userSchemas.ts
@@ -432,9 +432,6 @@ export const USER_REPLAYER_WORKFLOW_OPTIONS = z.object({
 }).describe("Kubernetes deployment-level options for the traffic replayer.");
 
 export const USER_REPLAYER_PROCESS_OPTIONS = z.object({
-    authHeaderValue: z.string().default("").optional()
-        .describe("Static value to use for the \"authorization\" header of each request (cannot be used with other auth arguments)")
-        .changeRestriction('gated'),
     kafkaTrafficEnableMSKAuth: z.boolean().default(false).optional()
         .describe("Enable SASL/IAM authentication for the replayer's Kafka consumer when connecting to Amazon MSK. Uses the pod's IAM role via EKS Pod Identity.")
         .changeRestriction('impossible'),
@@ -1120,6 +1117,23 @@ export const OVERALL_MIGRATION_CONFIG = //validateOptionalDefaultConsistency
                             code: z.ZodIssueCode.custom,
                             message: `Replayer '${replayerName}' dependsOnSnapshotMigrations[${j}] references unknown source '${dep.source}'. Available: ${Object.keys(data.sourceClusters).join(', ')}`,
                             path: ['traffic', 'replayers', replayerName, 'dependsOnSnapshotMigrations', j, 'source']
+                        });
+                    }
+                }
+
+                // When the target has sigv4 or basic auth, the workflow auto-derives the replayer's auth
+                // at deploy time. Setting removeAuthHeader alongside that causes a dual-auth startup crash.
+                const targetCluster = data.targetClusters[rc.toTarget];
+                if (targetCluster && rc.replayerConfig && rc.replayerConfig.removeAuthHeader === true) {
+                    const targetHasSigv4 = targetCluster.authConfig && HTTP_AUTH_SIGV4.safeParse(targetCluster.authConfig).success;
+                    const targetHasBasicAuth = targetCluster.authConfig && HTTP_AUTH_BASIC.safeParse(targetCluster.authConfig).success;
+                    if (targetHasSigv4 || targetHasBasicAuth) {
+                        const targetAuthType = targetHasSigv4 ? 'sigv4' : 'basic';
+                        ctx.addIssue({
+                            code: z.ZodIssueCode.custom,
+                            message: `Replayer '${replayerName}' sets 'removeAuthHeader' but target '${rc.toTarget}' uses ${targetAuthType} auth. ` +
+                                `The replayer cannot use both — the target's auth is applied automatically. Remove 'removeAuthHeader' from replayerConfig.`,
+                            path: ['traffic', 'replayers', replayerName, 'replayerConfig', 'removeAuthHeader']
                         });
                     }
                 }
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ orchestrationSpecs/dist/
 orchestrationSpecs/node_modules/
 orchestrationSpecs/packages/**/dist/
 orchestrationSpecs/packages/**/node_modules/
+orchestrationSpecs/packages/**/coverage/
 orchestrationSpecs/packages/migration-workflow-templates/k8sResources
 orchestrationSpecs/k8s-schemas
 orchestrationSpecs/packages/argo-workflow-builders/src/generated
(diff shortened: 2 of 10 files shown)


orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -432,8 +432,8 @@ export const USER_REPLAYER_WORKFLOW_OPTIONS = z.object({
 }).describe("Kubernetes deployment-level options for the traffic replayer.");
 
 export const USER_REPLAYER_PROCESS_OPTIONS = z.object({
-    authHeaderValue: z.string().default("").optional()
-        .describe("Static value to use for the \"authorization\" header of each request (cannot be used with other auth arguments)")
+    authHeaderValue: z.string().optional()
```
AndreKurait: Ideally i'd like an automatic validation to make sure this isn't specified at the same time the other ones are which would have caught this issue during our sample schema tests

jugal-chauhan: Added in the latest commit

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -511,6 +511,20 @@ export const USER_REPLAYER_OPTIONS = z.object({
             path: ['lookaheadTimeSeconds']
         });
     }
+    // Replayer auth options are mutually exclusive: at most one may be set.
+    // This mirrors the Java-side validation in TrafficReplayer.buildAuthTransformerFactory().
+    // Note: sigv4 auth is auto-derived from the target's authConfig at deploy time, not a user-facing field here.
+    const authOptionsSet = [
+        data.authHeaderValue !== undefined && data.authHeaderValue !== '' ? 'authHeaderValue' : null,
+        data.removeAuthHeader === true ? 'removeAuthHeader' : null,
```
AndreKurait: we should see sigv4 target username here etc. right?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -511,6 +511,20 @@ export const USER_REPLAYER_OPTIONS = z.object({
             path: ['lookaheadTimeSeconds']
         });
     }
+    // Replayer auth options are mutually exclusive: at most one may be set.
+    // This mirrors the Java-side validation in TrafficReplayer.buildAuthTransformerFactory().
+    // Note: sigv4 auth is auto-derived from the target's authConfig at deploy time, not a user-facing field here.
+    const authOptionsSet = [
+        data.authHeaderValue !== undefined && data.authHeaderValue !== '' ? 'authHeaderValue' : null,
```
AndreKurait: isn't authHeaderValue = "" what caused the issue? That means this check needs to differentiate between undefined and empty string

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: This should also be an expert setting since we would normally get auth from the target info (including basic auth? Please confirm)

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: By default, we should be getting the auth from the target config, which should set up the arguments for the replayer appropriately.  These two fields would only be for overriding those target defaults, right?  So if that's the case, I think that we don't include those in the validation.  We may want to show a warning/message to the user that we're overriding the default auth for this replayer run.

gregschohn (changes requested):
What behaviors do we actually want a user to be able to do?  But first...

1) If we assume that the target has at least some default way to make successful requests, is there value in having a replayer use an alternate?  
2) In this case, the value is insecure - it isn’t passed in via a secret, so the credentials that are set could be copied and pasted elsewhere and work just the same.

What do we want to allow: 
For right now, we should just strike the value altogether from the CRD and the user schemas.  Someday we could have the ability to use different identities to the target for different requests.  However, I don’t know if we’ll ever want to say this whole block of replayer requests should go through with a singular, and different identity than the RFS requests that we previously used.

jugal-chauhan:
> What behaviors do we actually want a user to be able to do? But first...
> 
> 1. If we assume that the target has at least some default way to make successful requests, is there value in having a replayer use an alternate?
> 2. In this case, the value is insecure - it isn’t passed in via a secret, so the credentials that are set could be copied and pasted elsewhere and work just the same.
> 
> What do we want to allow: For right now, we should just strike the value altogether from the CRD and the user schemas.  Someday we could have the ability to use different identities to the target for different requests. However, I don’t know if we’ll ever want to say this whole block of replayer requests should go through with a singular, and different identity than the RFS requests that we previously used.

Agreed. I went further and removed `authHeaderValue` entirely in the latest commits. Target auth (sigv4 via `authConfig.sigv4`, basic via `targetUsername/targetPassword` from K8s Secret) is auto-derived at deploy time. Added cross-cutting Zod validation for `removeAuthHeader` + sigv4/basic target conflicts too.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
jugal-chauhan: Addressed in latest commit

orchestrationSpecs/packages/schemas/src/userSchemas.ts
jugal-chauhan: Addressed in latest commit

AndreKurait (approved):
LGTM, thanks for this cleanup and validation additions

gregschohn (approved):
Thanks - sorry for having a brain freeze and bringing that in!

PR #2787, opened 2026-04-23
Remove OpenSearch dependency from Capture Proxy

gregschohn:
# Description

## Summary

Remove the OpenSearch libraries (`opensearch`, `opensearch-core`, `opensearch-common`,
`opensearch-security` plugin) as dependencies of the Capture Proxy. The proxy now uses
PEM-based TLS configuration exclusively via Netty's `SslContextBuilder`, eliminating the
need for the OpenSearch security plugin's `DefaultSecurityKeyStore`.  This results in the 
compressed size of the image going from 277MB to 171MB for arm64.

This is a breaking change: the `--sslConfigFile` CLI flag is removed. All deployment
paths (CDK, K8s, docker-compose) are updated to use `--sslCertChainFile` / `--sslKeyFile`.

## Motivation

The OpenSearch dependency existed solely to support the `--sslConfigFile` code path, which
parsed a YAML file with `plugins.security.ssl.http.*` keys and loaded TLS via
`DefaultSecurityKeyStore`. This pulled in the entire OpenSearch security plugin and its
transitive dependency tree — a significant amount of weight for what amounts to reading
PEM files.

Commit `b5cd38992` ("Add PEM-based TLS support to CaptureProxy") already added a clean
alternative that loads PEM certs directly through Netty. The K8s orchestration was already
using this path. This change completes the migration by removing the legacy path entirely.

### Java

- **`CaptureProxy.java`**: Removed `--sslConfigFile` parameter, `getSettings()` method,
  `HTTPS_CONFIG_PREFIX` / `SUPPORTED_TLS_PROTOCOLS_LIST_KEY` constants, the
  `DefaultSecurityKeyStore` branch in `buildSslEngineSupplier()`, and all imports that
  were only used by the legacy path (`Settings`, `DefaultSecurityKeyStore`,
  `SSLConfigConstants`, `TypeReference`, `YAMLFactory`, `ObjectMapper`, `Lombok`,
  `SneakyThrows`, `Paths`, `Optional`, `Collectors`).

- **`NettyScanningHttpProxyTest.java`**: Replaced `org.opensearch.common.collect.Tuple`
  (provided transitively by OpenSearch) with `java.util.Map.Entry`.

### Gradle

- **`trafficCaptureProxyServer/build.gradle`**: Removed the `opensearchSecurityPlugin`
  configuration, all four OpenSearch dependencies, the `zipTree` unpacking of the security
  plugin JARs, and the `jackson-core` / `jackson-databind` dependencies (no longer used
  anywhere in the proxy). Added explicit `kafka-clients` dependency which was previously
  resolved transitively through OpenSearch.

### Docker

- **`captureProxyBase/Dockerfile`**: Removed `PROXY_TLS_CONFIG_FILE` env var and the
  `proxy_tls.yml` generation (`plugins.security.ssl.http.*` YAML). The PEM certificate
  generation, copy, and Java truststore import are unchanged — the base image remains
  Corretto 11.

### Deployment configurations

- **CDK (`capture-proxy-stack.ts`)**: Changed from
  `--sslConfigFile /usr/share/captureProxy/config/proxy_tls.yml` to
  `--sslCertChainFile /usr/share/captureProxy/config/pub.pem --sslKeyFile /usr/share/captureProxy/config/key.pem`.
  The PEM files are already baked into the `capture_proxy_base` image.

- **Docker Compose (`proxy-single.yml`, `proxy-multi.yml`)**: Same flag swap as CDK.

- **K8s CRD (`migrationCrds.yaml`)**: Removed the `sslConfigFile` field from the
  CaptureProxy spec.

### Orchestration specs

- **`userSchemas.ts`**: Removed `sslConfigFile` from `USER_PROXY_PROCESS_OPTIONS` and
  the `superRefine` mutual-exclusion validation. Updated `tls` field description.

- **`migrationConfigTransformer.ts`**: Removed the `!normalized.proxyConfig?.sslConfigFile`
  guard from the secure-by-default TLS injection logic.

- **Snapshots**: Regenerated `argoSchemaSnapshot.test.ts.snap` (2 snapshots) and
  `setupCapture.snap.json` / `resourceManagement.snap.json` (13 snapshots).

### Tests

- **`CaptureProxySetupTest`**: Removed `testTlsParametersAreProperlyRead`,
  `testNoProtocolConfigDefaultsToSecureOnesOnly`, the `TLS_PROTOCOLS_KEY` constant,
  and the `testTlsParametersAreProperlyRead` helper method (all tested `getSettings()`).

- **`PemSslConfigurationTest`**: Removed `testBuildSslEngineSupplierRejectsBothPemAndLegacy`
  and the now-unused `@TempDir` / `Files` / `Path` imports.

- **Deleted** `orchestrationSpecs/packages/schemas/tests/fixtures/invalid/proxy-ssl-and-tls-both-set.json`
  (tested the removed mutual-exclusion validation).

### Who is affected

Anyone using `--sslConfigFile` directly:
- CDK deployments with custom `captureProxyExtraArgs` overriding `--sslConfigFile`
- Manual docker-compose invocations referencing `--sslConfigFile`
- Bare-metal deployments using the OpenSearch-style YAML TLS config

### Who is NOT affected

- K8s deployments using the `tls` config option (already on the PEM path)
- CDK deployments without custom `extraArgs` overriding `--sslConfigFile` (the stack
  now automatically uses the PEM flags)

### How to migrate

Replace:
```
--sslConfigFile <path-to-yaml>
```

With:
```
--sslCertChainFile <path-to-cert.pem> --sslKeyFile <path-to-key.pem>
```

Optionally add `--sslTrustCertFile <path-to-ca.pem>` for mutual TLS.

The PEM file paths correspond to the values that were previously in the YAML:

| YAML key | New CLI flag |
|---|---|
| `plugins.security.ssl.http.pemcert_filepath` | `--sslCertChainFile` |
| `plugins.security.ssl.http.pemkey_filepath` | `--sslKeyFile` |
| `plugins.security.ssl.http.pemtrustedcas_filepath` | `--sslTrustCertFile` |

### Issues Resolved
No Jira

### Testing
Automated tests

### Check List


buildImages/build.gradle
```diff
@@ -50,19 +50,14 @@ def jibProjects = [
                     ":TrafficCapture:transformationShim": ["syncVersionFile_transformation_shim"]
                 ]
         ],
-        "TrafficCapture:trafficCaptureProxyServer": [
+        "TrafficCapture:trafficCaptureProxyServer": correttoBase + [
                 // purposefully omit '-Dio.netty.handler.codec.http.defaultStrictLineParsing=false'
-                baseImageRegistryEndpoint: intermediateRegistry.hostUrl,
```
AndreKurait, line 55: Nice!

TrafficCapture/dockerSolution/build.gradle
```diff
@@ -191,6 +185,18 @@ javaContainerServices.each { dockerImageName, projectName ->
     CommonUtils.createDockerfile(project, artifactProject, baseImageOverrides[dockerImageName], dockerFilesForExternalServices, dockerImageName)
 }
 
+// Stage self-signed certs into the capture_proxy Docker build context
+tasks.named("copyArtifact_capture_proxy").configure {
```
AndreKurait, line 189: Where is this task getting called?

AndreKurait (approved):

TrafficCapture/dockerSolution/build.gradle
gregschohn: Don't we still do builds for some tests using the images builds in dockerSolution (like the docker compose tests)?  If not, we should remove those targets.  If we do, that's obviously something that we should clean up - but I don't think that we have done that yet.

PR #2791, opened 2026-04-23
ES 6.8 E2E: preDeployStep hook + explicit migration-then-source teardown

AndreKurait:
## Summary

Makes the ES 6.8 full-E2E Jenkins pipeline safe to re-run against a stage that
already has partial stacks deployed, and makes the teardown order explicit and
observable. Introduces a `preDeployStep` hook on `defaultIntegPipeline`, and in
`fullES68SourceE2ETest` uses it to destroy the migration CDK app FIRST, block
until every `OSMigrations-<stage>` CloudFormation stack reaches
`DELETE_COMPLETE`, then destroy the source (E2E solution) CDK app.

## Problems Solved

1. **Re-runs against a dirty stage were racy** — if a prior pipeline run left
   the migration CDK app and/or the source CDK app partially deployed, the next
   deploy would try to stand them up on top of the leftovers instead of
   cleaning first. There was no pipeline-level hook to run pre-deploy teardown.
2. **Teardown order was implicit inside one shell call** — the existing
   `--clean-up-all` path destroyed migration then source in a single bash
   function. The "migration finishes before source starts" guarantee was
   correct but invisible in the Jenkins log and hard to reason about if
   something hung.
3. **No explicit CFN wait on leftover migration stacks** — `npx cdk destroy` is
   synchronous, but if a stack had been orphaned outside the CDK app's tracking
   there was nothing guaranteeing it was fully gone before the source teardown
   began.

### Pipeline hook (`vars/defaultIntegPipeline.groovy`)

Adds an optional `preDeployStep` closure to the pipeline config. When
provided, it runs inside the deploy workspace with the standard
`sourceContextFileName` / `migrationContextFileName` / `sourceContextId` /
`migrationContextId` / `stage` args injected, so any test cover can drop in
pre-deploy teardown or preparation without forking the whole pipeline.

### ES 6.8 E2E wires the hook (`vars/fullES68SourceE2ETest.groovy`)

Invokes `awsE2ESolutionSetup.sh` twice in strict order:

```
./awsE2ESolutionSetup.sh ... --clean-up-migration-only
./awsE2ESolutionSetup.sh ... --clean-up-source-only
```

Two separate `sh` calls guarantee the second cannot start until the first has
fully returned, and each appears as its own line in the Jenkins log so a stuck
teardown is obvious.

### Script refactor (`test/awsE2ESolutionSetup.sh`)

- Splits the existing `clean_up_all` function into `clean_up_migration` and
  `clean_up_source`. `clean_up_all` is kept as a thin wrapper that calls them
  in order, so any existing caller that still passes `--clean-up-all` keeps
  working unchanged.
- `clean_up_migration` now additionally polls with
  `aws cloudformation wait stack-delete-complete` over any stack whose name
  starts with `OSMigrations-<stage>`, catching stacks that `cdk destroy` may
  have skipped (e.g. orphaned ones no longer tracked in the local app).
- Adds `--clean-up-migration-only` and `--clean-up-source-only` flags.
  `--clean-up-migration-only` short-circuits before the script touches the
  source CDK tree at all, so it cannot block on a missing source repo /
  missing source stack describe.

## Testing

- `bash -n test/awsE2ESolutionSetup.sh` passes.
- Manual review: verified the new flags route through the correct bash
  functions and that `--clean-up-all` still calls both phases in the same
  order as before.
- Rebased onto latest `upstream/main` with no conflicts.
- Runtime behavior will be exercised by the next scheduled run of the ES 6.8
  E2E pipeline, which exists specifically to catch regressions in this teardown
  path.

## Migration notes

Backwards compatible. Callers that don't set `preDeployStep` get the exact
prior behavior. Callers that still pass `--clean-up-all` to
`awsE2ESolutionSetup.sh` get the exact prior ordering.


test/awsE2ESolutionSetup.sh
```diff
@@ -79,21 +79,77 @@ create_service_linked_roles () {
...
+  # `source-map-support/register` side-effect import (and its type decls)
+  # resolves. In the full deploy flow `npm ci` runs before `cdk deploy`, but
+  # the --clean-up-migration-only short-circuit skips that path, so install
+  # deps here unconditionally. `npm ci` is cheap relative to the destroy.
+  npm ci
+  echo "Destroying migration CDK app stacks..."
+  CDK_SKIP_LOCAL_IMAGE_HASH=true npx cdk destroy "*" --force --c contextFile="$MIGRATION_GEN_CONTEXT_FILE" --c contextId="$MIGRATION_CONTEXT_ID"
+  local cdk_rc=$?
+  if [ $cdk_rc -ne 0 ]; then
+    echo "Error: cdk destroy for migration stacks exited with code $cdk_rc, aborting cleanup before source destroy."
```
gregschohn: what is source destroy?

test/awsE2ESolutionSetup.sh
```diff
@@ -79,21 +79,77 @@ create_service_linked_roles () {
...
+  local migration_stack_prefix="OSMigrations-${STAGE}"
+  local leftover
+  leftover=$(aws cloudformation list-stacks \
+      --stack-status-filter CREATE_COMPLETE UPDATE_COMPLETE UPDATE_ROLLBACK_COMPLETE ROLLBACK_COMPLETE DELETE_FAILED DELETE_IN_PROGRESS CREATE_IN_PROGRESS UPDATE_IN_PROGRESS \
+      --query "StackSummaries[?starts_with(StackName, \`${migration_stack_prefix}\`)].StackName" \
+      --output text)
+  if [ -n "$leftover" ]; then
+    echo "Waiting for lingering migration stacks to reach DELETE_COMPLETE: $leftover"
+    for s in $leftover; do
+      aws cloudformation wait stack-delete-complete --stack-name "$s" || {
```
gregschohn, line 121: are you waiting indefinitely here?

test/awsE2ESolutionSetup.sh
```diff
@@ -239,6 +307,31 @@ validate_required_options
 sed -i -e "s/<STAGE>/$STAGE/g" "$SOURCE_GEN_CONTEXT_FILE"
 sed -i -e "s/<STAGE>/$STAGE/g" "$MIGRATION_GEN_CONTEXT_FILE"
 
+# Short-circuit for --clean-up-migration-only: don't touch the source CDK tree at all,
+# just destroy the migration app stacks and wait for their CFN delete to complete.
+# The migration CDK app still synths on destroy, so resolve <VPC_ID> and
+# <SOURCE_CLUSTER_ENDPOINT> the same way the full path does (via CFN outputs of
+# the source stacks if they still exist), falling back to dummy values when the
+# source side has already been destroyed -- cdk destroy doesn't actually hit the
+# endpoint/vpc, it only needs synth to succeed so it can match CFN stack names.
```
jugal-chauhan, line 333: Similar to `--build` and `--version`, can we add a mutually exclusive logic between these 3 cleanup options ? 

```
if [ "$CLEAN_UP_ALL" = true ] && { [ "$CLEAN_UP_MIGRATION_ONLY" = true ] || [ "$CLEAN_UP_SOURCE_ONLY" = true ]; }; then
  echo "Error: --clean-up-all cannot be combined with --clean-up-migration-only or --clean-up-source-only"
  exit 1
fi
```

test/awsE2ESolutionSetup.sh
```diff
@@ -250,6 +343,15 @@ if [ "$BOOTSTRAP_REGION" = true ] ; then
...
 fi
 
+# Short-circuit for --clean-up-source-only: destroy the source CDK app stacks and exit.
+# vpc_id is best-effort here; if the source network stack is already gone we pass
+# an empty value and clean_up_source skips the SG-revert loop.
+if [ "$CLEAN_UP_SOURCE_ONLY" = true ] ; then
+  vpc_id=$(aws cloudformation describe-stacks --stack-name "$SOURCE_NETWORK_STACK_NAME" --query "Stacks[0].Outputs[?contains(OutputValue, 'vpc')].OutputValue | [0]" --output text 2>/dev/null || echo "")
+  clean_up_source "$vpc_id"
+  exit 0
+fi
```
jugal-chauhan, line 370: Why is this placed after we checkout `opensearch-cluster-cdk` ?

vars/defaultIntegPipeline.groovy
```diff
@@ -128,6 +128,31 @@ def call(Map config = [:]) {
...
+                        dir('test') {
+                            script {
+                                withCredentials([string(credentialsId: 'migrations-test-account-id', variable: 'MIGRATIONS_TEST_ACCOUNT_ID')]) {
+                                    withAWS(role: 'JenkinsDeploymentRole', roleAccount: "${MIGRATIONS_TEST_ACCOUNT_ID}", duration: 5400, roleSessionName: 'jenkins-session') {
+                                        config.preDeployStep(
+                                            stage: stage,
+                                            sourceContextFileName: source_context_file_name,
+                                            migrationContextFileName: migration_context_file_name,
+                                            sourceContextId: source_context_id,
+                                            migrationContextId: migration_context_id
```
jugal-chauhan, line 146: This is a great addition!

test/awsE2ESolutionSetup.sh
jugal-chauhan: I get it now, ignore this comment

jugal-chauhan (approved):
LGTM! The only change I'd like to see before we merge this is to ensure `--clean-up-all` cannot be combined with `--clean-up-migration-only` or `--clean-up-source-only`

vars/fullES68SourceE2ETest.groovy
```diff
@@ -102,6 +102,23 @@ def call(Map config = [:]) {
...
+            preDeployStep: { Map args ->
+                // Destroy any prior stacks in strict order before redeploying: migration CDK
+                // app first, BLOCK until all of its CloudFormation stacks reach DELETE_COMPLETE,
+                // then destroy the source (E2E solution) CDK app. Running these as two
+                // separate shell invocations makes the ordering explicit in the Jenkins log
+                // and guarantees the second call cannot start until the first has fully
+                // returned. `npx cdk destroy` already polls CFN synchronously, and
+                // --clean-up-migration-only additionally waits on `aws cloudformation wait
+                // stack-delete-complete` for any leftover "OSMigrations-<stage>" stacks.
+                def commonArgs = "--source-context-file './${args.sourceContextFileName}' " +
```
gregschohn, line 124: why do you need a context file?  What if it was different from what you're cleaning up from long ago?

AndreKurait:
Pushed f35b70c33 addressing the outstanding review feedback:

**@gregschohn**
- awsE2ESolutionSetup.sh:101 — expanded the error message to spell out that "source destroy" is the EC2 source-cluster CDK teardown that would otherwise run after the migration-stack destroy.
- awsE2ESolutionSetup.sh:116 — added an inline comment documenting that `aws cloudformation wait stack-delete-complete` is bounded (30s poll × 120 attempts ≈ 1h at AWS CLI defaults) and is additionally wrapped by the 1h Jenkins Pre-Deploy Cleanup stage timeout, so the pipeline cannot hang indefinitely.
- vars/fullES68SourceE2ETest.groovy:114 — added an in-situ note explaining why passing a context file is still safe even when cleaning up an older deploy: CloudFormation stacks are named purely by `--stage` (e.g. `OSMigrations-<stage>`, `opensearch-infra-stack-ec2-source-<stage>`), so cleanup is stage-scoped regardless of context content. The context file only needs to be valid enough for `cdk destroy` to synth and enumerate its own stack list; placeholder `<VPC_ID>` / `<SOURCE_CLUSTER_ENDPOINT>` substitutions in awsE2ESolutionSetup.sh keep that safe when the source side is already gone.

**@jugal-chauhan**
- awsE2ESolutionSetup.sh:316 — added up-front validation that rejects combining `--clean-up-all` with either of `--clean-up-migration-only`/`--clean-up-source-only`, and rejects combining the latter two with each other, instead of letting one silently win.

**github-advanced-security (CodeQL)**
- common-utilities.ts:563/:566 — replaced `execSync(\`docker … ${imageName}\`)` with `execFileSync('docker', [args…])` so no shell is spawned and `imageName` can no longer be interpolated into a command line. Added a restrictive `^[a-zA-Z0-9._/:@-]+$` allowlist as defense-in-depth before the exec. The 35 existing common-utilities.test.ts cases still pass locally.

test/awsE2ESolutionSetup.sh
AndreKurait: This is not customer facing so i'd like to leave this simpler as is

test/awsE2ESolutionSetup.sh
AndreKurait: source cluster destroy

vars/fullES68SourceE2ETest.groovy
AndreKurait: This could get completely broken if its different from what we were cleaning up (or even if the cdk was changed). Since this is on the path to deprecation i'm not too concerned here

AndreKurait: This is kinda a limitation in CDK, nested stacks could resolve this but i'm not interested in making that change now

PR #2780, opened 2026-04-22
Added Function Syntax Parser

akshay2000:
### Description
Added syntax parser for function syntax. A function may occur in a lot of contexts. This PR helps with creating a function AST.

### Testing
- Unit tests

### Check List
- [x] New functionality includes testing

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2792, opened 2026-04-23
feat(solr-transforms): Add bq boost query param (bq → bool.should)

niravpi:
## Description

   Add support for the Solr eDisMax `bq` (boost query) parameter. The `bq` parameter adds optional query clauses that additively influence scoring without affecting which documents match.

   ### Mapping

   Solr:  q=cheese&defType=edismax&bq=category:food^10&bq=category:deli^5

```
   OpenSearch:
   {
     "query": {
       "bool": {
         "must": [<translated q>],
         "should": [<translated bq1>, <translated bq2>]
       }
     }
   }
```

   Each `bq` value is parsed through `translateQ()` — the same query engine used for `q` and `fq`. Local params syntax (`{!...}`) in bq values is rejected at validation time to prevent silent lossy translation.

   ### Behavior

   - Multiple `bq` params are each added as separate `bool.should` clauses
   - Duplicate `bq` params are preserved (not deduplicated) — matches Solr's additive scoring behavior
   - Flattens into existing bool query when present (works alongside `fq` filters)
   - Always ensures a `must` clause exists (`match_all` when no base query) so `should` is purely scoring-only
   - Fails fast when `bq` is used without `defType=edismax`
   - Fails fast on negative boosts (`^-N`) and local params (`{!...}`)
   - `defType=dismax` is not supported for `bq` — use `defType=edismax` instead

   ### Files Changed

   | File | Description |
   |---|---|
   | `features/boost-query-bq.ts` | New transform — parse bq via translateQ, add to bool.should (edismax only) |
   | `features/boost-query-bq.test.ts` | 32 unit tests (including boost preservation, negative boost rejection, defType validation) |
   | `features/validation.test.ts` | 3 new tests for bq local params rejection |
   | `integration-tests/boost-query-bq.testcase.ts` | 11 e2e test cases (10 edismax success + 1 dismax rejection, Solr 8 + 9) |
   | `registry.ts` | Register in FEATURE_MODULES and select pipeline after fq |
   | `docs/LIMITATIONS.md` | 4 new entries (DISMAX, BQ-NEG-BOOST, BF, BOOST-MULTIPLICATIVE) |

   ### Testing

   - **Unit tests**: 855/855 passed (32 new bq tests + 3 validation tests)
   - **E2E integration**: 11 bq tests passed on both Solr 8 and Solr 9 (including dismax rejection)

   ### Known Limitations

   | Shortcode | Summary |
   |---|---|
   | DISMAX | `defType=dismax` not supported — use `defType=edismax` instead |
   | BQ-NEG-BOOST | Negative boost in `bq` (e.g., `^-10`) not supported |
   | BF | `bf` (boost functions) parameter — not supported, requires function query support |
   | BOOST-MULTIPLICATIVE | eDisMax multiplicative `boost` parameter — not supported, requires function_score |

   ### References

   - [Solr DisMax bq parameter](https://solr.apache.org/guide/solr/latest/query-guide/dismax-query-parser.html#bq-boost-query-parameter)
   - [Solr eDisMax query parser](https://solr.apache.org/guide/solr/latest/query-guide/edismax-query-parser.html)

### DIFF
--- a/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
+++ b/TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
@@ -26,6 +26,10 @@ the root cause, and provides a workaround where one exists.
 | [MM-EDISMAX-OPERATOR-DEFAULT](#mm-edismax-operator-default) | eDisMax `mm` default with explicit operators not fully replicated |
 | [MM-EDISMAX-MIXED-OPERATORS](#mm-edismax-mixed-operators) | eDisMax `mm` with mixed explicit and implicit operators applies to wrong scope |
 | [MM-STOPWORD-MULTIFIELD](#mm-stopword-multifield) | `mm` with per-field stopword removal differs from Solr |
+| [DISMAX](#dismax)                                | `defType=dismax` not supported — use `defType=edismax` instead |
+| [BQ-NEG-BOOST](#bq-neg-boost)                   | Negative boost in `bq` (e.g., `^-10`) not supported |
+| [BF](#bf)                                       | `bf` (boost functions) parameter not supported |
+| [BOOST-MULTIPLICATIVE](#boost-multiplicative)   | eDisMax multiplicative `boost` parameter not supported |
 
 ---
 
@@ -644,3 +648,94 @@ due to this difference should consider aligning stopword lists across fields
 or lowering the `mm` value.
 
 **Reference:** https://solr.apache.org/guide/solr/latest/query-guide/dismax-query-parser.html#mm-minimum-should-match-parameter
+
+---
+
+## DISMAX
+
+**Feature:** `defType=dismax` (DisMax query parser)
+
+**Solr behaviour:**
+The DisMax query parser provides a simplified query syntax with multi-field
+search via `qf`, boost queries (`bq`), boost functions (`bf`), and minimum
+match (`mm`). eDisMax (Extended DisMax) is a superset that adds support for
+field:value syntax, boolean operators, and additional parameters.
+
+**Current status:**
+Not supported. Requests with `defType=dismax` are rejected by validation
+with a clear error message. Use `defType=edismax` instead — eDisMax supports
+all DisMax features plus additional capabilities.
+
+**Workaround:**
+Replace `defType=dismax` with `defType=edismax` in all queries. eDisMax is
+backwards-compatible with DisMax for the supported parameter set (`q`, `qf`,
+`bq`, `mm`, `pf`, `ps`, `tie`). The only behavioral difference is that
+eDisMax supports explicit field:value syntax and boolean operators in `q`,
+which DisMax treats as literal text.
+
+**Reference:** https://solr.apache.org/guide/solr/latest/query-guide/dismax-query-parser.html
+
+---
+
+## BQ-NEG-BOOST
+
+**Feature:** Negative boost values in `bq` (e.g., `bq=category:spam^-10`)
+
+**Solr behaviour:**
+Solr accepts negative boost values via Java's `Float.parseFloat()`. A negative
+boost reduces the score contribution of matching documents additively — the
+document is still returned, just ranked lower.
+
+**Current status:**
+Not supported. The shim rejects `bq` values containing `^-` with a fail-fast
+error. OpenSearch does not support negative boost values — they produce
+undefined scoring behavior. Without fail-fast, the PEG grammar would
+reinterpret `category:spam^-10` as `-(category:spam^10)` (negation with
+positive boost), which would **exclude** matching documents instead of
+demoting them — a silent correctness bug.
+
+**Workaround:**
+Remove negative boosts from `bq` params. If score demotion is needed,
+consider restructuring the query to use positive boosts on the preferred
+documents instead.
+
+---
+
+## BF
+
+**Feature:** `bf` (boost functions) parameter
+
+**Solr behaviour:**
+The `bf` parameter in DisMax/eDisMax specifies function expressions with
+optional boosts that are added as scoring clauses. For example:
+`bf=div(1,sum(1,price))^1.5` is shorthand for
+`bq={!func}div(1,sum(1,price))^1.5`. Multiple `bf` values are supported.
+
+**Current status:**
+Not supported. The `bf` parameter is entirely function-based and requires
+a dedicated transform to parse the function shorthand syntax and delegate
+to the query engine. Sending `bf` will be rejected by validation as an
+unsupported parameter.
+
+**Reference:** https://solr.apache.org/guide/solr/latest/query-guide/dismax-query-parser.html#bf-boost-functions-parameter
+
+---
+
+## BOOST-MULTIPLICATIVE
+
+**Feature:** eDisMax multiplicative `boost` parameter
+
+**Solr behaviour:**
+The eDisMax `boost` parameter applies a multiplicative boost to the entire
+query score. Unlike `bq` (additive), `boost` acts as a scaling factor.
+For example: `boost=popularity` multiplies each document's score by its
+`popularity` field value. This is shorthand for wrapping the query in a
+`{!boost}` QParser.
+
+**Current status:**
+Not supported. Multiplicative boosting requires OpenSearch's `function_score`
+query with `script_score` or `field_value_factor`, which is architecturally
+different from additive `bq`/`bf`. Sending `boost` will be rejected by
+validation as an unsupported parameter.
+
+**Reference:** https://solr.apache.org/guide/solr/latest/query-guide/edismax-query-parser.html#extended-dismax-parameters
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
@@ -1,0 +1,455 @@
+import { describe, it, expect } from 'vitest';
+import { request } from './boost-query-bq';
+import type { RequestContext, JavaMap } from '../context';
+
+/** Create a mock RequestContext for testing. */
+function createCtx(
+  params: Record<string, string | string[]>,
+  existingQuery?: Map<string, any>,
+): RequestContext {
+  const body = new Map<string, any>();
+  if (existingQuery) body.set('query', existingQuery);
+
+  const searchParams = new URLSearchParams();
+  for (const [key, value] of Object.entries(params)) {
+    if (Array.isArray(value)) {
+      for (const v of value) searchParams.append(key, v);
+    } else {
+      searchParams.append(key, value);
+    }
+  }
+
+  return {
+    msg: new Map() as JavaMap,
+    endpoint: 'select',
+    collection: 'test',
+    params: searchParams,
+    body: body as JavaMap,
+  };
+}
+
+/** Convert nested Maps to plain objects for easier assertion. */
+function mapToObj(map: Map<string, any>): any {
+  const obj: any = {};
+  for (const [key, value] of map.entries()) {
+    if (value instanceof Map) {
+      obj[key] = mapToObj(value);
+    } else if (Array.isArray(value)) {
+      obj[key] = value.map((v) => (v instanceof Map ? mapToObj(v) : v));
+    } else {
+      obj[key] = value;
+    }
+  }
+  return obj;
+}
+
+describe('boost-query-bq request transform', () => {
+  it('has correct name', () => {
+    expect(request.name).toBe('boost-query-bq');
+  });
+
+  // --- match guard ---
+
+  describe('match', () => {
+    it('matches when bq is present', () => {
+      const ctx = createCtx({ bq: 'category:food', defType: 'edismax' });
+      expect(request.match!(ctx)).toBe(true);
+    });
+
+    it('does not match when bq is absent', () => {
+      const ctx = createCtx({ defType: 'edismax' });
+      expect(request.match!(ctx)).toBe(false);
+    });
+
+    it('matches when bq is empty string (apply handles the no-op)', () => {
+      const ctx = createCtx({ bq: '', defType: 'edismax' });
+      expect(request.match!(ctx)).toBe(true);
+    });
+  });
+
+  // --- fail-fast on invalid defType ---
+
+  describe('defType validation', () => {
+    it('throws when defType is absent', () => {
+      const ctx = createCtx({ bq: 'category:food' }, new Map([['match_all', new Map()]]));
+      expect(() => request.apply(ctx)).toThrow('[boost-query-bq] bq parameter requires defType=edismax');
+    });
+
+    it('throws when defType is lucene', () => {
+      const ctx = createCtx({ bq: 'category:food', defType: 'lucene' }, new Map([['match_all', new Map()]]));
+      expect(() => request.apply(ctx)).toThrow("got 'lucene'");
+    });
+
+    it('throws for defType=dismax', () => {
+      const ctx = createCtx({ bq: 'category:food', defType: 'dismax' }, new Map([['match_all', new Map()]]));
+      expect(() => request.apply(ctx)).toThrow("got 'dismax'");
+    });
+
+    it('does not throw for defType=edismax', () => {
+      const ctx = createCtx({ bq: 'category:food', defType: 'edismax' }, new Map([['match_all', new Map()]]));
+      expect(() => request.apply(ctx)).not.toThrow();
+    });
+  });
+
+  // --- single bq ---
+
+  describe('single bq', () => {
+    it('wraps existing query in bool.must and adds bq to bool.should', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'category:food', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.must).toHaveLength(1);
+      expect(result.bool.must[0].match_all).toBeDefined();
+      expect(result.bool.should).toHaveLength(1);
+      expect(result.bool.should[0].match.category.query).toBe('food');
+      expect(result.bool.minimum_should_match).toBeUndefined();
+    });
+
+    it('handles boosted bq value', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'category:food^10', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(1);
+      expect(result.bool.should[0].match.category.query).toBe('food');
+      expect(result.bool.should[0].match.category.boost).toBe(10);
+    });
+
+    it('handles phrase bq value', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'title:"aged cheese"', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should[0].match_phrase.title.query).toBe('aged cheese');
+    });
+
+    it('handles range bq value', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'price:[10 TO 100]', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should[0].range.price.gte).toBe('10');
+      expect(result.bool.should[0].range.price.lte).toBe('100');
+    });
+  });
+
+  // --- multiple bq ---
+
+  describe('multiple bq', () => {
+    it('adds all bq values as separate should clauses', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx(
+        { bq: ['category:food^10', 'category:deli^5'], defType: 'edismax' },
+        existingQuery,
+      );
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(2);
+      expect(result.bool.should[0].match.category.query).toBe('food');
+      expect(result.bool.should[0].match.category.boost).toBe(10);
+      expect(result.bool.should[1].match.category.query).toBe('deli');
+      expect(result.bool.should[1].match.category.boost).toBe(5);
+    });
+
+    it('filters out empty bq values', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx(
+        { bq: ['category:food', '', '   ', 'type:book'], defType: 'edismax' },
+        existingQuery,
+      );
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(2);
+    });
+
+    it('returns early when all bq values are empty', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: ['', '   '], defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      expect(ctx.body.get('query')).toBe(existingQuery);
+      expect(ctx.body.size).toBe(1);
+      expect(existingQuery.size).toBe(1);
+    });
+  });
+
+  // --- bool query flattening ---
+
+  describe('bool query flattening', () => {
+    it('flattens into existing bool query', () => {
+      const existingBool = new Map<string, any>([
+        ['must', [new Map([['match', new Map([['title', new Map([['query', 'java']])]])]])]],
+      ]);
+      const existingQuery = new Map([['bool', existingBool]]);
+      const ctx = createCtx({ bq: 'category:tech^5', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.must).toHaveLength(1);
+      expect(result.bool.must[0].match.title.query).toBe('java');
+      expect(result.bool.should).toHaveLength(1);
+      expect(result.bool.should[0].match.category.query).toBe('tech');
+      expect(result.bool.should[0].match.category.boost).toBe(5);
+    });
+
+    it('merges with existing should clauses', () => {
+      const existingBool = new Map<string, any>([
+        ['must', [new Map([['match_all', new Map()]])]],
+        ['should', [new Map([['match', new Map([['status', new Map([['query', 'featured']])]])]])]],
+      ]);
+      const existingQuery = new Map([['bool', existingBool]]);
+      const ctx = createCtx({ bq: 'category:food^10', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(2);
+      expect(result.bool.should[0].match.status.query).toBe('featured');
+      expect(result.bool.should[1].match.category.query).toBe('food');
+      expect(result.bool.should[1].match.category.boost).toBe(10);
+    });
+
+    it('works alongside fq (must + filter + should all present)', () => {
+      const existingBool = new Map<string, any>([
+        ['must', [new Map([['match', new Map([['title', new Map([['query', 'cheese']])]])]])]],
+        ['filter', [new Map([['match', new Map([['inStock', new Map([['query', 'true']])]])]])]],
+      ]);
+      const existingQuery = new Map([['bool', existingBool]]);
+      const ctx = createCtx({ bq: 'category:food^10', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.must).toHaveLength(1);
+      expect(result.bool.filter).toHaveLength(1);
+      expect(result.bool.should).toHaveLength(1);
+    });
+  });
+
+  // --- no existing query ---
+
+  describe('no existing query', () => {
+    it('creates bool with match_all must and should when no existing query', () => {
+      const ctx = createCtx({ bq: 'category:food', defType: 'edismax' });
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.must).toHaveLength(1);
+      expect(result.bool.must[0].match_all).toBeDefined();
+      expect(result.bool.should).toHaveLength(1);
+      expect(result.bool.minimum_should_match).toBeUndefined();
+    });
+  });
+
+  // --- complex bq expressions ---
+
+  describe('complex bq expressions', () => {
+    it('handles boolean expression in bq', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx(
+        { bq: 'category:food AND inStock:true', defType: 'edismax' },
+        existingQuery,
+      );
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(1);
+      expect(result.bool.should[0].bool.must).toHaveLength(2);
+    });
+
+    it('handles match_all bq (*:*)', () => {
+      const existingQuery = new Map([['match', new Map([['title', new Map([['query', 'java']])]])]]);
+      const ctx = createCtx({ bq: '*:*', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should[0].match_all).toBeDefined();
+    });
+
+    it('handles grouped bq expression', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx(
+        { bq: '(category:food OR category:drink)^5', defType: 'edismax' },
+        existingQuery,
+      );
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(1);
+    });
+
+    it('handles bare term bq', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'wireless^5', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(1);
+    });
+
+    it('handles negation bq', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: '-category:Books', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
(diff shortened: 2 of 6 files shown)


TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
```diff
@@ -494,3 +497,77 @@ Both JSON and XML content types are supported.
 Only `application/json` request bodies are supported. XML bodies
 (`text/xml`, `application/xml`) cannot be parsed by the shim and will
 fail with an error. Clients using XML format must switch to JSON.
+
+---
+
+## BQ-DISMAX-MM
```
k-rooot: As per my understanding, the dismax parser type is not supported for the MVP. Please double check and make the necessary changes if needed.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (bqValues.length === 0) return;
+
+    const shouldClauses = bqValues.map(parseBq);
+
+    const existingQuery = ctx.body.get('query');
+
+    if (existingQuery && isBoolQuery(existingQuery)) {
+      const existingBool = existingQuery.get('bool') as Map<string, any>;
+      const existingShould = existingBool.get('should') || [];
+      existingBool.set('should', [...existingShould, ...shouldClauses]);
```
k-rooot, line 86: In Solr, bq contributes additive scoring regardless of whether the clause matches, whereas in OpenSearch, bool.should behaves differently depending on context.
* If there is a must, should is optional (OK)
* But scoring is not strictly additive like Solr’s bq
* More importantly, minimum_should_match defaults can alter behaviour

Should we wrap each bq clause inside a function_score or explicitly use constant_score + boost, or enforce minimum_should_match: 0 to preserve optionality.

Solr: `bq=category:food^10` -> Adds +10 to score if matched.

OpenSearch: `{ "term": { "category": "food", "boost": 10 } }` -> is not always equivalent especially with coordination factors and query normalization.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
...
+/** Solr query params this feature handles. */
+export const params = ['bq'];
+
+const DISMAX_TYPES = new Set(['dismax', 'edismax']);
+
+/**
+ * Parse a single bq value using the same query engine as q and fq.
+ */
+function parseBq(bq: string): Map<string, any> {
+  return translateQ(new Map([['q', bq]])).dsl;
```
k-rooot, line 48: To future-proof the dependency, we could add unit tests to ensure boosts are always preserved correctly in the translation and that no normalization or rewriting strips them.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
...
+    const bqValues = ctx.params.getAll('bq').filter((v) => v.trim() !== '');
+    if (bqValues.length === 0) return;
+
+    const shouldClauses = bqValues.map(parseBq);
+
+    const existingQuery = ctx.body.get('query');
+
+    if (existingQuery && isBoolQuery(existingQuery)) {
+      const existingBool = existingQuery.get('bool') as Map<string, any>;
+      const existingShould = existingBool.get('should') || [];
```
k-rooot, line 85: Would `existingBool.get('should')` return a map if previously set?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
...
+    if (!DISMAX_TYPES.has(defType)) {
+      throw new Error(
+        `[boost-query-bq] bq parameter requires defType=dismax or defType=edismax, got '${defType || '(none)'}'`,
+      );
+    }
+
+    const bqValues = ctx.params.getAll('bq').filter((v) => v.trim() !== '');
+    if (bqValues.length === 0) return;
+
+    const shouldClauses = bqValues.map(parseBq);
```
k-rooot, line 79: Duplicate bq params can lead to inflated scoring and redundant clauses.

Solr allows duplicates but often they’re unintended. Should we optionally deduplicate (or at least document the behaviour)?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
...
+ *       }
+ *     }
+ *   }
+ *
+ * This transform runs AFTER filter-query-fq. If the existing query is already
+ * a bool query (from query-q or fq), we add should clauses directly to avoid
+ * unnecessary nesting.
+ *
+ * bq values are parsed through translateQ() — the same query engine used for
+ * q and fq. Any query syntax the engine supports works in bq automatically,
```
k-rooot: `Any query syntax the engine supports works in bq automatically`

This is optimistic. Some Solr constructs like local params, {!boost} may not map cleanly. We can instead fail fast when translation is lossy. Thoughts?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
k-rooot, line 86: If a bool query has only should clauses, OpenSearch defaults to `minimum_should_match = 1`. 

This changes semantics vs Solr. Should we instead set `existingBool.set('minimum_should_match', 0);`?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
+/**
+ * Boost query bq parameter — convert Solr bq params to OpenSearch bool.should.
```
k-rooot, line 2: Should we handle negative boosts? Solr allows `bq=category:spam^-10`. In OS, negative boosts behave unpredictably or are ignored. We should either explicitly guard or fail.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (existingQuery && isBoolQuery(existingQuery)) {
+      const existingBool = existingQuery.get('bool') as Map<string, any>;
+      const existingShould = existingBool.get('should') || [];
+      existingBool.set('should', [...existingShould, ...shouldClauses]);
+    } else {
+      const boolQuery = new Map<string, any>();
+      if (existingQuery) {
+        boolQuery.set('must', [existingQuery]);
+      }
```
k-rooot: If only bq is present (no q), Solr still returns documents (match all + boosts). 

Current o/p
```
{
  "bool": {
    "should": [...]
  }
}
```

OpenSearch interprets this as match documents that satisfy any should clause (NOT match-all)

Should we add a match_all when no base query exists.
```
if (!existingQuery) {
  boolQuery.set('must', [new Map([['match_all', new Map()]])]);
}
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,318 @@
+import { describe, it, expect } from 'vitest';
```
k-rooot, line 1: I'll review the unit tests in the next revision.

k-rooot (changes requested):

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
niravpi: `defType=dismax` is supported for the core query path, the parser (`parser.ts`) applies `qf` field distribution for both `dismax` and `edismax`, and `query-q.ts` declares `defType` and `qf` as supported params. The `bq` parameter works identically for both parser types.

That said, several DisMax-specific params are not yet implemented and will be rejected by validation as unsupported: `mm`, `pf`, `ps`, `qs`, `tie`, `q.alt`, `bf`. These are independent features that can be added incrementally they don't block `bq` support.

The LIMITATIONS.md entry (BQ-DISMAX-MM) documents the one edge case where DisMax `bq` with boolean/negation would interact differently with `mm` if/when `mm` support is added. IMO would be good to keep it here, but let me know if you think otherwise.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: The `minimum_should_match` concern was valid, fixed by always ensuring a `must` clause exists in the else branch (either the existing query or `match_all`). When `must` is present, `should` is purely optional (scoring-only), matching Solr's `bq` semantics.

   Regarding `function_score`/`constant_score`, `bool.should` with boost is the closest practical equivalent. Both Solr and OpenSearch use BM25 scoring, and `should` with boost produces additive scoring influence. `function_score` would add significant complexity and would break when `bq` contains complex queries
   (phrases, ranges, booleans). The scoring approximation is a known trade-off that applies to all query translation, not specific to `bq`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Added boost preservation tests: integer boost (`^10`), decimal boost (`^2.5`), boost on phrase query, boost on grouped expression, and different boosts across multiple `bq` params.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: No, `should` is always set as an array of clause Maps, never a single Map. The `|| []` fallback handles the case where no `should` key exists yet, and the spread `[...existingShould, ...shouldClauses]` works correctly on arrays. This is the same pattern used in `filter-query-fq.ts`: `existingBool.get('filter') ||
   []`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Solr explicitly allows and processes duplicate `bq` params — each one adds a separate scoring clause. If a user sends `bq=category:food^10&bq=category:food^10`, Solr doubles the boost contribution, and the shim matches that behavior. Deduplicating would silently change scoring semantics. If duplicates are
   unintended, that would be a client-side concern, not something the shim should second-guess.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Fail-fast is already inherited from `translateQ()`, it throws on unsupported syntax like local params and function queries. So `bq={!func}sum(1,price)` will throw today. When new syntax support lands (e.g., function queries via PR #2780), `bq` picks it up automatically through `translateQ()` with no changes needed here. The doc comment is more of a forward-looking to reflect this extensible design.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Ack. The else branch now always sets `must: [existingQuery ?? match_all]`. This ensures `should` is purely optional (scoring-only) regardless of whether a base query exists, matching Solr's `bq` semantics.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Ack. Added a fail-fast check that rejects `bq` values containing `^-`. Without this guard, the PEG grammar would reinterpret `category:spam^-10` as `-(category:spam^10)` (negation with positive boost), which **excludes** matching documents instead of demoting them a silent correctness bug that changes the
   result set. Solr uses negative boost for additive score reduction, but OpenSearch doesn't support negative boost values (undefined scoring behavior). Added a `BQ-NEG-BOOST` entry to LIMITATIONS.md with a workaround.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Ack. The else branch now always sets `must: [existingQuery ?? match_all]`. When `bq` is used without `q`, the query becomes `{ bool: { must: [match_all], should: [...] } }` matching all documents with `should` clauses purely influencing scoring, which matches Solr's behavior.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: ack

TrafficCapture/SolrTransformations/docs/LIMITATIONS.md
niravpi: Validated that defType=dismax is explicitly supported in the shim — not just in bq. The parser (parser.ts) checks defType === 'edismax' || defType === 'dismax' and applies qf field distribution for both. query-q.ts declares defType and qf as supported params. E2e tests in query-fields.testcase.ts already run 5
shared test cases for both parser types, and the bq e2e tests add another 7 shared cases per defType. DisMax-specific params not yet implemented (pf, ps, qs, q.alt, bf) are rejected by validation as unsupported
So I think we should be good

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Traced all code paths that produce a bool query before bq runs. When should is absent, existingBool.get('should') returns undefined, and the || [] fallback converts it to an empty array. When present, it's always an array. No code path sets should to a non-array. This is the same pattern used by fq (existingBool.get('filter') || []) which has been working correctly.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Confirmed that Solr processes duplicate bq params independently for additive scoring. The shim matches thi 2 identical ``` bq=category:food^10``` produce 2 separate should clauses. Deduplicating would change scoring semantics.

   Added edismax-bq-duplicate e2e integration test that sends identical duplicate bq values and validates Solr and OpenSearch produce the same result set. Passed on both Solr 8 and Solr 9.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Found a gap here. While `{!func}` correctly throws via `translateQ()` (no transform rule for `func` nodes), `{!boost b=2}category:food` was silently producing `match_all` — the boost metadata and body were both lost. This happens because the parser's `resolveLocalParamsBody` only knows how to parse bodies for
   `lucene`/`dismax`/`edismax`/`func` types; unknown types like `boost` get `body=null`, and the transformer returns `match_all`.

  The `q` param was already protected by a `rejectPattern: ^\{!` in `query-q.ts`, but `bq` had no `paramRules` at all.

   **Fix**: Added `paramRules` with `rejectPattern: ^\{!` to `boost-query-bq.ts` — all local params in bq are now rejected at validation time with a clear error message.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,94 @@
...
+function isBoolQuery(query: Map<string, any>): boolean {
+  return query.size === 1 && query.has('bool');
+}
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'boost-query-bq',
+  match: (ctx) => ctx.params.has('bq'),
+  apply: (ctx) => {
+    const defType = ctx.params.get('defType') ?? '';
+    if (!DISMAX_TYPES.has(defType)) {
```
k-rooot, line 63: Solr defaults to `lucene` if defType is not provided. Can this strict validation reject valid queries that implicitly rely on defaults?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+function parseBq(bq: string): Map<string, any> {
+  return translateQ(new Map([['q', bq]])).dsl;
+}
+
+/**
+ * Check if a query is a bool query (has 'bool' as its only key).
+ */
+function isBoolQuery(query: Map<string, any>): boolean {
+  return query.size === 1 && query.has('bool');
```
k-rooot, line 55: Can this fail for valid OpenSearch queries that include metadata alongside bool (e.g - boost, _name).

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!DISMAX_TYPES.has(defType)) {
+      throw new Error(
+        `[boost-query-bq] bq parameter requires defType=edismax, got '${defType || '(none)'}'`,
+      );
+    }
+
+    const bqValues = ctx.params.getAll('bq').filter((v) => v.trim() !== '');
+    if (bqValues.length === 0) return;
+
+    if (bqValues.some((v) => /\^-/.test(v))) {
```
k-rooot: The regex is too broad and may produce false positives (e.g., field values containing ^-)

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    const existingQuery = ctx.body.get('query');
+
+    if (existingQuery && isBoolQuery(existingQuery)) {
+      const existingBool = existingQuery.get('bool') as Map<string, any>;
+      const existingShould = existingBool.get('should') || [];
+      existingBool.set('should', [...existingShould, ...shouldClauses]);
+    } else {
+      const boolQuery = new Map<string, any>();
+      boolQuery.set('must', [existingQuery ?? new Map([['match_all', new Map()]])]);
```
k-rooot, line 89: Wrapping existingQuery in must introduce unintended nesting if it's already a bool or compound query?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
k-rooot: Solr bq supports inline boosts (field:value^10). This relies on translateQ to preserve boosts. Do we need any tests to validate this scenario?

**Response From Nirav [ As somehow git is not allowing to reply so editing the comment and adding response here]**

Already covered. The `boost preservation` section has 7 tests: integer boost (`^10`), decimal boost (`^2.5`), phrase boost (`^3`), grouped expression boost (`^5`), different boosts across multiple bq params, zero boost (`^0`), and large boost (`^99999`). Zero and large boost tests added in this revision per the edge case feedback.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+import { request } from './boost-query-bq';
+import type { RequestContext, JavaMap } from '../context';
+
+/** Create a mock RequestContext for testing. */
+function createCtx(
```
k-rooot, line 6: The helper does not allow injecting malformed or edge-case body.query shapes (e.g., non-Map, unexpected types), which limits negative testing.

We should allow passing raw any for existingQuery to test defensive handling. Please cover any missing negative tests

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+  it('has correct name', () => {
+    expect(request.name).toBe('boost-query-bq');
+  });
+
+  // --- match guard ---
+
+  describe('match', () => {
+    it('matches when bq is present', () => {
+      const ctx = createCtx({ bq: 'category:food', defType: 'edismax' });
+      expect(request.match!(ctx)).toBe(true);
```
k-rooot, line 56: Should we add test case where bq exists but is empty (bq=''). Current logic will match but later no-op.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+
+  describe('single bq', () => {
+    it('wraps existing query in bool.must and adds bq to bool.should', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'category:food', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.must).toHaveLength(1);
```
k-rooot, line 104: This checks structure but not semantic correctness (i.e., that should does not affect matching). Should we add assertion for minimum_should_match if enforced in implementation.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+      const ctx = createCtx(
+        { bq: ['category:food^10', 'category:deli^5'], defType: 'edismax' },
+        existingQuery,
+      );
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(2);
+      expect(result.bool.should[0].match.category.query).toBe('food');
```
k-rooot, line 159: Relies on ordering of should clauses, which may not be guaranteed if deduplication or normalization is introduced.

This comment is applicable in other tests as well

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx(
+        { bq: ['category:food', '', '   ', 'type:book'], defType: 'edismax' },
+        existingQuery,
+      );
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(2);
```
k-rooot, line 175: Should we also validate the correctness of retained values.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+    });
+
+    it('returns early when all bq values are empty', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: ['', '   '], defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      expect(ctx.body.get('query')).toBe(existingQuery);
```
k-rooot, line 184: Can we verify that hat no mutation occurred.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+      ]);
+      const existingQuery = new Map([['bool', existingBool]]);
+      const ctx = createCtx({ bq: 'category:tech^5', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.must).toHaveLength(1);
+      expect(result.bool.must[0].match.title.query).toBe('java');
+      expect(result.bool.should).toHaveLength(1);
```
k-rooot, line 205: Missing validation that existing should clauses are preserved when present.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+        ['should', [new Map([['match', new Map([['status', new Map([['query', 'featured']])]])]])]],
+      ]);
+      const existingQuery = new Map([['bool', existingBool]]);
+      const ctx = createCtx({ bq: 'category:food^10', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(2);
+      expect(result.bool.should[0].match.status.query).toBe('featured');
```
k-rooot, line 222: Again relies on ordering. Applicable for rest of the PR

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+    });
+
+    it('handles wildcard bq', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'title:wire*^5', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should).toHaveLength(1);
```
k-rooot, line 327: Does not verify correct query type. Should we assert wildcard presence.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
```diff
@@ -0,0 +1,399 @@
...
+
+  describe('boost preservation', () => {
+    it('preserves integer boost on field query', () => {
+      const existingQuery = new Map([['match_all', new Map()]]);
+      const ctx = createCtx({ bq: 'category:food^10', defType: 'edismax' }, existingQuery);
+
+      request.apply(ctx);
+
+      const result = mapToObj(ctx.body.get('query'));
+      expect(result.bool.should[0].match.category.boost).toBe(10);
```
k-rooot, line 371: Can we cover edge cases - zero boost, malformed boost and large boost

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Yes, and that's intentional. `bq` is a DisMax/eDisMax-only parameter, it has no meaning in the Lucene query parser. If a user sends `bq=category:food` without `defType=edismax`, Solr itself ignores the `bq` param (it's only processed by the DisMax/eDisMax QParser). The shim fails fast instead of silently dropping the param, which surfaces misconfiguration early. A query relying on Lucene defaults wouldn't use `bq` in the first place.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: No. `isBoolQuery` (`query.size === 1 && query.has('bool')`) only inspects the query Map produced by earlier transforms in the pipeline (`translateQ` → `query-q` → `fq` → `bq`). The shim's transformer never adds `boost` or `_name` at the top-level query Map, those are set inside the `bool` Map's children, not alongside `bool`. This is the same pattern used in `filter-query-fq.ts` and `filterRule.ts`. If a future transform adds top-level metadata, `isBoolQuery` would return false and the else-branch would wrap it correctly no silent failure.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: Ack. Tightened from `/\^-/` to `/\^-\d/` now only matches caret-minus followed by a digit (actual negative boost syntax like `^-10`, `^-2.5`). A quoted phrase like `title:"some^-thing"` no longer triggers a false positive. Added unit test: `'does not false-positive on quoted values containing ^-'`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.ts
niravpi: No. The else-branch only executes when `isBoolQuery(existingQuery)` returns false meaning the existing query is NOT a bool query (it's a `match`, `match_all`, `range`, etc.). If it were already a bool, the if-branch handles it by flattening `should` directly into the existing bool. Added test `'wraps range query in must without nesting'` to explicitly verify the else-branch with a non-bool query, including `minimum_should_match` undefined assertion.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: The `createCtx` helper accepts `Map<string, any>` which matches the runtime contract the shim pipeline always produces Maps via GraalVM interop. Testing with non-Map types (plain objects, strings) would test an impossible runtime state since all upstream transforms (`query-q`, `fq`) output Maps. For defensive handling of non-standard Map shapes, added `'wraps range query in must without nesting'` which tests the else-branch with a compound non-bool query. The defType validation tests cover the negative paths for invalid/missing defType, and the negative boost tests cover invalid boost values. Please let me know if anything might be missing and we should be adding.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Ack. Added `'matches when bq is empty string (apply handles the no-op)'` verifies `match` returns true for `bq=''`, and the existing `'returns early when all bq values are empty'` test already verifies `apply` is a no-op with no mutation (reference check + size assertions).

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Ack. Added `expect(result.bool.minimum_should_match).toBeUndefined()` to the single bq test, the no-existing-query test, and the range query test. The implementation ensures `should` is purely scoring-only by always having a `must` clause present (either the existing query or `match_all`). When `must` is present,
  OpenSearch treats `should` as optional (scoring-only) without needing explicit `minimum_should_match: 0`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: The ordering is deterministic by design `[...existingShould, ...shouldClauses]` always appends new clauses after existing ones. Same append-only contract used by `filter-query-fq.ts` (`[...existingFilters, ...filters]`). Testing the order verifies the actual contract: existing clauses preserved in position, new clauses appended. Using unordered assertions (`arrayContaining`) would weaken the tests and miss regressions where existing clauses get reordered.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Ack. Strengthened the `'merges with existing should clauses'` test to also assert `result.bool.should[1].match.category.boost === 10` verifying both the retained existing clause value and the new clause's boost value.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Ack. Strengthened the `'returns early when all bq values are empty'` test now asserts `ctx.body.size === 1` (no extra keys added to body) and `existingQuery.size === 1` (the query Map wasn't mutated). Combined with the existing `toBe` reference check, this proves no mutation at any level.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: The `'merges with existing should clauses'` test validates this, sets up a bool with existing `should: [status:featured]`, adds `bq=category:food^10`, asserts `should.length === 2` with first clause `status:featured` and second `category:food` with `boost: 10`. Also strengthened `'flattens into existing bool
  query'` to assert the boost value (`category.boost === 5`) in addition to the query text.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Same as the ordering comment above — `[...existingShould, ...shouldClauses]` is deterministic. The append order is the contract being tested.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Ack. Added `expect(result.bool.should[0].wildcard).toBeDefined()` and `expect(result.bool.should[0].wildcard.title).toBeDefined()`.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/boost-query-bq.test.ts
niravpi: Added `'preserves zero boost'` (`^0` → `boost: 0`) and `'preserves large boost'` (`^99999` → `boost: 99999`). Malformed boost (e.g., `^abc`) is caught by `translateQ()` which throws a parse error in fail-fast mode, this is the parser's responsibility, not bq-specific.

k-rooot (approved):

PR #2794, opened 2026-04-24
Added Facet Translations for Common Stat Functions

akshay2000:
### Description

Adds support for Solr stat/metric facet shorthand in `json.facet` conversion. String facet definitions like `"avg(price)"` are now parsed using the function query grammar and mapped to OpenSearch metric aggregations.

Supported mappings:
- `avg`, `sum`, `min`, `max` → direct OpenSearch equivalents
- `unique`, `hll` → `cardinality`
- `countvals` → `value_count`

Unsupported stat functions throw a clear error.

### Testing
- Unit tests

### Check List

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
@@ -11,6 +11,8 @@ import type { MicroTransform } from '../pipeline';
 import type { RequestContext, JavaMap } from '../context';
 import type { ParamRule } from './validation';
 import { convertSort, isMapLike, isSolrDateMathGap, convertSolrDateGap } from './utils';
+import { parseFuncQuery } from '../query-engine/parser/parser';
+import type { FuncNode } from '../query-engine/ast/nodes';
 
 /** Solr query params this feature handles. */
 export const params = ['json.facet'];
@@ -435,6 +437,56 @@ function convertQueryFacet(def: JavaMap): JavaMap {
 
 // endregion
 
+// region Stat (metric) facets
+
+/**
+ * Solr stat function → OpenSearch metric aggregation mapping.
+ *
+ * Solr's json.facet allows shorthand stat facets as plain strings:
+ *   "avg_price": "avg(price)"
+ * These map to OpenSearch metric aggregations:
+ *   "avg_price": { "avg": { "field": "price" } }
+ */
+const STAT_FUNC_MAP: Record<string, string> = {
+  avg: 'avg',
+  sum: 'sum',
+  min: 'min',
+  max: 'max',
+  unique: 'cardinality',
+  hll: 'cardinality',
+  countvals: 'value_count',
+};
+
+/**
+ * Convert a Solr stat facet string (e.g., "avg(price)") to an OpenSearch metric agg.
+ */
+function convertStatFacet(expr: string): JavaMap {
+  const func = parseFuncQuery(expr);
+  return convertFuncToMetricAgg(func);
+}
+
+function convertFuncToMetricAgg(func: FuncNode): JavaMap {
+  const osAggType = STAT_FUNC_MAP[func.name];
+  if (!osAggType) {
+    throw new Error(`Unsupported stat function '${func.name}' in json.facet`);
+  }
+
+  if (func.args.length !== 1) {
+    throw new Error(`Stat function '${func.name}' expects exactly 1 argument, got ${func.args.length}`);
+  }
+
+  const arg = func.args[0];
+  if (!('kind' in arg) || arg.kind !== 'field') {
+    throw new Error(`Stat function '${func.name}' expects a field reference argument`);
+  }
+
+  const inner = new Map<string, any>();
+  inner.set('field', arg.name);
+  return new Map<string, any>([[osAggType, inner]]);
+}
+
+// endregion
+
 /** Log a warning for any keys in a facet definition that are not in the known set. */
 function warnUnknownKeys(def: JavaMap, knownKeys: Set<string>, facetType: string): void {
   const unknownKeys: string[] = [];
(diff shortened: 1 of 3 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
```diff
@@ -498,7 +550,11 @@ export function convertJsonFacets(solrJsonFacet: JavaMap, ctx: RequestContext):
   const aggs = new Map<string, any>();
   for (const name of solrJsonFacet.keys()) {
     const facetDef = solrJsonFacet.get(name);
-    aggs.set(name, convertSingleFacet(facetDef, ctx));
+    if (typeof facetDef === 'string') {
```
k-rooot, line 553: Is it safe to assume all string values are stat functions or should we validate that the string is actually a function expression before parsing?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
```diff
@@ -435,6 +437,56 @@ function convertQueryFacet(def: JavaMap): JavaMap {
...
+
+/**
+ * Solr stat function → OpenSearch metric aggregation mapping.
+ *
+ * Solr's json.facet allows shorthand stat facets as plain strings:
+ *   "avg_price": "avg(price)"
+ * These map to OpenSearch metric aggregations:
+ *   "avg_price": { "avg": { "field": "price" } }
+ */
+const STAT_FUNC_MAP: Record<string, string> = {
```
k-rooot, line 450: Should we add support for `count()` which has a wide adoption?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
```diff
@@ -435,6 +437,56 @@ function convertQueryFacet(def: JavaMap): JavaMap {
...
+  return convertFuncToMetricAgg(func);
+}
+
+function convertFuncToMetricAgg(func: FuncNode): JavaMap {
+  const osAggType = STAT_FUNC_MAP[func.name];
+  if (!osAggType) {
+    throw new Error(`Unsupported stat function '${func.name}' in json.facet`);
+  }
+
+  if (func.args.length !== 1) {
```
k-rooot, line 474: Some Solr functions allow constants, nested func and expressions like `sum(product(price, qty))`. 

If its a conscious choice to not support these use cases, we could document this in limitations file.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
```diff
@@ -435,6 +437,56 @@ function convertQueryFacet(def: JavaMap): JavaMap {
...
+    throw new Error(`Stat function '${func.name}' expects exactly 1 argument, got ${func.args.length}`);
+  }
+
+  const arg = func.args[0];
+  if (!('kind' in arg) || arg.kind !== 'field') {
```
k-rooot, line 479: Solr supports expressions like `avg(price * qty)` and `sum(if(exists(field), field, 0))`. 

The current rule rejects these expressions.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/parser.ts
```diff
@@ -140,6 +140,15 @@ export function toParseError(err: unknown): ParseError {
   };
 }
 
+/**
+ * Parse a Solr function query string (e.g., "avg(price)") into a FuncNode.
```
k-rooot, line 144: Solr also supports `"avg_price": "avg(price,0)"

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
akshay2000: Before this PR: always an error if it is a string.
After the PR: Try out if we can treat it as a function. Error if we can't. So, we are validating by parsing.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
akshay2000: Count is when nothing is mentioned.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
akshay2000: Our intention with limitations list is to document the stuff that arises from differences in engines themselves and not the limitations that arise from (lack of) feature completeness. In this case, there will be an explicit error informing the user that this is not supported. When we (eventually) do support it the error will be gone.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
akshay2000: Binary operator like `avg(price * qty)` doesn't seem like a supported scenario at least in facets context. As for the nested functions, translating them will require scripting on OpenSearch side. The goal of this PR is to support the simplest expressions of type `stat(field)`. Everything else can be an enhancement.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/parser.ts
akshay2000: As called out earlier, we only want to support the simplest of the stats which have a direct equivalent in OpenSearch. Everything more complex than that should come as enhancements. Until then the user gets a clear errors when some unsupported scenarios occur.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
k-rooot: Interesting. I see the LIMITATIONS file is currently being used to document unsupported features as well.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/parser.ts
k-rooot: I’d lean toward prioritising the stats that have broader adoption rather than just the simplest ones. That said, this isn’t a blocker for the PR.

k-rooot (approved):

PR #2795, opened 2026-04-24
Restore Log4J tuple path as default when --tuple-s3-bucket is unset

AndreKurait:
## Description

`TrafficReplayer` has been producing NPEs on every tuple since the S3 tuple sink landed:

```
java.lang.NullPointerException: Cannot invoke
  "java.util.function.Consumer.accept(Object)" because "tupleConsumer" is null
  at TrafficReplayerCore.java:405
```

### Root cause

Commit `5b5c23f10` ("Replace Mountpoint S3 sidecar with direct S3 CRT client writes") replaced the `--enable-sync-tuples` toggle with `--tuple-s3-bucket` selector — but in the process **deleted the else-branch** that wired up the Log4J tuple path for the default (no-S3) case.

After that change, `TrafficReplayer.main` unconditionally invoked:
```java
var tupleWriter = createS3TupleWriterIfConfigured(params);  // returns null when --tuple-s3-bucket unset
tr.setupRunAndWaitForReplayWithShutdownChecks(..., tupleWriter, ...);
```

`createS3TupleWriterIfConfigured` returns `null` when no bucket is configured. The null `ThreadLocalTupleWriter` flowed into `TrafficReplayerCore.handleCompletedTransaction`, whose `tupleWriter == null` branch forwarded a null `Consumer<SourceTargetCaptureTuple>` into `packageAndWriteResponse` → NPE at line 405 on every single completed transaction.

### Fix

Restore the pre-regression behavior: when `--tuple-s3-bucket` is **not** set, construct `ResultsToLogsConsumer` + `TupleParserChainConsumer` (both still in tree, unchanged) and dispatch through the legacy `Consumer<SourceTargetCaptureTuple>` overload at `TrafficReplayerTopLevel.java:330`. This matches the pre-`c1c636962` wiring exactly.

### What this PR does NOT change

- `TrafficReplayerCore` — untouched. Its dual-path handling (`ThreadLocalTupleWriter` vs `Consumer`) was correct; it just had no caller for the Consumer path.
- `TrafficReplayerTopLevel` — untouched. The legacy overload was already present.
- `ResultsToLogsConsumer`, `TupleParserChainConsumer` — untouched. Still in tree, still tested by `ResultsToLogsConsumerTest`.
- The S3 path — unchanged behavior when `--tuple-s3-bucket` is set.
- No new flags, no new dependencies, no behavior changes to either sink.

### Diff

Commit 1 (`TrafficReplayer.java`): +22 / −8.
Commit 2 (`full_tests.py`): +59 / −0 — integration regression test (see below).

### Testing

**Unit suite** — `./gradlew :TrafficCapture:trafficReplayer:test` — BUILD SUCCESSFUL, all suites green. Existing `TupleWriteBlockingBehaviorTest`, `ResultsToLogsConsumerTest`, `OpenSearchDefaultRetryE2ETest`, etc. all pass.

**New integration regression test** — `test_e2e_0001_default` in `migrationConsole/lib/integ_test/integ_test/full_tests.py` now explicitly guards against the `5b5c23f10` shape of this regression. The original bug was silent to the existing doc-count-based tests because the NPE was caught per-tuple and logged — docs were still backfilled, replay still completed, but zero tuples reached disk. The new assertion closes that gap end-to-end in the `full-es68-e2e-aws-test` pipeline:

1. Record `pytest.test_start_time` in the `initialize` fixture.
2. Run the normal E2E traffic (`create_index` + three captured documents + backfill + replay + doc-count matching) — unchanged.
3. Sleep 65s to cross Log4J2's one-minute rotation boundary.
4. Issue one more captured source `get_document` so `TimeBasedTriggeringPolicy` has a post-boundary event to act on (the policy is event-driven, sleeping alone will not rotate).
5. Sleep 15s for the capture → Kafka → replayer → tuple emission → rotation → gzip pipeline to settle.
6. Glob `/shared-logs-output/traffic-replayer-*/*/tuples/tuples-*.log.gz` on the shared EFS mount, filter by `mtime >= pytest.test_start_time` (EFS is reused across pipeline runs).
7. Assert at least one rotated file exists; decompress each with `gzip.open` and count parseable JSON tuple lines.
8. Assert `tuple_count > 0`, with an error message that names the `5b5c23f10` regression shape so future breakage is self-diagnosing.

The assertion is intentionally `> 0`, not an exact count — the capture proxy currently records ALB `ELB-HealthChecker` requests alongside user traffic, which would make an exact count flaky. Wiring `--suppressCaptureForHeaderMatch User-Agent ELB-HealthChecker` in CDK is tracked as a follow-up; once in, this assertion can be tightened to an exact expected count.

### Check List
- [x] New functionality includes testing (new integration regression test + existing unit suite)

### DIFF
--- a/.github/workflows/jenkins_tests.yml
+++ b/.github/workflows/jenkins_tests.yml
@@ -42,7 +42,7 @@ jobs:
           job_params: "GIT_REPO_URL=${{ needs.sanitize-input.outputs.pr_repo_url }},GIT_BRANCH=${{ needs.sanitize-input.outputs.branch_name }},GIT_COMMIT=${{ needs.sanitize-input.outputs.commit_hash }}"
           jenkins_user: "${{ secrets.JENKINS_MIGRATIONS_USER }}"
           jenkins_api_token: "${{ secrets.JENKINS_MIGRATIONS_API_TOKEN }}"
-          job_timeout_minutes: "120"
+          job_timeout_minutes: "180"
 
   elasticsearch-5x-k8s-local-test:
     needs: [get-require-approval, sanitize-input]
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
@@ -700,14 +700,28 @@ public class TrafficReplayer {
 
             setupShutdownHookForReplayer(tr);
             var tupleWriter = createS3TupleWriterIfConfigured(params);
-            tr.setupRunAndWaitForReplayWithShutdownChecks(
-                Duration.ofSeconds(params.observedPacketConnectionTimeout),
-                serverTimeout,
-                blockingTrafficSource,
-                timeShifter,
-                tupleWriter,
-                Duration.ofMillis(params.quiescentPeriodMs)
-            );
+            if (tupleWriter != null) {
+                tr.setupRunAndWaitForReplayWithShutdownChecks(
+                    Duration.ofSeconds(params.observedPacketConnectionTimeout),
+                    serverTimeout,
+                    blockingTrafficSource,
+                    timeShifter,
+                    tupleWriter,
+                    Duration.ofMillis(params.quiescentPeriodMs)
+                );
+            } else {
+                var resultsToLogsConsumer = new ResultsToLogsConsumer(null, null,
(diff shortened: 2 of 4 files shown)


jugal-chauhan (approved):
LGTM! Thanks for brining in these changes. The regression test in `full_tests.py` is a great addition. This bug was particularly tricky as it was silent. Doc counts passed, replay completed, but zero tuples reached EFS. Having the end-to-end tuple file assertion closes that coverage gap and should catch this class of issue going forward (with the ECS solution)

PR #2798, opened 2026-04-24
Use minified Argo CRDs to fix air-gapped deployment failure

oeyh:
### Description
Set `crds.full: false` for the argo-workflows Helm chart. When `crds.full: true`, the chart
creates a pre-install Job that downloads full CRD YAML files from `raw.githubusercontent.com`.
This fails in air-gapped environments where the EKS cluster has no internet access, blocking
the Migration Assistant helm install.

With `crds.full: false`, minified CRDs are embedded inline in the chart templates and require
no internet access. The tradeoff is losing full OpenAPI schema support (`kubectl explain` on
Argo CRDs and Validating Admission Policy support), which is not critical for MA functionality.

### Issues Resolved
N/A (discovered during air-gapped deployment testing)

### Testing
Tested in an isolated VPC.

### Check List

By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -335,8 +335,9 @@ charts:
       images:
         pullPolicy: IfNotPresent
       crds:
-        # Full CRDs enable Validating Admission Policy support and kubectl explain on Argo CRDs
-        full: true
+        # Full CRDs require downloading from GitHub at install time, which fails in air-gapped environments.
+        # Minified CRDs are embedded inline in the chart and work without internet access.
+        full: false
```
AndreKurait, line 340: My preference would be to update our Jenkins non-release tests to this with true

oeyh:
Close in favor of https://github.com/opensearch-project/opensearch-migrations/pull/2839

PR #2799, opened 2026-04-25
Enable performance improvements for building and deploying k8s locally

gregschohn:
### Description
This branch restructures local image-build and local-test workflows to make developer environments more reliable and faster, especially for kind and minikube based testing.

It does five main things:

- Adds first-class kind local testing support, including cluster creation, cleanup, registry wiring, and a dedicated kindTesting.sh flow.
- Refactors local build setup into reusable backends so minikube and kind can share deployment logic while using different BuildKit/runtime wiring.
- Fixes BuildKit intermediate image handoff so downstream image builds consume the correct producer digests instead of relying on brittle tag-based behavior.
- Switches the Docker-hosted BuildKit path from a port-forwarded remote builder model to a docker-container builder model for local Docker-backed environments.
- Parallelizes BuildKit image builds and fixes Docker-hosted BuildKit DNS resolution issues that were causing image build failures.

`localTesting.sh` continues to bring up a minikube environment, build, and install MA into it.  It now uses a 'bake' feature of buildkit to enable building docker images in parallel.  `kindTesting.sh` has been added to bring up a kind environment, do builds, and deploy MA to it.  The kind environment doesn't use k8s for buildkit or registry containers.  Instead it uses docker directly.  The registry container is backed by a docker volume so that restarting the registry (or the now unrelated k8s cluster) won't cause the cached layers to be permanently lost - they should come back when the registry comes back.

You can run minikube and kind with MA simultaneously if you have the hardware to support that - the kind path puts the registry on the local port 5002 to not conflict with 5001.

### Issues Resolved
No Jira - but I have seen an issue in the localtesting script around where it waits for strimzi.

### Testing
Here are the numbers that I got between mainline minikube builds and ones for this branch.  I tested on an m1 pro with 32GB and an m5 pro with 64GB.  These tests were COMPLETELY cold - neither minikube nor kind had any layers cached.  Gradle build caches were purged.  These would be the times that a new developer would see if they've never run kind or minikube before.

```
FasterDevEnvironment:
 m1 pro
  minikube: 832s
  kind: 801s

 m5 pro
  minikube: 476s
  kind: 389s

Main branch (+ CoreDNS fix)
 m1 pro
  minikube: 916s

 m5 pro
  minikube: 559s
```

I also did a test on the m5 with orbstack replacing docker for this branch.  I noticed that kind was 12s faster and minikube came in 13s faster.  The number one thing dominating this time now is building the migration console image.  If we reduce its size and the number of build steps, that would shave the overall time for this build and many builds (since it often picks up changes, since it bundles so much together)

### Check List


buildImages/backends/k8sHostedBuildkit.sh
```diff
@@ -0,0 +1,173 @@
...
+  if [[ "${context}" =~ (eks:|gke_|aks-|migration-eks-) ]]; then
+    echo "Detected cloud K8s, using kubernetes driver with native multi-arch builds"
+    BUILDKIT_RESOURCE_OPTS=(
+      --driver-opt="requests.cpu=${BUILDKIT_REQUESTS_CPU:-4}"
+      --driver-opt="requests.memory=${BUILDKIT_REQUESTS_MEMORY:-8Gi}"
+      --driver-opt="limits.cpu=${BUILDKIT_LIMITS_CPU:-8}"
+      --driver-opt="limits.memory=${BUILDKIT_LIMITS_MEMORY:-16Gi}"
+    )
+
+    docker buildx create \
```
AndreKurait, line 104: nit: For developers we should consider adding an option to disable one of the architectures. For example, the default EKS nodeclass seems to only allow one

AndreKurait (approved):

PR #2810, opened 2026-04-27
Adds the OpenSearch pricing calculator into the migration agent

jzonthemtn:
### Description
Adds the OpenSearch pricing calculator into the agent.

Requires https://github.com/opensearch-project/opensearch-migrations/pull/2793 to get the Docker image built.

Screenshot of the pricing:

image

---


wrigleyDan:
@sumobrian & @AndreKurait The integration of the pricing calculator is the first of several tools that'll be integrated with the agent skill. 
Currently, there are no pre-built Docker containers for any of these which makes a combination of the different initiatives challenging without introducing dependencies.
We think the easiest way to combine all is through a `docker-compose.yml` file but we'd like to make certain this aligns with your thinking. With the current repository setup we can have custom built Docker images that we bring together in a `docker-compose.yml` based on the repository structure.
Is that also what you were thinking?

arunx2 (commented):
## Review Feedback

Nice work wiring the pricing calculator into the migration agent! The Python client is clean, well-documented, and the test coverage is thorough. A few items to address:

### 1. Broken symlink — `.claude/skills/migration-advisor/SKILL.md`

The symlink target `./tests/scripts/../../SKILL.md` doesn't resolve correctly from within `.claude/skills/migration-advisor/`. The `tests/scripts/` directories don't exist at that path, so this symlink will be dangling. The other symlink (`solr-opensearch-migration-advisor/SKILL.md → ./../../SKILL.md`) looks correct — this one needs a similar fix with the right relative path.

### 2. Dockerfile — `priceCache.json` COPY should stay

The `priceCache.json` file exists in the repo and caching the pricing API responses saves startup time. Rather than commenting out the COPY with a question, this line should remain active. If the file isn't being copied into the builder stage, the Dockerfile should ensure it's available (e.g., via a separate `COPY` of just that file into the build context).

### 3. `stop_container.sh` needs the compose file path

`start_container.sh` correctly uses `-f "$SCRIPT_DIR/docker-compose.yml"`, but `stop_container.sh` runs bare `docker compose down` which won't find the right project unless you happen to be in that directory. It should use the same `-f` flag. Also, `docker stop claude-container` after `docker compose down` is redundant since compose already stops the container — this line can be removed.

### 4. `_handle_pricing` doesn't route the "search" workload type

In `skill.py`, `_handle_pricing` checks for `"serverless"`, `"vector"`, and `"time series"` but never routes to `_pricing_prompt_search()`. If a user says "pricing for search workload", it falls through to the generic "which workload type?" prompt. Adding a check for `"search"` (being careful not to match "serverless" which also contains "search") would fix this.

### 5. Duplicate `if __name__ == "__main__"` in `test_pricing_calculator.py`

There are two `if __name__ == "__main__": unittest.main()` blocks — one at ~line 301 and another at the end. The first one will exit before the `TestPricingCalculatorIntegration` class is ever reached when running directly, which contradicts the docstring instructions. The first block should be removed or moved to the very end of the file.

### 6. Integration test `BASE_URL` vs. docstring mismatch

`TestPricingCalculatorIntegration.BASE_URL` is set to `http://opensearch-pricing-calculator:5050` (the Docker service name), but the class docstring tells users to start the calculator on `localhost:5050`. These should be consistent — either update the `BASE_URL` to use `localhost` for local runs, or read from an env var so it works in both contexts.

---

Overall the pricing client module, SKILL.md additions, and Docker Compose setup are well done. Just needs these fixes before merging.

jzonthemtn:
@arunx2 I don't see the `priceCache.json` file in the repository. A `find . -name "priceCache.json"` from a fresh clone on the root directory doesn't find one. Could it perhaps exist in your clone?

jzonthemtn:
Example pricing estimates after pushing PR changes:

```
  Cluster Configurations (ranked by OnDemand cost)                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                          
  ┌─────────────────┬──────────────────────┬───────────────────────┬──────────┬─────────────┬─────────────────────┐
  │      Rank       │    Manager nodes     │      Data nodes       │ JVM/node │ OnDemand/mo │ 3-yr All-Upfront/mo │                                                                                                                                                                       
  ├─────────────────┼──────────────────────┼───────────────────────┼──────────┼─────────────┼─────────────────────┤
  │ 1 — cheapest    │ m5.large.search × 3  │ t3.small.search × 6   │ 1 GB     │ $475.98     │ $156.57             │                                                                                                                                                                       
  ├─────────────────┼──────────────────────┼───────────────────────┼──────────┼─────────────┼─────────────────────┤
  │ 2 — recommended │ m5.large.search × 3  │ t3.medium.search × 6  │ 2 GB     │ $638.04     │ $374.07             │                                                                                                                                                                       
  ├─────────────────┼──────────────────────┼───────────────────────┼──────────┼─────────────┼─────────────────────┤                                                                                                                                                                       
  │ 3 — compute-opt │ m6g.large.search × 3 │ c6g.xlarge.search × 3 │ 4 GB     │ $778.92     │ $375.78             │                                                                                                                                                                       
  └─────────────────┴──────────────────────┴───────────────────────┴──────────┴─────────────┴─────────────────────┘         
  ```
  
  For testing, a shortcut is to use the prompt "Help me migrate Solr to OpenSearch. Jump to pricing estimates."

File: AIAdvisor/opensearch-pricing-calculator/Dockerfile [hunk 1]
```diff
@@ -28,7 +28,9 @@ RUN addgroup -S appgroup && adduser -S appuser -G appgroup
 WORKDIR /app
 
 COPY --from=builder /build/opensearch-pricing-calculator .
-COPY --from=builder /build/priceCache.json .
```
sumobrian, line 31: From @arunx2 - we should either build the cache at each run( which is not performant) or build the pricecache.json during the docker build. We can also check in the file but currently blocked by the .gitignore https://github.com/opensearch-project/opensearch-migrations/blob/main/AIAdvisor/opensearch-pricing-calculator/.gitignore#L28

sumobrian (approved):

PR #2825, opened 2026-04-27
Add frontend mTLS support for Capture Proxy

Akhil911:
### Description                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                 
  #### Summary                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
  Add frontend mTLS (mutual TLS) support to the Capture Proxy, enabling client certificate authentication for incoming connections. This allows clients outside the VPC (e.g., Lambda functions in a public domain) to authenticate securely with
  a Capture Proxy hosted in a private network.                                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
  This introduces three new CLI flags that work with the existing PEM-based TLS configuration:                                                                                                                                                   
  - `--sslTrustCertFile` — Path to PEM trusted CA certificate file for validating client certificates                                                                                                                                            
  - `--requireClientAuth` — Enforce mutual TLS; clients must present a valid certificate signed by the trusted CA                                                                                                                                
  - Both flags are optional and backward-compatible. Normal one-way TLS continues to work when they are not specified.                                                                                                                           
                                                                                                                                                                                                                                                 
  #### Motivation                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                 
  The Capture Proxy currently supports server-side TLS via PEM files (`--sslCertChainFile` + `--sslKeyFile`), but has no mechanism to authenticate incoming clients. For deployments where the Capture Proxy is hosted in a VPC but receives     
  traffic from public-facing services (e.g., Lambda functions in cell accounts), we need a way to verify the client's identity before accepting connections.                                                                                     
                                                                                                                                                                                                                                                 
  mTLS provides this authentication cleanly without requiring additional auth infrastructure — the client presents a certificate signed by a trusted CA, and the proxy validates it as part of the TLS handshake. This pairs naturally with the  
  existing PEM-based TLS implementation.                                                                                                                                                                                                         
                                                                                                                                                                                                                                                 
  #### Changes                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                 
  ##### Java                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                 
  - **`CaptureProxy.java`**: Added two new `@Parameter` annotations — `--sslTrustCertFile` (optional PEM trusted CA path) and `--requireClientAuth` (boolean flag). Modified `loadSslEngineFromPem()` to accept these parameters and configure   
  `SslContextBuilder` with `trustManager(trustCertFile)` and `clientAuth(ClientAuth.REQUIRE)` when applicable. Added parameter validation that throws `ParameterException` when `--requireClientAuth` is specified without `--sslTrustCertFile`. 
                                                                                                                                                                                                                                                 
  ##### Tests                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                 
  - **`FrontsideMtlsTest.java`** (new): Added unit tests covering all four parameter combinations:                                                                                                                                               
    - `testWithClientAuthRequiredReturnsEngineWithNeedClientAuth` — verifies `SSLEngine.getNeedClientAuth() == true` when both `--requireClientAuth` and `--sslTrustCertFile` are provided                                                       
    - `testWithoutClientAuthDoesNotRequireClientCert` — verifies client auth is not required when only `--sslTrustCertFile` is provided                                                                                                          
    - `testWithoutTrustCertAndNoClientAuthSucceeds` — verifies normal TLS (no mTLS) continues to work                                                                                                                                            
    - `testRequireClientAuthWithoutTrustCertThrows` — verifies `ParameterException` is thrown for the invalid combination                                                                                                                        
    - Tests use Netty's `SelfSignedCertificate` to generate test certs, so no external tooling is required                                                                                                                                       
                                                                                                                                                                                                                                                 
  #### Usage guide                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                 
  **Server (Capture Proxy):**                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                 
  --sslCertChainFile /path/to/server-cert.pem \                                                                                                                                                                                                  
  --sslKeyFile /path/to/server-key.pem \                                                                                                                                                                                                         
  --sslTrustCertFile /path/to/client-ca.pem \                                                                                                                                                                                                    
  --requireClientAuth                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                 
  **Client:** must present a certificate signed by the CA specified in `--sslTrustCertFile`. For example, with curl:                                                                                                                             
                                                                                                                                                                                                                                                 
  curl --cert client.crt --key client.key --cacert server-ca.crt https://<proxy-host (https://<proxy-host)>:9201/                                                                                                                                
                                                                                                                                                                                                                                                 
  Existing deployments using only `--sslCertChainFile` + `--sslKeyFile` (one-way TLS) are unaffected — mTLS is opt-in.                                                                                                                           
                                                                                                                                                                                                                                                 
  ### Issues Resolved                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                 
  No Jira                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                 
  ### Testing                                                                                                                                                                                                                                    
                                                                                                                                                                                                                                                 
  Automated tests (unit tests in `FrontsideMtlsTest.java`). All existing Capture Proxy tests continue to pass.                                                                                                                                   
                                                                                                                                                                                                                                                 
  ### Check List                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                 

gregschohn (commented):
Thanks for this!  This is a great addition.

To get complete support for k8s users, we need to figure out how to get this value set.

You’ll need to add these flags to userSchemas.ts so that you can pass them into the workflow.  We don't thread sslTrustCertFilePath into the proxy in the workflow yet.  The complexity is that it's a file path and getting the file onto the prod

Options would be
1) to have the config specify a secret, which the proxy would mount as a file, then fill the argument in appropriately
2) To inline the contents in the config and then write it as a file (probably as a config map, then mount it as a file) - probably more work for the developer and the user.  The only advantage would be that you’d know which secret you were using and you wouldn’t have any extra work to do as a user - just keep moving the config bundle along & run it.
3) set up a simple OCI image that you publish to your registry & specify that instead of the configmap name.  This way, you make sure that you deploy your image & your config just says, mount this image.

I’m looking into adding support for custom transforms this week using #3 - we’ll have users make an OCI image of their transform files(s)/directories.  They’ll push it & then specify …ecr.aws…/transforms… & that will get mounted via an init container.  It would be easy to extend that volume approach for additional files like trust stores or public certs.  Users have the option to include or not include the digest, which makes the resources easier to track.

I’d like to support #1 too, but then we’d need some way to load the data - obviously, that isn’t too hard since kubernetes gives you this out of the box (kc create secret…).  It's just something else that would need to be managed by users/separate code, etc.

Let me know what you're thinking - if you want to piggyback off of #3, that will probably land shortly (spec'ing it out now)

Akhil911:
Thanks for the thorough feedback! Agreed that the K8s user experience needs this plumbing to make mTLS actually usable end-to-end.                                                                                                             
                                                                                                                                                                                                                                                 
  Option 3 sounds ideal — piggybacking on the OCI image approach you're building for transforms makes sense since trust stores and custom transforms have similar requirements (immutable files mounted into the proxy pod). It also gives users a clean, versionable way to track which cert is deployed, which is important for cert rotation audits.                                                                                                                                         
                                                                                                                                                                                                                                                 
  A few thoughts:                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                 
  1. Happy to wait for Option 3 - simple OCI image, to land and extend it for trust certs rather than add another mechanism now. Can you share the tracking issue/PR once it's ready? I'll rebase on top once it merges and add the userSchemas.ts wiring.                 
  2. Config shape — assuming something like this in the user config?                                                                                                                                                                             
     
```                                                                                                                                                                                                                                            
     proxy:                                                                                                                                                                                                                                      
       tls:                                                                                                                                                                                                                                      
         mode: "mtls"                                                                                                                                                                                                                            
         requireClientAuth: true                                                                                                                                                                                                                 
         trustCertSource:                                                                                                                                                                                                                        
           oci:                                                                                                                                                                                                                                  
             image: "123.dkr.ecr.us-east-1.amazonaws.com/migration-ca:v1"                                                                                                                                                                        
             mountPath: "/etc/proxy-tls/trust"
```                                                                                                                                                                                          
                                                                                                                                                                                                                                                 
     With the workflow injecting --sslTrustCertFile /etc/proxy-tls/trust/ca.pem --requireClientAuth based on that.                                                                                                                               
                                                                                                                                                                                                                                                 
  3. For this PR — should I keep it scoped to the Java-side mTLS flags (no workflow wiring), so it can merge independently? Then I can follow up with a second PR that adds the userSchemas.ts + workflow plumbing once Option 3 is in. That keeps the reviews focused and unblocks anyone using the proxy outside K8s (e.g., docker-compose, bare metal) who can set the CLI flags directly.                                                                                                         
                                                                                                                                                                                                                                                 
Let me know if that approach works or if you'd prefer to hold this PR until the full K8s integration is ready.

gregschohn (approved):
See my previous comment.  We'll need to address that for this feature to work in EKS, but that can be a follow-up.

PR #2826, opened 2026-04-28
Translating Stat Aggs to Solr Stat Facets Buckets

akshay2000:
### Description

Adds response-side handling for metric aggregations in `aggs-to-facets.ts`. OpenSearch metric agg results (`{ "value": N }`) are now converted to scalar values, matching Solr's json.facet response format where stat facets are plain numbers (e.g., `"avg_price": 29.95`).

Also adds `count(*)` support: the PEG grammar now accepts `*` as a function argument, and `count(*)` maps to `value_count` on `_id` so the named aggregation appears in the OpenSearch response.

### Testing
- Unit tests

### Check List

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/aggs-to-facets.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/aggs-to-facets.ts
@@ -96,11 +96,16 @@ function convertFilterAgg(aggResult: JavaMap): JavaMap {
  * Supports:
  *   - Terms / range / histogram aggregations (which have a `buckets` array)
  *   - Filter aggregations (from query facets, which have `doc_count`)
+ *   - Metric aggregations (avg, sum, min, max, value_count, cardinality — which have a `value` key)
  *
  * For both bucket-based and filter aggregations, any nested aggregation results
  * are recursively converted.
+ *
+ * Metric aggregations return their scalar value directly, matching Solr's
+ * json.facet response format where stat facets are plain numbers:
+ *   Solr: { "facets": { "count": 100, "avg_price": 29.95 } }
  */
-function convertSingleAgg(aggResult: any): JavaMap {
+function convertSingleAgg(aggResult: any): JavaMap | number {
   if (!isMapLike(aggResult)) {
     throw new Error(`Expected aggregation result to be a Map, got: ${typeof aggResult}`);
   }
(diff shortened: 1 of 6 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
```diff
@@ -466,6 +466,13 @@ function convertStatFacet(expr: string): JavaMap {
 }
 
 function convertFuncToMetricAgg(func: FuncNode): JavaMap {
+  // count(*) → value_count on _id (every doc has one, equivalent to doc count)
+  if (func.name === 'count') {
```
k-rooot, line 470: Should we add integ test coverage for this PR?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/aggs-to-facets.ts
```diff
@@ -96,11 +96,16 @@ function convertFilterAgg(aggResult: JavaMap): JavaMap {
...
  *
  * For both bucket-based and filter aggregations, any nested aggregation results
  * are recursively converted.
+ *
+ * Metric aggregations return their scalar value directly, matching Solr's
+ * json.facet response format where stat facets are plain numbers:
+ *   Solr: { "facets": { "count": 100, "avg_price": 29.95 } }
  */
-function convertSingleAgg(aggResult: any): JavaMap {
+function convertSingleAgg(aggResult: any): JavaMap | number {
```
k-rooot, line 108: I’m not a fan of returning multiple data types. Let’s try to avoid that pattern if possible. If there are edge cases or unsupported scenarios, it’s better to return a error instead.

akshay2000: Discriminated unions are common when we're dealing with dynamic data. The underlying Solr structure itself is multi typed — the aggregate can either be a value or another nested object like buckets.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/json-facets.ts
akshay2000: Added.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/aggs-to-facets.ts
k-rooot: I think this is actually a plain union rather than a discriminated union. We don’t have a common discriminator field here, so callers would still need to rely on runtime checks like typeof to figure out what they got.

Returning JavaMap | number creates an ambiguity to callers. I’d prefer we normalise this into a single shape (or use a proper discriminated union) so it’s easier and safer to consume.

I’m approving the PR for now. If you think this is worth addressing, please feel free to raise a new revision.

k-rooot (approved):

Issue #2778, opened 2026-04-22
Solr-to-OpenSearch Migration Functional Parity Tracker


sumobrian:
## Overview

Migration parity is not query coverage. A migration that translates 100% of Solr query syntax but returns different results — because the analyzers tokenize differently, the field types map incorrectly, or the latency profile changes application behavior — has not achieved parity. The customer's application still breaks.

We measure parity across three dimensions because that is what it takes to deliver **equivalent business outcomes**:

- **Query Behavior** — Can the application's existing queries run against OpenSearch and return the same results in the same order? This is necessary but not sufficient. A query that parses correctly but searches against differently-analyzed data returns wrong results silently.
- **Data Fidelity** — Is the migrated data complete, correctly typed, and equivalently tokenized? If Solr's `text_general` analyzer produces different tokens than OpenSearch's `text` analyzer for the same input, every query touching that field returns different results — regardless of whether the query translation is perfect.
- **Operational Equivalence** — Does OpenSearch serve the workload at comparable latency, throughput, and error rate? A migration that doubles P99 latency or introduces timeout errors under production load has not achieved parity from the customer's perspective, even if every query returns correct results in a test environment.

Query translation coverage alone creates a false sense of progress. A 95% query pass rate means nothing if the 5% that fail are the queries that drive the customer's core search experience, or if the data underneath has diverged. This tracker measures all three dimensions because customers don't migrate query parsers — they migrate business outcomes.

---

### Read Path — Query Parsing & Translation

| Feature | Solr Syntax | OpenSearch Equivalent | Status | PR |
|---------|------------|----------------------|--------|-----|
| Term Query | `field:value` | `term` query | ✅ Merged | #2553 |
| Phrase Query | `field:"exact phrase"` | `match_phrase` | ✅ Merged | #2554 |
| Bare Query | Unfielded `value` using `df` | `match` on default field | ✅ Merged | #2554 |
| Bool Query | `AND`, `OR`, `NOT`, `+`, `-` | `bool` must/should/must_not | ✅ Merged | #2564 |
| Boost Query | `field:value^2.0` | `boost` parameter | ✅ Merged | #2567 |
| Wildcard Query | `field:te?t`, `field:test*` | `wildcard` query | ✅ Merged | In fieldRule.ts |
| Fuzzy Query | `field:roam~`, `field:roam~2` | `fuzzy` query | ✅ Merged | In fieldRule.ts |
[...]

PR #2827, opened 2026-04-28
feat(solr-transforms): NDJSON framework and shared _bulk helpers

niravpi:
## Summary

   PR 1 of the Solr→OpenSearch write-request translation series. Pure infrastructure — lays the rails for `_bulk` without changing any customer-visible behavior. No request handler invokes the new helpers yet; that happens in subsequent PRs. The existing pipeline registry is untouched.

   This PR aligns the shim with the traffic replayer's NDJSON framework contract so that subsequent PRs can produce `_bulk`-style requests through the same code path the replayer already uses.

   ## Motivation

   Solr's `/update` endpoint supports batch operations (`{"add":[...]}`, `{"delete":[...]}`, `/update/json/docs` with top-level array bodies, mixed commands). The current shim rejects all of these with a fail-fast. To translate them into OpenSearch's `_bulk` API, the shim needs:

   1. A way to emit NDJSON from a transform (one JSON document per line + trailing newline).
   2. A shared builder that produces valid `_bulk` actions for `index` / `create` / `update` / `delete`.
   3. A response aggregator that maps `_bulk`'s per-item status/error output back into a Solr-shaped response.
   4. The ability to parse top-level JSON-array request bodies on ingress (needed by `/update/json/docs`).

   This PR adds all four without wiring any of them into the active pipeline. 

   ## Framework alignment with the traffic replayer

   The shim previously had no path for NDJSON bodies. The traffic replayer already supports NDJSON on both sides via `NettyJsonBodyAccumulateHandler` (ingress) and `NettyJsonBodySerializeHandler` (egress), using `JsonKeysForHttpMessage.INLINED_NDJSON_BODIES_DOCUMENT_KEY` (`inlinedJsonSequenceBodies`) as the canonical payload key.

   This PR brings the shim to the same contract:

   | Concern | Replayer | Shim (before) | Shim (after PR 1) |
   |---|---|---|---|
   | Canonical batch payload key | `inlinedJsonSequenceBodies` | not used | `inlinedJsonSequenceBodies` |
   | Emit NDJSON from a transform | ✅ streaming | ❌ silently dropped | ✅ synchronous materializer |
   | Top-level JSON array ingress | ✅ as `Object[]` under `inlinedJsonBody` | ❌ raw String | ✅ as `List<Object>` under `inlinedJsonBody` |
   | Trailing newline on emit | ✅ (when no sibling body) | n/a | ✅ always |
   | Mutual exclusion with `inlinedJsonBody` | producer responsibility | n/a | enforced by `setBulkRequest` |

   ## Changes

   ### Java — `HttpMessageUtil.java`
   - `parseJsonOrText`: parses top-level `[...]` bodies into `List<Object>` (replayer parity).
   - `extractBodyString`: new branch for `inlinedJsonSequenceBodies` → NDJSON with trailing newline.

   ### TypeScript — `features/bulk/bulk-builder.ts` (new)
   - `buildBulkAction(op, {index, id?, doc?})` — returns NDJSON entries for `index`/`create`/`update`/`delete`.
   - `setBulkRequest(ctx, collection, actions, {refresh})` — rewrites URI to `/_bulk`, sets POST + `application/x-ndjson`, writes `inlinedJsonSequenceBodies`, clears `inlinedJsonBody`.

   ### TypeScript — `features/bulk/bulk-response.ts` (new)
   - Aggregates `_bulk` `items[]` into Solr-shaped response: `status:0` on all-success, `status:1` + `errors[]` on any failure. Uses `took` as `QTime`. Emulates Solr's `TolerantUpdateProcessorFactory` format.

   ### TypeScript — `features/update-router.ts` (modified)
   - Response transform refactored into match-and-dispatch: single-doc (`_id`+`result`) vs bulk (`items[]`). Existing single-doc behavior preserved exactly.

   ### TypeScript — `request.transform.ts` (modified)
   - Skips overwriting `inlinedJsonBody` when `inlinedJsonSequenceBodies` is already set. Tolerates both Map and List body shapes.

   ### TypeScript — `context.ts` (docstring only)
   - `getBodyMap` docstring updated for polymorphic return. No behavioral change.

   ### Docs — `LIMITATIONS.md`
   - **`BULK-PARTIAL-FAILURE`**: OpenSearch `_bulk` has no transactional rollback; aggregator returns HTTP 200 + `errors[]` to prevent client retry duplicates.
   - **`NDJSON-INGEST-PARITY`**: NDJSON egress works; multi-line NDJSON ingress not yet supported (not exercised by Solr clients).

   ## Tests added

   | Suite | Count | Coverage |
   |---|---|---|
   | `HttpMessageUtilTest.java` | 11 | JSON-object ingress → Map, top-level-array ingress → List, scalar/malformed → raw String, NDJSON egress with trailing newline, NDJSON precedence over `inlinedJsonBody`, List `inlinedJsonBody` serializes as JSON array, empty/null payload. |
   | `bulk-builder.test.ts` | 23 | `index` (happy, no id, numeric id, missing doc, missing index) × `create` × `update` × `delete` (3 each); `setBulkRequest` (URI, refresh, content-type case-safety, payload invariant, body clearing, validation). |
   | `bulk-response.test.ts` | 13 | `isBulkResponse` predicate (match/reject cases); aggregator (all-success, any-failure, error-without-status, all-failure, empty items, missing `took` → QTime 0, strips all OS fields, omits absent optional fields, skips malformed items). |
   | `update-router.test.ts` | +7 | Dispatch: bulk match, single-doc match, reject select, bulk all-success (QTime=took), bulk partial-failure with errors[], mixed-op response (index+delete+update) with delete failing, single-doc QTime=0 distinct from bulk. |

   Total new: **+43 TS tests** (863 total, up from 820) and **+11 Java tests**.

   ## Validation

   | Gate | Result |
   |---|---|
   | TS `vitest` | **863 / 863** passing (36 files) |
   | Gradle `:TrafficCapture:transformationShim:test` | **301 / 301** passing |
   | Gradle `:TrafficCapture:SolrTransformations:test` | green |
   | E2E `:isolatedTest` (Solr 8 + Solr 9 + OpenSearch) | **no regressions** |

   ## What this PR does NOT change

   - **Request registry**: unchanged. The new `features/bulk/*` files are dead code at runtime until until subsequent  imports them.
   - **Existing `/update` behavior**: unchanged. The `update-router` still rejects arrays, multi-commands, `commit`, `optimize`, delete-by-query, and XML with fail-fast. Single-doc add, JSON-command-format add, and delete-by-id work exactly as before.
   - **Select / schema / admin endpoints**: unchanged.
   - **Response for existing `_doc` flows**: byte-identical (same match predicate, same body transformation — verified by the full existing `update-router.test.ts` suite continuing to pass unmodified).

   ## PR sequence

   1. **This PR** — NDJSON framework + `_bulk` plumbing
   2. PR 2 — Batch add + bulk delete + commit (depends on this)
   3. PR 3 — Delete by query ( Independent ) 

   After PR 1+2+3, three of the five remaining Solr write-path gaps in issue #2778 close in one cycle: bulk/batch operations, delete-by-query, and commit.

   ## Commit

   - `feat(solr-transforms): NDJSON framework and shared _bulk helpers` — `1dd9efb08`
   - 11 files changed, 1,448 insertions(+), 45 deletions(-)

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
@@ -159,7 +159,8 @@ export function setBulkRequest(
     targetParams.set('refresh', 'true');
   }
   const query = targetParams.toString();
-  ctx.msg.set('URI', `/${collection}/_bulk${query ? `?${query}` : ''}`);
+  const suffix = query ? '?' + query : '';
+  ctx.msg.set('URI', '/' + collection + '/_bulk' + suffix);
   ctx.msg.set('method', 'POST');
 
   // Replace Content-Type case-insensitively. The Java proxy uses headers.add()
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
@@ -1,0 +1,188 @@
+/**
+ * Shared helpers for constructing OpenSearch `_bulk` API requests from Solr
+ * command-level operations (index / create / update / delete).
+ *
+ * ## Framework alignment
+ *
+ * - NDJSON payloads are written under `payload.inlinedJsonSequenceBodies`, the
+ *   canonical key defined in `JsonKeysForHttpMessage.INLINED_NDJSON_BODIES_DOCUMENT_KEY`.
+ * - The Java side (`HttpMessageUtil.extractBodyString`) materializes the list
+ *   as one JSON document per line with a trailing newline — matching both the
+ *   traffic replayer's `NettyJsonBodySerializeHandler` and OpenSearch's `_bulk`
+ *   API contract.
+ * - `Content-Type: application/x-ndjson` is set by case-insensitively removing
+ *   any existing Content-Type header before writing — the shim's Java proxy
+ *   uses `headers.add()` rather than `set()`, so casing-variant duplicates can
+ *   otherwise leak through.
+ */
+import type { RequestContext, JavaMap } from '../../context';
+
+/** Supported OpenSearch `_bulk` operation types. */
+export type BulkOp = 'index' | 'create' | 'update' | 'delete';
+
+/** Document body type — a JavaMap (GraalVM polyglot Map) or a JS Map. */
+type DocBody = JavaMap | Map<string, unknown>;
+
+/** Arguments to {@link buildBulkAction}. */
+export interface BulkActionInput {
+  /** OpenSearch target index — typically derived from `ctx.collection`. */
+  index: string;
+  /** Document id. Required for `create`/`update`/`delete`; optional for `index`. */
+  id?: string | number;
+  /**
+   * Document body — a Map representing the JSON document.
+   *   - `index` / `create`: the full source document (required)
+   *   - `update`: an update payload, e.g. `Map{doc → Map{field → value}}` (required)
+   *   - `delete`: must be omitted
+   */
+  doc?: DocBody;
+}
+
+/**
+ * Build the NDJSON entries for a single `_bulk` operation.
+ *
+ * Returns an array that the caller flattens into the full NDJSON sequence.
+ * Each entry becomes one line in the NDJSON body after Java-side serialization.
+ */
+export function buildBulkAction(op: BulkOp, input: BulkActionInput): Map<string, unknown>[] {
+  if (!input.index) {
+    throw new Error(`[bulk-builder] ${op}: "index" is required`);
+  }
+  const meta = buildMeta(input);
+
+  switch (op) {
+    case 'delete':
+      return buildDelete(meta, input);
+    case 'update':
+      return buildUpdate(meta, input);
+    case 'create':
+      return buildCreate(meta, input);
+    case 'index':
+      return buildIndex(meta, input);
+    default: {
+      const exhaustive: never = op;
+      throw new Error(`[bulk-builder] unsupported op '${exhaustive as string}'`);
+    }
+  }
+}
+
+function buildMeta(input: BulkActionInput): Map<string, unknown> {
+  const meta = new Map<string, unknown>();
+  meta.set('_index', input.index);
+  if (input.id != null && input.id !== '') {
+    meta.set('_id', String(input.id));
+  }
+  return meta;
+}
+
+/** Output: `[{"delete": {"_index": "mycore", "_id": "42"}}]` — one NDJSON line, no doc body. */
+function buildDelete(meta: Map<string, unknown>, input: BulkActionInput): Map<string, unknown>[] {
+  if (!meta.has('_id')) {
+    throw new Error('[bulk-builder] delete: "id" is required');
+  }
+  if (input.doc !== undefined) {
+    throw new Error('[bulk-builder] delete: must not include "doc"');
+  }
+  return [new Map<string, unknown>([['delete', meta]])];
+}
+
+/** Output: `[{"update": {"_index": "mycore", "_id": "1"}}, {"doc": {"title": "new"}}]` — two NDJSON lines. */
+function buildUpdate(meta: Map<string, unknown>, input: BulkActionInput): Map<string, unknown>[] {
+  if (!meta.has('_id')) {
+    throw new Error('[bulk-builder] update: "id" is required');
+  }
+  if (input.doc == null) {
+    throw new Error('[bulk-builder] update: "doc" is required (e.g. { doc: {...} } or { script: ... })');
+  }
+  return [new Map<string, unknown>([['update', meta]]), input.doc as Map<string, unknown>];
+}
+
+/** Output: `[{"create": {"_index": "mycore", "_id": "1"}}, {"id": "1", "title": "hello"}]` — two NDJSON lines. */
+function buildCreate(meta: Map<string, unknown>, input: BulkActionInput): Map<string, unknown>[] {
+  if (!meta.has('_id')) {
+    throw new Error('[bulk-builder] create: "id" is required');
+  }
+  if (input.doc == null) {
+    throw new Error('[bulk-builder] create: "doc" is required');
+  }
+  return [new Map<string, unknown>([['create', meta]]), input.doc as Map<string, unknown>];
+}
+
+/** Output: `[{"index": {"_index": "mycore", "_id": "1"}}, {"id": "1", "title": "hello"}]` — two NDJSON lines. `_id` optional. */
+function buildIndex(meta: Map<string, unknown>, input: BulkActionInput): Map<string, unknown>[] {
+  if (input.doc == null) {
+    throw new Error('[bulk-builder] index: "doc" is required');
+  }
+  return [new Map<string, unknown>([['index', meta]]), input.doc as Map<string, unknown>];
+}
+
+/** Options accepted by {@link setBulkRequest}. */
+export interface SetBulkRequestOptions {
+  /**
+   * When true, appends `?refresh=true` to the `_bulk` URI — used for Solr
+   * `commit=true` / `commitWithin` semantics (immediate visibility).
+   */
+  refresh?: boolean;
+}
+
+/**
+ * Mutate a {@link RequestContext} to emit a `_bulk` request to OpenSearch.
+ *
+ * `actions` is the flattened concatenation of `buildBulkAction(...)` results.
+ *
+ * Example flow:
+ *   Solr input:  POST /solr/mycore/update  {"add":[{"doc":{"id":"1"}},{"doc":{"id":"2"}}]}
+ *   Caller does: actions = [...buildBulkAction('index', {index, id:'1', doc}),
+ *                            ...buildBulkAction('index', {index, id:'2', doc})]
+ *                setBulkRequest(ctx, 'mycore', actions, {refresh: true})
+ *   OS output:   POST /mycore/_bulk?refresh=true
+ *                {"index":{"_index":"mycore","_id":"1"}}
+ *                {"id":"1","title":"hello"}
+ *                {"index":{"_index":"mycore","_id":"2"}}
+ *                {"id":"2","title":"world"}
+ */
+export function setBulkRequest(
+  ctx: RequestContext,
+  collection: string,
+  actions: Map<string, unknown>[],
+  options: SetBulkRequestOptions = {},
+): void {
+  if (!collection) {
+    throw new Error('[bulk-builder] setBulkRequest: collection is required');
+  }
+  if (!Array.isArray(actions) || actions.length === 0) {
+    throw new Error('[bulk-builder] setBulkRequest: actions must be a non-empty array');
+  }
+
+  const targetParams = new URLSearchParams();
+  if (options.refresh) {
+    targetParams.set('refresh', 'true');
+  }
+  const query = targetParams.toString();
+  ctx.msg.set('URI', `/${collection}/_bulk${query ? `?${query}` : ''}`);
+  ctx.msg.set('method', 'POST');
+
+  // Replace Content-Type case-insensitively. The Java proxy uses headers.add()
+  // (not set()), so any surviving casing variant would create a duplicate.
+  const headers = ctx.msg.get('headers');
+  if (headers && typeof headers.keys === 'function') {
+    const toDelete = Array.from(headers.keys() as Iterable<string>)
+      .filter((k: string) => k.toLowerCase() === 'content-type');
+    for (const k of toDelete) {
+      headers.delete(k);
+    }
+    headers.set('content-type', 'application/x-ndjson');
+  }
+
+  if (!ctx.msg.get('payload')) ctx.msg.set('payload', new Map());
+  const payload = ctx.msg.get('payload');
+
+  // Enforce single-body invariant: exactly one of inlinedJsonBody /
+  // inlinedJsonSequenceBodies is set at a time (matches replayer contract).
+  payload.delete('inlinedJsonBody');
+  payload.set('inlinedJsonSequenceBodies', actions);
+
+  // Clear ctx.body so request.transform.ts's post-pipeline write does not
+  // re-populate inlinedJsonBody from an earlier state.
+  ctx.body = new Map();
+}
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
@@ -1,0 +1,187 @@
+/**
+ * Aggregate OpenSearch `_bulk` response into a Solr-shaped update response.
+ *
+ * ## Design decisions
+ *
+ * ### Partial-failure mapping (BULK-PARTIAL-FAILURE)
+ *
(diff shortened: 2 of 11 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
```diff
@@ -0,0 +1,194 @@
...
+  index: string;
+  /** Document id. Required for `create`/`update`/`delete`; optional for `index`. */
+  id?: string | number;
+  /**
+   * Document body:
+   *   - `index` / `create`: the full source document (required)
+   *   - `update`: an update payload (`{doc: {...}}`, `{script: {...}}`, etc.) — required
+   *   - `delete`: must be omitted
+   */
+  doc?: unknown;
```
akshay2000: Is it possible to use a stronger type here? Maybe a sum-type of simpler types? I understand that there will be some parts that are dynamic, but any type enforcement would help.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+function buildDelete(meta: Map<string, unknown>, input: BulkActionInput): Map<string, unknown>[] {
+  if (!meta.has('_id')) {
+    throw new Error('[bulk-builder] delete: "id" is required');
+  }
+  if (input.doc !== undefined) {
+    throw new Error('[bulk-builder] delete: must not include "doc"');
+  }
+  return [new Map<string, unknown>([['delete', meta]])];
```
akshay2000, line 86: I do not know what they were thinking when they chose this syntax for TS maps. It is very difficult to read as compared to say C#. (Or Python.) Can you add a doc string to each of these `build*` methods that shows how the expected output JSON object is supposed to look like? Is it:
```
[{"delete": {"_index": "reviews", "_id": 42}}]
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
```diff
@@ -0,0 +1,194 @@
...
+ *   - `Content-Type` replaced with `application/x-ndjson` (casing-safe)
+ *   - `payload.inlinedJsonSequenceBodies` set to `actions`
+ *   - `payload.inlinedJsonBody` cleared if present (single-body invariant)
+ *   - `ctx.body` cleared so `request.transform.ts` does not re-populate
+ *     `inlinedJsonBody` from stale state
+ */
+export function setBulkRequest(
+  ctx: RequestContext,
+  collection: string,
+  actions: Map<string, unknown>[],
```
akshay2000, line 147: How will this look like in the real world? Is it Solr's input body?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
```diff
@@ -0,0 +1,194 @@
...
+): void {
+  if (!collection) {
+    throw new Error('[bulk-builder] setBulkRequest: collection is required');
+  }
+  if (!Array.isArray(actions) || actions.length === 0) {
+    throw new Error('[bulk-builder] setBulkRequest: actions must be a non-empty array');
+  }
+
+  const suffix = options.refresh ? '?refresh=true' : '';
+  ctx.msg.set('URI', `/${collection}/_bulk${suffix}`);
```
akshay2000: Isn't there a more structured way to set URL params?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
```diff
@@ -0,0 +1,194 @@
...
+  ctx.msg.set('method', 'POST');
+
+  const headers = ctx.msg.get('headers');
+  if (headers) {
+    // The Java proxy uses headers.add() (not set()) — strip any casing variant
+    // before writing to avoid duplicate Content-Type headers on the wire.
+    headers.delete('Content-Type');
```
akshay2000: Are we sure that only these two variants will exist? Can `Content-type` not exist?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
```diff
@@ -0,0 +1,194 @@
...
+    headers.delete('content-type');
+    headers.set('content-type', 'application/x-ndjson');
+  }
+
+  let payload = ctx.msg.get('payload');
```
akshay2000: Nit: Why so many words?
```
if (!ctx.msg.get('payload')) ctx.msg.set('payload', new Map());
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
```diff
@@ -0,0 +1,187 @@
+/**
+ * Aggregate OpenSearch `_bulk` response into a Solr-shaped update response.
+ *
+ * No transform consumes this helper yet — it will be wired into
+ * `update-router.response` (PR 1, step 8) and exercised by PR 2 when real
```
akshay2000: It is strange to have PR/commit sequence comments in the code. This is point in time information. Please ensure that this is removed in subsequent PRs.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
```diff
@@ -0,0 +1,187 @@
...
+ * `_bulk` traffic starts flowing.
+ *
+ * ## Design decisions
+ *
+ * ### Partial-failure mapping (BULK-PARTIAL-FAILURE)
+ *
+ * Solr's update semantics are "batch is all-or-nothing from the client's
+ * perspective" — clients typically observe `status: 0` on success or a 4xx/5xx
+ * HTTP error on failure with a specific failing document in the message. Real
+ * partial-success behavior is not exposed to clients.
```
akshay2000: But does the engine (Solr) partially apply the successful deltas? Or the whole thing is treated as a transaction that is rolled back even for a single failure? If so, there is a big behavioural difference between the engines that goes way beyond response structures.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
```diff
@@ -0,0 +1,187 @@
...
+  type: string | undefined;
+  reason: string | undefined;
+}
+
+/**
+ * Match predicate for the `_bulk` response shape. OpenSearch always returns
+ * `{took, errors, items}` — `items` is the unambiguous marker.
+ */
+export function isBulkResponse(body: JavaMap): boolean {
+  return body != null && typeof body.has === 'function' && body.has('items');
```
akshay2000: Do we need an assertion on the type of `items` or its existence itself is sufficient?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
```diff
@@ -0,0 +1,187 @@
...
+  const detail = firstOpDetail(entry);
+  if (!detail) return null;
+  const status = toNumber(detail.get('status'));
+  const errorDetail = toMapLike(detail.get('error'));
+  const failed = (status != null && status >= 400) || errorDetail != null;
+  if (!failed) return null;
+  return {
+    index: toStringOrUndefined(detail.get('_index')),
+    id: toStringOrUndefined(detail.get('_id')),
+    status,
```
akshay2000: Mixing named and ordered?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
```diff
@@ -0,0 +1,187 @@
...
+}
+
+function toMapLike(v: unknown): JavaMap | null {
+  if (v != null && typeof (v as any).get === 'function' && typeof (v as any).keys === 'function') {
+    return v as JavaMap;
+  }
+  return null;
+}
+
+function toNumber(v: unknown): number | undefined {
```
akshay2000, line 112: At some point consider moving these `to*` and `is*` to utils.

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/netty/HttpMessageUtil.java
```diff
@@ -169,17 +178,53 @@ static String extractBodyString(Map<String, Object> map) {
...
      */
     private static Object parseJsonOrText(String body) {
-        if (body.isEmpty() || body.charAt(0) != '{') return body;
-        try {
-            return MAPPER.readValue(body, MAP_TYPE_REF);
-        } catch (JsonProcessingException ignored) {
-            return body;
+        if (body.isEmpty()) return body;
+        var first = body.charAt(0);
+        if (first == '{') {
```
akshay2000, line 214: If we select `TYPE_REF` in if-else (or a switch-case) then we can have a single try-catch and avoid repeated code.

akshay2000 (changes requested):

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
niravpi: Ack, changed to ```doc?: DocBody``` where ``` type DocBody = JavaMap | Map<string, unknown>```. This enforces that callers pass a Map (either GraalVM polyglot or JS native), which is what all consumers produce.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
niravpi: Ack, each buildDelete/buildUpdate/buildCreate/buildIndex helper now has an output example in its docstring, e.g.:
```
   /** Output: [{"delete": {"_index": "mycore", "_id": "42"}}] — one NDJSON line, no doc body. */ ```

File: TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts (hunk 3 above)
Comment (niravpi) [reply TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts:147]: actions is the output of buildBulkAction, not Solr's input body. The flow is:

   1. Solr client sends: POST /solr/mycore/update {"add":[{"doc":{"id":"1"}},{"doc":{"id":"2"}}]}
   2. The batch handler loops and calls buildBulkAction('index', {index, id, doc}) for each doc
   3. The flattened actions array is passed to setBulkRequest(ctx, collection, actions)
   4. OpenSearch receives: POST /mycore/_bulk with NDJSON body

   Added a docstring example showing this full flow.

File: TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts (hunk 4 above)
Comment (niravpi) [reply TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts]: Ack updated it, now uses URLSearchParams (same pattern as update-doc.ts and delete-doc.ts).

File: TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts (hunk 5 above)
Comment (niravpi) [reply TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts]: As per our testing till now these are the two variants we observed but to be on safe side updated it to iterate all header keys case-insensitively:
```
   const toDelete = Array.from(headers.keys())
     .filter(k => k.toLowerCase() === 'content-type');
   for (const k of toDelete) headers.delete(k);
```
   This handles any casing variant.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-builder.ts
niravpi: Ack, simplified to if (!ctx.msg.get('payload')) ctx.msg.set('payload', new Map()).

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
niravpi: Yes, side effects or AI agents, had seen this during code and was supposed to remove it but seems like missed this out. Updated in latest version.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
niravpi: Solr's default behavior is strict/transactional, if any document in a batch fails, the entire batch is rejected and no documents are applied. The partial state is rolled back. OpenSearch has no such rollback successfully-indexed documents remain even if later items fail. This is a fundamental architectural difference (Lucene segments don't support multi-document rollback).

Updated the inline comment to explicitly state this. The full discussion is in LIMITATIONS.md → BULK PARTIAL-FAILURE, which explains why we chose HTTP 200 + errors[] (emulating Solr's TolerantUpdateProcessorFactory) rather than HTTP 4xx (which would cause clients to retry the entire batch, creating duplicates of already-indexed docs).

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
niravpi: Ack, made status explicit: status: status instead of the shorthand.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
niravpi: Agree, will move to utils.ts when a second consumer appears. Currently these are private to bulk-response.ts and tightly coupled to the _bulk response shape. For now keeping it as it as no other file needs them yet.

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/netty/HttpMessageUtil.java
niravpi: Ack, now selects the TypeReference based on the first character, then has a single MAPPER.readValue(body, typeRef) call in one try-catch.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/bulk/bulk-response.ts
niravpi: It should be sufficient but to be on safe side now added checks both body.has('items') && body.has('took'). The _bulk response always has both fields, and requiring both makes the predicate unambiguous.

akshay2000 (approved):

PR #2829, opened 2026-04-28
Load transformations via a mounted image or ConfigMap

gregschohn:
### Description

Add mountable transforms for all MA workflow supported components.

Introduce a user-facing model for mountable transform sources. Users define named transform sources once with `transformsSources`, then reference them from metadata migration, RFS, and replayer configs with `transformsSource`. This keeps long image or ConfigMap references out of each tool config and makes clear that `transformsSource` is an alias, not a filesystem path.

Add ordered transform lists for each transform surface: `metadataTransforms`, `documentTransforms`, `requestTransforms`, and `tupleTransforms`. A single transform object is normalized to a one-item list so the common case stays compact, while ordered lists allow multiple transforms to run in sequence. Replayer request and tuple transforms stay separate because they use different runtime config fields and operate on different message shapes.

Translate the new transform lists into the existing Java transformer config format. The config processor maps each item to the appropriate JavaScript or Python provider, expands user-relative script paths into full `/transforms/...` paths, applies default script names by transform kind and language, and writes the generated JSON into the existing transformer config fields.

Mount transform sources into transform-capable pods at `/transforms`. Workflow templates resolve either an OCI image volume or a ConfigMap volume, then pass only the resolved mount source through workflow-level options. Metadata migration uses the same conditional volume model as the deployment-based steps.

Make transform source and transform config changes visible to workflow sensitivity checks. The resolved transform source fields are carried into the migration CRDs so changes to transform images, ConfigMaps, or generated raw transform config make the managed resources dirty. Metadata and RFS transform changes remain impossible after creation, while replayer transform changes use the existing gated-change path.

Add packaging support for OCI image-backed transform volumes. The scratch Dockerfile and `package-transforms.sh` script package a transform directory, push it to a registry, and print a digest-pinned `transformsSources` snippet. Digest-pinned references make transform content changes explicit in workflow configuration.

Relax the Java transformer providers to match the new schema behavior. `bindingsObject` is optional, object-valued bindings are accepted directly, and the legacy JSON-string form still works. Script resolution preserves the source file path so Python transforms loaded from mounted files can import sibling helper modules through their script directory.

Make generated schemas compatible with arbitrary transform bindings. The new `bindingsObject` field accepts any JSON object, which exposes a zod-to-openapi edge case where bare `z.any()` nodes can become `nullable: true` without an associated schema type. Normalize those nodes into an explicit JSON-value type union so the generated workflow schema remains valid for AJV-based validation.

Update schema, config-processor, workflow-template, and provider tests to cover source alias validation, relative script paths, ordered transform translation, legacy raw config conflicts, optional/object bindings, Python sibling imports, and rendered workflow output.

Example using two document transforms from one transform image, run in order:

```yaml
transformsSources:
  my-transforms:
    image: "123456789012.dkr.ecr.us-east-1.amazonaws.com/migration-transforms@sha256:abc123..."
snapshotMigrationConfigs:
  - fromSource: "source"
    toTarget: "target"
    perSnapshotConfig:
      snap1:
        - metadataMigrationConfig: {}
          documentBackfillConfig:
            transformsSource: my-transforms
            documentTransforms:
              - language: javascript
                file: normalize-document.js
                bindingsObject:
                  sourceVersion: "7.10"
              - language: python
                file: add-marker.py
                bindingsObject:
                  markerField: "migrated_by_transform"
sourceClusters:
  source: ...
targetClusters:
  target: ...
```

The image root should contain:

```text
/normalize-document.js
/add-marker.py
```

Those user-relative paths become `/transforms/normalize-document.js` and `/transforms/add-marker.py` in the generated Java config. The list order is the execution order.

### Testing
See the [testing design file](https://github.com/gregschohn/opensearch-migrations/blob/3d8e74511e23d4376bf51a1d201a952de0263caa/docs/MountableTransformsE2ETestProposal.md) for what was added.  I tested that against a kind cluster on my local laptop.

### Check List


docs/MountableTransformsDesign.md
```diff
@@ -0,0 +1,368 @@
+# Mountable Transforms — Design Document
+
+## Overview
+
+Users define transform sources once in a top-level `transformsSources` map (aliased by name), then reference them by key in per-tool options. The workflow mounts `/transforms/` into every transform-eligible pod (MetadataMigration, RFS, Replayer). Each program has a default entrypoint filename it looks for, overridable by the user.
```
sumobrian: Are tuple transforms in or out of scope?

docs/MountableTransformsDesign.md
```diff
@@ -0,0 +1,368 @@
...
+
+Two sources are supported:
+
+| Source | Volume type | Version identity |
+|---|---|---|
+| `image` | K8s native `image` volume (1.35+ enabled by default, GA in 1.36) | Image digest in the reference |
+| `configMap` | K8s ConfigMap volume | ConfigMap name (user-managed) |
```
sumobrian, line 12: I'd start with this and accept 1MB limit.

docs/MountableTransformsDesign.md
```diff
@@ -0,0 +1,368 @@
...
+    .changeRestriction('impossible'),
+
+transformsEntrypoint: z.string().optional()
+    .describe("Override the default entrypoint filename within /transforms/. " +
+        "Defaults: 'metadata-transforms.json' (metadata), 'doc-transforms.json' (RFS), " +
+        "'replayer-transforms.json' (replayer).")
+    .changeRestriction('impossible'),
+```
+
+These flow automatically through `ARGO_*_OPTIONS` via `makeOptionalDefaultedFieldsRequired()`, and into `---INLINE-JSON` via `expr.omit(options, ...WORKFLOW_OPTION_KEYS)` — no manual parameter threading needed.
```
sumobrian: Doesn't JSON parser rejects unknown keys, so those should stay workflow-only fields and be resolved into the existing CLI fields before the CLI boundary.

docs/MountableTransformsDesign.md
gregschohn: I'm less concerned about the 1MB limit and more worried about the UX.  If I'm a user, I can update my OCI registry w/ latest or a tag value (either fast easy or measured and tight).  I can do that before my k8s cluster is up.  Even if the cluster gets reconfigured, I'm still good.  The story on how to configure the configmap is slightly more involved and it's harder to support the tight story if the user has access to change that configmap.  We also might run into more issues with concurrency issues.  User just reset transforms for replayer. but wants to stick with RFS transforms, which are currently running.  Having separate mounts on each container from a static resource is a lot easier to reason with for us and users.

docs/MountableTransformsDesign.md
gregschohn: Tuple transforms are well-within scope.  My thought is that we allow users to set up mountable FS & from there, the configs reach into the parts that they need as entry-points.

docs/MountableTransformsDesign.md
gregschohn: Yes, I may have adjustments to do here

docs/MountableTransformsDesign.md
AndreKurait: My pitch is the opposite, start with just the OCI Artifact which will handle the permissions story for this

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
```diff
@@ -204,7 +206,8 @@ public static class Parameters {
         String inputFilename;
         @Parameter(
             required = false,
-            names = {"-t", PACKET_TIMEOUT_SECONDS_PARAMETER_NAME, "--packetTimeoutSeconds" },
+            names = {"-t", PACKET_TIMEOUT_SECONDS_PARAMETER_NAME, "--packetTimeoutSeconds",
```
AndreKurait, line 209: Why are we adding new parameters for all of these?

orchestrationSpecs/packages/argo-workflow-builders/src/models/containerBuilder.ts
```diff
@@ -232,12 +232,15 @@ export class ContainerBuilder<
...
-            name,
-            configMap: config.configMap
-        }));
+        const volumes = Object.entries(this.volumeScope).map(([name, config]) => {
+            const {mountPath, readOnly, volume, ...volumeSource} = config as Record<string, PlainObject>;
+            return volume ?? {
+                name,
+                ...volumeSource
+            };
+        });
```
AndreKurait, line 241: The `...volumeSource` spread silently passes any extra key on `volumeScope[name]` through to the rendered `Volume`. A typo in a caller (e.g. `configmap` instead of `configMap`) becomes a malformed manifest at apply-time rather than a build-time type error. The `volume ??` escape hatch also has no documented use case I could find. Pick supported sources explicitly:

```suggestion
    getBody() {
        const volumes = Object.entries(this.volumeScope).map(([name, config]) => {
            const c = config as Record<string, unknown>;
            const source: Record<string, unknown> = {};
            if (c.configMap !== undefined) source.configMap = c.configMap;
            if (c.secret    !== undefined) source.secret    = c.secret;
            if (c.image     !== undefined) source.image     = c.image;
            if (c.emptyDir  !== undefined) source.emptyDir  = c.emptyDir;
            return { name, ...source };
        });
```

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -157,6 +157,89 @@ export const OPTIONAL_HTTP_ENDPOINT_PATTERN = `^(?:https?:\\/\\/${HOSTNAME_PATTE
...
+export const TRANSFORMS_IMAGE_SOURCE = z.object({
+    image: z.string()
+        .describe("OCI image reference (preferably with digest) whose root contains transform files. " +
+            "The workflow mounts the image at /transforms/. Build with deployment/k8s/package-transforms.sh.")
+});
+
+export const TRANSFORMS_CONFIGMAP_SOURCE = z.object({
+    configMap: z.string()
+        .describe("Name of a pre-existing Kubernetes ConfigMap. Each key becomes a file in /transforms/.")
+});
```
AndreKurait: Two tightenings on the source-of-truth schemas: require non-empty values, and `.strict()` the objects. Without `.strict()`, a config with both `image` and `configMap` set silently selects the first arm of the union; with `.strict()` it fails validation at the schema layer with a clear error.

```suggestion
export const TRANSFORMS_IMAGE_SOURCE = z.object({
    image: z.string().min(1)
        .describe("OCI image reference (preferably with digest) whose root contains transform files. " +
            "The workflow mounts the image at /transforms/. Build with deployment/k8s/package-transforms.sh.")
}).strict();

export const TRANSFORMS_CONFIGMAP_SOURCE = z.object({
    configMap: z.string().min(1)
        .describe("Name of a pre-existing Kubernetes ConfigMap. Each key becomes a file in /transforms/.")
}).strict();
```

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -461,6 +546,9 @@ export const USER_REPLAYER_WORKFLOW_OPTIONS = z.object({
 }).describe("Kubernetes deployment-level options for the traffic replayer.");
 
 export const USER_REPLAYER_PROCESS_OPTIONS = z.object({
+    transformsSource: z.string().optional()
+        .describe("Key into top-level transformsSources. Mounts /transforms/ into the replayer pod.")
+        .changeRestriction('impossible'),
```
AndreKurait: The map keys in `TRANSFORMS_SOURCES_MAP` are constrained to `K8S_NAMING_PATTERN` (line 178), but the per-tool reference here is a free-form string. A typo like `"My-Source!"` slips past schema validation and only surfaces at the cross-field reference check with a worse error. Apply the same regex:

```suggestion
    transformsSource: z.string().regex(K8S_NAMING_PATTERN).optional()
        .describe("Key into top-level transformsSources. Mounts /transforms/ into the replayer pod.")
        .changeRestriction('impossible'),
```

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -589,6 +717,10 @@ export const USER_METADATA_WORKFLOW_OPTIONS = z.object({
 }).describe("Workflow-level options for metadata migration, controlling JVM settings and approval gates.");
 
 export const USER_METADATA_PROCESS_OPTIONS = z.object({
+    transformsSource: z.string().optional()
+        .describe("Key into top-level transformsSources. Mounts /transforms/ into the metadata migration pod.")
+        .checksumFor('snapshot', 'replayer')
+        .changeRestriction('impossible'),
```
AndreKurait: Same as the replayer reference — pin the name to `K8S_NAMING_PATTERN` so a typo fails at schema validation rather than at the cross-field reference check.

```suggestion
    transformsSource: z.string().regex(K8S_NAMING_PATTERN).optional()
        .describe("Key into top-level transformsSources. Mounts /transforms/ into the metadata migration pod.")
        .checksumFor('snapshot', 'replayer')
        .changeRestriction('impossible'),
```

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -671,6 +814,10 @@ export const USER_RFS_WORKFLOW_OPTIONS = z.object({
 }).describe("Kubernetes deployment-level options for the Reindex From Snapshot (RFS) document backfill.");
 
 export const USER_RFS_PROCESS_OPTIONS = z.object({
+    transformsSource: z.string().optional()
+        .describe("Key into top-level transformsSources. Mounts /transforms/ into the RFS pod.")
+        .checksumFor('replayer')
+        .changeRestriction('impossible'),
```
AndreKurait: Same fix on the RFS reference for consistency with the map keys.

```suggestion
    transformsSource: z.string().regex(K8S_NAMING_PATTERN).optional()
        .describe("Key into top-level transformsSources. Mounts /transforms/ into the RFS pod.")
        .checksumFor('replayer')
        .changeRestriction('impossible'),
```

AndreKurait (commented):
Review feedback focused on the orchestration / schema layer of this PR. Comments 1–7 each have a concrete inline suggestion; 8–14 are structural observations where the fix spans more lines than a `suggestion` block can carry.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -453,6 +536,8 @@ export const USER_REPLAYER_WORKFLOW_OPTIONS = z.object({
         .describe(LOGGING_CONFIG_OVERRIDE_DESC),
     podReplicas: z.number().default(1).optional()
         .describe("Number of replayer pod replicas in the Kubernetes Deployment. Each replica independently consumes from Kafka and replays traffic to the target."),
+    useLocalStack: z.boolean().default(false).optional()
+        .describe("Mount local test AWS credentials for LocalStack-backed tuple S3 output. This is a workflow-only testing option and is not passed to the replayer process."),
```
AndreKurait: This is a workflow-only test hook, but it appears as a top-level user-facing field in the public JSON Schema with no marker. Tag it the same way other expert/internal fields are flagged so external users don't discover and start setting it:

```suggestion
    useLocalStack: z.boolean().default(false).optional()
        .describe("[Internal] Mount local test AWS credentials for LocalStack-backed tuple S3 output. Workflow-only testing hook; not passed to the replayer process and not intended for production use."),
```

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -12,6 +12,124 @@ export type ContainerVolumePair = {
...
+    transformsImage: BaseExpression<string>,
+    transformsConfigMap: BaseExpression<string>
+) {
+    const hasImage = expr.not(expr.isEmpty(transformsImage));
+    const hasConfigMap = expr.not(expr.isEmpty(transformsConfigMap));
+    const imageVolume = expr.makeDict({
+        name: TRANSFORMS_VOLUME_NAME,
+        image: expr.makeDict({
+            reference: transformsImage,
+            pullPolicy: expr.literal("IfNotPresent"),
```
AndreKurait: Hardcoded pull policy means a user iterating on a mutable tag during dev can't override (e.g. force `Always`). With digest-pinned refs `IfNotPresent` is fine, but plumbing this through is one line here once `pullPolicy` is added to `TRANSFORMS_IMAGE_SOURCE`. Mark it explicitly for now:

```suggestion
            pullPolicy: expr.literal("IfNotPresent"), // TODO: thread through TRANSFORMS_IMAGE_SOURCE.pullPolicy
```

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts
```diff
@@ -275,25 +310,81 @@ export const Replayer = WorkflowBuilder.create({
 
   .addParams(CommonWorkflowParameters)
 
-  .addTemplate("createDeployment", (t) =>
+  .addTemplate("createDeploymentWithImageTransforms", (t) =>
```
AndreKurait: **Collapse the three `createDeployment*Transforms` templates into one.**

Lines 313–420 define three near-identical templates that differ only by `transformsVolumeMode: "image" | "configMap" | "emptyDir"`, plus a 4th `createDeployment` dispatcher (421–443) that routes to them via `when` clauses. The bodies are otherwise an exact copy of ~30 lines of `getReplayerDeploymentManifest({...})`.

Two paths I'd try in order:

1. **First, try the runtime-ternary path everywhere.** `commonUtils/containerFragments.ts:19 makeTransformsVolume` already builds the image/configMap/emptyDir choice as an Argo expression ternary, and the metadata template uses something like it. If that expression renders into valid manifest YAML when interpolated, you don't need the per-mode templates at all — collapse to one `createDeployment` template, drop ~100 lines here and ~250 in snapshots, and remove `transformsVolumeMode` and `setupTransformsForContainerForMode` entirely. The fact that this PR has *both* `makeTransformsVolume` and `setupTransformsForContainerForMode` suggests this was attempted and hit a rendering issue — worth documenting what went wrong in the PR description, because if it works it's strictly better.

2. **If the runtime ternary doesn't render**, build and publish a tiny scratch image as part of the migration release (you already have `Dockerfile.transforms` and `package-transforms.sh`). The orchestration always emits `image: { reference: <transformsImage>, pullPolicy: ... }` on the volume, and the config-processor substitutes the empty-image digest when no user transforms are configured. Pod spec is then structurally identical in every case, you keep one template, and you don't take an external dependency on a public registry. Air-gapped clusters mirror the empty image alongside the other migration images.

Either approach replaces the dispatcher pattern. Whichever lands here should be applied to `documentBulkLoad.ts` (lines 367–486) and `metadataMigration.ts` (lines 205–453) too — they have the same three-template structure.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/metadataMigration.ts
```diff
@@ -181,29 +202,13 @@ export const MetadataMigration = WorkflowBuilder.create({
...
-        .addRequiredInput("sourceK8sLabel", typeToken<string>())
-        .addRequiredInput("targetK8sLabel", typeToken<string>())
-        .addRequiredInput("snapshotK8sLabel", typeToken<string>())
-        .addRequiredInput("fromSnapshotMigrationK8sLabel", typeToken<string>())
-        .addRequiredInput("crdName", typeToken<string>())
-        .addRequiredInput("crdUid", typeToken<string>())
-        .addRequiredInput("resourceCreationTimestamp", typeToken<string>())
-        .addOptionalInput("workflowCreationTimestamp", c => expr.getWorkflowValue("creationTimestamp"))
-        .addOptionalInput("workflowUid", c => expr.getWorkflowValue("uid"))
+    .addTemplate("runMetadataNoTransforms", t => t
```
AndreKurait: **The three metadata templates duplicate a ~60-line container body.**

`runMetadataNoTransforms` (205), `runMetadataWithImageTransforms` (267), and `runMetadataWithConfigMapTransforms` (336) repeat an identical `addContainer` block (env vars, command/args, artifact output, pod metadata). The only differences are:

- whether `'user-transforms'` is added to `addVolumesFromRecord`
- whether that entry resolves to an `image:` or `configMap:` source

Even if you keep the three-template structure (rather than collapsing per the comment on `replayer.ts:313`), factor out the shared body so the diff *between* templates is visible. Something like:

```ts
function withMetadataContainer<B>(b: B, extraVolumes: Record<string, VolumeEntry> = {}) {
    return b.addContainer(c => c
        .addImageInfo(...)
        .addVolumesFromRecord({ 'test-creds': { ... }, ...extraVolumes })
        .addEnvVar(...) // shared
        // ...
    );
}
```

Then each template body is one call. Today, a future reviewer changing (say) the `addArtifactOutput` s3Key has to remember to update three places.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
```diff
@@ -12,6 +12,124 @@ export type ContainerVolumePair = {
...
+                    name: TRANSFORMS_VOLUME_NAME,
+                    mountPath: TRANSFORMS_MOUNT_PATH,
+                    readOnly: true
+                }
+            ]
+        }
+    } as const;
+}
+
+function makeTransformsVolumeForMode(
```
AndreKurait, line 70: **Two transform-volume helpers for two callers — pick one.**

This file ships both `makeTransformsVolume` + `setupTransformsForContainer` (lines 19–76, expression-ternary form, used by the metadata template) and `makeTransformsVolumeForMode` + `setupTransformsForContainerForMode` (lines 78–131, mode-string form, used by replayer + RFS). They produce the same Kubernetes volume; the only difference is *when* the dispatch happens.

Leaving both in place means the next person adding a transforms-aware tool has to figure out which to call. If the runtime-ternary form works (see comment on `replayer.ts:313`), this whole `*ForMode` half goes away. If not, pick one and use it everywhere — the mode-string form is the cleaner of the two because the branch happens at TS-compile time rather than inside an Argo expression that has to be evaluated at workflow runtime.

orchestrationSpecs/packages/schemas/src/argoSchemas.ts
```diff
@@ -133,11 +133,21 @@ export const DYNAMIC_SNAPSHOT_CONFIG =
     }));
 
 export const ARGO_METADATA_OPTIONS = makeOptionalDefaultedFieldsRequired(
```
AndreKurait, line 141: **Triplicated `transformsImage` / `transformsConfigMap` extension blocks.**

`ARGO_METADATA_OPTIONS` (135), `ARGO_RFS_OPTIONS` (161), and `ARGO_REPLAYER_OPTIONS` (192) each do the same `omit({transformsSource, ...pipeline-fields}).extend({ transformsImage, transformsConfigMap })`, plus a matching `*_WORKFLOW_OPTION_KEYS` with `transformsImage: true, transformsConfigMap: true` (lines 149-150, 177-178, 208-209). Three copies of the same shape.

Pull this into a helper:

```ts
const TRANSFORMS_RESOLVED_FIELDS = {
    transformsImage: z.string().default("").optional(),
    transformsConfigMap: z.string().default("").optional(),
};

function resolveTransformsFields<T extends z.ZodObject<any>>(
    schema: T,
    pipelineKeys: Partial<Record<keyof T['shape'], true>>,
) {
    return schema.omit({ transformsSource: true, ...pipelineKeys })
                 .extend(TRANSFORMS_RESOLVED_FIELDS);
}
```

Then `ARGO_METADATA_OPTIONS = makeOptionalDefaultedFieldsRequired(resolveTransformsFields(USER_METADATA_OPTIONS.omit({skipEvaluateApproval, skipMigrateApproval}), {metadataTransforms: true}))`. New fields (e.g. `transformsPullPolicy`) get added once.

orchestrationSpecs/packages/schemas/src/argoSchemas.ts
```diff
@@ -133,11 +133,21 @@ export const DYNAMIC_SNAPSHOT_CONFIG =
...
-    USER_METADATA_OPTIONS.omit({skipEvaluateApproval: true, skipMigrateApproval: true})
+    USER_METADATA_OPTIONS.omit({
+        skipEvaluateApproval: true,
+        skipMigrateApproval: true,
+        transformsSource: true,
+        metadataTransforms: true,
+    }).extend({
+        transformsImage: z.string().default("").optional(),
```
AndreKurait: **`""` as the "absent" sentinel for `transformsImage` / `transformsConfigMap` is fragile.**

Encoding "no transforms configured" as the empty string forces every consumer downstream to test `expr.not(expr.isEmpty(...))`. That pattern is repeated 6+ times across this PR (e.g. `metadataMigration.ts:416`, `replayer.ts:425`, `documentBulkLoad.ts:469`, `containerFragments.ts:23`).

Two cleaner options that don't break wire format:

- Promote these to a discriminated union at the argo layer too: `transformsSource: { kind: 'image', reference } | { kind: 'configMap', name } | { kind: 'none' }`. The dispatcher branches on `kind`, not on string emptiness, and a future `kind: 'secret'` is a one-line addition rather than a third parallel field.

- If keeping flat fields, at minimum add a single `hasTransformsConfigured(opts)` helper exported from this file and use it everywhere instead of hand-rolling `not(isEmpty(...))` at each call site.

The current encoding also makes it impossible to distinguish "user explicitly set it to empty" from "user didn't set it" — not a problem today, but worth noting.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/documentBulkLoad.ts
```diff
@@ -377,6 +397,93 @@ export const DocumentBulkLoad = WorkflowBuilder.create({
...
+                    sourceK8sLabel: b.inputs.sourceK8sLabel,
+                    targetK8sLabel: b.inputs.targetK8sLabel,
+                    snapshotK8sLabel: b.inputs.snapshotK8sLabel,
+                    fromSnapshotMigrationK8sLabel: b.inputs.fromSnapshotMigrationK8sLabel,
+                    taskK8sLabel: b.inputs.taskK8sLabel,
+                })
+            }))
+        .addRetryParameters(K8S_RESOURCE_RETRY_STRATEGY)
+    )
+    .addTemplate("startHistoricalBackfill", t => t
```
AndreKurait, line 432: **The `withoutTransforms` step's `when` clause is provably redundant.**

Lines 472–484 dispatch with these conditions:

- `withImageTransforms`: `hasImage`
- `withConfigMapTransforms`: `!hasImage && hasConfigMap`
- `withoutTransforms`: `!hasImage && !hasConfigMap`

The third predicate is the negation of the union of the first two — it's definitionally true exactly when neither sibling matched. Since Argo evaluates step `when` clauses sequentially within a step group, you don't need the predicate at all for the fallback branch.

Cleaner: drop the explicit `when` from `withoutTransforms`, or restructure as a 2-arm conditional (image vs. configMap-or-empty) once one of the sentinel/runtime-ternary approaches lands.

Same redundant `withoutTransforms` predicate exists in `replayer.ts:437` and `metadataMigration.ts:443`.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/documentBulkLoad.ts
```diff
@@ -330,25 +364,8 @@ export const DocumentBulkLoad = WorkflowBuilder.create({
     )
```
AndreKurait, line 370: **Each per-mode template re-passes ~17 fields verbatim into `getRfsDeploymentManifest`.**

The 17-field manifest call (lines 374–395) is repeated 3× with only `transformsVolumeMode` differing. If the dispatcher pattern stays after addressing the comment on `replayer.ts:313`, extract a helper:

```ts
function buildBackfillManifestArgs(b: BuilderInputs, mode: TransformVolumeMode) {
    return {
        podReplicas: expr.deserializeRecord(b.inputs.podReplicas),
        loggingConfigMap: b.inputs.loggingConfigurationOverrideConfigMap,
        // ...all 17 fields...
        transformsVolumeMode: mode,
    };
}
```

Then each per-mode template body is `manifest: getRfsDeploymentManifest(buildBackfillManifestArgs(b, "image"))`. The diff between modes becomes literally the mode string. Reduces this file by ~60 lines and makes a future "I forgot to update one of the three when adding a new field" bug structurally impossible. Same mechanical extraction applies in `replayer.ts` and `metadataMigration.ts`.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
gregschohn: Does the commit answer your question?
https://github.com/opensearch-project/opensearch-migrations/pull/2829/changes/81b6a03b92527dbd1f7147b12f816252aae95422

AndreKurait: Yes, though i would prefer to clean these up to have fewer permutations even if that meant more changes in the workflow, unrelated to this PR though

gregschohn: Agree, but it's a breaking change, so I'd want to do that break in a smaller more targeted PR.

orchestrationSpecs/packages/argo-workflow-builders/src/models/containerBuilder.ts
gregschohn: The type safety is undone by the `const source: Record<string, unknown>` - it's a good idea, but only works if the type is built up directly.   Regardless, maintaining the type doesn't do us much good anyway unless it's advertised elsewhere that requires it?

orchestrationSpecs/packages/schemas/src/argoSchemas.ts
gregschohn: The union sounds like a bad idea for Argo - it's a more complicated type to unroll.  I'm adding a helper though.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/documentBulkLoad.ts
gregschohn: Argo does **not evaluate when clauses sequentially with a fall-through**.  We have many other cases of this already in the project.  We need all 3 sets of whens.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: done

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: done

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: done

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: done

orchestrationSpecs/packages/schemas/src/userSchemas.ts
gregschohn: done

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/documentBulkLoad.ts
AndreKurait: I'm fine without this, it's not user impacting either way

orchestrationSpecs/packages/schemas/src/argoSchemas.ts
AndreKurait: Thanks for adding the helper! I'm good with that approach

orchestrationSpecs/packages/argo-workflow-builders/src/models/containerBuilder.ts
AndreKurait: makes sense, i'm good closing this one out as it's a much larger refactor to keep the types all the way through

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
gregschohn: done - one was dead code

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts
gregschohn: TODO - I've not been happy w/ that triplication.  After the other changes settle, I'll try to figure it out.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
gregschohn: done - setupTransformsForContainer was dead code

orchestrationSpecs/packages/schemas/src/argoSchemas.ts
gregschohn: The function doesn't really work because it isn't strongly typed.  I did put the two fields into TRANSFORMS_RESOLVED_FIELDS and use zod's extend().

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/containerFragments.ts
gregschohn: I've made this default to IfNotPresent, but its an enumerated field defined alongside the transform source.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/metadataMigration.ts
gregschohn: same as the previous comment.  TODO

orchestrationSpecs/packages/schemas/src/getSchemaFromZod.ts
```diff
@@ -25,6 +51,27 @@
...
+        return;
+    }
+
+    if (jsonSchema.nullable === true && jsonSchema.type === undefined
+        && jsonSchema.$ref === undefined
+        && jsonSchema.oneOf === undefined
+        && jsonSchema.anyOf === undefined
+        && jsonSchema.allOf === undefined) {
+        jsonSchema.type = ["string", "number", "boolean", "object", "array", "null"];
+        delete jsonSchema.nullable;
```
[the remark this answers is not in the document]
AndreKurait: @gregschohn can we fix this?

AndreKurait (approved):

orchestrationSpecs/packages/schemas/src/getSchemaFromZod.ts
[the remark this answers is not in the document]
gregschohn: done

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts
gregschohn: We still have 3 templates, but I've collapsed the code paths

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/metadataMigration.ts
gregschohn: We still have 3 templates, but I've collapsed the code paths

Issue #2801, opened 2026-04-26
[BUG] [workflow CLI] "workflow output <workflow-name>" fails with "container not valid"


pkan010124:
## Summary
`workflow output <workflow-name>` returns errors for every pod in the workflow, preventing aggregated log retrieval.

## Environment
- MA version: 3.0.1
- Deployment: EKS (import-vpc)

## Current behavior
$ workflow output migration-workflow [migration-workflow-patchdatasnapshotready-2687195055] Error: error: container migration-workflow is not valid for pod ... [migration-workflow-readdatasnapshotname-263454980] Error: error: container migration-workflow is not valid for pod ... [all 8 step pods return the same error]

## Expected behavior
Aggregated logs across all step pods, similar to `kubectl logs --all-containers=true` behavior.

## Workaround
Manual: `kubectl logs -n ma <pod-name> --all-containers=true` for each step pod.

## Root cause hypothesis
Likely hardcoded container name lookup (`migration-workflow`) instead of using the Argo-generated container names (`main`, `wait`).

gregschohn:
To view all the logs for the workflow just `workflow output` should do it.  You can also select certain tasks or phases with label selectors - `workflow output -l task=snapshot` or `workflow output -l task=snapshot,source=logs`.

I'm looking into this now.  It's easy to reproduce.

jugal-chauhan:
This issue has now been addressed and resolved. The workflow output command was replaced by workflow log in PR #2834 (the root cause mentioned no longer exists). The new implementation uses k8s label selectors to find pods. The `workflow log` CLI provides subcommands like : `workflow log all`, `workflow log resource <name>`, `workflow log filter --task=<task_name>`. 

This will be available in the next release after 3.1.1

jugal-chauhan:
@pkan010124  This is available in release 3.2.0. The workflow output command has been replaced by two purpose-built commands:

- `workflow log` : pod-level logs with subcommands

```
(18:12:58) migration-console (~) -> workflow log --help
Usage: workflow log [OPTIONS] COMMAND [ARGS]...

  View or tail workflow logs.

  Subcommands:
    all       Show all logs for the selected workflow.
    resource  Show logs for one migration resource.
    filter    Show logs matching one or more label filters.

  Arguments after -- are forwarded to the underlying tool
  (kubectl logs for history, stern for follow mode).
  Use -- --help to see what the underlying tool supports.

Options:
  -h, --help  Show this message and exit.

Commands:
[...]

PR #2834, opened 2026-04-29
Reconfigure & UI bugfixes

gregschohn:
### Description

Workflow UI fixes: labels, sorting, approvals, logs, reset, secrets, and VAP messages.

Assorted fixes and improvements to the workflow CLI, TUI, Argo workflow templates, and Kubernetes resource policies discovered during end-to-end testing. The ordering below keeps the highest-impact correctness fixes first, followed by CLI and operator-experience improvements.

#### VAP approval gate fixes

- Added `Pending` to the ApprovalGate CRD phase enum. The retry loop's `resetGate` step was failing because `Pending` was not an allowed value.
- Added `approvalgates/status` to the `workflow-deployer-role` RBAC. The `resetGate` step patches the status subresource, so the workflow needs explicit status-subresource permissions.
- Reformatted RBAC resources as paired `resource` / `resource/status` entries to make future verification easier.

#### VAP denial messages

- Rewrote all gated VAP `messageExpression` blocks to use CEL `filter` + `join` instead of manual comma-prefix string concatenation.
- This fixes empty field names when a field is newly added and `has()` on `oldObject` is false.
- The new expressions are also easier to read and extend as gated fields change.
- Fixed Rich markup escaping in workflow tree labels and manage approval dialogs so bracket-containing text like `[noCapture]` renders literally.

#### Retry loop output fix

- `currentConfigChecksum` output on all `reconcile*WithRetry` templates now guards against nil outputs when `tryApply` fails.
- Failed `tryApply` branches now default to `""`, allowing the caller's `checksumNotDone` guard to proceed correctly after the retry loop completes.

#### Approval gate naming, labels, and behavior

- Standardized approval gate names to `type.resource[.vapretry]`.
- Added structured gate labels used by approval/reset/log workflows:
  - `migrations.opensearch.org/workflow`
  - `migrations.opensearch.org/resource-kind`
  - `migrations.opensearch.org/resource-name`
  - `migrations.opensearch.org/source`
  - `migrations.opensearch.org/target`
  - `migrations.opensearch.org/snapshot`
  - `migrations.opensearch.org/migration`
- Removed unused Kafka child-resource gates and kept approval at the root KafkaCluster gate, where the migration VAP applies.
- `ApprovalGate.status.phase` is treated as authoritative for display:
  - `workflow status` overlays live gate phase onto approval nodes.
  - `workflow manage` uses the same gate-phase overlay in the TUI tree.
  - `workflow approve --list` shows a gate as `Approved` even if Argo has not reconciled its waiting node yet.

#### Workflow approve command

- Restructured `workflow approve` into explicit subcommands:
  - `workflow approve step`
  - `workflow approve change`
  - `workflow approve retry`
- Added top-level `workflow approve --list` to list gates across all approval categories, replacing the separate `list-all` keyword.
- Approval list views now include completed gates while autocomplete remains focused on actionable gates.
- Explicit approval of already-approved gates is idempotent and exits successfully.
- Retry approvals use gate resource labels to check whether the underlying resource still exists before allowing approval.
- Approval output/status text was tightened so waiting, pending, approved, gated-change, and reset-required states are clearer.

#### Workflow log command

- Renamed `workflow output` to `workflow log`.
- Replaced the single output command with subcommands:
  - `workflow log all`
  - `workflow log resource`
  - `workflow log filter`
- Added resource-backed log selection from migration CR names.
- Added filter options for common labels: `--source`, `--target`, `--snapshot`, `--task`, and `--from-snapshot-migration`.
- Added value autocompletion for log filter options, with overconstrained values pruned from completion results.
- Preserved raw selector support with repeated `--label key=value`.
- Kept pass-through support for kubectl/stern arguments after `--`.

#### Pod and resource labels for log filtering

- Added `migrations.opensearch.org/*` labels to CaptureProxy and TrafficReplayer Deployment pods, matching what createSnapshot, metadataMigration, and documentBulkLoad already do.
- Added migration/workflow labels to generated workflow manifests and migration resources.
- Labels applied include `source`, `target`, `task`, and `workflows.argoproj.io/workflow` where applicable.
- Removed target labels from snapshot-related resources where target context was not meaningful.
- Updated config transformation and workflow template snapshots to carry the new label structure.

#### Reset command

- Updated reset input/output to use resource display names in `type.name` form.
- Renamed the reset argument from `path` to `names`.
- Allowed multiple reset targets in one command.
- Aligned reset display names with the approval resource naming model.

#### Configure secret commands

- Added `workflow configure secret` management commands for HTTP basic auth secrets:
  - `list`
  - `create`
  - `update`
  - `delete`
- Commands operate only on secrets carrying the existing HTTP basic credential use-case label.
- Added list/show-missing flows to make missing or managed secrets visible without relying on tab completion.
- Allowed secrets to be created before they are referenced by the active workflow config.

#### Workflow manage sort order

- Fixed top-level workflow sections (Kafka, Proxy, Snapshot, SnapshotMigration, Replay) appearing in timestamp order instead of the intended `sortOrder_view` order.
- `sortOrder_view` was being stripped by the streaming Argo API parser in `manage_injections.py`; it is now preserved.
- Added sorting to `TreeStateManager` matching the sort logic in `display_workflow_tree`.
- `_update_recursive` now rebuilds tree levels when nodes are added or removed to maintain sort order.

#### General CLI and console fixes

- Added `-h` as an alias for `--help` across workflow commands.
- Hid expert workflow-selection and connection flags from normal help output while preserving env-var/configuration support.
- Avoided slow shutdown behavior in the migration console startup/exit flow.
- Updated Kiro workflow steering/config to match the revised command patterns.

### Issues Resolved
https://github.com/opensearch-project/opensearch-migrations/issues/2801
Most of the bugs above were not opened

### Check List


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/migrationConsole.yaml
```diff
@@ -251,6 +252,20 @@ spec:
             - "-c"
             - |
               /root/start-console.sh
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                  - "/bin/sh"
+                  - "-c" # For the time described, notice not just the sleep value but also the terminationGracePeriodSeconds on the pod
+                  - |
```
AndreKurait, line 261: Neat!

migrationConsole/lib/console_link/console_link/workflow/commands/configure.py
```diff
@@ -181,6 +184,169 @@ def edit_config(ctx, stdin):
...
+        _show_secret_names(_managed_secret_names(ctx), NO_MANAGED_HTTP_BASIC_SECRETS_MESSAGE)
+        return
+    if not name:
+        click.echo("Error: specify a secret name or --list.\n", err=True)
+        click.echo(ctx.get_help())
+        ctx.exit(1)
+
+    secret_store = _credentials_secret_store(ctx)
+    _require_managed_secret(secret_store, name)
+    if not confirm and not click.confirm(f"Delete managed HTTP-Basic auth secret '{name}'?"):
```
AndreKurait, line 344: Can we gate interactive prompts to be non-default? So that an agent running delete won't hang on input requested?

migrationConsole/lib/console_link/console_link/workflow/tree_utils.py
```diff
@@ -6,10 +6,16 @@
...
 from rich.tree import Tree
+from kubernetes import client
 import json
+from .models.utils import load_k8s_config
 
 logger = logging.getLogger(__name__)
 
+CRD_GROUP = 'migrations.opensearch.org'
+CRD_VERSION = 'v1alpha1'
+APPROVAL_GATE_PLURAL = 'approvalgates'
```
AndreKurait, line 17: Should this reference an enum in migrationConsole/lib/console_link/console_link/workflow/commands/crd_utils.py?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -98,6 +99,38 @@ export const FullMigration = WorkflowBuilder.create({
...
+patch='{"metadata":{"ownerReferences":[{"apiVersion":"argoproj.io/v1alpha1","kind":"Workflow","name":"{{workflow.name}}","uid":"{{workflow.uid}}"}]}}'
+kubectl get approvalgates.migrations.opensearch.org -l "$selector" -o name \\
+  | xargs -r -n 1 kubectl patch --type merge -p "$patch" \\
+  || echo "WARN: failed to patch one or more approvalgate ownerReferences" >&2
+`])
+        )
+    )
+
+
+    .addTemplate("cleanupApprovalGates", t => t
```
AndreKurait, line 119: Should we add the retry policy here?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -98,6 +99,38 @@ export const FullMigration = WorkflowBuilder.create({
         .addSteps(b => b.addStepGroup(c => c)))
 
 
+    .addTemplate("addApprovalGateOwnerReferences", t => t
```
AndreKurait, line 102: Should we add the retry policy here

migrationConsole/lib/console_link/console_link/workflow/commands/configure.py
gregschohn: how are we handling this for configure in general?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
gregschohn: yes on both.  These aren't critical - they're to show NOT approvals if a workflow has finished or has been deleted - but there's no reason to have pod crash hold those up.

migrationConsole/lib/console_link/console_link/workflow/tree_utils.py
gregschohn: Is there an enum? We have a dict that maps plurals to singular versions and that includes approval gates, but also 6 other resources that aren't relevant.  Even the approval gate occurrence in crd_utils that is currently present is really only there for completeness.  I don't believe that it's necessary.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
gregschohn: I'll do onError retries (meaning that if the process ran but returned non-zero we won't retry).  That will let the container rerun even after scheduling errors, node crashes, etc.

We do NOT do that for createSnapshot, metadata, or other container templates.  Given the breadth of that change, I'll file another PR after this one merges to do so.

AndreKurait (approved):
Left nitpicks above, happy to get this in when you feel it's ready

PR #2835, opened 2026-04-29
feat(solr-transforms): Support delete-by-query via _delete_by_query

niravpi:
## Summary

   Adds support for Solr's delete-by-query command (`{"delete":{"query":"..."}}`) by translating it to OpenSearch's `_delete_by_query` API. The Solr query string is translated using the existing `translateQ`
   pipeline — the same parser → AST → transformer chain the read path uses — so any query that works on `/select` also works in delete-by-query.

   This PR is independent of PR #2827 (NDJSON/_bulk framework) — delete-by-query uses a single JSON body to `_delete_by_query`, not the `_bulk` API.

   ## Motivation

   Delete-by-query is one of the most commonly used Solr write operations after single-doc add/delete. SolrJ clients use it for bulk cleanup (`*:*`), conditional deletion (`status:draft`), and data lifecycle
   management. The shim previously rejected it with a fail-fast error.

   ## Changes

   ### `features/delete-by-query.ts` (new)
   - Request: passes Solr query through `translateQ` in **fail-fast mode** (no passthrough — destructive op should not execute an untranslated query). Forwards all URL params (`df`, `defType`, `qf`, etc.) to the query engine.
   - `wait_for_completion=true` always set (Solr clients expect synchronous response).
   - `commit=true` / `commitWithin` → `?refresh=true` (immediate visibility).
   - Response: `{took, total, deleted, ...}` → `{responseHeader: {status, QTime: took}}`. Version conflicts or `deleted < total` → `status: 1`.

   ### `features/update-router.ts` (modified)
   - `handleDelete`: routes `data.has('query')` to new handler instead of fail-fast.
   - `SUPPORTED_DELETE_FIELDS`: `{id, query}` — mutually exclusive (both present → error).
   - Response dispatch extended with `isDeleteByQueryResponse` branch.

   ### `update-router.testcase.ts` (modified)
   - Replaced old error-expectation test with 12 new E2E integration tests.

   ### `docs/LIMITATIONS.md`
   - New `DELETE-BY-QUERY` shortcode documenting: sync-only behavior, fail-fast-only (no passthrough), version-conflict handling.
   - `UPDATE-COMMANDS` table updated: delete-by-query row changed from ❌ to ✅.

   ## Tests

   ### Unit tests — 27 new
   | Suite | Count | Coverage |
   |---|---|---|
   | `delete-by-query.test.ts` | 20 | Request: simple/match-all/commit/no-commit/empty/null/whitespace/unparseable/trim. Predicate: match/reject cases. Response: all-success/version-conflicts/failures/deleted<total/strips-fields/missing-took/zero-total. |
   | `update-router.test.ts` | +7 | Dispatch: mutual exclusion id+query, SUPPORTED_DELETE_FIELDS includes query, unsupported field rejection, response match/dispatch, single-doc still works. |

   ## Validation

   | Gate | Result |
   |---|---|
   | TS vitest | **847/847** passing (+27 from 820 baseline) |
   | Gradle `:SolrTransformations:test` | green |
   | E2E `:isolatedTest` | **0 new regressions** |

   ## Limitations

   - **No passthrough mode** — queries that the read path would pass through as `query_string` will fail on delete-by-query. Intentional: a destructive operation should not execute an untranslated query.
   - **Version conflicts** — concurrent updates to matching documents can cause `version_conflicts`. OpenSearch continues deleting remaining documents (no abort). Reported as `status: 1`.
   - **Synchronous only** — `wait_for_completion=true` is always set. No async task-based mode exposed.

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-by-query.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-by-query.ts
@@ -1,0 +1,90 @@
+/**
+ * Delete by query — {"delete":{"query":"<solr-query>"}} → POST /_delete_by_query
+ *
+ * Translates a Solr delete-by-query command into OpenSearch's _delete_by_query API.
+ * The Solr query string goes through the existing translateQ pipeline (parser → AST
+ * → transformer), producing the same DSL the read path uses.
+ *
+ * Commit handling:
+ *   - commit=true → ?refresh=true&wait_for_completion=true
+ *   - No commit   → ?wait_for_completion=true only
+ *
+ * Fail-fast policy: delete is destructive — no passthrough on unparseable queries.
+ */
+import type { MicroTransform } from '../pipeline';
+import type { RequestContext, ResponseContext } from '../context';
+import { translateQ } from '../query-engine/orchestrator/translateQ';
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'delete-by-query',
+  match: (ctx) => ctx.body != null && typeof ctx.body.has === 'function' && ctx.body.has('query'),
+  apply: (ctx) => {
+    const queryStr = ctx.body.get('query');
+    if (queryStr == null || String(queryStr).trim() === '') {
+      throw new Error('[delete-by-query] query string is empty — nothing to delete');
+    }
+
+    const solrQuery = String(queryStr).trim();
+    const params = new Map<string, string>();
+    for (const [key, value] of ctx.params.entries()) {
+      params.set(key, value);
+    }
+    params.set('q', solrQuery);
+    const { dsl } = translateQ(params, 'fail-fast');
+
+    const body = new Map<string, unknown>([['query', dsl]]);
+
+    const targetParams = new URLSearchParams();
+    targetParams.set('wait_for_completion', 'true');
+    if (ctx.params.get('commit') === 'true' || ctx.params.has('commitWithin')) {
+      targetParams.set('refresh', 'true');
+    }
+
+    ctx.msg.set('URI', `/${ctx.collection}/_delete_by_query?${targetParams.toString()}`);
+    ctx.msg.set('method', 'POST');
+    ctx.body = body;
+  },
+};
+
+/**
+ * Response: _delete_by_query returns {took, total, deleted, batches, failures, ...}
+ * → Solr {responseHeader:{status, QTime:took}}
+ *
+ * Status mapping:
+ *   - deleted === total && no failures → status 0
+ *   - version_conflicts > 0 or failures non-empty → status 1
+ */
+export function isDeleteByQueryResponse(body: any): boolean {
+  return body != null && typeof body.has === 'function'
+    && body.has('deleted') && body.has('total');
+}
+
+export const response: MicroTransform<ResponseContext> = {
+  name: 'delete-by-query-response',
+  match: (ctx) => isDeleteByQueryResponse(ctx.responseBody),
+  apply: (ctx) => {
+    const body = ctx.responseBody;
+    const took = typeof body.get('took') === 'number' ? body.get('took') : 0;
+    const total = typeof body.get('total') === 'number' ? body.get('total') : 0;
+    const deleted = typeof body.get('deleted') === 'number' ? body.get('deleted') : 0;
+    const versionConflicts = typeof body.get('version_conflicts') === 'number' ? body.get('version_conflicts') : 0;
+    const failures = body.get('failures');
+    const hasFailures = (failures && typeof failures.length === 'number' && failures.length > 0);
+    const hasConflicts = versionConflicts > 0;
+    const incomplete = deleted < total;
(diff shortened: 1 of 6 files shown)


TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-by-query.ts
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * Fail-fast policy: delete is destructive — no passthrough on unparseable queries.
+ */
+import type { MicroTransform } from '../pipeline';
+import type { RequestContext, ResponseContext } from '../context';
+import { translateQ } from '../query-engine/orchestrator/translateQ';
+
+export const request: MicroTransform<RequestContext> = {
+  name: 'delete-by-query',
+  apply: (ctx) => {
```
akshay2000, line 21: Is `match` not required here?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-by-query.ts
```diff
@@ -0,0 +1,89 @@
...
+  return body != null && typeof body.has === 'function'
+    && body.has('deleted') && body.has('total');
+}
+
+export const response: MicroTransform<ResponseContext> = {
+  name: 'delete-by-query-response',
+  match: (ctx) => isDeleteByQueryResponse(ctx.responseBody),
+  apply: (ctx) => {
+    const body = ctx.responseBody;
+    const took = typeof body.get('took') === 'number' ? body.get('took') : 0;
```
akshay2000, line 67: Should we set all of these to -1 instead of 0? Be cause there could genuinely be 0 version conflicts, etc.

Should we also expose this extra metadata somewhere in the response?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
```diff
@@ -145,16 +146,22 @@ function handleDelete(ctx: RequestContext, data: any): void {
...
-  if (data.has('query')) {
-    throw new Error('[update-router] delete: delete-by-query is not supported yet');
-  }
   // Fail fast on unsupported Solr delete fields
-  const SUPPORTED_DELETE_FIELDS = new Set(['id']);
+  const SUPPORTED_DELETE_FIELDS = new Set(['id', 'query']);
   for (const key of data.keys()) {
     if (!SUPPORTED_DELETE_FIELDS.has(key)) {
-      throw new Error(`[update-router] delete: unsupported field '${key}' — only 'id' is supported`);
+      throw new Error(`[update-router] delete: unsupported field '${key}' — only 'id' and 'query' are supported`);
```
akshay2000, line 154: Nit: We should extract the log prefix into a constant.

akshay2000 (changes requested):

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-by-query.ts
niravpi: Added a match guard for defensive self-documentation:
```
   match: (ctx) => ctx.body != null && typeof ctx.body.has === 'function' && ctx.body.has('query'),
```
 In the current architecture this transform is only invoked explicitly by update-router.ts's handleDelete (when data.has('query') is true), so match is never evaluated at runtime. But adding it makes the intent clear and protects against future misuse if someone registers it directly in the pipeline. Same pattern as update-doc.ts and delete-doc.ts which are also called explicitly by the router.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/delete-by-query.ts
niravpi: Keeping defaults as 0 and not exposing extra metadata, this matches Solr's actual response contract. Solr's delete-by-query response is strictly:
```
   {"responseHeader": {"status": 0, "QTime": 5}}
```
No deleted count, no total, no version_conflicts — Solr never exposes this information to clients. Adding non-standard fields could break SolrJ clients that do strict response parsing.

The 0 default for took is purely defensive (OpenSearch always returns took in _delete_by_query responses). It maps to QTime where 0 means "no timing available" standard Solr convention for fast operations. Using -1 would be non-standard and confusing for Solr clients. But please let me know if you think otherwise.

The status: 1 signal is sufficient for clients to know something went wrong; detailed diagnostics (version conflicts, deleted count) can be available in the shim's logs and OpenTelemetry metrics.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/features/update-router.ts
niravpi: Agreed, [update-router] appears 15 times across the file. Will extract into a constant in PR #2846 (batch/bulk operations) which already modifies update-router.ts extensively.

akshay2000 (approved):

PR #2840, opened 2026-04-29
Fix sample-config renderer dropping 'resources' structure and field descriptions

AndreKurait:
## Summary

Fixes two distinct bugs in the sample-config renderer that together caused the `configure sample` workflow output to both hide structural detail and drop human-readable descriptions on many optional fields.

## Bug 1: `resources` rendered as `# unknown`

`RESOURCE_REQUIREMENTS` is defined via `z.preprocess(fn, innerSchema)` to deep-merge defaults. In Zod v4 this builds a `ZodPipe` whose `_def.in` is the `ZodTransform` (the preprocess function) and `_def.out` is the real schema. `unwrapSchema()` was unconditionally returning `_def.in`, handing the renderer a `ZodTransform` it could not introspect — so the field fell through to:

```
"resources": "",  # unknown
```

instead of rendering the nested `limits` / `requests` structure.

**Fix:** detect the preprocess shape (the `in` side is a `ZodTransform`) and unwrap to `_def.out`; keep existing behaviour for `schema.transform()` / `schema.pipe()` where the source schema is on `_def.in`.

## Bug 2: `# description` comments dropped from many fields

`.describe("...")` can be attached at any point in the construction chain, and the landing spot depends on call order:

| Chain | Description sits on |
|---|---|
| `z.string().default("").optional().describe("DOC")` | `ZodOptional` |
| `z.string().optional().default("").describe("DOC")` | `ZodDefault` |
| `z.string().describe("DOC").default("").optional()` | `ZodString` |
| `z.preprocess(fn, inner).describe("DOC").default(d)` | `ZodPipe` |

The renderer was reading `fullUnwrapType(fieldSchema).description`, but `fullUnwrapType` unwraps all the way to the innermost concrete type — losing any description attached to an outer wrapper. As a result, `jvmArgs`, `podReplicas`, `loggingConfigurationOverrideConfigMap`, the captureProxy `resources` block, SigV4 `service`, `indexAllowlist`, and many others rendered with no `# description` comment even though the schemas were correctly annotated in `userSchemas.ts`.

**Fix:** add a new `getDescription(schema)` helper in `schemaUtilities.ts` that walks the wrapper chain (`ZodOptional` / `ZodDefault` / `ZodNullable` / `ZodPipe` — mirroring `unwrapSchema`'s preprocess-aware pipe traversal) and returns the first non-empty `.description`. The single call site in `makeSample.ts` now uses it. `fullUnwrapType` itself is unchanged, keeping its "return the innermost concrete type" contract intact for any future callers.

## Result

The regenerated sample now shows:
- All 3 `resources` fields as nested `limits` / `requests` blocks with cpu / memory / ephemeral-storage and their component-specific descriptions (captureProxy / replayer / RFS).
- `# description` comments on `jvmArgs`, `podReplicas`, `loggingConfigurationOverrideConfigMap`, `indexAllowlist`, SigV4 `service`, snapshot tuning knobs (`maxSnapshotRateMBPerSec`, `onlyOneSnapshotCreationAtATime`, etc.), `auth`, `allowInsecure`, `endpoint`, and the previously-silent fields throughout the config.

Total sample grows from **282 → 462 lines** — entirely additive surfacing of existing schema structure and documentation. No schema changes.

## Tests

`packages/schemas` test suite: **44/44** pass (was 38/38; +6 new targeted tests).

- 3 new `unwrapSchema` regression tests covering `z.preprocess` / `.transform()` / `.pipe()` directions.
- 6 new `getDescription` tests covering every position the author can place `.describe()` on, plus the "no description present" and "multiple descriptions" edge cases.
- All 3 existing Argo snapshot tests continue to pass (no fixture regen needed — `newSample.wf.yaml` regenerates at pod start from `OVERALL_MIGRATION_CONFIG`).

## Out of scope

Remaining `# unknown` entries in the sample (`auth` discriminated union, `CLUSTER_SPEC_OVERRIDE` / `NODE_POOL_SPEC_OVERRIDE` / `TOPIC_SPEC_OVERRIDE` / `tls` open records) are caused by `z.discriminatedUnion` and `z.record(string, z.any())` — separate renderer limitations not addressed here.

---

<details>
<summary><b>Sample config BEFORE this PR</b> (282 lines, click to expand)</summary>

```yaml
{
#"skipApprovals": false,
## Map of Kafka cluster names to their configurations. Keys become Kubernetes resource names and must be valid DNS labels. If empty and proxies are configured, a 'default' auto-created cluster is used.
#"kafkaClusterConfiguration": {
#    # Kafka cluster configuration: either auto-create a new Strimzi cluster or connect to an existing one.
#    "KAFKA_CLUSTER_CONFIGURATION": {}
##    ## Option 1 (object):
##        {
##        "autoCreate": ""  # unknown
##        }
##    ## Option 2 (object):
##        {
##        "existing": {
###            "enableMSKAuth": false,
##            "kafkaConnection": "",  # string
##            "kafkaTopic": "",  # string
###            "auth": {}
##        }
##        }
#},
# Source Elasticsearch or OpenSearch clusters to migrate from.
"sourceClusters": {
    "source": {
#        "endpoint": "",  # string
#        "allowInsecure": false,
#        "authConfig": {},
##        ## Option 1 (object):
##            {
##            "basic": {
##                # Name of a Kubernetes Secret containing 'username' and 'password' keys for HTTP Basic authentication.
##                "secretName": ""  # string
##            }
##            }
##        ## Option 2 (object):
##            {
##            "sigv4": {
##                # AWS region for SigV4 request signing (e.g. 'us-east-1').
##                "region": "",  # string
###                "service": "es"
##            }
##            }
##        ## Option 3 (object):
##            {
##            "mtls": {
##                # PEM-encoded CA certificate or path to CA certificate file for verifying the server's TLS certificate.
##                "caCert": "",  # string
##                # Name of a Kubernetes TLS Secret containing the client certificate and private key for mutual TLS authentication.
##                "clientSecretName": ""  # string
##            }
##            }
        # Cluster version string in '<ENGINE> <VERSION>' format. Supported engines: 'ES' (Elasticsearch) versions 1, 2, 5, 6, 7, 8; 'OS' (OpenSearch) versions 1, 2, 3; 'SOLR' (Apache Solr) versions 8, 9. Examples: 'ES 7.10.2', 'OS 2.11.0', 'SOLR 9.7.0'.
        "version": "",  # string
#        # Snapshot repository and snapshot configuration for a source cluster.
#        "snapshotInfo": {
##            # Map of snapshot repository names to their S3 configurations. Keys are the repository names as registered in the source cluster.
##            "repos": {
##                # Configuration for an S3-backed snapshot repository used by the source cluster.
##                "REPO": {
##                    # AWS region where the S3 bucket resides (e.g. 'us-east-2'). Used for S3 client configuration and snapshot repository registration.
##                    "awsRegion": "",  # string
###                    "endpoint": "",  # string
##                    # S3 URI for the snapshot repository in the format 's3://BUCKET_NAME/OPTIONAL_PATH'. The bucket must already exist and be accessible from the source cluster.
##                    "s3RepoPathUri": "",  # string
###                    "s3RoleArn": ""  # string
##                }
##            },
#            # Snapshots to use or create for this source cluster.
#            "snapshots": {
#                # A snapshot configuration bound to a specific repository.
#                "SNAPSHOT": {
#                    # Snapshot configuration: either an externally managed snapshot name or settings to create a new snapshot.
#                    "config": {},
##                    ## Option 1 (object):
##                        {
##                        # Name of a pre-existing snapshot in the source cluster's repository. The workflow will use this snapshot directly without creating a new one.
##                        "externallyManagedSnapshotName": ""  # string
##                        }
##                    ## Option 2 (object):
##                        {
##                        # Configuration for creating a new snapshot of the source cluster.
##                        "createSnapshotConfig": {
###                            "snapshotPrefix": "",  # string
###                            "jvmArgs": "",  # string
###                            "loggingConfigurationOverrideConfigMap": "",  # string
###                            "indexAllowlist": [],  # string[]
###                            "maxSnapshotRateMbPerNode": 0,
###                            "compressionEnabled": false,
###                            "includeGlobalState": true
##                        }
##                        }
#                    # Name of the S3 snapshot repository. Must match a key in the source cluster's snapshotInfo.repos.
#                    "repoName": ""  # string
#                }
#            },
##            "serializeSnapshotCreation": ""  # boolean
#        }
    }
},
# Target OpenSearch clusters to migrate to.
"targetClusters": {
    # Connection configuration for a target OpenSearch cluster. Extends the base cluster config with a required endpoint.
    "target": {
        # HTTP(S) endpoint URL for the target cluster (e.g. 'https://target-cluster:9200/'). Required for target clusters.
        "endpoint": "",  # string
#        "allowInsecure": false,
#        "authConfig": {}
##        ## Option 1 (object):
##            {
##            "basic": {
##                # Name of a Kubernetes Secret containing 'username' and 'password' keys for HTTP Basic authentication.
##                "secretName": ""  # string
##            }
##            }
##        ## Option 2 (object):
##            {
##            "sigv4": {
##                # AWS region for SigV4 request signing (e.g. 'us-east-1').
##                "region": "",  # string
###                "service": "es"
##            }
##            }
##        ## Option 3 (object):
##            {
##            "mtls": {
##                # PEM-encoded CA certificate or path to CA certificate file for verifying the server's TLS certificate.
##                "caCert": "",  # string
##                # Name of a Kubernetes TLS Secret containing the client certificate and private key for mutual TLS authentication.
##                "clientSecretName": ""  # string
##            }
##            }
    }
},
# List of snapshot-based migration configurations. Each entry binds a source cluster to a target cluster and defines which snapshots to migrate with what settings.
"snapshotMigrationConfigs": [
    {
#    "skipApprovals": false,
    # Label of the source cluster to migrate from. Must match a key in sourceClusters.
    "fromSource": "",  # string
    # Label of the target cluster to migrate to. Must match a key in targetClusters.
    "toTarget": "",  # string
#    # Map of snapshot names to their migration configurations. Keys must match snapshot names defined in the source cluster's snapshotInfo.snapshots.
#    "perSnapshotConfig": {
#        "PER_SNAPSHOT_CONFIG": [
#            {
##            "label": "",  # string
##            "metadataMigrationConfig": {
###                "jvmArgs": "",  # string
###                "loggingConfigurationOverrideConfigMap": "",  # string
###                "skipEvaluateApproval": false,
###                "skipMigrateApproval": false,
###                "componentTemplateAllowlist": [],  # string[]
###                "indexAllowlist": [],  # string[]
###                "indexTemplateAllowlist": [],  # string[]
###                "allowLooseVersionMatching": true,
###                "clusterAwarenessAttributes": 1,
###                "multiTypeBehavior": "NONE",
###                "otelCollectorEndpoint": "http://otel-collector:4317",
###                "output": "HUMAN_READABLE",
###                "transformerConfigBase64": "",  # string
###                "transformerConfig": "",  # string
###                "transformerConfigFile": "",  # string
###                "enableSourcelessMigrations": false,
###                "useRecoverySource": false
##            },
##            "documentBackfillConfig": {
###                "podReplicas": 1,
###                "jvmArgs": "",  # string
###                "loggingConfigurationOverrideConfigMap": "",  # string
###                "skipApproval": false,
##                # [Expert] When true, uses the target OpenSearch cluster for RFS work coordination (lease management and shard assignment). When false (default), a dedicated single-node OpenSearch coordinator cluster is automatically deployed within the Kubernetes cluster, used for the lifetime of the migration, then torn down on completion. Using a dedicated coordinator avoids adding coordination overhead to the target cluster.
##                "useTargetClusterForWorkCoordination": false,
##                # Kubernetes resource limits and requests for the RFS container. Partial overrides are deep-merged with the built-in defaults. By default, limits equal requests, giving the pod 'Guaranteed' QoS (least likely to be evicted). Setting requests lower than limits results in 'Burstable' QoS. Ephemeral storage is auto-calculated from maxShardSizeBytes if not specified.
##                "resources": "",  # unknown
###                "indexAllowlist": [],  # string[]
###                "allowLooseVersionMatching": true,
###                "docTransformerConfigBase64": "",  # string
###                "docTransformerConfig": "",  # string
###                "docTransformerConfigFile": "",  # string
###                "documentsPerBulkRequest": 2147483647,
###                "documentsSizePerBulkRequest": 10485760,
###                "initialLeaseDuration": "PT1H",
###                "maxConnections": 10,
###                "maxShardSizeBytes": 85899345920,
###                "otelCollectorEndpoint": "http://otel-collector:4317",
###                "serverGeneratedIds": "AUTO",
###                "allowedDocExceptionTypes": [],  # string[]
###                "coordinatorRetryMaxRetries": 7,
###                "coordinatorRetryInitialDelayMs": 1000,
###                "coordinatorRetryMaxDelayMs": 64000,
###                "enableSourcelessMigrations": false,
###                "useRecoverySource": false
##            }
#            }
#        ]
#    }
    }
],
## Traffic capture and replay configuration. Proxies capture live traffic from source clusters to Kafka, and replayers consume from Kafka to replay against target clusters. All top-level items are independent, but replayers can declare dependencies on snapshot migrations to ensure data consistency.
#"traffic": {
#    # Map of proxy names to their capture configurations. Keys become the Kubernetes Service names and must be valid DNS labels.
#    "proxies": {
#        # Configuration for a single capture proxy instance, including its Kafka topic and source cluster binding.
#        "PROXIE": {
##            "kafka": "default",
##            "kafkaTopic": "",  # string
#            # Name of the source cluster this proxy sits in front of. Must match a key in sourceClusters.
#            "source": "",  # string
#            # Configuration for the capture proxy deployment and process options.
#            "proxyConfig": {
##                "loggingConfigurationOverrideConfigMap": "",  # string
##                "internetFacing": false,
##                "podReplicas": 1,
#                "resources": {},
##                "otelCollectorEndpoint": "http://otel-collector:4317",
##                "setHeader": [],  # string[]
##                "destinationConnectionPoolSize": 0,
##                "destinationConnectionPoolTimeout": "PT30S",
##                "kafkaClientId": "HttpCaptureProxyProducer",
#                # TCP port the capture proxy listens on for incoming HTTP(S) traffic. This port is exposed via the Kubernetes Service and used to construct the proxy endpoint URL.
#                "listenPort": "",  # number
##                "maxTrafficBufferSize": 1048576,
##                "noCapture": false,
##                "numThreads": 1,
##                # TLS configuration for the capture proxy. When omitted, a self-signed certificate is automatically provisioned via cert-manager. Specify mode 'plaintext' to opt out.
##                "tls": "",  # unknown
##                "enableMSKAuth": false,
##                "suppressCaptureForHeaderMatch": [],  # string[]
##                "suppressCaptureForMethod": "",  # string
##                "suppressCaptureForUriPath": "",  # string
##                "suppressMethodAndPath": ""  # string
#            }
#        }
#    },
#    # Map of replayer names to their replay configurations. Each replayer consumes from a proxy's Kafka topic and replays to a target cluster.
#    "replayers": {
#        # Configuration for a single traffic replayer instance, binding a proxy's captured traffic to a target cluster.
#        "REPLAYER": {
#            # Name of the capture proxy to replay traffic from. Must match a key in traffic.proxies.
#            "fromProxy": "",  # string
#            # Name of the target cluster to replay traffic to. Must match a key in targetClusters.
#            "toTarget": "",  # string
##            "dependsOnSnapshotMigrations": [
##                {
##                # Name of the source cluster. Must match a key in sourceClusters.
##                "source": "",  # string
##                # Name of the snapshot. Must match a key in the source cluster's snapshotInfo.snapshots.
##                "snapshot": ""  # string
##                }
##            ],
##            "replayerConfig": {
###                "jvmArgs": "",  # string
###                "loggingConfigurationOverrideConfigMap": "",  # string
###                "podReplicas": 1,
##                # Kubernetes resource limits and requests for the replayer container. Partial overrides are deep-merged with the built-in defaults. By default, limits equal requests, giving the pod 'Guaranteed' QoS (least likely to be evicted). Setting requests lower than limits results in 'Burstable' QoS, allowing the pod to use less resources when idle but burst up to the limit.
##                "resources": "",  # unknown
###                "kafkaTrafficEnableMSKAuth": false,
###                "kafkaTrafficPropertyFile": "",  # string
###                "lookaheadTimeSeconds": 400,
###                "maxConcurrentRequests": 10000,
###                "numClientThreads": 0,
###                "nonRetryableDocExceptionTypes": [],  # string[]
###                "observedPacketConnectionTimeout": 360,
###                "otelCollectorEndpoint": "http://otel-collector:4317",
###                "quiescentPeriodMs": 5000,
###                "removeAuthHeader": false,
###                "speedupFactor": 1.1,
###                "targetServerResponseTimeoutSeconds": 150,
###                "transformerConfig": "",  # string
###                "transformerConfigEncoded": "",  # string
###                "transformerConfigFile": "",  # string
###                "tupleTransformerConfig": "",  # string
###                "tupleTransformerConfigBase64": "",  # string
###                "tupleTransformerConfigFile": "",  # string
###                "tupleMaxBufferSeconds": 60,
###                "tupleMaxFileSizeMb": 256,
###                "userAgent": ""  # string
##            }
#        }
#    }
#}
}


```

</details>

<details>
<summary><b>Sample config AFTER this PR</b> (462 lines, click to expand)</summary>

```yaml
{
## Global flag to skip all manual approval gates across the entire migration. When true, overrides all per-component skipApproval settings.
#"skipApprovals": false,
## Kafka cluster configurations. If empty and traffic capture is configured, a default ephemeral Kafka cluster is auto-created for each referenced cluster label. Each entry defines a Kafka cluster (auto-created or external) referenced by proxies via 'kafka'.
#"kafkaClusterConfiguration": {
#    # Kafka cluster configuration: either auto-create a new Strimzi cluster or connect to an existing one.
#    "KAFKA_CLUSTER_CONFIGURATION": {}
##    ## Option 1 (object):
##        {
##        # Workflow-managed Strimzi Kafka cluster creation. Structural defaults for broker config, node pool, and topic settings are deep-merged here, while the auth default is resolved separately during transform-time policy application.
##        "autoCreate": {
###            # Workflow-owned Kafka client auth for auto-created Strimzi clusters. If omitted, transform-time resolution currently defaults workflow-managed Kafka to `scram-sha-512` as the secure-by-default policy. This default is intentionally resolved outside the deep-merged Strimzi object defaults so auth policy can change without being hidden inside the structural Kafka/NodePool/Topic default merge.
###            "auth": "",  # unknown
###            # Optional overrides merged into the generated Strimzi Kafka.spec. Workflow-managed fields such as resource names, required listeners, and workflow-owned auth settings may be overwritten by the workflow.
###            "clusterSpecOverrides": {
###                "CLUSTER_SPEC_OVERRIDE": ""  # unknown
###            },
###            # Optional overrides merged into the generated Strimzi KafkaNodePool.spec. Workflow-managed fields such as cluster labels may be overwritten by the workflow.
###            "nodePoolSpecOverrides": {
###                "NODE_POOL_SPEC_OVERRIDE": ""  # unknown
###            },
###            # Optional overrides merged into generated Strimzi KafkaTopic.spec values for workflow-created topics.
###            "topicSpecOverrides": {
###                "TOPIC_SPEC_OVERRIDE": ""  # unknown
###            }
##        }
##        }
##    ## Option 2 (object):
##        {
##        "existing": {
###            "enableMSKAuth": false,
##            "kafkaConnection": "",  # string
##            # Empty defaults to the name of the target label
##            "kafkaTopic": "",  # string
###            "auth": {}
##        }
##        }
#},
# Source Elasticsearch or OpenSearch clusters to migrate from.
"sourceClusters": {
    "source": {
#        # HTTP(S) endpoint URL for the cluster (e.g. 'https://my-cluster:9200/'). Leave empty if the cluster is not directly accessible or will be accessed through a proxy.
#        "endpoint": "",  # string
#        # When true, disables TLS certificate verification when connecting to the cluster. Use only for development or self-signed certificates.
#        "allowInsecure": false,
#        # Authentication configuration for connecting to the cluster. Supports HTTP Basic (Kubernetes Secret), AWS SigV4, or mutual TLS.
#        "authConfig": {},
##        ## Option 1 (object):
##            {
##            "basic": {
##                # Name of a Kubernetes Secret containing 'username' and 'password' keys for HTTP Basic authentication.
##                "secretName": ""  # string
##            }
##            }
##        ## Option 2 (object):
##            {
##            "sigv4": {
##                # AWS region for SigV4 request signing (e.g. 'us-east-1').
##                "region": "",  # string
###                # AWS service name for SigV4 signing. Use 'es' for Amazon OpenSearch Service or 'aoss' for OpenSearch Serverless.
###                "service": "es"
##            }
##            }
##        ## Option 3 (object):
##            {
##            "mtls": {
##                # PEM-encoded CA certificate or path to CA certificate file for verifying the server's TLS certificate.
##                "caCert": "",  # string
##                # Name of a Kubernetes TLS Secret containing the client certificate and private key for mutual TLS authentication.
##                "clientSecretName": ""  # string
##            }
##            }
        # Cluster version string in '<ENGINE> <VERSION>' format. Supported engines: 'ES' (Elasticsearch) versions 1, 2, 5, 6, 7, 8; 'OS' (OpenSearch) versions 1, 2, 3; 'SOLR' (Apache Solr) versions 8, 9. Examples: 'ES 7.10.2', 'OS 2.11.0', 'SOLR 9.7.0'.
        "version": "",  # string
#        # Snapshot repository and snapshot configurations for this source cluster. Required if any snapshot-based migrations reference this source.
#        "snapshotInfo": {
##            # S3 snapshot repositories registered with the source cluster.
##            "repos": {
##                # Configuration for an S3-backed snapshot repository used by the source cluster.
##                "REPO": {
##                    # AWS region where the S3 bucket resides (e.g. 'us-east-2'). Used for S3 client configuration and snapshot repository registration.
##                    "awsRegion": "",  # string
###                    # Override the S3 endpoint URL. Supports http://, https://, localstack://, and localstacks:// schemes. LocalStack endpoints are automatically resolved to IP addresses during config transformation.
###                    "endpoint": "",  # string
##                    # S3 URI for the snapshot repository in the format 's3://BUCKET_NAME/OPTIONAL_PATH'. The bucket must already exist and be accessible from the source cluster.
##                    "s3RepoPathUri": "",  # string
###                    # IAM role ARN that the source cluster will assume to read/write snapshots to S3. This is passed to the cluster when registering the snapshot repository. Leave empty if the cluster's own IAM role already has S3 access.
###                    "s3RoleArn": ""  # string
##                }
##            },
#            # Snapshots to use or create for this source cluster.
#            "snapshots": {
#                # A snapshot configuration bound to a specific repository.
#                "SNAPSHOT": {
#                    # Snapshot configuration: either an externally managed snapshot name or settings to create a new snapshot.
#                    "config": {},
##                    ## Option 1 (object):
##                        {
##                        # Name of a pre-existing snapshot in the source cluster's repository. The workflow will use this snapshot directly without creating a new one.
##                        "externallyManagedSnapshotName": ""  # string
##                        }
##                    ## Option 2 (object):
##                        {
##                        # Configuration for creating a new snapshot of the source cluster.
##                        "createSnapshotConfig": {
###                            # Prefix for auto-generated snapshot names. When set, the snapshot name is '<snapshotPrefix>_<uniqueId>'. When empty, defaults to '<sourceLabel>_<uniqueId>'.
###                            "snapshotPrefix": "",  # string
###                            # Additional JVM arguments passed via JDK_JAVA_OPTIONS (e.g. '-Xmx4g -XX:+UseG1GC'). If setting -Xmx, ensure the heap size is smaller than the container's memory limit in resources to account for off-heap memory usage.
###                            "jvmArgs": "",  # string
###                            # Name of a Kubernetes ConfigMap containing a custom Log4j2 properties configuration. The ConfigMap should have a single key whose value is the Log4j2 properties file content. When set, it is mounted into the container and passed via -Dlog4j2.configurationFile. See https://logging.apache.org/log4j/2.x/manual/configuration.html#properties for format reference.
###                            "loggingConfigurationOverrideConfigMap": "",  # string
###                            # Filters which indices are captured at the snapshot layer — evaluated by the source cluster when the snapshot is created. Entries use the cluster's native multi-index expression syntax (the same format accepted by the _snapshot API's 'indices' field): exact names (e.g. 'logs-2024-01'), wildcards (e.g. 'logs-*'), and exclusions via a leading '-' (e.g. '-*-archive'). The entries are joined with commas and sent verbatim as the 'indices' field of the snapshot creation request; this is NOT a regex. Only the matching indices end up in the snapshot, so downstream stages have no way to recover indices excluded here. To further narrow which indices are migrated after the snapshot is taken (including with 'regex:' patterns), use the metadata or RFS indexAllowlist, which filter client-side on the snapshot contents. An empty list includes all indices.
###                            "indexAllowlist": [],  # string[]
###                            # Maximum snapshot throughput in MB/s per data node. 0 means no rate limiting. Use to reduce I/O impact on the source cluster during snapshot creation.
###                            "maxSnapshotRateMbPerNode": 0,
###                            # [Expert] Enables metadata compression for the snapshot. Must be set to false for Elasticsearch 1.x sources, as compressed snapshot metadata is not supported by the snapshot reader for that version.
###                            "compressionEnabled": false,
###                            # [Expert] Includes cluster global state (persistent settings, templates, etc.) in the snapshot. Only disable if metadata migration encounters template processing issues that cannot be resolved via an allowlist.
###                            "includeGlobalState": true
##                        }
##                        }
#                    # Name of the S3 snapshot repository. Must match a key in the source cluster's snapshotInfo.repos.
#                    "repoName": ""  # string
#                }
#            },
##            # Controls whether snapshot creations for this source run one-at-a-time or in parallel. When true, all snapshot creations share a single semaphore so only one runs at any given time; when false, each snapshot can run in parallel with others for this source. When omitted, defaults to true for legacy sources (ES 1-7, OS 1) and false for modern sources (ES 8+, OS 2+). Set explicitly to override the version-based default. Common reason to force true on a modern source: the cluster only supports one snapshot at a time for the indices being captured (for example, OpenSearch UltraWarm indices, which only allow a single index per snapshot and cannot be snapshotted concurrently).
##            "serializeSnapshotCreation": ""  # boolean
#        }
    }
},
# Target OpenSearch clusters to migrate to.
"targetClusters": {
    # Connection configuration for a target OpenSearch cluster. Extends the base cluster config with a required endpoint.
    "target": {
        # HTTP(S) endpoint URL for the target cluster (e.g. 'https://target-cluster:9200/'). Required for target clusters.
        "endpoint": "",  # string
#        # When true, disables TLS certificate verification when connecting to the cluster. Use only for development or self-signed certificates.
#        "allowInsecure": false,
#        # Authentication configuration for connecting to the cluster. Supports HTTP Basic (Kubernetes Secret), AWS SigV4, or mutual TLS.
#        "authConfig": {}
##        ## Option 1 (object):
##            {
##            "basic": {
##                # Name of a Kubernetes Secret containing 'username' and 'password' keys for HTTP Basic authentication.
##                "secretName": ""  # string
##            }
##            }
##        ## Option 2 (object):
##            {
##            "sigv4": {
##                # AWS region for SigV4 request signing (e.g. 'us-east-1').
##                "region": "",  # string
###                # AWS service name for SigV4 signing. Use 'es' for Amazon OpenSearch Service or 'aoss' for OpenSearch Serverless.
###                "service": "es"
##            }
##            }
##        ## Option 3 (object):
##            {
##            "mtls": {
##                # PEM-encoded CA certificate or path to CA certificate file for verifying the server's TLS certificate.
##                "caCert": "",  # string
##                # Name of a Kubernetes TLS Secret containing the client certificate and private key for mutual TLS authentication.
##                "clientSecretName": ""  # string
##            }
##            }
    }
},
# List of snapshot-based migration configurations. Each entry binds a source cluster to a target cluster and defines which snapshots to migrate with what settings.
"snapshotMigrationConfigs": [
    {
#    # When true, skips all manual approval gates for migrations in this configuration block.
#    "skipApprovals": false,
    # Label of the source cluster to migrate from. Must match a key in sourceClusters.
    "fromSource": "",  # string
    # Label of the target cluster to migrate to. Must match a key in targetClusters.
    "toTarget": "",  # string
#    # Per-snapshot migration configurations. Each entry maps a snapshot name to one or more migration passes (metadata + document backfill).
#    "perSnapshotConfig": {
#        "PER_SNAPSHOT_CONFIG": [
#            {
##            # Unique label for this migration within its snapshot group. Auto-generated as 'migration-<index>' if not specified. Must start with a letter and contain only alphanumeric characters.
##            "label": "",  # string
##            # Configuration for migrating index metadata (mappings, settings, templates) from the snapshot to the target. Omit to skip metadata migration.
##            "metadataMigrationConfig": {
###                # Additional JVM arguments passed via JDK_JAVA_OPTIONS (e.g. '-Xmx4g -XX:+UseG1GC'). If setting -Xmx, ensure the heap size is smaller than the container's memory limit in resources to account for off-heap memory usage.
###                "jvmArgs": "",  # string
###                # Name of a Kubernetes ConfigMap containing a custom Log4j2 properties configuration. The ConfigMap should have a single key whose value is the Log4j2 properties file content. When set, it is mounted into the container and passed via -Dlog4j2.configurationFile. See https://logging.apache.org/log4j/2.x/manual/configuration.html#properties for format reference.
###                "loggingConfigurationOverrideConfigMap": "",  # string
###                # When true, skips the manual approval gate after the metadata evaluation step. The evaluation step analyzes what metadata changes would be applied without making changes.
###                "skipEvaluateApproval": false,
###                # When true, skips the manual approval gate after the metadata migration step. The migration step applies the evaluated metadata changes to the target cluster.
###                "skipMigrateApproval": false,
###                # List of component template names to include in the metadata migration. Each entry is either an exact name or a regex pattern prefixed with 'regex:'. An empty list includes all non-system component templates.
###                "componentTemplateAllowlist": [],  # string[]
###                # Filters which indices are migrated at the metadata stage — evaluated client-side on the snapshot contents after the snapshot has been taken. Each entry is either an exact index name (e.g. 'my-index') or a regex pattern prefixed with 'regex:' (e.g. 'regex:logs-.*'). Applies only among indices already captured in the snapshot; to exclude an index from the snapshot itself, use the CreateSnapshot indexAllowlist. An empty list includes all non-system indices.
###                "indexAllowlist": [],  # string[]
###                # List of index template names to include in the metadata migration. Each entry is either an exact name or a regex pattern prefixed with 'regex:'. An empty list includes all non-system index templates.
###                "indexTemplateAllowlist": [],  # string[]
###                # [Expert] Allows migration between clusters with non-exact version compatibility (e.g. ES 7.x to OS 2.x). Only disable if metadata has parsing issues on snapshots that require strict version matching.
###                "allowLooseVersionMatching": true,
###                # Number of shard allocation awareness attributes to preserve during metadata migration. Controls how index settings related to cluster topology are handled.
###                "clusterAwarenessAttributes": 1,
###                # Strategy for handling Elasticsearch multi-type indices (ES 5.x and earlier). 'NONE': fail if multi-type indices are encountered. 'UNION': merge all types into a single mapping. 'SPLIT': create separate indices for each type.
###                "multiTypeBehavior": "NONE",
###                # URL for the OpenTelemetry Collector endpoint used for metrics and traces (e.g. 'http://otel-collector:4317').
###                "otelCollectorEndpoint": "http://otel-collector:4317",
###                # Output format for the metadata migration evaluation report. 'HUMAN_READABLE' for formatted text, 'JSON' for machine-parseable output.
###                "output": "HUMAN_READABLE",
###                # Base64-encoded JSON transformer configuration. Metadata transformers modify index mappings and settings during migration.
###                "transformerConfigBase64": "",  # string
###                # Inline JSON transformer configuration. Keys are transformer names and values are their configuration. Metadata transformers modify index mappings and settings during migration.
###                "transformerConfig": "",  # string
###                # Path to a JSON file containing transformer configuration. Metadata transformers modify index mappings and settings during migration. [Expert] The file must be mounted into the container by the user (e.g. via Kyverno pod mutation or custom image). Not wired through the workflow by default.
###                "transformerConfigFile": "",  # string
###                # Enable migration of indices that have _source disabled or partially filtered (includes/excludes). When enabled, document backfill will reconstruct documents from stored fields and doc_values. Without this flag, metadata migration will fail if any selected index has _source disabled or partially filtered.
###                "enableSourcelessMigrations": false,
###                # When enabled, treat the _recovery_source stored field (present in ES 7+ / OpenSearch snapshots with soft-deletes) as _source. This field is transient and may not be present for all documents, so results can be inconsistent. Use only when reconstruction from doc_values and stored fields is insufficient.
###                "useRecoverySource": false
##            },
##            # Configuration for backfilling documents from the snapshot to the target using Reindex From Snapshot. Omit to skip document backfill.
##            "documentBackfillConfig": {
###                # Number of RFS (Reindex From Snapshot) pod replicas. Each replica processes shards independently.
###                "podReplicas": 1,
###                # Additional JVM arguments passed via JDK_JAVA_OPTIONS (e.g. '-Xmx4g -XX:+UseG1GC'). If setting -Xmx, ensure the heap size is smaller than the container's memory limit in resources to account for off-heap memory usage.
###                "jvmArgs": "",  # string
###                # Name of a Kubernetes ConfigMap containing a custom Log4j2 properties configuration. The ConfigMap should have a single key whose value is the Log4j2 properties file content. When set, it is mounted into the container and passed via -Dlog4j2.configurationFile. See https://logging.apache.org/log4j/2.x/manual/configuration.html#properties for format reference.
###                "loggingConfigurationOverrideConfigMap": "",  # string
###                # When true, skips the manual approval gate after the document backfill completes. Useful for automated pipelines where human approval is not needed.
###                "skipApproval": false,
##                # [Expert] When true, uses the target OpenSearch cluster for RFS work coordination (lease management and shard assignment). When false (default), a dedicated single-node OpenSearch coordinator cluster is automatically deployed within the Kubernetes cluster, used for the lifetime of the migration, then torn down on completion. Using a dedicated coordinator avoids adding coordination overhead to the target cluster.
##                "useTargetClusterForWorkCoordination": false,
##                # Kubernetes resource limits and requests for the RFS container. Partial overrides are deep-merged with the built-in defaults. By default, limits equal requests, giving the pod 'Guaranteed' QoS (least likely to be evicted). Setting requests lower than limits results in 'Burstable' QoS. Ephemeral storage is auto-calculated from maxShardSizeBytes if not specified.
##                "resources": {
##                    # Maximum resource limits for the container. The container will be terminated if it exceeds these limits.
##                    "limits": {
##                        # CPU allocation for the container in Kubernetes millicores.
##                        "cpu": "",  # string
##                        # Memory allocation for the container.
##                        "memory": "",  # string
###                        # Ephemeral storage allocation for the container. Used for temporary on-disk data such as Lucene index segments during RFS document migration.
###                        "ephemeral-storage": ""  # string
##                    },
##                    # Minimum guaranteed resources for the container. Used by the Kubernetes scheduler for pod placement.
##                    "requests": {
##                        # CPU allocation for the container in Kubernetes millicores.
##                        "cpu": "",  # string
##                        # Memory allocation for the container.
##                        "memory": "",  # string
###                        # Ephemeral storage allocation for the container. Used for temporary on-disk data such as Lucene index segments during RFS document migration.
###                        "ephemeral-storage": ""  # string
##                    }
##                },
###                # Filters which indices are migrated by the document backfill (RFS) — evaluated client-side on the snapshot contents after the snapshot has been taken. Each entry is either an exact index name or a regex pattern prefixed with 'regex:' (e.g. 'regex:logs-.*'). Applies only among indices already captured in the snapshot; to exclude an index from the snapshot itself, use the CreateSnapshot indexAllowlist. An empty list includes all non-system indices from the snapshot.
###                "indexAllowlist": [],  # string[]
###                # [Expert] Allows document migration between clusters with non-exact version compatibility. Only disable if snapshot parsing issues require strict version matching.
###                "allowLooseVersionMatching": true,
###                # Base64-encoded JSON transformer configuration. Document transformers modify each document during the snapshot backfill (e.g. field renaming, type conversion).
###                "docTransformerConfigBase64": "",  # string
###                # Inline JSON transformer configuration. Keys are transformer names and values are their configuration. Document transformers modify each document during the snapshot backfill (e.g. field renaming, type conversion).
###                "docTransformerConfig": "",  # string
###                # Path to a JSON file containing transformer configuration. Document transformers modify each document during the snapshot backfill (e.g. field renaming, type conversion). [Expert] The file must be mounted into the container by the user (e.g. via Kyverno pod mutation or custom image). Not wired through the workflow by default.
###                "docTransformerConfigFile": "",  # string
###                # Maximum number of documents per bulk indexing request to the target cluster. Lower values reduce per-request latency but increase overhead.
###                "documentsPerBulkRequest": 2147483647,
###                # Maximum aggregate document size in bytes per bulk indexing request. Individual documents larger than this limit are sent as single-document requests.
###                "documentsSizePerBulkRequest": 10485760,
###                # ISO 8601 duration for the initial work item lease in the coordination store (e.g. 'PT1H' = 1 hour, 'PT10M' = 10 minutes). If a worker fails to complete a shard within this duration, the lease expires and another worker can pick it up, doubling the lease duration on each retry. Increase for very large shards (>200GB) to reduce the number of re-downloads per shard needed to complete the migration.
###                "initialLeaseDuration": "PT1H",
###                # Maximum number of concurrent HTTP connections from each RFS worker to the target cluster for bulk indexing.
###                "maxConnections": 10,
###                # Expected maximum shard size in bytes. Used to auto-calculate ephemeral storage requirements as ceil(2.5 * maxShardSizeBytes). Set this to match your largest shard to ensure sufficient disk space for Lucene segment processing.
###                "maxShardSizeBytes": 85899345920,
###                # URL for the OpenTelemetry Collector endpoint used for metrics and traces (e.g. 'http://otel-collector:4317').
###                "otelCollectorEndpoint": "http://otel-collector:4317",
###                # Controls document ID generation on the target. 'AUTO': auto-detect serverless TIMESERIES/VECTOR collections and enable server-generated IDs. 'ALWAYS': always use server-generated IDs (discards source IDs). 'NEVER': always preserve source document IDs (may fail on serverless TIMESERIES/VECTOR collections).
###                "serverGeneratedIds": "AUTO",
###                # List of document-level exception types to treat as successful operations during bulk migration. Documents that fail with these errors are not retried and not counted as failures — they are silently accepted. Use this for idempotent migrations where certain errors are expected and harmless. For example, set to ['version_conflict_engine_exception'] when migrating into a data stream where backing index writes may conflict with existing documents. Defaults to empty (all errors are treated as failures). See BulkDocErrorTypes for common OpenSearch exception type strings.
###                "allowedDocExceptionTypes": [],  # string[]
###                # [Expert] Maximum number of retries when marking work items as completed on the coordinator.
###                "coordinatorRetryMaxRetries": 7,
###                # [Expert] Initial delay in milliseconds for coordinator completion retries. Doubles with each attempt up to coordinatorRetryMaxDelayMs.
###                "coordinatorRetryInitialDelayMs": 1000,
###                # [Expert] Maximum delay in milliseconds for any single coordinator completion retry.
###                "coordinatorRetryMaxDelayMs": 64000,
###                # Enable migration of indices that have _source disabled or partially filtered (includes/excludes). When enabled, documents are reconstructed from stored fields and doc_values instead of _source. Without this flag, migration of sourceless indices will fail with an error.
###                "enableSourcelessMigrations": false,
###                # When enabled, treat the _recovery_source stored field (present in ES 7+ / OpenSearch snapshots with soft-deletes) as _source. This field is transient and may not be present for all documents, so results can be inconsistent. Use only when reconstruction from doc_values and stored fields is insufficient.
###                "useRecoverySource": false
##            }
#            }
#        ]
#    }
    }
],
## Traffic capture and replay configuration. Proxies capture live traffic from source clusters to Kafka, and replayers consume from Kafka to replay against target clusters. All top-level items are independent, but replayers can declare dependencies on snapshot migrations to ensure data consistency.
#"traffic": {
#    # Map of proxy names to their capture configurations. Keys become the Kubernetes Service names and must be valid DNS labels.
#    "proxies": {
#        # Configuration for a single capture proxy instance, including its Kafka topic and source cluster binding.
#        "PROXIE": {
##            # Label of the Kafka cluster to use for captured traffic. Must match a key in kafkaClusterConfiguration.
##            "kafka": "default",
##            # Kafka topic name for captured traffic. If empty, defaults to the proxy name (the key in the proxies record).
##            "kafkaTopic": "",  # string
#            # Name of the source cluster this proxy sits in front of. Must match a key in sourceClusters.
#            "source": "",  # string
#            # Configuration for the capture proxy deployment and process options.
#            "proxyConfig": {
##                # Name of a Kubernetes ConfigMap containing a custom Log4j2 properties configuration. The ConfigMap should have a single key whose value is the Log4j2 properties file content. When set, it is mounted into the container and passed via -Dlog4j2.configurationFile. See https://logging.apache.org/log4j/2.x/manual/configuration.html#properties for format reference.
##                "loggingConfigurationOverrideConfigMap": "",  # string
##                # When true, the proxy's Kubernetes Service is annotated with 'internet-facing' load balancer scheme, making it accessible from outside the VPC.
##                "internetFacing": false,
##                # Number of proxy pod replicas in the Kubernetes Deployment. Increase for higher throughput or availability.
##                "podReplicas": 1,
#                # Kubernetes resource limits and requests for the capture proxy container. Partial overrides are deep-merged with the built-in defaults. By default, limits equal requests, giving the pod 'Guaranteed' QoS (least likely to be evicted). Setting requests lower than limits results in 'Burstable' QoS, allowing the pod to use less resources when idle but burst up to the limit.
#                "resources": {
#                    # Maximum resource limits for the container. The container will be terminated if it exceeds these limits.
#                    "limits": {
#                        # CPU allocation for the container in Kubernetes millicores.
#                        "cpu": "",  # string
#                        # Memory allocation for the container.
#                        "memory": "",  # string
##                        # Ephemeral storage allocation for the container. Used for temporary on-disk data such as Lucene index segments during RFS document migration.
##                        "ephemeral-storage": ""  # string
#                    },
#                    # Minimum guaranteed resources for the container. Used by the Kubernetes scheduler for pod placement.
#                    "requests": {
#                        # CPU allocation for the container in Kubernetes millicores.
#                        "cpu": "",  # string
#                        # Memory allocation for the container.
#                        "memory": "",  # string
##                        # Ephemeral storage allocation for the container. Used for temporary on-disk data such as Lucene index segments during RFS document migration.
##                        "ephemeral-storage": ""  # string
#                    }
#                },
##                # URL for the OpenTelemetry Collector endpoint used for metrics and traces (e.g. 'http://otel-collector:4317').
##                "otelCollectorEndpoint": "http://otel-collector:4317",
##                # List of static headers to add to proxied requests, each in 'Header-Name: value' format.
##                "setHeader": [],  # string[]
##                # Maximum number of persistent connections to the destination (source) cluster. 0 means unlimited connection pooling.
##                "destinationConnectionPoolSize": 0,
##                # ISO 8601 duration for how long idle connections in the destination pool are kept alive before being closed (e.g. 'PT30S' = 30 seconds, 'PT5M' = 5 minutes).
##                "destinationConnectionPoolTimeout": "PT30S",
##                # Kafka producer client ID used when publishing captured traffic to Kafka. Useful for identifying this proxy in Kafka broker logs and metrics.
##                "kafkaClientId": "HttpCaptureProxyProducer",
#                # TCP port the capture proxy listens on for incoming HTTP(S) traffic. This port is exposed via the Kubernetes Service and used to construct the proxy endpoint URL.
#                "listenPort": "",  # number
##                # Maximum size in bytes for buffering a single HTTP request/response payload before forwarding to Kafka.
##                "maxTrafficBufferSize": 1048576,
##                # When true, the proxy forwards traffic to the source cluster without capturing it to Kafka. Useful for TLS termination or routing without traffic recording.
##                "noCapture": false,
##                # Number of Netty worker threads for the proxy to handle concurrent connections.
##                "numThreads": 1,
##                # TLS certificate configuration for HTTPS termination at the proxy. When configured, the proxy serves HTTPS and the TLS secret is mounted at /etc/proxy-tls/.
##                "tls": "",  # unknown
##                # Enable SASL/IAM authentication for the proxy's Kafka producer when connecting to Amazon MSK. Uses the pod's IAM role via EKS Pod Identity.
##                "enableMSKAuth": false,
##                # List of header patterns. Requests matching any of these header patterns will be forwarded but not captured to Kafka.
##                "suppressCaptureForHeaderMatch": [],  # string[]
##                # HTTP method to suppress from capture (e.g. 'HEAD'). Requests with this method are forwarded but not recorded.
##                "suppressCaptureForMethod": "",  # string
##                # URI path pattern to suppress from capture. Requests matching this path are forwarded but not recorded.
##                "suppressCaptureForUriPath": "",  # string
##                # Combined method and path pattern for capture suppression in 'METHOD /path' format.
##                "suppressMethodAndPath": ""  # string
#            }
#        }
#    },
#    # Map of replayer names to their replay configurations. Each replayer consumes from a proxy's Kafka topic and replays to a target cluster.
#    "replayers": {
#        # Configuration for a single traffic replayer instance, binding a proxy's captured traffic to a target cluster.
#        "REPLAYER": {
#            # Name of the capture proxy to replay traffic from. Must match a key in traffic.proxies.
#            "fromProxy": "",  # string
#            # Name of the target cluster to replay traffic to. Must match a key in targetClusters.
#            "toTarget": "",  # string
##            # List of snapshot migrations that must complete before this replayer starts. Ensures data consistency when replaying traffic that depends on backfilled data.
##            "dependsOnSnapshotMigrations": [
##                {
##                # Name of the source cluster. Must match a key in sourceClusters.
##                "source": "",  # string
##                # Name of the snapshot. Must match a key in the source cluster's snapshotInfo.snapshots.
##                "snapshot": ""  # string
##                }
##            ],
##            # Optional replayer configuration overrides. If omitted, replayer runs with schema defaults.
##            "replayerConfig": {
###                # Additional JVM arguments passed via JDK_JAVA_OPTIONS (e.g. '-Xmx4g -XX:+UseG1GC'). If setting -Xmx, ensure the heap size is smaller than the container's memory limit in resources to account for off-heap memory usage.
###                "jvmArgs": "",  # string
###                # Name of a Kubernetes ConfigMap containing a custom Log4j2 properties configuration. The ConfigMap should have a single key whose value is the Log4j2 properties file content. When set, it is mounted into the container and passed via -Dlog4j2.configurationFile. See https://logging.apache.org/log4j/2.x/manual/configuration.html#properties for format reference.
###                "loggingConfigurationOverrideConfigMap": "",  # string
###                # Number of replayer pod replicas in the Kubernetes Deployment. Each replica independently consumes from Kafka and replays traffic to the target.
###                "podReplicas": 1,
##                # Kubernetes resource limits and requests for the replayer container. Partial overrides are deep-merged with the built-in defaults. By default, limits equal requests, giving the pod 'Guaranteed' QoS (least likely to be evicted). Setting requests lower than limits results in 'Burstable' QoS, allowing the pod to use less resources when idle but burst up to the limit.
##                "resources": {
##                    # Maximum resource limits for the container. The container will be terminated if it exceeds these limits.
##                    "limits": {
##                        # CPU allocation for the container in Kubernetes millicores.
##                        "cpu": "",  # string
##                        # Memory allocation for the container.
##                        "memory": "",  # string
###                        # Ephemeral storage allocation for the container. Used for temporary on-disk data such as Lucene index segments during RFS document migration.
###                        "ephemeral-storage": ""  # string
##                    },
##                    # Minimum guaranteed resources for the container. Used by the Kubernetes scheduler for pod placement.
##                    "requests": {
##                        # CPU allocation for the container in Kubernetes millicores.
##                        "cpu": "",  # string
##                        # Memory allocation for the container.
##                        "memory": "",  # string
###                        # Ephemeral storage allocation for the container. Used for temporary on-disk data such as Lucene index segments during RFS document migration.
###                        "ephemeral-storage": ""  # string
##                    }
##                },
###                # Enable SASL/IAM authentication for the replayer's Kafka consumer when connecting to Amazon MSK. Uses the pod's IAM role via EKS Pod Identity.
###                "kafkaTrafficEnableMSKAuth": false,
###                # [Expert] Path to a Java properties file with additional or overridden Kafka consumer configuration. The file must be mounted into the container by the user (e.g. via Kyverno pod mutation or custom image). Not wired through the workflow by default.
###                "kafkaTrafficPropertyFile": "",  # string
###                # Number of seconds of captured traffic to buffer ahead of the current replay position. Must be strictly greater than observedPacketConnectionTimeout. Larger values improve throughput but increase memory usage.
###                "lookaheadTimeSeconds": 400,
###                # Maximum number of HTTP requests that can be in-flight simultaneously to the target cluster. Limits concurrency to prevent overwhelming the target.
###                "maxConcurrentRequests": 10000,
###                # Number of threads used to send replayed requests to the target. 0 uses the Netty event loop (typically number of available processors).
###                "numClientThreads": 0,
###                # List of document-level exception types that should not be retried during bulk replay. These errors still count as failures in the output but are not retried because they are deterministic client or mapping errors that will produce the same result on every attempt. When omitted, defaults to a built-in set including version_conflict_engine_exception, mapper_parsing_exception, strict_dynamic_mapping_exception, and others. Set explicitly to override the defaults entirely (not additive). Common values: version_conflict_engine_exception, mapper_parsing_exception, illegal_argument_exception, resource_already_exists_exception.
###                "nonRetryableDocExceptionTypes": [],  # string[]
###                # Seconds of inactivity on a captured connection before assuming it was terminated in the original traffic stream. Must be strictly less than lookaheadTimeSeconds.
###                "observedPacketConnectionTimeout": 360,
###                # URL for the OpenTelemetry Collector endpoint used for metrics and traces (e.g. 'http://otel-collector:4317').
###                "otelCollectorEndpoint": "http://otel-collector:4317",
###                # Milliseconds to delay the first request on a resumed connection after a Kafka partition reassignment. Prevents request bursts during rebalancing.
###                "quiescentPeriodMs": 5000,
###                # Remove the Authorization header from replayed requests without replacing it. Useful when the target uses a different auth mechanism (e.g. SigV4) configured separately.
###                "removeAuthHeader": false,
###                # Multiplier to accelerate replay timing relative to the original captured traffic. 1.0 = real-time, 2.0 = double speed.
###                "speedupFactor": 1.1,
###                # Maximum seconds to wait for a response from the target cluster before timing out a replayed request.
###                "targetServerResponseTimeoutSeconds": 150,
###                # Inline request transformer configuration as a JSON string. Request transformers modify each captured HTTP request before it is replayed to the target cluster.
###                "transformerConfig": "",  # string
###                # Base64-encoded request transformer configuration, for configurations that would be cumbersome to otherwise encode in a JSON field. Request transformers modify each captured HTTP request before it is replayed to the target cluster.
###                "transformerConfigEncoded": "",  # string
###                # Path to a JSON file containing request transformer configuration. Request transformers modify each captured HTTP request before it is replayed to the target cluster. [Expert] The file must be mounted into the container by the user (e.g. via Kyverno pod mutation or custom image). Not wired through the workflow by default.
###                "transformerConfigFile": "",  # string
###                # Inline tuple transformer configuration as a JSON string. Tuple transformers operate on request-response pairs, enabling stateful transformations that depend on the source cluster's response.
###                "tupleTransformerConfig": "",  # string
###                # Base64-encoded tuple transformer configuration. Tuple transformers operate on request-response pairs, enabling stateful transformations that depend on the source cluster's response.
###                "tupleTransformerConfigBase64": "",  # string
###                # Path to a JSON file containing tuple transformer configuration. Tuple transformers operate on request-response pairs, enabling stateful transformations that depend on the source cluster's response. [Expert] The file must be mounted into the container by the user (e.g. via Kyverno pod mutation or custom image). Not wired through the workflow by default.
###                "tupleTransformerConfigFile": "",  # string
###                # Maximum seconds before rotating/uploading a tuple file to S3.
###                "tupleMaxBufferSeconds": 60,
###                # Maximum uncompressed size in MB before rotating a tuple file to S3.
###                "tupleMaxFileSizeMb": 256,
###                # String appended to the User-Agent header on all replayed requests to the target cluster. Useful for identifying replayed traffic in target cluster logs.
###                "userAgent": ""  # string
##            }
#        }
#    }
#}
}


```

</details>

### DIFF
--- a/orchestrationSpecs/packages/schemas/src/schemaUtilities.ts
+++ b/orchestrationSpecs/packages/schemas/src/schemaUtilities.ts
@@ -106,7 +106,13 @@ export function unwrapSchema(schema: z.ZodTypeAny, constructorNames = ZOD_OPTION
             return unwrapSchema(innerType as z.ZodTypeAny, constructorNames);
         }
     } else if (schema instanceof z.ZodPipe) {
-        const inner = schema._def.in;
+        // ZodPipe is used for both z.preprocess(fn, schema) and schema.transform(fn) / schema.pipe().
+        // - For z.preprocess: _def.in is the ZodTransform (the preprocess fn), _def.out is the real schema.
+        // - For .transform / .pipe: _def.in is the source schema, _def.out is the ZodTransform.
+        // In both cases we want the non-transform side so callers see the structured schema.
+        const d = schema._def as any;
+        const inIsTransform = d.in?.constructor?.name === 'ZodTransform';
+        const inner = inIsTransform ? d.out : d.in;
         return unwrapSchema(inner as any, constructorNames);
     }
 
@@ -150,5 +156,40 @@ export function fullUnwrapType<T extends z.ZodTypeAny>(schema: T) {
         return actualType;
     } else {
         return schema;
+    }
+}
+
+/**
+ * Walk through wrapper schemas (ZodOptional, ZodDefault, ZodNullable, ZodPipe) and
+ * return the first non-empty `.description` found on any level.
+ *
+ * `.describe()` can be called anywhere in the construction chain, e.g.:
+ *   z.string().default("").optional().describe("DOC")   -> description on ZodOptional
+ *   z.string().optional().default("").describe("DOC")   -> description on ZodDefault
+ *   z.string().describe("DOC").default("").optional()   -> description on ZodString
+ *   z.preprocess(fn, inner).describe("DOC").default(d)  -> description on ZodPipe
+ *
+ * fullUnwrapType() only returns the innermost concrete type, so callers reading
+ * `.description` off it miss descriptions attached to any outer wrapper.
+ * Use this helper instead when you want the field's authored documentation.
+ */
+export function getDescription(schema: z.ZodTypeAny): string | undefined {
+    let cur: z.ZodTypeAny = schema;
+    while (true) {
+        if (cur.description) return cur.description;
+        if (cur instanceof z.ZodOptional) {
+            cur = cur.unwrap() as z.ZodTypeAny;
+        } else if (cur instanceof z.ZodDefault) {
+            cur = cur.removeDefault() as z.ZodTypeAny;
+        } else if (cur instanceof z.ZodNullable) {
+            cur = cur.unwrap() as z.ZodTypeAny;
+        } else if (cur instanceof z.ZodPipe) {
+            // Match unwrapSchema()'s preprocess-aware behaviour: skip over the
+            // ZodTransform side and continue walking through the real schema.
+            const d = cur._def as any;
+            cur = (d.in?.constructor?.name === 'ZodTransform' ? d.out : d.in) as z.ZodTypeAny;
+        } else {
+            return undefined;
+        }
     }
 }
--- a/orchestrationSpecs/packages/schemas/src/schemaUtilities.ts
+++ b/orchestrationSpecs/packages/schemas/src/schemaUtilities.ts
@@ -110,10 +110,9 @@ export function unwrapSchema(schema: z.ZodTypeAny, constructorNames = ZOD_OPTION
         // - For z.preprocess: _def.in is the ZodTransform (the preprocess fn), _def.out is the real schema.
         // - For .transform / .pipe: _def.in is the source schema, _def.out is the ZodTransform.
         // In both cases we want the non-transform side so callers see the structured schema.
-        const d = schema._def as any;
-        const inIsTransform = d.in?.constructor?.name === 'ZodTransform';
-        const inner = inIsTransform ? d.out : d.in;
-        return unwrapSchema(inner as any, constructorNames);
+        const d = schema._def;
+        const inner = d.in instanceof z.ZodTransform ? d.out : d.in;
+        return unwrapSchema(inner as z.ZodTypeAny, constructorNames);
     }
 
     return schema;
@@ -131,7 +130,12 @@ export function fullUnwrapType<T extends z.ZodTypeAny>(schema: T) {
             const def = (actualType as any).def || (actualType as any)._def;
             const typeName = def?.typeName || def?.type;
             if (actualType instanceof z.ZodPipe) {
-                actualType = actualType._def.in;
+                // Match unwrapSchema()/getDescription(): z.preprocess puts the ZodTransform
+                // on _def.in, so unwrapping to .in would hand back the preprocess fn. Walk
+                // to _def.out in that case; keep .in for .transform()/.pipe() where the
+                // source schema is on .in and the ZodTransform is on .out.
+                const d = actualType._def;
+                actualType = d.in instanceof z.ZodTransform ? d.out : d.in;
             } else if (typeName === 'ZodEffects' || typeName === 'transform') {
                 // For transforms, get the input schema
                 if (typeof (actualType as any).innerType === 'function') {
@@ -186,8 +190,8 @@ export function getDescription(schema: z.ZodTypeAny): string | undefined {
         } else if (cur instanceof z.ZodPipe) {
             // Match unwrapSchema()'s preprocess-aware behaviour: skip over the
             // ZodTransform side and continue walking through the real schema.
-            const d = cur._def as any;
-            cur = (d.in?.constructor?.name === 'ZodTransform' ? d.out : d.in) as z.ZodTypeAny;
+            const d = cur._def;
+            cur = (d.in instanceof z.ZodTransform ? d.out : d.in) as z.ZodTypeAny;
         } else {
             return undefined;
         }
--- a/orchestrationSpecs/packages/schemas/src/makeSample.ts
+++ b/orchestrationSpecs/packages/schemas/src/makeSample.ts
@@ -1,6 +1,6 @@
 import { z } from 'zod';
 import {OVERALL_MIGRATION_CONFIG} from "./userSchemas";
-import {fullUnwrapType, unwrapSchema, ZOD_OPTIONAL_TYPES} from "./schemaUtilities";
+import {fullUnwrapType, getDescription, unwrapSchema, ZOD_OPTIONAL_TYPES} from "./schemaUtilities";
 
 // Path context for tracking descent through the schema
 export type SchemaPath = string[];
@@ -319,7 +319,7 @@ function schemaToJsonWithComments(
         keys.forEach((key, index) => {
             const fieldSchema = shape[key];
             const fieldCtx = extendPath(ctx, key);
-            const description = fullUnwrapType(fieldSchema).description;
+            const description = getDescription(fieldSchema);
             const typeName = getTypeName(fieldSchema);
             const optional = isOptional(fieldSchema);
             const nextIndent = currentIndent.incrementComment(optional);
--- a/orchestrationSpecs/packages/schemas/tests/schemaUtilities.test.ts
+++ b/orchestrationSpecs/packages/schemas/tests/schemaUtilities.test.ts
@@ -1,5 +1,5 @@
 import { z } from "zod";
-import { unwrapSchema, getDescription } from "../src/schemaUtilities";
+import { unwrapSchema, getDescription, fullUnwrapType } from "../src/schemaUtilities";
 
 describe("unwrapSchema through ZodPipe", () => {
     test("z.preprocess(fn, innerSchema) unwraps to the inner schema, not the transform", () => {
@@ -78,5 +78,30 @@ describe("getDescription walks wrapper chains", () => {
         // that's the most recently-applied / most specific override.
         const FIELD = z.string().describe("inner").default("").describe("outer");
         expect(getDescription(FIELD)).toBe("outer");
+    });
+});
+
+describe("fullUnwrapType through ZodPipe", () => {
+    test("z.preprocess(fn, innerSchema) wrapped in optional unwraps to the inner schema, not the transform", () => {
+        // Same latent bug as unwrapSchema/getDescription: z.preprocess puts the
+        // ZodTransform on _def.in, so naively picking .in would hand back the
+        // preprocess fn. fullUnwrapType must walk to _def.out in that case.
+        const INNER = z.object({
+            limits: z.object({ cpu: z.string() }).optional(),
+        });
+        const FIELD = z.preprocess((v) => v ?? {}, INNER).optional();
+
+        const unwrapped = fullUnwrapType(FIELD);
+        expect(unwrapped).toBeInstanceOf(z.ZodObject);
+    });
+
+    test("schema.transform(fn) wrapped in optional unwraps to the source schema", () => {
+        // For .transform() the source schema is on .in and the ZodTransform is
+        // on .out -- the existing .in branch is correct here. This test pins
+        // that the preprocess fix did not regress the transform direction.
+        const FIELD = z.string().transform((s) => s.toUpperCase()).optional();
+
+        const unwrapped = fullUnwrapType(FIELD);
+        expect(unwrapped).toBeInstanceOf(z.ZodString);
     });
 });
--- a/orchestrationSpecs/packages/schemas/tests/schemaUtilities.test.ts
+++ b/orchestrationSpecs/packages/schemas/tests/schemaUtilities.test.ts
@@ -1,0 +1,82 @@
+import { z } from "zod";
+import { unwrapSchema, getDescription } from "../src/schemaUtilities";
+
+describe("unwrapSchema through ZodPipe", () => {
+    test("z.preprocess(fn, innerSchema) unwraps to the inner schema, not the transform", () => {
+        // This mirrors the RESOURCE_REQUIREMENTS pattern in userSchemas.ts where
+        // z.preprocess is used to deep-merge defaults into a structured schema.
+        // Regression: before the fix, unwrapSchema returned the ZodTransform
+        // (preprocess fn), which caused the sample-config generator to render
+        // "resources": "",  # unknown  instead of the nested limits/requests object.
+        const INNER = z.object({
+            limits: z.object({ cpu: z.string(), memory: z.string() }).optional(),
+            requests: z.object({ cpu: z.string(), memory: z.string() }).optional(),
+        });
+        const PREPROCESSED = z.preprocess((v) => v, INNER);
+
+        const unwrapped = unwrapSchema(PREPROCESSED);
+        expect(unwrapped).toBeInstanceOf(z.ZodObject);
+    });
+
+    test("schema.transform(fn) unwraps to the source schema, not the transform", () => {
+        const SRC = z.string();
+        const TRANSFORMED = SRC.transform((s) => s.toUpperCase());
+
+        const unwrapped = unwrapSchema(TRANSFORMED);
+        expect(unwrapped).toBeInstanceOf(z.ZodString);
+    });
+
+    test("schema.pipe(otherSchema) unwraps to the source schema", () => {
+        const SRC = z.string();
+        const PIPED = SRC.pipe(z.string().min(1));
+
+        const unwrapped = unwrapSchema(PIPED);
+        expect(unwrapped).toBeInstanceOf(z.ZodString);
+    });
+});
+
+describe("getDescription walks wrapper chains", () => {
+    // Regression: .describe() can be attached anywhere in the construction
+    // chain. fullUnwrapType() only returns the innermost concrete schema, so
+    // the renderer was reading .description off a ZodString/ZodNumber that
+    // never carried the description, producing sample-config entries with no
+    // "# description" comment. getDescription() walks the whole chain.
+
+    test("description on outer ZodOptional (.default().optional().describe())", () => {
+        // Pattern used throughout userSchemas.ts for jvmArgs/podReplicas/etc.
+        const FIELD = z.string().default("").optional().describe("JVM args to append");
+        expect(getDescription(FIELD)).toBe("JVM args to append");
+    });
+
+    test("description on ZodDefault (.optional().default().describe())", () => {
+        const FIELD = z.string().optional().default("").describe("described default");
+        expect(getDescription(FIELD)).toBe("described default");
+    });
+
+    test("description on innermost concrete type (.describe().default().optional())", () => {
+        const FIELD = z.string().describe("innermost").default("").optional();
+        expect(getDescription(FIELD)).toBe("innermost");
+    });
+
+    test("description on ZodPipe from z.preprocess().describe().default()", () => {
+        // Pattern used for the `resources` fields in userSchemas.ts where
+        // the describe sits between the preprocess pipe and the default.
+        const INNER = z.object({ limits: z.object({ cpu: z.string() }).optional() });
+        const FIELD = z.preprocess((v) => v ?? {}, INNER)
+            .describe("resource overrides")
+            .default({ limits: { cpu: "1" } });
+        expect(getDescription(FIELD)).toBe("resource overrides");
+    });
+
+    test("returns undefined when no description is attached anywhere", () => {
+        const FIELD = z.string().default("").optional();
+        expect(getDescription(FIELD)).toBeUndefined();
+    });
+
+    test("returns the OUTERMOST description when multiple are attached", () => {
+        // If the author describes at several layers, the outermost wins --
+        // that's the most recently-applied / most specific override.
+        const FIELD = z.string().describe("inner").default("").describe("outer");
+        expect(getDescription(FIELD)).toBe("outer");
+    });
+});


orchestrationSpecs/packages/schemas/src/schemaUtilities.ts
jugal-chauhan, line 139: Should this also be treated the same way ? 
```
if (actualType instanceof z.ZodPipe) {
    const d = actualType._def as any;
    actualType = d.in?.constructor?.name === 'ZodTransform' ? d.out : d.in;
```

AndreKurait: Good catch — yes, that branch has the same latent bug (`_def.in` is the `ZodTransform` when the schema was built with `z.preprocess`). Aligned `fullUnwrapType` with `unwrapSchema` / `getDescription` in 309b8ec32 and added two regression tests (preprocess and transform directions).

`fullUnwrapType` happens to be imported but not invoked from `makeSample.ts` today, so this isn't an observable bug — purely a consistency fix so all three ZodPipe traversals handle preprocess the same way.

AndreKurait: Good call on the `as any` — dropped it. After `schema instanceof z.ZodPipe` narrows, `_def` is already typed as the $ZodPipeDef with typed `.in` / `.out`, so the cast was unnecessary. Also swapped the `constructor?.name === 'ZodTransform'` string sniff for `d.in instanceof z.ZodTransform` since we now have real types, and applied the same cleanup to the other two ZodPipe branches (`unwrapSchema` and `getDescription`) for consistency.

Pushed as 5885d014a.

jugal-chauhan (approved):
Thanks for brining in these changes and addressing the comments

PR #2852, opened 2026-04-30
Adding field mappings to shim and using term query for exact fields b…

nagarajg17:
…ased on field mappings

### Description
In the read paths (shim), we don’t have Solr collections field metadata info. Having this info help us to transform certain Solr queries correctly to OpenSearch. Ex: Let’s say we’ve query `q=category:electronics`, if category field is string (keyword in Opensearch) we need to use [term query](https://docs.opensearch.org/latest/query-dsl/term/term/), if it is text_category field then we need to use [match query](https://docs.opensearch.org/latest/query-dsl/full-text/match/). Currently since we don’t have field metadata info we are using match query for all cases as safer option. This can produce search results count difference since match query analyses fields compared to term query. So to solve this problem, we need to have field metadata info in transformation layer to use appropriate query types based on field types. We are Injecting field type metadata via bindings (same pattern as solrConfig): The Java side that configures the shim would need to supply a fieldTypes map in bindings (e.g., { “title”: “text”, “status”: “string”, “id”: “string” }). This will come from the operator providing it as config when setting up the shim

### Check List

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/orchestrator/translateQ.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/orchestrator/translateQ.ts
@@ -80,14 +80,21 @@ function passthroughDsl(query: string): Map<string, any> {
  *                 from this map and passes the full map to the parser
  *                 for configuration (e.g., `df`, `qf`, `pf`).
  * @param mode - 'fail-fast' (default) or 'passthrough-on-error'.
+ * @param fieldTypes - Optional map of field name → Solr type from the schema.
+ *                     Used by fieldRule to choose term vs match. Defaults to empty map
+ *                     (all plain field:value queries use match).
  *
  * Example:
  *   translateQ(new Map([['q', 'title:java'], ['df', 'content']]))
  *   → { dsl: Map{"match" → Map{"title" → "java"}}, warnings: [] }
+ *
+ *   translateQ(new Map([['q', 'status:active']]), 'fail-fast', new Map([['status', 'keyword']]))
+ *   → { dsl: Map{"term" → Map{"status" → "active"}}, warnings: [] }
  */
 export function translateQ(
   params: ReadonlyMap<string, string>,
   mode: TranslationMode = 'fail-fast',
+  fieldTypes: ReadonlyMap<string, string> = new Map(),
 ): TranslateResult {
   const query = params.get('q') || '*:*';
 
@@ -112,7 +119,7 @@ export function translateQ(
 
   // Stage 2: Transform
   try {
-    const mainDsl = transformNode(ast);
+    const mainDsl = transformNode(ast, fieldTypes);
 
     // Stage 3: pf phrase boost — wraps the main query in bool.must + bool.should
     // if pf is set. Delegated to phraseBoost.ts (transformer concern).
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.test.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.test.ts
@@ -1,12 +1,13 @@
 import { describe, it, expect, vi } from 'vitest';
 import { fieldRule } from './fieldRule';
+import { translateQ } from '../../orchestrator/translateQ';
 import type { FieldNode } from '../../ast/nodes';
 
 /** Stub transformChild — not used by fieldRule but required by signature. */
 const stubTransformChild = () => new Map();
 
 describe('fieldRule', () => {
-  // --- Plain match queries ---
+  // --- Plain match queries (no fieldTypes / unknown field) ---
 
   it.each([
     { field: 'title', value: 'java', desc: 'simple field:value' },
@@ -14,7 +15,7 @@ describe('fieldRule', () => {
     { field: '_text_', value: 'search', desc: 'underscore field name' },
     { field: 'metadata.author', value: 'doe', desc: 'dotted field name' },
     { field: 'title', value: 'foo-bar_baz.v1@test#123', desc: 'multiple special chars' },
-  ])('transforms $desc to match query', ({ field, value }) => {
+  ])('transforms $desc to match query when no fieldTypes provided', ({ field, value }) => {
     const node: FieldNode = { type: 'field', field, value };
 
     const result = fieldRule(node, stubTransformChild);
@@ -22,6 +23,88 @@ describe('fieldRule', () => {
     expect(result).toEqual(
       new Map([['match', new Map([[field, new Map([['query', value]])]])]]),
     );
+  });
+
+  it('uses match query for unknown field even when fieldTypes map is provided', () => {
+    const node: FieldNode = { type: 'field', field: 'unknown_field', value: 'foo' };
+    const fieldTypes = new Map([['status', 'solr.StrField']]);
+
+    const result = fieldRule(node, stubTransformChild, fieldTypes);
+
+    expect(result).toEqual(
+      new Map([['match', new Map([['unknown_field', new Map([['query', 'foo']])]])]]),
+    );
+  });
+
+  // --- term queries when fieldTypes identifies non-text fields ---
+
+  it.each([
+    { fieldTypeClass: 'solr.StrField',         desc: 'StrField (string)' },
+    { fieldTypeClass: 'solr.IntPointField',     desc: 'IntPointField (integer)' },
+    { fieldTypeClass: 'solr.LongPointField',    desc: 'LongPointField (long)' },
+    { fieldTypeClass: 'solr.FloatPointField',   desc: 'FloatPointField (float)' },
+    { fieldTypeClass: 'solr.DoublePointField',  desc: 'DoublePointField (double)' },
+    { fieldTypeClass: 'solr.DatePointField',    desc: 'DatePointField (date)' },
+    { fieldTypeClass: 'solr.BoolField',         desc: 'BoolField (boolean)' },
+    { fieldTypeClass: 'solr.UUIDField',         desc: 'UUIDField (keyword)' },
+    { fieldTypeClass: 'com.example.CustomExactField', desc: 'custom non-text class' },
+  ])('emits term query for $desc', ({ fieldTypeClass }) => {
+    const node: FieldNode = { type: 'field', field: 'status', value: 'active' };
+    const fieldTypes = new Map([['status', fieldTypeClass]]);
+
+    const result = fieldRule(node, stubTransformChild, fieldTypes);
+
+    expect(result).toEqual(
+      new Map([['term', new Map([['status', new Map([['value', 'active']])]])]]),
+    );
+  });
+
+  // --- match queries when fieldTypes identifies text fields ---
+
+  it.each([
+    { fieldTypeClass: 'solr.TextField',                          desc: 'solr.TextField' },
+    { fieldTypeClass: 'org.apache.solr.schema.TextField',        desc: 'fully-qualified TextField' },
+    { fieldTypeClass: 'com.example.CustomTextField',             desc: 'custom class containing TextField' },
+  ])('emits match query for $desc', ({ fieldTypeClass }) => {
+    const node: FieldNode = { type: 'field', field: 'title', value: 'java' };
+    const fieldTypes = new Map([['title', fieldTypeClass]]);
+
+    const result = fieldRule(node, stubTransformChild, fieldTypes);
+
+    expect(result).toEqual(
+      new Map([['match', new Map([['title', new Map([['query', 'java']])]])]]),
+    );
+  });
+
+  it('fieldTypes does not affect existence search — always emits exists', () => {
+    const node: FieldNode = { type: 'field', field: 'status', value: '*' };
+    const fieldTypes = new Map([['status', 'solr.StrField']]);
+
+    const result = fieldRule(node, stubTransformChild, fieldTypes);
+
+    expect(result).toEqual(
+      new Map([['exists', new Map([['field', 'status']])]]),
+    );
+  });
+
+  it('fieldTypes does not affect wildcard search — always emits wildcard', () => {
+    const node: FieldNode = { type: 'field', field: 'status', value: 'act*' };
+    const fieldTypes = new Map([['status', 'solr.StrField']]);
+
+    const result = fieldRule(node, stubTransformChild, fieldTypes);
+
+    expect(result).toEqual(
+      new Map([['wildcard', new Map([['status', new Map([['value', 'act*']])]])]]),
+    );
+  });
+
+  it('fieldTypes does not affect fuzzy search — always emits fuzzy', () => {
+    const node: FieldNode = { type: 'field', field: 'status', value: 'activ~' };
+    const fieldTypes = new Map([['status', 'solr.StrField']]);
+
+    const result = fieldRule(node, stubTransformChild, fieldTypes);
+
+    expect(result.keys().next().value).toBe('fuzzy');
   });
 
   // --- Existence search ---
@@ -72,7 +155,7 @@ describe('fieldRule', () => {
     const result = fieldRule(node, stubTransformChild);
 
     expect(result).toEqual(
-      new Map([['fuzzy', new Map([['title', new Map([['value', 'roam'], ['fuzziness', 2]])]])]]),
+      new Map([['fuzzy', new Map([['title', new Map<string, any>([['value', 'roam'], ['fuzziness', 2]])]])]]),
     );
   });
 
@@ -82,7 +165,7 @@ describe('fieldRule', () => {
     const result = fieldRule(node, stubTransformChild);
 
     expect(result).toEqual(
-      new Map([['fuzzy', new Map([['title', new Map([['value', 'roam'], ['fuzziness', 1]])]])]]),
+      new Map([['fuzzy', new Map([['title', new Map<string, any>([['value', 'roam'], ['fuzziness', 1]])]])]]),
     );
   });
 
@@ -92,7 +175,7 @@ describe('fieldRule', () => {
     const result = fieldRule(node, stubTransformChild);
 
     expect(result).toEqual(
-      new Map([['fuzzy', new Map([['title', new Map([['value', 'roam'], ['fuzziness', 0]])]])]]),
+      new Map([['fuzzy', new Map([['title', new Map<string, any>([['value', 'roam'], ['fuzziness', 0]])]])]]),
     );
   });
 
@@ -102,7 +185,7 @@ describe('fieldRule', () => {
     const result = fieldRule(node, stubTransformChild);
 
     expect(result).toEqual(
-      new Map([['fuzzy', new Map([['title', new Map([['value', 'roam'], ['fuzziness', 2]])]])]]),
+      new Map([['fuzzy', new Map([['title', new Map<string, any>([['value', 'roam'], ['fuzziness', 2]])]])]]),
     );
   });
 
@@ -153,5 +236,63 @@ describe('fieldRule', () => {
     fieldRule(node, spy);
 
     expect(spy).not.toHaveBeenCalled();
+  });
+
+  // --- fieldTypes propagates through bool queries ---
+
+  it.each([
+    { fieldTypeClass: 'solr.StrField',         desc: 'StrField (string)' },
+    { fieldTypeClass: 'solr.IntPointField',     desc: 'IntPointField (integer)' },
+    { fieldTypeClass: 'solr.LongPointField',    desc: 'LongPointField (long)' },
+    { fieldTypeClass: 'solr.FloatPointField',   desc: 'FloatPointField (float)' },
+    { fieldTypeClass: 'solr.DoublePointField',  desc: 'DoublePointField (double)' },
+    { fieldTypeClass: 'solr.DatePointField',    desc: 'DatePointField (date)' },
+    { fieldTypeClass: 'solr.BoolField',         desc: 'BoolField (boolean)' },
+    { fieldTypeClass: 'solr.UUIDField',         desc: 'UUIDField (keyword)' },
+  ])('term query is used for $desc field via translateQ', ({ fieldTypeClass }) => {
+    const fieldTypes = new Map([['status', fieldTypeClass]]);
+    const params = new Map([['q', 'status:active']]);
+
+    const { dsl } = translateQ(params, 'fail-fast', fieldTypes);
+
+    expect(dsl).toEqual(new Map([['term', new Map([['status', new Map([['value', 'active']])]])]]));
+  });
+
+  it('term query propagates through AND bool node', () => {
+    const fieldTypes = new Map([
+      ['status', 'solr.StrField'],
+      ['category', 'solr.StrField'],
+    ]);
+    const params = new Map([['q', 'status:active AND category:books']]);
+
+    const { dsl } = translateQ(params, 'fail-fast', fieldTypes);
+
+    expect(dsl).toEqual(
+      new Map([['bool', new Map([
+        ['must', [
+          new Map([['term', new Map([['status', new Map([['value', 'active']])]])]]),
+          new Map([['term', new Map([['category', new Map([['value', 'books']])]])]]),
+        ]],
+      ])]]),
+    );
+  });
+
+  it('term and match queries mixed in OR bool node', () => {
+    const fieldTypes = new Map([
+      ['status',   'solr.StrField'],   // exact → term
+      ['title',    'solr.TextField'],  // analyzed → match
+    ]);
+    const params = new Map([['q', 'status:active OR title:java']]);
+
+    const { dsl } = translateQ(params, 'fail-fast', fieldTypes);
+
+    expect(dsl).toEqual(
+      new Map([['bool', new Map([
+        ['should', [
+          new Map([['term',  new Map([['status', new Map([['value', 'active']])]])]]),
+          new Map([['match', new Map([['title',  new Map([['query', 'java']])]])]]),
+        ]],
+      ])]]),
+    );
   });
 });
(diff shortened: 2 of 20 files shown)


solrMigrationDevSandbox/docker-compose.yml
```diff
@@ -61,6 +61,7 @@ services:
     volumes:
       - transform-dist:/transforms:ro
       - ./config/solr/solrconfig.xml:/solr-config/solrconfig.xml:ro
+      - ./config/solr/schema.xml:/solr-config/managed-schema.xml:ro
```
k-rooot, line 64: Providing the solr schema via volume mount works well in a sandbox environment. However, since the shim will be distributed as a docker image hosted on AWS ECR for external customer use, we need to rethink how customers supply the solr schema as input.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/orchestrator/translateQ.ts
```diff
@@ -80,14 +80,21 @@ function passthroughDsl(query: string): Map<string, any> {
  *                 from this map and passes the full map to the parser
  *                 for configuration (e.g., `df`, `qf`, `pf`).
  * @param mode - 'fail-fast' (default) or 'passthrough-on-error'.
+ * @param fieldTypes - Optional map of field name → OpenSearch type from the schema.
```
k-rooot: 1. Solr schema would contain the field types for Solr not Opensearch right?
2. Can there be scenarios where customer have opted for the opposite configuration in OpenSearch?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -5,22 +5,32 @@
...
+ *   - field:value (plain, text field) → match query
+ *   - field:value (plain, any other type or unknown) → term query
  *
- * We use `match` for plain values because Solr's standard query parser
- * analyzes field values at query time, matching OpenSearch's match behavior.
+ * For plain values, the query type depends on the field's OpenSearch type from
+ * the schema (passed via fieldTypes):
+ *   - `text` → match query (analyzed, full-text search)
+ *   - anything else (keyword, integer, date, boolean, etc.) → term query (exact match)
+ *   - unknown (field not in fieldTypes) → match query (safe default — avoids breaking
```
k-rooot, line 15: This creates ambiguity between unknown vs non-text.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -32,9 +42,30 @@ const WILDCARD_PATTERN = /[*?]/;
...
+
+/**
+ * TransformRuleFn for FieldNode. Uses the optional `fieldTypes` parameter
+ * (field name → Solr fieldType class) to choose term vs match for plain values.
+ * Falls back to match when fieldTypes is absent or the field is not listed.
+ */
 export const fieldRule: TransformRuleFn = (
   node: ASTNode,
   _transformChild,
+  fieldTypes = new Map<string, string>(),
```
k-rooot: A new map is created on every invocation, even when unused.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -64,6 +95,19 @@ export const fieldRule: TransformRuleFn = (
...
+  //
+  // class.includes("TextField") → analyzed → match query.
+  // Everything else (solr.StrField, solr.IntPointField, etc.) → exact → term query.
+  // Unknown fields (not in fieldTypes) fall back to match: safer than term because
+  // a match query on an exact field degrades gracefully, whereas a term query on
+  // a text field misses analyzed tokens (e.g. won't find "java" if stored as "Java").
+  const fieldTypeClass = fieldTypes.get(field);
+  if (fieldTypeClass !== undefined && !isTextField(fieldTypeClass)) {
+    // Known non-text field: term query — exact match, no analysis.
+    return new Map([['term', new Map([[field, value]])]]);
```
k-rooot: All other queries use expanded form:
```
{ field: { param: value } }
```

But term uses a shorthand: `{ field: value }`

Can we maintain consistency?

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -32,9 +42,30 @@ const WILDCARD_PATTERN = /[*?]/;
 /** Detects fuzzy search: term~ or term~N at end of value */
 const FUZZY_PATTERN = /^(.+?)~(\d?)$/;
```
k-rooot, line 43: nit: `apple~a` edge case silently falls through

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
```diff
@@ -64,6 +95,19 @@ export const fieldRule: TransformRuleFn = (
     return new Map([['wildcard', new Map([[field, new Map([['value', value]])]])]]);
   }
 
-  // Plain value: field:value → match query (expanded form for boost compatibility)
+  // Plain value — choose query type based on field's Solr fieldType class.
+  //
+  // class.includes("TextField") → analyzed → match query.
+  // Everything else (solr.StrField, solr.IntPointField, etc.) → exact → term query.
+  // Unknown fields (not in fieldTypes) fall back to match: safer than term because
+  // a match query on an exact field degrades gracefully, whereas a term query on
```
k-rooot: This is not always true right. `match` on keyword fields may behave differently depending on analyzer config.

A match query on a non analysed (keyword) field does not behave like a term query it still applies query time analysis, which can change semantics.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.test.ts
```diff
@@ -71,39 +154,43 @@ describe('fieldRule', () => {
...
     const node: FieldNode = { type: 'field', field: 'title', value: 'roam~1' };
 
     const result = fieldRule(node, stubTransformChild);
 
-    expect(result).toEqual(
-      new Map([['fuzzy', new Map([['title', new Map([['value', 'roam'], ['fuzziness', 1]])]])]]),
-    );
+    const fuzzyMap = result.get('fuzzy') as Map<string, any>;
+    const params = fuzzyMap.get('title') as Map<string, any>;
+    expect(params.get('value')).toBe('roam');
```
k-rooot: You no longer validate that only expected keys exist. A regression here could silently add extra fields.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.test.ts
```diff
@@ -154,4 +241,31 @@ describe('fieldRule', () => {
 
     expect(spy).not.toHaveBeenCalled();
   });
+
+  // --- fieldTypes propagates through bool queries ---
```
k-rooot, line 241: Known text field via fieldTypes scenario can be included in the test coverage

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
```diff
@@ -18,10 +18,20 @@ const solrConfig = (typeof bindings !== 'undefined' && bindings?.solrConfig) //N
...
   : undefined;
 
+// Read fieldTypes from bindings once at init. Java provides a flat map of
+// fieldName → solrTypeClass (e.g. {"title":"solr.TextField","id":"solr.StrField"})
+// resolved from managed-schema.xml via solrSchemaXmlFile config.
+// Empty map when solrSchemaXmlFile is not configured — fieldRule falls back to match.
+const fieldTypes: ReadonlyMap<string, string> =
+  (typeof bindings !== 'undefined' && bindings?.fieldTypes) //NOSONAR — typeof required for undeclared closure var
+    ? new Map(Object.entries(bindings.fieldTypes as Record<string, string>))
+    : new Map();
```
k-rooot: nit: You always allocate a new map even when fieldTypes is absent.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
```diff
@@ -18,10 +18,20 @@ const solrConfig = (typeof bindings !== 'undefined' && bindings?.solrConfig) //N
   ? bindings.solrConfig
   : undefined;
 
+// Read fieldTypes from bindings once at init. Java provides a flat map of
+// fieldName → solrTypeClass (e.g. {"title":"solr.TextField","id":"solr.StrField"})
+// resolved from managed-schema.xml via solrSchemaXmlFile config.
+// Empty map when solrSchemaXmlFile is not configured — fieldRule falls back to match.
+const fieldTypes: ReadonlyMap<string, string> =
+  (typeof bindings !== 'undefined' && bindings?.fieldTypes) //NOSONAR — typeof required for undeclared closure var
+    ? new Map(Object.entries(bindings.fieldTypes as Record<string, string>))
```
k-rooot: This assumes fieldTypes is a plain object and values are strings. If the java send a null or non string value, this will produce incorrect mappings.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
```diff
@@ -18,10 +18,20 @@ const solrConfig = (typeof bindings !== 'undefined' && bindings?.solrConfig) //N
   ? bindings.solrConfig
   : undefined;
 
+// Read fieldTypes from bindings once at init. Java provides a flat map of
```
k-rooot, line 51: Do we need test coverage for this change?

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+     * {@code fieldName → fieldTypeClass} map.
+     *
+     * <p>Returns an empty map if the path is null, the file does not exist, or parsing fails.
+     *
+     * @param path path to {@code managed-schema.xml} or {@code managed-schema}
+     * @return immutable map of field name to Solr fieldType Java class string
+     */
+    public static Map<String, String> fromXmlFile(Path path) {
+        if (path == null || !Files.exists(path)) {
+            log.warn("managed-schema.xml not found at {}, skipping fieldTypes", path);
```
k-rooot: If path == null, the log will print `managed-schema.xml not found at null`

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+            return Map.of();
+        }
+        try {
+            var factory = DocumentBuilderFactory.newInstance();
+            factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
```
k-rooot, line 66: If the intent is to prevent XXE attacks, we may also need the following changes 
```
factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
factory.setXIncludeAware(false);
factory.setExpandEntityReferences(false);
```

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+
+            // Step 2: walk <field> elements and resolve each to its fieldType class.
+            var result = new LinkedHashMap<String, String>();
+            NodeList fields = root.getElementsByTagName("field");
+            for (int i = 0; i < fields.getLength(); i++) {
+                if (!(fields.item(i) instanceof Element el)) continue;
+                var name = el.getAttribute("name");
+                var type = el.getAttribute("type");
+                if (!name.isEmpty() && !type.isEmpty()) {
+                    result.put(name, typeNameToClass.getOrDefault(type, ""));
```
k-rooot: 1. If a <field> references a <fieldType> that doesn’t exist, we are using "" as the class. This will be treated as non-text -> term query potentially incorrect query behaviour
2. If duplicate `<field name="...">` entries exist, later entries overwrite earlier ones. While unlikely, this can happen in merged schemas

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+
+            log.info("Loaded {} field class mappings from {}", result.size(), path);
+            return Map.copyOf(result);
+        } catch (Exception e) {
+            log.warn("Failed to parse managed-schema.xml at {}: {}", path, e.getMessage());
```
k-rooot: nit: Do we need stack trace for debugging?

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    private static Map<String, String> buildTypeClassMap(Element root) {
+        var map = new LinkedHashMap<String, String>();
+        NodeList fieldTypes = root.getElementsByTagName("fieldType");
+        for (int i = 0; i < fieldTypes.getLength(); i++) {
+            if (!(fieldTypes.item(i) instanceof Element el)) continue;
+            var name = el.getAttribute("name");
+            var cls  = el.getAttribute("class");
+            if (!name.isEmpty()) {
+                map.put(name, cls);
```
k-rooot, line 111: When `<fieldType name="test" />`, cls becomes ""

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+    public static Map<String, String> fromXmlFile(Path path) {
+        if (path == null || !Files.exists(path)) {
+            log.warn("managed-schema.xml not found at {}, skipping fieldTypes", path);
+            return Map.of();
+        }
+        try {
+            var factory = DocumentBuilderFactory.newInstance();
+            factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            var doc = factory.newDocumentBuilder().parse(path.toFile());
+            var root = doc.getDocumentElement();
```
k-rooot, line 68: This can be updated to `doc.getDocumentElement().normalize();` to normalise the xml doc

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
```diff
@@ -0,0 +1,105 @@
...
+import java.nio.file.Path;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import lombok.extern.slf4j.Slf4j;
+import org.w3c.dom.Element;
+import org.w3c.dom.NodeList;
+
+/**
+ * Reads Solr field metadata from a {@code managed-schema.xml} and emits a flat
```
k-rooot, line 15: Solr schemas often use
```
<dynamicField name="*_txt" type="text_general" />
```

In these scenarios, queries on dynamic fields will fall back to `match`. Should we handle dynamic fields?

k-rooot (changes requested):

solrMigrationDevSandbox/docker-compose.yml
nagarajg17: As discussed offline, docker volume mount helps in mapping files from customer local dir to inside docker container dir. We'll need to update wiki which I'll do it post this PR closure

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/orchestrator/translateQ.ts
nagarajg17: Yes for Solr. Updated comment. Initially it mapped Solr to OpenSearch and that was sent in transformations. In offline discussion due to maintenance and other concerns we decided not to maintain this Solr to OpenSearch field mapping. Comment was missed to update post that mapping removal

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
nagarajg17: Safer default is match query when we don't know the field type. Having said that I don't have a practical example when it can happen. All Solr fields have a type

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
nagarajg17: Done

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
nagarajg17: It doesn't match wildcard or fuzzy pattern and defaults to term/match query depending on field type. Solr doesn't throw error for this
```
curl -s "http://localhost:8983/solr/testharness/select?q=product~a" | python3 -m json.tool
{
    "responseHeader": {
        "zkConnected": true,
        "status": 0,
        "QTime": 14,
        "params": {
            "q": "product~a"
        }
    },
    "response": {
        "numFound": 0,
        "start": 0,
        "numFoundExact": true,
        "docs": []
    }
}
```

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
nagarajg17: Updated the comment since it was misleading. For fields not present in fieldTypes, we have no schema information and are in best-effort territory. We default to match because it aligns with Solr's query-time analysis behavior for text fields, which are the most common case when schema metadata is absent

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.test.ts
nagarajg17: Good catch. Changed back to checking full query

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.test.ts
nagarajg17: Done

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
nagarajg17: Created a constant

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
nagarajg17: Java always sends a map

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
nagarajg17: No UTs exist

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: Updated to `solrSchemaXmlFile not configured, skipping fieldTypes`

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: No. The disallow-doctype-decl flag rejects the entire XML file the moment it sees a DOCTYPE declaration. Since all XXE attacks require a DOCTYPE to define the malicious entity, this single flag blocks every possible XXE variant before the parser even gets to process entities

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: 1. Correct. Update code to drop the field from fieldMapping, so fieldRule will use match query
2. Seems correct behaviour during conflicts

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: Updated to log entire error

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: This is removed now

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: Not needed for this use case. If we were reading element text content with `el.getTextContent()` or `el.getFirstChild().getNodeValue()`, then `normalize()` would be worth calling

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
nagarajg17: Good point. I think we need to discuss with SA's. Currently both default and this field mapping xml doesn't handle dynamic fields. Coming to changes wise - its not a trivial change. Resolving this fields at JS or Java layer both has its own merits and demerits, needs deeper discussion and changes for this. Its good to track as separate task if SA's agree. We would need changes for both xml files

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
nagarajg17: Created EMPTY_FIELD_TYPES as a module-level constant — allocated once when the module loads, reused on every call where fieldTypes is not provided. No per-call allocation

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
k-rooot: Can we add unit tests limiting the test coverage to the changes introduced in this PR? Otherwise, we will continue to add more functionality in this module without any UT coverage.

TrafficCapture/transformationShim/src/main/java/org/opensearch/migrations/transform/shim/SolrSchemaProvider.java
k-rooot: Sure. Please start a thread in the SAs group to reach consensus and add tasks accordingly.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/transformer/rules/fieldRule.ts
k-rooot: Thanks for making the changes in all modules in the PR.

TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/request.transform.ts
nagarajg17: `request.transform.ts` is the GraalVM entry point — it uses declare const bindings which is injected by Java at runtime. That makes it untestable in isolation in Vitest (Node.js environment) because bindings is never defined there. In tests, bindings is undefined, so both constants fall back to their defaults (undefined and EMPTY_FIELD_TYPES). So might not be useful to add UT for this case. LMK your thoughts

nagarajg17: Needed some code refactor to make dependency injectable for UTs. Made that and added UTs for `request.transform.ts` (all existing functions as well)

PR #2856, opened 2026-04-30
Verify HTTP-Basic cluster secrets exist before submitting a workflow

jugal-chauhan:
### Description
Every `workflow submit` invocation prints a somewhat confusing warning:
```
NOT checking if all secrets have been created. Run workflow configure edit to confirm
```

The secret-existence validation only runs inside `workflow configure edit`. Submit blindly trusts that whatever was validated at edit-time is still valid at submit-time.

But secrets can disappear between those two steps for e.g., a secret is deleted via `kubectl delete secret`, the namespace gets recreated, or the config is edited to reference a new secret that hasn't been created yet. When that happens, the Argo 
workflow is accepted and starts running, then fails mid-migration with an authentication error against the source or target cluster. The real cause (a missing secret) is buried inside an inner workflow failure, making it hard to diagnose.

### Changes

`workflow submit` now verifies that every HTTP-Basic secret referenced by the saved configuration still exists in Kubernetes **before** submitting the workflow.

- **Happy path:** the warning line is gone; submit proceeds silently.
- **Missing secret:** submit exits non-zero before any workflow resources are created, with a clear message naming the missing secret and pointing the user at `workflow configure edit` to create it.

### Example output, missing secret                                                                                                 
```
$ workflow submit                                                                                                                  
  Error: Found 1 missing secret that must be created to make well-formed HTTP-Basic requests to clusters:
  source-cluster-creds
  Run workflow configure edit to create them.
```

### Implementation

The secret-validation helpers that previously lived inside `configure.py` have been extracted into a new `commands/secret_utils.py` module so both `configure edit` and `submit` can share them. This follows the existing `argo_utils.py` / `crd_utils.py` pattern already used in the commands package.

- `validate_and_find_secrets(raw_yaml)` — runs the TypeScript Zod validator and returns referenced secret names (existing logic, moved)                                                                          
- `process_secrets(store, result, interactive)` — checks each referenced secret exists; interactive mode prompts to create missing ones, non-interactive raises (existing logic, moved)
- `verify_configured_secrets_exist(store, raw_yaml)` — new thin wrapper which the non-interactive guard `submit` uses
- `get_credentials_secret_store_for_namespace(namespace)` — namespace-based `SecretStore` factory, mirroring `get_credentials_secret_store(ctx)` in `workflow.models.utils` but compatible with `submit`'s command-level `--namespace` option
- `CREDENTIALS_USE_CASE_LABEL` — constant extracted from the previously-inline `"http-basic-credentials"` label string

### What this does not cover                                                                                                       

The check only verifies HTTP-Basic-auth secrets, which is the current scope of secrets tracked by the workflow CLI. Other cluster credential types (sigv4, mTLS, etc.) are not validated here. If those are added later, this module is the natural place to extend.

### Testing

- Updated four existing `workflow submit` unit tests to mock the new helpers so they don't hit Kubernetes or the TypeScript validator
- Added `test_submit_command_fails_fast_when_secrets_missing` that verifies:
    - Submit exits non-zero when a referenced secret is missing 
    - Error output names the missing secret
    - Error output points the user at `workflow configure edit`
    - The workflow-submit subprocess is never invoked (the guard short-circuits correctly before any Argo mutations occur)

- All 14 tests in `test_workflow_cli_commands.py` pass.                                                                              
- All 141 non-integration workflow tests pass. The 37 erroring live-integration tests are pre-existing and require a `kind-console-link-test` cluster; they are not affected by this change.

### Issues Resolved
N/A

### Check List


migrationConsole/lib/console_link/console_link/workflow/commands/secret_utils.py
```diff
@@ -0,0 +1,140 @@
...
+    if not missing:
+        return
+    if interactive:
+        _handle_add_basic_creds_secrets(secret_store, missing)
+    else:
+        raise click.ClickException(
+            f"Found {len(missing)} missing secret{'s' if len(missing) > 1 else ''} "
+            f"that must be created to make well-formed HTTP-Basic requests to clusters:\n  " +
+            "\n  ".join(missing) +
+            "\n\nRun `workflow configure edit` to create them."
```
AndreKurait, line 111: This is good for now, but still not ideal, a new command `workflow configure secrets` that directly went to this flow would be better

AndreKurait (approved):

PR #2865, opened 2026-05-01
Remove orphaned traffic-replay-default-e2e-test Jenkins pipeline

AndreKurait:
## Summary

Removes the `trafficReplayDefaultE2ETest` Jenkins integration test and its orphaned scaffolding. The test depended on the legacy source-CDK `captureProxyEnabled` flag, which has no working wiring in the current migration-CDK. The test deployed a stack with no capture proxy attached to the source, so the MSK `logging-traffic-topic` was never created and nothing was ever replayed. The failures observed in the Jenkins `traffic-replay-default-e2e-test` job (`test_replayer_0001_empty_index`, `test_replayer_0006_OSB`, `test_replayer_0007_timeBetweenRequestsOnSameConnection`) are all downstream symptoms of that root cause.

## Background: why a Kafka reset wouldn't fix this

The initial hypothesis was stale Kafka state between runs — i.e. add a `kafka delete-topic` before sending new traffic. Investigation showed the topic never existed in the first place:

```
2026-05-01 16:41:15  INFO  Executing command: kafka-topics.sh --delete --topic logging-traffic-topic
2026-05-01 16:41:18  INFO  Failed to perform Delete Topic command: ...
    java.lang.IllegalArgumentException: Topic 'Optional[logging-traffic-topic]' does not exist as expected
```

No topic → no captured traffic → nothing for the replayer to push to the target cluster → every assertion against the target returns `index_not_found_exception` (404). The test itself is asking the deployment to do something the deployment is not configured to do.

The `captureProxyEnabled` flag is a leftover from the old source-CDK repository. A repo-wide search finds zero consumers on the migration-CDK side — it's a dead option that the test invokes and the stack ignores.

## Problems Solved

1. **Orphaned test producing noise** — `trafficReplayDefaultE2ETest` runs hourly via `periodicCron.groovy`, fails every time on empty-target-cluster assertions, and can never pass without first rewiring capture-proxy support into the migration-CDK.
2. **Dead docs / cron entries** — `README.md` table row and `periodicCron.groovy` case statement pointed at a file that no longer exists (after the groovy deletion), producing dangling references.

### Deleted the test and its exclusive scaffolding (3 files, 185 lines)

- `vars/trafficReplayDefaultE2ETest.groovy` — the test itself
- `jenkins/migrationIntegPipelines/trafficReplayDefaultE2ETestCover.groovy` — cover job that invoked it
- `jenkinsTests/src/test/groovy/JenkinsPipelineTestRunner.groovy` — dedicated unit-test runner for only this one pipeline

### Removed orphan references (2 files, 2 lines)

- `vars/periodicCron.groovy` — dropped the `@hourly` cron case for `main-traffic-replay-default-e2e-test`
- `jenkins/migrationIntegPipelines/README.md` — removed the corresponding docs table row

### Retained shared infra

- `test/defaultSourceContext.json` stays — it's consumed by `awsE2ESolutionSetup.sh` as the default context and by `rfsDefaultE2ETest.groovy`, `fullES68SourceE2ETest.groovy`, and `defaultIntegPipeline.groovy` via the `source-single-node-ec2` context id.

## Verification

Post-change grep across the repo for either `trafficReplayDefaultE2ETest` or `traffic-replay-default-e2e-test` returns **0 hits** — the name is fully excised. Coverage for the surviving `source-single-node-ec2` fixture remains via `rfsDefaultE2ETest` and `fullES68SourceE2ETest`, both of which exercise the same shared source stack.

## Migration notes

No user-facing impact. The orphan was not on any documented surface and its Jenkins job was a cover for a test that hadn't passed in the observed run. If/when capture-proxy replay is reintroduced on migration-CDK, it should be built fresh against the current stack rather than resurrected from the legacy flag.


jugal-chauhan (commented):
Thanks for brining in these changes. A few suggestions on cleanup : 

1. After this PR, `jenkinsTests/` dir becomes a completely empty subproject. It was built for `JenkinsPipelineTestRunner.groovy` and has no other consumers.

- We can further drop `include 'jenkinsTests'` from `settings.gradle` [here](https://github.com/opensearch-project/opensearch-migrations/blob/main/settings.gradle#L49)
- We can also remove `jenkins-pipeline-unit` entries from `gradle/libs.versions.toml` as it is now [unused](https://github.com/opensearch-project/opensearch-migrations/blob/main/gradle/libs.versions.toml#L20)

2. Set `captureProxyEnabled: false` in [defaultSourceContext.json](https://github.com/opensearch-project/opensearch-migrations/blob/main/test/defaultSourceContext.json#L8). The remaining ECS tests (`rfsDefaultE2ETest`, `fullES68SourceE2ETest`) already inline their own contexts with `captureProxyEnabled: false`, so they're unaffected. The only runtime consumer of `defaultSourceContext.json` is `awsE2ESolutionSetup.sh` as a fallback default and setting it to `false` is the correct default since the flag does nothing. Or alternatively, remove the `captureProxyEnabled` key entirely from `defaultSourceContext.json`.

AndreKurait:
Thanks @jugal-chauhan — all three cleanups addressed in 901ba2700:

1. Deleted `jenkinsTests/` directory entirely (was just `README.md` + `build.gradle` with no sources after the orphan removal).
2. Dropped `include 'jenkinsTests'` from `settings.gradle`.
3. Removed both `jenkins-pipeline-unit` entries from `gradle/libs.versions.toml` (version pin + library coordinates).
4. Went with your alternative on `defaultSourceContext.json` — removed the `captureProxyEnabled` key entirely rather than setting it to `false`, since the flag is a no-op under the current CDK and there's no value carrying a dead key.

Verified no remaining references to `jenkinsTests`, `jenkins-pipeline-unit`, `JenkinsPipelineTestRunner`, or `lesfurets` anywhere in the tree (gradle, groovy, markdown).

AndreKurait:
Addressed in 901ba2700 — dropped `jenkinsTests/`, the `include` in `settings.gradle`, the `jenkins-pipeline-unit` entries in `libs.versions.toml`, and removed the `captureProxyEnabled` key outright from `defaultSourceContext.json` (went with your alternative since the flag is a no-op). `spotlessApply` clean.

jugal-chauhan (approved):
LGTM! Thanks for addressing these comments as well.

jugal-chauhan (approved):
LGTM! Thanks for addressing these comments as well.

PR #2870, opened 2026-05-03
Add workflow show command to display output artifacts

gregschohn:
### Description
Builds upon https://github.com/opensearch-project/opensearch-migrations/pull/2834

Add durable workflow show outputs for metadata migration

Add resource-centric workflow output retrieval so users can inspect retained metadata migration output after the producing workflow or pod is gone. Introduce `workflow show` as the user-facing command for durable output, with task-first selection for `evaluateMetadata` and `migrateMetadata`, resource filtering, context-sensitive `--list`, history/run selection, and clean output formatting.  Wire the command into the workflow CLI so the output path is described as resource output rather than ephemeral pod logs.

Add artifact-store helpers for reading, listing, and deleting retained output objects. The helper prefers a mounted artifact bucket when available and falls back to S3 access using the configured bucket, region, endpoint, or standard bucket environment variables. This gives both `workflow show` and `workflow reset` one shared path for durable output storage operations.

Persist metadata evaluate/migrate output in the workflow itself. The metadata workflow now writes command output to `/tmp/outputs/metadata-output.log`, stores that file as an explicit Argo artifact under `migration-outputs/snapshotmigration/<resource>/<uid>/<task>/<workflowUid>.log`, and patches the owning `SnapshotMigration` status with the artifact name, S3 key, resource UID, workflow name/UID/creation timestamp, and config checksum. This lets future commands start from the migration CR instead of trying to rediscover data through Argo's current workflow/node artifact API.

Add `--outputFile` to the migrate/evaluate argument model and have `MetadataMigration` initialize parent directories, truncate the file at startup, and append both human-readable and JSON command output to it while preserving the existing log behavior. The workflow passes `outputFile` through inline JSON rather than as a mixed command-line flag so it stays compatible with `JsonCommandLineParser`'s JSON-mode shape.

Extend the Argo workflow builder to render explicit artifact outputs with S3 keys. Container builders can now attach named artifact outputs, the renderer emits those artifacts into the template output block, and parameter schemas cover the new artifact shape. Add unit coverage for explicit artifact output rendering.

Update `workflow reset` to clean retained output artifacts when deleting migration resources. Reset now derives the resource UID before deletion, removes the durable output prefix for that resource by default, and supports `--keep-output-artifacts` for preserving output while printing the retained S3 prefix and resource creation/UID details. Add comments explaining why the reset cleanup removes the entire prefix rather than only the status-referenced object: Argo-compatible artifact writes can leave metadata/index side files next to the primary output object.

Wire the manage TUI to the same retained output path used by `workflow show`. Manage now detects metadata output patch steps, maps them back to the owning `SnapshotMigration` status output, reads the durable artifact through `read_managed_output`, and opens it in a clean pager view. The keybindings now reserve `l` for pod logs and use `o` for showing retained output. The Argo workflow filtering keeps managed output artifacts while avoiding unrelated payload expansion.

Document and align the workflow log behavior with the durable output strategy. `workflow log` remains the ephemeral pod-log/debugging path, while `workflow show` is the resource-backed path for user-visible retained metadata output. Disable Argo `archiveLogs` in the chart values so every workflow pod no longer writes an extra `argo-artifacts/<workflow>/<pod>/main.log` object; durable metadata output is kept through explicit `migration-outputs/snapshotmigration/<resource>/<resourceCreationTimestamp>_<resourceUid>/<task>/<workflowCreationTimestamp>_<workflowUid>.log` artifacts instead. The resource creation timestamp comes from the existing SnapshotMigration apply/reconcile step output, not from an extra Argo step, so the parent directory is stable for the CR instance while each workflow run still gets its own timestamped file.

Add source-level guards against hand-written Argo template markers in builder literal inputs. Callers should use expression helpers instead of embedding raw `{{...}}` strings in `AllowLiteralOrExpression` values. The guard rejects those strings at literal/default/config-map/artifact-key entry points while still allowing the renderer to emit normal Argo template syntax from expression helpers.

### Testing
Manual tests run:
- `./gradlew :MetadataMigration:test --tests org.opensearch.migrations.MetadataMigrationTest`
- `npm run type-check`
- `npm run -w @opensearch-migrations/migration-workflow-templates test -- --updateSnapshot --runInBand`
- `pipenv run pytest tests/workflow-tests/test_workflow_cli_commands.py tests/workflow-tests/test_reset.py tests/workflow-tests/test_manage.py tests/workflow-tests/test_manage_injections.py -q`
- `pipenv run flake8 console_link/workflow/commands/show.py console_link/workflow/tui/manage_injections.py console_link/workflow/tui/workflow_manage_app.py tests/workflow-tests/test_workflow_cli_commands.py tests/workflow-tests/test_manage.py tests/workflow-tests/test_manage_injections.py`
- `git diff --check`

### Check List


AndreKurait (approved):

migrationConsole/lib/console_link/console_link/workflow/commands/configure.py
```diff
@@ -262,89 +322,118 @@ def _secret_has_managed_labels(secret_store, secret: V1Secret) -> bool:
...
 
 
-@secret_group.command(name="delete")
-@click.option('--list', 'list_names', is_flag=True, default=False, help='List managed HTTP-Basic auth secrets')
+@credentials_group.command(name="delete")
+@click.option('--list', 'list_names', is_flag=True, default=False, help='List managed HTTP Basic credentials')
 @click.argument('name', required=False, shell_complete=_complete_managed_secret)
-@click.option('--confirm', is_flag=True, default=False, help='Skip confirmation prompt')
+@click.option('--confirm', 'yes', is_flag=True, hidden=True)
+@click.option('-y', '--yes', is_flag=True, default=False, help=SKIP_CONFIRMATION_HELP)
```
AndreKurait, line 418: Thank you!

PR #2881, opened 2026-05-04
FilterScheme: scope built-in exclusion rules per filter context

AndreKurait:
### Description

`FilterScheme.filterByAllowList` applies a single pool of built-in exclusion rules (prefixes, suffixes, exact names) to every callsite. That same pool is consulted when deciding whether to migrate

- indices (`IndexRunner`, `ShardWorkPreparer`, `MigratorEvaluatorBase`)
- legacy index templates, composable index templates, and component templates (`GlobalMetadataCreator_OS_2_11`, called three times with different `TemplateTypes`)

Most of the exact-name entries — `logs`, `metrics`, `traces`, `tenant_template`, `agentless`, etc. — and every `@`-suffix entry — `@settings`, `@mappings`, `@template`, etc. — are Elastic/Fleet-shipped *template* names, not names a user would give a data index. With those rules applied at every callsite, any user index that happens to be named literally `logs` gets silently dropped from the migration even though the source cluster has real data in it.

This PR fixes that by tagging each built-in exclusion rule with the set of contexts in which it applies.

### Changes

Introduces `FilterScheme.FilterContext` with four values — `INDEX`, `LEGACY_INDEX_TEMPLATE`, `INDEX_TEMPLATE`, `COMPONENT_TEMPLATE` — and threads it through `filterByAllowList(List<String>, FilterContext)`. The template callsite derives the context from the existing `TemplateTypes` enum; the three index callsites pass `FilterContext.INDEX` explicitly.

Internally, the three flat `Set<String>` exclusion tables collapse into one ordered list of `ExclusionRule` records, each carrying a pattern, a match type (prefix / suffix / exact), and an `EnumSet<FilterContext>` declaring where it applies.

**Scoped to `TEMPLATES_ONLY`** (`LEGACY_INDEX_TEMPLATE ∪ INDEX_TEMPLATE ∪ COMPONENT_TEMPLATE`):

- all `@`-suffix rules (`@ilm`, `@lifecycle`, `@mappings`, `@package`, `@settings`, `@template`, `@tsdb-settings`) — OpenSearch/Elasticsearch index names cannot contain `@`, so excluding them from the index context was lossless but misleading
- the `@abc_template@`, `apm@`, `data-streams-`, `data-streams@`, `ecs@` prefixes — same reasoning
- exact names that are stack-shipped template names and legitimate user-data index names: `agentless`, `elastic-connectors`, `ilm-history`, `logs`, `logs-mappings`, `logs-settings`, `logs-tsdb-settings`, `metrics`, `metrics-mappings`, `metrics-settings`, `metrics-tsdb-settings`, `profiling`, `search-acl-filter`, `synthetics`, `tenant_template`, `traces`, `traces-mappings`, `traces-settings`, `traces-tsdb-settings`

**Kept at `ALL` contexts** (still excluded everywhere):

- leading-dot system indices (`.`)
- kibana, searchguard, sg7-
- apm-, behavioral_analytics-, elastic-connectors-, elastic_agent., ilm-history-, metricbeat-, profiling-, synthetics-
- logs-elastic_agent, logs-endpoint., logs-index_pattern, logs-system.
- metrics-elastic_agent, metrics-endpoint., metrics-index_pattern, metrics-metadata-, metrics-system.
- the X-Pack watcher exact names (`triggered_watches`, `watches`, `watch_history_3`) — these are real indices on the source, not templates

### Testing

- Extended `FilterSchemeTest` from 22 to 27 tests; the existing cases are re-parameterized with the appropriate `FilterContext` and preserve their original assertions.
- Two new dedicated cases pin the fix down:
  - `testLogsIndex_isMigrated` — every template-only exact name (`logs`, `metrics`, `traces`, `profiling`, `synthetics`, `agentless`, `elastic-connectors`, `ilm-history`, `tenant_template`, `search-acl-filter`, and all the `-mappings`/`-settings`/`-tsdb-settings` variants) now passes the index filter.
  - `testLogsTemplate_isExcluded_inAllTemplateContexts` — those same names are still excluded across all three template contexts.
- One new case (`testAllowlistBypassesBuiltinExclusions`) documents that a user-supplied allowlist overrides built-in exclusions, preserving the prior contract.
- `./gradlew :RFS:test :MetadataMigration:test` passes cleanly.

### Backwards compatibility

`filterByAllowList(List<String>)` is replaced with `filterByAllowList(List<String>, FilterContext)`. All four production callsites are updated in this PR and no tests consumed the old single-arg form outside `FilterSchemeTest`, so no deprecated overload is kept — every caller is required to state which kind of thing is being filtered, which is the point of the change.

### Issues Resolved

N/A — driven by a user-reported regression where an index literally named `logs` was being silently dropped by the metadata filter.

### Is this a backport?

No.

### Testing

- [x] All tests pass
  - [x] `./gradlew :RFS:test :MetadataMigration:test`

### Check List

### DIFF
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
@@ -169,7 +169,7 @@ public abstract class MigratorEvaluatorBase {
     protected void validateSourcelessIndices(Clusters clusters) {
         var metadataFactory = clusters.getSource().getIndexMetadata();
         var repoDataProvider = metadataFactory.getRepoDataProvider();
-        var skipFilter = FilterScheme.filterByAllowList(arguments.dataFilterArgs.indexAllowlist).negate();
+        var skipFilter = FilterScheme.filterByAllowList(arguments.dataFilterArgs.indexAllowlist, FilterScheme.FilterContext.INDEX).negate();
         var sourcelessIndices = new ArrayList<String>();
 
         for (var index : repoDataProvider.getIndicesInSnapshot(arguments.snapshotName)) {
--- a/RFS/src/main/java/org/opensearch/migrations/bulkload/common/FilterScheme.java
+++ b/RFS/src/main/java/org/opensearch/migrations/bulkload/common/FilterScheme.java
@@ -1,77 +1,171 @@
 package org.opensearch.migrations.bulkload.common;
 
-import java.util.Arrays;
+import java.util.Collections;
+import java.util.EnumSet;
 import java.util.List;
+import java.util.Set;
 import java.util.function.Predicate;
 
+/**
+ * Built-in exclusion rules applied when no user allowlist is supplied.
+ *
+ * <p>Rules are tagged with the set of {@link FilterContext}s in which they apply.
+ * Names like {@code logs} or {@code metrics} are Elastic/Fleet-shipped
+ * <em>template</em> names, so excluding them from template creation is correct,
+ * but excluding them from index migration would silently drop any user index
+ * that happens to share the name. Each rule therefore declares which contexts
+ * it applies to.</p>
+ *
+ * <p>When an allowlist is supplied, built-in exclusions are bypassed entirely and
+ * only the allowlist is consulted.</p>
+ */
 public class FilterScheme {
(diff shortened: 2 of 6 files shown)


Davidding4718 (approved):

jugal-chauhan (approved):
LGTM

PR #2882, opened 2026-05-05
RFS sourceless: single-pass doc-banked sidecar for term-position index

AndreKurait:
### Problem

Sourceless `_source` reconstruction OOMs on analyzed-text fields at scale.

`SegmentTermIndex` materialized the entire field's `docId -> position-sorted List<String>` map in heap before serving the first doc. For a 3.8 GB segment with ~500M token-positions, that's **tens of GB** of `TreeMap` nodes and `String` refs, pinned for the segment's lifetime.

Observed: OOM on a **64 GB JVM** reconstructing a 3.8 GB segment (166k docs, one analyzed-text field, no `_source` / stored fields / doc_values).

### Fix

`SegmentTermIndex` delegates to a Lucene-native on-disk sidecar built by a single-pass, doc-banked streaming `SidecarBuilder`:

- `registerTerm` streams term bytes straight to `terms.dat` and writes the little-endian offset to `term-offsets.dat` — no in-heap term list.
- `accept` appends the incoming tuple's `(position, termId)` pairs into the target doc's amortized `int[]` bank. The walk is term-major, so pairs arrive in term-order, not position-order.
- `buildAndOpenReader` walks docs in ASC order; for each non-empty doc it packs pairs into a `long[]` keyed `(position << 32) | termId`, runs `Arrays.sort` (primitive dual-pivot quicksort — cache-friendly, no boxing, natural order matches the required `(position ASC, termId ASC)`), uvint-encodes `(numPairs, [posDelta, termId]...)` straight into `sidecar.dat`, and releases the doc's bank so peak off-heap live set shrinks as docs are consumed.
- `doc-index.dat` is a plain `long[maxDoc]` little-endian array of absolute offsets, with a `-1L` sentinel for empty docs.

`SidecarReader` mmaps the resulting files and binary-searches `doc-index.dat` per lookup — no eager heap materialization.

### Heap footprint: before vs. after

| Phase | Before | After |
|---|---|---|
| Per-segment field build | O(total_positions_in_field) as `TreeMap` nodes + `String` refs — tens of GB for a 3.8 GB segment | O(total_positions × 8 bytes) of primitive `int[]` banks + `long[]` sort buffers, released per doc as emitted |
| Per-doc read | O(positions_in_one_doc) | O(positions_in_one_doc) |
| Dominant term | `TreeMap<Integer, List<String>>` over every (doc, position) | per-doc `int[]` banks (primitive, no object headers, no boxing) |
| Scales with segment size? | No (OOMs on 64 GB JVM at 3.8 GB segment) | Proportional to positions, but with ~10× smaller constant vs. the old boxed `TreeMap` |

Target: **10 GB JVM** reconstructs a **600 GB** analyzed-text shard.

### On-disk format

```
terms.dat          — length-prefixed UTF-8 bytes, one per term, in termId order
term-offsets.dat   — little-endian int[numTerms] of byte offsets into terms.dat
sidecar.dat        — uvint-encoded (numPairs, [posDelta, termId]...) per doc
doc-index.dat      — long[maxDoc] of absolute offsets into sidecar.dat, or -1L for empty docs
```

### Tests

- `SidecarBuilderRoundTripTest` — 6 cases covering small fixed input, empty stream, sparse docs, tied-position tie-break by termId, fuzz vs. an in-heap reference implementation, and file-handle closure on reader open.
- `PostingsSinkContractTest` — `BytesRefLike` defensive copy, bounds checks, value equality, and `PostingsSink` emission ordering.
- `VarintCoderTest` — round-trip, byte-count agreement, truncation error clarity, zigzag signed encoding, fuzz over both encodings.

All 104 tests in the `SearchSnapshotExtractor` module pass.

### External contract

Reconstructed `_source` JSON bytes for any `(segment, docId)` are **byte-identical** to the previous implementation. Every position, every duplicate token, original order round-trips losslessly so downstream consumers re-index to identical postings.

### Tradeoff

Per-doc banks live on heap between the term-major walk and the doc-major emit, so peak heap during build is `O(total_positions × 8 bytes)` plus array-growth overhead. That's the cost of skipping external sort and keeping the builder to a single pass with no `postings.raw` scratch file, no run files, and no k-way merge. Acceptable for segments within a normal JVM heap budget; oversized segments would need a size-heuristic fallback to a disk-based builder, to be layered on separately if it becomes necessary in practice.

### Check List

- [x] New functionality has been documented in the commit body and PR description

### DIFF
--- a/SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneLeafReader.java
+++ b/SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/LuceneLeafReader.java
@@ -6,6 +6,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import org.opensearch.migrations.bulkload.lucene.sidecar.PostingsSink;
+
 public interface LuceneLeafReader {
 
     public LuceneDocument document(int luceneDocId) throws IOException;
@@ -62,17 +64,26 @@ public interface LuceneLeafReader {
     default String getValueFromTerms(int docId, String fieldName) throws IOException { return null; }
 
     /**
-     * Version-specific hook: walk the terms dictionary for {@code fieldName} once and return a
-     * docId -> position-ordered term list map. This performs the raw Lucene iteration over
-     * shadow-relocated {@code Terms}/{@code TermsEnum}/{@code PostingsEnum}, which is why it
-     * can't be written in this shared interface.
+     * Streams the inverted index for {@code fieldName} into {@code sink} as
+     * {@code (termId, docId, positions[])} callbacks, one per (term, doc) pair that has
+     * at least one non-negative position.
      *
-     * Called at most once per (segment, field) via {@link SegmentTermIndex}. The returned map
-     * is owned by the caller and lives only as long as the {@link SegmentTermIndex} that holds
-     * it, which is scoped to a single segment's Flux in {@link LuceneReader#readDocsFromSegment}.
+     * <p>Ordering contract: terms are visited in ascending-bytes order (matches Lucene's
+     * {@code TermsEnum.next()}), and for each term docs are visited in ascending-docId order
+     * (matches Lucene's {@code PostingsEnum.nextDoc()}). Positions within each callback are
+     * already in ascending order as emitted by {@code PostingsEnum.nextPosition()}.
+     *
+     * <p>The default implementation is a no-op: versions that don't support source
+     * reconstruction from the inverted index (e.g. Lucene 10 / OS 3.x which always have
+     * stored fields or doc_values) leave this unimplemented and callers see an empty stream.
+     *
+     * <p>Callers must first invoke {@link PostingsSink#registerTerm} on each new term (as it
+     * appears in the walk) to get its assigned {@code termId}, then invoke
+     * {@link PostingsSink#accept(int, int, int[], int)} once per (term, doc). The reusable
+     * {@code int[]} passed to {@code accept} is only borrowed for the duration of the call.
      */
-    default Map<Integer, List<String>> buildTermPositionIndex(String fieldName) throws IOException {
-        return Collections.emptyMap();
+    default void streamFieldPostings(String fieldName, PostingsSink sink) throws IOException {
+        // no-op: default reader has no terms to stream.
     }
 
     /**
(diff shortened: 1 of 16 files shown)


AndreKurait:
FYI — I built a prototype alternative (`SidecarBuilder` using JDK-native external merge sort over fixed-width 12-byte `(docId, pos, termId)` records) to probe whether we can get O(1) retained heap regardless of tuple count. Branch: [`AndreKurait:alt-offline-sort`](https://github.com/AndreKurait/opensearch-migrations/tree/alt-offline-sort) (single commit on top of this PR's head, `SidecarBuilder.java` only, +509/-105, on-disk format unchanged so `SidecarReader` is untouched).

Scale-up harness, seed 424242, 50K docs × 20K terms, `-Xmx512m`, sort buffer 64 MiB:

| avgPos | tuples | mainlike | PR OLD  | PR HEAD (this PR) | Alt-A (external sort)            |
|-------:|-------:|---------:|--------:|------------------:|----------------------------------|
| 2      | 274K   | 30 ms    | 929 ms  | **673 ms**        | 607 ms, 102 MiB alloc, 77 MiB ret |
| 10     | 6.4M   | 219 ms   | 3.9 s   | **1.05 s**        | 1.69 s, 173 MiB, 76 MiB          |
| 20     | 25.4M  | 675 ms   | 12.8 s  | **2.27 s**        | 4.51 s, 391 MiB, 76 MiB          |
| 40     | 100M   | OOM      | OOM     | OOM               | **17.8 s, 1255 MiB alloc, 76 MiB ret** ✅ |
| 80     | 401M   | OOM      | timeout | OOM               | **88.5 s, 4698 MiB, 76 MiB** ✅  |
| 100    | 626M   | OOM      | timeout | OOM               | **144.7 s, 7280 MiB, 76 MiB** ✅ |

Observations:
- This PR is a clear win vs. OLD (≈5× faster, ≈11× less retained heap at 25M tuples) and I'd merge it as-is.
- But retained heap still scales with tuples, so we still OOM on `-Xmx512m` at ~100M+ tuples. Whether that matters depends on the deployment heap and the upper bound on `Σ positions per segment` we expect in practice.
- Alt-A's retained heap is flat ~76 MiB across the entire sweep (O(sort_buffer + numRuns·64 KiB + maxDoc·8B), independent of tuple count), at the cost of ~2× small-scale latency from disk round-trip. 626M tuples on 512m heap complete in 145s sustaining ~100 MB/s temp I/O.

Posting as a reference in case the OOM ceiling is a hard requirement rather than a soft one — happy to either open it as a follow-up PR, fold it in here, or drop it if the current PR's ceiling is good enough. Full details and round-trip tests (4/4 pass) on the branch.

PR #2886, opened 2026-05-05
RFS sourceless: stream term-position index to disk — fixes OOM at scale, 9× less heap than mainline

AndreKurait:
## Summary

Replaces the sourceless RFS term-position index — today an in-heap `Map<BytesRef, Map<Integer, List<Integer>>>` that retains every (term, doc, position) tuple until reconstruction — with an on-disk external merge sort built on Lucene 10's `OfflineSorter`, producing four mmap-able sidecar files. Retained heap becomes a constant-factor function of a tunable sort buffer, independent of segment size, term cardinality, or positions-per-document.

This fixes `rfs-sourceless-scale-oom` OOMs observed in the integration environment: shards with analyzed-text fields whose tuple volume exceeds worker heap cannot complete today regardless of how much heap you throw at them.

## Where mainline's heap goes

Mainline retains, per field, per segment, until document reconstruction completes:

```
Map<BytesRef, Map<Integer, List<Integer>>>
 │                │                  │
 │                │                  └─ boxed Integer positions, ArrayList<Integer>
 │                └─ HashMap<Integer→List<Integer>>, boxed docIds
 └─ BytesRef keys (term bytes + header)
```

For a segment with **T** terms × **D** average docs-per-term × **P** average positions-per-doc, retained heap is on the order of `T × D × (HashMap.Node + Integer + ArrayList) + Σpositions × Integer`. Each boxed `Integer` is ~16 B + ~4 B pointer; each `HashMap.Node` is ~48 B; each `ArrayList` has ~40 B overhead before its `Object[]`. In practice, every 12 B of raw data bloats to **100–150 B retained**.

Concretely: the `md` benchmark workload (20 k terms × 5 k docs, avg ~3 positions/tuple) retains **2.1 GB** on mainline. The `lg` workload (80 k × 20 k) cannot be measured at all — **it OOMs on a 4 GiB heap**.

## What this PR changes

1. **`SidecarBuilder`** (`PostingsSink` + `AutoCloseable`) builds the on-disk index by delegating to Lucene 10's `OfflineSorter`:
   - `registerTerm(BytesRef)` streams term bytes into `terms.dat` as VInt-length-prefixed records and writes a long byte-offset into `term-offsets.dat`.
   - `accept(termId, docId, int[] positions, n)` packs each `(docId, pos, termId)` tuple as a 12-byte big-endian record into an `OfflineSorter.ByteSequencesWriter`. BE packing makes unsigned bytewise comparison sort naturally into `(docId ASC, pos ASC, termId ASC)`.
   - `buildAndOpenReader()` runs `OfflineSorter.sort(...)` — which handles spilling, k-way merge, and temp-file cleanup — then streams the sorted records through a `DocPayloadEmitter` that writes `sidecar.dat` (per-doc VInt-delta-pos + VInt-termId pairs) and `doc-index.dat` (long-array of per-doc offsets, sentinel for no-token docs).

2. **`SidecarReader`** opens the four files via `MMapDirectory` (native chunking past 2 GiB), exposes the two long-array files as `RandomAccessInput` slices for O(1) positional lookup, and serves `get(docId) -> List<String>` by cloning the sidecar `IndexInput` per call (clone = independent position state, shared mmap pages).

3. **`LeafReader5/7/9`**: per-version wiring to stream `(term, doc, positions)` into a `PostingsSink` during term enumeration — no intermediate object graph.

4. **Production defaults**: `SourcelessSpillConfig.DEFAULT_SORT_BUFFER_BYTES = 256 MiB`. Tunable via `-Drfs.reconstruction.sortBufferBytes`.

## Large-segment correctness

All file offsets are `long`; all record-counting fields (`docId`, `termId`, position, per-doc `pairCount`, term length) are `int`, matching Lucene's own ceilings:

| site | type | bound |
|---|---|---|
| `termsOut.getFilePointer()` → `termOffsetsOut.writeLong(...)` | `long` | `terms.dat` may exceed 2 GiB |
| `sideOut.getFilePointer()` → `docIndexOut.writeLong(...)` | `long` | `sidecar.dat` may exceed 2 GiB |
| `termOffsetsRA.readLong((long) termId * 8L)` | cast-before-multiply | safe to `Integer.MAX_VALUE` terms |
| `docIndexRA.readLong((long) docId * 8L)` | cast-before-multiply | safe to `Integer.MAX_VALUE` docs (`IndexWriter.MAX_DOCS ≈ 2³¹ − 129`) |
| `MMapDirectory` | chunked mmap | reads files > 2 GiB natively |
| `SidecarBuilder` ctor `sortBufferBytes` | `long` | widened in this PR (previously narrowed via `(int)` cast in `SegmentTermIndex`); clamped to `[MIN_BUFFER_SIZE_MB, Integer.MAX_VALUE]` MiB since Lucene's `BufferSize` is int-MiB-bounded |

`sidecar.dat`, `terms.dat`, and both offset-tables have been exercised end-to-end against a > 2 GiB payload by `SidecarReaderLargeFileTest.opensAndReadsPayloadsBeyondTwoGigabytes`.

## Retained heap bound

```
sort buffer           = sortBufferBytes     (config, default 256 MiB; OfflineSorter-managed)
streaming merge state = O(numRuns) run heads inside OfflineSorter
per-doc staging       = ByteBuffersDataOutput, reset between docs, holds one doc's VInt pairs
─────────────────────────────────────────────────────────────────────
steady-state retained ≈ sortBufferBytes + constants, independent of T, D, P, maxDoc
```

## Benchmark: sidecar vs mainline-sim at two heap sizes

Harness under `SearchSnapshotExtractor/src/test/java/.../sidecar/bench/`. `MainlineSimulationBenchmark` reproduces the mainline `Map<BytesRef, Map<Integer, List<Integer>>>` retention pattern on the same synthetic workloads `SidecarBuildBenchmark` drives. Wall clock = 3 warmup + 3 measured iters; heap peak = `MemoryMXBean.getHeapMemoryUsage().getUsed()` sampled at 25 ms.

| scale | path | heap | sortBuf | build (s) | peak heap (MB) | GC ms |
|---|---|---|---|---:|---:|---:|
| md  | sidecar      | 512m | 4 MiB   | **15.1** | **227**  | 2   |
| md  | sidecar      | 4g   | 256 MiB | 13.5     | 514      | 0   |
| md  | mainline-sim | 512m | —       | **OOM**  | —        | —   |
| md  | mainline-sim | 4g   | —       | 18.8     | **2,120** | 388 |
| lg  | sidecar      | 512m | 16 MiB  | **293.8** | **228** | 79  |
| lg  | sidecar      | 4g   | 1024 MiB| 256.4    | 2,072    | 24  |
| lg  | mainline-sim | 512m | —       | **OOM**  | —        | —   |
| lg  | mainline-sim | 4g   | —       | **OOM**  | —        | —   |

`md` workload: 20 k terms × 5 k maxDoc. `lg`: 80 k terms × 20 k maxDoc.

### Reading the numbers

- **`md` at 512m**: sidecar completes in 15 s with 227 MB peak; mainline OOMs. The PR's core claim.
- **`md` at 4g**: mainline finishes but retains **2.1 GB** — 9× sidecar's footprint — and is 25 % slower (18.8 s vs 13.5 s).
- **`lg` at 4g**: mainline OOMs even with 4 GiB. Sidecar runs in 5 min on 512m with a flat 228 MB heap.
- **Bigger sort buffer helps modestly**: 16 MiB → 1 GiB at `lg` saves 37 s (−13 %) for 9× the heap. 256 MiB is the sweet spot, hence the default.

### Measurement caveat

`MemoryMXBean.getHeapMemoryUsage().getUsed()` captures young-gen pre-GC, so sidecar's 227 MB peak at 512m-md overstates true retained heap. The theoretical bound (sort buffer + OfflineSorter merge state + per-doc staging) is a few MiB for this config — the rest is G1 young-gen headroom before the next collection. Mainline is measured the same way, so the 9× delta is apples-to-apples; and mainline's 2.1 GB is genuine retained object graph, not sampler artifact (confirmed by GC pressure: 8 GCs × 388 ms vs sidecar's 0 GCs).

## Testing

- `SidecarBuilderRoundTripTest` — 8 randomized round-trip cases: empty segment, single doc, dense/sparse doc-ids, long position streams, small sort-buffer fuzz (forces OfflineSorter to spill), tie-breaking at same position.
- `SidecarReaderLargeFileTest.opensAndReadsPayloadsBeyondTwoGigabytes` — verifies `MMapDirectory` chunking on files > 2 GiB.
- `PostingsSinkContractTest` — sink-contract invariants (accept ordering, termId range).
- Existing `SegmentTermIndex` and `LuceneReader` tests exercise the integration path.

## Reproducing

```
./gradlew :SearchSnapshotExtractor:testClasses
CP="SearchSnapshotExtractor/build/classes/java/test:$(./gradlew -q :SearchSnapshotExtractor:printTestClasspath)"
java -Xmx4g -Xms4g -cp "$CP" \
  org.opensearch.migrations.bulkload.lucene.sidecar.bench.BenchmarkRunner \
  {sidecar|mainline} {sm|md|lg} /tmp/out.md
# Override sort buffer for sidecar runs:
SIDECAR_BENCH_SORT_BUF_MB=256 java -Xmx4g ... sidecar lg /tmp/out.md
```


AndreKurait:
Integrated as f5d379a48
37fb8e799 on es-sourceless-experimental.

AndreKurait:
Integrated as f5d379a48 on es-sourceless-experimental.

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -25,6 +27,11 @@ public class SourceReconstructor {
 
     private SourceReconstructor() {}
 
+    // Dedup set for the "cannot nest under scalar" warning — the intent is "warn once per
+    // (fieldName, collidingParent, scalarType) tuple" so operators see the class of problem
+    // without the log being flooded once per document reconstructed.
+    private static final Set<String> WARNED_NEST_COLLISIONS = ConcurrentHashMap.newKeySet();
```
jugal-chauhan: This static Set grows without bound across the JVM lifetime. In a long-running RFS worker processing thousands of segments across hundreds of indices, unique (field, parent, type) collision keys accumulate forever.

Can we consider capping this to prevent unbounded growth during long migrations ?

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/sidecar/SidecarReader.java
```diff
@@ -0,0 +1,166 @@
...
+            rawEndOff[i]      = in.readZInt();
+        }
+
+        // Dedup: for each position group, keep the entry whose term is longest.
+        // The sidecar is sorted (docId, pos, termId) ASC so same-position entries are adjacent.
+        // Resolve all term strings up front so the inner loop is a simple length comparison.
+        String[] resolvedTerms = new String[numEntries];
+        for (int k = 0; k < numEntries; k++) {
+            resolvedTerms[k] = readTerm(rawTermId[k]);
+        }
```
jugal-chauhan: Why not resolve only within each position group and only materialize the winner ?

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/sidecar/SingleTermSidecarBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+ * <p>This builder deliberately does NOT use {@code OfflineSorter}: the walker
+ * drives docIds in ascending order within each term, but the cross-term
+ * interleaving would require sorting for a doc-major layout — which is
+ * unnecessary here because the output is already keyed by docId (O(1) random
+ * write per entry). The sparse seek-and-writeInt pattern is cheap: Lucene's
+ * {@code IndexOutput} is buffered, and the OS page cache coalesces adjacent
+ * writes.
+ */
+@Slf4j
+public final class SingleTermSidecarBuilder implements SingleTermSink, AutoCloseable {
```
jugal-chauhan: As a followup, can we add unit tests for this similar to `SidecarBuilderRoundTripTest` ? It will benefit catching edge cases like `maxDoc = 0` and `termId` at boundary limits

jugal-chauhan (approved):
Thanks for brining in these changes, the code changes LGTM! The comments below can be addressed in a follow up PR as they are non-blocking and NITs

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
AndreKurait: Today we do not process multiple shards on a single jvm

PR #2889, opened 2026-05-05
Make work item separator configurable

### LINKED ISSUE #2880: Make index name SEPARATOR configurable
The separator of index name for work items is [hard coded ](https://github.com/opensearch-project/opensearch-migrations/blob/3649e009c82221a34f0db721bc035685e512340d/RFS/src/main/java/org/opensearch/migrations/bulkload/workcoordination/IWorkCoordinator.java#L211)to be '__':

Most of my indices name happen to contain '__' which results in the migration tool not work for me totally. 

Please make this separator configurable.

sumobrian:
### Description

Resolves #2880.

The work item separator was hardcoded to `__`, which prevented migration of indices whose names contain `__`.

### Changes

- Make `WorkItem.SEPARATOR` a configurable static field with a default value of `__` for backward compatibility
- Add `--work-item-separator` CLI parameter to `RfsMigrateDocuments`
- Use `Pattern.quote()` in `split()` to handle regex-special characters in custom separators
- Use `volatile` for thread safety across workers

### Usage

Users with `__` in their index names can now run:
```
--work-item-separator ":::"
```

### Testing

- Default behavior unchanged (separator defaults to `__`)
- Existing tests pass without modification

### DIFF
--- a/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
+++ b/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
@@ -291,6 +291,13 @@ public class RfsMigrateDocuments {
         private ExperimentalArgs experimental = new ExperimentalArgs();
 
         @Parameter(required = false,
+            names = { "--work-item-separator" },
+            description = "Optional. The separator string used to delimit components in work item IDs. " +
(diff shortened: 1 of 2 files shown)


AndreKurait:
Closing in favor of https://github.com/opensearch-project/opensearch-migrations/pull/2890

gregschohn (commented):
Instead of this, can we base64 encode the index name before we drop the document into our index

PR #2890, opened 2026-05-05
Allow '__' in index names for work item parsing

### LINKED ISSUE #2880: Make index name SEPARATOR configurable

AndreKurait:
### Description

Fixes #2880: the work-coordinator rejected (or silently mishandled) indices whose names contain `__`.

The work-coordinator stores one document per `(index, shard, startingDocId)` triple in `.migrations_working_state`, keyed by a serialized id. Before this change that id was `indexName + "__" + shard + "__" + docId`, and the parser split on the first two `__` — which meant any index name containing `__` was unmigratable. The pre-check in the `WorkItem` constructor also rejected such names outright.

This change base64url-encodes (no padding) the index-name segment of the id:

```
<base64url(indexName)>__<shard>__<docId>
```

The base64url alphabet is `[A-Za-z0-9_-]`, so the encoded segment can never contain two consecutive underscores. The existing `split(SEPARATOR + "+")` parser keeps working as-is, unchanged. The `shard_setup` sentinel is still written verbatim.

For operator auditability, the plaintext index name is also written as an `indexName` keyword field on the work-item document body, so the working-state index is readable via a normal query without having to reverse the id encoding.

### Issues Resolved

Closes #2880

### Testing

- New `WorkItemStringParsingTest` unit test: round-trips plain names, names containing `__`, leading/trailing/repeated `__`, unicode, single-char / 2-char / 3-char / 4-char names (base64 padding boundaries), confirms the `shard_setup` sentinel still round-trips, confirms the serialized form always splits into exactly 3 components on `__`, and asserts that malformed ids throw `IllegalArgumentException`.
- Updated `WorkCoordinatorTest` integration-container fixtures that hand-assembled plaintext `"R0__0__0"`-style ids to go through `WorkItem.toString()` via a small `workId()` helper — this keeps them as readable as before while surviving any future serialization change.
- `Test0001SingleDocumentBackfill` now uses `test__0001__{unique_id}` as its target index name, so every run of the basic backfill integration test exercises an index name the pre-fix work-coordinator would have silently mishandled, across every supported ES→OS pair.

### Compatibility Note

This is a hard cutover: serialized ids are not forward- or backward-compatible with the previous plaintext format. To make the boundary explicit, the work-coordinator painless `scriptVersion` is bumped from `2.0` to `3.0`; the guard at the top of every script throws `IllegalArgumentException` when it encounters a document carrying a different version, so mixing binaries across the boundary fails loudly rather than silently corrupting state. Working-state indices written by an older build are not readable by this version and must be cleared before upgrading. The tool already assumes a fresh working-state index per migration run in normal operation.

### Check List
- [x] All tests pass
- [x] New functionality has been documented (commit messages + PR body + inline javadoc)

### DIFF
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/LeaseExpirationTest.java
@@ -360,10 +360,16 @@ public class LeaseExpirationTest extends SourceTestBase {
                 var successorItems = parentHits.get(0).path("_source")
                     .path(OpenSearchWorkCoordinator.SUCCESSOR_ITEMS_FIELD_NAME).asText();
 
-                // ASSERT: successor checkpoint doc is in early-trigger range
-                Assertions.assertTrue(successorItems.startsWith("geonames__0__"),
+                // ASSERT: successor checkpoint doc is in early-trigger range.  The work-item id is a
+                // base64url(indexName) + "__" + shard + "__" + docId string (see #2880), so decode it
+                // via the canonical parser rather than string-matching a plaintext prefix.
+                var successor = IWorkCoordinator.WorkItemAndDuration.WorkItem
+                    .valueFromWorkItemString(successorItems);
+                Assertions.assertEquals("geonames", successor.getIndexName(),
+                    "Successor should be for index geonames, got: " + successorItems);
+                Assertions.assertEquals(0, (int) successor.getShardNumber(),
                     "Successor should be for shard 0, got: " + successorItems);
-                var checkpointDoc = Integer.parseInt(successorItems.substring("geonames__0__".length()));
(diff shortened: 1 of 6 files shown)


gregschohn (commented):
What about base64 encoding the name?

AndreKurait:
> What about base64 encoding the name?

We can do that, the only issue that gives is it won't be backwards compatible with existing backfills if someone tries to upgrade without resetting the worker coordination index

sumobrian (approved):

AndreKurait:
@gregschohn updated to base64 based approach

PR #2891, opened 2026-05-05
Fix Docker Hub rate-limit failures in dockerComposeE2ETest

AndreKurait:
## Problem

`jenkins/migrationIntegPipelines/dockerComposeE2ETestCover.groovy` was failing on shared AL2023 Jenkins agents with `toomanyrequests: You have reached your unauthenticated pull rate limit` from Docker Hub. Two distinct paths were still hitting `docker.io` directly:

1. **`registry:2`** — `buildImages/backends/dockerHostedBuildkit.sh` `docker run`s this to stand up the local registry, and `vars/dockerComposeE2ETest.groovy` invokes that from a non-login `sh` step where `/etc/profile.d/ecr-pull-through.sh` doesn't apply.

2. **`docker.io/library/amazonlinux:2023`** (`elasticsearchTestConsole` target) — `buildImages/build.gradle` declared `serviceName: "elasticsearchTestConsole"` twice: once with PTC-rewritten `BUILDER_IMAGE`/`RUNTIME_IMAGE` buildArgs, once with no buildArgs. `registerBuildKitBakeAggregateTask` keys the included-projects map on `serviceName`, so the second entry overwrote the first and the generated bake spec had no build args — buildx fell back to the Dockerfile `ARG` defaults.

## Fix

- **Commit 1**: `dockerHostedBuildkit.sh` — add `ptc_rewrite_image()` (mirrors the `ptc_rewrite` helper already in `mirrorToEcr.sh` and `PullThroughCacheHelper.groovy`), wrap both `registry:2` call sites. `vars/dockerComposeE2ETest.groovy` — surface `ECR_PULL_THROUGH_ENDPOINT` into the `sh` step via `bash -l -c` and `withEnv`, matching the pattern in `vars/k8sLocalDeployment.groovy`.

- **Commit 2**: Remove the duplicate `elasticsearchTestConsole` entry from `buildKitProjects`.

## Behavior when `ECR_PULL_THROUGH_ENDPOINT` is unset

`ptc_rewrite_image` and `PullThroughCacheHelper.rewrite` both echo input unchanged when the endpoint is empty, so laptops and forks without a configured PTC are unaffected.

### DIFF
--- a/buildImages/backends/dockerHostedBuildkit.sh
+++ b/buildImages/backends/dockerHostedBuildkit.sh
@@ -14,9 +14,39 @@ set_docker_hosted_defaults() {
   : "${BUILD_CONTAINER_REGISTRY_ENDPOINT:=${EXTERNAL_REGISTRY_NAME}:5000}"
 }
 
+# Rewrite an image reference through the ECR pull-through cache when
+# ECR_PULL_THROUGH_ENDPOINT is set (e.g. on Jenkins hosts via
+# /etc/profile.d/ecr-pull-through.sh). Mirrors PullThroughCacheHelper.groovy
+# and the ptc_rewrite in deployment/k8s/.../mirrorToEcr.sh. Falls back to the
+# original reference when the endpoint is unset or the registry isn't mapped.
+ptc_rewrite_image() {
+  local image="$1"
+  local ptc="${ECR_PULL_THROUGH_ENDPOINT:-}"
+  [[ -z "$ptc" ]] && { echo "$image"; return; }
+  local prefix path
+  case "$image" in
+    docker.io/*)            prefix="docker-hub";               path="${image#docker.io/}" ;;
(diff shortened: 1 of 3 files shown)


jugal-chauhan (commented):
Can we also remove the redundant `elasticsearchTestConsole` entry. Or cherry-pick this [commit](https://github.com/opensearch-project/opensearch-migrations/pull/2892/changes/c21ed4cb859f6c6979f7d21a77e5855a390e0d55) onto here

AndreKurait:
> Can we also remove the redundant `elasticsearchTestConsole` entry. Or cherry-pick this [commit](https://github.com/opensearch-project/opensearch-migrations/pull/2892/changes/c21ed4cb859f6c6979f7d21a77e5855a390e0d55) onto here

Done

jugal-chauhan (approved):
LGTM! Thanks for brining in these changes

PR #2893, opened 2026-05-05
RFS sourceless: distribute recovered subfields into object-array seeds

AndreKurait:
## What

When sourceless RFS migration reconstructs a document whose partial `_source` seed
contains an object-array at an intermediate path, and doc_values / stored fields
carry subfield recovery values at dotted paths below that array, distribute those
subfield values into the seeded array elements instead of silently dropping them.

## Why

Consider a mapping with a `type:object` parent and three subfields, where
`_source.excludes` strips two of them at index time:

```json
{ "files": { "type":"object", "properties": {
    "cksum": {"type":"keyword"}, "size": {"type":"long"}, "name": {"type":"keyword"}
}}}
```

```json
{ "_source": { "excludes": ["files.size", "files.name"] } }
```

A document like `{"files":[{"cksum":"h1","size":100,"name":"a.txt"},{"cksum":"h2","size":500,"name":"b.txt"}]}`
is stored as:

- Partial `_source` seed: `{"files":[{"cksum":"h1"},{"cksum":"h2"}]}`
- `files.size` values in SortedNumeric doc_values
- `files.name` values in SortedSet doc_values

Before this change, the reconstructor walked `files.size` and found the parent
`files` was already a `List<Map>` from the partial `_source`. `putNested` refused
to descend, logged a warn, and dropped the recovered value. The migrated doc
ended up with `cksum` only; `size` and `name` were lost.

## Change

Extend `putNested` with an array-of-objects distribution branch:

- **Positional distribution** when the recovered value is a `List` the same
  length as the seeded object array: `list[i][leaf] = value[i]`.
- **Scalar broadcast** when the recovered value is a scalar: every element
  receives the same value.
- **Warn and drop** on size mismatch (no ordinal-to-element binding possible)
  or when the subfield path requires deeper nesting under the `List<Map>`
  parent (e.g. `files.meta.size`) — the distribute branch only guarantees
  immediate-leaf distribution.
- **Empty list** and **list of non-Maps** fall through to the existing
  warn-and-drop contract.

Extend `hasNested` so a leaf is considered present iff every `List<Map>`
element already carries it — makes re-merge idempotent and lets partial
element coverage still gap-fill.

## Ordering caveat

doc_values list order is traversal order (SortedNumeric: ascending numeric;
SortedSet: ascending term), **not** original array insertion order.
Element-to-element binding across subfields sourced from doc_values is
therefore approximate — fine for presence, search, and aggregation; not
display-accurate for per-element tuples where the original caller relied on
insertion order. Subfields sourced from **stored fields** preserve insertion
order per document.

This is documented on `distributeSubfieldAcrossList`.

## Test

`NoStoredSourceMigrationTest.testObjectArraySubfieldDistribution` is a real
end-to-end regression guard. For every `(source, target)` version pair in the
matrix it:

1. stands up source and target containers,
2. creates the index with the `type:object` parent + three subfields,
3. excludes `files.size` and `files.name` from `_source`,
4. indexes one document with a 2-element object array,
5. snapshots, then runs the full sourceless migration pipeline,
6. asserts the reconstructed `_source` preserves the object-array shape
   (length 2, not collapsed to columnar), keeps the seeded `cksum` on each
   element, and distributes `size` and `name` positionally from doc_values.

Test values are chosen so sorted-by-value equals insertion order, giving an
exact element-wise assertion without being flaky against the ordering caveat.
Gated to ES 5.x+ (matrix-consistent with `testUnrecoverableDocumentSkipped`).

## Scope

This PR is intentionally narrow — one production file, one integration test.
It is independent of and does not depend on the alt-offline-sort / SidecarBuilder
work on #2886.

## Checklist

- [x] Commits are signed per the DCO


AndreKurait:
Integrated as 032b361b2 on es-sourceless-experimental.

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -84,6 +99,23 @@ private static boolean putNested(Map<String, Object> target, String fieldName, O
...
+                // Only the LAST path segment is handled here; deeper nesting under a List<Map> parent
+                // (e.g. `files.meta.size` where `files[i].meta` would itself be an object) falls through
+                // to the warn branch below since per-element object construction from doc_values is
+                // outside the guarantees doc_values can provide.
+                if (i != parts.length - 2) {
+                    log.atWarn()
+                        .setMessage("Cannot write deeply-nested field '{}' under List<Map> at '{}'; dropping recovered value")
+                        .addArgument(fieldName)
+                        .addArgument(parts[i])
+                        .log();
```
jugal-chauhan: I like this guard : ` i != parts.length -2` - it checks limits distribution to immediate-leaf subfields only, making it easy to identify deep nested fields under a `List<Map>` parent. Am I understanding this correctly ? 

The warn message says "Cannot write deeply-nested field" but doesn't mention that single-leaf distribution is infact supported. Will this message always be printed for every dropped field ?

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -104,6 +136,78 @@ private static boolean putNested(Map<String, Object> target, String fieldName, O
...
+     * <p>
+     * Returns {@code true} iff at least one element was written.
+     */
+    @SuppressWarnings("unchecked")
+    private static boolean distributeSubfieldAcrossList(java.util.List<Object> list, String leaf,
+                                                        Object value, String fieldName) {
+        if (value instanceof java.util.List<?> valueList) {
+            if (valueList.size() != list.size()) {
+                log.atWarn()
+                    .setMessage("Cannot distribute '{}' across object array: got {} values for {} elements; dropping recovered value")
```
jugal-chauhan, line 195: Can we add a similar log line for the "ELSE" condition : `"Distributing {} values for '{}' across {}-element object array"` ?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/NoStoredSourceMigrationTest.java
jugal-chauhan, line 1: Would it be valuable to add a second test case where "sorted order" IS NOT EQUAL TO "insertion order", to explicitly document the "approximate binding" behavior?

jugal-chauhan (commented):
LGTM! Clean extraction from #2886. The addition of `hasNested` guarantee (leaf is considered present if and only if all elements carry it) is of significant value!

jugal-chauhan (approved):
LGTM! Thanks for addressing the comments and adding a well designed test `testObjectArraySubfieldDistributionApproximateBinding`. The `DEEP_NEST_UNDER_LIST_WARNED` AtomicBoolean is a better pattern (single boolean) than the `WARNED_NEST_COLLISIONS`

PR #2894, opened 2026-05-05
Add resource-centric view to workflow status and manage command

oeyh:
### Description

Makes the **resource-centric view** the default for `workflow status` and `workflow manage`, replacing the Argo workflow step tree with a clean, user-focused display. The old step tree is available via `--step-view`.

#### What it does

- Queries all migration CRs (KafkaCluster, CaptureProxy, DataSnapshot, SnapshotMigration, TrafficReplay, CapturedTraffic) from Kubernetes
- Organizes them into two top-level sections: **Snapshot Migration** and **Live Traffic Migration**
- Shows key spec fields, dependencies, and CR phase for each resource
- Filters workflow progress to only show actionable steps (running, failed, approval gates, output steps) — hides noise when all steps succeed
- Nests CapturedTraffic (Kafka topic) under its parent KafkaCluster
- Adds `resourceName` input parameter to all workflow template steps for reliable CR-to-workflow mapping
- Resource-centric TUI (`workflow manage`) with interactive expand/collapse, approval, log viewing, and output viewing on resource nodes
- Live backfill/snapshot status shown inline under workflow steps (default on, opt out with `--no-live-status`)
- Fixes `LiveCheckProcessor` to find `checkBackfillStatus` nodes and check `outputs.artifacts` for status data
- Resource-level log viewing (`l` key) in TUI — fetches merged logs for all pods belonging to a CR
- Gracefully degrades when Argo Server is unavailable

#### Example output (mid-migration)
```
Snapshot Migration
├── Snapshot
│   └── ✓ source-backfill-snapshot (Completed)
│       ├── snapshotPrefix: backfill
│       └── Depends on: capture-proxy
└── Backfill
    └── ○ source-target-backfill-snapshot-backfill
        ├── documentBackfillPodReplicas: 1 | sourceVersion: OS 2.11
        └── Workflow progress:
            └── ▶ checkBackfillStatus (Checked)
                ├── Backfill status: RUNNING
                ├── Percent completed: 30.0%
                ├── Estimated time to completion: 0h 6m 21s
                ├── Total shards: 10
                ├── Completed shards: 3
                └── Waiting shards: 6
Live Traffic Migration
├── Capture
│   └── ✓ capture-proxy (Ready)
│       └── podReplicas: 1 | listenPort: 9201
├── Buffer
│   └── ✓ default (Ready)
│       └── ✓ capture-proxy-topic (Ready)
└── Replay
    └── ○ capture-proxy-target-traffic-replayer
        ├── Depends on: capture-proxy, source-target-backfill-snapshot-backfill
        └── Workflow progress:
            └── ▶ Waiting for: source-target-backfill-snapshot-backfill (Running)
```
#### Example output (completed migration)
```
Snapshot Migration
├── Snapshot
│   └── ✓ source-backfill-snapshot (Completed)
└── Backfill
    └── ✓ source-target-backfill-snapshot-backfill (Completed)
Live Traffic Migration
├── Capture
│   └── ✓ capture-proxy (Ready)
├── Buffer
│   └── ✓ default (Ready)
│       └── ✓ capture-proxy-topic (Ready)
└── Replay
    └── ✓ capture-proxy-target-traffic-replayer (Ready)
```

#### Key design decisions

- **Resource view is now the default** — old Argo step tree available via `--step-view`
- **Dataclass model** (`ResourceNode`, `ResourceGroup`, `ResourceSection`) for type safety and extensibility
- **Filtered workflow progress** — only shows running/failed/pending/approval/output steps; hides entirely when all succeeded
- **`resourceName` added to all workflow template steps** — reliable CR-to-workflow mapping without heuristic fallbacks
- **Uses `get_node_phase`** (respects `overriddenPhase`) for accurate step filtering
- **Reuses `LiveCheckProcessor`** for live status — no custom backfill code
- **`--live-status` is default on** — opt out with `--no-live-status`
- **TUI uses stable IDs** (CR names, section/group names) to preserve expand/collapse state across refreshes
- **Resource-level log viewing** via extracted `get_resource_logs()` function

#### Files changed

- `workflow/resource_tree.py` (new) — Data model, tree building, workflow step extraction, step filtering, display rendering
- `workflow/tui/resource_tree_state_manager.py` (new) — Textual TUI tree state manager for resource view
- `workflow/tui/workflow_manage_app.py` — Resource view support, resource-level log/output bindings
- `workflow/commands/manage.py` — `--step-view` flag (default: resource view)
- `workflow/commands/status.py` — `--step-view` flag, `--live-status` default on, LiveCheckProcessor fixes
- `workflow/commands/log.py` — `get_resource_logs()` extracted function
- `workflow/commands/crd_utils.py` — Added `list_migration_resources_full()`
- `orchestrationSpecs/.../fullMigration.ts` — Added `resourceName` input to Kafka, Proxy, and Replay templates

### Issues Resolved
N/A

### Remaining tasks
- [ ] Merge workflow configuration into the resource tree - will tackle this in a followup PR.

### Testing
Manually tested against EKS cluster with running migrations (ES 2.11 → AOSS). Verified resource view with live backfill progress, filtered workflow steps, resource-level log viewing in TUI, output viewing on resource nodes, and graceful degradation when Argo Server unavailable.

### Check List


gregschohn:
Thanks - I'm loving the idea of removing workflow details and focusing on resources that are coming/going instead.  I'm ready to look at the code now - but I have one UX comment.  The `│   └── ✓ kafka-buffer (Ready)` line should be excised.  I see much value for it.  Let me know if you think that I might be missing something.

It would also be really valuable to show the source and target definitions.  Every other resource that we define in the config, including kafka and snapshot repos will have those configs explicitly included in some CRs.  The cluster settings get denormalized into many of the other steps.  

We don't track source/target endpoint values well today, but we should.  For a user, it would be useful to see that source is changing from sigv4 to http basic auth and to understand that through the status output.  

We could try to model sources and targets as resources, but when would we change them?  It wouldn't be immediately obvious which version each resource had picked up or was picking up.  That's also a much bigger change.

I'd shoot for preserving the cluster fields in the downstream programs that use them.  It's appropriately granular, scoped to each resource that uses it and aligns with all of the other work too.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    'Ready': ('✓', 'green'),
+    'Completed': ('✓', 'green'),
+    'Running': ('▶', 'yellow'),
+    'Initialized': ('○', 'cyan'),
+    'Error': ('✗', 'red'),
+    'Unknown': ('?', 'white'),
+}
+
+# Key spec fields to display per resource type
+SPEC_DISPLAY_FIELDS = {
```
gregschohn, line 47: It would be nice if we could get these from the schema.  We have the gated/impossible fields in the TS schema and have them flowing through with the "changeRestriction" field.  We could do something like "userVisible" too.  Then you'd grab the json schema and just do a filter.  Notice that I'm not sure that we have a way for python to pull the schema in & there would be a question of whether it should be runtime or build time.

For now, maybe put a note in the schema and a note here saying that they should be better linked, but leave that as a followup exercise for later.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    'kafkaclusters': ['version', 'auth.type', 'nodePool.replicas'],
```
gregschohn, line 48: can we change the name of this resource so that it doesn't include "kafka" across the board as a follow-up too

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+class ResourceNode:
+    """A single migration resource (CR instance)."""
+    name: str
+    plural: str
+    phase: str
+    depends_on: List[str]
+    spec: Dict[str, Any]
+    status: Dict[str, Any]
+    created_at: Optional[str] = None
+    children: List['ResourceNode'] = field(default_factory=list)
```
gregschohn, line 67: do we use this field?  Do we show sub resources for the top-level CRs?  Do you think that there could be value for a user to see that?

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+
+    # Future phases
+    config_diff: Optional[Dict[str, Any]] = None
+    workflow_step: Optional[Dict[str, Any]] = None
+    live_status: Optional[Dict[str, Any]] = None
```
gregschohn: not sure how this will be used, but making this a tuple w/ a timestamp that the live status was pulled might be helpful for the manage code

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    plural: str
+    phase: str
+    depends_on: List[str]
+    spec: Dict[str, Any]
+    status: Dict[str, Any]
+    created_at: Optional[str] = None
+    children: List['ResourceNode'] = field(default_factory=list)
+
+    # Future phases
+    config_diff: Optional[Dict[str, Any]] = None
```
gregschohn: I'm fine either way, but consider whether it's easier to track the diff or both pre-post values.  Will you only show pending changes (not just deployed ones)?  It seems like that and that probably makes the UI easier to follow.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    workflow_step: Optional[Dict[str, Any]] = None
```
gregschohn: What does 'step' mean here?  Is this the current step - is it the top-level step that runs tryApply on a resource?  We might have case where tryApplies might be at the same level - if so, we can adjust the workflow to simplify.  Does this have everything that we need to position ourselves properly to show all of the workflow context?

gregschohn:
After the completely merged resource-based graph is done - I wonder if it would be possible to do a tree in dependency order, like gradle task outputs.  The dependency graph isn't a tree, so you can't do the whole thing as a single consolidated tree because of diamonds - but we could say, for each given leave (e.g. a replayer, or snapshotmigration), here are the dependent (upstream) resources and everybody's state.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    """Walk the tree and extract steps for the innermost resource-tagged nodes."""
+    cr_name = _resolve_cr_name(node)
+    if cr_name:
+        has_explicit_name = (
+            get_node_input_parameter(node, 'resourceName') or
+            get_node_input_parameter(node, 'name')
+        )
+        has_inner_tagged = any(
+            child.get('group_name') for child in node.get('children', [])
+        )
```
gregschohn: do we need this check?  It would be nice to make the entire workflow completely canonical to reduce the complexity that other tools need to implement.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    for child in node.get('children', []):
+        _extract_steps_recursive(child, result)
+
+
+def _step_for_node(node: Dict[str, Any]) -> Dict[str, Any]:
+    """Get the step info for a resource node."""
+    completed = {'display_name': 'completed', 'phase': 'Succeeded', 'is_approval': False, 'denial_reason': None}
+    if node.get('phase') == 'Succeeded':
+        return completed
+    return _find_active_step(node.get('children', [])) or completed
```
gregschohn: why do we default to completed?  Is this necessary?

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+
+
+def _resolve_cr_name(node: Dict[str, Any]) -> Optional[str]:
+    """Resolve the CR name from a node with groupName_view."""
+    if not node.get('group_name'):
```
gregschohn, line 163: group_name is for how it will render.  We can either change the name to 'resourceName' or we can add another input for resourceName.  I'm worried about using a UI cue to infer a normative name.  I'm totally fine with the idea of decorating the steps with extra inputs that act as metadata.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    resource_name = get_node_input_parameter(node, 'resourceName')
+    if resource_name:
+        return resource_name
+    name = get_node_input_parameter(node, 'name')
+    if name:
+        return name
+    return node['group_name']
+
+
+def _find_active_step(children: List[Dict[str, Any]]) -> Optional[Dict[str, Any]]:
```
gregschohn: can this be extended to return - and later graft - an entire subtree rather than just one line?  For brevity, finding just one node will be really nice in the happy-path.  The line `return running_approval or running_leaf or failed_leaf or (` already shows that we have potential ambiguity.  We could have failed one step - but moved on to an approval.  We'll only show one of them though - the right answer would be to show both.

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -478,18 +478,64 @@ def _run_live_checks_parallel(self, in_progress_nodes: List[Dict[str, Any]]) ->
...
+                    for section in groups:
+                        for group in section.groups:
+                            for resource in group.resources:
+                                if resource.name in steps:
+                                    resource.workflow_step = steps[resource.name]
+                                for child in resource.children:
+                                    if child.name in steps:
+                                        child.workflow_step = steps[child.name]
+                    mark_not_configured_groups(groups, workflow_data)
+                    enrich_with_backfill_status(groups, workflow_data)
```
gregschohn: At this point, you should probably call this 'enrich with live status' and then within that function, break it up into RFS statuses - and maybe (eventually) live stats for proxy and replay.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+
+
+def _iter_nodes(nodes: List[Dict[str, Any]]):
+    """Recursively iterate all nodes in a tree."""
+    for node in nodes:
+        yield node
+        yield from _iter_nodes(node.get('children', []))
+
+
+def mark_not_configured_groups(sections: List[ResourceSection], workflow_data: Dict[str, Any]) -> None:
```
gregschohn: I don't have my head fully around this yet and I'm not concerned.
I'll give this advice, in case it  helps - you could make this the last post-processing step.  Scan over the completely augmented tree.  If there are empty groups, then you just slap the labels on there.  I can't tell - this code may actually be safer.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    console = Console()
+    has_any = any(
+        g.resources or g.not_configured
+        for s in sections for g in s.groups
+    )
+    if not has_any:
+        console.print("[dim]No migration resources found.[/dim]")
+        return
+
+    for section in sections:
```
gregschohn, line 228: I like the sections and groups ideas.  I like them being strongly typed too.
For code like this, it's a tax to have to know what level you're at.  If you had those data classes implement simpler interfaces (too), would you be able to print out status line + its children recursively and have less code?
Same comment as above - you need to so some amount of this - ask yourself if you've minimized it or not.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+def _add_resource_details(node, resource: ResourceNode) -> None:
+    """Add spec/status detail lines under a resource node."""
+    details = _format_spec_fields(resource)
+    if details:
+        node.add(f"[dim]{details}[/dim]")
+    if resource.depends_on:
+        deps = ", ".join(resource.depends_on)
+        node.add(f"[dim]Depends on: {deps}[/dim]")
+    if resource.live_status:
+        _add_live_status(node, resource.live_status)
```
gregschohn: is this synchronous or asynchronous?  Take a look at the existing status command.  There's a way that we pooled all of the live status call together so that we could kick them all off upfront and get them back ASAP rather than running them in serial.  From the cursory glance, this looks synchronous.  If it isn't let's describe that in the code better.  If it is, let's try to adjust the design.  I suspect that we'll want to pull this for every resource - and not have it take 30s if it could take 3.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
```diff
@@ -0,0 +1,497 @@
...
+    name = step['display_name']
+
+    if step.get('is_approval'):
+        reason = step.get('denial_reason') or ''
+        suffix = f" ({reason})" if reason else ""
+        label = f"[yellow]⟳ Workflow progress: WAITING FOR APPROVAL{suffix}[/yellow]"
+    elif phase == 'Succeeded':
+        label = f"[green]✓ Workflow progress: {name}[/green]"
+    elif phase in ('Running', 'Pending'):
+        label = f"[yellow]▶ Workflow progress: {name} ({phase})[/yellow]"
```
gregschohn: should these be aligned to use the formatting blocks above?
It's going to be weird if the workflow phase/states don't align intentionally and well with the resource ones

gregschohn (commented):
Good work.  Left some comments.  I'm fine if you want to address some of these next or move onto the diffs.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: The only use case for it now is to show `capture-proxy-topic` as a child of `kafka buffer` (kafka cluster). This was one of the feedback I got during discussion.

If we are removing the kafka cluster from the resource view (based on your other comment), then we have no use case for it and can be removed.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Added a TODO comment. Will follow up.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Changed to display a subtree from the workflow instead of just one step.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: This check has been removed. We added `resourceName` to all templates, simplified to just check `resourceName`.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Removed. We return the full subtree now, no default needed.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Updated to display a filtered subtree now.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Refactored to reuse `LiveCheckProcessor` from existing status command, which uses `ThreadPoolExecutor`. Removed custom backfill live status code.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: These code has been removed. We display a subtree of workflow steps now with the same formatting as existing status command.

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
oeyh: I refactored to reuse LiveCheckProcessor and have removed this custom backfill status code.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: This field has been removed. I now reuse LiveCheckProcessor and the format there, which contains timestamp.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Got it. Will consider this when integrating workflow config in the followup PR.

migrationConsole/lib/console_link/console_link/workflow/resource_tree.py
oeyh: Added resourceName as input to simplify the code here.

oeyh:
> I have one UX comment. The │   └── ✓ kafka-buffer (Ready) line should be excised. I (don't) see much value for it. Let me know if you think that I might be missing something.

This line indicates the status of the Kafka cluster, I can see some value in it especially if the workflow is configured to create the cluster. So I keep it there for now. Most comments should have been addressed now. Some are no longer applicable after refactor, others will be addressed in followup PRs.

The PR is ready for another look.

migrationConsole/lib/console_link/console_link/models/backfill_rfs.py
```diff
@@ -283,8 +283,29 @@ def get_detailed_status(target_cluster: Cluster, session_name: str) -> Optional[
     if (values.shard_total is None and values.shard_complete is None and
             values.shard_in_progress is None and values.shard_waiting is None):
         return "Shards are initializing"
-    
-    return "\n".join([f"Backfill {key}: {value}" for key, value in values.__dict__.items() if value is not None])
+
+    eta_str = _format_duration_ms(values.eta_ms) if values.eta_ms else "N/A"
+    start_time = str(values.started)[:19] if values.started else ''
+    finish_time = str(values.finished)[:19] if values.finished else ''
```
gregschohn, line 289: nit: Could you do something like `values.finished.isoformat(sep=" ", timespec="seconds")` - 19 isn't immediately obvious.

migrationConsole/lib/console_link/console_link/workflow/commands/crd_utils.py
```diff
@@ -90,6 +90,25 @@ def list_migration_resources(namespace, plurals=None):
     return results
 
 
+def list_migration_resources_full(namespace, plurals=None):
```
gregschohn, line 93: nit: plurals could be something like resource_type_filter.  The name of this function is also more specific than the plurals argument.  You could probably make list_migration_resources_full hardcode RESETTABLE_PLURALS and then call a more general list_resources_full(plurals) w/ no default.

migrationConsole/lib/console_link/console_link/workflow/tui/workflow_manage_app.py
```diff
@@ -314,6 +322,20 @@ def _show_logs_in_pager(self, node_data: Dict) -> None:
         if pod_name:
             self._logs.show_in_pager(self, pod_name, node_data.get('display_name', ''))
 
+    def action_view_resource_logs(self) -> None:
```
gregschohn, line 325: Let's consider something that handles the log COMMAND more directly. The TUI already does this with kubectl.  `LogManager.show_in_pager` and `follow_logs` both suspend the TUI and hand the terminal to kubectl/less. We didn't bother with `workflow output` (now `log`) before partially because we already had the pod name, but now we're expanding that to pick up more of the log command's (newer) functionality.

For the resource-level view, the TUI could do:

```python
def action_view_resource_logs(self) -> None:
    node = self.current_node_data
    if not node or not node.get('resource_path'):
        return
    resource_path = node['resource_path']
    with self.suspend():
        os.system('clear')
        cmd = f"workflow log resource {resource_path} | less -R"
        os.system(cmd)
```

We'll get consistency with the log command, less code, and importantly, remove the line buffering from the python manage process, which continues to spook me.  Sure, it might be in the subprocess, but that process has a shorter lifespan.  If it dies, it will be clearer why it has died than if the manage process runs out of memory.

That would also mean that you could drop the get_resource_logs() method, which has some replicated code.

migrationConsole/lib/console_link/console_link/workflow/commands/log.py
```diff
@@ -170,6 +170,49 @@ def _find_resource_object(namespace, resource_name):
     return None
 
 
+def get_resource_logs(namespace: str, resource_path: str) -> str:
```
gregschohn, line 173: See below.  We should drop this in lieu of just calling `workflow log` directly

migrationConsole/lib/console_link/console_link/workflow/commands/manage.py
```diff
@@ -95,15 +95,18 @@ def _initialize_k8s_client(ctx):
 @click.option('--namespace', default=get_current_namespace, hidden=True, envvar='WORKFLOW_NAMESPACE')
 @click.option('--insecure', is_flag=True, default=True, hidden=True, envvar='WORKFLOW_INSECURE')
 @click.option('--token', hidden=True, envvar='ARGO_TOKEN')
+@click.option('--step-view', is_flag=True, default=False,
```
gregschohn: I'll leave it up to you on when is the right time to flip this default.  We should just make sure that it's the same for status and manage commands.  I think that there are still some rough edges with the resource view, but I know that we're going to be patching them and that there are still rough edges with the workflow view (that aren't going to be smoothed)

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -476,10 +480,13 @@ def _run_live_checks_parallel(self, in_progress_nodes: List[Dict[str, Any]]) ->
 @click.option('--token', hidden=True, envvar='ARGO_TOKEN', help='Bearer token for authentication')
 @click.option('--all', 'show_all', is_flag=True, default=False,
               help='Show all workflows including completed ones (default: only running)')
-@click.option('--live-status', is_flag=True, default=False,
+@click.option('--live-status/--no-live-status', default=True,
```
gregschohn, line 487: This is a nice click feature that I was unaware of!

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -476,10 +480,13 @@ def _run_live_checks_parallel(self, in_progress_nodes: List[Dict[str, Any]]) ->
...
               help='Run a current status check for each snapshot and backfill still running')
+@click.option('--step-view', is_flag=True, default=False,
```
gregschohn: nit: Should this be renamed 'tree-view' or 'workflow-rollout-view'.

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -476,10 +480,13 @@ def _run_live_checks_parallel(self, in_progress_nodes: List[Dict[str, Any]]) ->
...
+              help='Show the Argo workflow step tree instead of the resource-centric view')
```
gregschohn: We should say "Show the current Argo Workflow's tree instead of ...".  It might be worth pointing out that such a view will not show historical actions to build out resources in a prior run.

We really should have done that before - it's arguably less critical now

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -489,7 +496,52 @@ def status_command(ctx, workflow_name, all_workflows, argo_server, namespace, in
     Example:
         workflow status
         workflow status --all
+        workflow status --step-view
     """
+    if not step_view:
+        try:
+            from ..resource_tree import (
```
gregschohn, line 507: nit: put this into a separate function & same for the next branch too

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -489,7 +496,52 @@ def status_command(ctx, workflow_name, all_workflows, argo_server, namespace, in
...
+            from ..models.utils import load_k8s_config
+            from ..tree_utils import build_nested_workflow_tree, filter_tree_nodes
+            load_k8s_config()
+            groups = build_resource_tree(namespace)
+
+            # Try to fetch Argo workflow data for progress info
+            workflow_unavailable = False
+            try:
+                service = WorkflowService()
+                fetcher = WorkflowDataFetcher(service, token)
```
gregschohn, line 520: this class already has `data_fetcher`, right?

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -489,7 +496,52 @@ def status_command(ctx, workflow_name, all_workflows, argo_server, namespace, in
...
+                    filtered_tree = filter_tree_nodes(tree_nodes)
+                    steps = extract_workflow_steps_by_resource(filtered_tree)
+                    for section in groups:
+                        for group in section.groups:
+                            for resource in group.resources:
+                                if resource.name in steps:
+                                    resource.workflow_progress = steps[resource.name]
+                                for child in resource.children:
+                                    if child.name in steps:
+                                        child.workflow_progress = steps[child.name]
```
gregschohn, line 536: A comment or variable name change to explain the types of the workflow_progress values would be helpful

gregschohn:
The ResourceTreeStateManager omits a number of useful features that are present in the TreeStateManager.  We should align the two.  I generated this document with kiro/claude as a quick sketch to show some of the remaining work and the impact if it isn't there.  Hopefully, this guide can help accelerate us making the changes.  This affects the `workflow manage` command only

# Task: Align `ResourceTreeStateManager` with `TreeStateManager` Conventions

## Context

The workflow manage TUI now supports a resource-centric view alongside its existing
step-centric view. The two are backed by separate state managers:

- `tui/tree_state_manager.py` — `TreeStateManager` (existing, mature)
- `tui/resource_tree_state_manager.py` — `ResourceTreeStateManager` (new)

The new manager works but diverges from the conventions established by the existing
manager in ways that create user-visible problems and architectural drift. This task
brings the new manager up to par.

## Files

Primary:
- `lib/console_link/console_link/workflow/tui/resource_tree_state_manager.py`
- `lib/console_link/console_link/workflow/tui/workflow_manage_app.py`
- `lib/console_link/console_link/workflow/resource_tree.py`

Reference (do not modify unless explicitly required):
- `lib/console_link/console_link/workflow/tui/tree_state_manager.py`
- `lib/console_link/console_link/workflow/tui/live_status_manager.py`
- `lib/console_link/console_link/workflow/tui/pod_name_manager.py`

## Goal

`ResourceTreeStateManager` should follow the same patterns as `TreeStateManager`:
incremental updates that preserve UI state, data fetching outside the manager,
shared helpers exposed publicly, and live-status checks executed off the main thread.

## Required Changes

### 1. Replace full-rebuild `update()` with incremental diffing

**Problem.** `update()` currently does `tree.clear()` and rebuilds from scratch every
refresh (every 3 seconds). This loses:
- cursor position (selection jumps to root)
- scroll position
- everything except a manually-collected set of collapsed-node IDs

It also causes flicker and re-evaluates dynamic key bindings against a freshly
recreated cursor node.

**Fix.** Mirror `TreeStateManager._update_recursive`:
- Build a stable identity for every level of the tree:
  - Sections: `f"section:{name}"`
  - Groups: `f"group:{display_name}"`
  - Resources: `f"resource:{plural}.{name}"` (use the `resource_path` already used
    elsewhere)
  - Workflow steps: existing Argo `id`
- At each level, diff `existing_children` (keyed by id) against the new node list:
  - If structural changes (insertions, removals, reordering), rebuild *that level
    only*, capturing and restoring expanded state.
  - Otherwise, walk in place: update labels only when relevant fields changed
    (phase, workflow_progress identity), recurse into children.
- Treat resource phase, child workflow steps, and live-check values as the inputs
  that warrant a label refresh; ignore cosmetic-only diffs.

### 2. Move data fetching out of the manager

**Problem.** `_build_sections` calls `build_resource_tree(namespace)` (which queries
the Kubernetes API) and runs `LiveCheckProcessor(...).enrich_tree_with_live_checks(...)`
synchronously on the main thread. The existing `TreeStateManager` is purely a view-state
manager.

**Fix.**
- Manager API mirrors `TreeStateManager`:
  ```python
  rebuild(self, sections: List[ResourceSection]) -> None
  update(self, sections: List[ResourceSection]) -> None
  ```
  Both accept already-built section data; no I/O inside the manager.
- Move the fetch + enrich logic into the app's data-fetch path
  (`_refresh_workflow_worker` / `_handle_workflow_data` in
  `workflow_manage_app.py`), running on the existing worker thread.
- `LiveCheckProcessor` invocation should run in the worker, with the enriched
  result delivered back to the main thread via `call_from_thread`.

### 3. Wire up `on_new_pod` for workflow-step subtrees

**Problem.** Workflow-progress subtrees inside the resource view contain Pod-type
Argo nodes (`data=step`). Without an `on_new_pod` callback firing,
`PodNameManager` never observes them, so actions like `action_follow_logs` and
`action_copy_pod_name` cannot resolve pod names from those nodes.

**Fix.**
- Add an `on_new_pod: Optional[Callable[[str], None]] = None` parameter to
  `ResourceTreeStateManager.__init__`, matching `TreeStateManager`.
- In `workflow_manage_app.py`, pass `self._pods.observe_node` when constructing
  the resource-view manager.
- Inside `_add_workflow_step` (and any place new Argo Pod nodes are added), call
  `self._on_new_pod_handler(step['id'])` for `type == 'Pod'` nodes that are new.
- During incremental updates, only fire for newly-added nodes, matching the
  pattern in `_update_recursive`.

### 4. Pass workflow data through for step labels

**Problem.** Step labels in the resource view always render with
`status_output=None`, so any step that would otherwise show a status-output
indicator does not.

**Fix.**
- Store `_workflow_data` on the manager (as `TreeStateManager` does).
- In the workflow-step rendering path, compute
  `status_output = get_step_status_output(self._workflow_data, step['id'])` and
  pass it to `get_step_rich_label`.
- Plumb `workflow_data` into `rebuild`/`update` alongside `sections` (or attach
  it to the `ResourceSection` payload — pick whichever keeps the API symmetric
  with `TreeStateManager`).

### 5. Stop reaching into private helpers in `resource_tree.py`

**Problem.** `resource_tree_state_manager.py` imports underscored helpers from
`resource_tree.py`:
`_format_spec_fields`, `_has_notable_steps`, `_collect_notable_steps`,
`_find_last_succeeded`, `_step_timestamp`, `_maybe_rewrite_wait_step`,
`DISPLAY_NAMES`, `DISPLAY_PHASES`. Several imports are also done inside method
bodies.

**Fix.**
- Promote the genuinely shared helpers to public names in `resource_tree.py`
  (drop the leading underscore) or move them into a new
  `workflow/resource_tree_helpers.py` module.
  Candidates: `format_spec_fields`, `has_notable_steps`, `collect_notable_steps`,
  `find_last_succeeded`, `step_timestamp`, `maybe_rewrite_wait_step`,
  `DISPLAY_PHASES`.
- Move all imports to the top of `resource_tree_state_manager.py`. Resolve any
  resulting circular-import issues by extracting shared types/constants into a
  neutral module rather than papering over with lazy imports.

### 6. Unify the two rendering passes (optional but recommended)

`resource_tree.py` renders sections to a Rich `Tree` (CLI); the manager renders
the same sections to a Textual `Tree` (TUI). The walk logic is duplicated.

If time permits, introduce a small builder that walks a `List[ResourceSection]`
and emits `(label, data, children)` tuples (or invokes a callback per node).
Both the Rich renderer and Textual manager become thin adapters. Keep this
refactor scoped — do not block tasks 1–5 on it. If skipped, leave a TODO
referencing this task.

## Acceptance Criteria

- [ ] Selecting a resource node in the TUI keeps that selection across refreshes.
- [ ] Scroll position is preserved across refreshes.
- [ ] Expanding/collapsing a node is preserved across refreshes (already works
      partially; should remain).
- [ ] No `tree.clear()` call inside `update()`; structural changes are localized
      to the changed level.
- [ ] No Kubernetes calls or live-check execution on the main thread inside the
      manager. `_build_sections` (or equivalent fetch logic) runs in a worker.
- [ ] `action_follow_logs` and `action_copy_pod_name` work on Pod-type Argo
      nodes inside the resource view's workflow-progress subtree.
- [ ] Workflow steps under resources show status-output indicators when present
      (verify by inspecting a workflow with `statusOutput` parameters).
- [ ] No imports inside method bodies in `resource_tree_state_manager.py`.
- [ ] No imports of underscored names from `resource_tree.py` outside the module
      itself.
- [ ] Existing CLI behavior (`workflow status` resource view) is unchanged.
- [ ] Existing step-view TUI behavior (`workflow manage --step-view`) is
      unchanged.
- [ ] All workflow tests still pass.

## Verification Steps

1. Run the existing test suite:
   ```
   cd lib/console_link
   pytest tests/workflow-tests
   ```
2. Manually exercise the resource view:
   - Launch `workflow manage` against a running workflow.
   - Select a non-root node, confirm selection persists across at least three
     refresh cycles.
   - Expand a workflow-progress subtree, confirm expansion persists.
   - Locate a Pod-type step under a resource, press the follow-logs and
     copy-pod-name keys; both must resolve the pod name.
   - Verify a step with a `statusOutput` parameter shows its indicator.
3. Manually exercise the step view (`workflow manage --step-view`) to confirm
   no regressions.
4. Run `workflow status` (resource view) and `workflow status --step-view` to
   confirm CLI parity.

## Non-Goals

- Do not change the resource-view feature set (groupings, sections, displayed
  fields).
- Do not change `TreeStateManager`.
- Do not refactor `LiveCheckProcessor` itself; only its invocation site moves.
- Do not address the separate `get_resource_logs` duplication in
  `commands/log.py` — that is tracked as a different task.

## Pitfalls

- Textual's `TreeNode.remove()` is destructive; capture expand state and child
  data before removing during a level rebuild, exactly as
  `_update_recursive` does.
- Workflow-progress subtrees can be re-evaluated by
  `_collect_notable_steps`/`_find_last_succeeded` between refreshes, which can
  legitimately reorder children. Treat that as a structural change at the
  workflow-progress level only; do not let it cascade into clearing the parent
  resource node.
- `is_ephemeral` is used in `TreeStateManager` to skip dynamic children during
  diffing. The resource view does not currently add ephemeral children; if you
  introduce any (e.g., live-check value lines), mark them so they are excluded
  from diff identity.
- When moving live-check execution into the worker, ensure the result is merged
  into `sections` *before* `call_from_thread(self._handle_workflow_data, ...)`
  so the manager only ever sees fully-enriched data.

migrationConsole/lib/console_link/console_link/workflow/tui/resource_tree_state_manager.py
```diff
@@ -0,0 +1,212 @@
...
+    mark_not_configured_groups, PHASE_SYMBOLS, RESOURCE_SECTIONS,
+)
+from console_link.workflow.tree_utils import (
+    get_step_rich_label,
+    build_nested_workflow_tree, filter_tree_nodes,
+)
+from console_link.workflow.commands.status import LiveCheckProcessor, ConfigConverter
+
+
+class ResourceTreeStateManager:
```
gregschohn, line 19: See https://github.com/opensearch-project/opensearch-migrations/pull/2894#issuecomment-4501570810 for overall comments about how we can bring this up to the level of the other tree state manager (for workflows only)

gregschohn (approved):

PR #2316, opened 2026-03-01
Enabling strimzi for workflows

gregschohn:
### Check List
- [ ] New functionality includes testing
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2905, opened 2026-05-06
Apply OnError retry to one-shot container templates

jugal-chauhan:
### Description
Add `CONTAINER_TEMPLATE_RETRY_STRATEGY` (OnError, limit 3) and apply it to six one-shot `.addContainer` templates that were missing retry policies. Also harden the two approval-gate kubectl scripts so real failures surface instead of being swallowed.

Extends the `.addResourceTask` retry pattern established by PR #2316 to direct container templates. New templates added by PR #2834 (`addApprovalGateOwnerReferences`, `cleanupApprovalGates`) are also covered.

#### 1. New retry constant
`commonUtils/resourceRetryStrategy.ts` — adds `CONTAINER_TEMPLATE_RETRY_STRATEGY` (`OnError`, `limit: 3`, same backoff as the existing resource-task constant). Separate constant because container reruns are more expensive than CR re-apply, so lower limit.

#### 2. `.addRetryParameters(...)` applied to 6 templates

| File | Template |
|---|---|
| `createSnapshot.ts` | `runCreateSnapshot` |
| `metadataMigration.ts` | `runMetadata` |
| `fullMigration.ts` | `addApprovalGateOwnerReferences` |
| `fullMigration.ts` | `cleanupApprovalGates` (onExit) |
| `testMigrationWithWorkflowCli.ts` | `configureAndSubmitWorkflow` |
| `testMigrationWithWorkflowCli.ts` | `evaluateWorkflowResult` |

#### 3. Approval-gate script hardening (`fullMigration.ts`)
Added `set -euo pipefail`; removed `|| echo WARN` and `|| true` fallbacks.

**Impact:**
- Transient pod/node failures → `OnError` retries (≤3 attempts).
- Real container errors (RBAC denial, API rejection, malformed patch) → fail
  the step, surface as workflow failure. Previously silently returned success.

### Templates explicitly NOT touched
- **`runMigrationCommandForStatus`** — wrapped by `Always, limit: 200` steps-templates; inner retry would compound to ~800 attempts on a scheduling blip.
- **`monitorWorkflow`** — intentional `Always` polling retry.
- **`checkSnapshotStatusInternal`** / **`waitForCompletionInternal`** — the `Always, limit: 200` wrappers already cover both failure classes.
- **`runConsole`** — interactive launcher.

### Idempotency (safe to rerun on OnError)
- `runCreateSnapshot` — `noWait: true`; snapshot repo is source of truth.
- `runMetadata` — `--evaluate` read-only; `--migrate` uses idempotent PUT APIs.
- `addApprovalGateOwnerReferences` — `kubectl patch --type merge` is idempotent.
- `cleanupApprovalGates` — ownerReferences GC the gates even if cleanup fails, so hard failure surfaces the issue without leaking resources.

### Issues Resolved
N/A

### Testing
- `npm run type-check`: clean
- `npx jest` under `migration-workflow-templates/`: 15/15 pass
- test snapshots updated

### Check List


orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -106,13 +107,16 @@ export const FullMigration = WorkflowBuilder.create({
...
             .addCommand(["/bin/bash", "-lc"])
             .addArgs([`
+set -euo pipefail
+
 selector='migrations.opensearch.org/workflow={{workflow.name}}'
 patch='{"metadata":{"ownerReferences":[{"apiVersion":"argoproj.io/v1alpha1","kind":"Workflow","name":"{{workflow.name}}","uid":"{{workflow.uid}}"}]}}'
+
 kubectl get approvalgates.migrations.opensearch.org -l "$selector" -o name \\
-  | xargs -r -n 1 kubectl patch --type merge -p "$patch" \\
-  || echo "WARN: failed to patch one or more approvalgate ownerReferences" >&2
```
gregschohn, line 113: This line could be better, but is there a downside to leaving it in?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -123,11 +127,14 @@ kubectl get approvalgates.migrations.opensearch.org -l "$selector" -o name \\
             .addResources(DEFAULT_RESOURCES.SHELL_MIGRATION_CONSOLE_CLI)
             .addCommand(["/bin/bash", "-lc"])
             .addArgs([`
+set -euo pipefail
```
gregschohn: do we need this for a single command?

gregschohn (approved):
Approving, but there are a couple minor bits of feedback on changes to the shell scripts themselves.

jugal-chauhan:
> Approving, but there are a couple minor bits of feedback on changes to the shell scripts themselves.

Ill address the feedback in a follow-up PR

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
jugal-chauhan: The `|| echo WARN >&2` forces the pipeline to `exit 0` regardless of kubectl's actual exit code. That's exactly the silent-failure behavior this PR is removing. kubectl's own error output is descriptive enough when patch fails

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
jugal-chauhan: Good catch! Dropped in aea68e72b. The behavioral fix comes entirely from removing `|| true`.
For a single `kubectl` command `set -euo pipefail` doesn't change anything.

jugal-chauhan:
> > Approving, but there are a couple minor bits of feedback on changes to the shell scripts themselves.
> 
> Ill address the feedback in a follow-up PR

Addressed the comments in this PR itself as followup commits

PR #2906, opened 2026-05-06
Add TCP readinessProbe to capture proxy Deployment to prevent workflow marking crashed pods as Ready

jugal-chauhan:
### Problem
The Argo workflow step `deployProxyDeployment` (and its TLS variant) uses `successCondition: "status.readyReplicas > 0"` to determine when the capture proxy is healthy. Without a `readinessProbe`, Kubernetes marks the container Ready immediately on start. If the proxy then crashes during initialization (e.g. TLS key loading fails), the workflow step has already succeeded and `patchCaptureProxyReady` runs, falsely reporting the proxy as healthy while the pod is in `CrashLoopBackOff`

### Changes Introduced

- **`setupCapture.ts`** — Add a `tcpSocket` `readinessProbe` on `listenPort` to both proxy Deployment manifests (TLS and non-TLS). The proxy only binds the port after SSL context initialization succeeds, so the probe naturally gates readiness behind full init.
- **`setupCapture.ts`** — Add `minReadySeconds: 2` to the Deployment spec to guard against a pod that flips Ready once and immediately crashes.
- **`readinessProbe.test.ts`** — New regression test asserting both proxy Deployment manifests include a `tcpSocket` probe targeting the container's declared `containerPort` and a non-zero `minReadySeconds`.
- **`setupCapture.snap.json`** — Regenerated to reflect the manifest changes.

### Issues Resolved
[MIGRATIONS-3042](https://opensearch.atlassian.net/browse/MIGRATIONS-3042)

### Testing
- Java compiles (`./gradlew :TrafficCapture:trafficReplayer:compileJava`)
- Snapshot tests updated

### Check List


orchestrationSpecs/packages/migration-workflow-templates/tests/readinessProbe.test.ts
```diff
@@ -0,0 +1,55 @@
+import * as yaml from "js-yaml";
+import {renderWorkflowTemplate} from "@opensearch-migrations/argo-workflow-builders";
+import {SetupCapture} from "../src/workflowTemplates/setupCapture";
+
+/** Ensures capture proxy Deployments have a readinessProbe that gates on real init. */
+describe("Capture proxy Deployments declare a readinessProbe", () => {
```
gregschohn, line 6: do the snapshot tests do this pretty much already?

gregschohn (approved):

orchestrationSpecs/packages/migration-workflow-templates/tests/readinessProbe.test.ts
jugal-chauhan: No, this is an additional test specific to capture proxy's readinessProbe

PR #2907, opened 2026-05-06
Sourceless: copy_to handling, _source filters, typed recovery API (CCE fix), and hot-path simplification

AndreKurait:
## Summary

Fixes three sourceless-reconstruction bugs surfaced on real migration snapshots, types the points/terms recovery API to eliminate a latent unchecked-cast crash, adds support for `_source.includes` / `_source.excludes` directives, and folds in a structural simplification + hot-path optimization pass over the same diff. Single unified gate, single codepath.

### 1. `copy_to` target fields leaking into reconstructed `_source`

`copy_to` destinations are index-time-only in ES/OS — they never appear in `_source` on the origin. The reconstructor was emitting them anyway because they have real stored/doc_values data in the segment. Reindexing the reconstructed doc would double-populate the targets.

### 2. Source fields unrecoverable from their own chain

When a source field has `store:false`, no `doc_values`, and `_source` is disabled, the stored → doc_values → points → terms chain returns nothing — but its `copy_to` *target* still has the data. A new Pass 5 reverse-derives the source's value from its lowest-lossiness target (constant_keyword → keyword/numeric/date → match_only_text → text), routing the recovered value through the source mapping for final JSON shape (date formatting, scaled_float, etc.).

To avoid an O(all-fields) walk on indices with thousands of leaves, Pass 5 iterates `mappingContext.getCopyToSourceFields()` — only fields with declared `copy_to` edges.

### 3. ES 5.6 / Lucene 6 text recovery returning only the alphabetically-first token

`LeafReader6` didn't override the term-position walk, so multi-token text fields were collapsing to one term. Fixed (and harmonized across LeafReader 5/7/9/10 with a `hasPositions` guard so requesting POSITIONS on a non-positional postings list doesn't yield bogus sentinel values).

### 4. `_source.includes` / `_source.excludes` ignored

The origin mapping's source-filter intent was silently dropped. Now parsed off the mapping and applied by the same gate that strips `copy_to` targets. Globs handled: `*`, `<prefix>.*`, `<prefix>.**`, bare-`*` patterns (`prefix*`, `*suffix`, `prefix*suffix`), and literal paths. Literal patterns match the path itself **and any descendant** — e.g. `excludes:["meta"]` drops `meta`, `meta.created_at`, `meta.updated_by`, `meta.audit.user`, mirroring ES's recursive-descent filtering (a `metadata` field is unaffected because the matcher uses `path startsWith "meta."`, not `"meta"`). Multi-`*`/`?` patterns fall back to literal match with a debug log. Excludes wins over includes (ES semantics). Pass 5 respects the same gate so an excluded field isn't resurrected through its `copy_to` target.

### 5. ClassCastException on copy_to reverse-derive into multi-valued doc_values

```
java.lang.ClassCastException: class java.lang.String cannot be cast to class [B
    at SourceReconstructor.decodePointValue(SourceReconstructor.java:937)
    at SourceReconstructor.convertFallbackValue(SourceReconstructor.java:886)
    at SourceReconstructor.populateFromSegment(SourceReconstructor.java:499)
    at SourceReconstructor.mergeWithDocValues(SourceReconstructor.java:363)
```

Two cooperating bugs:

- `convertFallbackValue` accepted `Object` and used `instanceof List<?>` to detect points-fallback shape, then unchecked-cast to `List<byte[]>`. Any other List that landed there — `List<String>` from `SORTED_SET` doc_values, `List<Long>` from `SORTED_NUMERIC` — passed `instanceof`, the cast succeeded silently, and the failure surfaced only on the first `byte[]` dereference inside `decodePointValue`.
- `probeFieldValue` (pass 5) returned an `Object` that was already shaped by the **target** mapping for tiers 1/2/4 (stored, doc_values, constant) but was raw points/terms output for tier 3. The pass-5 caller fed all of those back into `convertFallbackValue` with the **source** mapping, double-converting tiers 1/2/4 and only by accident crashing on the multi-valued doc_values shape.

Fixed by typing the recovery channel:

- `RecoveredValue { PointBytes, NumericTerm, TextTerm }` replaces the untyped `Optional<Object>` from `getValueFromPointsOrTerms`. The variants make the three actual recovery channels explicit; `convertFallbackValue` switches on them, so `List<String>` from doc_values cannot reach the byte-decoder branch by mistake — the type system rules it out.
- `ProbeResult { Final, Raw }` tags `probeFieldValue`'s output so the copy_to caller knows whether the value still needs source-mapping conversion. Tiers 1/2/4 yield `Final` (write verbatim). Only tier 3 yields `Raw`, and only `Raw` goes through `convertFallbackValue` with the source mapping. No more double-conversion path.

## Design

`FieldMappingContext` exposes one method:

```java
public boolean isSourceExcluded(String fieldPath)
```

returns `true` if the path is a `copy_to` target, matches a `_source.excludes` glob, or (when `_source.includes` is set) matches no include glob. `SourceReconstructor.shouldSkipField` calls it in passes 1–4 (stored, doc_values, points, terms, constant_keyword) and again in pass 5 (copy_to reverse-derive). Cheap no-op when neither `copy_to` nor a `_source` filter is declared.

`LuceneLeafReader.getValueFromPointsOrTerms` now returns `Optional<RecoveredValue>`. `SourceReconstructor.probeFieldValue` returns `ProbeResult` (a sealed `Final | Raw` sum type). The two sealed types are the load-bearing change: every points/terms decoder branch is now compile-time checked against the channel that produced it, and the copy_to caller can no longer feed already-converted target values back through a points decoder.

### Hot-path performance (per-document)
- **Cache `getDocValueFields()` per LeafReader** (5/6/7/9/10) — was rebuilt on every doc; now memoized via `volatile List<DocValueFieldInfo>`. Replaced the term-list materialization in boolean detection with a short-circuiting `fieldHasOnlyBooleanTerms` walking the `TermsEnum` directly.
- **`RecoveryContext` with O(1) lookups** — `populateFromSegment` and `probeFieldValue` now share a per-doc context that lazy-builds `Map<String, LuceneField>` from `document.getFields()` and `Map<String, DocValueFieldInfo>` from the cached reader fields. Eliminates the per-`copy_to`-target linear scans.
- **Cached dotted-path splits** — `FieldMappingContext.getPathSegments()` returns a pre-split `String[]` for mapped fields (cached at construction); arbitrary names use `splitPath` (manual single-pass, no `Pattern`). Replaces all `String.split("\\.")` calls.
- **Pre-compiled source-filter globs** — `_source.includes/excludes` stored as `List<Predicate<String>>`; `compileGlob` runs once at construction.
- **Pre-sorted copy_to targets / pre-filtered constants** — sort by lossiness once at construction; constant_value entries pre-extracted so pass 4 walks O(constant fields) instead of O(all fields).
- **`couldContributeToMerge()` fast-path** — `mergeWithDocValues` returns the original bytes (no JSON round-trip) when no recovery tier could contribute.

### Code quality
- Tighten `LuceneLeafReader.getDocValueFields()` return type from `Iterable` to `List`.
- Unified `walkMapAncestors` primitive shared between `descendsIntoExistingObjectArray` and (effectively) `hasNested`.
- Promoted ES field-type names to `EsFieldType.MappingTypes` constants (`constant_keyword`, `text`, `half_float`, `float`, `double`, `wildcard`, `integer`, `short`, `byte`, `match_only_text`, etc.).
- Trimmed verbose Javadocs that explained WHAT the code does; kept WHY-only callouts (hidden constraints, ordering caveats, version-specific quirks).
- Added missing `@Override` on new `toString()` methods (LeafReader9, LeafReader10).
- Removed a dead `JsonNode tree = ...` assignment from `SourceReconstructorTest`.

## Tests

**Unit** — `FieldMappingContextTest` (19 cases, all green): no-filters baseline, copy_to-destination exclusion (scalar + array form), `_source.excludes` exact path + `.*` glob + `.**` glob, bare-name parent-path exclusion + dotted-literal parent-path exclusion (with `metadata`-vs-`meta` prefix-collision negative case), `_source.includes` filtering, excludes wins over includes, `_source.includes:"*"` includes everything, mid-string glob falls back to literal, copy_to + `_source.excludes` combined.

**Unit** — `SourceReconstructorTest` (new cases):
- `isSourceExcluded_supportsBareStarPrefixSuffixGlobs` — covers `prefix*`, `*suffix`, `prefix*suffix` with positive and negative cases.
- `getCopyToSourceFields_returnsOnlyFieldsWithCopyToEdges` — pins the pass-5 iterator to the strict copy_to source subset.
- 7 copy_to handling tests: parse copy_to-as-string, copy_to-as-array, lossiness ranking (keyword before text), target stripping, reverse-derive when source has no data, less-lossy-target preference, fallback to text target, source-own-value-wins.
- `reverseDerive_recoversSourceFromSortedSetDocValuesTarget_noClassCast` — regression for the CCE: copy_to source whose target resolves via SORTED_SET doc_values. Would have crashed on the old code; now passes.

**Integration** — `NoStoredSourceMigrationTest` (`@Tag("isolatedTest")`) across the ES 5.6 / 6.8 / 7.10 / 7.17 / 8.17 / OS 1.3 / OS 2.19 → OS 3.5 matrix:
- `testCopyToFieldHandling` — Enron-style mapping; asserts (a) `users.*` targets are stripped, (b) `from` recovers via its own keyword chain, (c) `sender` (text, no doc_values, no store) reverse-derives from `users.sender` keyword target.
- `testSourceExcludesFieldHandling` — `_source.excludes:["secret","meta.*"]`; asserts `title` survives, `secret` is suppressed, all `meta.*` children are suppressed through full snapshot → reconstruct → reindex → search.

## Out of scope

The BKD point-values per-doc full scan (`compare()` returning `CELL_CROSSES_QUERY`) flagged by the efficiency review is left for a follow-up: a `SegmentTermIndex`-style per-segment cache would help but touches all five LeafReaders' point-traversal logic and warrants its own focused PR with benchmarks.

## Check List

- [x] All tests pass


SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/FieldMappingContext.java
```diff
@@ -111,4 +221,138 @@ public FieldMappingInfo getFieldInfo(String fieldName) {
...
+            return path.startsWith(glob.substring(0, glob.length() - 2));
+        }
+        if (glob.endsWith(".*")) {
+            return path.startsWith(glob.substring(0, glob.length() - 1));
+        }
+        if (glob.indexOf('*') >= 0 || glob.indexOf('?') >= 0) {
+            log.debug("Unsupported source-filter glob '{}', treating as literal", glob);
+        }
+        return path.equals(glob);
+    }
```
jugal-chauhan: nit : The glob matching handles *, <prefix>.*, and <prefix>.** but not <prefix>* (without the dot).

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -263,9 +270,122 @@ private static boolean populateFromSegment(Map<String, Object> target, LuceneLea
...
+     * the first non-null value found, or {@code null} if nothing matches. Used by the copy_to
+     * reverse-derivation path (pass 5) to probe a TARGET field without mutating the reconstructed
+     * source map — the caller decides how to route and convert the value.
+     *
+     * Unlike the primary passes, this does NOT apply type conversion — it returns the raw value
+     * from whichever tier hit. The caller runs it through the SOURCE field's mapping for final
+     * JSON shaping.
+     */
+    private static Object probeFieldValue(LuceneLeafReader reader, int docId, LuceneDocument document,
+            String fieldName, FieldMappingContext mappingContext, SegmentTermIndex termIndex) throws IOException {
```
jugal-chauhan: `probeFieldValue` is a great and clean addition! The tier ordering (stored → doc_values → points/terms → constant) mirrors the primary passes correctly.

For the 5th Pass, the reverse-derive loop iterates `mappingContext.getFieldNames()` which is all mapped fields. 

For indices with thousands of fields, this is O(fields × targets) per document. In practice, several fields won't have `copy_to` so `rankedTargets.isEmpty()` short-circuits early, this is good.

But is there a way to iterate only fields that HAVE `copy_to` edges? e.g., `mappingContext.getCopyToSourceFields()` returning just the keys of `copyToBySource`. Would avoid the `getFieldNames()` full scan + `hasNested` check for every field.

jugal-chauhan (commented):
Thanks for bringing in these changes! A couple of minor nits mentioned below. Happy to approve this once #2886 merges in (it will cause merge conflicts over here which will then need to be resolved)

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/FieldMappingContext.java
```diff
@@ -111,4 +221,158 @@ public FieldMappingInfo getFieldInfo(String fieldName) {
...
+                if (matchesGlob(glob, fieldPath)) {
+                    return false;
+                }
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private static boolean matchesGlob(String glob, String path) {
```
jugal-chauhan: In ES/OS, `_source.excludes: ["meta"]` excludes the entire meta object tree including `meta.created_at`, `meta.updated_by`, etc. The current matchesGlob treats `"meta"` as a literal and only matches `path.equals("meta")`, so `"meta.created_at"` is NOT excluded.

A user migrating an index with `_source.excludes: ["meta"]` may be surprised to find `meta.created_at` resurrected in the migrated docs

jugal-chauhan (approved):
LGTM! I have a minor concern regarding the intention behind the `matchesGlob` method in `FieldMappingContext.java` which currently does not handle parent-path exclusion. This can also be addressed in a followup PR.

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/FieldMappingContext.java
AndreKurait: Good catch — fixed in 3e64efda6 rather than as a follow-up since it's a one-line semantics fix.

The literal branch of `compileGlob` is now `path.equals(glob) || path.startsWith(glob + ".")`, mirroring ES's recursive-descent filtering. So `_source.excludes: ["meta"]` now drops `meta`, `meta.created_at`, `meta.updated_by`, `meta.audit.user`, etc. The trailing dot avoids a `metadata` prefix-collision against a `meta` rule.

Two new tests cover this: `testSourceExcludesBareNameSuppressesDescendants` (bare-name + the `metadata` negative case) and `testSourceExcludesDottedLiteralSuppressesDescendants` (dotted-literal parent-path with sibling-survival).

jugal-chauhan: Thanks for adding these specific test scenarios!

AndreKurait:
Superseded by https://github.com/opensearch-project/opensearch-migrations/pull/2985

PR #2908, opened 2026-05-06
fix(rfs): respect copy_to destinations and _source includes/excludes in reconstruction

AndreKurait:
## Problem

The sourceless reconstructor walks every stored field, doc_value, and indexed point it can reach, then serializes all of them into the rebuilt `_source`. That's the wrong shape for two categories of field that exist on the indexed side but never existed in the origin `_source`:

- **`copy_to` destinations.** ES writes `copy_to` output only to the inverted index and doc_values. It never writes it back into `_source`. A mapping with `{title, body, copy_to: "all_text"}` produces a document whose ES `_source` contains `{title, body}` but whose reconstructed OS `_source` (prior to this fix) contained `{title, body, all_text}`.
- **`_source.excludes` / `_source.includes` paths.** If the origin mapping declares `_source.excludes: ["secret"]`, ES never stores `secret` in `_source`, yet the field is typically still indexed — so the reconstructor emitted it.

The practical impact: an extraneous top-level object appears in the reconstructed document whenever the origin mapping funnels many leaves into a shared `copy_to` sink (for example, when an index maps per-bucket sub-fields to an umbrella "all" path for cross-field search).

## Before / after

Mapping:
```json
{
  "_source": {"excludes": ["secret"]},
  "properties": {
    "title":    {"type": "text", "copy_to": "all_text"},
    "body":     {"type": "text", "copy_to": "all_text"},
    "all_text": {"type": "text"},
    "secret":   {"type": "keyword"}
  }
}
```

Indexed document on origin: `{title, body, secret}` (the original `_source` — `all_text` never existed in `_source`; `secret` was filtered out).

| Path     | Origin `_source` | Reconstructed `_source` (before) | Reconstructed `_source` (after) |
|----------|:---:|:---:|:---:|
| title    | ✓ | ✓ | ✓ |
| body     | ✓ | ✓ | ✓ |
| all_text | — | **leaked** | — |
| secret   | — | **leaked** | — |

## Fix

- `FieldMappingContext` now parses `copy_to` at every mapped leaf and the index-level `_source.includes` / `_source.excludes` block, and exposes `isSourceExcluded(path)` as a single gate that combines copy_to-sink detection with include/exclude glob matching. Cheap no-op when neither directive is declared.
- `SourceReconstructor.shouldSkipField` consults that gate as its final decision. All existing call sites (stored fields, doc_values, points, terms) pick up the new behaviour without further changes.

Glob matching is a pragmatic subset sufficient for the common cases: exact paths, `"*"`, trailing `".*"`, trailing `".**"`. Mid-string wildcards fall back to literal match with a debug log so they don't silently drop matches.

## Tests

`RFS/.../FieldMappingContextTest`:
- `testNoFiltersMeansNothingExcluded` — no `copy_to` / `_source` filter declared → `isSourceExcluded` is a no-op.
- `testCopyToDestinationExcluded` — scalar and array-form `copy_to` destinations are excluded; source fields aren't.
- `testSourceExcludes` — exact-path and trailing-`.*` globs.
- `testSourceIncludes` — include-list gate; paths outside are excluded.
- `testSourceExcludesWinsOverIncludes` — ES semantics: excludes is applied after includes.

`SearchSnapshotExtractor/.../SourceReconstructorTest.reconstructSource_respectsCopyToAndSourceExcludes` — end-to-end behavioural test through `reconstructSource`: builds a segment whose stored fields include the `copy_to` sink and the `_source.excludes` field, runs the full reconstruction path, and asserts the rebuilt `_source` is shaped exactly `{title, body}`.

All 14 tests green locally:
```
./gradlew :SearchSnapshotExtractor:test --tests '*FieldMappingContext*' --tests '*SourceReconstructor*' :RFS:test --tests '*FieldMappingContext*'
```

## Scope

No behaviour change when the mapping declares neither `copy_to` nor `_source` filters. No changes outside the two production files and their two tests. No build, config, or unrelated refactors.

### DIFF
--- a/RFS/src/test/java/org/opensearch/migrations/bulkload/lucene/FieldMappingContextTest.java
+++ b/RFS/src/test/java/org/opensearch/migrations/bulkload/lucene/FieldMappingContextTest.java
@@ -133,4 +133,94 @@ class FieldMappingContextTest {
 
         assertTrue(context.getFieldNames().isEmpty());
     }
+
+    @Test
+    void testNoFiltersMeansNothingExcluded() throws Exception {
+        String json = """
+            {
+                "properties": {
+                    "title": {"type": "text"}
+                }
+            }
+            """;
+        FieldMappingContext context = new FieldMappingContext(MAPPER.readTree(json));
+        assertFalse(context.isSourceExcluded("title"));
+        assertFalse(context.isSourceExcluded("anything.else"));
+    }
+
+    @Test
+    void testCopyToDestinationExcluded() throws Exception {
+        String json = """
+            {
+                "properties": {
+                    "title": {"type": "text", "copy_to": "all_text"},
+                    "body":  {"type": "text", "copy_to": ["all_text", "combined.text"]},
+                    "all_text": {"type": "text"}
+                }
+            }
+            """;
+        FieldMappingContext context = new FieldMappingContext(MAPPER.readTree(json));
+        assertTrue(context.isSourceExcluded("all_text"), "copy_to destination must be excluded");
+        assertTrue(context.isSourceExcluded("combined.text"), "array-form copy_to entries must be excluded");
+        assertFalse(context.isSourceExcluded("title"), "source fields must not be excluded");
+        assertFalse(context.isSourceExcluded("body"), "source fields must not be excluded");
+    }
+
+    @Test
+    void testSourceExcludes() throws Exception {
+        String json = """
+            {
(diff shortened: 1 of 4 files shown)


AndreKurait:
Superseded by #2907 — the _source.includes / _source.excludes work from this PR has been folded into the unified source-exclusion gate there, alongside the copy_to fix, reverse-derivation (pass 5), and the ES 5.6 / Lucene 6 multi-token text recovery fix. One codepath, one test surface. Closing.

PR #305, opened 2023-09-08
Minor fixes to handle and propagate errors that had been getting lost through our system and into logs better.

gregschohn:
### Description
* Category (Enhancement, New feature, Bug fix, Test fix, Refactoring, Maintenance, Documentation)
* Why these changes are required?
* What is the old behavior before changes and new behavior after changes?

### Issues Resolved
[List any issues this PR will resolve]

Is this a backport? If so, please add backport PR # and/or commits #

### Testing
[Please provide details of testing done: unit testing, integration testing and manual testing]

### Check List
- [ ] New functionality includes testing
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #304, opened 2023-09-08
Minor update to Copilot readme for setting up an additional cluster

lewijacn:
### Description
Minor update to Copilot readme for setting up an additional cluster

### Issues Resolved
None

### Testing
N/A

### Check List
- [ ] New functionality includes testing
  - [ ] All tests pass, including unit test, integration test and doctest
- [ ] New functionality has been documented
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #2914, opened 2026-05-07
Replace coarse Job backoffLimit with podFailurePolicy + per-step retries

jugal-chauhan:
### Problem

Four Jobs in the `migrationAssistantWithArgo` Helm chart were declared with `backoffLimit: 1` (and `0` for `delete-s3-bucket`). These are the pre-install, pre-delete, and post-delete hooks that drive the entire Helm release:

| File | Job | Hook |
|---|---|---|
| `childrenChartInstaller/installJob.yaml` | `{release}-helm-installer` | `pre-install,pre-upgrade` |
| `childrenChartInstaller/uninstallJob.yaml` | `{release}-helm-uninstaller` | `pre-delete` |
| `childrenChartInstaller/uninstallJob.yaml` | `{release}-helm-uninstaller-logging` | `post-delete` |
| `resources/s3/deleteS3Bucket.yaml` | `delete-s3-bucket-{release}` | `pre-delete` |

The installer script runs ~10 serial steps (helm repo update, `kubectl wait` for the Kyverno admission webhook, CRD readiness, a canary policy probe, then 15+ parallel `helm upgrade --install`). Any single transient flake like a Kyverno webhook TLS startup race, or LocalStack wobble, or ECR throttling, or even a DNS blip can make the pod exit non-zero, and the Job is marked `BackoffLimitExceeded` after one retry, failing the whole Helm release.

#### Recent evidence from Jenkins `main/`

| Build | Job that hit `BackoffLimitExceeded` | Result |
|---|---|---|
| `main-k8s-local-integ-test` #303 | `delete-s3-bucket-ma` | ABORTED |
| `main-k8s-local-integ-test` #304 | `delete-s3-bucket-ma` | ABORTED |
| `main-k8s-local-integ-test` #305 | `ma-helm-installer` | FAILURE |
| `main-k8s-local-integ-test` #307 | `ma-helm-installer` | FAILURE |
| `main-elasticsearch-5x-k8s-local-test` #2078 | `ma-helm-installer` | FAILURE |

Only #307 had an actually-interesting root cause — Kyverno admission webhook returning `remote error: tls: internal error` before its cert was ready. The other four were ordinary transient flakes that `backoffLimit: 1`/`0` turned into pipeline failures.

### Fix

Two layers of recovery as per discussions : 

**Job-level — `podFailurePolicy` on all four Jobs:**
```
backoffLimit: 1
activeDeadlineSeconds: <600|900|1200>
podFailurePolicy:
    rules:
        - action: Ignore         # eviction/node-loss don't burn retries
           onPodConditions:
               - type: DisruptionTarget
        - action: FailJob        # deterministic non-zero exits fail fast
           onExitCodes:
               containerName: <per-job>
               operator: NotIn
               values: [0]
```

activeDeadlineSeconds bounds total wall-clock so a wedged Job still fails in reasonable time.

**Per-step : new `_retryHelper.tpl` defining a POSIX-safe `retry` function**, wrapping the network/dep-prone calls inside the install/uninstall scripts :

| Wrapped step | Where | Attempts × backoff |
|---|---|---|
| ECR registry login (token + IAM paths) | `installJob.yaml` | 2 × 5s |
| `helm repo add` per chart | `installJob.yaml` | 3 × 5s |
| `helm repo update` | `installJob.yaml` | 3 × 10s |
| `kubectl apply -f /policy/policy.yaml` (Kyverno) | `installJob.yaml` | 3 × 5s |
| `helm upgrade --install` (parallel + Kyverno serial) | `installJob.yaml` + `_installHelper.tpl` | 2 × 15s |
| `kubectl apply` CloudWatch dashboards | `installJob.yaml` | 3 × 5s |
| `kubectl apply` installation-notes ConfigMap | `installJob.yaml` | 2 × 5s |
| `helm uninstall` (parallel) | `_uninstallHelper.tpl` | 2 × 10s |
| `aws s3 rm --recursive` | `deleteS3Bucket.yaml` | 2 × 5s |
| `aws s3 rb` | `deleteS3Bucket.yaml` | 2 × 5s (non-fatal, preserves diagnostic) |

`kubectl wait` calls are intentionally **not** wrapped as they have explicit `--timeout` and represent state-machine waits, not API flake.

### Versioning

`kubeVersion: ">=1.26.0-0"` added to `Chart.yaml` so installs against pre-1.26 fail at `helm install` rather than mid-run.

### Issues Resolved
N/A

### Testing

Manually validated on a local minikube cluster (Docker driver) plus standalone shell + helm-template runs.

- **`helm template`** — rendered the chart against default values, kyverno-enabled values, the ECR-token auth path, and the ECR-IAM path; all 7 sub-chart names produce POSIX-valid `do_install_<name>` shell function identifiers (verified by grepping the rendered output for hyphenated names — none found).
- **`kubeVersion` gate (template-time)** — `helm template --kube-version 1.25.0` rejected with `chart requires kubeVersion: >=1.26.0-0`; `1.26.0`/`1.31.0`/`1.35.0` rendered OK.
- **`retry` helper unit tests** — extracted the rendered `retry` function into a standalone shell script and ran 6 cases under both `/bin/sh` and `bash` (success-on-first, retry-then-succeed, exhausted-failure, multi-arg cmd, optional `--` separator, `set -e` + handler interop) — 6/6 pass on each shell.
- **Live install Job on minikube k8s 1.31** — installed the chart with a stub values file (all sub-charts disabled), forced `helm repo update` to fail. Observed pod logs showed it exhausted 3 attempts followed by the non-fatal `|| echo "..."` handler taking over, exactly as intended:

  ```
  Starting Helm installation sequence...
  Error: no repositories found. You must add one before updating
  Transient: helm repo update failed (attempt 1/3), sleeping 10s...
  Error: no repositories found. You must add one before updating
  Transient: helm repo update failed (attempt 2/3), sleeping 10s...
  Error: no repositories found. You must add one before updating
  FAIL: helm repo update — exhausted 3 attempts
  No traditional helm repos to update (OCI repos don't need it)
  ```
- **`podFailurePolicy: FailJob`** : applied a synthetic Job copying our exact policy block with a worker that `exit 17`; Job condition came back as `reason: PodFailurePolicy, "exit code 17 matching FailJob rule at index 1"`, with **exactly 1 pod** (proving `backoffLimit: 1` was not consumed because the policy intercepted before backoff).
- **`podFailurePolicy: Ignore on DisruptionTarget`** : Applied a synthetic Job, then evicted its pod via the Kubernetes eviction API (`POST /api/v1/namespaces/.../pods/.../eviction`, which sets `DisruptionTarget: EvictionByEvictionAPI`). Observed that a second pod was spawned automatically and ran to completion, Job ended `succeeded=1` despite `backoffLimit: 1`.
- **`kubeVersion` gate (install-time)** : started a real k8s 1.25.16 minikube (`--force` since 1.25 is below minikube's normal floor); `helm install` was rejected with `chart requires kubeVersion: >=1.26.0-0 which is incompatible with Kubernetes v1.25.16`.

### Check List


gregschohn:
I'm concerned that the retries here are too coarse.  I'd prefer to see something more like this as the job policy
```
podFailurePolicy:
    rules:
      # 1. If the script exits with ANY non-zero code, fail the Job immediately.
      - action: FailJob
        onExitCodes:
          containerName: worker
          operator: NotIn
          values: [0]
      
      # 2. If the Pod is evicted or the Node disappears (transient), ignore it.
      # This means it will retry, but it won't count against your backoffLimit.
      - action: Ignore
        onPodConditions:
          - type: DisruptionTarget
```

...and then harden the 10 different steps that you're talking about in the shell script to retry.  We'll still have to handle cases where the install pod dies unexpectedly, but that might be 100x less common than a transient network issue for one of these dependencies.  If we put separate retries in around each step, the difference in behavior will have only minor differences. As we make the whole thing reentrant, that could require significantly more work in the shell script logic itself.  We'll eventually want to get to it - but it doesn't seem to be high priority and will take time to get right.

In the meantime, granular retries will allow users to recover faster and with a higher likelihood of success.  We still can put the transient retries in place, so we'll gather some data and see what else we might need to do in those rare cases.

jugal-chauhan:
> I'm concerned that the retries here are too coarse. I'd prefer to see something more like this as the job policy
> 
> ```
> podFailurePolicy:
>     rules:
>       # 1. If the script exits with ANY non-zero code, fail the Job immediately.
>       - action: FailJob
>         onExitCodes:
>           containerName: worker
>           operator: NotIn
>           values: [0]
>       
>       # 2. If the Pod is evicted or the Node disappears (transient), ignore it.
>       # This means it will retry, but it won't count against your backoffLimit.
>       - action: Ignore
>         onPodConditions:
>           - type: DisruptionTarget
> ```
> 
> ...and then harden the 10 different steps that you're talking about in the shell script to retry. We'll still have to handle cases where the install pod dies unexpectedly, but that might be 100x less common than a transient network issue for one of these dependencies. If we put separate retries in around each step, the difference in behavior will have only minor differences. As we make the whole thing reentrant, that could require significantly more work in the shell script logic itself. We'll eventually want to get to it - but it doesn't seem to be high priority and will take time to get right.
> 
> In the meantime, granular retries will allow users to recover faster and with a higher likelihood of success. We still can put the transient retries in place, so we'll gather some data and see what else we might need to do in those rare cases.

Thanks for pointing this out, it is a great suggestion! I have now reworked the PR since last review. Recent [changes](https://github.com/opensearch-project/opensearch-migrations/pull/2914/changes/f58c4fa911c78987252a53e9c7f8404a6fc7f52b..f489a9caa40cdf2f269dd2a1c197a40c51771073) are :

- `podFailurePolicy` on all four Jobs (`Ignore` on `DisruptionTarget`, `FailJob` on non-zero exit). `backoffLimit: 1` since the policy is now the primary defense.
- New `_retryHelper.tpl` providing a POSIX `retry <max> <sleep_s> <desc> -- <cmd…>` helper. 
- Wrapped the ~10 transient-prone steps you flagged `(helm repo add/update, ECR login, helm upgrade --install, helm uninstall, kubectl apply, aws s3 rm/rb)`. 
- `kubectl wait` left alone — it has its own timeouts and is a state-machine wait.
- Reentrancy of the install script has been deferred as of now — bounded retries cover the common cases.
- Added `kubeVersion: ">=1.26.0-0"` to `Chart.yaml` so a pre-1.26 install fails fast at `helm install` rather than mid-run.

Locally verified the policy and helper end-to-end on minikube (k8s 1.31): retry recovers from a forced flake, `FailJob` intercepts a synthetic non-zero exit with exactly 1 pod (no wasted retries), and an evicted pod gets a free retry via the `Ignore` rule with the Job still ending `succeeded=1`.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/childrenChartInstaller/installJob.yaml
```diff
@@ -13,7 +13,18 @@ metadata:
     "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
 spec:
   backoffLimit: 1
+  activeDeadlineSeconds: 1200
```
gregschohn, line 16: should we we make this a templated value (defaulting to 20 or 30 minutes) so that users could override if their network/hardware are significantly different/

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/helpers/_retryHelper.tpl
```diff
@@ -0,0 +1,35 @@
...
+  while true; do
+    if "$@"; then
+      return 0
+    fi
+    if [ "$_retry_i" -ge "$_retry_max" ]; then
+      echo "FAIL: $_retry_desc — exhausted $_retry_max attempts" >&2
+      return 1
+    fi
+    echo "Transient: $_retry_desc failed (attempt $_retry_i/$_retry_max), sleeping ${_retry_sleep}s..." >&2
+    sleep "$_retry_sleep"
```
gregschohn, line 31: should we bump these up to make them exponential?  You don't typically have many retries - and I think that's the right policy, so it doesn't make a big difference.

deployment/k8s/charts/aggregates/migrationAssistantWithArgo/Chart.yaml
```diff
@@ -3,3 +3,5 @@ name: migration-assistant
 version: 0.2.0
 type: application
 appVersion: "latest"
+# podFailurePolicy on Jobs requires Kubernetes >= 1.26 (beta) / GA in 1.31.
+kubeVersion: ">=1.35.0-0"
```
gregschohn, line 7: I've bumped this to 1.35

gregschohn (approved):

PR #2902, opened 2026-05-06
fix: deduplicate same-position tokens and preserve offset gaps in sourceless reconstruction

AndreKurait:
## Problem

Two bugs in the sourceless ES index reconstruction pipeline:

### Bug 1 — Duplicate tokens

When a Lucene token filter (e.g. `word_delimiter_graph`) produces a compound
token alongside its sub-tokens at the **same Lucene position**, all tokens were
stored individually in the sidecar and all returned by `SidecarReader.get()`,
yielding duplicates like:

```
date tue 26 sep 09 2000 2000 09 27 00 00 -0700 0700 pdt
```

instead of:

```
date  tue  26 sep 2000 09 27 00 -0700  pdt
```

Root cause: `word_delimiter_graph` emits the compound form (`"2000 09"`) **and** its
sub-tokens (`"2000"`, `"09"`) all at the same position increment=0.  All three records
land in the sidecar; all three came back from `get()`.

### Bug 2 — Spacing collapse

`String.join(" ", terms)` collapsed all inter-token gaps to a single space,
destroying double-spaces in values like `"Date: Tue, 26 Sep 2000..."`.

## Fix

**Dedup at read time (SidecarReader)**: group sidecar entries by Lucene position and
keep only the longest token per group. The longest token is always the unsplit compound
form; shorter co-positioned tokens are analysis artifacts carrying no extra characters.

**Offset-gap reconstruction**: sidecar now stores per-token character `start/endOffset`
from `PostingsEnum.OFFSETS`. `LuceneLeafReader.joinWithOffsets()` fills each inter-token
gap with exactly `(next.startOffset - prev.endOffset)` spaces. Falls back to
single-space join when offsets are unavailable (`NO_OFFSET = -1` sentinel).

## Changes

| File | Change |
|---|---|
| `PostingsSink` | `NO_OFFSET=-1`; new 6-arg `accept()` with `startOffsets[]`/`endOffsets[]`; old 4-arg kept as default shim |
| `SidecarBuilder` | Record 12→20 bytes: appends `startOff(4)`+`endOff(4)`. `DocPayloadEmitter` writes ZInt-encoded offsets |
| `SidecarReader` | New `TermEntry(term, startOffset, endOffset)` record; `get()` deduplicates by position keeping longest; `getTermStrings()` for compat callers |
| `SegmentTermIndex` | `getTermEntriesForDocument()` returning `List<TermEntry>`; `getTermsForDocument()` delegates to it |
[...]

PR #2915, opened 2026-05-08
feat(RFS): sourceless _source reconstruction from Lucene primitives

AndreKurait:
## Summary

This PR lands the full RFS sourceless reconstruction stack — reconstruct `_source` from Lucene postings / doc_values / points when the source index has `_source` disabled or partially excluded — onto `main`. It supersedes #2902 (which targeted the `es-sourceless-experimental` integration branch) by collapsing the entire incubating stack into a single PR against `main`.

## What this enables

Snapshot migrations where the source index set `_source.enabled: false` (or used `_source.excludes`) previously had no way to migrate docs — there was literally no `_source` blob in the segment to replay. This PR adds a reconstructor that rebuilds `_source` from the stored primitives Lucene keeps per-field:

- **STRING / TEXT** fields: reconstruct from postings with position + offset data (dedup co-positioned tokens; preserve inter-token offset gaps as real whitespace)
- **KEYWORD / numeric / binary / date** fields: reconstruct from doc_values / points
- **Object-array subfields**: distribute recovered subfields back into the correct array element via a scalar-broadcast seed

The sidecar (term-position index) is streamed to disk via external merge sort so reconstruction scales beyond the 2 GiB `MappedByteBuffer` limit.

## How it works

1. **SidecarBuilder** walks each segment's postings once, emits `(doc, field, position, startOffset, endOffset, term)` records to a spillable sidecar file.
2. **External merge sort** (new `SidecarBuilder.sortBufferBytes: long`) groups records by `(doc, field)` without holding the full index in RAM.
3. **SidecarReader** mmap-or-stream reads the sorted file, dedups co-positioned tokens (keeps longest — the compound form; shorter variants are analysis artifacts), and reconstructs the field value.
4. **LuceneLeafReader.joinWithOffsets** preserves exact inter-token spacing from offset deltas; falls back to single-space when offsets unavailable (`NO_OFFSET = -1` sentinel).
5. **Fallback chain**: if postings-based reconstruction fails → try `_recovery_source` → try doc_values → skip doc with a warning.

## Commits (17, oldest → newest)

```
0499b8f5a  stream term-position index to disk via external merge sort
4910cdf33  unbind SidecarReader from the 2 GiB mmap limit
10961e93f  widen sidecar term-offset to int64
f5eda3345  distribute recovered subfields into object-array seeds
097cc0e55  distribute recovered subfields into object-array seeds
7a35d0a7f  Merge branch 'main' into alt-offline-sort
02c2436e8  Delegate to Lucene 10 native primitives; drop hand-rolled utilities
0921d18ad  Further reduce sidecar LOC; trim docs
c73504ff7  Widen SidecarBuilder sortBufferBytes arg to long
8e8fcc677  Clear Sonar S1659 / S1905 in sidecar
d63efcaa3  E2E coverage for position-buffer grow and scalar-broadcast guards
032b361b2  Integrate #2893: distribute recovered subfields into object-array seeds
f5d379a48  Integrate #2886: stream term-position index to disk
341f6ef99  dedup same-position tokens; preserve offset gaps
c649c5ccb  guard PostingsEnum.OFFSETS to only request when field has offsets indexed
2cd172f7c  inline forField to resolve Sonar S1854 dead-assignment
72367cafa  always use _recovery_source as _source fallback when absent
```

## Scope

- Ships reconstruction machinery + unit + E2E tests (`NoStoredSourceMigrationTest` covers 9 source/target version pairs).
- No runtime toggle: source reconstruction runs automatically when a segment lacks `_source` and is needed by the document stream.

## Testing

Unit + E2E reconstruction tests run under `:DocumentsFromSnapshotMigration:isolatedTest` across the version matrix (ES 1.7.6, 2.4.6, 5.6.16, 6.8.23, 7.10.2, 7.17.29, 8.17.10 → OS 3.5.0).

```
./gradlew :SearchSnapshotExtractor:test --tests "*.sidecar.*" --tests "*SourceReconstructor*"
./gradlew :DocumentsFromSnapshotMigration:isolatedTest --tests "*NoStoredSourceMigrationTest*"
```

## Supersedes

Closes #2902 (re-homed from `es-sourceless-experimental` base → `main`).


SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/sidecar/SidecarReader.java
```diff
@@ -0,0 +1,175 @@
...
+public final class SidecarReader implements AutoCloseable {
+
+    /**
+     * A single token with its character-offset span in the original field value.
+     *
+     * <p>{@code startOffset} and {@code endOffset} are {@link PostingsSink#NO_OFFSET} (-1) when
+     * the field was not indexed with {@code index_options: offsets}. Callers must check for -1
+     * before using offsets for gap-preserving reconstruction.
+     */
+    public record TermEntry(String term, int startOffset, int endOffset) {}
```
jugal-chauhan: nit: Moving `TermEntry` from a top-level class to `SidecarReader.TermEntry` means all callers now reference `SidecarReader.TermEntry`. Was this intentional ?  This couples the public API to the reader implementation. If `TermEntry` is part of the contract between `SegmentTermIndex` and `LuceneLeafReader`, it might be cleaner as a standalone record in the sidecar package

PR #2920, opened 2026-05-08
Restore default SIGPIPE handling in CLI entry points

### LINKED ISSUE #2803: [BUG] [workflow CLI] "workflow configure sample" emits BrokenPipeError traceback when downstream reader closes pipe early
## Summary
`workflow configure sample` emits a Python traceback instead of exiting silently when piped to a command that closes stdin before reading all output. Standard Unix CLI tools handle SIGPIPE silently.

## Environment
- MA version: 3.0.1
- Deployment: EKS (import-vpc)

## Reproduction
$ workflow configure sample | head -0 ERROR:console_link.workflow.commands.configure:Failed to get sample configuration: [Errno 32] Broken pipe Traceback (most recent call last): File "/root/lib/console_link/console_link/workflow/commands/configure.py", line 315, in sample_config click.echo(sample_content) File "/.venv/lib64/python3.11/site-packages/click/utils.py", line 321, in echo file.write(out) BrokenPipeError: [Errno 32] Broken pipe Error: Failed to get sample: [Errno 32] Broken pipe

Also reproduces with `| true`.
Impact
Minor user-facing annoyance — the output is still generated correctly when the consumer reads everything.

jugal-chauhan:
### Problem
Running `workflow configure sample | head` (or any early-closing pipe consumer) prints a full Python traceback and `BrokenPipeError` to stderr. This happens because Python's default SIGPIPE handler raises an exception instead of exiting silently.

### Fix
Restore default Unix SIGPIPE handling (`signal.signal(signal.SIGPIPE, signal.SIG_DFL)`) at both CLI entry points (`console` and `workflow`). This is the standard approach used by `aws`, `kubectl`, and other CLI tools.

### Issues Resolved
Fixes https://github.com/opensearch-project/opensearch-migrations/issues/2803

### Testing
Added regression tests covering:
  - Import-time verification that SIG_DFL is installed
  - End-to-end pipe scenarios (`head -0`, `head`, `true`, `grep -m1`, `sed -n 1p`)
  - Happy-path sanity check (full read still works)

### Check List

### DIFF
--- a/migrationConsole/lib/console_link/console_link/cli.py
+++ b/migrationConsole/lib/console_link/console_link/cli.py
@@ -1,4 +1,5 @@
 # pyright: ignore[reportCallIssue]
+import signal
 from contextlib import contextmanager
 import json
 from pprint import pprint
@@ -6,6 +7,12 @@ import sys
 import time
 from typing import Dict
 import click
+
+# Restore default SIGPIPE handling so piped commands (e.g. `| head`) exit cleanly.
+try:
+    signal.signal(signal.SIGPIPE, signal.SIG_DFL)
+except (AttributeError, ValueError):
+    pass
 
 import console_link.middleware.clusters as clusters_
 import console_link.middleware.metrics as metrics_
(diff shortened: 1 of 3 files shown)


AndreKurait (approved):
Thanks! This is a great change for agents!

PR #2922, opened 2026-05-09
Solr: preserve dotted field names through migration

AndreKurait:
## Description

Solr permits literal `.` in field names (e.g. `category.name`, `attr_thing.withdot`). This PR makes the Solr→OpenSearch migration handle them correctly.

**Bug fix.** When a dotted-name doc field's top-level segment also matches a Solr `dynamicField` pattern (e.g. `attr_thing.withdot` under `attr_*`), OpenSearch auto-expands the dotted key into a parent object on write. The previous `buildDynamicTemplate` output used a plain `match` clause, which then fired on the synthesized parent — typing it as the template's target type and causing the child write to fail with `mapper_parsing_exception` (`NumberFieldMapper cannot be cast to ObjectMapper`). The bulk request fails and the doc is silently lost.

Fix: emit each `dynamic_template` with `path_match` (matches the field's full dotted path) plus `match_mapping_type` (gates on the JSON value shape, which intermediate object containers never carry). The combination scopes the template to leaves only.

**Test coverage.** Adds:
- Unit tests pinning the new template shape and the per-OS-type → `match_mapping_type` mapping.
- SolrReader E2E `migratesFieldsWithDotsInName` — declared dotted fields of every common Solr type round-trip through `_field_caps`, `_source`, and dotted-path queries.
- SolrReader E2E `migratesDynamicFieldsWithDotsInLeafName` — six dynamic-field patterns (`attr_*`, `*_s`, `*_i`, `*_b`, `*_f`, `*_dt`) each with a dotted leaf, exercising the regression path end-to-end.
- Backfill E2E `SolrSnapshotToOpenSearchTest.backfillsDottedFieldNames` — runs the full production flow (`MetadataMigration` pushes the converted schema, then `RfsMigrateDocuments` CLI writes documents) and verifies dotted-name mappings, `_source` preservation, and dotted-path queries on the target. Pre-fix, the dynamic-leaf doc would have been silently dropped at bulk-write.
- Python integ scenario in `solr_tests.py` covering the same path on the live console workflow.

The two SolrReader E2E tests are also restructured around explicit SOURCE → MIGRATE → TARGET phases with a tabular Javadoc preamble — each test reads top-to-bottom and clearly states what's on Solr vs. what's expected on OpenSearch (mapping types + `_source` + queries).

### Is this a breaking change?

No. `buildDynamicTemplate` is package-private; only its emitted template shape changed. Existing dynamicField behavior (no dotted leaves) is unchanged — `path_match` matches non-dotted leaves with the same wildcard semantics, and `match_mapping_type` accepts the JSON shape OpenSearch reports natively for each target type. All prior E2E fixtures (`dynamicFieldsAndCopyFieldsMigration`, `copyFieldDestinationsResolveTypeFromDynamicFields`, `dateFieldsGetProperFormatMapping`) still pass.

### Issues Resolved

n/a — internal investigation.

### Check List

- [x] Commits are signed per the DCO

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/opensearch-migrations/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

### DIFF
--- a/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SolrSnapshotToOpenSearchTest.java
+++ b/DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SolrSnapshotToOpenSearchTest.java
@@ -12,6 +12,9 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Stream;
 
+import org.opensearch.migrations.MetadataMigration;
+import org.opensearch.migrations.MigrateOrEvaluateArgs;
+import org.opensearch.migrations.Version;
 import org.opensearch.migrations.bulkload.common.DocumentExceptionAllowlist;
 import org.opensearch.migrations.bulkload.common.OpenSearchClientFactory;
 import org.opensearch.migrations.bulkload.common.RestClient;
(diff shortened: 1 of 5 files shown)


jugal-chauhan (approved):
Thanks for brining in these changes with a thorough test coverage!

PR #2924, opened 2026-05-11
RFS: release lease on shutdown when no docs were migrated

AndreKurait:
## Summary
- When a worker exits before checkpointing any documents (slow shard download, k8s pod kill, etc.), the work item used to stay leased until natural expiration, blocking other workers. Both shutdown paths — SIGTERM clean shutdown and the early lease-timeout trigger — hit the same dead end with `progressCursor == null`: no checkpoint to write and no useful successor to seed.
- Add `IWorkCoordinator.releaseWorkItem` that clears `expiration` and `leaseHolderId` via a Painless script, gated on the caller still owning the lease (`leaseHolderId == workerId`) and the item not being completed. Late or racing releases are no-ops, so they can't clobber a new owner or resurrect a completed item.
- `nextAcquisitionLeaseExponent` is intentionally left untouched — the work hasn't been done yet, and any prior backoff bump is still meaningful for the next acquirer.
- Wire the new helper into both `executeCleanShutdownProcess` and `exitOnLeaseTimeout` via a shared `releaseLeaseWithoutProgress` branch, so both shutdown paths follow the same code.

## Why
Doc-zero shutdowns (especially on short-lived migration pods) currently waste an entire lease window before the next worker can retry. With short test leases this can stall a migration end-to-end; with long production leases it can stall it for many minutes per failure.

## Test plan
- [x] New container integration tests in `WorkCoordinatorTest`:
  - `testReleaseWorkItemAllowsImmediateReacquisition` — workerA acquires + releases, workerB picks up the same item with no wait.
  - `testReleaseWorkItemIsNoopIfLeaseRolledOver` — late release from workerA must not steal workerB's active lease.
  - `testReleaseWorkItemIsNoopIfAlreadyCompleted` — release after completion must not resurrect the item.
- [x] Existing coordinator tests unchanged.

### DIFF
--- a/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
+++ b/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
@@ -551,7 +551,8 @@ public class RfsMigrateDocuments {
                         arguments.initialLeaseDuration,
                         () -> Optional.ofNullable(cancellationRunnableRef.get()).ifPresent(Runnable::run),
                         cleanShutdownCompleted,
-                        context.getWorkCoordinationContext()::createSuccessorWorkItemsContext),
+                        context.getWorkCoordinationContext()::createSuccessorWorkItemsContext,
+                        context.getWorkCoordinationContext()::createReleaseWorkItemContext),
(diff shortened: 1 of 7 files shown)


jugal-chauhan (approved):
LGTM! Thanks for brining in these changes

PR #2925, opened 2026-05-11
RFS: log bulk response body at INFO on any failure

AndreKurait:
## Summary
- Promote per-response error log in `OpenSearchClient.executeBulkWithRetry` from DEBUG to INFO so document migration failures surface immediately on the first attempt, not only after retries are exhausted at WARN/ERROR.
- Truncate the body to `BULK_TRUNCATED_RESPONSE_MAX_LENGTH` to match the existing WARN log.

## Test plan
- [ ] CI green
- [ ] Manual: trigger a bulk index failure and confirm the response body appears at INFO on the first attempt (not just on the final retry)


jugal-chauhan (approved):
LGTM

PR #2928, opened 2026-05-11
Solr→OS shim: support date-math, ISO timestamps, and field:(a OR b) groups in queries

AndreKurait:
## What

Three real Solr→OpenSearch query translation gaps in `TrafficCapture/SolrTransformations/transforms/`, each surfaced by running the SolrCloud sandbox migration and inspecting the queries that the shim either failed to parse or translated into something OpenSearch then rejected.

| | Before | After |
| --- | --- | --- |
| `review_date:[NOW-365DAYS TO NOW]` | parser error: unexpected `-` after `[` | translates to `range: { review_date: { gte: "now-365d", lte: "now" } }` |
| `review_date:[2020-01-01T00:00:00Z TO 2022-12-31T23:59:59Z]` | parser error: unexpected `:` | translates to `range: { review_date: { gte: "2020-01-01T00:00:00Z", lte: "2022-12-31T23:59:59Z" } }` (passthrough — OS accepts ISO natively) |
| `marketplace:(US OR UK OR JP)` | parser error: unexpected `(` | translates to `bool: { should: [ term:{marketplace:US}, term:{marketplace:UK}, term:{marketplace:JP} ] }` (group hoisting) |

These are the top three sources of refused queries in the sandbox sample I ran — each one breaks a common idiomatic Solr query shape that workloads use heavily.

## Changes

Three small, orthogonal changes:

**1. `solr.pegjs` `rangeVal` — extend allowed character class.** Was `[a-zA-Z0-9._\-]+`, which excluded `:`, `T`, `Z`, `/`, and `+` — every character ISO timestamps and Solr date-math need. Whitespace deliberately stays excluded; `_` rules around the bound handle whitespace and `TO` is unambiguous.

**2. `dateMath.ts` (new) + `rangeRule.ts` — translate Solr date-math into OpenSearch date-math.** OpenSearch wants lowercase `now` and single-letter units (`d`, `h`, `m`, `s`, `M`, `y`). The helper handles bare `NOW`, offsets (`NOW-365DAYS`), rounding (`NOW/MONTH`), and compound expressions (`NOW/MONTH-6MONTHS+1DAY/HOUR`). ISO-8601 timestamps and other unrecognized values pass through unchanged so OpenSearch can reject them loudly rather than the shim swallowing the discrepancy. The helper is colocated with the other transformer rules so future callers (legacy `facet.range`, `stats.field` percentiles over time) can reuse it.

**3. `solr.pegjs` `fieldExpr` — accept `field:(subquery)` group sugar.** Solr canonicalises this as `(field:a OR field:b OR field:c)`. A new `fieldExpr` alternative captures the inner subquery, walks it, and hoists the outer field onto every bare leaf. Nested explicit-field nodes inside the group keep their own field — `tag:(foo OR other:bar)` keeps `other:bar` qualified by `other`, matching Solr semantics. Boost on the outer group is preserved (`marketplace:(US OR UK)^2.0`).

## Tests

Three new vitest specs:

- `dateMath.test.ts` — 27 tests covering the date-math translator (bare `NOW`, offsets, rounding, compound expressions, ISO passthrough, malformed-input passthrough).
- `grammarExtensions.test.ts` — 15 tests for the grammar additions, including regression coverage on the existing range and group forms.
- `rangeRule.test.ts` +4 tests — date-math integration through `rangeRule`.

Suite count goes from **935 → 977 passing tests**. No prior test was modified or relaxed.

```
$ npm run test:nocov
 Test Files  40 passed (40)
      Tests  977 passed (977)
```

## What this does NOT include

This PR is intentionally narrow — only the three gaps above. There are four more real gaps in the same area (legacy `facet=true & facet.field/.range/.query/.pivot`, `stats.field`, `group.field`, `track_total_hits`) that I have working in a follow-up branch and will open separately so each PR carries its own evidence.

## Compatibility

- No Solr query that parsed before this PR will parse to a different AST after it. Verified by 977/977 passing including the existing `parser.test.ts` (82 tests) and `rangeRule.test.ts` (8 prior tests).
- No new dependencies. `solr.pegjs` is loaded at runtime and parsed by `peggy.generate()` on first call, so the grammar changes pick up automatically.
- Shim-internal only; no impact on Console, Replayer, Reindex-from-Snapshot, or migration-orchestration paths.

### DIFF
--- a/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/grammarExtensions.test.ts
+++ b/TrafficCapture/SolrTransformations/transforms/src/solr-to-opensearch/query-engine/parser/grammarExtensions.test.ts
@@ -1,0 +1,242 @@
+/**
+ * Unit tests for grammar fixes:
+ *
+ *   1. `rangeVal` accepts ISO-8601 timestamps and Solr date-math
+ *      (`NOW`, `NOW-365DAYS`, `NOW/MONTH-6MONTHS`, `2020-01-01T00:00:00Z`).
+ *   2. `fieldExpr` accepts the `field:(a OR b OR c)` parenthesized
+ *      value-group sugar and hoists the field onto every bare leaf
+ *      inside the group.
+ *
+ * These are pure parser-level tests: we only assert on the AST shape, not
+ * on downstream OpenSearch DSL output. Range/fieldExpr translation tests
+ * live in `rangeRule.test.ts` and `astToOpenSearch.test.ts`.
+ */
+
+import { describe, it, expect } from 'vitest';
+import { parseSolrQuery } from './parser';
+
+const emptyParams = new Map<string, string>();
+
+describe('grammar — rangeVal extensions (Fix 1)', () => {
+  it('parses an ISO-8601 timestamp range', () => {
+    const { ast, errors } = parseSolrQuery(
+      'review_date:[2020-01-01T00:00:00Z TO 2022-12-31T23:59:59Z]',
+      emptyParams,
+    );
+    expect(errors).toEqual([]);
+    expect(ast).toMatchObject({
+      type: 'range',
+      field: 'review_date',
+      lower: '2020-01-01T00:00:00Z',
+      upper: '2022-12-31T23:59:59Z',
+      lowerInclusive: true,
+      upperInclusive: true,
+    });
+  });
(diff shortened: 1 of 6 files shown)


jugal-chauhan (approved):
LGTM! Thanks for addressing these three production query translation failures in a focused fix.

PR #2937, opened 2026-05-11
Sourceless reconstruction + simplification pass on top of #2907

AndreKurait:
## Summary

Builds on #2907 (sourceless copy_to fields) and applies a structural simplification pass identified by a code-reuse / quality / efficiency review of the same diff. Functional behaviour is unchanged from #2907; this PR is intended to land alongside (or fold into) it.

### Hot-path performance (per-document)
- **Cache `getDocValueFields()` per LeafReader** (5/6/7/9/10) — was rebuilt on every doc; now memoized via `volatile List<DocValueFieldInfo>`. Replaced the term-list materialization in boolean detection with a short-circuiting `fieldHasOnlyBooleanTerms` walking the `TermsEnum` directly.
- **`RecoveryContext` with O(1) lookups** — `populateFromSegment` and `probeFieldValue` now share a per-doc context that lazy-builds `Map<String, LuceneField>` from `document.getFields()` and `Map<String, DocValueFieldInfo>` from the cached reader fields. Eliminates the per-`copy_to`-target linear scans.
- **Cached dotted-path splits** — `FieldMappingContext.getPathSegments()` returns a pre-split `String[]` for mapped fields (cached at construction); arbitrary names use `splitPath` (manual single-pass, no `Pattern`). Replaces all `String.split("\\.")` calls.
- **Pre-compiled source-filter globs** — `_source.includes/excludes` stored as `List<Predicate<String>>`; `compileGlob` runs once at construction.
- **Pre-sorted copy_to targets / pre-filtered constants** — sort by lossiness once at construction; constant_value entries pre-extracted so pass 4 walks O(constant fields) instead of O(all fields).
- **`couldContributeToMerge()` fast-path** — `mergeWithDocValues` returns the original bytes (no JSON round-trip) when no recovery tier could contribute.

### Code quality
- Tighten `LuceneLeafReader.getDocValueFields()` return type from `Iterable` to `List`.
- Unified `walkMapAncestors` primitive shared between `descendsIntoExistingObjectArray` and (effectively) `hasNested`.
- Promoted ES field-type names to `EsFieldType.MappingTypes` constants (`constant_keyword`, `text`, `half_float`, `float`, `double`, `wildcard`, `integer`, `short`, `byte`, `match_only_text`, etc.).
- Trimmed verbose Javadocs that explained WHAT the code does; kept WHY-only callouts (hidden constraints, ordering caveats, version-specific quirks).
- Reverted an unrelated logging regression in `OpenSearchClient.bulkRequest` (INFO → DEBUG, dropped `truncateMessageIfNeeded`) that snuck in via #2907.
- Added missing `@Override` on new `toString()` methods (LeafReader9, LeafReader10).
- Removed a dead `JsonNode tree = ...` assignment from `SourceReconstructorTest`.

### Out of scope
The BKD point-values per-doc full scan (`compare()` returning `CELL_CROSSES_QUERY`) flagged by the efficiency review is left for a follow-up: a `SegmentTermIndex`-style per-segment cache would help but touches all five LeafReaders' point-traversal logic and warrants its own focused PR with benchmarks.

## Test plan
- [x] `:SearchSnapshotExtractor:test` (SourceReconstructorTest, FieldMappingContextTest) — all passing
- [x] `:RFS:test` (OpenSearchClientTest including `testBulkRequest_warningLogContainsDetails_onError` and `testBulkRequest_truncatesLargeLog_onError`) — all passing
- [x] `:DocumentsFromSnapshotMigration:compileTestJava` (NoStoredSourceMigrationTest is `@Tag("isolatedTest")` so isn't run by default; compiles cleanly)
- [x] `./gradlew compileJava` — every module builds
- [x] `./gradlew spotlessApply` — no formatting changes needed
- [ ] Run isolated/integration tests (`isolatedTest` tag) in CI


jugal-chauhan (approved):
Thanks for bringing this in! The optimizations look great : caching `getDocValueFields()`, pre-splitting paths, and the`RecoveryContext O(1)` maps all target real per-document overhead that compounds across several docs.


A couple of questions:

- The `volatile List<DocValueFieldInfo>` cache on `LeafReader` : Is there a concern about the list being mutated after caching ? Since `getDocValueFields()` builds a new list each time on the current code, the cached version should be immutable. Could you confirm the cached list is never modified after construction?
- [couldContributeToMerge()](https://github.com/opensearch-project/opensearch-migrations/pull/2937/changes#diff-5bc34106f5727d2c8f92bb191b0a6e6789a3bcc4c53c3abfa3485a09063bef68R267-R275) : what's the heuristic here? If it returns `false` but a field IS recoverable (e.g., a new field type added later that the heuristic doesn't know about), the merge would silently skip recovery. Is there a safe fallback (e.g., return true for unknown field types) ?

AndreKurait:
Superseded by https://github.com/opensearch-project/opensearch-migrations/pull/2985

PR #2938, opened 2026-05-11
Fix replayer Kafka drain stalls (CDC LAG never advancing); add per-partition time-lag to console kafka describe-consumer-group; SCRAM in k8s

AndreKurait:
## Summary

Fixes a class of CDC drain stalls that left the replayer's Kafka consumer-group offset frozen indefinitely while the migration itself succeeded — and adds the operator-facing observability that surfaces them.

## Drain-stall fixes (root causes that motivated this PR)

OffsetLifecycleTracker holds Kafka offsets in a priority queue and only commits when the head of the queue clears. If any single record gets pinned at the head, every subsequent commit returns `BLOCKED_BY_OTHER_COMMITS` and `CURRENT-OFFSET` stops advancing. The replayer can be writing to the target perfectly while LAG climbs forever. Five distinct ways this happened, each fixed at its source:

1. **Synchronous tuple-writer failure leaked the head record** — `handleCompletedTransaction` deferred the Kafka commit to `writeFuture.whenComplete`. When `packageAndWriteTuple` threw synchronously (e.g. AWS SDK `BlockingOutputStreamAsyncRequestBody` 10s timeout), the future was never built and the commit was never scheduled.
2. **Tuple-finalization re-throw cancelled the recovery commit** — the old `catch (Exception)` block logged "Proceeding" then immediately `throw e`, defeating its own intent. Split the inner/outer try so finalization failures fall through with one canonical commit at the bottom of the success path.
3. **Connection-close commit was deferred to a future that never fired** — `onConnectionClose` chained the commit off `replayEngine.closeConnection().whenComplete(...)`. When the channel close stalled, the offset stayed pinned 11+ minutes. Commit eagerly on the normal path (matching the other three callers) — channel close is independent network cleanup, not a migration-semantics gate.
4. **Held streams were dropped when the accumulator closed in WAITING / IGNORING states** — `fireAccumulationsCallbacksAndClose` only committed in the `ACCUMULATING_*` branches. Restructured the `finally` block so any held keys still reach `onTrafficStreamsExpired` regardless of state.
5. **Partition revoke→reassign round-trip left a stale-generation accumulation** (#3013, merged in) — when the consumer was fenced and the same partition came back, the broker reset the fetch position and re-delivered records. The old "trulyLost = pending − new" filter excluded round-tripped partitions, so no `TrafficSourceReaderInterruptedClose` was enqueued and the new-generation record fell through to the accumulator's defensive ERROR path. `TrackingKafkaConsumer.onPartitionsAssigned` now treats every pending-cleanup partition as truly lost and dispatches the synthetic close *before* the generation bump; `KafkaTrafficCaptureSource.readNextTrafficStreamSynchronously` drains and prepends synthetic closes from mid-`poll()`.

## Operator-facing observability

The native `kafka-consumer-groups.sh --describe` output the console wraps shows message-lag (`LAG = LOG-END-OFFSET - CURRENT-OFFSET`) but tells operators nothing about how *stale* the consumer position is in wall-clock time. For migration traffic, "the last record consumed was produced 8 minutes ago" is far more actionable than "50,000 messages behind."

`console kafka describe-consumer-group` now appends a **TIME LAG** section after the unmodified native describe table:

```
TOPIC  PARTITION  PROBED-OFFSET  TIMESTAMP (UTC)         TIME-LAG  NOTE
t1     0          100            2024-01-02T03:04:05Z    8m12s
t1     1          199            2024-01-02T03:11:50Z    27s       caught up; showing last record
t1     2          -              -                       -         partition empty
```

The native table is preserved verbatim (downstream scrapers keep working). Probe failures fall through to the original `CommandResult` with a logged warning. Per-partition probe failures (timeout, missing timestamp, parse error) render dashes and never abort the rest of the table. Implementation probes `CURRENT-OFFSET` for active-lag partitions, `LOG-END-OFFSET - 1` for caught-up partitions, and skips empty partitions entirely.

`assert_replay_drained()` is the test-side counterpart added in `cdc_base.py`: at `[replay-end]` it polls `describe-consumer-group` until per-partition LAG ≤ 1 (default), logs the snapshot, and raises `ReplayLagDrainTimeout` on stall. Without it, `verify_clusters` was happily passing whenever target docs matched — even though the replayer never committed past offset 0.

## SCRAM-SHA-512 auth fix for k8s

The migration-console StatefulSet does not mount `KAFKA_SCRAM_PASSWORD` or the Strimzi cluster CA cert file, but `_get_kafka_from_workflow_config` hardcoded references to both — causing `console kafka` commands to fail in any EKS deployment with SCRAM auth enabled.

`_resolve_strimzi_scram_credentials()` now reads the SCRAM password from the Strimzi KafkaUser Secret (`<cluster>-migration-app`) and the CA cert from `<cluster>-cluster-ca-cert` directly via the k8s API at runtime. The password is passed inline; the CA cert is written to a temp file. The migration-console RBAC rule for `kafka.strimzi.io` was extended with `delete`/`patch` (and the kafkatopics/kafkanodepools/strimzipodsets resources) so `workflow reset` can tear down KafkaUsers/Kafkas during cleanup.

## CI infrastructure fixes

- `wait-for-cluster` Argo template was silently succeeding on timeout (`set -e` missing, redundant unconditional "ready" print). Replaced both broken check blocks with one retry loop, increased the readiness timeout to 300s, and surfaced an explicit failure when neither pod nor deployment appears.
- Kyverno ghost mutating webhooks could survive a timed-out previous "Cleanup MA" stage and block the BuildKit Helm install with `mutate.kyverno.svc-fail: service "kyverno-svc" not found`. Added a defensive webhook cleanup at the start of "Build Docker Images."
- `generate-data` fails fast (rather than spinning forever) when bulk inserts make zero forward progress.
- Reverted sample-cdk to v0.4.1 — v0.4.2 inline AccessPolicies break runtime auth.
- DEBUG plumbing for the drain-stall investigation: log4j override is now sysprop-driven and parameterized so CDC integ tests can opt in without rebuilding the image.

## Tests

Replayer (Java): 106 trafficReplayer tasks pass under `--rerun-tasks`. New unit tests pin each fix:
- `HandleCompletedTransactionCommitTest` — synchronous tuple-writer failure still commits offsets.
- `TrafficSourceReaderInterruptedCloseWiringTest` (3 tests) — interrupted-close path drains correctly and does not skip `replayEngine.closeConnection`.
- `PartitionRevocationStaleStateTest` (5 tests) — channel-context manager and accumulator behaviour across generation bumps.
- `StaleAccumulationCancelOnRejoinTest` + `StaleAccumulationCancelOnRejoinKafkaTest` — synthetic-close ordering on round-trip, both in-memory and against an embedded Kafka.

Console (Python): 38 `test_kafka.py` pass; 11 `test_environment_workflow.py` pass; 466 in the rest of the console_link suite (excluding `workflow-tests/`). 14 new tests cover parser/formatter/probe-command/auth-flag/failure-isolation paths plus the SCRAM k8s secret resolution.

E2E (kind, both standard and SCRAM-SHA-512 paths): `console kafka list-topics` / `list-consumer-groups` / `describe-topic-records` / `describe-consumer-group` (with active-lag, caught-up, and empty-partition rows) all pass.

## Issues Resolved

N/A

## Is this a backport?

No.

## Check List

- [x] All tests pass — new and existing console_link unit tests pass locally; full trafficReplayer test suite passes under `--rerun-tasks`


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/migrationConsole.yaml
```diff
@@ -46,17 +46,24 @@ rules:
...
+  - apiGroups: ["kafka.strimzi.io"]
+    # Strimzi's topic operator owns kafkatopics; reset cascades through them too.
+    resources: ["kafkatopics", "kafkanodepools"]
+    verbs: ["get", "list", "watch", "delete", "deletecollection", "patch"]
   - apiGroups: ["core.strimzi.io"]
+    # strimzipodsets are Strimzi-generated and lag behind the Kafka CR in busy
+    # clusters; reset.py:_OWNED_RESOURCE_CLEANUP deletes them by label.
     resources: ["strimzipodsets"]
-    verbs: ["get", "list", "watch", "delete", "deletecollection"]
+    verbs: ["get", "list", "watch", "delete", "deletecollection", "patch"]
```
gregschohn: Why do we want to let users delete/update clusters, etc. Why not leave these fully under the management of workflow configure/submit/reset?

gregschohn (commented):
I'd like to re-view this after https://github.com/opensearch-project/opensearch-migrations/pull/3013 makes it into the mainline to make sure that I fully understand the kafka changes.  Other than that, this LGTM for right now.

I plan to do a follow-up PR to remove a lot of the parsing and resolution of kafka resources so that the console command derives those from the TS models and works well if there are 0-N kafka clusters being managed.  That's all for later though and for right now, all of these changes are better than what we have right now.

gregschohn (approved):

libraries/testAutomation/testAutomation/test_runner.py
```diff
@@ -418,7 +418,6 @@ def run(self, skip_delete: bool = False, keep_workflows: bool = False,
                                              target_version=target_version,
                                              keep_workflows=keep_workflows,
                                              reuse_clusters=reuse_clusters,
-                                             skip_workflow_reset=skip_delete,
```
gregschohn, line 421: don't need this any more?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -406,6 +406,31 @@ const DEFAULT_AUTO_CREATE_KAFKA = {
                         ],
                     },
                 },
+                // Spread brokers one-per-node so a single node disruption
```
gregschohn, line 409: we probably should drop podAntiAffinity

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -710,6 +712,20 @@ private void fireAccumulationsCallbacksAndClose(
                     accumulation.getLastTimestamp(),
                     getTrafficStreamsHeldByAccum(accumulation)
                 );
+            } else if (!expirationCommitted) {
+                // Held traffic stream keys must be committed even when no HTTP request was
+                // signaled. OffsetLifecycleTracker keeps any uncommitted entry at the head of
+                // its priority queue and every subsequent commit becomes BLOCKED_BY_OTHER_COMMITS,
+                // freezing the consumer-group offset indefinitely (observed in CDC E2E runs as
+                // 30m+ TIME-LAG).
+                var heldStreams = getTrafficStreamsHeldByAccum(accumulation);
```
gregschohn, line 721: This code path can be called when the replayer shuts down.  We could have been accumulating a response - so this would stop a replayer starting back up from replaying that message.

We should revert changes to this method.

TrafficCapture/tupleSink/src/main/java/org/opensearch/migrations/replay/sink/S3TupleSink.java
```diff
@@ -118,64 +125,120 @@ public S3TupleSink(
...
+     * Only rotates buffered tuples once the file has reached its max age (size/count rotation is
+     * handled inline in accept()). */
+    private void periodicFlushOnWorker() {
+        if (!pendingFutures.isEmpty() && hasReachedMaxAge()) {
             rotate(true);
         }
     }
 
     @Override
     public void close() {
```
gregschohn, line 188: We should make sure that we don't drop things on the floor until we're done waiting for all of the pending transactions to come back as responses.  We DO currently give the replayer time to flush remaining transactions through the system already.  We have an at-least-once delivery policy, but it's also nice to re-replay messages as few times as possible.

libraries/testAutomation/testAutomation/test_runner.py
AndreKurait: This was a bug

migrationConsole/lib/console_link/console_link/models/factories.py
```diff
@@ -65,13 +65,13 @@ def get_replayer(config: Dict, client_options: Optional[ClientOptions] = None):
     raise UnsupportedReplayerError(', '.join(config.keys()) if config else '<empty>')
 
 
-def get_kafka(config: Dict):
+def get_kafka(config: Dict, scram_password: Optional[str] = None):
```
AndreKurait, line 68: No

gregschohn (approved):
Please address the critical close issues and areas in dead code in a fast-follow

PR #2946, opened 2026-05-12
Workflow show improvements

gregschohn:
### Description
Improve workflow show and approve completions

Add resource and task tab completion for workflow show, rename the visible positional arguments to RESOURCE_NAME and TASK, and support `workflow show --all` for printing the latest artifacts across current resources. Make `workflow show` with no args print help.
Also dedupe workflow approve completions so already selected gate names are not offered again.
Add workflow CLI tests for show completion, --all/help behavior, task-name resource selection, and approve completion dedupe.

### Testing
Manual testing of the workflow command

### Check List


AndreKurait (approved):
I'm good with this as is, though i would love to excercise this in the minikube tests. E.g.

`workflow show` reads `snapshotmigrations` CRDs via `migrations.opensearch.org/v1alpha1` and fetches S3 artifacts referenced in their `.status.outputs` fields. The only tests for this are unit tests (in `test_workflow_cli_commands.py`) that mock the k8s client and `read_artifact_text`. The command has **never been validated against a real cluster with real CRDs and real artifacts**.

This is particularly relevant now that PR #2946 changes argument semantics (positional arg renamings, new `--all` flag, completion sources that enumerate resources). A real-cluster test would catch:
- CRD schema mismatches (if the `status.outputs` structure doesn't match what the code expects)
- S3/artifact store connectivity issues
- Completion functions timing out against clusters with many resources
- The new `--all` flag actually printing something useful after a migration runs

### Where `workflow show` Should Be Added

**Best fit: `ma_workflow_test.py`** (the minikube Jenkins tests)

The minikube tests already produce the exact state that `workflow show` reads from — after `workflow_perform_migrations()` completes, the `snapshotmigration` CRDs exist with output refs. The test could exercise `workflow show` between `workflow_perform_migrations()` and `verify_clusters()`:

```python
# In ma_workflow_test.py, after workflow_perform_migrations():
def _verify_workflow_show_outputs(namespace: str = "ma"):
    """Verify 'workflow show --all' returns output after migrations complete."""
    cmd = ["workflow", "show", "--all", "--namespace", namespace]
    result = subprocess.run(cmd, capture_output=True, text=True, timeout=60)
    if result.returncode != 0:
        logger.warning("workflow show --all failed: %s", result.stderr)
        return
    # At minimum, we expect some output referencing snapshotmigration resources
    assert "snapshotmigration" in result.stdout.lower() or "metadataEvaluate" in result.stdout, \
        f"workflow show --all produced no recognizable output: {result.stdout}"
    logger.info("workflow show --all output:\n%s", result.stdout)
```

**Secondary fit: `test_workflow_integration.py`** (Kind cluster)

This would require installing the `snapshotmigration` CRD and creating a test resource with mock output refs. More isolated but doesn't validate the full artifact-reading path. Example addition to `TestArgoWorkflows`:

```python
def test_workflow_show_with_real_crd(self, argo_workflows, test_namespace):
    """Test workflow show against a real snapshotmigration CRD with output refs."""
    # Install CRD, create resource with status.outputs, invoke workflow show
    ...
```

### Recommendation

Add a lightweight `workflow show --all` smoke test to `ma_workflow_test.py` as a post-migration assertion — it's free (the state already exists) and validates the full path from CRD to artifact content. The Kind cluster tests could add CRD-level testing later for the completion and argument-resolution logic specifically.

PR #2952, opened 2026-05-12
Remind users to clear target indexes after snapshot migration reset

AndreKurait:
## Description

When `workflow reset` deletes a `snapshotmigration` resource, the target cluster may still hold the indexes the migration populated. Reset only operates on migration CRDs and their owned k8s resources — it never touches the target cluster, so those indexes silently linger and can interfere with a rerun.

This adds a post-reset reminder, printed only when one or more `snapshotmigrations` were among the deleted targets, pointing users at the existing console command:

```
Note: reset does not delete indexes on the target cluster. If the snapshot
migration created indexes on the target, remove them with:
  console clusters clear-indices --cluster target
```

The reminder fires for both `workflow reset <name>` and `workflow reset --all` paths whenever the delete set includes a snapshot migration.

## Issues Resolved

N/A — UX improvement.

## Testing

- Added 3 unit tests in `test_reset.py`:
  - reminder shown when resetting a snapshotmigration by name
  - reminder not shown for non-snapshot-migration resets (e.g., trafficreplay)
  - reminder shown for `reset --all` when delete set includes a snapshotmigration
- All 22 tests in `tests/workflow-tests/test_reset.py` pass.
- flake8 clean.


jugal-chauhan (approved):
LGTM! This improves the user experience. 

I like the addition of new tests here. Can we also add a test for `warning` not being print when `_delete_targets` returns `False` in the `by-name` path ?

PR #2955, opened 2026-05-12
Force --allow-existing-indexes on EKS metadata migration

AndreKurait:
## Summary

Adds a `--allow-existing-indexes` flag to the `MetadataMigration` CLI so that pre-existing target indexes can be reported as a non-fatal `WARN` instead of a fatal `ERROR`, and forces it on for the EKS path so that re-applying a `SnapshotMigration` resource doesn't abort on indexes the prior run created.

- Standalone CLI default: `false` — existing fatal behavior preserved for direct callers and standalone integration tests (`EndToEndTest`, `TargetMetadataWriteEndToEndTest`, `ItemsTest`).
- EKS/K8s path: hardcoded `true` in the metadata-migration template's inline-JSON params dict. No new CRD field, no new user-config knob, no VAP rule. The flag is invisible to operators editing `SnapshotMigration` and to `values.yaml`.

## Why

In EKS/K8s, `metadata migration` and `document backfill` are two phases of one custom resource (`SnapshotMigration`). When that resource is reconciled again — after a workflow restart, after a config edit, after a partial failure — the metadata phase currently re-runs against indexes the previous backfill phase already created. The `IndexCreator` reports `INDEX_ALREADY_EXISTS`, which `Items` treats as fatal, the metadata phase exits non-zero, and the whole reconciliation aborts. The user has to either delete target indexes or narrow the allowlist before retrying — neither of which a controller can do automatically.

Earlier revisions of this PR exposed the toggle as a user-configurable schema field with a `gated` change restriction. That added CRD surface area for a knob whose `false` setting actively breaks the EKS reconciliation loop, so the option doesn't carry its weight. Hardcoding the override in the workflow template removes the footgun and keeps the user-facing config minimal.

For the standalone Java CLI, the existing fatal-on-conflict behavior is still appropriate — a one-shot operator running `MetadataMigration migrate` against a cluster they didn't expect to have these indexes wants to know about it. Hence the split default.

### `MetadataMigration/src/main/java/.../MigrateOrEvaluateArgs.java`
New `--allow-existing-indexes <bool>` JCommander parameter, default `false`. Description explains the idempotent re-run case.

### `MetadataMigration/src/main/java/.../cli/Items.java`
- Optional `allowExistingIndexes` builder field (default `false`).
- New private `isFatal(CreationResult)` that checks the flag at result-reporting time and downgrades `INDEX_ALREADY_EXISTS` to non-fatal when the flag is set.
- `getAllErrors()`, `failureMessage()`, and JSON output (`buildArray`) now route through `isFatal()` — so a downgraded result becomes `WARN` in CLI output, `fatal: false` in JSON, and is dropped from the top-level `errors` array (i.e. doesn't bump the exit code).
- The "To resolve…" suggestion block is suppressed when the result is no longer fatal, and gains a third bullet pointing users at `--allow-existing-indexes true` for the "resuming a previous run" case.

### `MetadataMigration/src/main/java/.../commands/MigratorEvaluatorBase.java`
Wires `arguments.allowExistingIndexes` into the `Items.builder()` call alongside the existing `succeedOnEmpty` bit.

### `orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/metadataMigration.ts`
In `makeParamsDict`, append `"allowExistingIndexes": expr.literal(true)` to the inline-JSON params dict next to `outputFile`. Because the override goes through `expr.mergeDicts` after the user options, it always wins, so the flag flows through as `--allow-existing-indexes true` to the MetadataMigration CLI. The override sits inside the metadata template, so it doesn't leak to SnapshotMigration spec, the CRD, or the user-facing config.

### Snapshot regen
- `metadataMigration.snap.json` — adds `"allowExistingIndexes", true` inside the `sprig.dict(...)` literal that becomes the metadata CLI inline-JSON args.
- `resourceManagement.snap.json` — diff is unrelated to this change (re-generated for consistency).
- `argoSchemaSnapshot.test.ts.snap` — no longer carries an `allowExistingIndexes` field on the user/argo schemas (reverted from the earlier schema-based approach).

## Testing

- `:MetadataMigration:test --tests org.opensearch.migrations.cli.*` — 60+ tests pass, including:
  - Existing `ItemsTest.testIndexAlreadyExistsSuggestions` — still asserts `ERROR`, `fatal: true`, suggestion block. Confirms the Java-CLI default is unchanged.
  - New `ItemsTest.testAllowExistingIndexesDowngradesIndexAlreadyExistsToWarn` — `WARN`, `fatal: false`, no top-level `errors` entry, no suggestion block.
  - New `ItemsTest.testAllowExistingIndexesLeavesOtherFatalErrorsAlone` — only `INDEX_ALREADY_EXISTS` is downgraded; `TARGET_CLUSTER_FAILURE` stays fatal and still appears in the `errors` array.
  - Existing `EndToEndTest.alreadyExists_indexFatalTemplatNonFatal` — still passes because it doesn't set the flag, so the Java CLI default of `false` keeps `INDEX_ALREADY_EXISTS` fatal.
- `npm test` (full `orchestrationSpecs` workspace) — 4 suites green: schemas (50), argo-workflow-builders (36 + 2 skipped), config-processor, migration-workflow-templates (snapshot regenerated to embed the override in the metadata CLI args dict).
- Verified by grep that the override only appears inside the `metadataMigration` template's params dict and not in the `resourceManagement.snap.json` SnapshotMigration manifest.

## Check List

- [x] New functionality includes testing (3 new `ItemsTest` cases; existing tests preserved as a guard against accidental Java-CLI default flip).
- [x] New functionality has been documented (CLI `--help` description, suggestion block).


jugal-chauhan (approved):
LGTM! Thanks for brining in these changes.

PR #2956, opened 2026-05-12
Fix Solr snapshot completion detection when async-task state lags

AndreKurait:
## Summary

`_accumulate_collection_result` (called from `_get_solr_snapshot_status` for the SolrCloud snapshot path) only marked a collection as completed when its REQUESTSTATUS top-level `state` field literally read `"completed"`. Solr's overseer task is asynchronous, so `state` can stay at `"running"` (or `"submitted"`) for a non-trivial window after every shard has actually finished. During that window the workflow saw output like:

```
Snapshot status: Running
Finished time:
Percent completed: 100.00%
Estimated time to completion: 0h 0m 0s
Data processed: 0.000/0.000 MiB
Total shards: 90
Successful shards: 90
```

…and the createSnapshot template's `checkScript` polled for `SUCCESS`, never saw one, and looped until its 200-attempt retry cap. The backup was actually done.

## Why

Looking at `_accumulate_collection_result`, the `else` branch (state ≠ `"completed"` and ≠ `"failed"`) unconditionally sets `accum["all_completed"] = False`. That flag is the only thing `_determine_solr_step_state` looks at, so a single collection in the lagging-state case — even with 30/30 shards reporting in `success` — keeps the whole snapshot status `RUNNING`.

The per-shard `success` map in REQUESTSTATUS is the ground truth: each entry there is a shard that reported a successful backup. If that map has as many entries as the collection has shards, the backup of that collection is done regardless of whether the async-task overseer has flipped the state field yet.

### `migrationConsole/lib/console_link/console_link/models/snapshot.py`

`_accumulate_collection_result`:

- Compute `completed = state == "completed" or (num_shards > 0 and success_count >= num_shards)`.
- When `completed`, increment `completed_shards += num_shards` and leave `all_completed = True` alone.
- Failure path (`state == "failed"`) is unchanged.
- Otherwise, the existing in-progress fallback applies: `all_completed = False`, `completed_shards += success_count`.
- Added a docstring noting the Solr overseer-lag behavior so the next reader doesn't reintroduce the bug.

### `migrationConsole/lib/console_link/tests/test_snapshot.py`

Five new tests covering the accumulator and the end-to-end status path:

- `test_accumulate_collection_completed_state_marks_completed` — happy path, state="completed".
- `test_accumulate_collection_running_state_with_all_shards_succeeded_marks_completed` — the regression: state="running", success_count == num_shards, must mark completed. Verified to fail before the fix and pass after.
- `test_accumulate_collection_submitted_state_with_partial_progress_stays_running` — guard against false positive: state in-progress with partial shards stays RUNNING.
- `test_accumulate_collection_failed_state_marks_failed` — failure path preserved.
- `test_solr_snapshot_status_returns_success_when_all_shards_completed_with_lagging_state` — full `_solr_backup_status(deep_check=False)` simulation that reproduces the user's scenario (3 collections, 30 shards each, all 30 in success, top-level state still "running") and asserts the result is the literal string `SUCCESS` that the workflow's `checkScript` polls for.

## Testing

- `python -m pytest tests/test_snapshot.py -k "accumulate_collection or solr_snapshot_status or snapshot_status"` — all 9 tests pass (4 existing + 5 new).
- Sanity-check: reverted only the source change and re-ran. The new tests fail with `assert False is True / "all shards successful must imply completed regardless of async-task state"`, confirming they actually exercise the regression rather than the new code path.
- `flake8 console_link/models/snapshot.py tests/test_snapshot.py` — clean.

The other unrelated `mocker` fixture failures in `test_snapshot.py` (e.g. `test_fs_snapshot_create_calls_subprocess_run_with_correct_args`) are pre-existing — they require `pytest-mock` which isn't installed in the local venv. None are touched by this change.

## Check List

- [x] New functionality includes testing — 5 new pytest cases, including one that reproduces the exact production scenario.
- [x] New functionality has been documented — docstring on `_accumulate_collection_result` explaining the Solr overseer-lag behavior.


jugal-chauhan (approved):
Thanks for investigating and addressing this bug! The `success_count >= num_shards` heuristic is correctly used as the ground-truth as per-shard success reports are more reliable.

LGTM

Issue #2961, opened 2026-05-13
Sourceless RFS reconstruction: unbounded per-segment per-field spill causes worker ENOSPC and lease-cycle stalls


AndreKurait:
# Sourceless RFS reconstruction: unbounded per-segment per-field on-disk spill causes worker ENOSPC and lease-cycle stalls

## Summary

Sourceless RFS document reconstruction (introduced by #2941) writes per-(segment, field) sidecars to a worker-local spill directory under `<lucene-dir>/.rfs-spill/seg-N/<field>/`. The current implementation has **no aggregate cap on spill bytes**, **no per-field eviction within a segment**, and **no fail-fast / disk-precheck**. For an index with multiple analyzed-text fields and an analyzer whose token expansion is non-trivial (e.g. `pattern_capture`), the worker's scratch volume is amplified by ~14× the indexed segment size, exhausting ephemeral storage and triggering a lease-expiry restart loop.

## Reproduction

Snapshot of an OpenSearch 7.10-compatible index with:

- one shard ~50 GiB indexed
- mapping that excludes large analyzed text fields from `_source` (`_source.excludes`: `texts.user.content`, `files.content`, `actions.content`, `policies.content`, `ed.comment`, `sv.curr.comment`, `sv.past.comment`, etc.)
- analyzer with token expansion (`pattern_capture` filter, `preserve_original: true`)

Run `migration-assistant` mainline against this snapshot with `enableSourcelessMigrations: true` on a worker with 200 GiB ephemeral storage.

Observed:

| Metric | Value |
|---|---|
| Single-field `sort-input.bin` peak | **91.9 GiB** |
| `<lucene-dir>/.rfs-spill/` peak | **145.5 GiB** |
| Indexed segment size | ~10 GiB |
| Amplification | **~14×** |
| Outcome | Worker JVM exit 1/4 → lease expires → coordinator reassigns same partition → restart loop (4–6 restarts/worker observed) |

## Root cause

Three architectural defects compound:

### 1. Unbounded raw `sort-input.bin`

`SidecarBuilder` writes one 20-byte tuple `(docId, position, termId, startOffset, endOffset)` per analyzed token occurrence to `sort-input.bin` in entirety **before** `OfflineSorter` runs. Per-field worst-case size:

```
size(sort-input.bin) = total_tokens_for_field_in_segment * 20 bytes
                    ≈ raw_field_bytes * tokens_per_byte * 20
```

[...]

PR #2962, opened 2026-05-13
feat(rfs): cap sourceless reconstruction spill bytes with fail-fast budget

AndreKurait:
### Description
Sourceless reconstruction (SearchSnapshotExtractor) writes per-field spill files via `SidecarBuilder` during external merge sort. On hot-loop or large-postings shapes, the spill tree can grow to many multiples of the source segment, eventually filling the worker's disk and surfacing only as a generic `ENOSPC` `IOException` deep in `OfflineSorter` — by which point the worker lease is wedged for minutes (#2961).

This PR is the first of three follow-ups to that issue. It adds a **fail-fast aggregate budget** at the spill write boundary so operators can put a hard ceiling on bytes-in-flight without changing the per-builder algorithm.

Two opt-in JVM-level knobs:

| Property | Default | Effect |
|---|---|---|
| `rfs.reconstruction.maxSpillBytes` | `Long.MAX_VALUE` (unbounded) | Aggregate ceiling on bytes-in-flight in spill files across all concurrent `SidecarBuilder`s in this JVM. |
| `rfs.reconstruction.spillFreeSpaceMinBytes` | `0` (disabled) | Floor on the spill volume's usable space (`FileStore#getUsableSpace`). Coarse-grained probe (every 64 MiB written) catches the case where another process — or another shard's Lucene tree — fills the disk. |

When either limit is breached, `SidecarBuilder.accept()` throws the new `SpillBudgetExceededException` (an `IOException` subclass) instead of letting the write run to ENOSPC. The segment `Flux`'s existing `concatMapDelayError` path treats it as a per-segment failure: the spill tree is freed on `close()`, the worker lease is released cleanly, and neighboring segments keep making progress.

### Issues Resolved
First of three PRs addressing #2961. Follow-ups will land:
- Per-field LRU eviction in `SegmentTermIndex` (PR-2)
- Eager disk-space precheck in `SidecarBuilder` constructor (PR-3)

### Implementation notes
- **Bookkeeping** is a JVM-wide `AtomicLong` in the new `SpillBudget` singleton. Reservation is balanced with release on `close()`, so a failed segment doesn't permanently shrink headroom for the rest of the worker.
- **Reservation points:**
  - `accept()` reserves `positionCount × RECORD_BYTES` per call (the actual bytes that hit `sort-input.bin`).
  - `buildAndOpenReader()` reserves an additional `~peak` for the `OfflineSorter` merge phase (transient runs roughly double the on-disk footprint), then releases the full reservation after the sort scratch is unlinked from disk.
  - `close()` releases any leftover reservation on the failure path.
- **Behavior with both knobs unset is byte-for-byte identical** to the existing default: reservations against an unbounded cap are accounting-only and the volume probe short-circuits.

### Testing
- New `SourcelessSpillConfigTest` (5 cases) — pins `-D` parsing, defaults, clamp, and parse-failure fallback for all three props.
- New `SidecarBuilderBudgetTest` (3 cases) — drives the cap path end-to-end:
  - unbounded default accepts large writes,
  - breach throws the typed exception with the prop name in the message,
  - partial-progress-then-throw leaves bookkeeping consistent (failed reservation rolls back, successful one stays committed until `close()`).
- Existing `SidecarBuilderRoundTripTest` (9 cases) unchanged — all still pass.
- Full `SearchSnapshotExtractor:test` suite green locally.

```text
SourcelessSpillConfigTest                 5/5 PASSED
SidecarBuilderBudgetTest                  3/3 PASSED
SidecarBuilderRoundTripTest               9/9 PASSED
```

### Check List
- [x] All tests pass
- [x] New functionality has been documented (Javadoc on every new public surface + the three properties block in `SourcelessSpillConfig`)
- [x] Commits are signed per the [DCO](https://github.com/opensearch-project/.github/blob/main/CONTRIBUTING.md) using `--signoff`

### DIFF
--- a/SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourcelessSpillConfig.java
+++ b/SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourcelessSpillConfig.java
@@ -1,5 +1,8 @@
 package org.opensearch.migrations.bulkload.lucene;
 
+import java.io.IOException;
+import java.nio.file.FileStore;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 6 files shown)


AndreKurait:
I'm considering moving away from this all together as heap based is much faster

Issue #2947, opened 2026-05-12
Helm charts deploy resources with references to images tagged to 'latest'


gregschohn:
### What is the bug?
A user deploys helm 3.2.0 (or any version).
MA is configured to launch images for migration-console, RFS, proxy, replayer, etc with the image tag set to 'latest'
When a new version is released, the latest tags are updated.
When pods/nodes are recycled or recreated in the user's deployed 3.2.0 deployment, they'll pick up the latest images - e.g. 3.3.0.
That's bad enough, but they won't do that uniformly and users will be stuck in a very frankenstein environment.

Users would be hit by this whenever we do a deployment.  They have very little control to prevent it w/out overriding the image tags manually when doing a helm deployment.

### What are your migration environments?
Any helm/EKS release has this bug.

### How can one reproduce the bug?
Wait for a release & don't upgrade.

### What is the expected behavior?
When a user installs version X.Y.Z, all of the images and resources stay at X.Y.Z unless the user explicitly changes them through an upgrade or manual patch.  MA should version lock all of its resources.

PR #2970, opened 2026-05-14
Use chart appVersion for default migration image tags

gregschohn:
### Description
Default migration image tags to Chart.appVersion instead of 'latest' when values omit an explicit tag. The default value of appVersion for the Chart.yaml is 'latest' so source/development installs will still use 'latest'. This lets release-packaged Helm charts resolve project images to VERSION via the existing --app-version packaging step.

Make sure that the gha release-drafter validate that release tag pushes match VERSION.

Some helm lint errors popped up, so the templates were tidied up too.

### Issues Resolved
https://github.com/opensearch-project/opensearch-migrations/issues/2947

### Testing
Tested helm installs as a dev (no build) and a helm install after running `helm package deployment/k8s/charts/aggregates/migrationAssistantWithArgo --version 3.2.0 --app-version 3.2.0 `

### Check List

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -39,7 +39,12 @@ jobs:
       - id: get_data
         run: |
           echo "approvers=$(cat .github/CODEOWNERS | grep @ | tr -d '*\n ' | sed 's/@/,/g' | sed 's/,//1')" >> $GITHUB_OUTPUT
-          echo "version=${GITHUB_REF#refs/tags/}" >> $GITHUB_OUTPUT
+          version="$(tr -d '[:space:]' < VERSION)"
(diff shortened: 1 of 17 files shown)


.github/workflows/release-drafter.yml
```diff
@@ -14,6 +14,7 @@ env:
 permissions:
   id-token: write
   contents: write
+  issues: write
```
AndreKurait: @gregschohn  this isn't using issues anymore so we can remove this

gregschohn: i can't figure out why this appeared.  I don't see it as a change that I made.  You're to blame for it from many months ago.

AndreKurait: This is a merge conflict issue since it was removed in one of the commits on main

AndreKurait (approved):

PR #2976, opened 2026-05-14
Print advisory about surviving snapshot data on workflow reset datasnapshot

### LINKED ISSUE #2960: [BUG] `workflow reset datasnapshot.<name>` leaves the `source-cluster` snapshot in place
### What is the bug?
`workflow reset datasnapshot.<name>` deletes the DataSnapshot CR, and the snapshot data the CR pointed at appears to remain in the configured `source-cluster` snapshot repository (and its backing S3 prefix) after the reset. I was able to observe it by going to the source cluster directly and reading its `_snapshot/<repo>/_all` listing.

This may well be by design : `snapshot data` is owned by the `source-cluster` snapshot repository. We just want to surface the observation : the CR carries `.status.snapshotName`, the console already knows where the repo lives (via `sourceClusters.<source>.snapshotInfo.repos`), and right now the reset stops at the CR boundary without surfacing what survived. 

If the intent is to retain snapshots, a one-line note from the `workflow reset` output (using `aws s3` API) would be beneficial. If cleanup was intended, the data could plausibly be wired to follow the CR.

- MA version: 3.2.0
- Deployment: EKS, region `us-east-1`
- S3 repo: `s3://migrations-default-116981770455-jdc-0508-us-east-1/jdc-0508-snap`
- CloudWatch log group: `/migration-assistant-jdc-0508-us-east-1/logs`

Run a basic RFS migration workflow and wait until `datasnapshot.source-snap1` reaches `Completed`.

```
$ kubectl -n ma exec migration-console-0 -- workflow reset --list
…
       datasnapshot.source-snap1     (Completed) (depends on: capture-proxy)
…
```

Capture the snapshot name the CR points at:
```
$ kubectl -n ma get datasnapshot source-snap1 -o jsonpath='{"snapshotName="}{.status.snapshotName}'
snapshotName=source_snap1_1778642920
```

Source repo currently has exactly that one snapshot:
```
$ kubectl -n ma exec migration-console-0 -- console clusters curl source -- "_snapshot/default/_all" \
    | jq '[.snapshots[] | {snapshot, state}]'
[ { "snapshot": "source_snap1_1778642920", "state": "SUCCESS" } ]
```

S3 has its data:
```
$ aws s3 ls s3://<bucket>/<prefix>/ | grep snap-
2026-05-12 22:33:52   513  snap-tDarfI2jTkS9NC5An07y_Q.dat
```

Reset the CR:
```
$ kubectl -n ma exec migration-console-0 -- workflow reset datasnapshot.source-snap1
  Deleted datasnapshot.source-snap1
```

Verify CR is gone:
```
$ kubectl -n ma get datasnapshot
No resources found in ma namespace.
```

But noticed the source-side snapshot survives:
```
$ kubectl -n ma exec migration-console-0 -- console clusters curl source -- "_snapshot/default/_all" \
    | jq '[.snapshots[] | {snapshot, state}]'
[ { "snapshot": "source_snap1_1778642920", "state": "SUCCESS" } ]
```

and so does its S3 data:
```
$ aws s3 ls s3://<bucket>/<prefix>/ | grep snap-
2026-05-12 22:33:52   513  snap-tDarfI2jTkS9NC5An07y_Q.dat
```

A clean `workflow reset datasnapshot.<name>` should leave the user with no orphan data in the snapshot repository OR, if intention is to prefer to preserve repo contents by default, the console should point the user at the surviving snapshot name and suggest a command to remove it.

A few things I noticed while looking around : 
- The DataSnapshot CR seems to have `.status.snapshotName` (the per-run snapshot id, e.g. `source_snap1_1778642920`), and the workflow config has `sourceClusters.<source>.snapshotInfo.repos.<repo>` for the repo name + S3 location. Between those steps, the values needed to call `_snapshot/_delete` look reachable, if cleanup turns out to be the right direction.
- The existing `--keep-output-artifacts` flag on `workflow reset` could be a nice reference point. It already expresses "preserve durable storage on reset" semantics for the `migration-outputs/` prefix, so a similar shape for snapshot data would feel familiar to existing users.

Davidding4718:
## Description

When `workflow reset datasnapshot.<name>` is run, the DataSnapshot CR is deleted but the actual snapshot data remains in the source cluster's snapshot repository (and its S3 backing store). This leaves orphaned data that the user may not be aware of.

This PR adds an advisory message that is printed **before** the CR is deleted, informing the user:
- The snapshot name that will survive the reset
- The source cluster label
- A suggested `_snapshot/<repo>/<name>` DELETE command to clean it up manually

### Example output

```
⚠️  Snapshot data will survive this reset:
   Snapshot name: source_migration-snapshot_1778534584
   Source cluster: source
   To delete it manually, run:
     console clusters curl source -- "_snapshot/<repo-name>/source_migration-snapshot_1778534584" -X DELETE
   Or use: console snapshot delete

  Deleted datasnapshot.source-migration-snapshot
```

### Issues Resolved

Fixes #2960

### Testing

- Added 3 unit tests in `TestSnapshotAdvisory` class
- All 22 tests in `test_reset.py` pass
- Manually reproduced the issue on an EKS cluster and confirmed the advisory prints correctly

### Check List

### DIFF
--- a/migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
+++ b/migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
@@ -328,9 +328,34 @@ def _cleanup_output_artifacts(plural, name, uid, created_at=None):
         click.echo(f"  Warning: {e}", err=True)
 
 
+def _snapshot_advisory(namespace, plural, name):
+    """Print advisory about surviving snapshot data when deleting a datasnapshots CR."""
+    if plural != 'datasnapshots':
+        return
+    try:
+        custom = client.CustomObjectsApi()
+        item = custom.get_namespaced_custom_object(
+            group=CRD_GROUP, version=CRD_VERSION,
+            namespace=namespace, plural=plural, name=name,
+        )
+        snapshot_name = item.get('status', {}).get('snapshotName')
+        if not snapshot_name:
+            return
+        click.echo(
+            f"\n⚠️  Snapshot data will survive this reset:\n"
+            f"   Snapshot name: {snapshot_name}\n"
+            f"   To delete it manually, run:\n"
+            f"     console clusters curl source "
+            f"\"_snapshot/<repo-name>/{snapshot_name}\" -X DELETE\n"
+        )
+    except ApiException:
+        pass
+
+
 def _delete_and_wait(namespace, plural, name, delete_output_artifacts=True):
     """Delete a single CRD and poll until it is gone."""
     uid, created_at = _resource_metadata(namespace, plural, name)
+    _snapshot_advisory(namespace, plural, name)
     _mark_deleting(namespace, plural, name)
     _cleanup_owned_resources(namespace, plural, name)
     if delete_output_artifacts:
--- a/migrationConsole/lib/console_link/tests/workflow-tests/test_reset.py
+++ b/migrationConsole/lib/console_link/tests/workflow-tests/test_reset.py
@@ -12,6 +12,7 @@ from console_link.workflow.commands.reset import (
     _find_ancestors,
     _find_dependents,
     _prune_ancestors_of_protected_proxies,
+    _snapshot_advisory,
 )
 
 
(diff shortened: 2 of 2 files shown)


migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
```diff
@@ -328,9 +328,38 @@ def _cleanup_output_artifacts(plural, name, uid, created_at=None):
...
+        if not snapshot_name:
+            return
+        labels = item.get('metadata', {}).get('labels', {})
+        source_label = labels.get('migrations.opensearch.org/source', 'source')
+        click.echo(
+            f"\n⚠️  Snapshot data will survive this reset:\n"
+            f"   Snapshot name: {snapshot_name}\n"
+            f"   Source cluster: {source_label}\n"
+            f"   To delete it manually, run:\n"
+            f"     console clusters curl {source_label} -- "
```
AndreKurait: i don't believe `source_label` here is correct since console clusters curl only supports source and target (maybe proxy)

AndreKurait (changes requested):

migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
```diff
@@ -328,9 +328,38 @@ def _cleanup_output_artifacts(plural, name, uid, created_at=None):
...
+            f"\"_snapshot/<repo-name>/{snapshot_name}\" -X DELETE\n"
+            f"   Or use: console snapshot delete\n"
```
AndreKurait: 1. On EKS, Context.__init__ (cli.py:78) detects the K8s ConfigMap-based workflow config store via can_use_k8s_config_store() and calls
  Environment.from_workflow_config(...).
  2. from_workflow_config (environment.py:160-187) wires up target_cluster, source_cluster, and kafka — but explicitly hard-codes instance.snapshot =
   None (line 185), instance.metadata = None, instance.backfill = None. There's no codepath that populates snapshot from the K8s workflow config.
  3. The console snapshot group (cli.py:471-474) gates every subcommand:
  def snapshot_group(ctx):
      if ctx.env.snapshot is None:
          raise click.UsageError("Snapshot is not set")
  4. Therefore console snapshot delete (and create, status, unregister-repo) always raise UsageError("Snapshot is not set") on EKS, regardless of the
   underlying source cluster being reachable.

migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
Davidding4718: Good catch. I've removed the dynamic `source_lable` and replaced it with `source`.

migrationConsole/lib/console_link/console_link/workflow/commands/reset.py
Davidding4718: I wasn't aware that `Environment.from_workflow_config` leaves `snapshot = None` on EKS. I've removed the `console snapshot delete` suggestion entirely and now only suggest the `console clusters curl source` command

PR #2981, opened 2026-05-15
RFS: unify ES and Solr migration orchestration to prevent exit-code divergence

AndreKurait:
## Summary

**Commit 1: Unify orchestration** — Extracts the duplicated orchestration scaffolding (work coordinator, shutdown hook, exit-code handling) from the ES and Solr code paths into a single `runMigration()` method. Introduces `MigrationSourceFactory` functional interface so each source only provides pipeline construction logic.

**Commit 2: Shared Lucene utilities** — Extracts two shared utilities that both the ES and Solr reading paths can use:

- **`SegmentDocStream`**: Shared segment iteration (sort by name, cumulative docBase, binary-search resume, bounded-concurrency flatMapSequential). Replaces the duplicated loop in `LuceneReader` and `SolrBackupSource`.
- **`StoredFieldDocReader`**: Shared stored-field-to-JSON assembly (ID extraction, internal-field filtering, multi-value accumulation, mapping-aware type conversion). Replaces `SolrLuceneDocReader.extractFieldValue()` and the stage-1 path of `SourceReconstructor`.

Additionally wires `FieldMappingContext` into the Solr path: `SolrMultiCollectionSource` now converts the Solr schema into a `FieldMappingContext` (via `SolrSchemaConverter`), so Solr stored-field extraction gets proper type-aware conversion (dates → ISO, booleans, IPs, scaled_float) for free.

## Motivation

The Solr and ES paths had duplicated orchestration and Lucene reading logic. When behavior changed in one path (like the exit-code fix in #2965), the other path would silently diverge. By unifying:
- Exit-code handling is structurally guaranteed consistent (prevents the #2965 class of bug)
- Segment iteration is single-source-of-truth
- Stored-field reading with type conversion is shared (Solr gets richer type handling for free)

## Test plan

- [x] `./gradlew :DocumentsFromSnapshotMigration:compileJava` passes
- [x] `./gradlew :DocumentsFromSnapshotMigration:compileTestJava` passes
- [x] `./gradlew :SolrReader:compileJava` passes
- [x] `./gradlew :SearchSnapshotExtractor:compileJava` passes
- [ ] Existing `SolrSnapshotToOpenSearchTest` integration tests pass (require Docker)
- [ ] Existing ES snapshot integration tests pass (require Docker)
- [ ] Follow-up: migrate `LuceneReader` to use `SegmentDocStream` (left for a separate PR to keep this one focused)

### DIFF
--- a/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
+++ b/DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
@@ -464,11 +464,20 @@ public class RfsMigrateDocuments {
         }
     }
 
+    @FunctionalInterface
+    interface MigrationSourceFactory {
+        CompletionStatus buildAndRun(
+            IWorkCoordinator workCoordinator,
+            LeaseExpireTrigger processManager,
+            AtomicReference<WorkItemCursor> progressCursor,
+            AtomicReference<Runnable> cancellationRunnableRef,
+            WorkItemTimeProvider workItemTimeProvider
+        ) throws IOException, InterruptedException, NoWorkLeftException;
+    }
+
     public static void main(String[] args) throws Exception {
         var workerId = ProcessHelpers.getNodeInstanceName();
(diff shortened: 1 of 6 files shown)


AndreKurait:
Superseded by #2982 which has a leaner approach (net -65 lines instead of +362).

PR #2985, opened 2026-05-15
Sourceless RFS + analysis-compat transform: streaming postings (OOM), copy_to/object-array recovery, position-gap stopword

AndreKurait:
## Summary

Two co-equal themes:

1. **End-to-end sourceless `_source` reconstruction for RFS** — full recovery chain from Lucene segments when `_source` is disabled or partially excluded. Adds streaming postings reconstruction, multi-term streaming (OOM fix), copy_to reverse-derivation, object-array subfield distribution, position-gap stopword preservation, and several bug fixes uncovered during regression testing on customer-shaped mappings.
2. **Preemptive analysis-component compatibility transform** — strips/renames analyzer, tokenizer, char_filter, and token-filter names that the target cluster will reject (e.g. `standard` filter, `delimited_payload_filter`, camelCase nGram/edgeNGram). Runs at metadata-transform time so the create-index/template request never carries the offending name. Coverage spans ES 1.x→9.x, OS 1.x→3.x, Solr 6.x→9.x, including ES 7 sources holding ES 6 restored indices.

### Performance

- **Streaming field postings via min-heap** — avoids materializing full term indexes to disk via OfflineSorter; one `PostingsEnum` per term kept in a min-heap, drained on-demand per document
- **Primitive long sort in Lucene 9** — uses JDK 21's vectorized `Arrays.sort(long[])` intrinsic with bit-packed `(pos << 32 | idx)` encoding
- **DocValues iterator caching per-segment** — eliminates per-call iterator re-acquisition overhead

### Memory: bounded segment-wide multi-term recovery (OOM fix)

The eager `Map<Integer, List<String>>` built by `buildMultiTermIndex` / `buildSingleTermIndex` materialized the entire field's posting list per `(segment, field, worker)`. On high-cardinality analyzed-text fields this produced 12+ GB of heap per worker and OOMed at `READER_PARALLELISM` workers.

- **`StreamingMultiTermPostings`** — FREQS-only streaming counterpart of `StreamingFieldPostings`. One `PostingsEnum` + one decoded term `String` per term; `advance(docId)` drains heads in dictionary order. Memory bounded by `uniqueTerms × ~80 B`, not by corpus size.
- **`AbstractStreamingPostingsHeap`** — shared min-heap skeleton extracted from `GenericStreamingFieldPostings`. Both streaming paths now share heap, sift-down, skip-to-doc, lifecycle, and close-eligibility logic in one place. Subclasses provide only the per-doc emission strategy via `drainAtDoc(docId)`.
- **`SegmentTermIndex`** — replaces `singleTermByField` + `multiTermByField` with one `multiTermStreamingByField`. `getSingleTermForDocument` reduces to "first emitted term" of the same streaming cursor. Deterministic close via the segment Flux's `doFinally`; per-field heap entries are field-granular, segment-bound, and never escape the worker pipeline.
- **Per-version `LuceneStreamingAdapters`** (v5/v6/v7/v9/v10) — one `wrap(PostingsEnum)` helper per Lucene shadow replaces 7 inline anonymous-class adapter blocks across the LeafReaders.
- **One `SegmentTermIndex` per worker** — fixes a parallel-worker race where forward-only streaming cursors were silently dropping documents under round-robin partitioning. Each worker's bucket is now a strictly-ascending docId subsequence.
- **Sidecar disk-spill path removed** — `SidecarBuilder` / `SidecarReader` / `BytesRefLike` / `SourcelessSpillConfig` and ~1200 LOC of OfflineSorter-based fallback deleted; the streaming-postings path fully supersedes it.
- **Dead-code removal** — `buildSingleTermIndex` / `buildMultiTermIndex` deleted from the interface and from `LeafReader{9,10}`. No remaining callers anywhere in the repo; never reachable after the rewrite.

Net heap-cost change per `(segment, field, worker)`:

| Path | Before | After |
|------|--------|-------|
| Multi-term recovery | `O(docs × tokensPerDoc × 60 B)` | `O(uniqueTerms × ~80 B)` |
| Single-term recovery | `O(docs × ~40 B)` HashMap | shared with multi-term cursor |

### Source reconstruction features

- **Copy_to reverse-derivation** — when a source field has no stored/doc_values/points data, recovers its value from copy_to targets (keyword targets preferred over text)
- **Object-array subfield distribution** — distributes recovered values across seeded `List<Map>` elements from partial `_source`
- **Per-element text reconstruction** — position-gap splitting (gap ≥ 100) recovers individual array element values for text fields
- **Best-effort prefix distribution** — when fewer values than array elements (SORTED_SET dedup or empty-string elements), distributes available values into prefix elements
- **Position-gap stopword preservation** (`--position-gap-stopword`, default `a`) — when ES indexes `"i like the tree"` with stopword `the`, the postings carry positions `[0:i, 1:like, 3:tree]` (position 2 consumed but term dropped). Naive space-join produced `"i like tree"` which OS re-tokenizes at `[0,1,2]`, silently changing slop / proximity / phrase semantics on every migrated doc with a stopword in an analyzed-text field. The flag splices `(positionGap - 1)` copies of the configured stopword into each gap so OS re-creates the original increments. Default `a` works for the english/standard analyzers; pass `--position-gap-stopword=""` to keep the legacy multi-space behaviour.

### Pipeline / CLI

- **`--emit-doc-type` is now a tri-state (`AUTO` / `ON` / `OFF`, default `AUTO`)** — auto-enables `_type` emission only when the source is ES ≤ 6 *and* a document transformer is configured (the combination that exercises `TypeMappingSanitizationTransformerProvider` on multi-type indices and was previously crashing the bulk-loader with a `PolyglotList → Map` `ClassCastException`). Surfaced through `USER_RFS_PROCESS_OPTIONS` and the `SnapshotMigration` CRD spec.
- **`RfsTunables`** — new constants class cataloging JVM system-property and env-var tunables for the sourceless reconstruction pipeline. Default reader parallelism changed from 1 to `availableProcessors-1` (min 1).

### Bug fixes

- **Copy_to target leakage** — reverse-derivation no longer writes into source fields with `index:false, doc_values:false` (no Lucene footprint, only exists in `_source`)
- **Source-excluded copy_to targets recovered** — fields like `texts.user.content` that are both a copy_to target AND source-excluded are correctly recovered from the inverted index
- **`matchesGlob` sub-path matching** — ES semantics: `"texts"` in `_source.excludes` now matches `texts.user.content`, `texts.sys.attrs`, etc.
- **Length-based de-dup when offsets are absent** (`#42`) — keyword recovery uses term length instead of offsets when the postings format doesn't store them
- **SonarQube S1186/S3776/S6541** — reduced cognitive complexity in `StreamingFieldPostings`, added comment to empty `close()`
- **`TermEntry` 4-arg constructor** — fixed call sites after position field addition

## Analysis-component compatibility transform

`AnalysisCompatibility` defines a per-(sourceVersion, targetVersion) table of analyzer/tokenizer/char_filter/token-filter components to **strip** or **rename** before the create-index/template request goes out. The reactive `InvalidResponse`-based retry path is preserved as a safety net; the preemptive transform just prevents the round-trip failure when we already know the name will be rejected.

- **Coverage** — ES 1.x→9.x, OS 1.x→3.x, Solr 6.x→9.x. Source predicates are version-pair-aware: ES 7 sources fire ES 6 legacy rules because ES 7 clusters can hold ES 6 indices restored via snapshot (e.g. legacy `standard` filter, `delimited_payload_filter`, camelCase `nGram` / `edgeNGram`, `htmlStrip`).
- **`MetadataTransformationRegistry`** — wires the table into the existing baked-in transformer chain so it runs alongside the field-data-type, knn_vector, and flattened transforms.
- **JS rule applier** (`transformation/standardJavascriptTransforms/src/analysis-component-removal.js`) — engine-agnostic, applies `removed` and `renames` rule sets to both nested and dotted `index.*` body forms; the rule set itself is computed in Java and passed in via the bindings context.

## ES tarball Docker hardening

`custom-es-images/dockerfiles/Dockerfile`: force HTTP/1.1, use curl exponential backoff (`--retry 10 --retry-max-time 600`), and verify `/opt/elasticsearch/bin/elasticsearch` exists and is executable before the layer is committed. Without the verification a transient HTTP/2 stream error from `artifacts.elastic.co` left a tagged image whose entrypoint died with exit 127, surfacing as `ContainerLaunchException` on every gradle-tests shard that needed that image.

## Tests

- `SourceReconstructorTest` — 110+ unit tests covering all recovery tiers, copy_to, object arrays
- `copyToTargets_neverLeakIntoSource` — verifies reverse-derivation doesn't leak into unrecoverable fields
- `sourceExcludedCopyToSource_reverseDerivesFromTarget` — verifies source-excluded copy_to source fields are recovered from their targets
- `mergeWithDocValues_fewerDocValues_distributesIntoPrefixBestEffort` — best-effort prefix distribution
- `reconstructSource_objectArrayFromTextTermList_duplicateValues` — full 4-element array with duplicates via positions
- `NoStoredSourceMigrationTest` — E2E guards on real ES 7.10 snapshot data; `testPositionBufferGrowAcrossFrequencyRegimes` exercises the per-worker `SegmentTermIndex` race fix
- `testSourceDisabledWithAnalyzerMigration_subjAndFsubj` — verifies position-gap stopword preservation end-to-end via `match_phrase` slop=0 on ES 6.8 → OS 2.19 with custom analyzer + stopword filter
- `AnalysisCompatibilityTest` — 32 tests covering version-pair boundaries, Solr-specific rules, same-version pairs, null returns
- `EndToEndTest` — ES 6 index restored into ES 7 → OS 2.x verifies the analysis-compat transform fires for ES 7 sources holding ES 6 indices
- `GenericStreamingFieldPostingsTest` / `GenericStreamingMultiTermPostingsTest` — heap iteration, forward-only semantics, offset collection, close lifecycle (zero Mockito)
- `FieldMappingContextTest` — 31 tests covering copy_to resolution, glob compilation, source filtering, nested fields
- `InvalidResponseTest` — 25 tests covering `getRemovedTokenFilters`, `containsAwarenessAttributeException`, `getIllegalArguments`, `getUnsupportedMappingParameters`
- `ObjectNodeUtilsTest` / `RfsTunablesTest` — supporting-utility coverage

## Follow-ups (out of scope)

- Empirical heap + throughput benchmark on the failing partition to quantify the OOM fix's savings against the eager-map baseline (memory-allocation profiler + per-segment heap-dump comparison).
- Pre-flight memory estimator using `Terms.size()` + `Terms.getSumTotalTermFreq()` to predict per-partition heap needs before the run.


DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -336,6 +346,22 @@ public static class ExperimentalArgs {
...
+                "by \"the\" but the term itself is dropped). Without this flag the reconstructor joins on spaces " +
+                "and OS re-tokenizes the document at consecutive positions [0,1,2], silently changing slop / " +
+                "proximity / phrase semantics. The reconstructor splices this token into the gap so OS — assumed " +
+                "to have the same token configured as a stopword — re-creates the original [0,1,3] postings while " +
+                "indexing. The token MUST be on the target's stopword list or it leaks into search results; " +
+                "'a' is a safe default for the english / standard analyzers. " +
+                "Pass an empty string to opt out and fall back to the legacy multi-space behaviour. " +
+                "Default: 'a'."
+        )
+        public String positionGapStopword = "a";
```
gregschohn, line 364: this is language specific, right?

gregschohn (approved):
This approach has a lot of caveats.  Memory consumption can be very high; we can only reconstitute documents on the source whose ingress is an identity function for a reconstructed document; and metadata is forcibly changed/dropped between solr and OS w/out much of an alternative or warning provided to the user.

Regardless of all that though, this is the best that we've got.  I've got some experiments to do staging work in the filesystem, but that suffers from having too many files and having many more parameters to manage.  As for the other gaps, those are issues with trying to do a reindex rather than a field-by-field direct reindex operation.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
AndreKurait: Yes, may have to be tuned based on the language

Issue #2132, opened 2026-01-07
[BUG] Error submitting workflow


nocoli:
### What is the bug?
Attempting to follow https://github.com/opensearch-project/opensearch-migrations/wiki/Backfill-Workflow and the `workflow submit` command is resulting in the following error in Cloudshell

```
(05:52:06) migration-console (~) -> workflow submit
NOT checking if all secrets have been created.  Run `workflow configure edit` to confirm
Initializing workflow from session: default
Submitting workflow to namespace: ma
ERROR:console_link.workflow.services.script_runner:Script failed with exit code 1
ERROR:console_link.workflow.services.script_runner:stderr: Error: Found 4 errors

Error submitting workflow: Command '['/root/configProcessor/createMigrationWorkflowFromUserConfiguration.sh', '/tmp/tmpcy9mouvd.yaml', '--prefix ma', '--etcd-endpoints http://etcd.ma.svc.cluster.local:2379']' returned non-zero exit status 1.
ERROR:console_link.workflow.commands.submit:Workflow submission failed
Traceback (most recent call last):
  File "/root/lib/console_link/console_link/workflow/commands/submit.py", line 139, in submit_command
    submit_result = runner.submit_workflow(config_yaml, args)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/lib/console_link/console_link/workflow/services/script_runner.py", line 205, in submit_workflow
    output = self.run_script("createMigrationWorkflowFromUserConfiguration.sh", None,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/lib/console_link/console_link/workflow/services/script_runner.py", line 113, in run_script
    return self.run(self.script_dir / script_name, input_data, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/lib/console_link/console_link/workflow/services/script_runner.py", line 75, in run
    result = subprocess.run(
             ^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/subprocess.py", line 571, in run
    raise CalledProcessError(retcode, process.args,
[...]

PR #2987, opened 2026-05-15
Use K8s cronjobs instead of argo-recurring checks for RFS and CreateSnapshot

gregschohn:
### Description
The RFS deployment's teardown has always been controlled by the workflow that spun it up - or adopted it.  However, there are several cases where the deployments (including the coordinator) may never be torn down.  The fix is to move the responsibility of teardown outside of the workflow.  That's accomplished here with a cronjob that's spun up right before we spin up the deployments.  The cronjob works through a state machine to manage updating status values on the CR and determining when to tear the resources down.  It also has checks to make sure that it itself won't run in perpetuity in cases where the workflow crashed or otherwise skipped the creation of the RFS deployment.

See the [design doc](https://github.com/gregschohn/opensearch-migrations/blob/cec21daac9fa78359c5539d97be8f5c8bc6c61da/docs/rfsCronJobMonitor.md) for how this is achieved in a way that should hold up to pressure.

I've also made the same update to createSnapshot checks - we now check those outside of the workflow.  These two checks make the workflow size more bounded and deterministic.  Transient errors and approval gate will still cause a variable number of retries, but those are capped.

As the cronjobs run and check for completion, they also update a status field in the parent CR.  That status field is used by workflow status/manage to show the progress - at least as of the last cronjob check. Notice that the cronjobs are currently configured to bump their wait time from 1 minute to up to 5 minutes in 1 minute increments.  I'll probably change that shortly to stay at 1 minute to keep the status values fresher.

### Issues Resolved
https://github.com/opensearch-project/opensearch-migrations/issues/2132 - because we now let the cronjob run as long as it needs to and the cronjob itself cleans itself up.

### Testing
Lots of manual testing for full migration configs.  Checked manually that status and manage show the new backfill stats.  Confirmed that the status in the CR looks reasonable - the datasnapshot summary at the end looks weird, but that's because our console command could use some improvement.  That can be fixed independently.

### Check List


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/templates/resources/migrationCrds.yaml
```diff
@@ -201,6 +201,36 @@ spec:
...
+                    message:
+                      type: string
+                    summary:
+                      type: object
+                      properties:
+                        shardsTotal:
+                          type: integer
+                        shardsSuccessful:
+                          type: integer
+                        shardsFailed:
```
AndreKurait, line 221: These should probably all be nullable for solr right?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/createSnapshot.ts
```diff
@@ -172,42 +127,44 @@ export const CreateSnapshot = WorkflowBuilder.create({
...
+            .addResources(DEFAULT_RESOURCES.SHELL_MIGRATION_CONSOLE_CLI)
+            .addEnvVar("CRONJOB_NAME", getSnapshotDoneCronJobName(b.inputs.dataSnapshotName))
+            .addEnvVarsFromRecord({
+                SESSION_NAME: b.inputs.dataSnapshotName,
+                ...workflowIdentityEnvVars(),
+                DATASNAPSHOT_NAME: b.inputs.dataSnapshotName,
+                DATASNAPSHOT_UID: b.inputs.dataSnapshotUid,
+                SNAPSHOT_NAME: b.inputs.snapshotName,
+                CONFIG_CHECKSUM: b.inputs.configChecksum,
+                CONSOLE_IMAGE: b.inputs.imageMigrationConsoleLocation,
```
AndreKurait, line 151: Why do we need console image here when we have 
```
            .addImageInfo(b.inputs.imageMigrationConsoleLocation, b.inputs.imageMigrationConsolePullPolicy)
```

orchestrationSpecs/packages/migration-workflow-templates/src/makeTemplates.ts
```diff
@@ -43,6 +45,18 @@ function createDirectoryIfNotExists(dirPath: string): void {
...
+function copyWorkflowScripts(outputDirectory: string): void {
+    const scriptsOutputDirectory = path.join(outputDirectory, workflowScriptsOutputDirectoryName);
+    fs.rmSync(scriptsOutputDirectory, {recursive: true, force: true});
+    fs.cpSync(workflowScriptsSourceDirectory, scriptsOutputDirectory, {recursive: true});
+    for (const entry of fs.readdirSync(scriptsOutputDirectory)) {
+        if (entry.endsWith(".sh")) {
+            fs.chmodSync(path.join(scriptsOutputDirectory, entry), 0o755);
+        }
+    }
+    console.log(`Workflow scripts copied successfully (${scriptsOutputDirectory}).`);
```
AndreKurait, line 57: nit: make this log debug

AndreKurait (approved):

PR #2990, opened 2026-05-15
Check for missing dependencies in the bootstrap script

epugh:
### Description
For folks running this script for the first time, they may not have the dependencies like `tac`, `aws` or `curl` installed locally.  `tac` is one that osx folks aren't likely to have.

Obviously after you run through things once you will have it!   

I was trying to not just list every command needed, like `docker` or `gradlew`....

### Testing
manual testing.

### Check List

### DIFF
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -640,6 +640,7 @@ if [[ "$deploy_cfn" == "true" ]]; then
     seen_file=$(mktemp)
     trap "rm -f '$seen_file'" RETURN
     while true; do
+      # awk reverses the event list (oldest-first) — portable substitute for `tac`, which isn't in the BSD userland on macOS.
       aws cloudformation describe-stack-events --stack-name "$cfn_stack_name" ${region:+--region "$region"} \
         --query 'StackEvents[].[EventId,Timestamp,ResourceStatus,ResourceType,LogicalResourceId,ResourceStatusReason]' \
         --output text 2>/dev/null | awk '{a[NR]=$0} END{for(i=NR;i;i--) print a[i]}' | while IFS=$'\t' read -r eid ts status rtype logical reason; do
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -533,7 +533,7 @@ check_existing_ma_release() {
 
 # Check required tools
 missing=0
-for cmd in jq kubectl tac aws curl; do
+for cmd in jq kubectl aws curl; do
   if ! command -v $cmd &>/dev/null; then
     echo "Missing required tool: $cmd"
     missing=1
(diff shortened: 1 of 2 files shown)


deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -533,7 +533,7 @@ check_existing_ma_release() {
 
 # Check required tools
 missing=0
-for cmd in jq kubectl; do
+for cmd in jq kubectl aws curl; do
```
gregschohn, line 536: Thanks!

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -551,6 +551,12 @@ fi
 
 # --- CFN deployment (optional) ---
 if [[ "$deploy_cfn" == "true" ]]; then
+  if ! command -v tac &>/dev/null; then
```
gregschohn: Let's ditch tac altogether - we're only using it to make the diagnostic output easier to read.
Claude gives me this alternative w/ awk...

```
while true; do
  aws cloudformation describe-stack-events --stack-name "$cfn_stack_name" ${region:+--region "$region"} \
    --query 'StackEvents[].[EventId,Timestamp,ResourceStatus,ResourceType,LogicalResourceId,ResourceStatusReason]' \
    --output text 2>/dev/null \
    | awk '{a[NR]=$0} END{for(i=NR;i>0;i--) print a[i]}' \
    | while IFS=$'\t' read -r eid ts status rtype logical reason; do
        grep -qxF "$eid" "$seen_file" 2>/dev/null && continue
        echo "$eid" >> "$seen_file"
        [[ "$reason" == "None" ]] && reason=""
        printf "  %-26s %-30s %-40s %s%s\n" "$ts" "$status" "$rtype" "$logical" "${reason:+  ($reason)}"
      done
  sleep 10
done
```

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -551,6 +551,12 @@ fi
...
+    echo "Missing required tool: tac (used for CloudFormation event streaming)" >&2
```
AndreKurait: @epugh What do you think about removing the dependency on tac altogether? 

```
  aws cloudformation describe-stack-events ... --output text 2>/dev/null \
    | awk '{lines[NR]=$0} END{for(i=NR;i>=1;i--) print lines[i]}' \
    | while IFS=$'\t' read -r ...
```

epugh: claude gave me a onel ine `awk` command...   You know, sometimes i fall in love with cute things like "tac" being the opposite of "cat".    I like the conciseness and "neato" of TAC.  

Having said that, this all gets simpler witn one line of unreadable `awk` and a comment.   I'm on onboard with that, give me a thumbs up or comment, and I'll tweak this PR to remove tac.

epugh: So I got excited and added the `awk` since I had it in my buffer.

sumobrian:
@epugh - Can you fix DCO issue.

AndreKurait (approved):

sumobrian (approved):

gregschohn:
Thanks for the find!  We're now deprecating the aws-bootstrap script in favor of the rust installer - partially because of finds like this (thanks!), so this is now a moot point.

PR #2994, opened 2026-05-16
Fix chunked Transfer-Encoding capture and Write-during-READS accumulator handling

AndreKurait:
## Summary

- **Capture proxy fix**: `PassThruHttpHeaders` was preserving `Content-Transfer-Encoding` (a MIME header) but missing `Transfer-Encoding` (the HTTP/1.1 header). This caused Netty's `HttpRequestDecoder` to not recognize chunked requests, producing premature `EOM(firstLine=-1, headers=-1)` and splitting the chunk terminator `0\r\n\r\n` into a separate Read observation after the EOM.
- **Replayer fix**: `CapturedTrafficToHttpTransactionAccumulator` did not handle Write/WriteSegment observations during `ACCUMULATING_READS` state, causing the state machine to get permanently stuck and eventually NPE on a subsequent SegmentEnd observation.
- **Improved asserts**: Added descriptive messages to segment invariant assertions for easier diagnosis of future issues.

## Root Cause

When a client sends an HTTP request with `Transfer-Encoding: chunked`, the capture proxy dropped the header before Netty's decoder could see it. This caused Netty to treat the request as having no body, emitting `LastHttpContent` immediately after headers. The chunked body data and terminator were then captured as subsequent observations, corrupting the replayer's state machine on long-lived connections with segment splits.

## Test plan

- [x] `ChunkedTransferEncodingCaptureTest` - verifies the proxy correctly handles chunked requests (single read, split at terminator, followed by normal request)
- [x] `WriteObservationDuringReadStateTest` - verifies the replayer handles Write-during-READS without NPE and recovers for subsequent requests
- [x] Tests verified to FAIL without the fix and PASS with it (TDD)
- [x] All existing tests pass (no regressions)

### DIFF
--- a/TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/PassThruHttpHeaders.java
+++ b/TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/PassThruHttpHeaders.java
@@ -24,7 +24,7 @@ public class PassThruHttpHeaders extends DefaultHttpHeaders {
             Stream.concat(
                 Stream.of(
                     HttpHeaderNames.CONTENT_LENGTH.toString(),
-                    HttpHeaderNames.CONTENT_TRANSFER_ENCODING.toString(),
+                    HttpHeaderNames.TRANSFER_ENCODING.toString(),
                     HttpHeaderNames.TRAILER.toString()
                 ),
                 Optional.ofNullable(extraHeaderNames).stream().flatMap(Arrays::stream)
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
@@ -514,15 +514,39 @@ public class CapturedTrafficToHttpTransactionAccumulator {
                 .log();
         } else if (observation.hasSegmentEnd()) {
             var rrPair = accum.getRrPair();
-            assert rrPair.requestData.hasInProgressSegment();
+            assert rrPair.requestData != null && rrPair.requestData.hasInProgressSegment() :
+                "SegmentEnd received but no in-progress request segment for "
+                + accum.trafficChannelKey + ". This indicates a missing ReadSegment "
+                + "or a malformed observation sequence in the capture stream.";
             rrPair.requestData.finalizeRequestSegments(timestamp);
         } else if (observation.hasRequestDropped()) {
             requestCounter.decrementAndGet();
             handleDroppedRequestForAccumulation(accum);
+        } else if (observation.hasWrite() || observation.hasWriteSegment()) {
+            return handleWriteDuringReadState(accum, observation, trafficStreamKey, timestamp);
         } else {
             return Optional.empty();
         }
         return Optional.of(CONNECTION_STATUS.ALIVE);
+    }
+
+    private Optional<CONNECTION_STATUS> handleWriteDuringReadState(
+        @NonNull Accumulation accum,
+        TrafficObservation observation,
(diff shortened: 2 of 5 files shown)


TrafficCapture/nettyWireLogging/src/main/java/org/opensearch/migrations/trafficcapture/netty/PassThruHttpHeaders.java
```diff
@@ -24,7 +24,7 @@ public HttpHeadersToPreserve(String... extraHeaderNames) {
             Stream.concat(
                 Stream.of(
                     HttpHeaderNames.CONTENT_LENGTH.toString(),
-                    HttpHeaderNames.CONTENT_TRANSFER_ENCODING.toString(),
+                    HttpHeaderNames.TRANSFER_ENCODING.toString(),
```
AndreKurait, line 27: This is the primary change that fixes the bug, everything else is defensive gates

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -514,17 +514,41 @@ private Optional<CONNECTION_STATUS> handleObservationForReadState(
...
+            + "for {}. This indicates the capture proxy emitted a premature EOM or the "
+            + "stream has malformed observation ordering. Transitioning to ACCUMULATING_WRITES.")
+            .addArgument(accum.trafficChannelKey)
+            .log();
+        if (accum.hasRrPair()) {
+            handleEndOfRequest(accum);
+        } else {
+            accum.state = Accumulation.State.ACCUMULATING_WRITES;
+        }
+        return handleObservationForWriteState(accum, observation, trafficStreamKey, timestamp);
```
gregschohn, line 549: This should fail loudly rather than try to recover (assertion, exception, metrics). The recovery path it implements may be worse than the silent drop it or hang that replaces.

There is a legitimate issue with 100 continue responses, but we should change capture to properly track/capture those.  That isn't relevant for this issue though, so we can punt on those.  This patch wouldn't make the 100 continue cases work any better either since the client upload is split.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -617,7 +644,16 @@ private boolean handleEndOfRequest(Accumulation accumulation) {
...
-        assert (!httpMessage.hasInProgressSegment());
+        assert (!httpMessage.hasInProgressSegment()) :
+            "Request has in-progress segment data at end-of-request for " + accumulation.trafficChannelKey
+            + ". This indicates a malformed capture stream missing a SegmentEnd.";
+        if (httpMessage.hasInProgressSegment()) {
+            log.atWarn().setMessage("Finalizing request with in-progress segment data for {}. "
+                + "This indicates a malformed capture stream missing a SegmentEnd.")
+                .addArgument(accumulation.trafficChannelKey)
+                .log();
+            httpMessage.finalizeRequestSegments(httpMessage.getLastPacketTimestamp());
```
gregschohn, line 655: I'm worried about continuing on with any request after a malformed connection.
I'd lean more on the side of converting the assert to an always-on exception.  I might feel that way about all of the assertions in this file

AndreKurait:
Followup in #2999 and #3000

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
AndreKurait: Thanks! I'm adding a change for 100 continue in #3000  and i'll make it fail loudly here

PR #2998, opened 2026-05-18
Generate migration resources and persist migration run history

gregschohn:
### Description

This builds off of [PR-2987](https://github.com/opensearch-project/opensearch-migrations/pull/2987)....  See [here for the changes since cronJobRfsChecks](https://github.com/gregschohn/opensearch-migrations/pull/17/changes)

This PR tightens the loop between the TS workflow schemas, the Kubernetes resources we install, and the user-facing parameters we can explain before a workflow runs. Instead of maintaining CRD and VAP YAML separately from the schema, migration resource specs and admission rules are now generated from shared schema/projection metadata. That gives us one place to define which fields exist on each migration resource, which changes are safe, which require approval, and which are impossible without deleting/recreating the resource. The same projection layer also powers `resolveMigrationResources`, so users can inspect the resolved resource parameters from a workflow config and dry-run the relevant policy behavior without waiting for a long-running workflow to hit admission failures.

A [design doc can be found within the files of this PR.](https://github.com/gregschohn/opensearch-migrations/blob/404455fad13b06b95bedf2d98672c187dad8417d/docs/resolvingMigrationParametersFromConfigs.md)

The PR also adds Kubernetes-native run history through immutable `MigrationRun` resources. Each workflow submit gets a millisecond-based run number, and the initializer writes a `MigrationRun` containing the resolved config that was submitted. The workflow then stamps its UID back onto that record from inside the workflow startup path, so run history remains tied to the eventual Argo Workflow without relying on the caller’s shell after `kubectl create`. Approval semantics move to that run number as well via the `approved-during-run` annotation, reducing the coupling to transient Workflow UIDs while still preserving a durable correlation path for debugging and audit.

On the installation side, generated CRDs and VAPs are staged into the migration-console image and applied by the Helm workflow install job, replacing static chart templates that could drift from the TS source. The rendered workflow outputs and docs have been refreshed around the new run-number metadata, generated resource installation, resolved parameter inspection, and removal of the separate resolved-config artifact step.

### Issues Resolved
This will track changes and answer questions about the relationship between the workflow configuration and the deployed resources, for all revisions of each of those resources.

### Check List


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/values.yaml
```diff
@@ -238,14 +238,26 @@ charts:
     values:
       crds:
         enabled: true
+      replicaCount: 2
```
AndreKurait, line 241: Ideally we'll have a taint here so they all don't land on the same node

AndreKurait (approved):

PR #3010, opened 2026-05-23
EKS CDC pipeline reliability fixes

AndreKurait:
## Summary

Four independent reliability fixes for the EKS CDC integ-test pipelines. Each fix corresponds to a specific CI failure mode and stands alone — they are bundled here only because they share a context (CDC pipeline runs).

### `wait-for-cluster` Argo template silent success

The template's first if/else block ran `kubectl wait --timeout=10s 2>/dev/null` as a probe and printed "cluster is ready!" regardless of whether the wait succeeded. The script lacked `set -e`, so even when kubectl had timed out the success message landed and the workflow proceeded. Downstream steps then failed against a not-yet-ready cluster with confusing connection errors.

Fix: add `set -e`, drop the buggy probe block. The second block (existence check + `kubectl wait` / `kubectl rollout status` with a real 200s timeout) was already correct.

### Pod-readiness wait observability

`wait_for_pod_ready` issued a single `watch.stream(timeout_seconds=3600)` with DEBUG-level event logs, so a stuck wait left a 60-minute silent gap in the pytest log and the eventual `TimeoutError` carried no information about *why* the pod never became Ready (was it never created? ImagePullBackOff? FailedScheduling?).

Fix: switch to a 30s polling loop. Each iteration logs an INFO heartbeat naming the matching pods and their phases, so a stuck wait surfaces inside the next 30s. On timeout, dump `kubectl describe pod -l <selector>` and `kubectl get events --field-selector=type!=Normal` so the cause is visible alongside the `TimeoutError`.

### Helm chart pull retry + Strimzi RBAC for `workflow reset` cleanup

Two CI failures previously caused by the same kind of transient infrastructure blip:

* `mirrorToEcr.sh` ran each `helm pull` once with stderr discarded. A single transient failure (DNS, github.io 5xx, OCI registry blip) under `set -euo pipefail` took down the entire bootstrap with no diagnostic. Image pulls already retry 5x; chart pulls now do the same, capturing stderr from the last attempt and surfacing it on final failure.
* `workflow reset` (run between integ-test cases under SA `migration-console-access-role`) walks the Strimzi resource tree to clean up. The role had no `kafka.strimzi.io` / `core.strimzi.io` rules at all, so the cleanup 403'd mid-walk and left residual broker pods that deadlocked the next test's setup. Add `delete` + `deletecollection` on `kafkas`, `kafkausers`, `kafkatopics`, `kafkanodepools`, and `core.strimzi.io/strimzipodsets` — the full set `reset.py` iterates.

Also reformats the long `migrations.opensearch.org` `resources:` list into one-per-line so a missing `/status` companion is obvious at a glance.

### cert-manager survives single-node loss

The chart already sets `replicaCount: 2` and a `minAvailable: 1` PDB on the cert-manager controller, cainjector, and webhook, but with no anti-affinity both replicas can land on one node. A single node loss then drains the Service to zero endpoints and Argo `provisionCert` dies with `failed calling webhook "webhook.cert-manager.io": context deadline exceeded`.

Fix: add soft `preferredDuringSchedulingIgnoredDuringExecution` pod anti-affinity keyed on `kubernetes.io/hostname` to each of the three deployments. Soft (not required) so single-node clusters still schedule both pods.

## Test plan

- [ ] EKS CDC pipeline runs the integ-test matrix end-to-end without the 60-minute silent hangs.
- [ ] `workflow reset` between test cases cleans up Strimzi CRs without 403s.
- [ ] `mirrorToEcr.sh` survives a transient github.io / OCI registry blip during chart pulls.
- [ ] cert-manager controller/cainjector/webhook survive a node drain without webhook calls timing out on dependent Argo steps.


gregschohn (commented):
These changes LGTM.  I've merged mainline & some other independent changes that I've made too.  I was planning to use this branch for  integ test hardening.  We can approve and merge once all the tests pass

gregschohn (approved):

PR #3013, opened 2026-05-23
Recover from inline rebalance: drop+seek touched partitions, cancel channel sessions on partition round-trip

AndreKurait:
## What this PR fixes

Under cooperative-sticky rebalancing, a Kafka consumer that misses `max.poll.interval.ms` (slow target, GC pause, or shutdown) gets fenced. On the next `poll()` Kafka runs a rebalance inline: `onPartitionsRevoked` then `onPartitionsAssigned` for the same partitions ("round-trip"). The broker resets the fetch position to the last committed offset on every reassignment — round-trip or not — so any records past the last commit re-deliver under a new generation.

The original code papered over this with a deferred `pendingCleanupPartitions` set that filtered out round-trip partitions as "not really lost", and the accumulator caught the fallout with a defensive backstop that closed the connection via `CLOSED_PREMATURELY`. That close path leaves `ConnectionReplaySession.cancelled = false`, so the cached Netty channel in `ClientConnectionPool.connectionId2ChannelCache` could self-heal onto the re-delivered request stream — half-finished old request and new request multiplexed on the same target socket.

## Mental model

A partition revocation is always treated as a loss event. The recovery is shaped by four invariants:

1. **Always synthesize an interrupted-close at the OLD generation.** Fired from a shared `cleanupRevokedPartitions(parts, attemptCommit)` helper invoked by both `onPartitionsRevoked` (with `attemptCommit=true`) and `onPartitionsLost` (with `false`). The truly-lost callback runs **outside** `commitDataLock` and at the OLD generation, so synthesized close session keys reference the OLD generation that matches `session.generation` on channels opened during that gen.

2. **Inline-rebalance recovery is a uniform pre-poll snapshot + post-poll min-rewind.** Before every `poll()`, `TrackingKafkaConsumer.safePollWithSwallowedRuntimeExceptions` clears a `rebalanceDuringPoll` flag and snapshots `kafkaConsumer.position(tp)` for every assigned partition (both local — no broker RPC). If the rebalance callbacks fired inline they set the flag. After `poll()`, on a set flag, the entire returned batch is dropped and every CURRENTLY-assigned partition (skipping ones not in the pre-poll snapshot — newly assigned mid-poll) is seeked to `min(prePollPosition, lowestOffsetReturnedForThisPartition)`:
   - **Touched partition** (broker reset to last committed): lowest returned == reset offset, ≤ pre-poll → seek to reset → records past the last commit re-deliver under the new gen.
   - **Untouched partition**: lowest returned == pre-poll → seek to pre-poll → only the polled-and-dropped records re-deliver, NOT thousands of in-flight records past last commit.
   - **No records this poll**: not in min-returned map → skipped.
   - **Newly assigned mid-poll**: not in pre-poll snapshot → skipped (left wherever rebalance set it).
   - **Revoked outright**: not in current assignment → not iterated (`position`/`seek` would throw).

   Post-poll `position(tp)` is *not* usable as the current reference here because it has already advanced past records this poll returned, hiding the broker's reset value for touched partitions. The lowest returned offset is the un-advanced reference frame.

3. **A returned chunk never mixes dropped-and-rewound records with live records.** The recovery branch returns an empty `ConsumerRecords`, so the caller sees either nothing (followed by re-delivery on a later poll) or a fully live batch. Synth-close batches enqueued by `cleanupRevokedPartitions` flow through `KafkaTrafficCaptureSource.readNextTrafficStreamSynchronously`'s pre-poll drain on the next call, before any new-generation records.

4. **Use `TRAFFIC_SOURCE_READER_INTERRUPTED` for any close caused by partition reassignment.** It routes through `replayEngine.cancelConnection()`, which sets `cancelled = true` on the session and prevents cached-channel reuse. A shared `closeAsTrafficSourceReaderInterruptedAndRemove(existingAccum, tsk)` helper covers both the synth-close branch and the accumulator's defensive backstop, so the two cannot drift in close semantics. `commitTrafficStreams(status, …)` suppresses the broker commit for both `CLOSED_PREMATURELY` (partial replay) and `TRAFFIC_SOURCE_READER_INTERRUPTED` (broker will re-deliver past the last commit).

## Changes

**`TrackingKafkaConsumer`**
- Single private `cleanupRevokedPartitions(parts, attemptCommit)` shared by `onPartitionsRevoked` and `onPartitionsLost`. Fires `onPartitionsTrulyLostCallback` outside `commitDataLock` at the OLD generation.
- New `rebalanceDuringPoll` `AtomicBoolean` + per-poll `prePollPositions` snapshot. `safePollWithSwallowedRuntimeExceptions` recovers via `recoverFromInlineRebalance` which seeks every assigned partition to `min(pre-poll, min-returned-offset)` and returns an empty batch.
- `onPartitionsAssigned` is just generation-bump + `OffsetLifecycleTracker` install (`computeIfAbsent` guards against re-stamping). No deferred drain.

**`KafkaTrafficCaptureSource`**
- `readNextTrafficStreamSynchronously` is three phases: pre-poll drain one synth-close batch, park-and-empty if `outstandingTrafficSourceReaderInterruptedCloseSessions > 0`, else poll. No post-poll prepend, no all-synth-or-all-real assertion.
- `CooperativeStickyAssignor` configured by default in `buildKafkaProperties`.
- `DEFAULT_POLL_INTERVAL_MS` raised from `60000` to `300000` to match kafka-clients 4.x library default. Touch cadence stays at 30s via `DEFAULT_KEEP_ALIVE_PERIOD` (decoupled from the fence threshold). Operators can still override via `--kafkaPropertyFile`; subclasses via `protected static defaultPollIntervalMs()`.
- `PENDING_CLOSE_SESSION_NUMBER_PLACEHOLDER` (= 0) is the named constant shared between the synth-close registration site and `TrafficReplayerTopLevel`'s close-callback lookup, so they cannot drift before the `GenerationalSessionKey` upgrade lands.

**`CapturedTrafficToHttpTransactionAccumulator`**
- Shared `closeAsTrafficSourceReaderInterruptedAndRemove(existingAccum, tsk)` helper used by both the synth-close branch and the stale-generation defensive backstop.
- Defensive backstop closes with `TRAFFIC_SOURCE_READER_INTERRUPTED` (was `CLOSED_PREMATURELY`), restoring ERROR-level logging since correct source-layer behavior should never trigger it.

**`TrafficReplayerCore`**
- `commitTrafficStreams(status, …)` suppresses commits for both `CLOSED_PREMATURELY` and `TRAFFIC_SOURCE_READER_INTERRUPTED`.
- `onConnectionClose` for `TRAFFIC_SOURCE_READER_INTERRUPTED` short-circuits with `commitTrafficStreams(false, …)` then `replayEngine.cancelConnection()` — bypasses the sorter and time-shifting so the channel close runs synchronously, before any new-generation records flow.

## Tests

- New `StaleAccumulationCancelOnRejoinTest` (in-memory `MockConsumer`) and `StaleAccumulationCancelOnRejoinKafkaTest` (real Kafka container) reproduce the production trace and verify the synth-close and re-delivered record arrive in the correct order across chunks.
- Existing `KafkaKeepAliveTests.testBlockedReadsAndBrokenCommitsDontCauseReordering` (real Kafka) covers heartbeat-loss + multi-cycle fence + commit-after-rejoin.
- `TrackingKafkaConsumerTest` asserts the truly-lost callback fires at the OLD generation directly from `onPartitionsRevoked`.
- `PartitionRevocationStaleStateTest` covers the accumulator's stale-generation defensive backstop and channel-context replacement on gen bump.
- All trafficReplayer unit tests pass; `isolatedTest` (containerized) `KafkaKeepAliveTests` and `StaleAccumulationCancelOnRejoinKafkaTest` pass.

## Doc

`docs/replayerArchitecture.md` is updated to match the merged code: shared `cleanupRevokedPartitions` helper, `rebalanceDuringPoll` flag, pre-poll-snapshot + min-rewind recovery, the no-mix invariant, the `closeAsTrafficSourceReaderInterruptedAndRemove` shared close path, and the dual-status commit suppression.


docs/replayerArchitecture.md
```diff
@@ -420,7 +432,10 @@ called from `kafkaExecutor`.
 
 If records are in-flight but no new records are being read (backpressure is blocking the read
 loop), the consumer must still call `poll()` periodically to maintain its group membership
-(heartbeat). The `keepAliveInterval` is set to half of `max.poll.interval.ms`.
+(heartbeat). The `keepAliveInterval` is pinned to `DEFAULT_KEEP_ALIVE_PERIOD = 30s` in
+`KafkaTrafficCaptureSource`. It is intentionally decoupled from `max.poll.interval.ms` (the
+broker-enforced fence threshold, which inherits the kafka-clients library default of 5 minutes)
+so the touch cadence stays tight even when the fence threshold is lenient.
```
gregschohn, line 438: What would happen if the poll interval was set to something smaller like 20s?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -224,30 +255,17 @@ public static KafkaTrafficCaptureSource buildKafkaSource(
         @NonNull KafkaBehavioralPolicy behavioralPolicy
     ) throws IOException {
         var kafkaProps = buildKafkaProperties(brokers, groupId, authType, kafkaUserName, kafkaPassword, propertyFilePath);
-        kafkaProps.putIfAbsent(MAX_POLL_INTERVAL_KEY, DEFAULT_POLL_INTERVAL_MS);
-        var pollPeriod = Duration.ofMillis(Long.valueOf((String) kafkaProps.get(MAX_POLL_INTERVAL_KEY)));
-        var keepAlivePeriod = getKeepAlivePeriodFromPollPeriod(pollPeriod);
+        kafkaProps.putIfAbsent(MAX_POLL_INTERVAL_KEY, defaultPollIntervalMs());
         return new KafkaTrafficCaptureSource(
```
gregschohn, line 259: We should still tie this in programmatically, but this won't really be an issue until we're running multiple replayers in parallel AND are trying to keep tighter time collars.

gregschohn (commented):
.

gregschohn (approved):

docs/replayerArchitecture.md
AndreKurait: This would cause a problem, at the moment we're not parsing this poll interval in our classes and if that's needed i'd like to leave it in a future commit

PR #3014, opened 2026-05-23
feat(replayer): Add JsonJSPredicateProvider for JS-based request filtering

niravpi:
## Description

Adds a JavaScript-based predicate provider that enables request filtering via JS scripts. This complements the existing `JsonJMESPathPredicateProvider` by providing full JS power for complex filtering logic.

## Changes

- **`JsonJSPredicateProvider.java`** — Implements `IJsonPredicateProvider`, loads JS via file/resource/inline (same config keys as `JsonJSTransformerProvider`)
- **ServiceLoader registration** — `META-INF/services/org.opensearch.migrations.transform.IJsonPredicateProvider`
- **12 unit tests** — URI filtering, regex, method checks, file loading, error handling

## Usage

```bash
--requestFilterConfig '{"JsonJSPredicateProvider":{"initializationScriptFile":"/path/to/filter.js"}}'
```

Example `filter.js`:
```javascript
(function(bindings) {
  return function(msg) {
    return /\/(select|update|get)/.test(msg.get('URI'));
  };
})
```

## Motivation

The existing `JsonJMESPathPredicateProvider` only supports inline JMESPath expressions (no file loading, no regex, limited string operations). This provider enables:
- Regex-based path matching
- HTTP method filtering
- Header/body inspection
- File-based filter scripts (consistent with `--transformerConfig` pattern)

## Testing

- 12 unit tests (all passing)
- End-to-end validated with Traffic Replayer: filter correctly skips admin/schema/config requests while allowing select/update/get through to transformation

### DIFF
--- a/transformation/transformationPlugins/jsonMessageTransformers/jsonJSTransformerProvider/src/main/java/org/opensearch/migrations/transform/JsonJSPredicateProvider.java
+++ b/transformation/transformationPlugins/jsonMessageTransformers/jsonJSTransformerProvider/src/main/java/org/opensearch/migrations/transform/JsonJSPredicateProvider.java
@@ -1,0 +1,82 @@
+package org.opensearch.migrations.transform;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+/**
+ * JS-based request filter predicate provider. Loads a JavaScript file that returns
+ * a function evaluating to true (allow) or false (skip).
+ *
+ * <p>Supports the same config keys as {@link JsonJSTransformerProvider}:
+ * {@code initializationScriptFile}, {@code initializationResourcePath}, or
+ * {@code initializationScript}.
+ *
+ * <p>Config example:
+ * <pre>{@code {"JsonJSPredicateProvider":{"initializationScriptFile":"/path/to/filter.js"}}}</pre>
+ *
+ * <p>The JS must follow the standard closure pattern and return a boolean:
+ * <pre>{@code
+ * (function(bindings) {
+ *   return function(msg) {
+ *     var uri = msg.get('URI');
+ *     return uri.indexOf('/select') >= 0;
+ *   };
+ * })
+ * }</pre>
+ */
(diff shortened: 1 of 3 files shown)


transformation/transformationPlugins/jsonMessageTransformers/jsonJSTransformerProvider/src/main/java/org/opensearch/migrations/transform/JsonJSPredicateProvider.java
```diff
@@ -0,0 +1,82 @@
...
+ * <pre>{@code
+ * (function(bindings) {
+ *   return function(msg) {
+ *     var uri = msg.get('URI');
+ *     return uri.indexOf('/select') >= 0;
+ *   };
+ * })
+ * }</pre>
+ */
+public class JsonJSPredicateProvider implements IJsonPredicateProvider {
```
AndreKurait, line 31: Food for thought, could we adapt the existing request interface to allow skipping the request? For backfill we do this by returning an array of transformed items that can be 0 in length (in the future also split requests)?  This would allow the same function in the transformation interface to have rules for how to transform the request including skipping.

niravpi: Thanks for the suggestion Andre. Explored this approach as well, transform returning `null`/empty to signal skip and validated it end-to-end. It works, but I am leaning towards to keep the separate filter for the long term replay use case due to following reasons:

   - **Debuggability:** Clear distinction between "intentionally filtered" (predicate returned false) vs "transform bug returned null" in metrics/logs
   - **Performance:** Filtered requests never enter GraalVM, skips loading the full transform engine for filtered request traffic
   - **Operational independence:** Filter config can change without touching transform (e.g., temporarily allow `/admin` for debugging without rebuilding JS)

 This PR (`JsonJSPredicateProvider`) adds JS file-based filtering for more complex cases (regex, body inspection, method+URI combos) in a long run. Let me know if you think otherwise and if I might be missing any usecase in here.

AndreKurait:
Closing to focus on replayer direct transformations that can handle this dual behavior

PR #3015, opened 2026-05-24
Make the MA helm installation fail when its dependencies fail

gregschohn:
The dependency installer could previously hide failures from child chart installs. The installer launched chart installs in the background and then aggregated chart notes, but a failed inner `helm upgrade --install` could still leave notes behind while the outer MA release was broken or misleading. Track each background install PID explicitly, record per-chart failure markers, wait for every child install, write whatever notes were collected, and exit nonzero if any child chart failed.

Rename the pre-install/pre-upgrade hook Job from `helm-installer` to `dependency-installer` so Helm's own failure message gives users a better clue about what failed. Helm still reports hook Job failures tersely, but `ma-dependency-installer` is more actionable than `ma-helm-installer`.

Derive dependency wait time from the same retry policy used by chart installs. This keeps dependency checks aligned with the actual install retry budget: chart timeout * retry attempts + retry sleep + a small buffer. Also make failed dependencies unblock dependents immediately instead of waiting for the full timeout.

Make hook Job failures propagate to Helm across the rest of the chart. Add `podFailurePolicy` and `activeDeadlineSeconds` to the resource-install hook Jobs that create CRDs, admission policies, issuers, S3 buckets, and monitoring resources. The policy intentionally ignores `DisruptionTarget` pod failures so node disruption can be retried without burning the Job backoff budget, but fails the Job immediately when the hook container exits nonzero.

Switch affected hook Jobs from `restartPolicy: OnFailure` to `restartPolicy: Never` because Kubernetes requires `Never` when `podFailurePolicy` is used. This also makes script failures clearer: a nonzero container exit becomes a failed pod, the Job failure policy marks the Job failed, and Helm fails the release. We keep nonzero `backoffLimit` values for general reliability cases where the pod dies for reasons not covered by the explicit failure policy.

Tighten hook scripts so errors are not swallowed. Use `set -e` where missing, avoid pipelines that mask intermediate failures, verify generated CRDs and ValidatingAdmissionPolicy resources after applying them, and make certificate issuer setup fail if the expected resources never become ready.

Enable `waitForInstallation` for production/EKS dependency charts where we want the install to validate that the installed components can actually start. Since dependency chart installs still run in parallel, this should be bounded mostly by the slowest dependency rather than the sum of every dependency install.

Rename `installWorkflows.yaml` to `migrationsResources.yaml` because that hook now installs more than Argo WorkflowTemplates. Rename the install hook Job to `install-migrations-resources`; leave the delete hook named `remove-argo-migration-templates` because its cleanup still only removes Argo WorkflowTemplate resources.

### Check List


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/valuesEks.yaml
```diff
@@ -140,13 +140,18 @@ otelConfiguration:
           exporters: [awsemf/cadvisor, prometheus]
 
 charts:
+  aws-privateca-issuer:
+    waitForInstallation: true
+
   argo-workflows:
+    waitForInstallation: true
```
AndreKurait: should we have any more waitForInstallation in the minikube values? like i think we would need to wait on argo-workflows for actually installing the workflow templates.

AndreKurait (approved):
I'd like to wait for the EKS tests to pass before we merge, but otherwise i'm good

PR #3017, opened 2026-05-25
Fix bootstrap script help text to list all 7 VPC endpoint types

sumobrian:
## Summary
- The `--create-vpc-endpoints` help text and unknown-type warning only listed 5 of the 7 valid endpoint types (`s3,ecr,ecrDocker,cloudwatchLogs,efs`)
- The `all` expansion already includes `sts` and `eksAuth`, but these were missing from the user-facing messages
- Updated both the `--help` output and the warning message to list all 7 types: `s3,ecr,ecrDocker,cloudwatchLogs,efs,sts,eksAuth`

## Test plan
- [ ] Run `./aws-bootstrap.sh --help` and verify the `--create-vpc-endpoints` description lists all 7 types
- [ ] Pass an invalid endpoint type and verify the warning lists all 7 valid types

### DIFF
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -133,7 +133,7 @@ while [[ $# -gt 0 ]]; do
       echo "                                            (required with --deploy-import-vpc-cfn)."
       echo "  --create-vpc-endpoints [list]             Create VPC endpoints for private subnet connectivity."
       echo "                                            Only valid with --deploy-import-vpc-cfn."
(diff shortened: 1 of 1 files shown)


gregschohn:
aws-bootstrap is being deprecated in favor or the rust installer.
We don't currently have a great way to install the VPC Endpoints that are required to operate in an otherwise isolated network right now.

I'm purposefully keeping this open because 1) we need that feature - at least for consistent testing and 2) we should make sure that the spirit - to show all vpc endpoings - of this PR makes it into the installer.

PR #3020, opened 2026-05-26
feat(replayer): Add file-based config flags for request filter and response post-processor

niravpi:
## Description

   Adds file-based and base64-encoded config flags for request filter and response post-processor, following the same `TransformerParams` pattern as the existing `--transformer-config-file`.

   ## Changes

   `TrafficReplayer.java` — replaces simple `@Parameter` fields with proper `TransformerParams` implementations:

   - `RequestFilterParams` — supports `--request-filter-config`, `--request-filter-config-file`, `--request-filter-config-encoded`
   - `ResponsePostProcessorParams` — supports `--response-post-processor-config`, `--response-post-processor-config-file`, `--response-post-processor-config-encoded`

   Both use `TransformerConfigUtils.getTransformerConfig()` for resolution — same utility, same validation (error if multiple sources specified), same precedence.

   ## Motivation

   EKS/ECS deployments pass configs via file paths or base64-encoded parameters. The existing inline-only flags (`--request-filter-config`, `--response-post-processor-config`) don't support this. Adding file and encoded variants enables the same deployment patterns used by `--transformer-config-file` and `--transformer-config-encoded`.

   ## Testing

   - All existing replayer tests pass (no regression)

   ## Backward Compatibility

   Fully backward compatible. The existing `--request-filter-config` and `--response-post-processor-config` inline flags continue to work unchanged. The new file and encoded variants are additive.

### DIFF
--- a/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
+++ b/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/TrafficReplayer.java
@@ -191,22 +191,11 @@ public class TrafficReplayer {
         @ParametersDelegate
         private TupleTransformationParams tupleTransformationParams = new TupleTransformationParams();
 
-        @Parameter(
-            required = false,
-            names = {"--request-filter-config", "--requestFilterConfig"},
-            arity = 1,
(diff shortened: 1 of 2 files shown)


AndreKurait:
We're going a different direction with replayer transforms being able to skip themselves so we don't need this additional interface

PR #3022, opened 2026-05-27
Add migration-assistant CLI (preview) alongside aws-bootstrap.sh; bake via Jenkins

AndreKurait:
## What this PR does

This PR introduces a new bash-native CLI for OpenSearch Migration
Assistant deploys (the TUI in `deployment/k8s/aws/cli/`), packaged
alongside the existing `aws-bootstrap.sh`. Both ship as release
artifacts; `aws-bootstrap.sh` remains the **production** deploy path.

> ⚠️ **The CLI is PREVIEW.** It runs through Jenkins by default (see
> "Bake mechanism" below) so we accumulate stability coverage on real
> CI runs before promoting it to the production artifact.

## Bake mechanism

Every EKS Jenkins pipeline now has a `USE_RELEASE_BOOTSTRAP` boolean
param. The default routes:

| `USE_RELEASE_BOOTSTRAP` | Deploy backend |
| --- | --- |
| `false` (default for source builds) | New TUI CLI from the source checkout — `./deployment/k8s/aws/cli/bin/migration-assistant` |
| `true` | Legacy `aws-bootstrap.sh` downloaded from the GitHub release |

This means **every PR check / scheduled cron currently exercises the
TUI** as the deploy script. Operators who want the stable path flip
`USE_RELEASE_BOOTSTRAP=true` and run aws-bootstrap.sh exactly as today.

`vars/resolveBootstrap.groovy` returns the same `{script, flags}` map
in both branches, so `bootstrapMA.groovy` doesn't care which backend
is running.

## Layout

```
deployment/k8s/aws/
├── aws-bootstrap.sh          # PRODUCTION (unchanged from main)
├── assemble-bootstrap.sh     # PRODUCTION (unchanged from main)
└── cli/                      # PREVIEW
    ├── bin/migration-assistant
    ├── lib/                  # 18 shell modules — std, term, manifest,
    │                         # ui, log, state, discover, cfn, helm,
    │                         # crane, agent, pack, …
    ├── install.sh            # curl-pipe installer for the CLI tarball
    └── test/                 # bats suite (326 tests)

agent-skills/                 # Renamed from kiro-cli/, hoists agent-
                              # agnostic skills out from kiro-specific
                              # config; shared by claude/codex/kiro.
```

Release assets:
- `aws-bootstrap.sh` (production)
- `install.sh` + `migration-assistant-cli-<ver>.tar.gz` (preview CLI)
- `agent-skills.tar.gz` (the shared skill bundle)

## What's in the preview CLI

The CLI ships an extensible framework so partner distributions can
build custom variants without forking:

- **`manifest.json`** at the tarball root — declares branding, mode
  list, and MCP servers. Schema v1.
- **Skill auto-discovery** — every `skills/<name>/SKILL.md` in the
  bundle is auto-installed at agent handoff. Drop in a skill dir,
  it ships.
- **`migration-assistant pack`** subcommand — repacks an upstream
  release tarball with `--add-skill <dir>`, `--add-mcp <json>`,
  `--branding <json>`, `--brand-name`, `--brand-binary`, etc.
  Produces a new tarball with a `build.packs[]` audit entry.
- **Per-stage state** lives in `./migration-assistant-workspace/<stage>/`.
- **Mode picker**: `Manual` only by default. `Agent` (LLM-driven
  deploy via claude/codex/kiro) is a preview-within-a-preview gated
  behind `MIGRATE_ENABLE_AGENT=1`. Without it the mode picker hides
  the option and `--mode Agent` errors out with a hint.

## Trying the CLI yourself

```sh
# Install:
curl -fsSL https://github.com/opensearch-project/opensearch-migrations/releases/latest/download/install.sh | bash

# Use:
cd ~/my-migration-project
migration-assistant                 # default: Manual mode
MIGRATE_ENABLE_AGENT=1 migration-assistant   # to see / pick AI mode
migration-assistant cleanup
```

Or run from the source checkout:
```sh
./deployment/k8s/aws/cli/bin/migration-assistant
```

## Notes for reviewers

- `aws-bootstrap.sh` and `assemble-bootstrap.sh` are byte-identical to
  main HEAD. The diff against main for those files is **zero**.
- Only `vars/resolveBootstrap.groovy` differs from main meaningfully:
  the source-checkout branch points at the new CLI instead of building
  `dist/aws-bootstrap.sh` from source. That's the bake mechanism.
- Test count: 326 bats tests across the CLI, all passing on Ubuntu
  22.04 (matches GHA) + macOS. shellcheck + bash-3.2 lint clean.
- The agent fresh-handoff prompt is configurable per-pack via
  `branding.agentFreshPrompt` in manifest.json.

## Why this got reshuffled

I originally pushed harder — replacing aws-bootstrap.sh outright. After
review feedback, decided the new code needs a few releases of bake time
before it earns the production deploy slot. This commit (`d33a042a7`)
restores aws-bootstrap.sh as the canonical production path; the CLI
ships as a preview artifact and runs through Jenkins by default to
gather coverage. When the team is satisfied, a follow-up PR can flip
`USE_RELEASE_BOOTSTRAP`'s default or remove aws-bootstrap.sh entirely.

## Behind a feature flag

Even within the preview CLI, anything riskier is gated:

| Feature | Gate | Purpose |
| --- | --- | --- |
| The CLI itself | `USE_RELEASE_BOOTSTRAP=false` (Jenkins default) | Bake before promote |
| AI/Agent mode | `MIGRATE_ENABLE_AGENT=1` (env var) | UX is preview within preview |
| Per-MCP skip | `MIGRATE_SKIP_MCP=1` / `MIGRATE_SKIP_MCP_<NAME>=1` | Operator escape hatch |


deployment/k8s/aws/README.md
gregschohn, line 1: The README needs an overhaul.  There are still some references to aws-bootstrap.sh.  The first sentence is "This guide is for **developers** that are interested in building everything..." but the documentation seems like it's user facing, etc.

.github/workflows/workflow-schema.yaml
gregschohn, line 1: let's rename this file

.github/workflows/generate-workflow-schema.yaml
```diff
@@ -7,6 +7,9 @@ on:
       - main
     paths:
       - 'deployment/k8s/**'
+      - '!deployment/k8s/aws/cli/**'
+      - '!deployment/k8s/aws/aws-bootstrap.sh'
```
gregschohn: why is this still here?

deployment/k8s/aws/cli/bin/migration-assistant
```diff
@@ -0,0 +1,79 @@
...
+# shellcheck source=lib/console.sh
+source "$LIB_DIR/console.sh"
+# shellcheck source=lib/agent.sh
+source "$LIB_DIR/agent.sh"
+# shellcheck source=lib/cleanup.sh
+source "$LIB_DIR/cleanup.sh"
+# shellcheck source=lib/resume.sh
+source "$LIB_DIR/resume.sh"
+
+case "${1:-resume}" in
```
gregschohn, line 84: I'd like to be able to flip to the agent or be able to run the console with the command when MA was deployed - I'm not sure how to rig that though.

deployment/k8s/aws/cli/lib/agent.sh
gregschohn, line 1: This DID successfully start my codex agent with some instructions.  It took ~1 minute to get to the point where it realized that I didn't have credentials and then wanted to know if it should deploy to kind.  Overall, the experience was rough.  I think that this branch should be hidden for now until the agent gets merged - maybe this gets merged without this branch and the agent work includes activating this branch.

gregschohn: Another weird thing here that will be a big UX question - I run agent mode.  I stop (ctrl-C).  When I exit, codex, etal show me how to enter back into the session.  If I do that directly, I am RIGHT where I left off and it takes ~1s.  When I re-enter with the migration-assistant command, I have two prompts (probably a bug) and about 25s of re-reading the Startup.md file - and it seems like it doesn't know where I left off anyway (I think one time it might have known).

We should try to tap right back where they left off.  There are some various patterns to do resume with tracked session ids, but we could also just do last and figure that we're within a narrowly scoped directory

deployment/k8s/aws/cli/lib/artifacts.sh
```diff
@@ -0,0 +1,114 @@
...
+
+ART_REPO='opensearch-project/opensearch-migrations'
+ART_TAG_PREFIX=''   # opensearch-migrations tags as "3.2.1", not "v3.2.1"
+
+# artifacts_fetch <name> <version>  →  echoes the local symlink path.
+artifacts_fetch() {
+  local name="$1" version="$2"
+  local tag="${ART_TAG_PREFIX}${version}"
+  local primary="https://github.com/${ART_REPO}/releases/download/${tag}/${name}"
+  local fallback="https://raw.githubusercontent.com/${ART_REPO}/${tag}/${name}"
```
gregschohn, line 36: why do we need a fallback?

deployment/k8s/aws/cli/lib/cfn.sh
```diff
@@ -0,0 +1,536 @@
...
+  fi
+
+  if _cfn_stack_healthy "$stack_name" "$region"; then
+    ui_ok "CFN stack $stack_name already healthy; skipping deploy"
+    state_set CFN_STACK_NAME "$stack_name"
+    state_save
+    return 0
+  fi
+
+  ui_step "Downloading CFN template (MA v$ma_ver)"
```
gregschohn: why does this start w/ 'v'

deployment/k8s/aws/cli/lib/cfn.sh
```diff
@@ -0,0 +1,536 @@
...
+  local tmpl
+  tmpl=$(artifacts_fetch "$CFN_TEMPLATE_NAME" "$ma_ver")
```
gregschohn: Do any of these scripts work from a local repo or do I need to download them from the sources via artifacts.sh?

deployment/k8s/aws/aws-bootstrap.sh
gregschohn, line 1: There's a lot to like and a lot to be scared of (like the hundreds of lines of UI tracking code in cfn.sh).  I'd like to see us keep the aws-bootstrap.sh experience around a bit longer, just as a fallback in case we hit bugs in this installer.  After 2-4 weeks, we can deprecate the non-modular and non-interactive version.  Until then, it won't take much to maintain it and it could be a 'get-out-of-jail-free card'

Are you creating this file somewhere for the user?  I still see many references to it in comments, code and docs.

deployment/migration-assistant-solution/README.md
```diff
@@ -33,9 +33,11 @@ sequenceDiagram
 
 The EKS solution has a simpler initial configuration surface.  A user deploys an
 EKS cluster via one of the vended CloudFormation templates.  Once deployed, the
-user runs the aws-bootstrap.sh script, which can be downloaded directly from
-github and run as-is.  Both of these commands can be run from an AWS CloudShell 
-or from anywhere else with credentials and network access.  
+user runs the migration-assistant CLI (legacy entry-point: `aws-bootstrap.sh`,
```
gregschohn, line 36: not sure why you're mentioning a dead file... are you just trying to tell a user what it was?  If that's the case, I'd be more explicit.  I also don't think that we should kill it off yet.

deployment/k8s/aws/cli/install.sh
gregschohn, line 1: I need a readme to explain what aws-bootstrap is for

deployment/k8s/aws/cli/Makefile
gregschohn, line 1: Can we do this with gradle?
```
plugins {
    id 'base'
}

def cliVersion = "0.1.0"
def shellFiles = fileTree(dir: projectDir).matching {
    include 'bin/migration-assistant'
    include 'install.sh'
    include 'lib/*.sh'
    include 'test/helpers/*.sh'
}

tasks.register('lint', Exec) {
    description = 'Run shellcheck on shell scripts'
    commandLine 'shellcheck', '-x', '-P', 'lib'
    args shellFiles.files
}

tasks.register('testBats', Exec) {
    description = 'Run bats-core tests'
    commandLine 'bats', 'test/'
}

tasks.register('releaseTar', Tar) {
    dependsOn lint, testBats
    archiveBaseName.set('migrate-cli')
    archiveVersion.set(cliVersion)
    archiveExtension.set('tar.gz')
    destinationDirectory.set(file("$buildDir/dist"))
    
    // Gradle handles the archiving natively, no mktemp or bash tar needed
    from('bin') { into 'bin' }
    from('lib') { into 'lib' }
    from('skills') { into 'skills' }
    from('install.sh', 'README.md', 'LICENSE')
}

build.dependsOn releaseTar
```

I'm a recovering make user.  You could have portability problems (now users need make installed too, which might include a LOT of other baggage with it).  It also tracks dependencies via timestamps, which might get screwy with branch switches.  It is a lot weaker than content hashes and something else that we need to figure out.

gregschohn:
Here's feedback from an automated codex review (I'm doing my own separately).  I don't understand some of the points, the others do seem like significant issues.

The rewrite has several blocking compatibility regressions in core deployment paths: stage selection, CFN mode/skip behavior, image mirroring, kubeconfig handling, and cleanup/install flows. These break documented and CI invocations, so the patch should not be considered correct as-is.

Full review comments:

- [P1] Preserve --stage as the deployment stage — ./dev/m5/deployment/k8s/aws/cli/lib/resume.sh:56-56
  When callers pass `--stage prod` and run non-interactively or accept defaults, the first pass only moves the state directory and the second pass discards the flag without setting `STAGE_NAME`. Downstream `wizard_collect`, CFN, Helm, and console code then default to `ma`, so documented/legacy aws-bootstrap invocations deploy to the wrong stage/namespace.

- [P1] Honor the selected CFN deployment mode — ./dev/m5/deployment/k8s/aws/cli/lib/cfn.sh:11-11
  For `--deploy-import-vpc-cfn --vpc-id ... --subnet-ids ...`, the parser records import-VPC state, but this module always downloads the Create-VPC template and the deploy command only passes `Stage`. That documented existing-VPC path will create/use the wrong template or fail instead of importing the supplied VPC/subnets.

- [P1] Gate CFN deployment on the deploy flags — ./dev/m5/deployment/k8s/aws/cli/lib/resume.sh:208-208
  When no `--deploy-*-cfn` flag is supplied—the old bootstrap-only path and the updated `awsRunEksValidation.sh` invocation—manual mode still calls `cfn_deploy_or_skip`. Unless the user explicitly adds `--skip-cfn-deploy`, the CLI probes/deploys a default Create-VPC stack instead of discovering the already verified stack by stage, so existing-stack runs can modify the wrong CFN stack.

- [P1] Stop reading image list from the new shim — ./dev/m5/deployment/k8s/aws/cli/lib/crane.sh:45-45
  When `MIRROR_IMAGES` remains the default `Y` for any release produced by this PR, `artifacts_fetch aws-bootstrap.sh` returns the generated shim from `assemble-bootstrap.sh`, not the deleted monolithic script with `IMAGES="..."`. `_extract_images` then returns empty and default private-ECR installs abort before mirroring.

- [P1] Keep kubeconfig creation when only skipping context switch — ./dev/m5/deployment/k8s/aws/cli/lib/helm.sh:43-47
  For CI/fresh hosts using `--skip-setting-k8s-context`, this branch skips `aws eks update-kubeconfig` entirely. The old flag only avoided changing the current context after creating the kubeconfig entry, so subsequent Helm/Kubectl commands here have no usable cluster context unless it already existed.
Yes. I’d add these, strongest first:

- **[P1] Restore required Helm values from CFN exports** — [helm.sh](./dev/m5/deployment/k8s/aws/cli/lib/helm.sh:133)
  The legacy script always passed `stageName`, `aws.region`, `aws.account`, and `defaultBucketConfiguration.snapshotRoleArn` to Helm. The new path only passes region/account and image flags, and `_write_helm_values` writes an empty placeholder. That drops `SNAPSHOT_ROLE` and chart stage wiring, so installs can come up with wrong defaults or broken snapshot permissions.

- **[P1] Preserve local build/source mode** — [resume.sh](./dev/m5/deployment/k8s/aws/cli/lib/resume.sh:77), [cfn.sh](./dev/m5/deployment/k8s/aws/cli/lib/cfn.sh:37), [helm.sh](./dev/m5/deployment/k8s/aws/cli/lib/helm.sh:50)
  `--build`, `--base-dir`, `--ma-chart-dir`, `--image-tag`, and `--skip-test-images` were functional in `aws-bootstrap.sh`. The rewrite stores or warns on some of them, but CFN and Helm always fetch release artifacts and image mirroring uses release tags. Developer/local validation flows using source-built templates, chart, and images are effectively broken.

- **[P1/P2] Honor `--ma-images-source` for private image mirroring** — [crane.sh](./dev/m5/deployment/k8s/aws/cli/lib/crane.sh:146)
  Legacy copied MA images from `--ma-images-source` when provided. The new crane path always pulls MA images from `public.ecr.aws`, so the documented “copy from another ECR registry” path for restricted clusters silently does the wrong thing.

- **[P2] Preserve custom kube context behavior** — [resume.sh](./dev/m5/deployment/k8s/aws/cli/lib/resume.sh:83), [helm.sh](./dev/m5/deployment/k8s/aws/cli/lib/helm.sh:133)
  `--kubectl-context` is parsed into state, but `aws eks update-kubeconfig` never uses `--alias`, and Helm/Kubectl calls never pass `--kube-context`/`--context`. This breaks CI and multi-cluster hosts that relied on a predictable context.

- **[P2] Re-implement TLS and EKS access flags** — [resume.sh](./dev/m5/deployment/k8s/aws/cli/lib/resume.sh:95), [helm.sh](./dev/m5/deployment/k8s/aws/cli/lib/helm.sh:133)
  `--tls-mode`, `--pca-arn`, and `--eks-access-principal-arn` are accepted, but I don’t see the old Helm TLS flags or EKS access-entry / Pod Identity setup being applied. Those invocations now look successful while installing without the requested TLS/access configuration.

- **[P2] Restore validation for unsupported/invalid flag combinations** — [resume.sh](./dev/m5/deployment/k8s/aws/cli/lib/resume.sh:89)
  Legacy failed fast for mutually exclusive CFN modes, missing `--stack-name`, import mode without `--vpc-id`/`--subnet-ids`, `--build` with `--version`, TLS mode errors, and VPC endpoint misuse. The new parser mostly records state and proceeds, so bad invocations can run much later and fail in AWS/Helm with less obvious symptoms.

- [P2] Apply user-supplied Helm values — ./dev/m5/deployment/k8s/aws/cli/lib/helm.sh:133-137
  When callers pass `--helm-values custom.yaml`, the parser stores `HELM_EXTRA_VALUES_FILE`, but this Helm invocation never appends it. Deployments that rely on extra values for custom node pools, TLS, or overrides silently run with only generated defaults.

- [P2] Derive the cleanup namespace before using it — ./dev/m5/deployment/k8s/aws/cli/lib/cleanup.sh:32-32
  After a successful deploy saves `HELM_RELEASE`, `migration-assistant cleanup` expands `$HELM_NS` under `set -u` before it is ever assigned. Cleanup therefore exits with an unbound variable and never uninstalls the release or deletes the stack.

- [P2] Download the artifact name that releases publish — ./dev/m5/deployment/k8s/aws/cli/install.sh:76-77
  For non-local installs, this constructs `migrate-cli-<version>.tar.gz` under `releases/download/v<version>`, but `assemble-bootstrap.sh`/release-drafter publish `migration-assistant-cli-<version>.tar.gz` under the actual tag. The curl installer will 404 for published releases.

deployment/k8s/aws/cli/README.md
```diff
@@ -0,0 +1,71 @@
+# migrate-cli
+
+A bash-native CLI for the OpenSearch Migration Assistant. Replaces the
+14k-line Go TUI proposed in
+[opensearch-project/opensearch-migrations#3008](https://github.com/opensearch-project/opensearch-migrations/pull/3008)
+with a single-binary install of pure shell scripts.
```
gregschohn, line 24: not relevant

deployment/k8s/aws/assemble-bootstrap.sh
gregschohn, line 1: I'm getting my head around this now - what's the downside to trying to maintain was-bootstrap.sh as it is today for a couple more releases - as a break-glass sort of escape?

deployment/k8s/aws/cli/lib/crane.sh
```diff
@@ -0,0 +1,371 @@
...
+  # match it here.
+  if ! _crane_mirror_ma_images "$registry" "$ma_ver" "$region"; then
+    return 1
+  fi
+
+  state_set CRANE_REGISTRY "$registry"   # full <host>/<repo>, for helm
+  state_set CRANE_ECR_HOST "$ecr_host"   # bare host, for diagnostics
+  state_set CRANE_MIRRORED "1"
+  state_set last_step "crane_done"
+  state_save
```
gregschohn, line 135: There are 18 different state_save calls that I count in the lib directory.  Each one of those should have tests (& reviews of tests around them).  Is the increased test surface worth the value of a one-time process that will rarely be terminated.

If you tell me that CloudShell will be fighting with users here, that's fair too.

deployment/k8s/aws/cli/lib/resume.sh
```diff
@@ -0,0 +1,282 @@
...
+    2|Agent|agent)   printf 'Agent\n' ;;
+    *)               die "invalid selection: $pick" ;;
+  esac
+}
+
+# manual_path — full deploy + console exec.
+manual_path() {
+  ui_step "Discover environment"
+  discover_os
+  if ! discover_aws; then
```
gregschohn, line 372: doesn't this do the same thing that line 196 is doing?

deployment/k8s/aws/cli/lib/discover.sh
```diff
@@ -0,0 +1,110 @@
...
+      --output text 2>/dev/null || echo "")
+    state_set EKS_CLUSTERS    "${clusters//$'\t'/ }"
+    state_set CFN_MA_STACKS   "${stacks//$'\t'/ }"
+    log_info "discover_resources: clusters='$clusters' stacks='$stacks'"
+  fi
+}
+
+# Internal helper: extract a top-level JSON field. Uses jq if present,
+# otherwise a coarse grep+sed. Good enough for sts:GetCallerIdentity output.
+_jq_or_grep() {
```
gregschohn, line 106: I'm fine requiring jq.  Trying to equate jq and grep is a maintenance bomb.

deployment/k8s/aws/cli/lib/helm.sh
```diff
@@ -0,0 +1,808 @@
...
+
+_write_helm_values() {
+  local out="$1"
+  # The chart's default values.yaml + valuesEks.yaml + the --set flags
+  # we pass on the command line cover everything we need today. This file
+  # exists as a hook so future per-stage overrides can land here without
+  # changing the helm invocation.
+  cat >"$out" <<EOF
+# Generated by migration-assistant — edit at your own risk; overwritten on rerun.
+# Per-stage overrides go here. Chart's values.yaml + valuesEks.yaml are
```
gregschohn, line 1105: this line/file is confusing?  Do I put values in here an then expect them to be overwritten?

deployment/k8s/aws/cli/lib/helm.sh
```diff
@@ -0,0 +1,808 @@
...
+  state_set HELM_RELEASE "$stage"
+  state_set last_step "helm_done"
+  state_save
+}
+
+# helm_ensure_namespace <ns>
+#
+# Idempotent: if the namespace exists, nothing is printed; if it doesn't,
+# we create it and announce it on a single line. No YAML spam.
+helm_ensure_namespace() {
```
gregschohn, line 323: I've seen a few steps like this where we could collapse steps together - in this case w/ helm --create-namespace.  Is there a reason that you are NOT doing that. To get more granular state tracking?

deployment/k8s/aws/cli/lib/helm.sh
gregschohn, line 1: There's a lot going on in this script - and a lot of which I look at and realize that I want this for my local development.  Please provide some documentation for how devs can start to integrate this into their workflows.  The more testing, especially goofy developer cases, the faster we'll be able to squash any issues or discover better patterns.

deployment/k8s/aws/cli/lib/install_tools.sh
```diff
@@ -0,0 +1,175 @@
...
+    brew) _install_with brew "$name" ;;
+    apt)  _install_with apt  "$name" ;;
+    dnf)  _install_with dnf  "$name" ;;
+    yum)  _install_with yum  "$name" ;;
+    *)    _install_static "$name" ;;
+  esac
+}
+
+ensure_tools_basic()    { local t; for t in aws kubectl jq;  do ensure_tool "$t"; done; }
+ensure_tools_extended() { local t; for t in helm crane;       do ensure_tool "$t"; done; }
```
gregschohn, line 43: nit: We should break this up

deployment/k8s/aws/cli/lib/install_tools.sh
```diff
@@ -0,0 +1,175 @@
...
+    crane)   echo "" ;;   # apt has no crane; static path will handle it
+    *)       echo "$1" ;;
+  esac
+}
+_dnf_pkg() { _apt_pkg "$1"; }
+
+# Internal: download a static release into TOOLS_DIR. Verified by SHA-256.
+_install_static() {
+  local name="$1"
+  mkdir -p "$TOOLS_DIR"
```
gregschohn, line 112: Should we do this & put it on the path every time so that when we run ensure_tool, we'll pick up anything that we downloaded in a previous run?

deployment/k8s/aws/cli/lib/install_tools.sh
```diff
@@ -0,0 +1,175 @@
...
+  mv "$tmp/crane" "$TOOLS_DIR/crane"
+  chmod +x "$TOOLS_DIR/crane"
+  rm -rf "$tmp"
+}
+
+_dl_aws() {
+  ui_warn "aws CLI v2 static install requires unzip and a multi-step extract."
+  ui_warn "Please install via your package manager:"
+  case "$(state_get OS_NAME)" in
+    darwin) printf '    brew install awscli\n' ;;
```
gregschohn, line 172: A user might not have brew?  Why do we download jq for everybody, but not was?

deployment/k8s/aws/cli/lib/install_tools.sh
```diff
@@ -0,0 +1,175 @@
...
+    *)      printf '    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o awscliv2.zip && unzip awscliv2.zip && sudo ./aws/install\n' ;;
+  esac
+  die "aws CLI must be installed before re-running migration-assistant"
+}
+
+_dl_jq() {
+  local os="$1" arch="$2"
+  local url
+  case "$os/$arch" in
+    linux/amd64)  url='https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-linux-amd64' ;;
```
gregschohn, line 182: we should be consistent about using arch/OS variables to drive these

deployment/k8s/aws/cli/lib/log.sh
```diff
@@ -0,0 +1,149 @@
...
+
+log_init() {
+  stage_dir_init
+  LOG_FILE="$STAGE_DIR/log/migrate.log"
+  if [[ -f "$LOG_FILE" ]]; then
+    local sz
+    sz=$(wc -c <"$LOG_FILE" 2>/dev/null | tr -d ' ' || echo 0)
+    if [[ "$sz" -gt 5242880 ]]; then
+      mv -f "$LOG_FILE" "$LOG_FILE.1"
+    fi
```
gregschohn: Is there a reason that we only want to keep one file?  I'd open every file with timestamp (seconds).  You can compress the old one(s) when you come in and create a new one.  Using a separate file for each run will make it obvious and easy to keep logs for runs straight

deployment/k8s/aws/cli/lib/log.sh
```diff
@@ -0,0 +1,149 @@
...
+  } >>"$LOG_FILE"
+}
+
+# log_announce — tell the operator where the log is. Called by the banner
+# at startup and registered as an on_exit handler so it's the last thing
+# printed even on a normal exit. Stderr-only — never pollutes stdout.
+log_announce() {
+  [[ "$LOG_FILE" == "/dev/null" || -z "$LOG_FILE" ]] && return 0
+  if [[ "${1:-}" == "--exit" ]]; then
+    printf '\n%sLog: %s%s\n' "$__UI_C_DIM" "$LOG_FILE" "$__UI_C_RESET" >&2
```
gregschohn, line 77: nit, this file shouldn't depend on anything in the ui file.

deployment/k8s/aws/cli/lib/log.sh
```diff
@@ -0,0 +1,149 @@
...
+# log_announce_exit — on_exit_register hook: print the log path one more
+# time as the very last operator-visible line. Receives <rc> as $1.
+log_announce_exit() {
+  log_announce --exit
+}
+
+# log_stream <prefix> CMD ARGS… — run a long-running command and tee its
+# combined output to the log file AND stderr, one prefixed line at a time.
+#
+# Why this exists: `aws cloudformation deploy` (and helm install, crane
```
gregschohn, line 144: nit: call out why you need all of this machinery & mkfifo over just tee (the reasons are good ones)

deployment/k8s/aws/cli/lib/resume.sh
```diff
@@ -0,0 +1,282 @@
...
+  # First pass: parse only --stage, so we know which stage's state to
+  # load. Everything else is parsed in the second pass below, AFTER
+  # state_load — otherwise the flag-driven state_set calls would write
+  # against an empty in-memory state, blowing away last_step etc.
+  local i
+  for ((i = 1; i <= $#; i++)); do
+    case "${!i}" in
+      --stage)
+        local next=$((i + 1))
+        STAGE="${!next}"
```
gregschohn, line 30: Stage continues (in all of our versions) sounds confusing to me.  Maybe it makes some sense since we use it as a tag in a lot of other names.  Tag doesn't make sense since that's another concept altogether.  Name won't be a complete name for many of the resources that will use stage as PART of its name.  Just wondering if you've thought of anything better, or a better way to explain what it is in help/docs.  It always sounds like it's an optional disambiguator for another value.

deployment/k8s/aws/cli/lib/ui.sh
```diff
@@ -0,0 +1,218 @@
+#!/usr/bin/env bash
+# ui.sh — terminal UI primitives. Color, prompts, spinner, banners.
+#
+# Design rules:
+#   * No external deps. Everything is printf + tput.
+#   * If stderr is not a TTY, drop colors silently (CI-safe output).
+#   * Prompts read from /dev/tty when stdin is not a TTY (curl-piped
+#     installer scenario), so install.sh can still ask questions.
```
gregschohn, line 12: should this be an option?  It seems like it could limit some e2e testing for the outermost scripts

deployment/k8s/aws/cli/lib/ui.sh
```diff
@@ -0,0 +1,218 @@
...
+#
+# *** CRITICAL: ALL UI CHROME GOES TO STDERR. ***
```
gregschohn: I read chrome and I think of the browser.  Can you use another word or is this a known shell-UI term?

deployment/k8s/aws/cli/lib/ui.sh
```diff
@@ -0,0 +1,218 @@
...
+    answer="$default_yn"
+  fi
+
+  # Lowercase first letter for the comparison. Bash 3.2 has no ${var,,}.
+  local first
+  first=$(printf '%s' "$answer" | cut -c1 | tr '[:upper:]' '[:lower:]')
+  case "$first" in
+    y) return 0 ;;
+    n) return 1 ;;
+    *) return 1 ;;   # unknown input — be conservative
```
gregschohn: we should re-ask if they didn't type yes/no variants

deployment/k8s/aws/cli/test/helpers/stub.sh
```diff
@@ -0,0 +1,66 @@
...
+  export HOME="$TMPHOME"
+  export MIGRATE_HOME="$TMPHOME/.opensearch-migrate"
+  export STAGE="${STAGE:-default}"
+  export STAGE_DIR="$MIGRATE_HOME/$STAGE"
+  export STUB_DIR="$TMPHOME/stubs"
+  mkdir -p "$STUB_DIR"
+  export PATH="$STUB_DIR:$PATH"
+}
+
+teardown_isolated_home() {
```
gregschohn, line 22: Should you have a TRAP to do this too?

agent-skills/skills/Startup.md
```diff
@@ -0,0 +1,94 @@
...
+   reindex (or backfill), live-traffic capture/replay if needed, and
+   cutover/teardown. Show the plan to the user. **Wait for explicit
+   approval** before executing anything that modifies the source or target.
+
+## Where to put things
+
+- `plan/`        — your planning documents. One file per plan revision.
+- `history/`     — append-only operator-visible action log (`history/<ts>.md`)
+                   for anything you ran that touched AWS, the source, or
+                   the target.
```
gregschohn, line 228: what if you used the filesystem as a DB a bit more.  You could write each record as a separate readonly file (umask 277...).  I'm a bit worried that the agent could rewrite history.  I know that it could always chmod the files, but it's an extra hoop and one that you could block w/out approval.

agent-skills/skills/Startup.md
gregschohn, line 1: as I've noted elsewhere, the re-entry for the agent should be tightened up.  That might make these instructions a little bit tighter.  I suspect you might need to move steps 2-5 (first things to do) to a file that will be picked up by users just installing the skill too - and step two here just defers to it.  This file probably should JUST be about establishing the conventions that match the rest of this script - when you don't use this launcher, agents shouldn't need to play along with any of those conventions.  That seems like a better split of responsibilities.

deployment/k8s/aws/cli/lib/state.sh
```diff
@@ -0,0 +1,153 @@
...
+    local i k v
+    for ((i=0; i < ${#STATE_KEYS[@]}; i++)); do
+      k="${STATE_KEYS[$i]}"
+      v="${STATE_VALS[$i]}"
+      v=${v//\"/\\\"}
+      printf '%s="%s"\n' "$k" "$v"
+    done
+  } >"$env_tmp"
+
+  if command -v jq >/dev/null 2>&1; then
```
gregschohn, line 111: if I have a deployment from another machine, can I launch that install w/out jq?  I guess we can because the aws client has some filtering that we can use for cfn?  I'd love to drop these parallel-tool conditional branches (& just require jq)

deployment/k8s/aws/cli/lib/ui.sh
```diff
@@ -0,0 +1,218 @@
...
+# ui_table <header_line> <row1> <row2>…
+# Renders a simple monospace table. Caller pre-formats columns with tabs.
+ui_table() {
+  local header="$1"; shift
+  printf '%s%s%s\n' "$__UI_C_BOLD" "$header" "$__UI_C_RESET" >&2
+  local row
+  for row in "$@"; do
+    printf '  %s\n' "$row" >&2
+  done
+}
```
gregschohn, line 202: Maybe callers already did this, but here's a width-normalized table implementation

```
# ui_table <header_line> <row1> <row2>…
# Auto-aligned monospace table. Columns are tab-separated within each line;
# widths come from the widest visible cell per column. ANSI color escapes are
# ignored when measuring, so padding stays correct. Output -> stderr.
# bash 3.2 compatible.
ui_table() {
  local header="$1"; shift
  local -a lines=("$header" "$@")
  local -a widths=()
  local -a __UI_FIELDS=()      # scratch — kept out of the global namespace
  local __UI_VISLEN=0          # scratch   via dynamic scope (see below)
  local line i out last pad sp gap='  ' first=1

  # pass 1: widest visible cell per column
  for line in "${lines[@]}"; do
    __ui_split_tab "$line"                       # fills __UI_FIELDS
    for i in "${!__UI_FIELDS[@]}"; do
      __ui_vislen "${__UI_FIELDS[i]}"            # fills __UI_VISLEN
      (( __UI_VISLEN > ${widths[i]:-0} )) && widths[i]=$__UI_VISLEN
    done
  done

  # pass 2: emit, padding every cell except the last in each row
  for line in "${lines[@]}"; do
    __ui_split_tab "$line"
    out=''; last=$(( ${#__UI_FIELDS[@]} - 1 ))
    for i in "${!__UI_FIELDS[@]}"; do
      out=$out${__UI_FIELDS[i]}
      if (( i < last )); then
        __ui_vislen "${__UI_FIELDS[i]}"
        pad=$(( ${widths[i]:-0} - __UI_VISLEN )); (( pad < 0 )) && pad=0
        printf -v sp '%*s' "$pad" ''
        out=$out$sp$gap
      fi
    done
    if (( first )); then
      printf '  %s%s%s\n' "$__UI_C_BOLD" "$out" "$__UI_C_RESET" >&2; first=0
    else
      printf '  %s\n' "$out" >&2
    fi
  done
}

# split $1 on tab -> array __UI_FIELDS (empty cells preserved)
__ui_split_tab() {
  local s=$1
  __UI_FIELDS=()
  while [[ $s == *$'\t'* ]]; do
    __UI_FIELDS+=( "${s%%$'\t'*}" )
    s=${s#*$'\t'}
  done
  __UI_FIELDS+=( "$s" )
}

# visible length of $1 (ANSI CSI escapes removed) -> int __UI_VISLEN
__ui_vislen() {
  local s=$1 esc=$'\x1b' re
  re="$esc"'\[[0-9;]*[A-Za-z]'
  while [[ $s =~ $re ]]; do
    s=${s//"${BASH_REMATCH[0]}"/}
  done
  __UI_VISLEN=${#s}
}
```

deployment/k8s/aws/cli/lib/version.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Where we look for new CLI releases.
+CLI_RELEASES_API='https://api.github.com/repos/opensearch-project/opensearch-migrations/releases/latest'
+
+# version_check — print a one-liner if a newer CLI version is published.
+# Never auto-updates. Network errors are silent.
+#
+# `set -e + pipefail` would kill the whole CLI on a 404, so we run the
+# pipeline in a forgiving subshell (`|| true`) and cache stderr to a temp.
+version_check() {
```
gregschohn: nit: could be renamed - I thought that this was getting the versions, not just checking that I was running w/ the latest version

deployment/k8s/aws/cli/lib/version.sh
```diff
@@ -0,0 +1,60 @@
...
+# CLI_VERSION is baked at install time. MA_VERSION is what we pin when
+# downloading CFN/helm/skills artifacts; user-overridable via wizard.
+
+[[ -n "${__MIGRATE_VERSION_LOADED:-}" ]] && return 0
+__MIGRATE_VERSION_LOADED=1
+
+CLI_VERSION="${CLI_VERSION:-0.1.0}"
+
+# Default Migration Assistant artifact version (matches PR 3008 MA_VERSION).
+DEFAULT_MA_VERSION="3.2.1"
```
gregschohn: We need to pick this up some other way so that releases don't need to update this file

deployment/k8s/aws/cli/lib/wizard.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+# wizard.sh — collect deploy-time parameters only.
+#
+# Three fields. That's it. Anything migration-specific (source endpoint,
```
gregschohn: nit: s/fields/inputs/ ?  I'm not sure what fields means here

deployment/k8s/aws/cli/lib/wizard.sh
```diff
@@ -0,0 +1,64 @@
...
+    ui_dim "    mirror_images=$(state_get MIRROR_IMAGES)"
+    ui_dim "    ma_version=$(state_get MA_VERSION)"
+    return 0
+  fi
+
+  local stage_name mirror ma_ver
+
+  # Default to "ma" — the kubernetes namespace the chart expects.
+  # See lib/helm.sh comment for why release name == namespace.
+  ui_prompt "Stage name (helm release + k8s namespace + CFN suffix)" \
```
gregschohn: nit: 
Stage name (helm release + k8s namespace + CFN suffix)
->
Stage name (used for helm release, k8s namespace, and CFN suffix)

deployment/k8s/aws/cli/lib/wizard.sh
```diff
@@ -0,0 +1,64 @@
...
+      && [[ -n "$(state_get MIRROR_IMAGES)" ]] \
+      && [[ -n "$(state_get MA_VERSION)" ]]; then
+    ui_dim "  resuming — using saved deploy config:"
+    ui_dim "    stage_name=$(state_get STAGE_NAME)"
+    ui_dim "    mirror_images=$(state_get MIRROR_IMAGES)"
+    ui_dim "    ma_version=$(state_get MA_VERSION)"
+    return 0
+  fi
+
+  local stage_name mirror ma_ver
```
gregschohn: Does this flow have the ability to use an existing CFN stack or EKS deployment (terraform, etc).  I'm not seeing how if those were installed outside of this script?  I'd expect to see a discovery check here and if found, skip the stage name.

gregschohn (commented):
Finishing the first round of a review.  I need to run this and get my head around how some things work (like overall terminal formatting).  I think that this is a very strong addition and one that we should pursue!  I'll make a pass over the newer edits shortly

Is this bash 3.2 or bash 4.3 compatible?  Mac is installed with 3.2, but you can brew install 4.3.  
Here's what I've got on my Mac
```
% bash --version
GNU bash, version 5.3.9(1)-release (aarch64-apple-darwin25.1.0)
Copyright (C) 2025 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
% which bash
/opt/homebrew/bin/bash
```

deployment/k8s/aws/cli/lib/discover.sh
epugh: agreed.   and once you know you have `jq`, you will find many more uses for it!

deployment/k8s/aws/cli/LICENSE
jzonthemtn, line 1: Is this `LICENSE` file needed here?

AndreKurait: Removed thanks!

AndreKurait:
Moving forward with rust based approach https://github.com/opensearch-project/opensearch-migrations/pull/3082

Issue #83, opened 2023-01-20
Shadow Validation Replayer - Add error handling for replayed requests


mikaylathompson:
As a user, I would like failed write requests to be retried a few (TBD) times and logged in such a way that they can be visualized by the Analyzer. This task is a follow-up to #81.

### Acceptance Criteria
- Requests from the original cluster should be identified as mutating or non-mutating and mutating requests should be retried a reasonable number of times before ultimately failing.
- A data interface change should be documented to explain how initially-failing requests are indicated so they can be identified by the analyzer.

Issue #82, opened 2023-01-20
Create or Select Tooling to Simulate Cluster Traffic


chelma:
## Task
In order to test our cluster duplication tooling, we need to have tooling to simulate traffic being sent to the Cluster.  While in the long term we'll probably want to have a fairly sophisticated tool that can be run on a collection of hosts and pointed at a Cluster across a real network, for the time being we can likely get away with something simple that's run off a developer laptop.

As a starting point, you may investigate this existing tool [1] to see if it solves the problem.

[1] https://github.com/opensearch-project/opensearch-benchmark-workloads

## Acceptance Criteria
* Tooling should be runnable on a developer laptop via command line and pointed at a hostname/port
* Tooling should be able to simulate data reads (get indexes/docs), data writes (put indexes/docs), configuration reads (get cluster status), and configuration writes (change index settings, etc).
* The ratio and rate of the requests sent by the tooling should be configurable.  For this task, we'll focus on constant-rate requests instead of variable rate requests. - sumobrian@: this is optional
* The behavior of the tooling should be repeatable.  That is, if an issue is identified with one run of the tooling, we should be able to do that exact same run again to provoke the same response from the system under test.
* Demo tool to team

chelma:
Task might be obviated by this existing tool: https://github.com/opensearch-project/opensearch-benchmark-workloads

sumobrian:
Estimate: 5 - Assumption is that we will not have to write code but if we do it will be basic.

PR #3026, opened 2026-05-28
Fix(jenkins): tag isolated VPC subnets for AWS Load Balancer Controller discovery

jugal-chauhan:
### Summary
Tag the EKS subnets created by `createIsolatedVpc` with `kubernetes.io/role/internal-elb=1` so AWS Load Balancer Controller can discover them when provisioning the internal NLB for `capture-proxy`.

### Problem
`main-eks-full-e2e-isolated-vpc-test` builds [#82](https://migrations.ci.opensearch.org/job/main/job/main-eks-full-e2e-isolated-vpc-test/82/) and [#83](https://migrations.ci.opensearch.org/job/main/job/main-eks-full-e2e-isolated-vpc-test/83/) fail in Phase 2 at `waitForProxyEndpointReady`, surfacing as `captureproxy/capture-proxy entered Error phase:` (empty message).

The underlying event on the `capture-proxy` Service:

```
Warning  FailedBuildModel  (x19 over 22m)  service
Failed build model due to unable to resolve at least one subnet (0 match VPC and tags: [kubernetes.io/role/internal-elb])
```

AWS LBC uses tag-based subnet auto-discovery for internal NLBs. With zero matching subnets, no NLB is provisioned, the Service stays `EXTERNAL-IP: <pending>` for the full 20-minute readiness window, and the workflow exits 1.

### Why CDK tagging didn't cover this path

The standard pipeline lets CDK create the VPC, and `eks-infra.ts` tags the private subnets via `Tags.of(subnet).add(...)`. The isolated-VPC pipeline supplies a pre-existing VPC, so CDK takes the Bring-your-own-VPC route and `Subnet.fromSubnetId(...)` returns imported constructs. `Tags.of()` is a no-op on imported resources, so the tag never lands.

### Fix applied

In `vars/eksIsolatedIntegTest.groovy`, immediately after `createIsolatedVpc(...)`, run `aws ec2 create-tags` against the returned subnet IDs.

The Jenkins host already holds the IAM perms (the same role calls `ec2:CreateVpc` / `ec2:CreateSubnet` with tag specs in the preceding step). The call is idempotent and runs ~10 minutes before LBC reads the tags, so eventual consistency is a non-issue.

With this change, now if the `aws ec2 create-tags` call ever fails, the build fails within seconds at the new step with the AWS error attached, instead of after a 20-minute silent timeout in `waitForProxyEndpointReady`.

### Testing
- Waiting on https://migrations.ci.opensearch.org/job/pr-checks/job/pr-eks-full-e2e-isolated-vpc-test/7/

### Check List

### DIFF
--- a/vars/eksIsolatedIntegTest.groovy
+++ b/vars/eksIsolatedIntegTest.groovy
@@ -143,6 +143,14 @@ def call(Map config = [:]) {
                                 isolatedVpcId = vpc.vpcId
                                 env.isolatedVpcId = vpc.vpcId
 
+                                // Tag subnets so AWS Load Balancer Controller can discover them for internal NLBs.
+                                // CDK's Tags.of() is a no-op on imported subnets, so we tag them here at the source.
+                                sh """
+                                    aws ec2 create-tags --region ${params.REGION} \
+                                      --resources ${vpc.subnetIds.replace(',', ' ')} \
+                                      --tags Key=kubernetes.io/role/internal-elb,Value=1
+                                """
+
                                 // Deploy EKS + MA into isolated subnets
                                 def bootstrap = resolveBootstrap(
                                     useReleaseBootstrap: params.USE_RELEASE_BOOTSTRAP,


oeyh (approved):
Nice!

PR #3027, opened 2026-05-28
Fix workflow log resource failing to find pods after run-number label added to CRs

oeyh:
### Description
`workflow log resource` builds a pod label selector from ALL `migrations.opensearch.org/*` labels on the CR. After run-number was added to CRs, the resulting AND selector includes labels that workflow pods don't carry (`run-number`, `workflow-name`), causing no pods to match.

This fix adds a `_CR_ONLY_LABELS` exclusion set so those CR-metadata labels are filtered out of the pod selector.

### Issues Resolved
N/A

### Testing
- Added unit test verifying CR-only labels are excluded from the selector
- Manually verified `workflow log resource` works for all resources on a live cluster

### Check List


migrationConsole/lib/console_link/console_link/workflow/commands/log.py
```diff
@@ -26,6 +26,11 @@
 
 logger = logging.getLogger(__name__)
 _RESOURCE_OUTPUT_LABELS = {'strimzi.io/cluster'}
+# Labels present on CRs but not propagated to workflow pods.
+_CR_ONLY_LABELS = {
```
gregschohn: This name could be more descriptive...

gregschohn (commented):
I wonder if we should change the names of the labels to include a longer prefix to disambiguate the keys taht we want to automatically pick up vs not - like 'migrations.opensearch.org/config/source' and search for migrations.opensearch.org/config instead of just migrations.opensearch.org.  Right now, we're set to repeat this pattern again and the test below can't help it if we think of a new label to put onto the resource

gregschohn (approved):

migrationConsole/lib/console_link/console_link/workflow/commands/log.py
oeyh: Renamed it to `_CR_ONLY_LABELS to _LABELS_NOT_PROPAGATED_TO_PODS` for clarity.

oeyh: > I wonder if we should change the names of the labels to include a longer prefix to disambiguate the keys taht we want to automatically pick up

This design makes more sense. But I'd prefer to keep this PR as a targeted fix for the immediate breakage and address the label namespace design separately.

PR #3028, opened 2026-05-30
Mountable data v2

gregschohn:
### Description

The previous transform mount design had two distinct concepts (`TRANSFORMS_SOURCE` for code, ad-hoc fields elsewhere) that all expanded into the same kind of pod volume but could not be combined or deduplicated. This PR replaces those scattered definitions with a single `FILE_REF` union — either a ConfigMap key or a path inside a mountable OCI image — and uses it uniformly for transform entry points, transform contexts, transformer config files, and proxy mTLS trust roots. The names also stop colliding with "source cluster" terminology that appears nearby in the same schema.

The config processor lowers every `FILE_REF` it finds in a user config into deduped `fileSourceVolumes` and `fileSourceVolumeMounts` lists per compute resource, so a single ConfigMap or image referenced by multiple transforms produces exactly one mount, and each pod only carries the mounts it actually needs. Workflow-only fields are stripped from custom resources with a trace annotation so the migration CRDs stay clean while the workflow retains enough information to debug what was resolved.

Transform providers can now read their context values from mounted files instead of inline JSON. Each entry in a transform's `context.values` is either an inline literal (`{value: ...}`) or a file reference (`{fromFile: {...}}`), and providers can declare per-key materialization types so structured config files (e.g. type mapping rules) can be parsed as JSON instead of treated as plain strings. Whole-directory contexts via `valueDirectories` let a single ConfigMap or image directory expose all of its files as context entries.

The capture proxy accepts mTLS client trust roots from either a mounted file (`trustedClientCaFile`) or an inline PEM environment variable (`trustedClientCaPem`), with `requireClientAuth` derived automatically from which trust source is configured. Schema refinements reject paths that escape the mount root, ConfigMap keys that contain nested path segments, and client auth blocks with neither or both trust sources, so misconfigurations fail at config load instead of at pod startup.

Workflow templates project the resolved volumes and mounts into metadata, RFS, replayer, and capture proxy pods through new container builder helpers, removing the per-template image/configMap plumbing that was previously copy-pasted across each pipeline stage.

A design doc (`docs/FileBundlesDesign.md`) walks through the lowering rules and the deduplication contract.

### Testing

Unit tests cover invalid file refs (absolute paths, parent traversal, nested ConfigMap keys), inline transform context values, deduplicated mounts across pipeline entries, mTLS trust root selection, and the workflow-only-field annotation path. Integration tests under `migrationConsole/lib/integ_test/integ_test/test_cases/mountable_transform_tests.py` exercise the full mount path end-to-end with both ConfigMap- and image-backed transforms and contexts. Snapshot tests under `migration-workflow-templates` and `schemas` are regenerated against the merged config surface.

### Check List


docs/FileBundlesDesign.md
```diff
@@ -0,0 +1,917 @@
...
+  }
+}
+```
+
+Final config passed to the named provider:
+
+```json
+{
+  "directoryKey": "value loaded from a value directory",
+  "regexMappings": "JSON value parsed from regexMappings.json",
```
akshay2000, line 318: How does the plumbing know to parse the text as a JSON? I see the following:

> For named providers, it asks the selected `IJsonTransformerProvider` what type each file-backed config key expects.

Are we extending the interface to return value types for config key? 

**EDIT:** Progressed further in the PR. Yes, we are adding a new method to resolve the config keys into value types.

docs/FileBundlesDesign.md
```diff
@@ -0,0 +1,917 @@
...
+  "featureFlags": "JSON value parsed from featureFlags.json",
+  "sourceProperties": {
```
akshay2000, line 320: How does this translate to a Java type? Will `sourceProperties` be a nested Map?

docs/FileBundlesDesign.md
```diff
@@ -0,0 +1,917 @@
...
+    }
+  }
+}
+```
+
+Final config passed to the named provider:
+
+```json
+{
+  "directoryKey": "value loaded from a value directory",
```
akshay2000, line 317: It is not clear what this key signifies. Where is the value coming from? Does the provider need to declare this as a config param in its constructor? What if it doesn't?

docs/FileBundlesDesign.md
gregschohn: It will just be a nested map.  It will be up to the provider to interpret this however it wants to

docs/FileBundlesDesign.md
gregschohn: this would be another value in the bindings object.  It doesn't declare it in a formal, strongly-typed registry, but it should parse for such a key.  It should also prominently document what the key's value does and what types it accepts.  It's just an example for right now of something that can be passed from the outside to the inside.

docs/FileBundlesDesign.md
gregschohn: It's totally up to the provider.

orchestrationSpecs/packages/schemas/tests/fixtures/valid/solr-source.json
```diff
@@ -2,7 +2,24 @@
   "sourceClusters": {
     "solr-source": {
       "version": "SOLR 8.11.4",
-      "endpoint": "http://solr:8983"
+      "endpoint": "http://solr:8983",
+      "clusterConfiguration": {
```
AndreKurait: Let's revert this and just leave it in the file transforms

AndreKurait (approved):
I'm aligned after making the changes discussed offline about removing changes to the "sourceClusters" workflow config

PR #3029, opened 2026-05-30
Make S3 tuple output durable before replay commits

gregschohn:
### Description
Tuple S3 output now writes each gzip JSON-lines batch to a local temp file before uploading it to S3, then retries transient S3 upload failures indefinitely with the same object key while keeping the tuple futures pending. This policy keeps replay progress tied to durable tuple output without buffering full objects in memory or losing already captured tuple data when S3 is temporarily unavailable.

Hard tuple write failures now fail fast instead of quietly leaving offsets uncommitted. Local file creation, serialization, gzip close, or sink-level exceptional futures shut down the replayer with a clear fatal error and do not commit the held traffic stream offsets, so operators see an explicit failure rather than a replayer that restarts and stalls at the same point.

The tuple sink and replayer tests cover same-key S3 retry, source commit gating on tuple future completion, nonblocking handling of later requests while tuple futures are pending, and fatal shutdown without offset commit for hard tuple write failures.

### Check List


AndreKurait (approved):
LGTM, let's consider this approach when adding DLQ to RFS

PR #3032, opened 2026-06-02
Feature/adding further solr source version tests

awagen:
### Description
- added Solr 6, Solr 7 and Solr 9 to full migration test
- added custom image build for Solr 7
- fixed issue of failing numerical field detection for Lucene 6, 7
- added solr 6 and solr 9 tests to jenkins tests (left 7 out for now due to apache archive reliance for pulling solr-7 archive, which makes the download very slow)

### Check List


.github/workflows/jenkins_tests.yml
```diff
@@ -84,6 +84,26 @@ jobs:
           jenkins_api_token: "${{ secrets.JENKINS_MIGRATIONS_API_TOKEN }}"
           job_timeout_minutes: "120"
 
+  solr-9x-k8s-local-test:
```
AndreKurait: Let's remove this, the idea we have is to not run this on every single Pr, but to run https://migrations.ci.opensearch.org/job/main/job/main-k8s-matrix-test/ if requested or on mainline for all the versions

buildImages/build.gradle
```diff
@@ -228,7 +246,8 @@ def customESEntries = [
 
 // Test/CI-only images — not built by buildKitAll, use buildKitTestAll instead
 def testBuildKitProjects = [
-    *(includeSolr660TestImage ? [solr660Entry] : []),
+    *(excludeSolr660TestImage ? [] : [solr660Entry]),
+    *(includeSolr773TestImage ? [solr773Entry] : []),
```
AndreKurait, line 250: does it make sense here to just have one flag to exclude all solr images, like we have excludeESCustomTestImages

awagen: In general agree. Yet: distinction right now comes from the fact that we have a fast mirror for Solr 6 and a slow archive one for Solr 7, so benefit of exclusion is mainly 7 atm. Haven't found a faster public one so far.

orchestrationSpecs/packages/migration-workflow-templates/resources/scripts/configureAndSubmitWorkflow.sh
```diff
@@ -48,6 +50,6 @@ echo "$CONFIG_OUTPUT"
 echo "Submitting workflow..."
 WORKFLOW_OUTPUT=$(
   kubectl exec --namespace "$WORKFLOW_NAMESPACE" "$MIGRATION_CONSOLE_POD" -- \
-    /bin/bash -lc 'workflow submit' 2>&1
+    /bin/bash -lc "workflow submit --unique-run-nonce $RUN_NONCE" 2>&1
```
AndreKurait: Why was this needed? I think we can revert changes to this file and leave the rest of the PR as is.

awagen: Hi @AndreKurait . This was needed for the following reason: for the backup functionality in solr 6 / 7 we only have local backup available (and we emulate the localstack backup via sync). Local backup fails if the path is not known upfront (and created). The path passed in the backup call are generated as UUID if unique-run-nonce is not set, effectively leading to a fail.

wrigleyDan: reverted with https://github.com/opensearch-project/opensearch-migrations/pull/3032/commits/1077da02cae016d9981a9b4efdffb829b813325a

AndreKurait (approved):

PR #3048, opened 2026-06-03
Fix CI 403 failures: Jenkins webhook trigger + SonarQube scanner

AndreKurait:
## Description

Fixes two unrelated 403 Forbidden failures that are currently blocking CI on every PR.

---

## 1. Jenkins webhook trigger 403

The `pull_request_target` and `push` paths fail with:

```
2026-06-03 02:46:53 [WARNING] Unable to parse trigger request response: Expecting value: line 1 column 1 (char 0)
2026-06-03 02:46:53 [ERROR] Failed to trigger webhook for URL: https://migrations.ci.opensearch.org/generic-webhook-trigger/invoke
... requests.exceptions.HTTPError: 403 Client Error: Forbidden
```

### Root cause

The Jenkins [Generic Webhook Trigger plugin](https://plugins.jenkins.io/generic-webhook-trigger/) authenticates via the `?token=XXX` URL query parameter — **not** via an `Authorization: Bearer` header. The current code in `default_webhook_trigger.py` sends the token as an `Authorization` header, so Jenkins responds with 403 and `Did not find any jobs with GenericTrigger configured!`.

This was previously diagnosed in #2971 (closed without merge) — that PR carried the 3-line auth fix. This PR re-applies that fix and layers in robustness improvements so the next failure is straightforward to debug.

### Changes

**Auth (the actual fix):**
- Pass `pipeline_token` as the URL-encoded `?token=` query parameter
- Drop the `Authorization: Bearer ...` header

**Error reporting:**
- Surface the real Jenkins status code and response body on non-2xx (e.g. `status=403 body='Did not find any jobs with GenericTrigger configured!'`). Previously the body was buried under a `json.JSONDecodeError` logged as a WARNING — the real 403 was visible only in the trailing traceback.
- Strip the token from logged URLs (`webhook_base_url` is used for logs; `webhook_url` only for the actual POST) so CI logs cannot leak the secret.

**Network resilience:**
- Add `(connect=10s, read=60s)` HTTP timeouts on every `requests.get` / `requests.post` so a hung Jenkins endpoint cannot stall the action indefinitely.
- Reuse a single retry-enabled `Session` across queue and build polling for consistent transient-5xx retry behaviour.
- Set `Retry(allowed_methods=…)` explicitly so urllib3 doesn't silently skip retries on POST.

**Burst-load 403 retry** (commit `bb207c2`):
- Even with the URL-token auth fix, the Generic Webhook Trigger plugin returns intermittent 403s under the workflow's burst load (~14 concurrent webhook calls when a PR opens). Two calls 2 seconds apart with identical credentials can return 200 and 403 — purely a server-side flake in the plugin's in-memory token/job index.
- 5 attempts with exponential backoff (2s, 4s, 8s, 16s) wrapping the trigger POST.
- Retries on connection errors plus 403, 404, 408, 425, 429, 5xx. **401 is not retried** — that's a real auth failure.
- Each retry logs the failed attempt with status + truncated body.
- Application-level wrapper rather than urllib3's `Retry` because urllib3 short-circuits on 403/404 with no clean per-request override.

### Tests

`pytest tests/ -v` — 20 tests, all pass; flake8 clean.

New test coverage:
- Token rides in the URL, never in the `Authorization` header
- Special characters in the token are URL-encoded
- 403 / 401 / non-JSON bodies all surface `status=N body=...` in error logs
- Token does not appear anywhere in error log output
- `_summarize_response` truncates long bodies and handles `None`
- `WebhookRetryTest`: 403-then-success recovery, persistent 403 still fails after exhausting retries, 401 fails fast (no retry), connection errors trigger retry
- Patched `time.sleep` at class level so the suite runs in 0.05s instead of 3 minutes

Existing tests migrated from `requests.get` to `requests.Session.get` to match the new polling path.

### Testing in real Jenkins

Per the action's [README](https://github.com/opensearch-project/opensearch-migrations/blob/main/.github/actions/jenkins-trigger/README.md), end-to-end verification requires a `jenkins-trigger-test-*` branch on the upstream repo since `pull_request_target` reads workflow YAML from `main`. Happy to push to such a branch after this PR is reviewed.

---

## 2. SonarQube scanner download 403 (commit `825e036`)

The SonarQube job is also failing on every run:

```
curl: (22) The requested URL returned error: 403
curl: (22) The requested URL returned error: 403
... (× 6, one per retry)
```

### Root cause

`binaries.sonarsource.com` has started returning 403 for GitHub-hosted runner egress IPs. The URL is correct and the file still exists — SonarSource appears to be geo/IP-filtering that host. Every retry hits the same 403, so the workflow's existing `--retry 5` cannot recover.

### Fix

- Switch the download URL to Maven Central (`repo1.maven.org/maven2/org/sonarsource/scanner/cli/sonar-scanner-cli/...`), which is the canonical mirror SonarSource publishes to. It's Cloudflare-backed and does not exhibit the same blocking behavior.
- Bump the scanner from `6.2.1.4610` (Oct 2024) to the latest stable `8.1.0.6389` (Apr 2026).
- The zip layout (`sonar-scanner-${VERSION}-linux-x64/bin/sonar-scanner`) is unchanged, so the rest of the script (extract, `PATH` export, `sonar-scanner` invocation) needs no other edits.
- SHA256 updated to `bb8f709f9cb73352f8d1260a3b3c506c0f41146754bc630762c126d795499d0b`, verified against the `.sha256` published alongside the zip on Maven Central.

### DIFF
--- a/.github/actions/jenkins-trigger/webhook-trigger/default_webhook_trigger.py
+++ b/.github/actions/jenkins-trigger/webhook-trigger/default_webhook_trigger.py
@@ -7,13 +7,18 @@ import sys
 import time
 from dataclasses import dataclass
 from typing import Optional
+from urllib.parse import urlencode
 
 import requests
-from requests import Response
+from requests import Response, Session
 from requests.adapters import HTTPAdapter
 from urllib3.util import Retry
 
 logging.basicConfig(format='%(asctime)s [%(levelname)s] %(message)s', level=logging.INFO)
+
+# HTTP timeouts: (connect, read). Read timeout is generous because Jenkins endpoints
+# can be slow under load, but we want to fail rather than hang forever.
+DEFAULT_TIMEOUT = (10, 60)
 
 
 @dataclass
@@ -43,15 +48,18 @@ class JenkinsConfig:
         return bool(self.jenkins_user and self.jenkins_api_token)
 
     @property
-    def webhook_url(self) -> str:
+    def webhook_base_url(self) -> str:
         return f"{self.jenkins_url}/generic-webhook-trigger/invoke"
 
     @property
+    def webhook_url(self) -> str:
+        # The Jenkins Generic Webhook Trigger plugin authenticates via the `token`
+        # query parameter, not via Authorization headers.
+        return f"{self.webhook_base_url}?{urlencode({'token': self.pipeline_token})}"
+
+    @property
     def auth_headers(self) -> dict[str, str]:
-        return {
-            "Content-Type": "application/json",
-            "Authorization": f"Bearer {self.pipeline_token}",
-        }
+        return {"Content-Type": "application/json"}
 
     @property
     def payload(self) -> dict[str, str]:
@@ -88,7 +96,7 @@ def _cancel_jenkins_build(signum: int, frame: object) -> None:
         try:
             stop_url = f"{_active_build_url}stop"
             auth = (_active_config.jenkins_user, _active_config.jenkins_api_token)
-            response = requests.post(stop_url, auth=auth)
+            response = requests.post(stop_url, auth=auth, timeout=DEFAULT_TIMEOUT)
             logging.info(f"Sent stop request to {stop_url}, response: {response.status_code}")
         except Exception as e:
             logging.error(f"Failed to cancel Jenkins build: {e}")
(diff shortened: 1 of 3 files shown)


AndreKurait:
jenkins failure in this PR will be addressed AFTER merging since it's still running the old version

gregschohn (approved):

PR #3061, opened 2026-06-04
ci: open replacement PR when Pipfile.lock hash fix can't push to dependabot branch

AndreKurait:
## Summary

- The `Fix Dependabot Pipfile.lock hash` workflow tries to commit the corrected `_meta.hash` straight back to the `dependabot/**` PR branch. That now fails because an opensearch-project org ruleset enforces PR-only updates on `dependabot/**` refs:

  ```
  remote: error: GH013: Repository rule violations found for refs/heads/dependabot/pip/migrationConsole/lib/console_link/idna-3.15.
  remote: - Cannot update this protected ref.
  remote: - Changes must be made through a pull request.
  remote: - Required status check "DCO" is expected.
  remote: - Required status check "all-ci-checks-pass" is expected.
  To https://github.com/opensearch-project/opensearch-migrations
   ! [remote rejected] dependabot/pip/migrationConsole/lib/console_link/idna-3.15 -> dependabot/pip/migrationConsole/lib/console_link/idna-3.15 (push declined due to repository rule violations)
  ```

  The most recent failure is on #3057 (idna 3.13 → 3.15); the workflow run is [26927000716](https://github.com/opensearch-project/opensearch-migrations/actions/runs/26927000716). The bot isn't a bypass actor on the ruleset, and no token swap can change that.

- Keep the fast path (still attempt direct push first, in case bypass is granted later) but on push rejection open a **replacement PR** from a non-protected branch:

  ```
  bot/pipfile-hash-fix/<suffix-of-dependabot-branch>
  ```

  The replacement PR carries the same dependency bump plus the corrected `_meta.hash` so `all-ci-checks-pass` is green. A sticky comment (deterministic HTML marker) is left on the original Dependabot PR pointing to the replacement, so synchronize events don't spam comments — the comment is updated in place. The replacement PR itself is also reused across runs — if one is already open for the same dependabot branch, only its body is refreshed.

- Adds `pull-requests: write` to the workflow's permission scope so `gh pr create`, `gh pr comment`, and the `PATCH /issues/comments/:id` call succeed.

## Test plan

- [ ] Trigger via the next dependabot Pipfile.lock bump and confirm:
  - direct push attempt is logged (and may either succeed if bypass is granted, or fail and continue)
  - on push failure, a `bot/pipfile-hash-fix/...` PR is opened against `main` with the corrected hash
  - the original dependabot PR gets exactly one sticky comment linking to the replacement, even after multiple `synchronize` events
- [ ] Manually re-trigger the workflow on #3057 and confirm a replacement PR is opened (or update the open Dependabot PR to refresh the trigger)
- [ ] `actionlint` is clean on the modified workflow

### DIFF
--- a/.github/workflows/fix-dependabot-pipfile-hash.yml
+++ b/.github/workflows/fix-dependabot-pipfile-hash.yml
@@ -4,7 +4,15 @@ name: Fix Dependabot Pipfile.lock hash
 # Dependabot generates Pipfile.lock hashes using plette's native method,
 # but our CI uses pipenv 2026.5.0 which computes hashes using PEP 503
 # canonicalized package names. This workflow detects the mismatch and
-# pushes a fixup commit so CI passes.
+# fixes it.
+#
+# The original strategy was to push the fix back to the dependabot PR
+# branch directly. That was blocked when an opensearch-project org
+# ruleset began enforcing PR-only updates on dependabot/** refs:
+#   "Cannot update this protected ref"
+# So now we open a replacement PR from a non-protected branch and
+# comment on the dependabot PR so reviewers can close it in favor of
+# the corrected one.
 
 on:
   pull_request_target:
@@ -14,6 +22,7 @@ on:
 
 permissions:
   contents: write
+  pull-requests: write
 
 jobs:
   fix-pipfile-hash:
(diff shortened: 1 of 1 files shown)


AndreKurait:
Closing in favor of a simpler approach: since the org ruleset now blocks direct pushes to `dependabot/**` branches and the workaround of opening replacement PRs adds complexity for marginal value, the cleanest path is to simply remove this workflow entirely. Follow-up PR removes `.github/workflows/fix-dependabot-pipfile-hash.yml`.

PR #3070, opened 2026-06-05
Fix: Strip workflow-only mTLS fields from CaptureProxy CRD apply

### LINKED ISSUE #3069: Title: upsertcaptureproxyresource fails with strict decoding error — workflow-only fields not stripped before CRD apply
### Describe the bug

When configuring clientAuth on the capture proxy with trustedClientCaPem, the workflow correctly identifies requireClientAuth and sslTrustCertPemEnvVar as workflow-only fields (annotation: migrations.opensearch.org/workflow-only-fields:
"sslTrustCertPem,sslTrustCertPemEnvVar,requireClientAuth"), but the upsertcaptureproxyresource step includes them in the kubectl apply payload. Kubernetes rejects the patch because the CaptureProxy CRD schema does not define these fields and
x-kubernetes-preserve-unknown-fields is not set.

### Error:

strict decoding error: unknown field "spec.requireClientAuth", unknown field "spec.sslTrustCertPemEnvVar"

### Expected behavior

Fields listed in migrations.opensearch.org/workflow-only-fields should be stripped from the CRD resource body before kubectl apply and passed only to the proxy container as args/env vars at deployment time.

### Steps To reproduce

1. Deploy MA 3.3.0 via aws-bootstrap.sh --version 3.3.0
2. Configure workflow with clientAuth on capture proxy:

```
traffic:
  proxies:
    capture:
      source: "source"
      proxyConfig:
        listenPort: 443
        internetFacing: true
        tls:
          mode: certManager
          issuerRef:
            name: migrations-ca
            kind: ClusterIssuer
          dnsNames:
            - capture
            - "*.elb.us-east-1.amazonaws.com"
          clientAuth:
            required: true
            trustedClientCaPem: |
              -----BEGIN CERTIFICATE-----
              ...
              -----END CERTIFICATE-----
```

3. Submit workflow: workflow submit
4. Observe upsertcaptureproxyresource pod in Error state

### Environment

- Migration Assistant: 3.3.0 (Helm chart migration-assistant-3.3.0)
- EKS: Kubernetes 1.35
- Region: us-east-1

### CRD verification

The CaptureProxy CRD spec only accepts these fields:

```
dependsOn, destinationConnectionPoolSize, destinationConnectionPoolTimeout,
enableMSKAuth, internetFacing, kafkaClientId, listenPort, ...
noCapture, numThreads, otelCollectorEndpoint, podReplicas, resources,
serviceType, setHeader, suppressCaptureForHeaderMatch, ...
tls
```

`requireClientAuth` and `sslTrustCertPemEnvVar` are NOT in the CRD schema.

### Workaround

None currently — the workflow is blocked at proxy reconciliation. Removing clientAuth from the config allows the workflow to proceed but without mTLS.

Akhil911:
### Description

Fixes #3069

When `clientAuth` is configured on the capture proxy, the workflow correctly resolves `requireClientAuth`, `sslTrustCertFile`, and `sslTrustCertPemEnvVar` as workflow-only fields (they're passed as container args/env vars, not stored in the CRD). However, `ARGO_PROXY_WORKFLOW_OPTION_KEYS` was missing these fields from its pick list, causing them to be included in the `kubectl apply` payload for the CaptureProxy CRD.

The CRD schema does not define these fields and `x-kubernetes-preserve-unknown-fields` is not set, so Kubernetes rejects the patch with:
```
strict decoding error: unknown field "spec.requireClientAuth", unknown field "spec.sslTrustCertPemEnvVar"
```

### Changes

- Added `sslTrustCertFile`, `sslTrustCertPemEnvVar`, and `requireClientAuth` to `ARGO_PROXY_WORKFLOW_OPTION_KEYS` in `argoSchemas.ts`
- Updated workflow template snapshots to reflect the corrected omit list

### Testing

- All schema tests pass (78/78)
- All config-processor tests pass (55/55)
- All migration-workflow-template tests pass (31/31)
- Verified end-to-end: deployed MA 3.3.0, configured `clientAuth.trustedClientCaPem`, confirmed the error. After this fix, the fields are correctly omitted from the CRD manifest and passed only to the container runtime.

### Related

- Issue: #3069
- Design: `docs/FileBundlesDesign.md` (section: Capture Proxy mTLS Trust Roots)


gregschohn (commented):
The fix in https://github.com/opensearch-project/opensearch-migrations/pull/3071 omits these fields, which removes the fields that we push altogether and folds the number of copies of the TLS data fields back to 1.

gregschohn:
The fix in https://github.com/opensearch-project/opensearch-migrations/pull/3071 omits these fields, which removes the fields that we push altogether and folds the number of copies of the TLS data fields back to 1.

PR #2690, opened 2026-04-13
WIP: Workflow tests

gregschohn:
### Description
This builds upon https://github.com/opensearch-project/opensearch-migrations/pull/2447 to provide some E2E testing for workflows.  It is still a WIP.

[Here's the embedded design](https://github.com/gregschohn/opensearch-migrations/blob/20f35314d2cb73a9e4a637f3b6fe6a8d75d1ec6c/docs/e2eOrchestrationTestDesign.md) for this testing framework.

### Check List
- [ ] New functionality includes testing
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3071, opened 2026-06-05
Fix CaptureProxy mTLS clientAuth failing CRD apply (strict decoding error)

### LINKED ISSUE #3069: Title: upsertcaptureproxyresource fails with strict decoding error — workflow-only fields not stripped before CRD apply

AndreKurait:
### Description

Configuring `clientAuth` on a capture proxy makes the `upsertcaptureproxyresource` step fail with:

```
strict decoding error: unknown field "spec.requireClientAuth", unknown field "spec.sslTrustCertPemEnvVar"
```

**Root cause.** The CaptureProxy CRD is projected from `USER_PROXY_OPTIONS` and does not define the resolved-only fields produced by `clientAuth` lowering (`sslTrustCertFile`, `sslTrustCertPem`, `sslTrustCertPemEnvVar`, `requireClientAuth`, `fileSourceVolumes`, `fileSourceVolumeMounts`). The runtime CR builder `makeCaptureProxyManifest` stripped only `ARGO_PROXY_WORKFLOW_OPTION_KEYS` — a set that intentionally *retains* those flat fields because `setupCapture` reuses it for the Java-args split — so they leaked into `CaptureProxy.spec` and the API server rejected the apply. There are two separate proxy-CR builders; the config-processor one (`captureProxyParameters`) already strips these, but the live Argo path (`makeCaptureProxyManifest`) did not. mTLS has never worked on the capture proxy since it was introduced in #3028 — this is the first real exercise of that path.

**Fix.** Rather than only stripping the fields (which would lose the ability to gate the change), this keeps mTLS configuration gate-able:
- Strip the resolved-only bridge fields (in addition to the workflow-option keys) in `makeCaptureProxyManifest`, so they never reach the CR.
- Stop deleting `clientAuth` from `tls` during lowering, so it rides into the CR's already-`gated` `spec.tls` subtree. Enabling/disabling mTLS, switching trust source, or rotating the PEM/file now correctly trip the `tls` ValidatingAdmissionPolicy and the proxy config checksum. The flat `sslTrustCert*`/`requireClientAuth` fields remain the Deployment/Java-process projection and are still stripped from the CR.

Note: the mTLS config is *not* foreign to the CRD — it rides declaratively under `spec.tls.clientAuth` (which the CRD accepts and gates). The flat `sslTrustCert*`/`requireClientAuth` fields are just the resolved/lowered projection of that subtree for the Java process; they are stripped only because they would be **top-level** `spec.*` fields the CRD does not define, and would duplicate `spec.tls.clientAuth`.

**Key-set hygiene.** The strip set is derived from the schema rather than hand-maintained:
- The fields lowering adds over the user schema are named once as a single shape (`PROXY_RESOLVED_FIELDS`), used for the `USER_PROXY_OPTIONS` → `ARGO_PROXY_OPTIONS` `.extend()`, and `ARGO_PROXY_RESOLVED_ONLY_KEYS` is derived from it. By construction this equals `keys(ARGO_PROXY_OPTIONS) - keys(USER_PROXY_OPTIONS)`, and since the CaptureProxy CRD is projected from `USER_PROXY_OPTIONS`, that delta is exactly the set of top-level keys the CRD does not define — so a newly-added resolved field can't silently drift out of the strip set and re-introduce the leak.
- `makeCaptureProxyManifest` strips a single deduped union (`ARGO_PROXY_CR_OMITTED_KEYS`) of the workflow-option keys and the resolved-only keys; the two sets legitimately overlap (e.g. `sslTrustCertPem`, `fileSource*`), so the dedup keeps the rendered `sprig.omit` listing each key once.

### Issues Resolved

Fixes #3069

### Testing

- New offline conformance test (`captureProxyCrdConformance.test.ts`) renders the real `ResourceManagement` template, parses the rendered `sprig.omit` list, and asserts the runtime CaptureProxy CR builder emits **no** `spec` field the generated CaptureProxy CRD does not define. It fails red on the pre-fix code (leaks `sslTrustCertFile`, `sslTrustCertPemEnvVar`, `requireClientAuth`) and passes after. The pre-existing CR round-trip test missed this because it exercised the separate config-processor builder, not the live Argo path.
- New `argoProxyResolvedKeys.test.ts` locks in the structural invariant: `ARGO_PROXY_RESOLVED_ONLY_KEYS == keys(ARGO_PROXY_OPTIONS) - keys(USER_PROXY_OPTIONS)`, and the deduped CR omit set contains no duplicate keys.
- Updated proxy lowering tests to assert `clientAuth` is retained in `spec.tls` while the flat bridge fields are stripped.
- A complementary live E2E test (a `proxy-clientAuth` mutator + gated spec) is contributed against the workflow-testing framework in #2690.

### Check List


AndreKurait:
Companion live-cluster E2E coverage (a gated `proxy-clientAuth` mutator + spec) is proposed against the workflow-testing framework in #2690: gregschohn/opensearch-migrations#18.

gregschohn (approved):

PR #3072, opened 2026-06-05
Fix infinite create-index retry loop on ES 7.x index-level knn settings

AndreKurait:
### Description

ES 7.x with the AWS k-NN plugin stores knn parameters as **flat index-level settings keys with embedded dots** (e.g. `index.knn.algo_param.m`, `index.knn.algo_param.ef_construction`, `index.knn.space_type`). When migrating to OpenSearch 2.x+/3.x, these keys are rejected as `unknown setting` at create-index time.

`IndexCreator_OS_2_11` strips reported illegal args and retries, but the strip logic only walked nested-object paths — it never matched the flat keys — so the request body was never actually modified and index creation **looped forever**. Observed in a dev migration: a single `knn_vector` index (`retail-products`, ES 7.10 → OS 3.3) produced **220k+ identical create attempts** with no progress, blocking the rest of the migration.

This PR fixes the loop and also stops users from hitting the broken path in the first place:

**1. Robust flat-key removal (`RFS`)**
- `ObjectNodeUtils.removeFieldsByPath` now removes the literal flat key first (e.g. `settings["index.knn.algo_param.m"]`), then falls back to the existing nested-object walk.
- `IndexCreator_OS_2_11.handleInvalidResponse` passes both the full `index.`-prefixed argument (matching the serialized flat keys) **and** the stripped form (for legacy nested settings), so the strip-and-retry path always terminates.

**2. New default metadata transformer `es-knn-index-to-field-level-metadata.js`**
- Lifts index-level knn build params (`m`, `ef_construction`, `space_type`) onto each `knn_vector` field's `method` config — the shape OpenSearch 2.x+ expects — and removes the now-illegal index-level keys. Walks both flat-keyed and nested settings forms, and recurses into nested/object mappings.
- Preserves any pre-existing field-level `method` config, and leaves the valid runtime setting `index.knn.algo_param.ef_search` in place.
- Registered as a baked-in transformer for any source whose target is OpenSearch; it's a no-op when no index-level knn params are present, so it's safe to default-enable.

### Issues Resolved

N/A — found during dev-cluster migration troubleshooting.

### Testing

- `./gradlew :RFS:test` — full suite passes, including new `ObjectNodeUtilsTest` flat-key cases and a new `IndexCreator_OS_2_11Test` case reproducing the exact `retail-products` payload (second create call now succeeds).
- `npm test` in `transformation/standardJavascriptTransforms` — **71 passing** (10 new Jest cases for the transformer + 61 existing).
- `./gradlew :MetadataMigration:test` passes; verified the existing isolated `KnnSettingsTransformationTest` contract (OS 1.3 → OS 2.19) is upheld — `index.knn.algo_param.ef_search` is intentionally **not** stripped.
- Confirmed the new JS file is packaged onto the classpath via `:transformation:standardJavascriptTransforms:packageJs`.

### Check List

### DIFF
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/MetadataTransformationRegistry.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/MetadataTransformationRegistry.java
@@ -57,6 +57,20 @@ public class MetadataTransformationRegistry {
                 .descriptionLine("Convert field data type dense_vector to OpenSearch knn_vector")
                 .build())
             .build(),
+        // Any source whose target is OpenSearch: pre-OS index-level knn params
+        // are rejected on create-index, so lift them onto field-level method
+        // config. No-op when absent (see the js file for the full rationale).
+        TransformerConfigs.builder()
+            .filename("js/es-knn-index-to-field-level-metadata.js")
+            .isRelevantForVersions(andSourceTargetVersionPredicate(
+                    v -> true,
+                    UnboundVersionMatchers.anyOS
+            ))
+            .transformerInfo(Transformers.TransformerInfo.builder()
+                .name("Index-level knn settings to field-level method")
+                .descriptionLine("Lift index.knn.* settings onto knn_vector field method config")
+                .build())
+            .build(),
         TransformerConfigs.builder()
             .filename("js/knn-to-serverless-metadata.js")
             .isRelevantForVersions(andSourceTargetVersionPredicate(
--- a/MetadataMigration/src/test/java/org/opensearch/migrations/EsKnnIndexLevelSettingsMigrationTest.java
+++ b/MetadataMigration/src/test/java/org/opensearch/migrations/EsKnnIndexLevelSettingsMigrationTest.java
@@ -1,0 +1,166 @@
+package org.opensearch.migrations;
+
+import java.io.File;
+import java.util.stream.Stream;
+
+import org.opensearch.migrations.bulkload.framework.SearchClusterContainer;
+import org.opensearch.migrations.bulkload.framework.SearchClusterContainer.ContainerVersion;
+import org.opensearch.migrations.commands.MigrationItemResult;
+import org.opensearch.migrations.snapshot.creation.tracing.SnapshotTestContext;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import lombok.SneakyThrows;
+import lombok.extern.slf4j.Slf4j;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+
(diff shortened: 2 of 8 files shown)


AndreKurait:
Force-pushed an updated revision (rebased onto `main`, so the branch is now mergeable). Changes since the prior revision:

- **Rebased onto `main`** — resolved the `ObjectNodeUtilsTest` conflict by keeping both `main`'s `removeAnalyzerFilters` tests and this PR's flat-key tests.
- **Engine: `lucene` → `faiss`, `space_type` defaults to `l2`** in `es-knn-index-to-field-level-metadata.js`. The ES 7.x source uses the k-NN plugin (nmslib); the existing `knn-nmslib-to-faiss` transform is gated OS→OS so it never fires for an ES source. faiss is valid across OS 1.x–3.x and matches the intended `nmslib → faiss, space defaulting to l2` behavior, so the synthesized method must be self-sufficient.
- **Added an ODFE 1.13.3 (ES 7.10.2 + k-NN, Apache-2.0) → OpenSearch 3.5 Testcontainers E2E test** (`EsKnnIndexLevelSettingsMigrationTest`, `@Tag("isolatedTest")`). It creates a method-less `knn_vector` index with index-level flat `index.knn.algo_param.*` settings — the exact shape that triggered the infinite loop — snapshots, migrates, and asserts the migration completes and the params land as field-level faiss/l2 method config (with `index.knn` and runtime `ef_search` preserved). **Verified passing locally against real containers.**
- **Trimmed comments** to essential rationale across the transform, `ObjectNodeUtils`, `IndexCreator`, and the registry.

Validation: JS 86/86 (97.5% line coverage on the new transform), `ObjectNodeUtilsTest` 17/17, `IndexCreator_OS_2_11Test` 7/7, and the new ODFE→OS3 E2E test green.

AndreKurait:
Updated the engine handling after reviewing the [Lucene-on-Faiss](https://opensearch.org/blog/lucene-on-faiss-powering-opensearchs-high-performance-memory-efficient-vector-search/) work.

**Engine = faiss (default), overridable.** The index-level `index.knn.algo_param.*` / `index.knn.space_type` settings are read only by the **nmslib** engine, which was removed in OS 3.0 — so any source carrying them is an nmslib index, and faiss is the canonical successor (consistent with `knn-nmslib-to-faiss-metadata.js` and the original ask). The one historical reason to prefer the `lucene` engine — faiss's all-in-memory requirement — is what Lucene-on-Faiss removes: partial loading + early termination now run *on a Faiss index* via `index.knn.memory_optimized_search` / on-disk mode, so defaulting to faiss no longer costs the memory advantage. For the legitimate lucene-deployment cases, engine is overridable via `context.engine` (no code change).

**space_type correctness fix.** nmslib accepted `l1`/`linf`; faiss and lucene do not. The transform now normalizes unsupported space types to `l2` with a warning (matching `knn-to-serverless-metadata.js`'s supported set: `l2, innerproduct, cosinesimil`) — otherwise an `l1`/`linf` source would have reproduced the original create-index rejection on a different key.

Validation: JS 92/92 (97.75% line coverage on the transform), and the ODFE 1.13.3 → OS 3.5 E2E test re-run with `--rerun-tasks` (genuine execution, not cached) **PASSED**.

gregschohn (approved):

PR #3080, opened 2026-06-05
Multi-type union default + S3TupleSink shutdown NPE fix

AndreKurait:
This PR bundles two independent changes that landed on the same branch.

---

## 1. Union multi-type indices by default in `TypeMappingsSanitizationTransformer`

Moves legacy Elasticsearch (<7.x) multi-type index/mapping resolution out of the Java `IndexMappingTypeRemoval` version-transformer rule and into the JS `TypeMappingsSanitizationTransformer`, which now **unions** all types into the base index **by default** and **auto-applies**.

### What changed
- **Union by default** — `TypeMappingsSanitizationTransformer.makeContext`'s default `regexMappings` now collapses every type into the base index (`index/<anyType>` → `index`) instead of splitting it (`index_<type>`). Document backfill routes to the same merged index as the metadata side.
- **Auto-applied** — `MetadataTransformationRegistry` emits a `TypeMappingSanitizationTransformerProvider` config (with the source version injected), gated on **source ES ≤6 → target ES 6+ or any OpenSearch**.
- **Composes, doesn't replace** — the TypeMappingSanitization now runs first and composes alongside the baked-in transforms (`MigratorEvaluatorBase.selectTransformer`), so the union still applies when a user supplies their own `transformerConfig`.
- **Removed across the project** — `IndexMappingTypeRemoval`, `IndexMappingTypeRemovalWithMergedSupport`, `MetadataTransformerParams`, and the `--multi-type-behavior` / `multiTypeBehavior` option (Java CLI args, orchestration schema, workflow templates, and sample config).
- Regenerated affected jest snapshots and updated docs.

### Behavior change / compatibility
- Previously, multi-type indices failed by default (`--multi-type-behavior=NONE` → throw). They are now unioned into a single mapping by default for ES ≤6 → ES 6+/OpenSearch migrations.
- **Known gap (out of scope):** `IndexMappingTypeRemovalWithMergedSupport` (ES 1.x–5.4) merged same-`type` fields with differing `copy_to` by unioning the `copy_to` arrays. The JS `mergeFieldProperties` uses strict `deepEquals`, so that narrow edge case now errors instead of merging.

---

## 2. Fix NPE in `S3TupleSink.accept()` after `close()`

Fixes a `NullPointerException` in `S3TupleSink.accept()` when called after `close()`.

### The race
`accept()` runs on Netty event-loop threads. `close()` runs on the main thread (via `ThreadLocalTupleWriter.close()` iterating per-thread sinks). The sequence that crashes:

1. Main thread: `close()` → `closeRequested.set(true)` → `runAndAwaitOnWorker(drain)` → `clearCurrentStream()` sets `gzipOut = null`
2. Event-loop thread (concurrent): `accept()` → serializes JSON → `runOnWorker(lambda)` — the executor hasn't been shut down yet, so this succeeds
3. Worker thread: lambda runs → `gzipOut.write(json)` → **NPE**

### The fix
Two guards, belt-and-suspenders:

1. **Early check on the calling thread**: `if (closeRequested.get())` at the top of `accept()` — fails fast before serialization or enqueue
2. **Null guard in the worker lambda**: `if (gzipOut == null)` — catches the race window between the `closeRequested` check and actual executor rejection

Both paths complete the caller's `CompletableFuture` with `IllegalStateException` rather than crashing.

### Propagation is safe
The failed future propagates through:
- `ThreadLocalTupleWriter.writeTuple()` returns it to the caller
- `TrafficReplayerCore.tryPackageAndWriteTuple()` → `writeFuture.whenComplete()` → `handleTupleWriteCompletion()`
- That method calls `commitTrafficStreams()` regardless of write success — so the Kafka offset advances and the record is not re-delivered

The tuple is "handled" (offset commits) but not persisted to S3, which is the correct behavior during shutdown — the replayer is tearing down and these are trailing records that arrived after the drain decision.

---

## Test plan
- [x] `TypeMappingsSanitization*` unit tests pass (incl. union-default doc-backfill)
- [x] `MetadataTransformationRegistryTest`, `MultiTypeMappingTransformationTest`, `TransformerMapperTest` pass
- [x] `orchestrationSpecs` jest suites pass (snapshots regenerated)
- [x] Existing replayer/tupleSink unit tests pass
- [ ] Manual: call `accept()` after `close()` — future completes exceptionally with `IllegalStateException`, no NPE


MetadataMigration/src/main/java/org/opensearch/migrations/MetadataTransformationRegistry.java
```diff
@@ -136,9 +136,47 @@ public class MetadataTransformationRegistry {
...
+        // Resolve legacy multiple-types-per-index mappings (ES <7) into a single type by
+        // unioning the type mappings. Auto-applied for any source at ES 6 or below migrating
+        // to a target at ES 6 or above (including any OpenSearch). This supersedes the old Java
+        // IndexMappingTypeRemoval rule + --multi-type-behavior flag: the TypeMappingsSanitization
+        // transformer unions by default and runs ahead of the version transformer (see
+        // MigratorEvaluatorBase.selectTransformer), so version rules see single-type mappings.
+        TransformerConfigs.builder()
+            // Sentinel filename: this entry emits its own provider config via providerConfigProvider
+            // (TypeMappingSanitizationTransformerProvider) rather than a JsonJSTransformerProvider.
+            .filename("<TypeMappingSanitizationTransformerProvider>")
```
gregschohn, line 149: why are there brackets around the name?  Why not just skip it?

gregschohn (approved):
I agree with the direction of moving more transforms into js implementations and treating them as user-substitutable.  It feels that this is a pretty gradual move right now, but still will have destabilization on the interfaces.

PR #3092, opened 2026-06-08
Surface snapshot read failures in logs

### LINKED ISSUE #3044: Surface snapshot read failures in logs
When RFS fails to read a snapshot, users currently have no clear, consistent place to find out what went wrong. The failure reason may be buried in pod logs or absent from the `SnapshotMigration` CR status entirely.

When a non-retriable snapshot read failure occurs, the full error, including cause, snapshot path, and any relevant context, must be logged at `ERROR` level in a way that is:

- Visible in the workflow log (Argo / Step Functions execution history).
- Visible in CloudWatch (pod stdout/stderr for EKS deployments).
- Emitted before the process exits so it is not lost if the pod is immediately terminated.

Currently the failure may be swallowed, logged at a low level, or logged only to a pod that is subsequently deleted before the user can inspect it. This must be reliable regardless of how quickly the kill sequence runs after the failure is detected.

**Acceptance Criteria**

 - [ ] A non-retriable snapshot read failure produces an ERROR-level log entry containing the failure reason, snapshot path, and relevant context, visible in both the workflow log and CloudWatch before the process exits.

jzonthemtn:
### Description
Puts snapshot read failures in the logs early, for both Solr and Elasticsearch sources.

### Issues Resolved
Closes #3044

### Testing
Unit tests and integration tests. 

### Check List


jzonthemtn:
Marked as draft while doing testing on EKS.

jzonthemtn:
I modified the S3 location to be an invalid key and this is from the log:

```
Starting Metadata Evaluation
Results:
   Issue(s) detected
Issues:
   Snapshot read failure: Failed to list objects in S3 bucket: migrations-default-371015648411-dev-us-east-1, prefix: my_local_backups/snapshot.main-snapshot/ | snapshot=snapshot.main-snapshot, repo=s3://migrations-default-371015648411-dev-us-east-1/my_local_backups, region=us-east-1
   No source was defined
   No target was defined


time=2026-06-09T13:26:18.388Z level=INFO msg="sub-process exited" argo=true error="exit status 5"
Error: exit status 5
```

image

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/SolrS3SnapshotReadFailureProcessTest.java
```diff
@@ -0,0 +1,336 @@
...
+
+    @Container
+    static final LocalStackContainer LOCAL_STACK = new LocalStackContainer(
+            DockerImageName.parse("localstack/localstack:4.3.0"))
+        .withServices(LocalStackContainer.Service.S3)
+        .withNetwork(NETWORK)
+        .withNetworkAliases(LOCALSTACK_ALIAS);
+
+    @Container
+    static final GenericContainer<?> SOLR_CLOUD = new GenericContainer<>(
```
gregschohn: nit: can you break this out into a test fixture package (we might need to create a new one).  Doesn't need to be a priority.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/RfsMigrateDocumentsTest.java
```diff
@@ -70,4 +73,55 @@ static Stream<Arguments> emitDocTypeCases() {
             Arguments.of(RfsMigrateDocuments.EmitDocTypeMode.AUTO, null, someTransformer, false)
         );
     }
+
+    /** Minimal stand-in so the helper test doesn't couple to any specific production exception. */
+    private static class FakeSnapshotReadFailure extends RuntimeException implements SnapshotReadFailure {
```
gregschohn: these should go be in the RFSCommon directory.  They should be under SnapshotReadFailuresTest

MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java
```diff
@@ -85,6 +86,14 @@ protected void run(String[] args) {
         reportLogPath();
         reportTransformationPath();
 
+        if (result.getExitCode() == SnapshotReadFailures.EXIT_CODE) {
```
gregschohn: nit: the EXIT_CODE values should be managed by each main, since they have other exit codes too, to prevent collision.  They should also be harmonized, but managed in a decoupled way.

MetadataMigration/src/test/java/org/opensearch/migrations/commands/MigrateTest.java
```diff
@@ -41,6 +43,46 @@ void migrate_failsUnexpectedException() {
...
+        var migrate = spy(meta.migrate(args));
+        // A snapshot read failure surfacing during source/snapshot setup, wrapped the way real
+        // callers wrap it (e.g. inside a metadata-read wrapper).
+        var readFailure = new SnapshotRepo.CannotParseRepoFile("corrupt repo metadata: index-0");
+        doThrow(new RuntimeException("reading snapshot failed", readFailure))
+            .when(migrate).createClusters();
+
+        var result = migrate.execute(context);
+
+        assertThat(result.getExitCode(), equalTo(SnapshotReadFailures.EXIT_CODE));
```
gregschohn: do you want to also assert that we don't see a backtrace?  Maybe match the opening line of a java stack trace w/ a regex?

RfsCommon/src/main/java/org/opensearch/migrations/bulkload/common/SnapshotReadFailures.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Exit code emitted when a non-retriable snapshot read failure terminates a process. Shared by
+     * the document-migration worker and the metadata-migration command so a workflow can detect a
+     * snapshot read failure by a single code regardless of which step hit it.
+     */
+    public static final int EXIT_CODE = 5;
+
+    /** Bound the cause-chain walk so a pathological cyclic chain can never loop forever. */
+    private static final int MAX_CAUSE_CHAIN_DEPTH = 100;
```
gregschohn: Can such pathological chain even be created in java? Either way, 100 is pretty deep - we probably SHOULD throw w/ a backtrace then! (of course, we'd possibly fill up disk too when we log the output - :) )

gregschohn (commented):
Here are some other comments from an agent's pass

The main behavior is close, but the patch introduces user-visible regressions around JSON output and over-broad snapshot-read classification.

Full review comments:

- Preserve JSON output when surfacing snapshot errors — /Users/schohn/review/MetadataMigration/src/main/java/org/opensearch/migrations/MetadataMigration.java:86-92
  When `metadata migrate/evaluate --output json` hits a snapshot read failure, this logs a plain-text error to the same `MetadataMigration` console appender after `writeOutput(result.asJsonOutput())`. The JSON result already contains `errorMessage`, so stdout becomes JSON followed by `Non-retriable...`, which breaks callers expecting valid JSON from `--output json`; gate this extra console log to human-readable output or send it somewhere that doesn't corrupt JSON.

- Don't classify unsupported versions as snapshot reads — /Users/schohn/review/SnapshotReader/src/main/java/org/opensearch/migrations/cluster/SnapshotReaderRegistry.java:123-123
  When a live source or user-specified source version has no compatible reader, this registry throws `UnsupportedVersionException`; marking it as `SnapshotReadFailure` makes `classifyFailure`/RFS report exit code 5 and `Non-retriable snapshot read failure` even though no snapshot read was attempted. Workflows that branch on the new code will treat version/configuration problems as corrupt or unreadable snapshots, so this should remain unmarked or use a snapshot-only subtype.

- Prefer the specific snapshot-read cause — /Users/schohn/review/RfsCommon/src/main/java/org/opensearch/migrations/bulkload/common/SnapshotReadFailures.java:33-34
  When an inner read failure is wrapped by another new marker, such as an S3 blob 404 becoming `S3Repo.CouldNotReadFromS3` and then being wrapped as `SnapshotShardUnpacker.CouldNotUnpackShard`, this returns the wrapper before seeing the detailed cause. The emitted `describe()` message then loses the bucket/key or parse error that this change is trying to surface; keep walking and prefer the deepest/specific marker, or avoid marking broad wrappers.

RfsCommon/src/main/java/org/opensearch/migrations/bulkload/common/SnapshotReadFailures.java
jzonthemtn: I might have got a little excessive on that one.

jzonthemtn:
Addressed the individual comments and looked into the others. They were legit enough to warrant changes.

* In MetadataMigration.run(), the snapshot-read-failure surfacing now writes to System.err instead of the stdout-bound console logger.
* Stop classifying unsupported-version errors as snapshot read failures.
* Return deepest snapshot-read marker so specific cause is surfaced.

AndreKurait (approved):
Looks good! Thanks for the improvements here!

PR #3095, opened 2026-06-08
feat(proxy): Improve capture proxy throughput and concurrency handling

Akhil911:
## Description

Resolves capture proxy throughput bottleneck that prevents handling high-concurrency workloads (300+ concurrent connections across 8 pods).

### Changes

1. **`numThreads` default `1` → `0`** — Netty auto-detects thread count based on available processors (`availableProcessors * 2`). With 3.5 vCPU pods, this gives ~7 worker threads/pod instead of 1.

2. **`--reliableCapture` flag** (default: `true`) — When set to `false`, mutating HTTP methods (POST/PUT/DELETE/PATCH) are forwarded to the destination immediately without waiting for Kafka acknowledgment. Capture still happens asynchronously.

3. **`--maxConnections` flag** (default: `0`/disabled) — When set, new connections exceeding the limit receive a `503 Service Unavailable` response immediately, providing back-pressure signaling instead of unbounded queue buildup.

4. **`ConnectionLimitHandler`** — New sharable Netty handler that enforces the connection limit.

All new parameters support env var override via the `CAPTURE_PROXY_` prefix (e.g., `CAPTURE_PROXY_RELIABLE_CAPTURE=false`).

### Testing

- 20 new unit tests covering all new features
- All 61 existing tests pass without modification
- Backward-compatible constructor overloads preserve existing API

### Background

During live traffic testing (~960 msgs/sec, 300 concurrent Lambda invocations), the single-threaded proxy hit its capacity ceiling with zero headroom. Increasing to 350 concurrent connections caused cascading failure (p50 → 30s) within 2 minutes.

### Configuration for high-throughput deployments

```bash
CAPTURE_PROXY_NUM_THREADS=0
CAPTURE_PROXY_RELIABLE_CAPTURE=false
CAPTURE_PROXY_DESTINATION_CONNECTION_POOL_SIZE=25
CAPTURE_PROXY_MAX_CONNECTIONS=500
```

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxy.java
@@ -123,8 +123,9 @@ public class CaptureProxy {
         @Parameter(required = false,
             names = { "--numThreads" },
             arity = 1,
-            description = "How many threads netty should create in its event loop group")
-        public int numThreads = 1;
+            description = "How many threads netty should create in its event loop group. "
+                + "A value of 0 will use the default number of threads (2 * number of available processors).")
+        public int numThreads = 0;
         @Parameter(required = false,
             names = { "--destinationConnectionPoolSize" },
             arity = 1,
--- a/TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxyNumThreadsTest.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/test/java/org/opensearch/migrations/trafficcapture/proxyserver/CaptureProxyNumThreadsTest.java
@@ -1,0 +1,41 @@
+package org.opensearch.migrations.trafficcapture.proxyserver;
+
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+public class CaptureProxyNumThreadsTest {
+
+    @Test
+    public void testNumThreadsDefaultsToZeroForAutoDetect() {
+        var params = CaptureProxy.parseArgs(new String[]{
+            "--destinationUri", "http://localhost:9200",
+            "--listenPort", "80",
+            "--noCapture"
+        });
+        Assertions.assertEquals(0, params.numThreads,
+            "numThreads should default to 0 (auto-detect based on available processors)");
+    }
+
+    @Test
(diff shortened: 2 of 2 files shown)


Akhil911:
Hi Greg,

Thanks for the thoughtful feedback — I've updated the PR based on your guidance.

**What I kept:** The numThreads default change (1 → 0 for auto-detect). This directly resolved our throughput bottleneck — with 3.5 vCPU pods we now get ~7 worker threads per pod instead of 1, which handles our 300+ concurrent Lambda invocation workload comfortably.

**What I removed:** Both --reliableCapture and --maxConnections. Your points made sense:
  - reliableCapture=false breaks the replay contract, so that's a non-starter.
  - Static connection limits are a poor approximation of actual load and don't scale meaningfully across different pod sizes. A user shouldn't need to guess whether they need 100 or 1000.

**On back-pressure (for later):** I spent some time thinking about what a dynamic approach might look like. The options I considered were latency-based circuit breaking, in-flight request limiting (AIMD-style adaptive concurrency), and Kafka producer buffer pressure. Each has tradeoffs — the core challenge is picking a signal that reflects actual system capacity without tightly coupling to a specific bottleneck. I'd love to discuss this with you when the need arises — for now, the 7x thread improvement gives us sufficient headroom.

The updated PR is minimal: one parameter default change + tests. Let me know if there's anything else you'd like me to address. Thanks!

AndreKurait (approved):
The description is a bit out of date, but i'm fine with this as the only change
`numThreads default 1 → 0 `

Here's a summary of the threading model:
```
  Threading model — why this is safe:
  
  The proxy's concurrency model is thread-confined-per-connection: Netty assigns each accepted socket to exactly
  one event loop thread for its lifetime. All handler state (LoggingHttpHandler, FrontsideHandler,
  BacksideHandler, StreamChannelConnectionCaptureSerializer, OrderedStreamLifecyleManager) is instantiated
  per-connection in initChannel and only accessed from that connection's assigned thread — no synchronization
  needed, no races possible.
  
  The only shared resources are:
  - KafkaProducer — documented thread-safe, and sendFullyAsync dispatches to ForkJoinPool.commonPool() anyway
  - BacksideConnectionPool — uses FastThreadLocal, giving each event loop its own isolated
  ExpiringSubstitutableItemPool instance (explicitly lock-free by design) 
  
  The whenComplete callback in ConditionallyReliableLoggingHttpHandler fires on a ForkJoinPool thread after
  Kafka ACK, but Netty Channel operations are safe to call cross-thread (they queue to the owning event loop).
  
  Changing from 1→0 (auto = availableProcessors * 2) simply spreads connections across more event loops. Each
  connection's isolation guarantees are identical. The bottleneck was CPU saturation of a single selector thread
  under high concurrency, not a blocking call.
```

PR #3098, opened 2026-06-09
Fix isolated tests

gregschohn:
### Description

Two fixes for local k8s test infrastructure, plus one for the EKS pipeline.

**EKS: use ECR-mirrored buildkit image** (`eksKubernetesBuildkit.sh`)

`mirror_images_to_ecr` already copies `moby/buildkit:buildx-stable-1` to the ECR mirror, but `BUILDKIT_IMAGE` was never set so the kubernetes buildx driver ignored it and pulled from Docker Hub on every run. Under rate limits or Karpenter cold-start, that raced with buildx's 110s bootstrap timeout. Now sets `BUILDKIT_IMAGE` to the ECR path before `setup_build_backend`.

**Local k8s: fix docker-registry network routing conflict** (`dockerHostedBuildkit.sh`, `k8sLocalDeployment.groovy`, `localTesting.sh`, `kindTesting.sh`)

Connecting any Docker container to a second network fails when it already has a default gateway: Docker tries to program a second default route and the kernel rejects it with "cannot program address … conflicts with existing route". This broke the `docker network connect` between the registry container and the minikube/kind cluster network on every run.

Fix: bind the registry on `0.0.0.0` (all interfaces) instead of `127.0.0.1`, then configure each cluster node to reach it via the host gateway IP rather than a cross-network Docker connection. `connect_cluster_to_registry_network` now resolves the gateway IP from inside each node (`ip route show default`), injects an `/etc/hosts` entry so the hostname resolves without Docker bridge DNS, and writes a `hosts.toml` so containerd accepts plain HTTP on the host-bound port. No `docker network connect` calls remain.

Also adds `teardown_registry_container` and calls it before `minikube delete` and `docker network rm` so the network has no connected containers when it is removed (a connected container silently blocks network deletion). The `registry-data` volume persists across runs so image layers are reused.

### Issues Resolved
Failing k8s local tests: https://migrations.ci.opensearch.org/job/pr-checks/  
Failing EKS isolated VPC test: https://migrations.ci.opensearch.org/job/main/job/main-eks-full-e2e-isolated-vpc-test/

### Check List


AndreKurait (approved):

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -1021,6 +1021,10 @@ if [[ "$build" == "true" && -z "$ma_images_source" ]]; then
     # EKS/bootstrap builds are always Kubernetes-hosted; they cannot assume
     # direct access to a local Docker daemon from the target environment.
     source "${base_dir}/buildImages/backends/eksKubernetesBuildkit.sh"
+    # Use the ECR-mirrored buildkit image so the kubernetes driver doesn't pull
+    # from Docker Hub. mirror_images_to_ecr already copied this image above.
+    ECR_HOST="${MIGRATIONS_ECR_REGISTRY%%/*}"
+    export BUILDKIT_IMAGE="${ECR_HOST}/mirrored/moby/buildkit:buildx-stable-1"
```
AndreKurait, line 1027: it is really hard to not use public images 😅

PR #3106, opened 2026-06-11
spike out adding deleteExistingIndexes option

### LINKED ISSUE #3105: `console clusters clear-indices` fails with 400 when target has `action.destructive_requires_name=true`
## Summary

I am trying to clear out existing target indexes that are NOT managed by the MA `config.json` template.  

I expect `console clusters clear-indices --cluster target --acknowledge-risk` to work, and I expect it to only delete target cluster indexes that it knows about.  So the wildcard pattern actually is a surprise!   If my target cluster has *nyc_taxies*, well that is the only one I expect to see deleted!  So maybe two issues?

`console clusters clear-indices --cluster <name>` always issues a **wildcard** delete
expression (`*,-.*,…`). On any cluster that has the cluster setting
`action.destructive_requires_name=true`, OpenSearch/Elasticsearch rejects wildcard and
`_all` deletes with **HTTP 400**, so `clear-indices` can never succeed on such a cluster —
regardless of `--acknowledge-risk`. There is no flag to make `clear-indices` delete by
explicit name instead.

This is a common configuration (it's the recommended hardening, and is the default in
several managed/secured OpenSearch deployments), so `clear-indices` is effectively unusable
against those targets.

## Environment

- Component: migration console / `console_link` (`console clusters clear-indices`)
- Invoked from: `migration-console-0` pod (`kubectl exec -it migration-console-0 -n ma -- /bin/bash`)
- Target cluster setting: `action.destructive_requires_name: true` (persistent)

## Steps to reproduce

1. Have a target OpenSearch cluster with the destructive-delete guard enabled:
   ```bash
   console clusters curl target -X PUT "/_cluster/settings" \
     -H 'Content-Type: application/json' \
     -d '{"persistent":{"action.destructive_requires_name":true}}'
   ```
2. Run:
   ```bash
   console clusters clear-indices --cluster target --acknowledge-risk
   ```

## Actual behavior

```
Performing clear indices operation...
Error encountered when clearing indices: 400 Client Error: Bad Request for url:
http://<target>:9200/*,-.*,-searchguard*,-sg7*,.migrations_working_state*?ignore_unavailable=true
```

The request deletes by wildcard, which the cluster refuses:

```json
{"persistent":{"action":{"destructive_requires_name":"true"}}}
```

The wildcard does resolve to real indices server-side (so the indices exist and are
deletable by name) — it is only the wildcard *form* of the delete that is rejected:

```
$ console clusters curl target -X GET "/_cat/indices/*,-.*,-searchguard*,-sg7*?h=index"
read_me
top_queries-2026.06.11-14716
```

## Expected behavior

`clear-indices` should succeed (or offer a mode that succeeds) against a cluster with
`action.destructive_requires_name=true`. Possible fixes:

1. **Resolve, then delete by explicit name.** Expand the wildcard via `_cat/indices`
   (the tool already knows how to list them) and issue the delete with the concrete,
   comma-separated index names instead of a wildcard pattern. Named deletes are permitted
   even when the guard is on.
2. **Detect the setting and give an actionable error.** If a wildcard delete returns 400
   and `action.destructive_requires_name=true`, surface a clear message ("target forbids
   wildcard deletes; delete indices by name") instead of a raw `400 Bad Request`.
3. **Add an explicit `--index <name>` / `--indices a,b` option** so callers can target
   specific indices, which also avoids accidentally wiping unrelated indices.

## Workaround

Delete by explicit name, which the guard allows:

```bash
console clusters curl target -X DELETE "/nyc_taxis"
# or several at once:
console clusters curl target -X DELETE "/read_me,top_queries-2026.06.11-14716"
```

(Or temporarily set `action.destructive_requires_name=false`, run `clear-indices`, then set
it back — not recommended for routine use, and it wipes all non-system indices.)

## Additional notes

- A named-delete path is also safer in general: `clear-indices` currently removes **all**
  non-system indices, which is easy to run by accident against the wrong cluster. An
  explicit-index option would reduce blast radius.

epugh:
### Description
Add `"metadataMigrationConfig": { "deleteExistingIndexes": true }` to allow console to delete target index.

### Issues Resolved
Fixes #3105 

### Testing
Unit tests.   Now working on manual test.

### Check List

### DIFF
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/MigrateOrEvaluateArgs.java
@@ -98,6 +98,17 @@ public class MigrateOrEvaluateArgs {
     )
     public boolean allowExistingIndexes = false;
 
+    @Parameter(required = false,
+        names = { "--delete-existing-indexes" },
+        description = "When true, each in-scope index (subject to the index allowlist) is deleted from the " +
+            "target cluster by name before metadata migration recreates it, giving a clean re-run. Deleting by " +
+            "exact name works even when the target has action.destructive_requires_name=true, and only touches " +
+            "indexes being migrated (unlike a wildcard clear). Only takes effect during the migrate phase, not " +
+            "evaluate. When false (default), existing indexes are left untouched. [Destructive]",
+        arity = 1
+    )
+    public boolean deleteExistingIndexes = false;
+
     // Accepted for parity with RfsMigrateDocuments but not used by MetadataMigration.
     // The orchestration layer forwards a shared config bag to both CLIs.
     @Parameter(required = false,
--- a/MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
+++ b/MetadataMigration/src/main/java/org/opensearch/migrations/commands/MigratorEvaluatorBase.java
@@ -9,6 +9,7 @@ import org.opensearch.migrations.MigrateOrEvaluateArgs;
 import org.opensearch.migrations.MigrationMode;
 import org.opensearch.migrations.Version;
 import org.opensearch.migrations.bulkload.common.FilterScheme;
+import org.opensearch.migrations.bulkload.common.OpenSearchClientFactory;
 import org.opensearch.migrations.bulkload.transformers.FanOutCompositeTransformer;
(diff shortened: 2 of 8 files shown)


gregschohn:
Thanks for the issue and PR... 
Please see the comment [here](https://github.com/opensearch-project/opensearch-migrations/issues/3105#issuecomment-4858811149)

I'd like to keep today's metadata migration steps add-only and would rather invest in some better dedicated rollback tools, as per the comment.

PR #3109, opened 2026-06-11
Fix YAML parsing error when trustedClientCaPem contains PEM certificate

### LINKED ISSUE #3108: [BUG] deployProxyWithTls fails with YAML parsing error when trustedClientCaPem contains PEM certificate
### What is the bug?
When using `trustedClientCaPem` (inline PEM) in the capture proxy TLS clientAuth config, the 
`deployProxyWithTls` Argo step fails because the rendered Kubernetes manifest contains 
`-----END CERTIFICATE-----` which kubectl interprets as a YAML document separator (`---`).

  - Source cluster - ES 6.8
  - Target cluster - OS 3.5
  - MA version: 3.3.3
  - Platform: EKS (aws-bootstrap)

  1. Deploy MA 3.3.3
  2. Configure workflow with:
```yaml
  traffic:
    proxies:
      capture:
        proxyConfig:
          tls:
            mode: certManager
            clientAuth:
              trustedClientCaPem: "-----BEGIN CERTIFICATE-----\nMIIF...content...\n-----END CERTIFICATE-----"
```
  
  3. Run workflow submit
  4. The deployProxyWithTls step fails

The PEM content in trustedClientCaPem should be properly escaped/quoted when rendered into the Kubernetes Deployment manifest YAML before kubectl apply.

This is a follow-up to #3069 . Workaround I followed to mitigate the issue - Omit clientAuth from config and patch the proxy deployment post-creation to inject 
` --requireClientAuth` and `--sslTrustCertFile` flags with the CA mounted from a K8s Secret.

Akhil911:
### Description

Fixes #3108.

When using `trustedClientCaPem` (inline PEM) in the capture proxy TLS clientAuth config, the `deployProxyWithTls` Argo step fails because `-----END CERTIFICATE-----` contains `---` which kubectl interprets as a YAML document separator.

### Changes

- **setupCapture.ts**: Base64-encode the PEM value using `expr.toBase64()` before embedding it in the K8s Deployment manifest env var
- **CaptureProxy.java**: `readTrustCertPemFromEnv()` now auto-detects base64-encoded PEM values and decodes them (backward-compatible — raw PEM containing `-----BEGIN` still works as before)
- Updated `readinessProbe.test.ts` to verify toBase64 encoding
- Updated snapshots

### Testing

- All 33 TypeScript tests pass in `migration-workflow-templates`
- Backward-compatible: if the env var already contains raw PEM (e.g. mounted from a file), it is used as-is

### Issues Resolved

- [#3108](https://github.com/opensearch-project/opensearch-migrations/issues/3108)


gregschohn:
Thanks for this.  I suspect that this issue is a lot more widespread.  I'm interested in what it would take to do a yaml-escape so that we didn't have to require every program to decode the base64.  Let me know if that's something that you'd like to look into.  In the meantime, to get unblocked, does putting the cert data in the configmap work?

Akhil911:
Thanks Greg! I've updated the PR based on your feedback. The new revision replaces the base64 approach with a generalized YAML-escaping helper — no consumer-side changes needed.
  
  What changed:
  
  - Added expr.yamlSafeString() to the argo-workflow-builders package (ExprBuilder class). It chains regexReplaceAll calls to escape \, ", \r, and \n — all characters that can break double-quoted YAML scalars after Argo parameter substitution.
  - setupCapture.ts now wraps the PEM value with expr.yamlSafeString(args.sslTrustCertPem) instead of toBase64.
  - Zero changes to CaptureProxy.java — YAML parsing unescapes \\n → \n transparently, so the proxy receives the correct PEM with no decoding logic.
  - The helper is generic and reusable for any future parameter that may contain YAML-special characters.
  
  Tests added:
  
  - Unit tests validating all 4 escape operations are chained in correct order
  - Parity tests that simulate the full pipeline (TS model → YAML render → Argo substitution → YAML parse → value recovery) to verify PEM certificates survive the round-trip — following the side-by-side comparison pattern you mentioned for the argo-builder package.
  
To answer your question about the configmap workaround — yes, that works (it's the trustedClientCaFile path) and I've verified it. But this fix makes the inline PEM path (trustedClientCaPem) work correctly without any workarounds.

AndreKurait:
For this yaml change, i'd like to either have integ tests added or a manual test done
Just to make sure it actually get's unescaped by the time it makes it to the proxy

AndreKurait:
@Akhil911 DCO is also failing

AndreKurait:
@Akhil911 Most recent test failed with error
```
topicConfigChecksum:fad717ea50e04a36)[0].setupProxy[8].setupProxy[4].deployProxyWithTls error: failed to evaluate expression: invalid char escape
```

I won't be able to keep approving these jenkins runs on each change. It's clear this integ test wasn't validated locally. We'll need you to run this integ test locally and post the logs of the run before continuing the review. The instructions for running this test suite locally is in the repo.

Akhil911:
Thanks for the review feedback. Stepping back on the full picture here:
  
**On Greg's YAML-escape question:** I investigated this thoroughly. The transparent YAML-escape approach (where YAML parsing auto-unescapes and consumers see the raw value) is the ideal solution architecturally. However, it's not achievable today because:
  1. The builder's regexReplaceAll is documented as broken at Argo runtime (https://github.com/opensearch-project/opensearch-migrations/blob/main/orchestrationSpecs/packages/argo-workflow-builders/tests/integ/parity/stringOps.parity.test.ts) (argument mapping mismatch)
  2. sprig.regexReplaceAll works at runtime but isn't exposed through the builder API
  3. There's no sprig.replace (non-regex) wrapper available either
  
These are builder infrastructure gaps, not limitations of the approach itself. Fixing the builder's regex support would unblock transparent YAML-escaping for all parameters — not just this one.
  
**Why toBase64 is the right call today:**
  - It's the established pattern in this codebase — setupCapture.ts already uses expr.toBase64() for jsonConfig on the same line, and the proxy consumes it successfully
  - The "N consumers" concern doesn't apply here — readTrustCertPemFromEnv is the sole consumer of this env var. There's no fan-out
  - The decode is backward-compatible: raw PEM (from trustedClientCaFile path or manual env var setting) still works unchanged
  - It's proven at Argo runtime — no expression evaluation risk
  
**Path forward:** I'd suggest filing an issue to expose sprig.regexReplaceAll (or a sprig.replace) in the builder API. Once that's available, we can replace toBase64 with transparent YAML-escaping across all parameters that may contain special characters — making it a one-line change back to the ideal approach. Happy to file that issue if you'd like.
  
**On Andre's integ test concern:** I can't run the K3s integ tests locally (no Docker on Mac/dev desktop). All 160 TS unit tests pass, and I've added Java unit tests covering the decode logic (3 cases: raw PEM passthrough, base64 decode, invalid base64 fallback). Since toBase64 is already running in production for jsonConfig in this exact workflow step, the runtime risk is minimal. Would appreciate one more CI approval to validate end-to-end.

AndreKurait (changes requested):
@Akhil911 we can close this since i've implemented the transparant yaml escaping in https://github.com/opensearch-project/opensearch-migrations/pull/3155 as Greg had originally suggested instead of pushing this as a temporary fix and immediately having to do the right fix for broad application

Akhil911:
Thanks for the confirmation @AndreKurait!

PR #3110, opened 2026-06-12
Manage config edit

gregschohn:
### Description
This adds an Edit ('e') mode to `workflow manage` and also now allows users to Delete ('d') resources from the main page.  Workflows can be submitted in either the edit mode or the status mode with Submit ('s').  There have been numerous iterative refinements to the UX already and it undoubtedly could use more.

Some other changes/bugfixes have also been included, mostly on the developer-experience, like tagging images properly so that one can test multiple workspaces against the same repo.

### Testing
Manual testing - building up a config by hand and making many runs over it to correct and resubmit.

### Check List


docs/manageExternalConfigurationReferencesDesign.md
malliaridis, line 1: Not a fan of these big documentation files. I believe they are generated with an Agent and therefore hard to read and repeating themselves? I found a couple information repeated in docs/manageResourceEditingDesign.md.

.github/workflows/jenkins_tests.yml
```diff
@@ -162,7 +162,7 @@ jobs:
           job_params: "GIT_REPO_URL=${{ needs.sanitize-input.outputs.pr_repo_url }},GIT_BRANCH=${{ needs.sanitize-input.outputs.branch_name }},GIT_COMMIT=${{ needs.sanitize-input.outputs.commit_hash }}"
           jenkins_user: "${{ secrets.JENKINS_MIGRATIONS_USER }}"
           jenkins_api_token: "${{ secrets.JENKINS_MIGRATIONS_API_TOKEN }}"
-          job_timeout_minutes: "120"
+          job_timeout_minutes: "180"
```
malliaridis, line 165: Giving the job 1 more hour seems like we lost a lot of performance here, either it is build or execution performance. I also noticed that while trying the example out, not a great developer experience.

orchestrationSpecs/packages/config-processor/src/index.ts
```diff
@@ -6,3 +6,5 @@ export * from './consoleResources';
 export * from './resolvedMigrationResources';
 export * from './streamSchemaTransformer';
 export * from './fileSourceUtils';
+export {applyEditOperation, applyEditOperationToObject, buildEditStateFromObject, validationForConfig} from './editConfig';
```
malliaridis, line 9: I know that barrel files like this are bad, and that exporting all fields with `*` is also not ideal, but for consistency I would use a wildcard here as well.

Doing so will raise a conflict with `main` function being already exported, which I am not sure if we need these individual main entry points in each file. Are we not following a "gateway" pattern here for the CLI?

malliaridis (changes requested):
The idea in general looks good, and seems to solve some UX problems the vim editor currently has when using `workflow configure edit`.

As recommended I stopped reviewing this PR file by file and try to make sense out of it, and I tried it out via manually testing it in the CLI. You will still fine some code-specific review points from before, but you may ignore these.

### What this PR improves / addresses

The `workflow manage [e]` option addresses specifically the following issues:
- Bad overview in the configuration file
  - This is addressed by providing a way of collapsing configuration sections
  - Comments are not populated in the tree, but in sub-windows that are shown during the editing of specific fields
- With the new manage overview (is this the resource view you were talking about?) it is clear when submitting what changes exactly are submitted

--- 

### What could be addressed before proceeding

- [ ] Extract kafka restructuring (put under `traffic`) to separate PR
  - This is likely a refactoring that should be addressed separately, as it introduced breaking changes
- [ ] Extract addition of new CLI parameters `--quiet` and `--verbose` to separate PR
- [ ] `workflow manage` displays an empty screen without any information. This may confuse users that try it for the first time, but also when they are used to "something being displayed".

### What should be addressed before merging

- [ ] Configuration sections are missing if they are not already present in the configuration file
  - Entire sections like `snapshotMigrationConfigs` are missing, but also sub-sections like `sourceClusters.[source].snapshotInfo.snapshots.[snapshot].config.createSnapshotConfig` or fields like `sourceClusters.[source].snapshotInfo.repos.[repo].awsRegion`
  - Some sections like `snapshotMigrationConfigs` appear after adding specific fields, which is kind of confusing
- [ ] Significant longer build times, about 3 times longer
- [ ] Figure out why existing validation errors during configuration are only displayed when trying to save the changes, and not displayed in the tree
  - Perhaps this is addressed once we fix the issue with the missing sections?
- [ ] Cleanup and simplify the new documentation that is being added
  - The documentation is hard to read and adds little value if it is huge and repetitive
- [ ] When adding a new source the user is not asked about the endpoint, and it is set to empty string without a validation error
  - When adding a new target the endpoint is explicitly requested
- [ ] There is a high possibility there is a validation error during submit, but without any information what is invalid
  - Error message just says "Validation still reports 1 issue. Fix before submit, or press w to save  the draft."
  - Likely addressed after fixing the missing fields not being displayed
- [ ] When adding a valid repository, then a snapshot with `createSnapshotConfig`, once the `repoName` is selected, the repository is rendered invalid
  - **This renders the editor useless, as it is not possible to complete the configuration through it.** It does not allow configuring snapshot migrations this way.
  - This is likely caused by the `localstack://` schema that I am using. It is first accepted and later rejected.

### What could be addressed after this PR is merged

- Provided hints are not always clear what needs to be filled in
	- If the hints are read from the schema, review these and update wherever necessary
	- Provide examples in cases where examples make sense (rule of thumb: always)
- Regex patterns are not human-readable, so they add little value (perhaps they are useful if they are copied to check outside the TUI for matching strings?)
	- If regex checks for valid URI, provide additional information like "a valid URI"
	- If users tend to copy and use the regex outside for testing (e.g. regex101), then provide an easy way to copy regex
	- If regex is not readable, nor used outside TUI, simply hide regex patterns
- Too many symbols are shown around the status, making the `workflow manage` fuzzy
	- Display changes in form of a table with "previous value" column and "new value" column (where first column can be the field that is updated)
- ESC not always available to the user, this requires the user to exit via `q` and re-enter with `workflow manage`
	- ESC can be problematic for some users (e.g. terminals inside IDEs)
	- Consider using different key in place of ESC as it may be problematic for some users (e.g. terminals inside IDEs)
- Colors do not match the standard colors, like red for errors, yellow/orange for warnings, green for success
	- Define color rules and apply correctly across TUI
- Some fields are changed by default and marked as "pending", even though they are using the defaults and not set by the user. This may add unnecessary configurations to the overview, making it harder to read
	- Do not display fields that already use default values in the `workflow manage` overview
- The same error message can be spammed multiple times, e.g. when clicking on "add source snapshot" without defining prior a repository
	- Dismiss messages from the same source/trigger before displaying a new message (is this supported?)
- Branching in schema replaces entire sections if fields like the `sourceType` are changed, but data is also cached
  - This causes a whole change in the displayed data irritating the users and making them wonder where the data / configuration go
  - Avoid branching in schema (undo recent changes of snapshot / backup?) so that existing configuration can be reused no matter what is changed

**What do you think about these improvements as follow-ups?** Let me know if I should create issues for these.

malliaridis:
After updating my workflow configuration to pass validations in `workflow configure edit` using Solr as source and the new `backups` definitions, the `workflow manage` editor complains that:

> Source `solr-source` has no snapshots. Define snapshots under sourceClusters.solr-source.snapshotInfo.snapshots before configuring per-snapshot migrations. [WARN 1]

image

I believe the new schema changes confused the editor from this PR.

.github/workflows/jenkins_tests.yml
malliaridis: This may not be related to this PR, so it can be ignored here and addressed separately.

gregschohn:
In a couple hallway tests, the terminal UI itself was too complicated for users to understand how to navigate and use the interface let alone the larger system.

The middleware work has been continued in https://github.com/opensearch-project/opensearch-migrations/pull/3314.  That PR extends this branch, but replaces the Textualize frontend for a React front-end backed by a FastAPI.

PR #3118, opened 2026-06-14
migration console tui adjustments

awagen:
### Description
The draft PR contains code changes for:
- approval gate addition for backfill and proxy setup
- addition of next command hint for commands in the migration console

Rather than finalized PR this represents a state for discussion of the approach as discussed with @gregschohn (thorough testing pending).
One aspect wed also need to cover is the granularity of hints, which right now seems a bit coarse. 

Thoughts Id like to add:
would it be beneficial to structure (and configure centrally with clear reference where configurations / workflow definitions are pulled from) steps a bit more around single resources that can either be defined as existing resources or steps that create those resources, where each step is centrally represented as basically a function (or a config with dependencies and output type) where workflow is a sequence of steps.
Then a possible step by step config guidance within TUI and additions within the workflow could simply resolve the dependency, suggest steps for creation of missing resources and config for existing resources to build step by step.
Its possible this is more informed by the flow of the full migration local tests than migration reality though
in which case resource setups and configs are partially separated.

### Check List


migrationConsole/lib/console_link/console_link/workflow/commands/submit.py
```diff
@@ -42,11 +43,15 @@ def _handle_workflow_wait(
...
 
+        return phase
+
     except TimeoutError as e:
         click.echo(f"\n{str(e)}", err=True)
         click.echo(f"Workflow {workflow_name} is still running", err=True)
+        return 'Running'
     except Exception as e:
         click.echo(f"\nError monitoring workflow: {str(e)}", err=True)
+        return ''
```
gregschohn: Why should we return an empty string here?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/setupCapture.ts
```diff
@@ -856,6 +856,15 @@ export const SetupCapture = WorkflowBuilder.create({
                     managedByWorkflow
                 )}) }
             )
+            .addStep("approveProxySetup", ResourceManagement, "waitForUserApproval", c =>
```
gregschohn: I was expecting that this step would be after the proxy was completely configured.
I'd expect many users would then swing their client traffic through the proxy.  Once that's done, they approve of the step & the migration moves onto backfill.

migrationConsole/APPROVAL_GATES.md
gregschohn, line 1: I presume that these md files were mostly to help you build out the rest of the code.  I have a couple question on the APPROVAL_GATES, but in general, don't include these files as they are - or any files like this in migrationConsole.

I _DO_ think there are gaps in the documentation (MigrationAsAWorkflow.md and reconfiguringWorkflow.md) where a lot of this information could go.  Could you work some of this into those documents and anywhere else within the docs/ folder that needs some more clarity.  

My hope is that a user can point an agent to the docs directory and create any specific notes about the system that a dev/user might need to be concerned with.

migrationConsole/APPROVAL_GATES.md
```diff
@@ -0,0 +1,465 @@
...
+
+The TUI's gate discovery (`tree_utils.py`) works in two passes:
+
+1. Walk the Argo Workflow node tree for nodes whose `templateName` is `waitforuserapproval`.
+2. Query the matching `ApprovalGate` CRDs from Kubernetes and overlay their status.
+
+If you only create `ApprovalGate` CRDs with `kubectl apply` — without a running workflow
+that has `waitforuserapproval` nodes — the TUI has nothing to display.
+
+### `devApprovalGateTest.yaml`
```
gregschohn: I have 2 questions - what are you trying to prove/disprove with the tests?  Do we have other tests and patterns that already get us suitable coverage?

I think that the answer is yes if we consider the integ tests.  Sure, we'll have to wait a bit for the workflow to reach the gates.  But - on the other hand, we have integ tests that run all of the stages anyway, so inserting the approvals within those existing integ tests could cover these very stable (& closely related) features.  Do you think that that would be adequate?  My issue with the current test is that it's very dependent upon the implementation details of the approvals themselves.

gregschohn (commented):
Thanks @awagen - This looks like it's on the right path.  I left some nits around approvals.  Let me know if you think that you agree.

migrationConsole/lib/console_link/console_link/workflow/commands/submit.py
awagen: Not intented. Will be fixed to reflect error state and handling in hints.

migrationConsole/APPROVAL_GATES.md
awagen: Yes, the file was not intended to make it to a final PR. This serves as temporary overview also since we will have a little handover of this during my holiday so for the moment that includes a bit more context. We will distribute relevant info into the respective docs.

migrationConsole/APPROVAL_GATES.md
awagen: This was mainly as temporary test to have quick step subs to validate the sequence of events rather than needing to wait for workflow to reach state (and in the current form not intended to land in the final PR). Agree that it makes sense as part of the end to end test. One part that I find suboptimal during development is that such step logic and quick validations are baked into more complex flows that, as you mention, require quite a wait for validating a detail. I feel the feedback loop here can be improved.

gregschohn: In my development workflows, I usually work off of one of the (poorly located) config-processor/scripts/samples workflow configurations.  I install the testClusters chart with default values.  When I submit - usually manually, via the sequence listed at the bottom of orchestrationSpecs/README.md (which means I need to update the IP address of localstack on my host or config file if I run from outside the cluster).

With empty clusters, those don't take too long & you'll see a number of workflows there for different variants - some run only backfill, some set up the proxy different ways.  This is where I put configs for manual tests.  

There are also parity tests (orchestrationSpecs/packages/argo-workflow-builders/tests/integ/parity/...) that let you put argo workflows together simultaneously with the direct argo-yaml expressions and our modeled type-script.  In this case, I wouldn't use those tests, but I do when I'm filling a gap in the argo-workflow-builders or trying to figure out a new pattern.

Let me know if you think that those two approaches help.  I'm happy to entertain more or for us to surface documentation to advertise them better to devs.

awagen: These lived practices are gold, @gregschohn , thank you for pointing out. Surfacing best practices in this regard makes sense imo. It's easy to land in a certain mode of development process that is suboptimal in terms of fast feedback, makes sense to emphasize those in the docs. Happy to contribute (well, after PTOs :)).

migrationConsole/lib/console_link/console_link/workflow/commands/submit.py
wrigleyDan: addressed in https://github.com/opensearch-project/opensearch-migrations/pull/3118/changes/cdf1f888f1e02aecabd53d383267705a02757fcd

migrationConsole/APPROVAL_GATES.md
wrigleyDan: Addressed in https://github.com/opensearch-project/opensearch-migrations/pull/3118/changes/37045e9b600c34fe5c27936a6ef0c12b396385d8

Removed scratch files and updated existing documentation files.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/setupCapture.ts
wrigleyDan: addressed in https://github.com/opensearch-project/opensearch-migrations/pull/3118/commits/96396cbf26b8c6e82469d5a41ef23061eddebc19

gregschohn (commented):
I made a couple small modifications in https://github.com/opensearch-project/opensearch-migrations/pull/3189.

Let me know what you think of those.  If you think that they're OK & want to accept them, you can pull them into this branch & I can approve/merge.

migrationConsole/APPROVAL_GATES.md
malliaridis: Linking this to #3183 as it could potentially solve some parts of the documentation request if addressed.

PR #3119, opened 2026-06-15
Load Testing the Traffic Capture & Replay Pipeline: Background and Problem Statement

wrigleyDan:
### Description
Add a document describing the background and problem statement for Traffic Capture & Replay load testing.

Update 15.07.26:
Changed draft to PR with initial scope on the load testing setup itself, while further aspects will be addressed in follow up PRs. 

Notes:
- This PR, besides the abovementioned document, provides Initial standalone variant based on k6 with local (not yet k8s / Minikube / kind) setup, distinct scenarios and two examples of index / document formats
- Follow up PR will address local k8 integration (Minikube) and cloud setup integration
- Part of the variants (e.g mixed profile) utilize a redis-based buffer to track ingested docs to be able to retrieve them. If deemed useful for the current tests, we can bake this into a custom k6s image that comes with it on the given node. The current variant assumes a central redis to enable multiple k6 instances
- While k6 is configured via ts files, the structure is particular to k6, meaning it does not provided the capability to add arbitrary libraries such as a redis client. To go around it in cases where redis is utilized a webdis instance is added which allows http communication to the redis instance
- Currently there are a bunch of scripts running the different scenarios and checking a range of metrics. These are likely temporary and more for illustration purpose atm.
- As per the metric collection, the existing scheme of the current k8s setup will be kept, e.g collector per node via daemonset, then either scraped by prometheus or exported to cloud watch. For managed Kafka these would be available in cloud watch, while for custom kafka deployments the mode would be scraping admin endpoints by the otel-collector. The implementation of this will be contained in above mentioned follow-up PRs.

### Issues Resolved
n/a

### Testing
n/a

### Check List


gregschohn (commented):
This document captures the context well.  I'd like to see it start to move ahead w/ some POCs for traffic generation.

A couple of high-level comments are
1) Don't worry about solr specifics - pick some http servers that accept traffic that can include requests and responses that are up to 100MB large; supports some type of data updates; has persistence of identifiers (users or the server can create ids) that can be used later; and lastly, one that can handle a LOT of traffic.  This work eventually dovetails to the validation story, so keeping it ES->OS might make a lot of sense, but OS->OS or a test server specifically for proxy testing could make sense too (if one exists).  You have some freedom there.
2) Appendix A's items are good callouts.  We should plan to exercise those - but they won't work with the current system.  We'll want to include those as optional so that we can do the features to support them with a test-driven process.

For next steps - I'd be interested in getting some traffic generated (that can be fanned out to multiple senders) - and then start to figure out what other pieces we need to confirm that things are working well.  Some of that will be in the test infrastructure, but we'll probably find some other features in the core application that we'll need too.  

We can start small and have a series of PRs build this out

wrigleyDan:
@awagen can you take this forward?
Thanks!

TrafficCapture/trafficLoadTest/scripts/run_test_chaos.sh
gregschohn, line 1: Why are these files called "validate"?  The validation story right now is just on thresholds of a couple really simple metrics, right?  Validation will probably eventually eclipse everything else in these load tests, so I'd like to keep the word validate around for something that _only_ has validation as a responsibility.  Would these be just as validly named if they were "run" or "run_test"?

gregschohn (approved):
I have two naming comments and a k8s over docker-compose comment.

On the naming - can you root the loadTestTrafficGenerator to TrafficCapture/trafficLoadTest (it isn't JUST going to be a generator, right).  For the other, please see the validate comment.

The k8s comment isn't blocking, but please map a change to migrate the docker-compose configs into k8s.  The testClusters chart could be extended to pick up these configs & the k6 operator.  That would let you hardcode names like "capture-proxy:9200" within a single chart too.  

I want to test our EKS deployments that users would be running at scale with multiple test clients.  Now that you know how k6 works and we have the patterns established, I don't see value in keeping the compose files around the repo because we need the orchestration in k8s too.  That can be a fast follow - and it's good to get the docker stuff into git history anyway for posterity

TrafficCapture/trafficLoadTest/scripts/run_test_chaos.sh
awagen: Yes, these are not serious validations in that sense, they are just example runs. So definately run_* will do. Id also see those as temporary to give an overview of test runs, but expect them to become obsolete shortly.

awagen: As per the moving to k8s, this is already planned as quick follow-up iteration on the initial setup, which will then also incorporate the setup consolidation.

PR #3132, opened 2026-06-16
Unify Solr snapshot: single CreateSnapshot entry point with mode flag

AndreKurait:
## Description

Refactors the Solr snapshot flow so that **both standalone and SolrCloud clusters always invoke `CreateSnapshot`** with a `--mode` flag (`create` or `import`), unifying what was previously split across multiple code paths.

### Key Changes

**`SolrBackupStrategy.java`** — unified entry point:
- Shared steps run in both modes: collection auto-discovery → config file check-and-upload to S3 → mode-specific branching
- **CREATE mode**: performs full snapshot backup (Collections API for SolrCloud, replication API for standalone)
- **IMPORT mode**: validates external snapshot accessibility without performing a backup; config files are still fetched and uploaded
- Config retrieval uses Admin File Handler with Schema API fallback (handles all Solr versions)
- Standalone filesystem config synthesis creates the `zk_backup_0/configs/` structure that `SolrSchemaXmlParser` expects

**`snapshot.py`** — always passes `--mode` for Solr sources (defaults to `create`)

**`SnapshotMode.java`** — enum with `fromString()` factory for CLI parsing

**`CreateSnapshot.java`** — `--mode` CLI parameter with validation

### Motivation

Previously, standalone Solr required a separate orchestration path. This unification means:
1. `CreateSnapshot` is the **single entry point** for all Solr snapshot operations
2. Config files are **always verified/uploaded** regardless of mode — no more missing schemas in downstream metadata migration
3. Standalone clusters get first-class support through the same unified flow

### Testing

- Unit tests: `TestCreateSnapshotModeFlag` (mode parsing, validation, defaults)
- Integration tests: `TestCreateSnapshotSolr`, `TestCreateSnapshotSolrS3` (Docker/Testcontainers, `@Tag("isolatedTest")`)
- `SolrReader` (119 tests), `RFS` (433 tests) — no regressions

### Checklist
- [x] New functionality has been documented (Javadoc on all public methods)
- [x] Commits are signed (`Signed-off-by`)


AndreKurait:
## Converged onto PR #3092's shared test-container interface

Updated this PR to adopt the `SolrClusterContainer` test-fixture abstraction introduced by #3092, so the two merge cleanly.

**What changed**
- **`SolrClusterContainer` (testFixtures)** now owns S3-backup container construction. Added a public `Mode {CLOUD, STANDALONE}` enum and a general `withS3Backup(version, mode, network, coreOrCollection, bucket, region, alias)` factory that encapsulates every version/mode launch detail (command, env, wait strategy, **Solr 8 contrib-jar vs Solr 9 `SOLR_MODULES`** layout). `cloudWithS3Backup(version, network, bucket, region, alias)` — #3092's exact signature — is provided as a thin `CLOUD` delegate. Added a `SOLR_8_10` (8.10.1) constant (earliest S3-capable Solr).
- **Moved `solr-s3-backup.xml`** into `SolrReader/src/testFixtures/resources/` (matching #3092's rename), with a `${sharedLib}` default so one resource serves both Solr 8 (contrib path) and Solr 9 (modules path). Both `MetadataMigration` and `CreateSnapshot` already depend on `testFixtures(project(':SolrReader'))`, so it's on their classpath.
- **`SolrS3MetadataMigrationTest`** replaces its inline `startSolr()`/`GenericContainer` construction with `SolrClusterContainer.withS3Backup(...)`, parameterized over the fixture's `SolrVersion`/`Mode` types and using `getSolrUrl()`. ~63 lines of container plumbing moved into the shared fixture.

**Merge note for #3092:** my `cloudWithS3Backup` and #3092's are functionally identical (mine delegates to the general factory). The only likely textual overlap is that method + the `Network`/`MountableFile` imports — a trivial resolution (keep either body). The resource move is recorded as a git rename, same destination as #3092.

**Verified (testcontainers, isolatedTest):**
- `SolrS3MetadataMigrationTest` — 7/7 (8.10.1 / 8.11.4 / 9.8.1 × cloud/standalone + incremental `zk_backup_1`)
- `TestCreateSnapshotSolr` — 12/12 (6/7/8/9 × standalone/cloud + detection)
- `TestCreateSnapshotSolrS3` — green (validates the resource move resolves cross-module)
- `SolrStandaloneBackupCreatorTest` — 17/17

AndreKurait:
## Verified the #3092 merge by trial-merging (and fixed a real clash)

I trial-merged this branch against #3092's head (`o19s:3044-solr-snapshot-read-failures`) rather than assume it was clean — and found that my earlier "keep either body" note was **wrong**. Both branches added a `cloudWithS3Backup(SolrVersion, Network, String, String, String)`; git applied both hunks without a textual conflict, but the result had **two identical methods → a compile error** a textual merge does not surface.

**Fix (pushed):** dropped my `cloudWithS3Backup` delegate. This branch now exposes only the general `withS3Backup(version, mode, network, coreOrCollection, bucket, region, alias)` factory (the Solr 8 cloud case is `withS3Backup(version, Mode.CLOUD, ...)`); nothing on this branch referenced `cloudWithS3Backup`. After #3092 lands, its `cloudWithS3Backup` is the sole copy and complements `withS3Backup`.

**Verified on the merged tree (this branch + #3092):**
- `SolrClusterContainer` auto-merges with exactly one `cloudWithS3Backup` (#3092's) + one `withS3Backup` (this PR's). The two private resource-name constants (`S3_BACKUP_SOLR_XML`, `S3_REPOSITORY_RESOURCE`, same value) coexist legally.
- Full compile of the merged tree (testFixtures + both test modules + main) — **passes**.
- Merged-tree unit tests — `SolrReader:test` (my Solr 9 parser fix + #3092's `SolrBackupSourceTest`) and `RfsCommon:test` (#3092's `SnapshotReadFailuresTest`) — **pass**. My production changes (`SolrBackupStrategy`, `SolrStandaloneBackupCreator`) and #3092's (`SnapshotReadFailure`, `S3Repo.CouldNotReadFromS3`) don't interact.

**Remaining merge resolution** (one file, mechanical): `SolrS3MetadataMigrationTest.java` — take this PR's parameterized version. It's a strict superset of #3092's single `solrCloudS3BackupProducesCorrectOpenSearchMappings()` (same five mapping assertions + exit-code check, now also covering 8.10/9 and standalone), so #3092's single method and its static `SOLR_CLOUD` field are intentionally superseded. No assertions are lost.

CreateSnapshot/src/test/java/org/opensearch/migrations/TestCreateSnapshotSolr.java
```diff
@@ -72,63 +97,95 @@ public void testIsSolrCloud_cloud_returnsTrue() {
...
+                    "-z", "localhost:9983");
+                if (up.getExitCode() != 0) {
+                    throw new IllegalStateException("zk upconfig failed: " + up.getStderr());
+                }
+                createConfig = "&collection.configName=basic";
+            }
+
+            solr.execInContainer("curl", "-s",
+                "http://localhost:8983/solr/admin/collections?action=CREATE"
+                    + "&name=cloudcoll&numShards=1&replicationFactor=1" + createConfig + "&wt=json");
```
malliaridis, line 157: It's probably worth adding similar test with 2+ shards and 2+ replication factor, as it has an impact in the directory naming conventions.

migrationConsole/lib/integ_test/integ_test/test_cases/solr_tests.py
```diff
@@ -36,6 +37,13 @@
     (SolrV6_X, OpensearchV3_X)
 ]
 
+SOLR_IMPORT_ALLOW_COMBINATIONS = [
+    (SolrV9_X, OpensearchV2_X),
+    (SolrV9_X, OpensearchV3_X),
+    (SolrV8_X, OpensearchV2_X),
+    (SolrV8_X, OpensearchV3_X),
+]
+
```
malliaridis, line 46: Why are v6 and v7 not included here?

And how is this list different from `SOLR_ALLOW_COMBINATIONS`? Is one for `create` and one for `import` explicitly? Or is the first one used differently too?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -798,6 +802,9 @@ export const USER_CREATE_SNAPSHOT_WORKFLOW_OPTIONS = z.object({
 export const USER_CREATE_SNAPSHOT_PROCESS_OPTIONS = z.object({
     otelTraceCollectorEndpoint: OTEL_TRACE_COLLECTOR_ENDPOINT,
     otelMetricsCollectorEndpoint: OTEL_METRICS_COLLECTOR_ENDPOINT,
+    mode: z.enum(["create", "import"]).default("create").optional()
+        .describe("Snapshot mode used by the workflow-generated CreateSnapshot step. User configs should leave this as 'create' (default). For Solr externally-managed snapshots, use externallyManagedSnapshotName with importConfig; the workflow will inject 'import' internally."),
```
malliaridis, line 806: Based on the description,
- users should use by default `create`
- when users already define `externallyManagedSnapshotName` it could default to `import` automatically

Are there any cases where none of these two apply? If not, the `--mode` could be internal and derived from the rest of the workflow configuration, reducing the required user input by an important, source-specific parameter, allowing less input errors and causing simpler configurations.

SolrReader/src/testFixtures/java/org/opensearch/migrations/bulkload/solr/framework/SolrClusterContainer.java
```diff
@@ -123,6 +125,143 @@ public void close() {
...
+     * @param mode             {@link Mode#CLOUD} or {@link Mode#STANDALONE}
+     * @param network          shared docker network (must reach the LocalStack alias)
+     * @param coreOrCollection core name to precreate in STANDALONE mode (the wait strategy looks for
+     *                         it); ignored in CLOUD mode
+     * @param bucketName       S3 bucket name
+     * @param region           S3 region
+     * @param localstackAlias  LocalStack network alias; S3 endpoint becomes {@code http://<alias>:4566}
+     */
+    @SuppressWarnings("resource")
+    public static SolrClusterContainer withS3Backup(
```
malliaridis, line 187: Is there any guard this method is not called with Solr Version lower than 8.10? In the method documentation this limitation is not mentioned, and in the class there are Solr version constants below 8.10 that may give false impressions.

jzonthemtn:
I compared this against #3147 with respect to the solr collection naming.

- #3147 removes configuration - the Solr collection/index name is now always derived from the backup.
- This pull request adds configuration and structure. A `CreateSnapshot` entry point with
`--mode create|import`, plus `solrCollections` and an step that can upload a schema (and currently requires the live Solr to be reachable).

The two pull requests collide in `userSchemas.ts` and the Solr backup path. But I think the conflict/choice between the two comes down to, does the tool learn about a backup by inferring from the artifact (#3147), or by being told the config + live source (this PR)?

(#3147 also removes the need for the Solr reshaping scripts prior to migrating. The Solr collection naming is just part of it.)

malliaridis (commented):
It seems that this PR breaks the "source-independent" configuration that applies to all sources / targets and follows a more source-specific configuration. This would make it harder for users to figure out which params are needed and used. Specifically,

- there are source-specific parameters `solrCollectionNames` and `--mode` that are introduced in the snapshot creation config, which could eventually be avoided by using source-independent parameters or derive the missing information from other configuration info
- there is a naming similarity of `externalSnapshotName`, `importExternalSnapshotName` and `externallyManagedSnapshotName` which may confuse users during configuration, and could likely refer to the same thing / element

In PR (#3187) we also decided to avoid adding a solr-specific param for simplicity.

Filtering the collections may be a out of this PR's scope, and we could consider approaching this "requirement" differently, for example by defining a workflow-scoped index list that applies to all steps. This would allow us to merge `externalSnapshotName` and `externallyManagedSnapshotName`, as well as using the `indexAllowList` in the workflow context, allowing the snapshot creation and RFS to maintain the indexes to migrate across the workflow (which is currently not the case).

This change would reduce complexity, simplify the configuration, reduces potential input errors and abstract away individual workflow step configurations, making everything more stable.

For the missing information of Solr (and ES), it we could also have a configuration field like "fetch or generate missing snapshot information via API", lets call it `fetchMissingSnapshotDataFromApi: boolean`, or complementary `generateMissingSnapshots`, which could be used for generating
- missing snapshots (e.g. only 2 out of 3 indexes that were mentioned exist in the backup)
- missing snapshot configurations (e.g. in Solr standalone the missing metadata)

That way we stay source-independent in the workflow configuration, simplify the configuration fields (less fields to set by the users), and automate more steps inernally without user intervention.

This feedback was created without #3209 in mind, which likely addreses some of these concerns.

_A small side note: we may have to clarify project-wide what a "Snapshot" is in terms of our migration assistant, as a snapshot or backup may be considered something else in the Solr context than in ElasticSearch or OpenSearch context, and therefore the user expectations during snapshot creation or RFS may differ._

malliaridis:
@gregschohn here are some late feedback points that may got lost.

PR #3155, opened 2026-06-19
Harden workflow manifests, readiness, and EKS metric checks

### LINKED ISSUE #3108: [BUG] deployProxyWithTls fails with YAML parsing error when trustedClientCaPem contains PEM certificate

AndreKurait:
## Summary

Harden Argo workflow manifest rendering, migration-service readiness, and the EKS workflow app-metric assertion.

This PR fixes #3108 by making dynamic `resource.manifest` string values safe across Argo's text-substitution boundary, then adds the operational hardening uncovered while validating CDC on Kubernetes/EKS: Kafka broker resource defaults, `minPodReplicas`-backed PodDisruptionBudgets, a replayer readiness gate that waits for the capture proxy to actually serve traffic, and CloudWatch metric-check hardening for backfill-only EKS runs.

Closes #3108.

### YAML-safe resource manifests

Argo substitutes workflow parameters into `resource.manifest` before `kubectl` parses the YAML. This PR makes that boundary explicit:

- Dynamic string scalars in resource manifests render through a YAML-safe path so PEMs, quotes, newlines, document separators, booleans, numbers, and mapping-like strings round-trip as strings.
- Dynamic numbers, booleans, arrays, and objects continue to render as direct YAML values.
- Known-safe string expressions can opt into direct YAML rendering where needed.
- Type-level checks reject raw string expressions on the non-string direct rendering path.
- Raw YAML marker handling now emits unquoted Argo placeholders deterministically instead of relying on string stripping.

This fixes inline capture-proxy client CA PEM handling for `tls.clientAuth.trustedClientCaPem` without requiring consumer-side decoding.

### Argo expression and parity fixes

The builder's regex helpers now match real Argo behavior:

- Sprig regex functions are rendered under the `sprig.` namespace.
- `regexReplaceAll` argument ordering matches Sprig.
- String literal escaping handles regex patterns correctly.
- Previously skipped or known-broken parity cases are restored.

### CDC PEM end-to-end coverage

Adds a CDC client-auth PEM test path that parameterizes the shared CDC overlay and verifies the deployed capture proxy receives the original PEM value intact.

### Kafka broker scheduling defaults

Adds `DEFAULT_RESOURCES.KAFKA_BROKER` with 1 CPU / 2Gi defaults and merges it under `KafkaNodePool.spec`, so broker pods avoid `BestEffort` scheduling while user-supplied `nodePoolSpec` values still take precedence.

### Migration service disruption controls

Adds shared scalable-workload schema and workflow-template helpers for `podReplicas` and `minPodReplicas`.

`minPodReplicas` is validated as a nonnegative integer no greater than `podReplicas`, and workflow templates now render `policy/v1` PodDisruptionBudgets for:

- capture proxy
- traffic replayer
- RFS workers
- dedicated RFS coordinator

The default `minPodReplicas` remains `0` so existing single-replica and single-node development clusters keep their current drain behavior.

### Capture proxy and replayer sequencing

Splits `CapturedTraffic.status.checksumForReplayer` readiness across the two setup phases:

- Phase 1 patches `CapturedTraffic` ready for snapshot/topic purposes but leaves `checksumForReplayer` empty.
- Phase 2, after the capture proxy endpoint is confirmed ready, re-patches `CapturedTraffic` with `checksumForReplayer`.

That makes the replayer wait for the same proxy readiness boundary that snapshots already depend on. S3 captured-traffic sources keep their existing loader-owned readiness path.

### EKS CloudWatch metric assertion hardening

Jenkins `pr-eks-integ-test/1039` failed after both `Test0001` and `Test0002` workflows succeeded and target documents were verified. The failure was the post-test CloudWatch app-metric assertion: it used one non-paginated `ListMetrics` response and queried only the first dimension set matching `qualifier`/`OTelLib`. In the shared `OpenSearchMigrations` namespace, that can miss the run's actual time series or stop on a stale matching dimension set.

Backfill-only EKS tests were also checking `bytesSent` before any deterministic RFS completion counter. The assertion now applies CloudWatch server-side dimension filters for `qualifier` and `OTelLib`, still paginates `ListMetrics`, scans all matching dimension sets until one has recent data, and checks `pipelineDocsMigrated`/`pipelineBytesMigrated` before the lower-level `bytesSent` fallback. Because `ListMetrics` discovery for new dimension sets can lag behind metric ingestion, the bounded poll window is extended from about 4 minutes to 15 minutes. The log and assertion wording now refer to workflow app metrics because this check covers RFS-only tests as well as capture/replay flows.

## Validation

Coverage added or updated for dynamic manifest scalar rendering, YAML-sensitive string round trips, type-level misuse checks, workflow-template output changes, `minPodReplicas` schema validation, CDC PEM env-var handling, and CloudWatch metric lookup pagination/dimension scanning.

Local verification for this PR includes:

- `npm -w @opensearch-migrations/argo-workflow-builders run type-check`
- `npm -w @opensearch-migrations/migration-workflow-templates run type-check`
- `npm -w @opensearch-migrations/argo-workflow-builders test -- resourceExpressionYaml.test.ts expression.type.test.ts --runInBand`
- `npm -w @opensearch-migrations/migration-workflow-templates test -- outputMatch.test.ts --runInBand`
- `npm -w @opensearch-migrations/schemas test`
- `npm -w @opensearch-migrations/config-processor test -- --runInBand`
- `npm run -w @opensearch-migrations/migration-workflow-templates type-check`
- `npm run -w @opensearch-migrations/migration-workflow-templates test -- --runInBand`
- `PYTHONPATH=migrationConsole/lib/integ_test:migrationConsole/lib/console_link pytest migrationConsole/lib/integ_test/tests/test_metric_operations.py -q`
- `PYTHONPATH=migrationConsole/lib/integ_test:migrationConsole/lib/console_link pytest migrationConsole/lib/integ_test/tests -q`

EKS failure analysis:

- Reviewed Jenkins `pr-eks-integ-test/1039`: `Test0001` and `Test0002` completed their workflows and verified the target documents, then failed only on the CloudWatch workflow app-metric assertion for qualifier `eksint-p1039`.


gregschohn (commented):
High-level review courtesy of agent-assistance...  I'll do a thorough dive myself within the next couple days.

Scope replayer PDBs to one replay — /Users/schohn/dev/review/orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/replayer.ts:374-374
  When a workflow creates more than one traffic replay, each new PDB uses only `app: replayer` as its selector, while every replayer pod carries that label. That makes all replayer PDBs overlap the same pods, so Kubernetes evictions can be rejected because a pod matches multiple PDBs and a nonzero `minPodReplicas` on one replay can affect unrelated replays. Please include labels unique to this replay/deployment in the selector.

migrationConsole/lib/integ_test/integ_test/metric_operations.py
```diff
@@ -144,19 +150,43 @@ def _metric_has_recent_data(client, metric_name: str, dimensions, lookback_minut
...
+                lookback_minutes):
+            return True
     return False
 
 
 def assert_cloudwatch_capture_replay_metrics_for_workflow_run(
         namespace: str = "ma",
         lookback_minutes: int = 20,
-        attempts: int = 5,
+        attempts: int = 16,
```
gregschohn, line 189: should we make this backoff exponential - or a lot bigger.

migrationConsole/lib/integ_test/tests/test_metric_operations.py
gregschohn, line 1: Do we have any non-test code reading from CW metrics yet?  This is a lot of effort to test test-code.
However, we SHOULD be able to read CW and prometheus metrics to tie it into workflow commands so that users don't need to switch contexts - so this should eventually pay off anyway.

orchestrationSpecs/packages/argo-workflow-builders/src/renderers/argoExpressionRender.ts
```diff
@@ -32,6 +33,16 @@ export type ArgoFormatted = {
 
 const formattedResult = (text: string, compound = false): ArgoFormatted => ({text, compound});
 
+/** Escapes a value for a single-quoted expr-lang / govaluate string literal. */
+function escapeGovaluateStringLiteral(s: string): string {
```
gregschohn, line 37: THANK YOU!

AndreKurait:
@gregschohn addressed those concerns in the latest changes

orchestrationSpecs/packages/argo-workflow-builders/tests/integ/infra/argoCluster.js
```diff
@@ -84,7 +84,7 @@ export function startCluster() {
                 case 3:
                     _c.sent();
                     console.log("Installing Argo Workflows...");
-                    argoVersion = "v4.0.0";
+                    argoVersion = "v4.0.3";
```
gregschohn, line 87: did we bump this?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/podDisruptionBudget.ts
```diff
@@ -0,0 +1,66 @@
+import {
+    BaseExpression,
+    makeDirectTypeProxy,
+    makeStringTypeProxy,
+    ResourceWorkflowDefinition,
+} from "@opensearch-migrations/argo-workflow-builders";
+import {OwnerReference} from "@opensearch-migrations/k8s-types";
+
+export type PodDisruptionBudgetManifest = {
```
gregschohn: let's see if we can pull this in via the k8s server.  I have a patch for this.

gregschohn (approved):

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/commonUtils/podDisruptionBudget.ts
```diff
@@ -4,25 +4,9 @@ import {
...
-    spec: {
-        minAvailable: number | object,
-        selector: {
-            matchLabels: Record<string, string | object>,
-        },
-    },
-};
-
-function stringifyLabels(labels: Record<string, string | BaseExpression<string>>): Record<string, string | object> {
+function stringifyLabels(labels: Record<string, string | BaseExpression<string>>): Record<string, string> {
```
gregschohn, line 9: let's see if we can pull this in via the k8s server.  I have a patch for this.

Issue #3141, opened 2026-06-17
[BUG] Need an approval gate between 'Create Proxy: capture-proxy' and 'Create Snapshot: source'


yidaimi:
### What is the bug?
I found there is no approval gate between Create Proxy: capture-proxy' and 'Create Snapshot: source', which will cause data lose.

### What are your migration environments?

From Elasticsearch 8.17  migrate to AWS Opensearch v3.5.

Migration Assistant v 3.3.3

### What is the expected behavior?

I realize the way to achieve 0 downtime migration should be like this:
1. Create kafka and capture proxy --automatically.
2. Reroute client traffic to capture proxy --by manual.
3. create full snapshot on source ES -- should by triggered by manual
4. metadata migration... --automatically
5. snapshot recovery to target Opensearch for full migration... --automatically
6. Replay captured traffic from kafka  --automatically
7. Final client traffic cutover from capture proxy to target opensearch --by manual. 

But I found the step 3 will start automatically after step 1 fininshed even when I set"skipApprovals": false . So I have no opportunity to reroute my application traffic to capture proxy which will cause the data lost. 

I think there should be an approval gateway between step 1 and step 3 to leave me enough time to reroute client traffic.

### Do you have any additional context?
My workflow configuration, screenshots were attached.

[config.json](https://github.com/user-attachments/files/29036165/config.json)

<img width="1891" height="1257" alt="Image" src="https://github.com/user-attachments/assets/3dfe5c54-14d5-4748-b0b8-7874fd04e085" />

PR #3169, opened 2026-06-24
Add approval gate 'confirmtrafficrouting'

yidaimi:
### Description
I found there is no approval gate between Create Proxy: capture-proxy' and 'Create Snapshot: source', which will cause data lose.
Because it's impossible for anyone to reroute the traffic in such a short time.  However if we can't reroute traffic to capture-proxy before creating snapshot will cause data lose. 

So I create an approval gate 'confirmtrafficrouting' in this PR before 'Create Snapshot: source' to make users have enough time to reroute traffic to capture-proxy.

### Issues Resolved
[](https://github.com/opensearch-project/opensearch-migrations/issues/3141)

### Testing
Here is the screenshots of my local testing results.

approvallist-confirmtrafficrouting
workflow-status-confirmtrafficrouting

### Check List


AndreKurait:
@yidaimi Thank you for your contribution. This is something that @gregschohn had been looking at so i want to check in with him to see if this is already added in one of his PRs.

The pattern we would expect in lieu of this is to do incremental workflow submits. So do one with just the kafka and proxy settings, and then shift over to the proxy, and then do another with the full migration workflow with backfill and replay.

yidaimi:
> @yidaimi Thank you for your contribution. This is something that @gregschohn had been looking at so i want to check in with him to see if this is already added in one of his PRs.
> 
> The pattern we would expect in lieu of this is to do incremental workflow submits. So do one with just the kafka and proxy settings, and then shift over to the proxy, and then do another with the full migration workflow with backfill and replay.

@AndreKurait Thank you for you reply.
I totally agree with you.
In my opinion ,there are 2 ways to leave enough time for users to reroute the traffic.
1. Reroute after deployment's ready before workflow submit. 
-- However, the capture-proxy is not ready at that time. You should move part of helm installation scripts including kafka, capture-proxy from workflow submit phase to deployment phase. So users can get the ingress location and reroute the traffic after deployment. This is the way you are working at.

2. Add approval gate after workflow submitted. 
-- This is the way what my PR worked at. 

Both approaches work — it depends on which one you choose.

gregschohn:
Thanks @yidaimi!  We've needed this approval gate for a while.  
I have one high-level question - as the code is now, we set up a proxy, let the workflow move on, then have an approval before each snapshot.  For users that aren't doing snapshot migrations, they'll skip those approval gates.  For users not setting up a proxy, the workflow will have to bring this concern in.

Would it be just as acceptable if we put the gate right along with the proxy setup?  The proxy steps would only finish after the gate was approved (or skipped).  That way, no matter what the downstream dependencies are, they should block or flow just fine - and there's a single point of control for each proxy.

If there's a reason to do a check before we start a snapshot, that's probably a separate feature.  Let me know what you think.

yidaimi:
> Thanks @yidaimi! We've needed this approval gate for a while. I have one high-level question - as the code is now, we set up a proxy, let the workflow move on, then have an approval before each snapshot. For users that aren't doing snapshot migrations, they'll skip those approval gates. For users not setting up a proxy, the workflow will have to bring this concern in.
> 
> Would it be just as acceptable if we put the gate right along with the proxy setup? The proxy steps would only finish after the gate was approved (or skipped). That way, no matter what the downstream dependencies are, they should block or flow just fine - and there's a single point of control for each proxy.
> 
> If there's a reason to do a check before we start a snapshot, that's probably a separate feature. Let me know what you think.

Hi @gregschohn , thanks for your reply.

I think an approval gate is still needed even when a user only needs to do a full restore. Imagine that once you reach the cutover window for migration, you'd have to carry out "creating the base resources" + "creating the snapshot" + "restoring the snapshot" all together — this would significantly extend the downtime.

On top of that, the "creating the base resources" step can sometimes get blocked due to factors like resources, networking, or permissions, so it carries a fair amount of risk. Therefore, for the full-migration-only scenario, I think the ideal steps should be:

**Before cutover**: infrastructure deployment is completed, and it then waits for approval to create the snapshot.
**During the cutover window**: create the snapshot + restore the snapshot.
Of course, if possible, I'd like the workflow to support creating and restoring snapshots repeatedly. That would make it more robust — if the user makes a mistake, they'd still have a chance to try again, without having to tear down the infrastructure and recreate it.

**Based on the above thinking**:

For the second scenario you mentioned — "putting the gate right along with the proxy setup" — it seems it still doesn't achieve the separation of "creating the base resources" and "creating the snapshot."

For the first scenario, it seems to satisfy the principle of separating "creating the base resources" from "creating the snapshot," but I'm not sure whether it allows creating and restoring snapshots repeatedly.

As for treating snapshot creation as a separate feature — I think that sounds good, and I rather like this approach.

gregschohn:
> > Thanks @yidaimi! We've needed this approval gate for a while. I have one high-level question - as the code is now, we set up a proxy, let the workflow move on, then have an approval before each snapshot. For users that aren't doing snapshot migrations, they'll skip those approval gates. For users not setting up a proxy, the workflow will have to bring this concern in.
> > Would it be just as acceptable if we put the gate right along with the proxy setup? The proxy steps would only finish after the gate was approved (or skipped). That way, no matter what the downstream dependencies are, they should block or flow just fine - and there's a single point of control for each proxy.
> > If there's a reason to do a check before we start a snapshot, that's probably a separate feature. Let me know what you think.
> 
> Hi @gregschohn , thanks for your reply.
> 
> I think an approval gate is still needed even when a user only needs to do a full restore. Imagine that once you reach the cutover window for migration, you'd have to carry out "creating the base resources" + "creating the snapshot" + "restoring the snapshot" all together — this would significantly extend the downtime.
> 
> On top of that, the "creating the base resources" step can sometimes get blocked due to factors like resources, networking, or permissions, so it carries a fair amount of risk. Therefore, for the full-migration-only scenario, I think the ideal steps should be:
> 
> **Before cutover**: infrastructure deployment is completed, and it then waits for approval to create the snapshot. **During the cutover window**: create the snapshot + restore the snapshot. Of course, if possible, I'd like the workflow to support creating and restoring snapshots repeatedly. That would make it more robust — if the user makes a mistake, they'd still have a chance to try again, without having to tear down the infrastructure and recreate it.
> 
> **Based on the above thinking**:
> 
> For the second scenario you mentioned — "putting the gate right along with the proxy setup" — it seems it still doesn't achieve the separation of "creating the base resources" and "creating the snapshot."
> 
> For the first scenario, it seems to satisfy the principle of separating "creating the base resources" from "creating the snapshot," but I'm not sure whether it allows creating and restoring snapshots repeatedly.
> 
> As for treating snapshot creation as a separate feature — I think that sounds good, and I rather like this approach.

Thanks for the feedback - I can see some value in having an approval up-front.  We discussed such a thing internally too long ago. Please take a look at [3203](https://github.com/opensearch-project/opensearch-migrations/pull/3203).  That adds an approval gate at the very _beginning_ of a migration.  That PR is stacked on top of other PRs that add gates after proxy setup and after backfill completes.  I hope to get them merged within the next couple days.

gregschohn:
With 3203 in, the issues discussed in this ticket should all be addressed.
Thanks for the contributions!

Issue #3045, opened 2026-06-02
Improve version validation for Solr backups in RFS metadata migration


jzonthemtn:
The Solr migration already contains version-aware branching logic that selects the correct Lucene reader, backup layout handler, and schema parser based on the declared `--source-version`. This logic should be kept, but the validation backing it is incomplete.

Currently the user-declared Solr major version is trusted almost entirely. The only active cross-check is that a SIP-12 incremental backup layout (`shard_backup_metadata/`) rejects a declared version below 8. All other mismatches, a user declaring Solr 8 against a Solr 6/7 layout, for example, pass silently and produce confusing downstream failures.

The fix is to cross-check the declared version against observable layout markers in the backup directory and emit a clear, actionable error early, ideally during the metadata migration phase (`SolrSnapshotReader`) where failures are cheap and fast, before any document work has been scheduled.

**Proposed Change**

 Add a `validateVersionAgainstLayout(Path collectionDir, int declaredMajorVersion)` method to `SolrBackupLayout` that inspects the markers above and returns a list of warnings or throws on a clear mismatch. Call this from `SolrSnapshotReader` (or `SolrBackupIndexMetadataFactory`) during the metadata pass, before any shard work items are created.

Where the mismatch is unambiguous (e.g., `snapshot.shardN/` dirs present with a declared version of 8), throw with a clear message. Where the mismatch is a strong signal but not definitive (e.g., numbered `zk_backup_N/` with declared version 6), emit a warning and continue, consistent with the existing `--allow-loose-version-matching` mechanism.

  Example error messages:
  - "Backup contains `snapshot.shardN/` directories (Solr 6 layout) but declared version is SOLR 8. Check `--source-version`."
  - "Backup contains `backup_0.properties` (Solr 8+ layout) but declared version is SOLR 7. Check `--source-version`."
  - "Backup contains `shard_backup_metadata/` (Solr 8.9+ SIP-12 layout) but declared version is SOLR 8.0. Consider declaring SOLR 8.9 or later."

**Acceptance Criteria**

- [ ] A declared Solr version that clearly contradicts an observed backup layout marker produces an actionable error during the metadata phase, before any shard work items are created.
- [ ] Ambiguous mismatches (strong signal, not definitive) produce a warning and respect `--allow-loose-version-matching`.
- [ ] No existing version-aware branching logic is removed; this change only adds validation on top of it.
[...]

PR #3187, opened 2026-06-29
Read Solr 6, 7, 8, and 9 cloud/standalone backups in their native layout without reshaping

### LINKED ISSUE #3147: Integrate "Reshape" Logic for SolrCloud / standalone 6, 7, 8, 9 into RFS
For Solr running as SolrCloud we output the files in:

Solr 7

```
./backups/solrcloud/nyc_taxis_7/
├── backup.properties
├── snapshot.shard1
│   ├── _0_Lucene50_0.doc
│   ├── _0_Lucene50_0.pos
│   ├── blah blah
│   ├── _6.si
│   └── segments_2
├── snapshot.shard2
│   ├── _0_Lucene50_0.doc
│   ├── _0_Lucene50_0.pos
│   ├── blah blah
│   ├── _6.si
│   └── segments_2
├── snapshot.shard3
│   ├── _0_Lucene50_0.doc
│   ├── _0_Lucene50_0.pos
│   ├── _0_Lucene50_0.tim
│   ├── blah blah
│   ├── _7.si
│   └── segments_2
└── zk_backup
    ├── collection_state.json
    └── configs
        └── nyc_taxis
            ├── schema.xml
            ├── solrconfig.xml
            ├── stopwords.txt
            └── synonyms.txt
```

This currently requires the user to modify the shape of the data before uploading to S3 for the Migration Assistent to:

```
./backups/solrcloud/nyc_taxis_7/
└── nyc_taxis
    ├── backup.properties
    ├── snapshot.shard1
    │   ├── _0_Lucene50_0.doc
    │   ├── blah blah
    │   ├── _7.si
    │   └── segments_2
    ├── snapshot.shard2
    │   ├── _0_Lucene50_0.doc
    │   ├── blah blah
    │   ├── _6.si
    │   └── segments_2
    ├── snapshot.shard3
    │   ├── _0_Lucene50_0.doc
    │   ├── _0_Lucene50_0.pos
    │   ├── blah blah
    │   ├── _7.si
    │   └── segments_2
    └── zk_backup
        ├── collection_state.json
        └── configs
            └── nyc_taxis
                ├── schema.xml
                ├── solrconfig.xml
                ├── stopwords.txt
                └── synonyms.txt
```

jzonthemtn:
### Description
Previously, SolrCloud 7 backups required reshaping by the user prior to being uploaded to S3. These changes removes the need for the user to do any manual modifications.

### Issues Resolved
Closes #3147. Related to #3045.

### Testing
Added unit/integration tests, and manually tested on EKS.

### Check List


jzonthemtn:
This PR was originally just for Solr 7, but the changes were small or non-existent (only added tests) for 6, 8, and 9, so I included those versions in this PR, too.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -165,6 +158,13 @@ public static class Args {
             description = ("Repository URI. Schemes: file:///path, s3://bucket/path, gs://bucket/path (or bare absolute path)"))
         public String repoUri = null;
 
+        @Parameter(required = false,
+            names = { "--solr-collection-name", "--solrCollectionName" },
```
malliaridis: I know that "collection" and "core" are two well-known terms in the Solr ecosystem, but since "--solr-collection-name" refers to both even though "collection" is used, and the description calls it "index" at the very beginning, would it not be better to use `--solr-index-name` instead and mention in the description that it means "collection name" for Solr cloud and "core name" for Solr standalone?

Of course, that's just a suggestion for avoiding confusion and attempts to use `--solr-core-name`.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/SolrBackupDiscovery.java
```diff
@@ -0,0 +1,202 @@
...
+        return schemas;
+    }
+
+    /** Only S3 needs lazy per-shard downloads; filesystem backups are already local. */
+    public boolean shardPreparationNeeded() {
+        return s3Repo != null && bareMode != SolrBackupLayout.SolrBackupMode.STANDALONE;
+    }
+
+    public void prepareCollection(String collection) {
+        if (s3Repo != null && bareMode == SolrBackupLayout.SolrBackupMode.STANDALONE) {
```
malliaridis, line 91: In case `bareMode` is `null` but the underlying Solr backup is from standalone, would that not execute the wrong logic?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/SolrBackupDiscovery.java
```diff
@@ -0,0 +1,202 @@
...
+        return dataDirByCollection;
+    }
+
+    public Map<String, JsonNode> schemas() {
+        return schemas;
+    }
+
+    /** Only S3 needs lazy per-shard downloads; filesystem backups are already local. */
+    public boolean shardPreparationNeeded() {
+        return s3Repo != null && bareMode != SolrBackupLayout.SolrBackupMode.STANDALONE;
```
malliaridis, line 87: In case `bareMode` is `null` and the underlying Solr backup is standalone, this would return true. Is that fine?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/SolrBackupDiscoveryTest.java
```diff
@@ -0,0 +1,310 @@
...
+    @Test
+    void s3CloudBare_nameOverrideWins_andSkipsPropertiesDownload() throws Exception {
+        var s3Repo = s3Repo();
+        when(s3Repo.listTopLevelDirectories()).thenReturn(List.of("zk_backup"));
+
+        var discovery = SolrBackupDiscovery.discover(s3Repo, backupDir, "override_idx", List.of());
+
+        assertThat(discovery.collections(), contains("override_idx"));
+        verify(s3Repo, never()).downloadFile(anyString());
+    }
```
malliaridis: Can we add tests that confirm that the override is ignored when we have multiple collections?According to documentation of `solrCollectionName` it should ignore the override:

> Ignored for wrapped multi-collection layouts.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/SolrBackupDiscoveryTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+    @Test
+    void s3CloudBare_propertiesReadFailure_leavesNameNull_fallsBackToTopLevelListing() throws Exception {
+        var s3Repo = s3Repo();
+        when(s3Repo.listTopLevelDirectories()).thenReturn(List.of("zk_backup"));
+        when(s3Repo.downloadFile("backup.properties")).thenThrow(new RuntimeException("boom"));
+
+        // CLOUD with null name is not a single-collection bare backup; falls back to the listing.
```
malliaridis, line 68: Is this fallback actually expected?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/SolrBackupDiscoveryTest.java
```diff
@@ -0,0 +1,310 @@
...
+    void s3StandaloneFlatRoot_nameOverrideWins() throws Exception {
+        var s3Repo = s3Repo();
+        when(s3Repo.listTopLevelDirectories()).thenReturn(List.of());
+        when(s3Repo.listFilesInS3Root()).thenReturn(List.of("segments_2"));
+
+        var discovery = SolrBackupDiscovery.discover(s3Repo, backupDir, "my_index", List.of());
+
+        assertThat(discovery.collections(), contains("my_index"));
+        assertThat(discovery.dataDirByCollection().get("my_index"), equalTo(""));
+    }
```
malliaridis: What about multiple segments being present? Should the name override still work?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/SolrBackupDiscoveryTest.java
```diff
@@ -0,0 +1,310 @@
...
+    void s3StandaloneFlatRoot_atBucketRoot_emptyKeyYieldsEmptyDerivedName() throws Exception {
+        var s3Repo = s3Repo();
+        when(s3Repo.listTopLevelDirectories()).thenReturn(List.of());
+        when(s3Repo.listFilesInS3Root()).thenReturn(List.of("segments_2"));
+        when(s3Repo.getS3RepoUri()).thenReturn(new S3Uri("s3://bucket"));
+
+        var discovery = SolrBackupDiscovery.discover(s3Repo, backupDir, null, List.of());
+
+        // Empty repo key -> empty derived core name (defensive guard in lastPathSegment).
+        assertThat(discovery.collections(), contains(""));
```
malliaridis: Is this not a problem for continuing the path? I don't believe it is possible to create a collection with an empty string as name, right?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/SolrBackupDiscoveryTest.java
```diff
@@ -0,0 +1,310 @@
...
+        assertThat(discovery.collections().isEmpty(), is(true));
+    }
+
+    @Test
+    void s3FlatRootProbe_listFilesFailure_notBare() throws Exception {
+        var s3Repo = s3Repo();
+        when(s3Repo.listTopLevelDirectories()).thenReturn(List.of());
+        when(s3Repo.listFilesInS3Root()).thenThrow(new RuntimeException("cannot list root"));
+
+        var discovery = SolrBackupDiscovery.discover(s3Repo, backupDir, null, List.of());
```
malliaridis: What happens if collection name is always provided, but no data / backup is present?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/SolrBackupDiscoveryTest.java
```diff
@@ -0,0 +1,310 @@
...
+
+        discovery.prepareCollection("col_a");
+
+        verify(s3Repo).downloadPrefix("col_a/zk_backup_0");
+        verify(s3Repo).downloadPrefix("col_a/shard_backup_metadata");
+        assertThat(discovery.dataDirByCollection().get("col_a"), equalTo("col_a"));
+    }
+
+    @Test
+    void prepareCollection_cloudUnknownCollection_unresolved_warnsAndSkipsDownloads() throws Exception {
```
malliaridis, line 225: We do not explicitly check for warnings / logs to make sure that the user is notified (e.g. here "warned")

malliaridis (commented):
_TL;DR_

I believe we are better of if we skip the new parameter `--solr-collection-name`, as it seems to not work as expected. Instead, we should solely rely on the name of the backup directory, since the user is already copying files to s3 in the cases where it is needed, and in cases where it is not (automatic backups), the override may not work and be limited to a single snapshot / idnex.

_Full story_

This is only a partial feedback that resulted from manual testing. The process was very error-prone and confusing, so any discoveries below have to be double-checked before addressing them.

I will provide some insights for documentation purposes, so bear with me.

About Solr 6.6.6 Standalone one core (call it `products`) backup is made externally and uploaded to s3:
- In standalone, when using an external snapshot, the user will provide a `externallyManagedSnapshotName` (lets say `products`), the backup directory in s3 would need to match that name, so renaming the directory `snapshot.products` (if `&name=products` was provided during creation) or `snapshot.20260709102030123` to `products` should suffice, otherwise the user cannot point to the right directory. The user ends up with `s3://<bucket>/<snapshotName>` (here `s3://solr-backups/products`)
- It should also be possible to keep the prefix `snapshot.` and it will likely be stripped off

When having multiple cores (`tech_products`, `beauty_products`) under the same snapshot name (`products`):
- The user would copy the backups to s3 probably like this:
	- `s3://solr-backups/products/tech_products`
	- `s3://solr-backups/products/beauty_products`
- Using `externallyManagedSnapshotName=products` will identify both backups and replay them to target correctly
- Having the prefix `snapshot.*` in core snapshots is correctly stripping it off
- When any of the core backups is prefixed with `snapshot.*`, the rest that are not prefixed are ignored, including those in `indexAllowList`

- If the user wants to replay only a set of backups, I believe they would use `indexAllowList`, and not overwrite the core via `solrCollectionName`, but rather by naming the backup accordingly
	- It seems that `indexAllowList` does not work as expected in the current configuration, perhaps I misuse the "snapshotName" and it has to be just a s3 path prefix instead?
	- I tested configurations with `indexAllowList` because I saw that in one of your examples

- Using `solrCollectionName` in Solr standalone with multiple core backups makes only sense when `indexAllowList`  is limiting to a single core backup as well, as otherwise all backups would be replayed with the same collection name (I guess?)


When creating in Solr 6.6.6 Standalone a backup via the workflow:
- Only local filesystem is supported I believe, s3 repository not available in Solr < 8.10.0
- The user still has to manually copy the data to s3, where a rename could potentially happen

I believe in non of the above cases would `--solr-collection-name` be necessary.

For Solr 6.6.6 Cloud mode I could not create a backup yet to test.

For Solr 9.7.0 workflow with backup and RFS works in Cloud mode, configurations like `indexAllowList` were not tested here, and probably out of scope.

More Solr versions feedback will follow.

I have also added a few notes about the nullability of `bare` variable. Since it is coupled with multiple conditions, it is hard to make sure it is not `null`, or how the implementation behaves in case it is `null`.
Under which circumstances would it be `null`? Perhaps that's what I should ask here.

malliaridis:
I have to update a few details related to my previous feedback related to multiple snapshots under the same directory:

Using `externallyManagedSnapshotName=products` in the setup before is probably wrong, as it should point to the actual snapshot, not the parent directory. If pointing to a specific snapshot directory / a single core backup, it should behave correctly, as it contains only that. The name in this case is retrieved from the snapshot directory `snapshot.tech_products` with `snapshot.` stripped off if present, which is the correct behavior as described in cases of single snapshot backups.

This however may be cumbersome for users with standalone instances with hundreds of cores.

What my feedback perhaps makes clear is that it may be confusing for end-users as well, that they can configure like 3 to 5 different configurations values just to point to one or more directories / snapshots. Adding another one new makes it not less complicated.

Since the backup of standalone cores is not working in this state of the project (addressed in separate PR probably), it should not be the responsibility of this PR to address the renaming of target index names if the snapshot is named like `snapshot.20260709102030123` (if that was the main reason this extra value was introduced).

EDIT: It also seems that even when providing a single snapshot like `snapshot.20260709102030123`, the collection name override is not used as expected, it still uses the second half of the snapshot name.

jzonthemtn:
## How the Solr to  OpenSearch index name is set

The `solrCollectionName` override was removed in the last commit and the name is always derived from the backup. RFS classifies the layout and then applies that layout's rule:

| Layout | Name comes from |
|--------|-----------------|
| Wrapped, multi-collection (`<root>/<collection>/…`) | each top-level directory name |
| Bare SolrCloud | `collection` in `backup.properties` |
| Bare standalone, wrapped (`snapshot.<name>/`) | wrapper dir name, `snapshot.` stripped |
| Bare standalone, flat root (`segments_N` at root) | last path segment, `snapshot.` stripped |

SolrCloud records its collection name in the backup, standalone doesn't, so standalone names can only be inferred from the storage path.

A flat standalone index at the `bucket/filesystem` root has no path segment to name it after, so `requireDerivedStandaloneName` errors instead of migrating into a blank index name. In this case the migration fails with an error in the log, exit code 5, and a failed Argo step.

gregschohn:
Thanks for this PR.  I like the refinments that you're doing on solr backup handling!

Some of the changes here are parallel to changes in [PR-3209](https://github.com/opensearch-project/opensearch-migrations/pull/3209) (which is an offshoot of https://github.com/opensearch-project/opensearch-migrations/pull/3132).  

I'd like to add 3209 in ASAP.  In addition to the solr tweaks, it has some material changes on hardening field-handling for dependency graphs and approval checks.  Those fields include the new solr-backup ones and others that were found in exhaustive sweeps.  There are some minor conflicts right now between the two PRs, but I don't think anything has been made redundant yet.  

Let me know what you think of the plan to merge 3209 into main, merge main into this PR, and to migrate the SolrBackupStrategy, etc code into the CreateSnapshot case to prepare/canonicalize the snapshot for all other paths.  I'm open to some other options too, but that was one that seemed to create even less net user-config.

## Next Steps

### UX Enhancement: Zero-Config Prepare for SolrCloud IMPORT Mode

Both live-cluster calls in `SolrBackupStrategy` — topology detection (`isSolrCloud`) and
collection name discovery (`resolveCollections`) — can be answered from the backup layout
for SolrCloud IMPORT mode using the detection primitives added in this PR:

| Call | Backup equivalent |
|---|---|
| `isSolrCloud()` | presence of `zk_backup/` + `backup.properties` markers |
| `resolveCollections()` | collection name in `backup.properties` |

Since `isCloud = true` also skips `ensureConfigFiles*()` (SolrCloud backups already include
the schema in `zk_backup/`), a successful layout detection means **zero live API calls**
for SolrCloud IMPORT.

Standalone IMPORT is unchanged — standalone backups don't include the schema, so the live
cluster is still required to fetch it. CREATE mode is also unchanged — the live cluster is
needed to trigger the backup.

**What the implementation looks like:**

1. **`SolrBackupLayout.java`** — Extract a static helper that accepts lambdas for listing and
   downloading instead of taking `S3Repo` directly, so both `ClusterReaderExtractor` and
   `SolrBackupStrategy` can share the detection logic without a new coupling.

2. **`SolrBackupStrategy.java`** — In the constructor, probe the backup layout before calling
   `isSolrCloud()`. If layout detection succeeds and mode is IMPORT, derive `isCloud` and
   the collection name from the backup and skip both live-cluster calls. Fall back to the
   existing live-cluster path if detection fails or mode is CREATE.

3. **`ClusterReaderExtractor.java`** — Update to use the new lambda-based helper. No logic
   change.

4. **Tests** — Unit tests for the detection path in `SolrBackupStrategy` reusing this PR's
   layout fixtures. No new module dependency — `CreateSnapshot` already depends on `SolrReader`.

---

### Merging into main after PR #3209 Lands

PR #3209 (migration-resource contract cleanup, targeting the Solr backup branch) touches two
files that this PR also modifies. Both conflicts are mechanical.

#### `orchestrationSpecs/packages/schemas/tests/__snapshots__/argoSchemaSnapshot.test.ts.snap`

3187 makes a small path fix (`tests/fixtures/` → `packages/schemas/tests/fixtures/`).
3209 regenerates the entire snapshot to reflect renamed symbols and the new Solr schema types.

#### `migrationConsole/lib/console_link/console_link/models/snapshot.py`

Both PRs independently add `solr_collections` handling with identical intent:
- Add `'solr_collections'` to the Cerberus validator schema
- Read `self.solr_collections = config.get("solr_collections") or None`  
- Use `self.solr_collections or self._get_solr_collections()` in `create()`

3209 also adds `'mode'` to the validator and `command_args["--mode"]` for all Solr paths.

**Resolution:** Take 3209's version as the base (it's a superset). Verify that the
`solr_collections` field and the `mode` field are both present in the merged validator schema
and both wired into `create()`. Run the snapshot tests to confirm:

```sh
pytest migrationConsole/lib/console_link/tests/test_snapshot.py
```

SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrBackupLayout.java
```diff
@@ -297,6 +299,312 @@ private static boolean containsBackupDataMarkers(Path dir) {
...
+        if (Files.isRegularFile(bare)) {
+            return bare;
+        }
+        try (var entries = Files.list(dir)) {
+            return entries.filter(Files::isRegularFile)
+                .filter(p -> {
+                    var name = p.getFileName().toString();
+                    return name.startsWith("backup_") && name.endsWith(".properties");
+                })
+                .max(Comparator.comparing(p -> p.getFileName().toString()))
```
gregschohn: Should we parse more of the string to extract the numerical values so that ..._9... isn't greater than ..._10...?  If not, explain why extractZkBackupIndex isn't necessary.

MetadataMigration/src/main/java/org/opensearch/migrations/cli/ClusterReaderExtractor.java
```diff
@@ -151,18 +166,42 @@ private void downloadZkBackupForCollection(S3Repo s3Repo, String collection) {
...
+    void downloadZkBackupForDataDir(S3Repo s3Repo, String dataDir) {
+        var zkName = SolrBackupLayout.findLatestZkBackupName(s3Repo.listSubDirectories(dataDir));
+        if (zkName == null) {
+            log.warn("No zk_backup found at bare Solr backup root '{}'", dataDir);
+            return;
+        }
+        s3Repo.downloadPrefix(SolrBackupLayout.joinPrefix(dataDir, zkName));
+    }
+
+    static SolrBackupLayout.BareBackupLayout detectBareSolrLayoutInS3(S3Repo s3Repo) {
```
gregschohn: I think this is a dup of https://github.com/opensearch-project/opensearch-migrations/pull/3187/changes#diff-2c1ec7304ef433b23dae958259e6d9c196185a9a1d8c4f369632d4c537223a6bR73-R84

SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrBackupLayout.java
```diff
@@ -297,6 +299,312 @@ private static boolean containsBackupDataMarkers(Path dir) {
...
+        return name.startsWith("snapshot.") ? name.substring("snapshot.".length()) : name;
+    }
+
+    /**
+     * Reads the collection name from a SolrCloud {@code backup.properties} (or the latest
+     * {@code backup_N.properties}) under the given directory. Returns {@code null} if no properties
+     * file is present or it carries no recognised collection-name key.
+     *
+     * <p>NOTE: the exact key Solr writes should be confirmed against a real Solr 7 backup; both
+     * {@code collection} and {@code collectionName} are accepted here.
```
gregschohn: Do you know the answer to this?  It's between 'collection' and 'collectionName'?

MetadataMigration/src/main/java/org/opensearch/migrations/cli/ClusterReaderExtractor.java
jzonthemtn: Yes, I agree. Will update.

SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrBackupLayout.java
jzonthemtn: It is `collection`. Will update to drop `collectionName`.

SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrBackupLayout.java
jzonthemtn: The only thing read from the files is the `collection` which is the same in all files so the ordering does not matter. But it does feel weird and could raise questions later on so I changed it to order them.

malliaridis:
```java
.max(Comparator.comparing(p -> p.getFileName().toString()))
```

Is this sorting rule not failing for `backup_6.properties` and `backup_29.properties` (due to 2 < 6)? If I am not mistaken the number is the backup count and it is always incremented by one whenever a new backup is created (saw the number increasing for scheduled backups).

jzonthemtn: Yes, and I must have fixed it right before you posted that. I did add a unit test for `6` and `29` just now.

gregschohn (approved):

PR #3189, opened 2026-07-01
Harden migration workflow approval gates and console hints

gregschohn:
### Description

This PR stacks on top of #3118 and finishes the approval-gate and workflow-hint work so it is ready to merge.

#3118 introduced the initial shape:
- approval gates for migration workflow phases, including proxy setup and document backfill
- migration-console hints for common workflow commands
- supporting docs and examples for manually testing approval gates

This PR adds the follow-up hardening and review fixes:
- Adds explicit `skipApproval` support for capture-proxy approval gates.
- Resolves skip behavior using the most granular configured value first, falling back to global `skipApprovals` only when the per-gate value is unset.
- Keeps approval-gate `when` expressions aligned with the existing evaluate/migrate metadata gate structure.
- Hardens `Test0003ApprovalGateIntegration` so it exercises a fuller migration flow and verifies gates by phase rather than relying only on `approve --all`.
- Improves workflow hint handling so `status` returns the displayed phase explicitly and `manage` reuses the TUI’s last-known phase instead of making an extra `get_workflow(...)` call after exit.
- Cleans up dead/unreachable hint-control flow and keeps hint imports consistent.
- Updates snapshots and schema coverage for the new approval and skip-approval wiring.

### What This Adds Over #3118

#3118 proves the concept and wires in the initial gates/hints. This PR makes that behavior more precise and testable:

- Per-proxy approval skipping is now configurable and does not rely on blanket global skip behavior.
- Approval skip resolution is deterministic: local setting first, global fallback second.
- Approval-gate integration coverage verifies that expected gates actually trip.
- The console hint path avoids extra cluster calls where the needed phase is already known.
- The implementation incorporates review feedback from #3118/#3189 rather than leaving those concerns for a later cleanup.

### Testing

- `npm run type-check` from `orchestrationSpecs`
- `git diff --check`
- Approval-gate integration workflow was previously exercised locally against kind during this stack.
```

I’d consider changing the PR title too. Current title is vague. Better:

```text
Harden migration workflow approval gates and console hints
```
### Check List


malliaridis:
@gregschohn in your description you mention #3169 instead of #3118. I believe that was a typo?

gregschohn:
@malliaridis - correct, I meant 3118.  3169 is related, but the functionality that that was aiming for should be wrapped up in 3118 + 3189 (here).  Thanks!

awagen:
@gregschohn LGTM, just due to merge of main there are two unsigned commits right now, after signing we can merge this one in main and close the #3118. Also still a bunch of checks failing, possibly due to config changes?

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/setupCapture.ts
```diff
@@ -761,6 +767,16 @@ export const SetupCapture = WorkflowBuilder.create({
...
                 )
+                // Gate here, after the proxy is deployed and its endpoint is serving, so the
+                // operator can swing client traffic through the proxy and verify capture before
+                // approving. Only once approved do we patch the CaptureProxy to Ready below, which
+                // publishes checksumForSnapshot and releases the downstream snapshot/backfill flow.
+                .addStep("approveProxySetup", ResourceManagement, "waitForUserApproval", c =>
+                    c.register({
+                        resourceName: expr.concat(expr.literal("captureproxysetup."), b.inputs.proxyName)
+                    }),
+                    { when: c => ({templateExp: widenComplexity(expr.not(skipProxyApproval))}) }
```
k-rooot: This renders (per tests/__snapshots__/setupCapture.snap.json, the approveProxySetup step) to:
```
"when": "{{=!(inputs.parameters.skipApproval)}}"
```

`inputs.parameters.skipApproval` is an Argo string parameter (default "false"), so this negates a string inside an expr-lang block. That is exactly the case the earlier comment at fullMigration.ts:476-483 says throws interface {} is string, not bool, and why the backfill gate wraps its value in sprig.dig(...) to get a real boolean instead.

The nearby metadata/backfill gates don't hit this because they negate a boolean operand (!(0 == len(...)), !(sprig.dig(...))), not a bare string param. The proxy-setup gate is the only when that applies ! directly to a string parameter.

So either this throws when a proxy actually reaches the gate (and no test catches it — snapshot tests only assert rendered YAML, not evaluation), or the sprig.dig workaround in fullMigration.ts was unnecessary. Please confirm with a live Argo run (or an expr-eval test) that this when evaluates rather than throws. If it does throw, mirror the backfill fix — e.g. sprig.dig/a bool-typed expression instead of expr.not on the raw param.

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
```diff
@@ -652,13 +652,17 @@ export class MigrationConfigTransformer extends StreamSchemaTransformer<
...
+        // globally or for a proxy, the proxy-setup approval gate is auto-skipped.
+        const globalSkipApprovals = userConfig.skipApprovals ?? false;
         const proxiesWithChecksums = proxies.map(p => ({
             ...p,
             kafkaConfig: { ...p.kafkaConfig, configChecksum: kafkaChecksums.get(p.kafkaConfig.label) ?? '' },
             configChecksum: cs(p.proxyConfig, kafkaChecksums.get(p.kafkaConfig.label)),
             topicConfigChecksum: cs(p.kafkaConfig.kafkaTopic, p.kafkaConfig.topicSpecOverrides, kafkaChecksums.get(p.kafkaConfig.label)),
             checksumForSnapshot: csDep(PROXY_SCHEMA, p.proxyConfig as Record<string, unknown>, 'snapshot', kafkaChecksums.get(p.kafkaConfig.label)),
             checksumForReplayer: csDep(PROXY_SCHEMA, p.proxyConfig as Record<string, unknown>, 'replayer', kafkaChecksums.get(p.kafkaConfig.label)),
+            skipApproval: globalSkipApprovals || (p.skipApproval ?? false),
```
k-rooot: 1. The global OR per-proxy logic means a user who sets skipApprovals: true globally cannot force the gate back on for one proxy. If that's intended, fine.
2. The validation test at migrationConfigTransformer.validation.test.ts:123 only covers global=false, per-proxy=true. Please add the global=true case so the override precedence is locked in.

buildImages/scripts/startMinikubeAndDeployCharts.sh
```diff
@@ -11,5 +11,5 @@ export imageRegistryPrefix="${HOST_IP_FROM_WITHIN_MINIKUBE}:5001/"
 echo "HOST_IP_FROM_WITHIN_MINIKUBE: $HOST_IP_FROM_WITHIN_MINIKUBE"
 # installs kyverno, argo, argo-workflow-controller, cert-manager, fluent, jaeger, localhost, ma-helm-installer, strimzi-cluster-operator
 # kube-prometheus, kube-grafana, otel service manager, s3 bucket (e.g localstack)
-helm install ma ../deployment/k8s/charts/aggregates/migrationAssistantWithArgo -n ma --create-namespace --wait --timeout 20m \
+helm install ma ../deployment/k8s/charts/aggregates/migrationAssistantWithArgo -n ma --debug --create-namespace --wait --timeout 20m \
```
k-rooot: `--debug` dumps the full rendered manifests and is very noisy. Is this intentional?

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -566,13 +568,15 @@ def _run_step_view(ctx, workflow_name, all_workflows, argo_server, namespace, in
...
     try:
+        from .hints import hint_after_status
         service = WorkflowService()
         handler = StatusCommandHandler(service, token)
 
         if all_workflows:
             handler.handle_status_command(None, argo_server, namespace, insecure, show_all, live_status)
         else:
             handler.handle_status_command(workflow_name, argo_server, namespace, insecure, show_all, live_status)
+            hint_after_status(handler._last_phase or '')
```
k-rooot: 1. _last_phase is private and is only populated inside _display_workflow_with_tree. If a future path doesn't reach that method, the hint silently degrades to no-op. Prefer returning the phase from handle_status_command. 

2. nit: Also, the from .hints import hint_after_status here is a function-local import while every other command imports at module top. We can make it consistent.

migrationConsole/lib/console_link/console_link/workflow/commands/manage.py
```diff
@@ -113,3 +114,11 @@ def manage_command(ctx, workflow_name, argo_server, namespace, insecure, token,
     except Exception as e:
         click.echo(f"Error: {str(e)}", err=True)
         ctx.exit(ExitCode.FAILURE.value)
+        return
+
+    try:
+        wf = get_workflow(namespace, workflow_name)
```
k-rooot: After the TUI exits we call get_workflow(...) again just to read the phase. The TUI already had this state; on a slow/unreachable cluster the user now waits on a call whose only purpose is a hint. Acceptable, but hand the last-known phase back from the TUI if you can. (nit: the return after ctx.exit(...) is dead code - ctx.exit raises)

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/setupCapture.ts
gregschohn: Thanks!   I agree that this is off.  I'll make sure it is consistent with how we do metadata eval and migrate

buildImages/scripts/startMinikubeAndDeployCharts.sh
awagen: I had added this for the local minikube case only since I found additional logging helpful, but agree this is more of an optional here and can be added on demand.

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/setupCapture.ts
gregschohn: I've updated this.  The subsequent PRs stacked on top of this should make the pattern consistent for every approval gate.  I've also strengthened the tests on approval gates to check every gate.

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
gregschohn: fixed

migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
```diff
@@ -768,7 +786,10 @@ def _run_subcommand(ctx, category, names, list_flag, all_flag, pre_approve,
     )
 
     if list_flag:
-        _run_list(gates, show_prereq=(category == 'retry'))
+        if output_format == 'json':
```
awagen, line 801: Here the difference in conditions would be normal or json format. Its intentional that there is no show_prereq in the json variant?

migrationConsole/lib/integ_test/integ_test/test_cases/basic_tests.py
```diff
@@ -127,19 +132,98 @@ def workflow_perform_migrations(self, timeout_seconds: int = MIGRATION_COMPLETIO
...
 
     def _approve_expected_step_gates(self, timeout_seconds: int):
         for gate_name in self._approval_gate_names():
             self._wait_for_step_gate(gate_name, "waiting", timeout_seconds)
+            self._assert_gate_prerequisite_completed(gate_name)
             self._approve_step_gate(gate_name)
             self._wait_for_step_gate(gate_name, "approved", timeout_seconds)
 
+    def _assert_gate_prerequisite_completed(self, gate_name: str):
+        if gate_name.startswith("evaluatemetadata."):
```
awagen, line 150: is a success of call to _assert_workflow_show_output_available actually ensuring the criteria to be checked in _assert_gate_prerequisiste_completed or is success ambiguous in the current implementation? Comes down to the question whether workflow show for a task name would only succeed under the condition we want to assert here. In the documentbackfill case in contras status checks are applied

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
```diff
@@ -79,17 +79,16 @@ def _handle_single_workflow(self, workflow_name: str, argo_server: str,
...
         workflow_list = self.data_fetcher.list_workflows(
             argo_server, namespace, insecure, exclude_completed=not show_all)
 
         if not workflow_list:
-            self._handle_no_workflows(show_all, argo_server, namespace, insecure, live_check)
-            return
+            return self._handle_no_workflows(show_all, argo_server, namespace, insecure, live_check)
 
         click.echo(f"Found {len(workflow_list)} workflow(s) in namespace {namespace}:")
         click.echo("")
```
awagen, line 94: Its counting rice corns, I know, but do we need that echo with empty string here?

awagen:
@gregschohn only had a few minor comments / questions in the above, nothing major that would block approval. Thanks for the PR additions! One thing: some commits need signing. Can approve afterwards (assuming approvals are removed on new commits)

migrationConsole/lib/console_link/console_link/workflow/commands/status.py
gregschohn: Those two lines are a header.  We drop into a loop where we print each found workflow with empty lines between each workflow status output.

migrationConsole/lib/integ_test/integ_test/test_cases/basic_tests.py
gregschohn: The evaluatemetadata (and separately migratemetadata) create outputs that are left behind in the argo artifacts bucket.  Those should be there for a sucessful step.  Only the metadata steps currently produce output artifacts, so that's why only those steps have output available assertions.

migrationConsole/lib/console_link/console_link/workflow/commands/approve.py
gregschohn: It was intentional, but since json fields can be easily ignored, I'm adding that to the json.  It could be valuable for other UIs.

awagen (approved):

PR #3196, opened 2026-07-06
fix: use connectionId as Kafka partition key to prevent fragment scattering

Aman-bv:
## Summary

This PR fixes three interconnected bugs that cause data loss, proxy 502s, and replayer stalls when handling large `_bulk` requests (up to 25MB) in multi-partition Kafka deployments.

### Bug 1: Fragment Scattering Across Partitions
The capture proxy uses `connectionId.index` as the Kafka record key. Since each fragment has a different index, Kafka's murmur2 partitioner routes fragments of the same connection to different partitions. With 24 partitions, a 50-fragment request scatters across ~21 partitions, making reassembly impossible when multiple replayer consumers are active.

### Bug 2: Proxy 502s on Kafka Errors
When Kafka rejects a record (e.g. `RecordTooLargeException`), the error handler in `ConditionallyReliableLoggingHttpHandler` releases the ByteBuf via `ReferenceCountUtil.release(msg)` and then still forwards it downstream. The `FrontsideHandler` → `SslHandler` path encounters `refCnt: 0` → `IllegalReferenceCountException` → channel dies → ALB returns 502.

### Bug 3: Accumulator Expiry Deadlock (amplified by Bug 1)
The expiry mechanism in `ExpiringTrafficStreamMap` is purely event-driven — it only triggers within `expireOldEntries()`, called from the observation processing path in `CapturedTrafficToHttpTransactionAccumulator`. There is no independent timer-based expiry. This creates a circular deadlock:

```
expiry needs observations → observations need polling → polling needs backpressure relief → backpressure needs expiry
```

When fragments scatter across partitions (Bug 1), each partition's consumer accumulates incomplete connection state that can never be reassembled (the other fragments are on other partitions assigned to other consumers). These zombie entries consume accumulator slots, creating backpressure that blocks polling, which prevents new observations from arriving, which prevents `expireOldEntries()` from being called — deadlock.

## Fixes

| Bug | Fix | File |
|-----|-----|------|
| Fragment Scattering | Use `connectionId` (without index) as Kafka partition key | `KafkaCaptureFactory.java` |
| Proxy 502s | Remove premature `ReferenceCountUtil.release(msg)` in error path | `ConditionallyReliableLoggingHttpHandler.java` |
| Accumulator Deadlock | Mitigated by fixing Bug 1 (no more zombie entries from scattered fragments) | — |
| Future Hang | Catch `Throwable` instead of `Exception` in `sendFullyAsync` | `KafkaCaptureFactory.java` |
| Silent Null Scattering | `Objects.requireNonNull(connectionId)` prevents silent round-robin | `KafkaCaptureFactory.java` |

## Impact

This is a **breaking change for existing Kafka topics** with multiple partitions — existing records were keyed with `connectionId.index` and new records will be keyed with `connectionId`. Consumers that rely on the old key format for deduplication or ordering should drain existing records before deploying this change. For most deployments (single replayer consumer or fresh topic), this is a transparent fix.

### Bug 1: Fragment Scattering
```
KafkaCaptureFactory.java line 125:
  ProducerRecord<>(topic, connectionId + "." + index, payload)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                          Different key per fragment → different partition
```

Simulation with 50 fragments across 24 partitions:
- Before fix: fragments scatter across 21/24 partitions
- After fix: all fragments land on 1 partition (determined by `hash(connectionId) % 24`)

### Bug 2: Proxy 502s
```
ConditionallyReliableLoggingHttpHandler.java:
  whenComplete((result, t) -> {
      if (t != null) {
          ReferenceCountUtil.release(msg);  // ← releases ByteBuf
      }
      super.channelFinishedReadingAnHttpMessage(ctx, msg, ...);  // ← forwards released ByteBuf!
  });
```

Sequence: RecordTooLargeException → release(msg) → forward msg → SslHandler.flush(refCnt:0) → IllegalReferenceCountException → channel dead → 502

### Bug 3: Accumulator Deadlock
```
┌─────────────┐     ┌──────────────┐     ┌─────────────────┐     ┌──────────────┐
│ expiry needs │────▶│ observations │────▶│ polling needs   │────▶│ backpressure │
│ observations │     │ need polling │     │ backpressure    │     │ needs expiry │
│              │◀────│              │◀────│ relief          │◀────│              │
└─────────────┘     └──────────────┘     └─────────────────┘     └──────────────┘
```

Fragment scattering fills the accumulator with unreassemblable zombie entries → backpressure → deadlock. Fixing partition locality eliminates the zombie entries.

### Production code
- `KafkaCaptureFactory.java`: Kafka record key `recordId` → `connectionId`; null-check; catch `Throwable`
- `ConditionallyReliableLoggingHttpHandler.java`: Remove erroneous `ReferenceCountUtil.release(msg)` in error path

### Tests (`KafkaCaptureFactoryTest.java`)
- `testAllFragmentsUseSameKafkaKeyForPartitionLocality` — verifies all fragments share one key
- `testLargerBufferSizeReducesFragmentation` — 1MB vs 8MB buffer comparison
- `testMaxRequestSizeWithLargeBufferProducesValidRecords` — 7MB payload within 8MB limit
- `testDifferentConnectionsProduceDifferentKeys` — different connections get different keys
- `testSingleFragmentUsesConnectionIdAsKey` — non-fragmented traffic also uses correct key

### Docker test infrastructure
- `docker-compose.large-requests.yml` — full stack with 24-partition Kafka, 8MB settings
- `kafka-producer.properties` / `kafka-consumer.properties` — 8MB producer/consumer configs
- `test-large-requests.sh` — end-to-end validation script with partition locality checks

## Test plan

- [x] `testAllFragmentsUseSameKafkaKeyForPartitionLocality` — all fragments share one key
- [x] `testLargerBufferSizeReducesFragmentation` — 1MB buffer=many fragments, 8MB=1 record
- [x] `testMaxRequestSizeWithLargeBufferProducesValidRecords` — 7MB payload fits in 8MB buffer
- [x] `testDifferentConnectionsProduceDifferentKeys` — multi-connection isolation
- [x] `testSingleFragmentUsesConnectionIdAsKey` — single-fragment correctness
- [x] All existing tests pass (8/8 in captureKafkaOffloader, all in nettyWireLogging, all in trafficCaptureProxyServer)
- [ ] Local docker-compose end-to-end test with `test-large-requests.sh`

## Related Issues

- Accumulator deadlock: [ExpiringTrafficStreamMap.expireOldEntries()](https://github.com/opensearch-project/opensearch-migrations/blob/main/TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/ExpiringTrafficStreamMap.java) — purely event-driven, no timer-based fallback
- JIRA: [MIGRATIONS-1276](https://opensearch.atlassian.net/browse/MIGRATIONS-1276) — behavioral policy for offload failures


TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
```diff
@@ -178,8 +184,8 @@ public static <K, V> CompletableFuture<RecordMetadata> sendFullyAsync(
                         completableFuture.complete(metadata);
                     }
                 });
-            } catch (Exception exception) {
-                completableFuture.completeExceptionally(exception);
+            } catch (Throwable throwable) {
```
gregschohn: Why widen this?  My concern here would be that a fatal error like OOM now might get trapped in the CompletableFutures and the process might not terminate because of that.

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpHandlerTest.java
```diff
@@ -302,4 +302,49 @@ public void testThatAPostInASinglePacketBlocksFutureActivity_withLeakDetection(b
         // MyResourceLeakDetector.dumpHeap("nettyWireLogging_"+COUNT+"_"+ Instant.now() +".hprof", true);
     }
 
+    @Test
+    @WrapWithNettyLeakDetection(repetitions = 4)
```
gregschohn: Let's bump this up to something considerably higher - maybe 32.  It seems that the test will be quick, so 8x more time for 8x more likelihood to catch a leak should be worth it.

gregschohn (commented):
Thanks a lot for these critical fixes!

Let me know if you're comfortable removing the change to catch Throwable, or if you think that there's another reason that we should have it.

TrafficCapture/captureKafkaOffloader/src/main/java/org/opensearch/migrations/trafficcapture/kafkaoffloader/KafkaCaptureFactory.java
Aman-bv: Makes sense

TrafficCapture/nettyWireLogging/src/test/java/org/opensearch/migrations/trafficcapture/netty/ConditionallyReliableLoggingHttpHandlerTest.java
Aman-bv: Done 🚀

gregschohn (approved):
Thanks again.  LGTM.  Just waiting for tests to pass to merge this in.

PR #3197, opened 2026-07-06
Fix ES 6.8 startup failure caused by unsupported gcs.client settings

gregschohn:
## Problem

`main-k8s-matrix-test` has been failing on all ES 6.8 source combinations since build 171 (~June 29). Every test against ES 6.8 fails with:

```
WorkflowEndedBeforeSuspend: The workflow 'full-migration-with-clusters-...' reached ending phase of Failed before reaching suspend state
```

The root cause: commit 856bb9ad added `gcs.client.default.endpoint` and `gcs.client.default.token_uri` to the ES 6.8 `elasticsearch.yml` template in `clusterWorkflows.yaml`. These are ES **7.x+** settings — ES 6.8 rejects them as unknown at startup, so the source cluster never comes up and every test that depends on it fails immediately.

ES 7.10 and 8.19 templates correctly include these settings and are unaffected.

## Fix

Remove the two unsupported `gcs.client.*` lines (and the now-unneeded `__NAMESPACE__` sed substitution) from the ES 6.8 `elasticsearch.yml` heredoc only.

GCS snapshot migration from an ES 6.8 source can still be configured via the keystore or the cluster settings API if needed in future.

## Testing

This fixes `main-k8s-matrix-test` for all ES 6.8 → OS_x.x combinations. No other templates are changed. The ES 7.10 and 8.19 GCS configurations are untouched.


migrationConsole/lib/integ_test/testWorkflows/clusterWorkflows.yaml
```diff
@@ -603,15 +603,8 @@ spec:
             discovery.type: single-node
             xpack.security.enabled: false
             xpack.ml.enabled: false
-            # repository-gcs client settings for local GCS testing. The endpoint
```
k-rooot, line 606: 1. The same commit that added these settings (856bb9ad) also wired a gcs-test-credentials ConfigMap into the 6.8 pod: it's mounted at /gcs-credentials in the init container and declared as a volume. But unlike the 7.10/8.19 blocks, the ES 6.8 init container never runs elasticsearch-keystore add-file gcs.client.default.credentials_file ...,  and 6.8 only installs repository-s3, not repository-gcs. So after this PR, that GCS mount is dead config. I'd remove the gcs-test-credentials volume + mount from the 6.8 block too, so the cleanup is complete.


2. ES 6.8 rejects unknown settings at startup while 7.x+ silently accepts these gcs.client.* keys, and the config is duplicated per-version by hand - so it's easy to paste a 7.x-only setting into a 6.x block again.

AndreKurait (approved):

PR #3198, opened 2026-07-06
Retry cluster connection checks in reuse-clusters mode

gregschohn:
## Problem

`main-eks-full-e2e-isolated-vpc-test` build 186 failed because the pre-deployed target AOS domain was transiently unreachable at test start time. The test uses `--reuse-clusters` (long-lived domains), and `import_existing_clusters` did a single `connection_check` call with no retry — so any momentary AOS domain unavailability (restart, maintenance, brief routing hiccup) causes an immediate hard failure with a bare `AssertionError` and no details about why.

## Fix

1. **Retry up to 10 times with 30s waits** (5-minute total budget) before giving up — long enough for AOS domains to recover from typical transient issues.
2. **Include `connection_message` in the assertion failure** so logs show the actual error (timeout, connection refused, DNS failure, etc.) rather than a bare `AssertionError`.

## Why 10×30s

AOS domain restarts typically complete in 1–3 minutes. A 5-minute retry window covers that with margin, and `connection_check` itself has a 5s timeout so the total wall time for a fully-down domain is ~5min not ~50min.

### DIFF
--- a/migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
+++ b/migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
@@ -1,6 +1,7 @@
 from enum import Enum
 import logging
 import subprocess
+import time
 
 from ..cluster_version import ClusterVersion, is_incoming_version_supported
 from ..operations_library_factory import get_operations_library_by_version
@@ -10,6 +11,24 @@ from console_link.middleware.clusters import cat_indices, connection_check, clea
 from ..integration_test_argo_service import IntegrationTestArgoService
 
 logger = logging.getLogger(__name__)
+
+_CLUSTER_CONNECT_RETRIES = 10
+_CLUSTER_CONNECT_WAIT_S = 30
+
+
+def _wait_for_connection(cluster, retries: int = _CLUSTER_CONNECT_RETRIES,
+                         wait_s: int = _CLUSTER_CONNECT_WAIT_S) -> ConnectionResult:
+    """Retry connection_check to tolerate transient AOS domain unavailability."""
+    for attempt in range(1, retries + 1):
+        result = connection_check(cluster)
+        if result.connection_established:
+            return result
+        logger.warning("Connection attempt %d/%d failed for %s: %s",
+                       attempt, retries, cluster, result.connection_message)
+        if attempt < retries:
+            time.sleep(wait_s)
+    return result
+
 
 # Maps wildcard ClusterVersion (e.g. ES_7.x) to the concrete template name that exists in clusterWorkflows.yaml.
 # When minor_version is 'x', we pick the canonical representative minor version for that major.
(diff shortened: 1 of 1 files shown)


vars/k8sLocalDeployment.groovy
```diff
@@ -112,7 +112,7 @@ def call(Map config = [:]) {
 
             stage('Perform Python E2E Tests') {
                 steps {
-                    timeout(time: 5, unit: 'HOURS') {
+                    timeout(time: 6, unit: 'HOURS') {
```
k-rooot: The PR body ("Problem", "Fix", "Why 10×30s") only describes the Python retry logic - it says nothing about the timeout changes. The retry adds at most ~5 min per cluster (~10 min for both), which doesn't explain a full 1-hour bump. Either these are unrelated scope creep that should be a separate PR, or there's a real reason that needs to be in the description. Please explain why +1 hour, and why now.

vars/k8sLocalDeployment.groovy
k-rooot, line 1: PR #3200 edits this exact file to split jobs specifically so runs stay under the time limit, while this PR raises the limit.

migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
```diff
@@ -11,6 +12,24 @@
...
 
+_CLUSTER_CONNECT_RETRIES = 10
+_CLUSTER_CONNECT_WAIT_S = 30
+
+
+def _wait_for_connection(cluster, retries: int = _CLUSTER_CONNECT_RETRIES,
+                         wait_s: int = _CLUSTER_CONNECT_WAIT_S) -> ConnectionResult:
+    """Retry connection_check to tolerate transient AOS domain unavailability."""
+    for attempt in range(1, retries + 1):
+        result = connection_check(cluster)
```
k-rooot, line 30: The whole design assumes connection_check returns a ConnectionResult with connection_established=False on failure. If a transient timeout/DNS failure causes it to raise an exception instead, this loop won't catch it and the retry does nothing. Please confirm connection_check never raises on transient network errors; if it can, wrap the call in try/except inside the loop.

migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
```diff
@@ -11,6 +12,24 @@
...
 logger = logging.getLogger(__name__)
 
+_CLUSTER_CONNECT_RETRIES = 10
+_CLUSTER_CONNECT_WAIT_S = 30
+
+
+def _wait_for_connection(cluster, retries: int = _CLUSTER_CONNECT_RETRIES,
+                         wait_s: int = _CLUSTER_CONNECT_WAIT_S) -> ConnectionResult:
+    """Retry connection_check to tolerate transient AOS domain unavailability."""
+    for attempt in range(1, retries + 1):
```
k-rooot, line 29: nit: If someone calls this with retries=0, the loop body never runs and result is unbound at the return, giving a confusing NameError instead of a clean failure. Defaults are safe (10), but since retries is a public parameter, either assert retries >= 1 or initialize result = None up front.

vars/k8sLocalDeployment.groovy
gregschohn: Removed, thanks - that will make the next PR merge cleaner.

migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
gregschohn: Added the assert

migrationConsole/lib/integ_test/integ_test/test_cases/ma_argo_test_base.py
gregschohn: I've added a comment.  I'm purposefully letting unexpected exceptions go through the loop and to be handled as fatal errors here.  Each of the server interactions has its own catchall catch-block.

k-rooot (approved):

PR #3200, opened 2026-07-06
Split k8s-local-integ-test into per-source dedicated jobs

gregschohn:
## Problem

`main-k8s-local-integ-test` ran all source versions sequentially in one job, causing timeouts: OS_2.19 and OS_3.1 target runs consistently exceeded the 5-hour test stage limit (~307–335 min measured), aborting before finishing

If a failure is going to happen, we should be able to find out quickly and reproduce relatively quickly.  For tests that take only a few minutes, those remain pooled together to amortize the cost of building and deploying.

## Solution

Replace the single conjoined job with 9 per-source dedicated/small-pooled jobs that run independently and in parallel on CI:

| Job | Source | Target | Tests | Est. time |
|-----|--------|--------|-------|-----------|
| `k8s-local-es1x-test` | ES_1.5 | OS_3.1 | 0001-0003,0010 | ~48 min |
| `k8s-local-es2x-test` | ES_2.4 | OS_3.1 | 0001-0003,0010 | ~48 min |
| `k8s-local-es5x-test` *(renamed)* | ES_5.6 | OS_3.1 | 0001-0005 | ~84 min |
| `k8s-local-es6x-test` | ES_6.8 | OS_3.1 | 0001-0003,0006,0010 | ~60 min |
| `k8s-local-es7x-test` | ES_7.10 | OS_3.1 | 0001-0003,0006,0010,0020 | ~71 min |
| `k8s-local-es8x-test` *(renamed)* | ES_8.19 | OS_3.1 | 0001-0003,0020,0035,0040 | ~73 min |
| `k8s-local-os1x-test` | OS_1.3 | OS_3.1 | 0001-0003,0010 | ~20 min |
| `k8s-local-solr8x-test` *(renamed)* | SOLR_8.11 | OS_2.19 | Solr0001 | ~20 min |
| `k8s-local-solr-other-test` | SOLR_6.6,7.7,9.8 | OS_2.19 | Solr0001 | ~38 min |

Critical path across all parallel jobs: **~84 min** (ES_5.6), well within a 2-hour PR feedback window.

## Changes

- 6 new `vars/k8sLocal*Test.groovy` files (3–8 lines each, same pattern as existing)
- 6 new `jenkins/migrationIntegPipelines/*Cover.groovy` files (boilerplate)
- 3 renamed vars/ and jenkins/ files (`elasticsearch-5x` → `k8s-local-es5x`, etc.)
- `k8sLocalIntegTestCover.groovy` deleted
- `periodicCron.groovy` updated with new job names (hourly for ES/OS, 6h for Solr)
- `test_runner.py`: `--source-version` now accepts multiple values (`nargs='+'`) so `solr-other-test` can pass `SOLR_6.6 SOLR_7.7 SOLR_9.8` in one invocation
- `k8sLocalDeployment.groovy`: converts comma-separated `sourceVersion` config to space-separated CLI args

## CI config (manual step required)

The Jenkins job definitions live outside this repo. The following jobs need to be created/renamed in the CI configuration:
- Create: `main-k8s-local-es1x-test`, `main-k8s-local-es2x-test`, `main-k8s-local-es6x-test`, `main-k8s-local-es7x-test`, `main-k8s-local-os1x-test`, `main-k8s-local-solr-other-test`
- Rename: `main-elasticsearch-5x-k8s-local-test` → `main-k8s-local-es5x-test`, `main-elasticsearch-8x-k8s-local-test` → `main-k8s-local-es8x-test`, `main-solr-8x-k8s-local-test` → `main-k8s-local-solr8x-test`
- Delete: `main-k8s-local-integ-test`


vars/k8sLocalSolrOtherTest.groovy
```diff
@@ -0,0 +1,8 @@
+def call(Map config = [:]) {
+    k8sLocalDeployment(
+            jobName: config.jobName ?: 'k8s-local-solr-other-test',
+            sourceVersion: 'SOLR_6.6,SOLR_7.7,SOLR_9.8',
```
k-rooot, line 4: The motivation for this PR is that one broken source shouldn’t turn the entire job red without making it obvious which source failed. 

Here we're bundling 3 solr sources into a single job. If SOLR_7.7 breaks, the job goes red and we won't know which one failed without digging into logs - the same opaque failure we're fixing.

libraries/testAutomation/testAutomation/test_runner.py
```diff
@@ -547,7 +547,8 @@ def _generate_unique_id() -> str:
 
 
 def get_version_combinations(source_version, target_version, target_type):
-    source_list = VALID_SOURCE_VERSIONS if source_version == "all" else [source_version]
+    versions = source_version if isinstance(source_version, list) else [source_version]
+    source_list = VALID_SOURCE_VERSIONS if versions == ["all"] or versions == "all" else versions
```
k-rooot: `versions` is always a list after line 550, so `versions == "all"` can never be true - that half of the `or` is dead code.

Also, this silently mishandles a mixed input like ["all", "ES_5.6"] - it won't expand "all" and will instead pass the literal string "all" as a version.

libraries/testAutomation/tests/test_test_runner.py
```diff
@@ -215,3 +215,11 @@ def test_aoss_all_sources(self):
         combos = get_version_combinations("all", "OS_2.19", TargetType.AOSS)
         assert all(t == "AOSS" for _, t in combos)
         assert len(combos) == len(VALID_SOURCE_VERSIONS)
+
+    def test_multi_source_list(self):
```
k-rooot, line 219: Since the CLI default is now a list and nargs='+' always yields a list, the ["all"] path is the one that actually runs in production and we don't have test coverage for this scenario. Can we add UT coverage for this?

vars/k8sLocalDeployment.groovy
```diff
@@ -118,6 +118,8 @@ def call(Map config = [:]) {
                                 def sourceVer = sourceVersion ?: params.SOURCE_VERSION
                                 def targetVer = targetVersion ?: params.TARGET_VERSION
                                 currentBuild.description = "${sourceVer} → ${targetVer}"
+                                // --source-version accepts one or more space-separated values; convert commas for multi-source jobs
+                                def sourceVerArg = sourceVer.replace(',', ' ')
```
k-rooot: sourceVer comes from `sourceVersion ?: params.SOURCE_VERSION`, which can be `null`. The old code interpolated `null` gracefully into the string; now `.replace()` on `null` throws a `NullPointerException` and fails the build.

vars/k8sLocalElasticsearch1xTest.groovy
```diff
@@ -0,0 +1,8 @@
+def call(Map config = [:]) {
+    k8sLocalDeployment(
+            jobName: config.jobName ?: 'k8s-local-elasticsearch1x-test',
```
k-rooot, line 3: nit: The PR description table lists jobs as `k8s-local-es1x-test`, `k8s-local-es5x-test`, etc., but the code uses `k8s-local-elasticsearch1x-test`.

This also has to line up exactly with the cron keys in periodicCron.groovy (`main-k8s-local-elasticsearch1x-test`) and the real Jenkins job names, or the cron schedule silently won't apply. Please make the description match the code and double-check the Jenkins job names are the elasticsearch1x form, not es1x.

vars/periodicCron.groovy
```diff
@@ -18,12 +18,18 @@ def call(String jobName) {
...
-        case 'main-elasticsearch-5x-k8s-local-test':  return '@hourly'
-        case 'main-elasticsearch-8x-k8s-local-test':  return '@hourly'
+        case 'main-k8s-local-elasticsearch1x-test':              return '@hourly'
+        case 'main-k8s-local-elasticsearch2x-test':              return '@hourly'
+        case 'main-k8s-local-elasticsearch5x-test':              return '@hourly'
+        case 'main-k8s-local-elasticsearch6x-test':              return '@hourly'
+        case 'main-k8s-local-elasticsearch7x-test':              return '@hourly'
+        case 'main-k8s-local-elasticsearch8x-test':              return '@hourly'
+        case 'main-k8s-local-opensearch1x-test':              return '@hourly'
+        case 'main-k8s-local-solr8x-test':            return 'H H(0-23)/6 * * *'
```
k-rooot, line 28: nit: The returns no longer line up (the opensearch1x and ES cases have extra padding vs the solr ones).

jenkins/migrationIntegPipelines/k8sLocalElasticsearch1xTestCover.groovy
```diff
@@ -9,4 +9,4 @@ library identifier: "migrations-lib@${gitBranch}", retriever: modernSCM(
 // This ensures the GenericTrigger regex stays in sync with the actual job name
 // (e.g., main-* vs pr-*) across runs.
 def jobNameOverride = params.JOB_NAME_OVERRIDE ?: env.JOB_BASE_NAME ?: ''
-elasticsearch8xK8sLocalTest(jobName: jobNameOverride ?: null)
+k8sLocalElasticsearch1xTest(jobName: jobNameOverride ?: null)
```
k-rooot, line 12: The old k8sLocalIntegTestCover.groovy (which called k8sLocalDeployment directly) and the main-k8s-local-integ-test / elasticsearch-5x-k8s-local-test / solr-8x-k8s-local-test job names are gone. Please grep the repo (docs, k8s-matrix-test, README badges, any dashboard config) for the old names to make sure there are no dangling references now pointing at jobs that no longer exist.

vars/k8sLocalSolrOtherTest.groovy
gregschohn: I've updated the PR description with a justification as to why these are still pooled together.  I've also removed the PR description advocating the signal point.  I'm not holding onto this opinion too strongly, so I'm happy to break this out into separate tests too - it's a bit more strain on the hardware, doesn't reduce the critical path, but DOES give a better signal (& a more consistent pattern).  

Let me know what you think.

.github/workflows/jenkins_tests.yml
```diff
@@ -44,60 +44,189 @@ jobs:
...
           api_token: "${{ secrets.JENKINS_MIGRATIONS_GENERIC_WEBHOOK_TOKEN }}"
           job_params: "GIT_REPO_URL=${{ needs.sanitize-input.outputs.pr_repo_url }},GIT_BRANCH=${{ needs.sanitize-input.outputs.branch_name }},GIT_COMMIT=${{ needs.sanitize-input.outputs.commit_hash }}"
           jenkins_user: "${{ secrets.JENKINS_MIGRATIONS_USER }}"
           jenkins_api_token: "${{ secrets.JENKINS_MIGRATIONS_API_TOKEN }}"
           job_timeout_minutes: "120"
 
-  elasticsearch-8x-k8s-local-test:
+  k8s-local-elasticsearch2x-test:
     needs: [get-require-approval, sanitize-input]
+    continue-on-error: true  # TODO: remove once Jenkins jobs are created
```
AndreKurait: We don't fail or block merge on Jenkins so I'm not sure what this gives us even in the short term

AndreKurait (approved):

.github/workflows/jenkins_tests.yml
gregschohn: agree - rolled that back

PR #3201, opened 2026-07-07
Add external Kafka client properties

gregschohn:
Implement the Kafka client configuration design for bring-your-own Kafka clusters. External Kafka configs now accept separate producer and consumer clientProperties maps, validate them before workflow submission, reject workflow- and application-owned keys, and render deterministic Java .properties file bodies during config processing.

The design (docs/KafkaClientConfigurationProposal.md) intentionally limits this raw property surface to existing Kafka clusters. Workflow-managed Strimzi Kafka still uses broker/topic/resource-oriented settings so future high-level migration settings can derive coordinated broker, topic, producer, and consumer limits without asking users to synchronize raw Kafka knobs manually.

orchestrationSpec changes:
* .../schemas adds the user-facing schema for clientProperties plus rendered internal Kafka client property fields. It also exports shared Kafka client file path constants so config-processor-rendered paths stay bound to the workflow-template mounts.
*  .../config-processor resolves the final producer and consumer property files. It validates flat scalar property names and values, rejects protected keys based on the resolved Kafka auth/connection model, merges workflow-generated TLS properties with user tuning values, and renders stable Java properties text.
* .../migration-workflow-templates changes setupCapture and replayer to consume the pre-rendered property file bodies, mount them through the existing workflow-owned ConfigMap path, and pass the capture proxy/replayer property-file arguments whenever the rendered content is non-empty.

migrationConsole integration coverage adds an explicit CDC test that provisions Strimzi Kafka outside the migration workflow, runs CDC against it as an external Kafka cluster, verifies the generated client ConfigMaps, and confirms the migration workflow does not create its own KafkaCluster CR. The cdcOnlyImportedClusters workflow is parameterized so that test can inject the external Kafka config.

Jenkins pipeline files add a focused k8s-local wrapper for the slower explicit external-Kafka CDC test instead of adding it to the broad version matrix.

Tests cover schema validation, protected-key rejection, renderer escaping, transformer output, workflow snapshots, Python test import wiring, and the new explicit integration path.

### Check List


orchestrationSpecs/packages/config-processor/src/kafkaClientProperties.ts
```diff
@@ -0,0 +1,169 @@
...
+export type KafkaClientProperties = Record<string, KafkaClientPropertyValue>;
+export type KafkaClientRole = "producer" | "consumer";
+
+type KafkaClientConfig = z.infer<typeof KAFKA_CLIENT_CONFIG>;
+type ProtectedKeyReasons = Map<string, string>;
+
+// Protected keys are owned by the workflow or by the producer/consumer app contract.
+// They are protected by rejecting user-provided file entries during config processing,
+// before the rendered properties file is mounted into the Argo workflow.
+function addAuthProtectedKeys(keys: ProtectedKeyReasons, kafkaConfig: KafkaClientConfig): void {
```
k-rooot, line 14: sasl.jaas.config, security.protocol, and sasl.mechanism are only protected when auth is already enabled. For a no-auth external cluster (authType: "none", no MSK), a user can set sasl.jaas.config freely and JAAS config strings can load arbitrary login modules (the classic JNDI/LDAP → RCE surface), interpreted inside the capture-proxy / replayer pods.

For bring-your-own-Kafka this may be an acceptable trust model, but it should be a conscious decision, not a side effect of the condition above.

I'd suggest either (a) protecting sasl.jaas.config / security.protocol / sasl.mechanism unconditionally, or (b) documenting explicitly in the proposal doc that the raw-property surface is intentionally trusted and why.

vars/externalKafkaK8sLocalTest.groovy
```diff
@@ -0,0 +1,8 @@
+def call(Map config = [:]) {
+    k8sLocalDeployment(
+            jobName: config.jobName ?: 'external-kafka-k8s-local-test',
```
k-rooot, line 3: 1. Not in periodicCron.groovy. Every other k8s-local-* job has a cron entry; this one has none, so external-kafka-k8s-local-test will never run on a schedule - only when triggered manually. If the intent is periodic CDC coverage, it's missing its cron line.
2. PR #3200 renames all these wrappers to the k8s-local-<source>-test convention and rewrites periodicCron.groovy. This PR adds external-kafka-k8s-local-test in the old <name>-k8s-local-test form and adds files in the same vars/ and jenkins/migrationIntegPipelines/ dirs #3200 is restructuring.

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
k-rooot, line 682: Editing clientProperties won't restart pods. Rendered props aren't in any checksum, and the ConfigMap has no pod-template hash annotation -> edit updates the ConfigMap but proxy/replayer keep the old value until an unrelated reconcile.

Should we fold the rendered props into the proxy/replayer checksum or stamp a properties-hash annotation on the pod template?

orchestrationSpecs/packages/schemas/src/userSchemas.ts
k-rooot, line 370: No "TLS without SASL" auth type. Only none | scram-sha-512, so a plain-TLS external broker has zero protected TLS keys.

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
```diff
@@ -584,6 +600,19 @@ function buildKafkaClientConfig(
         topicSpecOverrides: cluster.autoCreate.topicSpecOverrides,
         label: kafkaClusterKey
     };
+    return {
```
k-rooot, line 603: For autoCreate clusters we call buildKafkaClientPropertiesFileContent with no userProperties, so it only ever emits the workflow-generated truststore lines (or ""). But it means the protected-key logic and validators run against an always-empty map for managed clusters. More notably, it couples managed-cluster truststore rendering to the same renderer whose escaping rules are only tested via the mirror-decoder.If the renderer regresses, it now breaks managed SCRAM clusters too, not just the new BYO-Kafka feature.

PR #3203, opened 2026-07-08
Add opt-in begin approval gate

gregschohn:
## What changed

This is related to the final ask brought up in https://github.com/opensearch-project/opensearch-migrations/pull/3169.

- Adds an opt-in root `requireBeginApproval` setting.
- Propagates that setting into the transformed Argo migration config.
- Creates a `begin` ApprovalGate for the MigrationRun when the setting is true.
- Adds `approveBegin` as the first `main` workflow step, before the startup/migration step group begins.
- Threads the setting through the CDC-with-clusters integration workflow.
- Extends `Test0003ApprovalGateIntegration` to approve the `begin` gate first.
- Hardens the same integration test to confirm the capture proxy is not Ready at the begin gate, then waits for the proxy to become Ready after the proxy setup gate is approved.

## Validation

- `npm run -w @opensearch-migrations/schemas test`
- `npm run -w @opensearch-migrations/migration-workflow-templates test`
- `npm run -w @opensearch-migrations/config-processor test -- --runTestsByPath tests/migrationInitializer.customMigrationResources.test.ts`
- `python3 -m py_compile migrationConsole/lib/integ_test/integ_test/test_cases/basic_tests.py`
- `npm run make-templates`
- `git diff --check`


orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
```diff
@@ -772,7 +772,12 @@ export const FullMigration = WorkflowBuilder.create({
         .addRequiredInput("uniqueRunNonce", typeToken<string>())
         .addInputsFromRecord(defaultImagesMap(t.inputs.workflowParameters.imageConfigMapName))
 
-        .addSteps(b => b.addStepGroup(g => g
+        .addSteps(b => b
+            .addStep("approveBegin", ResourceManagement, "waitForUserApproval", c =>
```
jzonthemtn, line 775: Am I reading this right that the `approveBegin` step ignores `skipApprovals`?

jzonthemtn (approved):
Looks good. Left one question about the begin approval gate not respecting `skipApprovals` (assuming I interpreted that part of the code correctly).

AndreKurait (approved):
@jzonthemtn left a good comment about the interaction with skipApprovals and require begin approval. Do we have tests that go through all 4 combinations?

gregschohn:
> @jzonthemtn left a good comment about the interaction with skipApprovals and require begin approval. Do we have tests that go through all 4 combinations?

Thanks - the last commit added exhaustive tests for a bunch of combinations

orchestrationSpecs/packages/migration-workflow-templates/src/workflowTemplates/fullMigration.ts
gregschohn: Thanks!  I've removed the dig statement and now am making the field required in the incoming schema.  The configTransformer is responsible for populating all skip flags appropriately from the global default

PR #3204, opened 2026-07-08
Fix cancelled-session permit leak after partition rebalance

Aman-bv:
### Description

Fixes the cancelled-session permit leak that causes the traffic replayer to stop consuming after Kafka consumer group rebalances.

#### The Bug

After a partition reassignment, `cancelConnection()` marks sessions as cancelled and drains the `OnlineRadixSorter`. However, requests holding `TrafficStreamLimiter` permits were **never completed** in two places:

1. **Send-schedule timers** — requests queued in `TimeToResponseFulfillmentFutureMap` (waiting for Netty timer turn)
2. **Transformation-phase timers** — standalone timers created by `scheduleWork` (used by `ReplayEngine.scheduleTransformationWork`), deliberately NOT registered in `session.schedule`

These held `TrafficStreamLimiter` permits permanently, starving the consumer thread.

#### The Fix

1. **`drainWithCancellation()`** — snapshots the `TimeToResponseFulfillmentFutureMap` deque, clears it, then completes all futures exceptionally (snapshot avoids `ConcurrentModificationException` from reentrant `removeFirstItem()` in synchronous `whenComplete` callbacks)
2. **`drainTransformationTimers()`** — new `ConcurrentLinkedQueue<CompletableFuture>` on `ConnectionReplaySession` tracks standalone transformation timers; `cancelConnection()` drains them immediately
3. **`isCancelled()` guards** — in `scheduleWork` (transformation callback) and `submitUnorderedWorkToEventLoop` (sorter callback) to prevent cascading work after cancellation
4. **Null guard in `whenComplete`** — handles the case where `removeFirstItem()` returns null because the deque was already drained
5. **Span leak fix** — `scheduledContext` in `scheduleSendRequestOnConnectionReplaySession` is now closed on the exceptional path too (previously only inside `thenCompose`, which is skipped on drain)

<details>
<summary>Cancel flow diagram</summary>

```mermaid
sequenceDiagram
    participant Kafka as Kafka Consumer Thread
    participant Pool as ClientConnectionPool
    participant Session as ConnectionReplaySession
    participant EL as Netty EventLoop
    participant Limiter as TrafficStreamLimiter

    Note over Kafka: Partition reassignment triggers
    Kafka->>Pool: cancelConnection(ctx, sessionNumber)
    Pool->>Session: setCancelled(true)

    Note over Pool: Drain transformation timers (thread-safe, immediate)
    Pool->>Session: drainTransformationTimers(cause)
    Session->>Session: poll pendingTransformationTimers queue
    Session-->>Limiter: completeExceptionally → whenComplete → permit.release()

    Note over Pool: Drain send-schedule on event loop
    Pool->>EL: submit(drain task)
    EL->>Session: schedule.drainWithCancellation(cause)
    Note over Session: snapshot deque → clear → completeExceptionally each
    Session-->>Limiter: completeExceptionally → whenComplete → permit.release()
    EL->>Session: scheduleSequencer.cancelAllWork()

    Note over Pool: Close channel and invalidate cache
    Pool->>Pool: closeConnection(ctx, sessionNumber)

    Note over EL: Late timer fires (race window)
    EL->>EL: isCancelled() guard → fail fast
    EL-->>Limiter: CancellationException → whenComplete → permit.release()
```

</details>

### Issues Resolved

- Consumer thread starvation due to leaked `TrafficStreamLimiter` permits after partition rebalance
- Replayer consumer group eviction cascade (no polls → session timeout → rebalance → more leaks)
- Tracing span leak on cancel path (one span per in-flight request per rebalance)

### Testing

- `CancelledSessionPermitLeakTest.cancelledSession_releasesAllPermits` — schedules 5 requests at far-future timestamps, cancels connection, asserts all permits release within 2s
- `CancelledSessionPermitLeakTest.cancelledSessionBeforeSchedule_failsFastOnSorterCallback` — exercises `isCancelled()` in `submitUnorderedWorkToEventLoop`
- `CancelledSessionPermitLeakTest.cancelAfterSorterButBeforeTimerFire_failsFastInScheduleWork` — exercises `isCancelled()` in `scheduleWork` (timer fires naturally after cancel)
- `CancelledSessionPermitLeakTest.cancelConnection_drainsTransformationTimers` — exercises the transformation timer drain path via `scheduleWork` + `cancelConnection`
- `CancelConnectionDrainTest.cancelConnection_drainsSorterSlots` — verifies sorter is empty after cancel

### Check List


Aman-bv:
Also noting:
Even though we had 16 replayer instances running, the partitions got assigned to just 4 of them or sometimes even less. Having a concurrency of 500 for each container on a single host created a lot of requests in flight, which in turn triggered back pressure and stopped polling and entered a deadlock talked about in the previous PR #3196 

That consequently kicked off the consumer from the consumer group and triggered rebalancing where the bug in this PR came into play.

Rebalancing and assigning partitions or handling them could be improved further.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
```diff
@@ -147,10 +147,15 @@ public TrackedFuture<String, Void> cancelConnection(IReplayContexts.IChannelKeyC
...
+            // from leaving requestWorkTracker entries and TrafficStreamLimiter slots unreleased.
+            var cancellationCause = new java.util.concurrent.CancellationException(
+                "Session cancelled due to partition reassignment for " + connId);
+            session.eventLoop.submit(() -> {
+                session.schedule.drainWithCancellation(cancellationCause);
+                session.scheduleSequencer.cancelAllWork();
+            });
             closeConnection(ctx, sessionNumber);
         }
         return TextTrackedFuture.completedFuture(null, () -> "cancelled");
```
Aman-bv, line 166: **Code Review Finding (pre-existing, informational):**

`cancelConnection()` returns a pre-completed future but the actual drain runs async on the Netty event loop. Callers decrement `totalCountOfScheduledTasksOutstanding` before drain completes, so there is a window where the orchestrator believes all work is done while timer futures are still being cancelled.

Not introduced by this PR — noting for future hardening.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
```diff
@@ -147,10 +147,15 @@ public TrackedFuture<String, Void> cancelConnection(IReplayContexts.IChannelKeyC
...
-            // entries and TrafficStreamLimiter slots unreleased.
-            session.eventLoop.submit(() -> session.scheduleSequencer.cancelAllWork());
+            // Drain all pending schedule and sorter slots on the event loop thread so they
+            // complete exceptionally immediately. This prevents orphaned scheduleFuture entries
+            // from leaving requestWorkTracker entries and TrafficStreamLimiter slots unreleased.
+            var cancellationCause = new java.util.concurrent.CancellationException(
+                "Session cancelled due to partition reassignment for " + connId);
+            session.eventLoop.submit(() -> {
+                session.schedule.drainWithCancellation(cancellationCause);
+                session.scheduleSequencer.cancelAllWork();
```
Aman-bv, line 157: **Code Review Finding (pre-existing, informational):**

`OnlineRadixSorter.cancelAllWork()` (line 175 in `OnlineRadixSorter.java`) completes `signalWorkCompletedFuture` with `null` (success) rather than exceptionally. Downstream processors that check for exceptional completion to detect cancellation will treat a cancelled session as if it completed normally.

Consider using `completeExceptionally(new CancellationException(...))` instead of `complete(null)` so that consumers of the sorter output can distinguish cancellation from success.

Not introduced by this PR — noting for future hardening.

gregschohn:
Thanks for the feedback!  I'm still in the process of reviewing this.  If you separate the two bugs into two separate PRs, we might be able to get the first bugfix merged quicker.

CLOSED_PREMATURELY should only be used during the shutdown sequence where we're really just dropping things on the floor and definitely don't want to commit those.  It sounds like the concept of CLOSED_PREMATURELY and EXPIRED_PREMATURELY might be conflated in the remedy for bug #2.

I'd like to focus on bug #1 first, then get an understanding for how to reproduce bug #2 and to consider the fix.  I'll do both of these in the interim anyway.

gregschohn:
I used claude's "ultra-code" to do a review on this PR - I've already spent a considerable amount of time modeling this and working through some of these issues.  I obviously still forgot about some things.  Here's the full report that came back.  

For the following report: I disagree with the severity of the first two minor issues concerning bug #2's fix - just because it doesn't cause data loss today shouldn't stop us from building a system that we can reason about.  I would like to see a reproduction pattern for why the CLOSED_PREMATURELY might be contributing to a stuck commit cursor.  For the second issue, the leak discovered, that could eventually be a real problem that adds up with lots of rebalances.

---

I now have a complete, cross-verified picture. The concurrency dimension stalled in the workflow, but I've analyzed it directly myself (the drain-then-clear task ordering is FIFO-safe on the single-threaded event loop, the snapshot prevents CME, and `completeExceptionally` is idempotent so no double-completion — no defects there). Every substantive claim is grounded in code I or a verifier read, and the one alarming hypothesis (data loss) was adversarially refuted. Here's the review.

---

# Review: PR #3204 — Fix replayer cancelled-session leak & offset blocking

**Verdict: Request changes.** The direction is right and the diagnosis of Bug 1 is real, but the fix for Bug 1 is **incomplete** (it leaves the larger part of the permit-holding window undrained), and Bug 2's change is justified by a **factually incorrect comment** and may not actually address the production symptom it targets. None of the issues are data-corrupting — the scariest hypothesis (data loss on restart) was checked and **refuted**.

Methodology note: I reviewed across 5 dimensions (concurrency, offset-correctness, resource-lifecycle, general-correctness, tests) with each finding independently, adversarially verified against the checked-out branch. A transient auth outage killed part of the first pass; I re-ran the affected dimensions and verified the concurrency dimension by hand.

## 🔴 Major — Bug 1's fix is incomplete: transformation-phase timer is never drained

`RequestSenderOrchestrator.java:96` (`scheduleWork`), `ClientConnectionPool.java:155-158`

The permit is acquired *before* transformation (`TrafficReplayerCore.java:172`), and a request's lifecycle is:

1. `scheduleWork` schedules a **standalone Netty timer** at `start − EXPECTED_TRANSFORMATION_DURATION` (`ReplayEngine.java:167`, i.e. ~`start−1s`). This timer is deliberately **not** registered in `session.schedule` nor in `scheduleSequencer` (see the comment at `RequestSenderOrchestrator.java:110-119`).
2. Only after transformation does `scheduleRequest` schedule the send timer at `start` — this one *is* in `session.schedule`.

`cancelConnection()` drains **only** `session.schedule.drainWithCancellation()` and `scheduleSequencer.cancelAllWork()`. The transformation timer from step 1 is untouched, so a request cancelled while waiting in transformation holds its permit until that far-future timer fires **naturally**. The new `isCancelled()` guard at `scheduleWork` line 128 runs *when the timer fires* — it prevents the work from executing but does nothing to release the permit early.

Because the send window is only ~1s wide while the transformation window spans `[dequeue, start−1s]`, under the exact scheduling-lag scenario this PR targets (timers hours in the future), **the undrained window is the majority of the permit's lifetime**. The consumer-starvation the PR set out to eliminate persists for any request cancelled during transformation.

This is also why the regression test passes despite the gap: `CancelledSessionPermitLeakTest` calls `orchestrator.scheduleRequest()` directly (lines 98, 168, 211), bypassing `ReplayEngine.scheduleTransformationWork` → `scheduleWork` entirely, so it never exercises the leaking path.

**Fix:** track `scheduleWork`'s `ScheduledFuture` (or its trigger future) on the `ConnectionReplaySession` so `cancelConnection` can complete/cancel it exceptionally, or route `scheduleWork` through `session.schedule` so `drainWithCancellation` covers it. Then add a test that drives the full `ReplayEngine` path.

## 🟡 Minor — Bug 2: the justifying comment is wrong, and the change may not fix the described symptom

`TrafficReplayerCore.java:347-355`

The new comment says CLOSED_PREMATURELY is *"packet timeout expired on incomplete data … Retrying is pointless — the same incomplete fragments will time out again."* This is inaccurate:

- The packet/connection **timeout** status is `EXPIRED_PREMATURELY` (`CapturedTrafficToHttpTransactionAccumulator.java:~157`, `onExpireAccumulation`), and it **already committed** both before and after this PR.
- `CLOSED_PREMATURELY` is fired **only** by `accumulator.close()` (`…Accumulator.java:640`), which is called **only** at terminal shutdown (`TrafficReplayerTopLevel.java:258`, `finally` block). It is not a timeout, and on restart the records *can* be re-read and complete — so "retrying is pointless" is backwards.

The concerning implication, which the author should confirm: since CLOSED_PREMATURELY fires only at shutdown, and the running-consumer head-of-line case is EXPIRED_PREMATURELY (already committing), **this change may not address the "infinite lag after rebalance" symptom described in the PR body.** Please pin down the exact production repro — which `ReconstructionStatus` and which callback path actually blocked the offset? If it was EXPIRED_PREMATURELY, this change is orthogonal to it.

**Good news (adversarially verified, and the reason this is only Minor):** committing on CLOSED_PREMATURELY is **not** a data-loss risk. `commitTrafficStream` → `commitKafkaKey` only *stages* offsets into `nextSetOfCommitsMap`; the broker flush (`commitSync`) happens only in `touch()`/`cleanupRevokedPartitions()`/`getNextBatchOfRecords()`. At shutdown those never run again (close happens after the poll loop exits), and `enable.auto.commit=false` (`KafkaTrafficCaptureSource.java:280`) means `kafkaConsumer.close()` does **not** flush the staged commit. So the CLOSED_PREMATURELY commit is staged and silently dropped — restart behavior is identical to the old code. At worst this change is a no-op at shutdown.

**Fix:** correct the comment (CLOSED_PREMATURELY = accumulator close at shutdown, not a timeout; drop "retrying is pointless"), and clarify in the PR description what the real production trigger was.

## 🟡 Minor — Tracing span leak on the cancel/drain path

`RequestSenderOrchestrator.java:313, 324-330`

`scheduledContext = ctx.createScheduledContext(startTime)` (line 313) opens an OTel span registered in `ActiveContextTracker`. It's closed at line 325 — but that `close()` lives **inside** `trigger.thenCompose(voidVal -> { scheduledContext.close(); … })`. `thenCompose`'s function is skipped on exceptional upstream completion, and the new `drainWithCancellation` completes the trigger **exceptionally**. So on cancel the span is never `.end()`ed and stays in the tracker's `ConcurrentSkipListSet` — one leaked span per in-flight scheduled request, per rebalance. (`packetProducer` is still released by the outer `whenComplete` at line 332, so no ByteBuf leak.) Contrast `scheduleWork` at line 125, which closes its context unconditionally and is unaffected.

**Fix:** close `scheduledContext` on exceptional completion too, e.g. `trigger.whenComplete((v,t) -> { if (t != null) scheduledContext.close(); })`, ensuring exactly-once close.

## 🟡 Minor — Headline Bug-2 test is tautological

`ClosedPrematurelyOffsetBlockingTest.java:105-117, 146`

`closedPrematurelyStatus_commitsOffset_unblockingSubsequentRecords` re-implements the fix predicate *inside its own anonymous callback* (`if (status != TRAFFIC_SOURCE_READER_INTERRUPTED) commitCalls.add(true)`), so the key assertion at line 146 passes regardless of production code — reverting the fix leaves it green. Production `commitTrafficStreams`/`removeAndReturnNewHead` are never invoked, despite the class docstring claiming otherwise. Only the `assertEquals(CLOSED_PREMATURELY, …)` line is meaningful. **Not a coverage gap** — `CommitTrafficStreamsStatusTest` does exercise the real path — but this test gives false confidence and should be rewired to call `TrafficReplayerAccumulationCallbacks` with a mock source, or deleted as redundant. Fix the docstring's `removeAndReturnNewHead` claim too.

## 🟢 Minor / nit — Test timing

- `CancelledSessionPermitLeakTest.java:~211` (`cancelAfterSorterButBeforeTimerFire_failsFastInScheduleWork`) relies on a 300ms margin (500ms timer vs 800ms `sleep`). The `eventLoop.submit().sync()` barrier doesn't fully protect it: under a GC/CPU stall the no-op barrier task can be polled before the due timer in the same batch. Low-probability CI flake. Prefer a `CountDownLatch`/`CompletableFuture` completed inside `whenComplete` and awaited.
- `Thread.sleep(200)` in the fast-path test (`~line 187`) is redundant (the `sync()` barrier already guarantees completion) — harmless, not flaky. Could drop it.

## ✅ Verified clean

- **Concurrency of the cancel path** (analyzed directly, since this dimension stalled in automation): the FIFO ordering of event-loop tasks means `drainWithCancellation` runs *before* `schedule.clear()` (which only runs after the async channel-close completes), so `clear()` can't drop futures out from under the drain. The `ArrayList` snapshot in `drainWithCancellation` correctly avoids the CME from reentrant `removeFirstItem()`, and `completeExceptionally` is idempotent so a racing natural timer fire is a safe no-op. `OnlineRadixSorter.cancelAllWork` iterates a snapshot, so `items.remove` in the cleanup callback is CME-safe.
- **`removeFirstItem()` null-guard** (`RequestSenderOrchestrator.java:386`) and the volatile `cancelled` happens-before (established by `eventLoop.submit`) are correct.
- **Permit release on the drained path**: the exceptional completion propagates through to the `whenComplete` at `TrafficReplayerCore.java:178` (`doneProcessing`), releasing the permit exactly once — this part of Bug 1 works.

---

**Bottom line:** merge-blocking item is the incomplete Bug-1 fix (transformation timer) — it means the primary bug is only partially resolved and the test masks it. The Bug-2 change is safe but rests on a wrong comment and an unverified premise about the production symptom; please clarify. The span leak and test-quality items are worth addressing but not blocking.

Want me to draft the specific code changes for the transformation-timer drain and the span-close fix, or post these as inline PR comments?

Aman-bv:
@gregschohn Thanks for the thorough review. Here's where things stand:

**Bug 2 split out** — moved to PR #3207 as requested. This PR now contains only the Bug 1 fix.

**All findings from your review are addressed in the latest commits:**

| Finding | Status |
|---------|--------|
| ❗ Major — transformation-phase timer never drained | Fixed: added `pendingTransformationTimers` queue + `drainTransformationTimers()` + dedicated test (`cancelConnection_drainsTransformationTimers`) |
| ❗ Span leak on cancel path | Fixed: `AtomicBoolean` guard ensures `scheduledContext.close()` is called exactly once (success path OR error path, not both) |
| ❗ Tautological test | That was Bug 2's test — lives in PR #3207 now |
| ❗ Test timing / Thread.sleep | Fixed: replaced all `Thread.sleep` with `CountDownLatch` await patterns for deterministic synchronization |
| ❗ O(n) queue remove during drain | Fixed: conditional remove only on normal completion (skip during drain since `poll()` already removed the item) |

Ready for re-review when you get a chance.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/RequestSenderOrchestrator.java
```diff
@@ -120,13 +121,26 @@ public <T> TrackedFuture<String, T> scheduleWork(
         //
         // Making them more independent means that the work item being enqueued is lighter-weight and
         // less likely to cause a connection timeout.
-        return bindNettyScheduleToCompletableFuture(connectionSession.eventLoop, timestamp)
+        var timerFuture = bindNettyScheduleToCompletableFuture(
+            connectionSession.eventLoop, timestamp);
+        connectionSession.pendingTransformationTimers.add(timerFuture.future);
+        timerFuture.future.whenComplete((v, t) -> {
+            if (t == null) {
+                connectionSession.pendingTransformationTimers.remove(timerFuture.future);
```
gregschohn: Is there a downside to making pendingTransformationTimers 1) hidden within the session (connectionSession.remove(timerFuture)) and 2)putting it within a ConcurrentHashMap (as a key) for quicker removal?

gregschohn (commented):
Thanks for this PR!  I have one minor non-blocking change.  I'll approve this once the CI test runs finish (spurious maven download error nailed it before... now merging from main).

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/RequestSenderOrchestrator.java
Aman-bv: No downside, good call. Done in [076d8b0](https://github.com/opensearch-project/opensearch-migrations/pull/3204/commits/076d8b0a3af09598c611ea587b327ea8e2e87911): made the collection private with add/remove/has/drain methods on the session, and switched to a `Set` backed by `ConcurrentHashMap` for O(1) removal on the normal completion path.

gregschohn (approved):

TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/kafka/CancelledSessionPermitLeakTest.java
```diff
@@ -0,0 +1,299 @@
+package org.opensearch.migrations.replay.kafka;
```
jzonthemtn, line 1: Super minor - is just `org.opensearch.migrations.replay` a more appropriate package?

jzonthemtn (approved):

PR #3207, opened 2026-07-09
Fix CLOSED_PREMATURELY offset commit suppression causing head-of-line blocking

Aman-bv:
## Summary

Fixes offset commit suppression for `CLOSED_PREMATURELY` status that causes permanent head-of-line blocking in `OffsetLifecycleTracker`. Adds two defensive mechanisms (wall-clock expiry watchdog and stale-head reaper) to prevent similar commit-pointer stalls from other causes. Fixes a backpressure deadlock caused by time-shifted close events and hardens consumer keepalive during backpressure.

### Problem overview

```mermaid
flowchart TD
    A[Offset added to OffsetLifecycleTracker PriorityQueue] --> B{Does the offset complete normally?}
    B -->|Yes| C[removeAndReturnNewHead called → commit pointer advances]
    B -->|No| D{Why not?}

    D -->|accumulator.close at shutdown| E[Status = CLOSED_PREMATURELY]
    D -->|Connection stale, holds permits| F[Main thread blocked in blockIfNeeded]
    D -->|Callback never fires| G[Offset orphaned at head of queue]
    D -->|Close event time-shifted to future| H[hookWorkFinishingUpdates never runs]

    E -->|ON MAIN| E1[❌ Commit suppressed → offset stuck forever]
    E -->|THIS PR| E2[✅ Bug Fix: commit fires, offset removed]

    F -->|ON MAIN| F1[❌ Event-driven expiry needs new data → deadlock]
    F -->|THIS PR| F2[✅ Wall-clock watchdog expires stale connection → permits freed]

    G -->|ON MAIN| G1[❌ No recovery mechanism → stuck forever]
    G -->|THIS PR| G2[✅ Stale-head reaper evicts after 5 min]

    H -->|ON MAIN| H1[❌ stopReadingAt frozen → backpressure permanent]
    H -->|THIS PR| H2[✅ Close events fire immediately]

    style E1 fill:#ffcccc
    style F1 fill:#ffcccc
    style G1 fill:#ffcccc
    style H1 fill:#ffcccc
    style E2 fill:#ccffcc
    style F2 fill:#ccffcc
    style G2 fill:#ccffcc
    style H2 fill:#ccffcc
```

---

## Bug Fix: `CLOSED_PREMATURELY` suppresses offset commits

**On main:** `TrafficReplayerCore.commitTrafficStreams()` skips the Kafka offset commit when status is `CLOSED_PREMATURELY`.

**Why it's wrong:** `CLOSED_PREMATURELY` fires when `accumulator.close()` is called at shutdown with incomplete streams. These offsets have been fully read from Kafka — they just didn't complete the HTTP transaction. Suppressing the commit leaves the offset stuck in `OffsetLifecycleTracker`'s PriorityQueue. Since only the head (lowest offset) can advance the commit pointer, one stuck offset blocks ALL subsequent offsets on that partition permanently.

**Correct semantic:** Only `TRAFFIC_SOURCE_READER_INTERRUPTED` should suppress commits — that fires on partition reassignment where re-delivery to the new owner is required.

**Change:** Allow `CLOSED_PREMATURELY` to commit.

---

## Bug Fix: Backpressure deadlock from time-shifted close events

**On main:** `ReplayEngine.closeConnection()` time-shifts the close event via `timeShifter.transformSourceTimeToRealTime(timestamp)`, scheduling it on a Netty timer that fires minutes in the future (e.g., 18min wall-clock at 10x speedup for a 3h source-time offset).

**Problem this solves:** When the Kafka backlog contains only connection-close events (no request data), all scheduled close events have future-timestamped Netty timers. Until those timers fire, `hookWorkFinishingUpdates` never runs, so `lastCompletedSourceTime` can't advance, so `stopReadingAt` stays frozen, so backpressure permanently blocks the consumer from polling more records. The replayer is completely stuck.

```mermaid
sequenceDiagram
    participant Kafka as Kafka Consumer
    participant BTS as BlockingTrafficSource
    participant RE as ReplayEngine
    participant Netty as Netty Timer
    participant BP as Backpressure (stopReadingAt)

    Note over Kafka: Backlog contains only close events (no requests)

    Kafka->>BTS: poll() → batch of close events
    BTS->>RE: closeConnection(timestamp=3h ago)

    RE->>RE: atTime = timeShifter(3h ago) → 18min from now
    RE->>Netty: schedule close at T+18min
    RE->>RE: tasksOutstanding++

    Note over BTS: All closes scheduled in future
    Note over BTS: No hookWorkFinishingUpdates fires
    Note over BP: lastCompletedSourceTime frozen
    Note over BP: stopReadingAt = lastCompleted + buffer (frozen)

    BTS->>BTS: blockIfNeeded() → lastTimestamp > stopReadingAt
    Note over BTS: ❌ DEADLOCK: consumer blocked, can't poll, timers haven't fired

    Note over RE: THIS PR: atTime = Instant.now()
    RE->>Netty: schedule close immediately
    Netty->>RE: hookWorkFinishingUpdates fires
    RE->>BP: stopReadingAt advances
    BP-->>BTS: ✅ Unblocked, resumes reading
```

**Change:** Close events now use `Instant.now()` instead of the time-shifted timestamp. There's no replay fidelity concern — a close is pure cleanup of a connection that already ended on the source side. The `OnlineRadixSorter` still enforces per-connection ordering so a close can't race ahead of pending sends. The original source `timestamp` is still passed to `hookWorkFinishingUpdates` so watermarks advance correctly.

---

## Bug Fix: Consumer eviction during backpressure (getNextRequiredTouch)

**On main:** `TrackingKafkaConsumer.getNextRequiredTouch()` returns `Optional.empty()` when `kafkaRecordsLeftToCommitEventually == 0`, causing `blockIfNeeded()` to call `readGate.acquire()` without any timeout. The consumer thread blocks indefinitely, never calls `poll()`, and after `max.poll.interval.ms` (10 min) Kafka kicks it — triggering a cascading rebalance.

**Change:** Only return empty when BOTH `partitionToOffsetLifecycleTrackerMap.isEmpty()` AND `kafkaRecordsLeftToCommitEventually == 0`. If the consumer has assigned partitions (which it does during backpressure), it will always schedule a keepalive touch.

---

## New: Wall-clock expiry watchdog

**On main:** `ExpiringTrafficStreamMap` only expires connections inside `expireOldEntries()`, called from `addObservationToAccumulation()` — i.e., only when new data arrives.

**Problem this solves:** When stale connections hold all `TrafficStreamLimiter` permits, the main thread blocks in `blockIfNeeded()`. No new records are read → event-driven expiry never fires → permits never freed → permanent deadlock requiring manual restart.

**Change:** A `ScheduledExecutorService` (30s interval) calls `expireByWallClock()`, sweeping accumulations whose source timestamp exceeds the configured timeout. This unblocks the main thread by releasing permits via the commit callback.

```mermaid
sequenceDiagram
    participant Main as Main Thread
    participant BTS as BlockingTrafficSource
    participant Watchdog as Expiry Watchdog (30s)
    participant Map as ExpiringTrafficStreamMap
    participant Gate as readGate

    Main->>BTS: blockIfNeeded()
    Note over Main: Blocked — stale connections hold all permits
    Note over Main: On main: deadlock forever

    loop Every 30 seconds
        Watchdog->>Map: expireByWallClock()
        Map->>Map: Find connections with source timestamp > timeout
        Map->>Gate: commitTrafficStream() → readGate.release()
    end

    Gate-->>BTS: Permit available
    BTS-->>Main: Unblocked, resumes reading
```

---

## New: Stale-head offset reaper

**On main:** `OffsetLifecycleTracker` has no mechanism to evict orphaned offsets from the head of its PriorityQueue. If `removeAndReturnNewHead()` is never called for the head offset (e.g., response callback dropped due to network timeout), the commit pointer is stuck forever.

**Change:**
- `OffsetLifecycleTracker.reapStaleHead(Duration)` — bulk-removes all consecutive stale offsets (>3 min) from the head and returns the new commit offset
- Threshold deliberately set below Kafka's default `max.poll.interval.ms` (5 min) so the reaper can unblock the consumer before Kafka evicts it — gives a 2-minute safety margin
- Called from both `getNextBatchOfRecords()` and `touch()` (the latter fires during backpressure, ensuring the reaper runs even when the read loop is blocked)
- `removeAndReturnNewHead()` now tolerates already-reaped offsets (returns empty instead of throwing) to handle the race where a normal completion arrives after the reaper already evicted that offset
- `onPartitionsAssigned()` passes the consumer's clock to the tracker for consistent age calculations

```mermaid
flowchart LR
    subgraph PriorityQueue [OffsetLifecycleTracker PriorityQueue]
        direction LR
        O100[offset 100\nage=6min ☠️] --> O101[offset 101\nage=6min ☠️] --> O102[offset 102\nage=6min ☠️] --> O103[offset 103\nage=2min ✓]
    end

    subgraph Reaper [reapStaleHead - 5min threshold]
        direction TB
        R1[Remove 100, 101, 102 in one sweep]
        R2[Return new commit offset = 103]
    end

    PriorityQueue --> Reaper
```

---

## Testing

| Test Class | Coverage |
|---|---|
| `ClosedPrematurelyOffsetBlockingTest` | CLOSED_PREMATURELY fires commit; PriorityQueue head-of-line blocking demo |
| `CommitTrafficStreamsStatusTest` | Production commit logic: CLOSED_PREMATURELY commits, INTERRUPTED suppresses |
| `ExpiringTrafficStreamMapWallClockExpiryTest` | Watchdog sweeps stale connections, skips fresh, handles edge cases |
| `OffsetLifecycleTrackerTest` | Bulk reap, stops at fresh offset, tolerates double-remove, race safety |
| `TrackingKafkaConsumerTest` | Consumer-level reapStaleHeads stages commits; logHeartbeat safety |

---

## Check List


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -164,6 +169,28 @@ public KafkaTrafficCaptureSource(
         kafkaConsumer.subscribe(Collections.singleton(topic), trackingKafkaConsumer);
```
Aman-bv, line 169: ## Design Note: keepAlive tradeoffs

**Problem solved:** The main loop blocks 26+ minutes in `forEach(accept)` processing large batches. During this time `kafkaExecutor` is idle and nobody calls `poll()`, so `max.poll.interval.ms` (10 min) expires, the consumer is evicted, triggering a cascading rebalance that kills all 16 consumers simultaneously.

**How this fixes it:** A background scheduler submits `keepAlive()` to the idle `kafkaExecutor` every 30s. Since all KafkaConsumer operations are serialized on this single-threaded executor, there's no concurrent access — `keepAlive()` only runs when `kafkaExecutor` has no other work queued.

**Pros:**
- Prevents cascading rebalance death (the primary goal)
- Minimal code (~30 lines), preserves all existing invariants
- `reapStaleHeads()` + `safeCommit()` during long batches means offsets advance sooner, reducing reprocessing on crash
- Thread-safe by construction (single-threaded executor serialization)
- Shutdown-safe (RejectedExecutionException caught, isClosed gate, cancel-before-shutdown ordering)

**Cons / risk to be aware of:**
- If the consumer is *processing-dead* but JVM-alive (deadlock, infinite loop, bug in the accumulator), keepAlive artificially keeps the partition assigned to a consumer that will never make progress. Previously, `max.poll.interval.ms` would evict it after 10 min and rebalance the partition to a healthy consumer.
- **Mitigation:** ECS health checks will kill truly stuck tasks. `session.timeout.ms` (the *network-level* heartbeat, separate from `max.poll.interval.ms`) still detects JVM-frozen scenarios. Monitoring on `schedulingLag` growth + `tasksOutstanding=0` can alert on the processing-stuck case.
- **Why this tradeoff is correct:** The old "auto-recovery" (eviction after 10 min) was worse than the disease — it cascaded and killed ALL consumers, not just the stuck one. One partition held hostage (detectable, recoverable via ECS) is strictly better than total cluster death.

gregschohn:
Thanks for the PR.  I've used this PR to work through all of the kafka handling parts of the replayer.  I've found a couple of issues that may have been the root cause of your head-of-line blocker.

Please take a look at https://github.com/opensearch-project/opensearch-migrations/pull/3225, which has now been merged to main.  I'm in the process of releasing 3.3.5 now.  Please let me know if you still see issues after you start using that.

We can still consume too much memory and get an OOM or hit a very long stall because we're limiting requests and they're taking a long time to work through - but all paths that I've checked SHOULD be constantly making forward progress as time progresses.  None of the waits or holds will be indefinite.  If we slip and don't make a callback, that's a bug.  I don't believe that the design will allow for a live-lock.  With regard to the memory consumption, scaling horizontally will help mitigate that - and THANK YOU FOR YOUR PRs!  As such, I don't want to adjust the replayer semantics that guarantee at least-once delivery (attempts).  If we start to weaken that, the whole platform's reliability comes into question and its much harder to reason about and the utility will be lowered catastrophically for many users.

I'm beginning to work through an approach is to buffer fewer Kafka records and set up a second scanning reader to expire old messages that never come back into scope.  I'll tag that design/PR to here too once I have it.  That will probably be next week.

Let me know if the latest mainline changes help you out / if you're still seeing issues / still think that we still need something else for commit control.

Aman-bv:
> Please take a look at #3225, which has now been merged to main. I'm in the process of releasing 3.3.5 now. Please let me know if you still see issues after you start using that.

I am actually building the images pushing to ECR and using it from there for both the capture-proxy and the traffic replayer with rest of the infra on OMA 3.2.1

> We can still consume too much memory and get an OOM or hit a very long stall because we're limiting requests and they're taking a long time to work through - but all paths that I've checked SHOULD be constantly making forward progress as time progresses. None of the waits or holds will be indefinite. If we slip and don't make a callback, that's a bug. I don't believe that the design will allow for a live-lock. With regard to the memory consumption, scaling horizontally will help mitigate that

I don't see a lot of memory usage on the ECS task. I think the bottleneck here is still the replayer. Also checked the logs, did not have any OOM or memory related logs.
 
> I'm beginning to work through an approach is to buffer fewer Kafka records and set up a second scanning reader to expire old messages that never come back into scope. I'll tag that design/PR to here too once I have it. That will probably be next week.

Thank you! : )

> Let me know if the latest mainline changes help you out / if you're still seeing issues / still think that we still need something else for commit control.

Hey @gregschohn ,
Thank you so much for the fixes and such a quick turnaround on all the issues. Much much appreciated.

About this PR: I kind of made this my playground to try different things and figure out the bugs, hence I moved this to draft. Will probably close this out after we complete the migration.

I tried out the fixes from the main and came across a few more edge cases which might have been missed if I am not wrong. 


# My current setup
capture proxy is creating kafka fragments of 29 MB, since replayer was not able to form the complete request and I was facing too many errors so decided to bump the fragment size. Although I don't think our batch indexing requests should be touching 29 MB. I have suppressed the GET requests from being captured. We have external versioning in place for the docs.

--speedup-factor 25  --max-concurrent-requests 1024

Kafka Consumer Properties (/app/kafka.properties)
  max.poll.interval.ms=600000      (10 minutes) --> to give enough time in order to avoid rebalance
  max.poll.records=500
  max.partition.fetch.bytes=31457280  (30 MB) 
  fetch.max.bytes=31457280            (30 MB) 

Still the offset is getting stuck and the consumer is getting kicked out, triggering a rebalance which never completes.

Screenshot 2026-07-17 at 10 23 42 PM

Screenshot 2026-07-17 at 10 25 41 PM


Overtime it grows a lot:

Screenshot 2026-07-17 at 10 25 57 PM


I also had a deadlock where the consumer rebalancing never finished, I had to stop the replayer and start again (yet I still get stuck at offset blocking again and again).

Deadlock sequence as given by Claude:

  1. 16 consumers join simultaneously → cooperative rebalance starts
  2. onPartitionsRevoked fires inline during poll() → calls enqueueTrafficSourceReaderInterruptedClosesForPartitions → increments outstandingTrafficSourceReaderInterruptedCloseSessions
  3. Next readNextTrafficStreamSynchronously call drains the close queue (line 332) and returns synthetic closes
  4. But on the NEXT iteration, outstandingTrafficSourceReaderInterruptedCloseSessions > 0 (line 339) because the accumulator hasn't yet called onNetworkConnectionClosed
  5. The code parks for 5ms and returns empty — never reaching getNextBatchOfRecords (line 349)
  6. poll() is NEVER called because it only happens inside getNextBatchOfRecords → safePollWithSwallowedRuntimeExceptions
  7. The cooperative rebalance protocol requires poll() to send the SyncGroup response
  8. Without poll(), no consumer can finalize their assignment
  9. All 16 consumers appear as - in describe-consumer-group — rebalance hangs forever

Created a PR that hopefully addresses this: https://github.com/opensearch-project/opensearch-migrations/pull/3231

Aman-bv:
# Another bug is the expire connections not working

The logs showed:
  - BackpressureHeartbeat - engaged=true — reads are blocked
  - oldestInWrites={conn=..., lastPacketAge=56m51s} with expiryConfig=6m0s
  
The connection has been idle for 57 minutes but expiry is 6 minutes. It's not expiring because backpressure blocks reads → no new observations → no expiry sweep runs → stuck connections live forever → their uncommitted offsets block progress.

CC: @gregschohn

Aman-bv:
Also, the target opensearch cluster is not the problem, it is being heavily underutilised given the cluster size.

Aman-bv:
I am trying to debug the problem, given that:
 
  1. Target cluster is at 0% CPU : it's barely receiving any work
  2. targetResponses={} 
  3. accumulatingRequest age=20m — requests stuck being assembled at the replay layer
  4. scheduled age=20m 

I think the stuck connections in accumulatingRequest are ones holding the oldest offsets and blocking all commits. Not sure why.

Edit:
Found this in the logs:
```
[WARN ] 2026-07-17 18:30:45,547142 [main] TrafficReplayerCore - Batch processing took 25286ms (1 records). This delays the next Kafka poll. max.poll.interval.ms may be at risk.
     
[WARN ] 2026-07-17 18:37:35,261961 [main] TrafficReplayerCore - Batch processing took 52273ms (1 records). This delays the next Kafka poll. max.poll.interval.ms may be at risk.
```

Aman-bv:
Hey @gregschohn, wanted to share a detailed breakdown of the problems we faced with stuck offsets during production deployment and how each fix addresses them (This is using the fix from master and PR #3231 along with the fixes below):

## Root Cause Analysis: CURRENT-OFFSET Stuck / Consumer Lag Growing Indefinitely

### Symptoms
- `CURRENT-OFFSET` freezes on multiple partitions while `LOG-END-OFFSET` keeps growing
- `targetResponses={}` (zero responses) despite the target cluster being responsive
- Connections stuck in `ACCUMULATING_WRITES` state for hours beyond their expiry timeout
- Eventually all semaphore permits exhausted with 100K+ items queued

### Root Cause

**Wall-clock expiry comparing wall-clock time against source-time timestamps**

The `ExpiringTrafficStreamMap` expiry is source-time-driven — it only fires when new packets arrive and uses source timestamps for comparison. When backpressure is engaged and no new records arrive for a specific connection, expiry never fires. The wall-clock fallback in the heartbeat thread was comparing `System.currentTimeMillis()` against the source-time `lastPacketTimestamp`, which could be hours in the past due to speedup-factor — so the comparison was meaningless and expiry never triggered.

Additionally, when expiry does fire, `closeConnection` tries to drain the schedule. But the schedule's head item is a Netty timer future that won't complete for up to `targetResponseTimeout` seconds. The `closeConnection` path blocks waiting for it, which blocks the heartbeat thread, which prevents further expiry checks.

### Fix: Wall-Clock Expiry + Expedite Connection

1. **Track real wall-clock activity** — Added `lastWallClockActivityMs` (real monotonic time) on each `Accumulation`, updated on every packet. The heartbeat sweep now compares this against real elapsed time.

2. **Expedite instead of close** — Introduced `expediteConnection` which marks the session as expedited, completes all pending transformation timers and schedule futures immediately (fires them rather than cancelling). This bypasses time-shift delays while still sending all requests to the target, preserving **at-least-once delivery**.

### Side Effect: Log4j2 Thread Starvation Under Burst Load

When expedited connections fire all queued requests simultaneously (no time-shift delays), this creates a burst of logging activity. The replayer uses `AsyncLoggerContextSelector` (LMAX Disruptor ring buffer) with `WaitStrategy=Block` and `RollingRandomAccessFile` appenders writing to EFS. Under burst load:

1. EFS write latency spikes due to the sudden volume
2. File appender threads stall, ring buffer fills
3. `WaitStrategy=Block` causes every application thread that logs to block — including Netty event loops and the heartbeat thread
4. The entire JVM freezes while ECS reports HEALTHY

**Mitigation:** Removed EFS-backed file appenders. All logging now goes to STDERR only. Additionally set `AsyncQueueFullPolicy=Discard` with `DiscardThreshold=INFO` and changed `WaitStrategy` from `Block` to `Timeout` as a safety net against future ring buffer pressure.

### Changes Summary

| File | Change |
|------|--------|
| `Accumulation.java` | Added `volatile long lastWallClockActivityMs` field |
| `CapturedTrafficToHttpTransactionAccumulator.java` | Wall-clock expiry using real timestamps + double-fire guard |
| `ClientConnectionPool.java` | Added `expediteConnection()` method |
| `ConnectionReplaySession.java` | Added `expedited` flag + `expediteTransformationTimers()` |
| `TimeToResponseFulfillmentFutureMap.java` | Added `expediteImmediately()` |
| `RequestSenderOrchestrator.java` | Skip time-shift delays when expedited; retry immediately |
| `TrafficReplayerCore.java` | Trigger `expediteConnection` on `EXPIRED_PREMATURELY` / `CLOSED_PREMATURELY` |
| `ReplayEngine.java` | Delegate `expediteConnection` |
| `log4j2.properties` | Removed EFS file appenders, STDERR only |
| `log4j2.component.properties` | `WaitStrategy=Timeout`, `AsyncQueueFullPolicy=Discard` |

### Key Design Decision: At-Least-Once Preserved

`expediteConnection` does NOT discard requests. It completes timer futures immediately so queued requests execute without time-shift delays. Every request is still sent to the target and the offset is only committed after the response is received. This is fundamentally different from `cancelConnection` (used for partition rebalance) which discards work.

PR #3209, opened 2026-07-13
Harden migration resource contracts: dependsOn, DataSnapshot recovery, solrCollections, doc cleanup

gregschohn:
## Purpose

PR to run CI and verify the migration-resource contract cleanup. This is follow-on work to the CRD/VAP/checksum hardening that landed on the Solr backup PR branch. The plan is documented in `docs/migrationResourceContractCleanupPlan.md`.

**Note:** This targets `main` to run CI. The 7 cleanup commits are intended to land on Andre's Solr backup branch once verified.

## Changes

1. **`dependsOn` written by `tryApply`, stripped from the initializer** — `workflow reset` builds its deletion DAG from `spec.dependsOn` on live CRs, but the DataSnapshot/SnapshotMigration apply manifests never wrote it. After the first reconcile those terminal CRs looked dependency-free and could be deleted out of safe order. Now written by `tryApply` (sole writer; reflects only established edges) and omitted from the initializer bootstrap.

2. **DataSnapshot VAP-retry recovery loop** — every other reconcile parks a VAP-rejected apply on an `ApprovalGate` and retries; DataSnapshot had no recovery path (silent dead end). Added `waitForFix → patchApproval → resetGate → retryLoop` + initializer gate, matching the other resources.

3. **`solrCollections` moved off the user schema** — Solr-only internal field that leaked onto the ES/OS-facing `USER_CREATE_SNAPSHOT_OPTIONS` where it was silently ignored. Moved to internal `ARGO_CREATE_SNAPSHOT_OPTIONS`; still projected into `DataSnapshot.spec`.

4. **Resolved-param ↔ applied-CR alignment** — default-less string fields (`otelTraceCollectorEndpoint`, transformer configs) that the apply manifests write as `""` are now emitted by the resolved params too, so MigrationRun history/dry-run match the live CR.

5. **Create-path snapshot name lowercased in the reported output** — matches the name actually created.

6. **Docs** — `.describe()` on internal `mode`/`snapshotSourceType` enums; updated `reconfiguringWorkflows.md`, `resolvingMigrationParametersFromConfigs.md`, `WorkflowCrdDesign.md`, `vapCrdChecksumHardeningPlan.md`.

`authMtlsCaCertHash` kept `impossible` intentionally (auth-trust change = delete/recreate; no safe gated path).

## Verification

- `npm run build` (full workspace) — pass
- `npm -ws run test` (372 tests) — pass
- `npm run test:integ:crds` — pass


CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -42,25 +48,74 @@ public class SolrBackupStrategy implements SourceBackupStrategy {
...
+    private final boolean isCloud;
 
     public SolrBackupStrategy(CreateSnapshot.Args args) {
         this.args = args;
+        this.repoUri = RepoUri.parse(args.repoUri);
+        validateSolrRepoSupported(repoUri);
         this.connectionContext = args.sourceArgs.toConnectionContext();
         this.httpClient = new SolrHttpClient(connectionContext);
+        this.mode = CreateSnapshot.getSnapshotMode(args);
+        this.isCloud = isSolrCloud(connectionContext.getUri().toString(), httpClient);
```
jzonthemtn, line 62: If this returns `200` it is SolrCloud. Anything else (timeout, error)  is Solr standalone. So even if the endpoint is wrong, it will be assumed to be Solr standalone and keep going. Probably not worth fixing here, but in a separate PR to change that logic a bit. Maybe increase the timeout to avoid network hiccups, and check the exceptions returned.

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -69,28 +124,382 @@ public void run() {
...
+
+        try (var s3Client = buildS3Client(args.s3Region, args.endpoint)) {
+            var response = s3Client.listObjectsV2(
+                software.amazon.awssdk.services.s3.model.ListObjectsV2Request.builder()
+                    .bucket(repoUri.bucketName)
+                    .prefix(snapshotPrefix)
+                    .maxKeys(1)
+                    .build());
+            if (response.contents().isEmpty()) {
+                log.warn("No objects found at s3://{}/{} — downstream pipeline may fail if snapshot data "
```
jzonthemtn, line 456: This line and 462 on down -- should it be loud and error that there are no files in S3? What is the reasoning to keep going even if no objects were found?

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
```diff
@@ -124,13 +124,27 @@ public static class Args {
                 description = "Source cluster type: 'elasticsearch' (default) or 'solr'")
         public String sourceType = "elasticsearch";
 
+        @Parameter(
```
jzonthemtn, line 127: I think `--mode` only applies to Solr backups. I suppose if provided for a non-Solr source it will just be ignored? Should it fail with an error and notify the user in that case? Can also be a separate issue/PR.

I don't think it's worth renaming to `--solr-mode` but I'll defer that judgment call.

jzonthemtn:
I see this PR and #3187 as being independent, and if this is the chosen path forward I would merge this PR first since it is focused on the create/import side while #3187 is focused on the backup-reading side. Then, update #3187 as needed to merge.

jzonthemtn (approved):
I will write low priority follow-up issues for the couple of things I noted during the review.

CreateSnapshot/src/main/java/org/opensearch/migrations/CreateSnapshot.java
gregschohn: Thanks - good point and I think that you're right, at least for now.  CreateSnapshot doesn't do a lot - but I could potentially see more value in maybe doing ES/OS snapshot validation with the import command or possibly doing some extraction step in the future.  It's only used for solr right now, but the idea of importing vs purely creating from scratch might be relevant for some other modes too.
I'm not holding onto this belief tightly though.

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
jzonthemtn: Captured as https://github.com/opensearch-project/opensearch-migrations/issues/3233

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
jzonthemtn: Captured as https://github.com/opensearch-project/opensearch-migrations/issues/3232

PR #3216, opened 2026-07-13
Handle frontside TLS handshake failures

gregschohn:
## What changed

Adds a frontside TLS exception handler immediately after Netty SslHandler in the capture proxy pipeline. TLS handshake failures are now logged as warnings and the channel is closed locally instead of letting the exception reach the tail of the pipeline.

## Why

Malformed or failed client TLS handshakes can be triggered before the proxy has authenticated or decoded HTTP. Previously these exceptions could propagate as unhandled Netty pipeline exceptions, matching the observed crash-risk report.

## Impact

The proxy should report frontside TLS handshake failures in logs and keep serving subsequent valid TLS traffic.

## Validation

- ./gradlew :TrafficCapture:trafficCaptureProxyServer:test --tests org.opensearch.migrations.trafficcapture.proxyserver.netty.NettyScanningHttpProxyTest
- Added coverage for a plaintext request sent to a TLS proxy port, asserting the TLS failure is logged and a subsequent HTTPS request still succeeds.


gregschohn:
Will reopen with DDOS protections against over-logging shortly (I'd meant to do that, but didn't push)

PR #3219, opened 2026-07-14
Harden proxy tls handshake logging

gregschohn:
### Description

Adds a frontside TLS exception handler immediately after Netty SslHandler in the capture proxy pipeline. TLS handshake failures are now logged as warnings and the channel is closed locally instead of letting the exception reach the tail of the pipeline.

Malformed or failed client TLS handshakes can be triggered before the proxy has authenticated or decoded HTTP. Previously these exceptions could propagate as unhandled Netty pipeline exceptions, matching the observed crash-risk report.

Also put some protections in place to suppress logging after repeated errors to diminish the impact of using this as a DOS attack.

### Testing
- ./gradlew :TrafficCapture:trafficCaptureProxyServer:test --tests org.opensearch.migrations.trafficcapture.proxyserver.netty.NettyScanningHttpProxyTest
- Added coverage for a plaintext request sent to a TLS proxy port, asserting the TLS failure is logged and a subsequent HTTPS request still succeeds.

### Check List

### DIFF
--- a/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/UnauthenticatedClientLogDeduper.java
+++ b/TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/UnauthenticatedClientLogDeduper.java
@@ -1,0 +1,89 @@
+package org.opensearch.migrations.trafficcapture.proxyserver.netty;
+
+import java.time.Duration;
+import java.util.EnumMap;
+import java.util.Objects;
+import java.util.function.LongSupplier;
+
+class UnauthenticatedClientLogDeduper {
+
+    enum KnownEvent {
+        FRONTSIDE_TLS_HANDSHAKE_FAILURE
+    }
+
+    private final EnumMap<KnownEvent, EventState> eventStates = new EnumMap<>(KnownEvent.class);
+    private final long dedupeWindowNanos;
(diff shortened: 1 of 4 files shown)


TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/UnauthenticatedClientLogDeduper.java
```diff
@@ -0,0 +1,89 @@
...
+            return LogDecision.suppress();
+        }
+
+        var suppressedCount = eventState.suppressedCount;
+        eventState.suppressedCount = 0;
+        eventState.nextLogAfterNanos = now + dedupeWindowNanos;
+        return LogDecision.log(suppressedCount);
+    }
+
+    int knownEventCount() {
```
jzonthemtn: Minor - I think `knownEventCount()` is only used by the tests.

jzonthemtn (approved):

TrafficCapture/trafficCaptureProxyServer/src/main/java/org/opensearch/migrations/trafficcapture/proxyserver/netty/UnauthenticatedClientLogDeduper.java
gregschohn: Thanks - that doesn't look very useful, removing.

AndreKurait (approved):

PR #3220, opened 2026-07-14
feat(gcp): make live-migration capture proxy private on GKE

tonypiazza:
### Description

Makes the live-migration **capture proxy** deploy behind an internal (VPC-local) load balancer on non-AWS clouds. Previously the proxy's Kubernetes `Service` emitted only AWS-specific load-balancer annotations, so on GKE it received a **public IP** — leaving the live capture-and-replay ingress exposed on the public internet.

The load-balancer scheme is now selected by cloud provider:

- **`setupCapture.ts`** — the proxy `Service` annotations are built from a `cloudProvider` value:
  - `aws` (default): NLB, internal/internet-facing per `internetFacing` (**unchanged from today**)
  - `gcp`: `networking.gke.io/load-balancer-type: Internal`
  - `azure`: `service.beta.kubernetes.io/azure-load-balancer-internal: true`

  `cloudProvider` is resolved as a **template input** via `configMapKeyRef` against the `provider-config` ConfigMap. Global workflow arguments do not resolve `valueFrom` at submit time, so the value must enter at the input layer (this mirrors how image params already resolve).
- **`providerConfig.yaml`** — renders the `provider-config` ConfigMap unconditionally from `.Values.cloudProvider` (key always present; Argo v4 errors on a missing `configMapKeyRef` key even with `optional: true`). Defaults to `aws` when unset.
- **`terraform/gcp`** — fully qualifies the capture-proxy and traffic-replayer image repositories to the Artifact Registry path (as already done for the console and reindex-from-snapshot images). Their chart defaults are unqualified (`migrations/...`) and resolve against docker.io, which fails to pull on any non-local deploy.
- **Docs** — note that live capture-proxy ingress is now private on GKE, and that image repositories must be fully qualified at deploy time.

**AWS/EKS behavior is unchanged when `cloudProvider` is unset** (defaults to `aws`). This is a no-op for existing AWS deployments — the only added resource is one `provider-config` ConfigMap.

### Issues Resolved

N/A — follow-up to #3205 (GCP private networking for snapshot/RFS migrations), extending private-by-default connectivity to the live-migration capture path.

### Testing

- **Unit:** Adds `proxyServiceLoadBalancer.test.ts` locking the per-cloud annotation branches; regenerates the `setupCapture` workflow-template snapshot. Full `orchestrationSpecs` suite green (`npm run test`, `npm run type-check`).
- **CRD integration:** `npm run test:integ:crds` (k3s testcontainer) passes.
- **AWS no-regression, verified at three layers on a clean Ubuntu 22.04 + Docker environment matching CI:**
  1. *Generated template diff* — of 14 workflow templates only `setup-capture.yaml` changes; the AWS annotation branch reproduces the three original `aws-load-balancer-*` annotations byte-for-byte.
  2. *Helm render diff* — rendering the chart with `valuesEks.yaml` (cloudProvider unset), the entire chart diff is a single **additive** `provider-config` ConfigMap (`cloudProvider: "aws"`); zero modifications or deletions to existing resources.
  3. *Runtime on kind with argo-workflows 1.0.5 (Argo v4.0.3)* — submitting the `setup-capture` `deployproxyservice` template, `cloudProvider` resolves from the ConfigMap at submit time and the created `Service` carries exactly the three original AWS annotations; flipping the ConfigMap to `gcp` yields `networking.gke.io/load-balancer-type: Internal`.
- **GCP end-to-end:** the private live-migration path (backfill + capture + replay) has been validated on real GKE over private networking with exact per-index document-count parity.

### Check List


tonypiazza:
**CI note — `pr-eks-cdc-full-e2e-test` #400 failure looks like a teardown race, not a migration failure**

Dug into [build 400](https://migrations.ci.opensearch.org/job/pr-checks/job/pr-eks-cdc-full-e2e-test/400/). Summary: `1 passed, 5 errors`. The migration itself worked — test 0031 captured 10 docs through the capture proxy and all 10 landed on the target (source and target `_count` both = 10). The single real failure is in **teardown**, which then cascaded to error the other 4 cases at setup (dirty namespace tripped the `_fail_if_migration_resources_exist()` pre-check).

The failing step is `workflow reset`:

Error: Timed out waiting for services deletion: capture-proxy
Error: refusing to delete captureproxy.capture-proxy until owned services resources are gone.

Root cause looks independent of this PR. In `reset.py`, `_wait_for_owned_deletion` waits `OWNED_RESOURCE_DELETION_TIMEOUT_SECONDS = 120` for the proxy's owned `Service` to disappear, and its finalizer-stripping fallback is gated on `if group:` — so it only strips finalizers on CustomObjects, never on a core-v1 `Service`. On EKS the `LoadBalancer` Service carries the AWS LB controller's finalizer and isn't removed until the NLB is fully deprovisioned; if that exceeds 120s, the wait fails and the parent CR won't delete. This PR doesn't touch   `reset.py`.

On the PR side, the capture-proxy Service annotation change is byte-equivalent on the AWS branch (AWS NLB annotations are the default fallthrough when `cloudProvider` isn't `gcp`/`azure`), and the green data path confirms provisioning is unaffected.

Requests:

1. Could someone re-run this job for PR #3220 against the current commit? Build 400 was auto-triggered by PR activity, and the `build` endpoint is locked down, so I can't retrigger it myself. If the job only re-fires on a new push rather than a manual run, let me know and I'll push a no-op commit instead. Same params as #400: test IDs `0031,0032,0033,0040,0042`, `ES_7.10 → OS_1.3`. If it comes back green, this is a slow-NLB-teardown flake rather than a code defect.
2. Separately, worth considering hardening `_wait_for_owned_deletion` to also strip finalizers on core-v1 Services (or a longer timeout for `LoadBalancer` Services) — this failure mode is latent regardless of this PR.

AndreKurait (approved):

PR #2986, opened 2026-05-15
Write failed documents to a durable DLQ (S3)

jzonthemtn:
### Description
Writes failed documents to a durable DLQ (S3).

### Issues Resolved
Closes #2975 

### Testing
Added unit and integration tests.

### Check List
- [X] New functionality includes testing
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #3229, opened 2026-07-17
Write failed documents to a durable failed-document stream (S3) — with review follow-ups

### LINKED ISSUE #2975: Expose a durable failed-document DLQ for Reindex-from-Snapshot on EKS
## Description

Reindex-from-Snapshot (RFS) already detects document failures and has a `FailedRequestsLogger`, but the EKS/Argo deployment path currently disables that logger because there is no durable, customer-accessible way to collect the output. As a result, customers can see shard progress and worker health, but they cannot reliably answer which documents failed, why they failed, or where to find those failures for follow-up.

We should add a first-class failed-document DLQ for RFS on EKS that is durable, session-scoped, customer-accessible, and surfaced through workflow status/commands.

## Problem

- RFS can identify failed bulk requests, but the current EKS path drops that detail instead of persisting it durably.
- Current workflow/status views focus on shard progress and do not expose failed-document inventory.
- Results from one RFS run must be isolated from earlier runs of the same workflow/resource.
- The current exception model is not clear enough for DLQ semantics:
  - `allowedDocExceptionTypes` should still be treated as success and excluded from DLQ output.
  - retryable vs non-retryable failures need clearer handling so only terminal failures appear in the DLQ.
- Operators need a supported way to inspect failures from workflow UX, not by manually spelunking pod-local logs.

## Proposal

Implement a durable DLQ sink for unexpected RFS document failures on EKS, with a workflow-facing UX around it.

The initial implementation should:

- Persist terminal document failures to a durable sink rather than pod-local files.
- Scope DLQ data to an RFS session/configuration so a new run does not mix with prior runs.
- Surface DLQ summary information in workflow status.
- Add a workflow command to list DLQ entries in a stable ordering for the most recent session.
- Define a bounded, documented DLQ record schema with enough detail for investigation without making each entry excessively verbose.

A reasonable default would be one shared bucket/artifact location with per-session prefixes, while still allowing sink configuration if we want that flexibility. We should reuse existing reliable sink patterns where possible rather than inventing an entirely new best-effort logger path.

## Suggested Scope Boundaries

- In scope:
  - durable persistence of failed-document records
  - stable listing/querying for a session
  - workflow status and workflow command support
  - clear exception classification semantics for DLQ inclusion
  - operator documentation
- Not required for the first cut:
  - automatic re-drive from the DLQ back into RFS
  - a generic cross-component DLQ framework for every migration subsystem

If automated retry/redrive is not part of the first implementation, the docs and workflow UX should say so explicitly and position the DLQ as an inspection/export mechanism.

## Acceptance Criteria

- Unexpected terminal document failures during RFS backfill on EKS are written to a durable DLQ that customers can inspect during or after a migration.
- The DLQ is scoped to the latest RFS session/configuration and does not mix records from prior runs of the same workflow/resource.
- The implementation no longer relies on `FAILED_REQUESTS_LOGGER_LEVEL=OFF` with no durable replacement.
- `allowedDocExceptionTypes` are excluded from DLQ records and failure counts.
- Retryable vs non-retryable exception behavior is defined and implemented for DLQ purposes:
  - retryable failures are not emitted as final DLQ entries unless they still fail after retry exhaustion
  - non-retryable terminal failures are emitted once
  - allowed exceptions are treated as success and emitted zero times
- Each DLQ record uses a documented, stable schema that includes at least:
  - migration/session identifier
  - target index name
  - document identifier when available
  - failure classification/type
  - a concise failure reason or error summary
  - timestamp
  - enough request context or payload to support investigation
- The DLQ record format is intentionally bounded so it does not become an unstructured dump of every possible log detail.
- Customers can deterministically find the DLQ location for a given backfill session from workflow/session metadata.
- A workflow command lists DLQ entries for the latest session in a stable ordering.
- Workflow status shows:
  - failed document count for the latest session
  - a pointer/reference to the DLQ location
- The failed document count reported by workflow status matches the persisted DLQ contents for the latest session.
- At least one automated integration test forces document-level failures and verifies:
  - successful documents still migrate
  - failed documents are persisted to the DLQ
  - the reported failure count matches the DLQ contents
  - records from a prior run are not incorrectly reported for a later run
- Documentation explains:
  - where the DLQ lives by default
  - how sink location/prefixing is configured
  - how to query/list DLQ entries for a session
  - how to inspect/export failures
  - whether retry/redrive is supported, and if not, what the supported operator flow is instead

## Notes

- One shared bucket is probably operationally simpler than a dedicated bucket per sink, as long as prefixes clearly isolate sessions/runs.
- Existing reliable sink implementations in the repo are worth reusing as design references.
- The first implementation should optimize for durable, queryable observability rather than for full automated retry orchestration.

If you want, I can also turn this into a shorter, more opinionated GitHub issue that reads more like something ready to paste directly into `opensearch-migrations/issues`.

gregschohn:
### Description

Writes failed documents to a durable failed-document stream (S3), building on #2986 by @jzonthemtn / @o19s. This branch includes all of the original PR's commits, plus follow-up fixes addressing review-comment items and two additional correctness/robustness issues found while reviewing.

**Follow-up commits added on top of #2986's head:**

1. **Fix failed-document-stream retry-exhaust failure/document correlation.** `emitRetryExhaustedToFailedDocumentStream` keyed the last bulk response's per-item failures by response *position* and then indexed `pendingOps` (already compacted by `compactPendingDocs`) by *position*. Those two position spaces diverge whenever a success or non-retryable item precedes a surviving retryable item, so the retryable doc got the wrong `failureType`/`responseItem` attached (or null). Now correlates by **document id**, which is stable across compaction. Includes a regression test that fails on the old positional code and passes with the fix.

2. **Make local append failures durable-gating in `RotatingGzipS3ObjectWriter`.** A local gzip append failure (e.g. disk full) previously returned a failed future that the sink only logged, so the gating flush still succeeded and the work item was marked complete even though a terminal-failure record was lost. Local append IO failures are now retained so the next `flush()` surfaces them and the caller refuses to complete the work item (at-least-once: a successor reprocesses). Per-record serialization "poison pills" are intentionally still dropped so they can't loop forever.

3. **Run failed-document-stream writes off the netty event loop.** Serialization + gzip + blocking temp-file I/O were running inline on the reactor-netty event-loop thread that delivered the bulk response, which could stall other in-flight requests during failure spikes. Writer access is now marshalled onto a single dedicated worker thread, mirroring the replayer's `S3TupleSink`. The append is enqueued synchronously before the returned Mono is subscribed, so a batch's records are still appended before that batch's gating flush runs. Public API and the fail-fast gating contract are unchanged.

4. **Restore the full Solr version matrix for `k8s-local-solr-other-test`** (`SOLR_9.8` → `SOLR_6.6,SOLR_7.7,SOLR_9.8`), which #2986 had narrowed; `main` still runs all three.

### Issues Resolved
Closes #2975

### Testing
Original PR added unit and integration tests. The follow-ups add: a retry-exhaust position-shift regression test, two `RotatingGzipS3ObjectWriter` tests (retained-and-surfaced append failure vs. non-retained serialization failure), and update one sink test to the new async-write contract. Verified green locally: `s3Common`, `TrafficCapture:tupleSink`, the RFS failed-document-stream / client / parser / integration suites, and the `DocumentsFromSnapshotMigration` end-to-end flush-gating tests.

### Check List


jzonthemtn:
The PR description says correlated by document id but I think that is out of date. The code comment says "Correlation is by retryable-failure index". Nothing to update unless document id was the intended method.

jzonthemtn (approved):

PR #3231, opened 2026-07-17
fix(replayer): resolve rebalance deadlock and drain-gate bugs blocking Kafka consumers

Aman-bv:
## 🔍 Description

Fixes a production issue where Kafka consumer `CURRENT-OFFSET` gets permanently stuck, causing lag to grow indefinitely. The root cause is a combination of bugs that prevent offset commitment when exceptions occur during request processing, compounded by a backpressure gate that freezes when work is outstanding.

---

### 1. Core Fix — Offset commit on exception
**File:** `TrafficReplayerCore.java`

When `packageAndWriteResponse` threw a `CompletionException` (e.g. from a `CancellationException` during connection drain), `commitTrafficStreams` was never called. This left the offset permanently stuck in the `OffsetLifecycleTracker`, blocking all subsequent commits for that partition.

> **Fix:** Move `commitTrafficStreams` into a `finally` block in `processCompletedTransaction` so offsets are always committed (or released) regardless of downstream exceptions. The status-based guard (`TRAFFIC_SOURCE_READER_INTERRUPTED` / `CLOSED_PREMATURELY` → no commit) is preserved, **maintaining at-least-once delivery semantics**.

---

### 2. Backpressure gate advancement
**File:** `ReplayEngine.java`

The `isWorkOutstanding()` guard in `updateContentTimeControllerWhenIdling()` prevented the content-time controller from advancing when work was outstanding (the *normal steady state* under load). This froze the backpressure gate (`stopReadsPast` never advanced), which blocked all reads from `BlockingTrafficSource`, which meant zero `addObservation()` calls, which meant zero expiry sweeps — leaving connections stuck indefinitely.

> **Fix:** Remove the `isWorkOutstanding()` guard so the gate always advances unconditionally.

---

### 3. TrackedFuture duplicate parent crash
**File:** `TrackedFuture.java`

During connection drain, `setParentDiagnosticFuture` could be called on a future that already had a parent (from a retry path). The `IllegalStateException` crashed the completion handler, leaving offsets permanently stuck.

> **Fix:** Convert from throwing ISE to logging a warning and retaining the original parent.

---

### 4. Connection drain — head-of-line blocking
**File:** `ClientConnectionPool.java`

`schedule.clear()` silently discarded pending work items without completing their futures. Those futures are tracked by the `OffsetLifecycleTracker` — if they never complete, their offset never advances.

> **Fix:** Replace `clear()` with `drainWithCancellation()` + `cancelAllWork()` to properly complete futures with `CancellationException`.

---

### 5. Wall-clock expiry safety net (defense-in-depth)
**File:** `CapturedTrafficToHttpTransactionAccumulator.java`

The existing expiry (`ExpiringTrafficStreamMap.expireOldEntries()`) is packet-timestamp-driven — it is **only** called from `addObservation()`, which is only called when new records are delivered from Kafka. The causal chain that made this fail:

1. `updateContentTimeControllerWhenIdling()` was guarded by `isWorkOutstanding()` (see change 2 above)
2. With work outstanding (normal steady state), the backpressure gate froze entirely
3. `BlockingTrafficSource` blocked all reads because `stopReadsPast` never advanced
4. Zero reads → zero `addObservation()` calls → zero expiry sweeps
5. Connections sat idle for 57+ minutes despite a 6-minute configured timeout

Now that change 2 removes the `isWorkOutstanding()` guard, this specific scenario won't recur. However, if any future bug freezes the read pipeline, stuck connections would silently block offset advancement again.

> **Fix:** Add a wall-clock fallback in the periodic heartbeat that force-expires connections idle beyond 1.5x the configured timeout. This is defense-in-depth — it ensures stuck connections can never permanently block offset advancement regardless of the state of the read pipeline.

---

### 6. Rebalance deadlock
**File:** `KafkaTrafficCaptureSource.java`

Three bugs in the drain-gate logic caused consumers to permanently lose partition assignments:
1. **Session-number key mismatch** — `remove()` never matched registration key
2. **BlockingTrafficSource missing delegation** — callback never reached `KafkaTrafficCaptureSource`
3. **Drain-gate blocking `poll()`** — cooperative rebalance couldn't complete

> **Fix:** Correct key construction, add delegation overrides, replace `parkNanos` with `touch()` keepalive + 5-minute circuit-breaker timeout.

---

### 7. Observability enhancements

| Component | What it reports |
|-----------|----------------|
| `TrafficStreamLimiter.logHeartbeat()` | Semaphore permit utilization |
| `ReplayEngine.logHeartbeat()` | Scheduling lag, response code counters, stale-duration warning |
| `TrackingKafkaConsumer` | Worst stuck commit head across partitions |

---

## 📋 Issues Resolved

Resolves stuck `CURRENT-OFFSET` / indefinitely growing consumer lag in the traffic replayer when exceptions occur during request processing or when backpressure gates freeze.

---

## ✅ Testing

| Test | Covers |
|------|--------|
| `BackpressureGateAdvancesWithWorkOutstandingTest` | Gate advances even when `isWorkOutstanding()==true` |
| `TrackedFutureDuplicateParentTest` | Duplicate parent doesn't throw, retains original |
| `sessionNumberMismatch_counterStillDecrements` | Regression for key mismatch bug |
| `drainGate_callsTouchToKeepConsumerAlive` | Regression for `poll()` keepalive |
| `drainGate_timeoutResetsCounterAndUnblocksReadLoop` | Circuit-breaker timeout |
| `drainGate_touchRuntimeExceptionIsCaughtAndLoopContinues` | Exception safety |
| `blockingTrafficSource_delegatesLifecycleCallbacks` | Delegation overrides |

- All existing tests in `org.opensearch.migrations.replay.kafka.*` continue to pass
- Codecov confirms all modified and coverable lines are covered

---

## 📝 Check List


coreUtilities/src/main/java/org/opensearch/migrations/utils/TrackedFuture.java
```diff
@@ -99,12 +99,12 @@ protected void setParentDiagnosticFuture(TrackedFuture<D, ?> parent) {
...
         var wasSet = parentDiagnosticFutureRef.compareAndSet(null, parent);
         if (!wasSet) {
-            throw new IllegalStateException(
-                "dependencyDiagnosticFutureRef was already set to " + parentDiagnosticFutureRef.get()
-            );
+            log.atWarn().setMessage("dependencyDiagnosticFutureRef was already set — ignoring duplicate parent. " +
+                    "Existing={}")
+                .addArgument(parentDiagnosticFutureRef::get)
+                .log();
+            return;
```
Aman-bv, line 106: Why are we changing this?

coreUtilities/src/main/java/org/opensearch/migrations/utils/TrackedFuture.java
```diff
@@ -99,12 +99,12 @@ protected void setParentDiagnosticFuture(TrackedFuture<D, ?> parent) {
...
         }
-        // the parent is a pretty good breadcrumb for the current stack... but the grandparent of the most recently
```
Aman-bv, line 106: Do we really need to remove this comment?

TrafficCapture/dockerSolution/src/main/docker/kafka-consumer.properties
```diff
@@ -1,2 +1,4 @@
-fetch.max.bytes=8388608
-max.partition.fetch.bytes=8388608
+max.poll.interval.ms=600000
+max.poll.records=500
+max.partition.fetch.bytes=52428800
+fetch.max.bytes=52428800
```
Aman-bv: Leave default values as is.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
```diff
@@ -324,26 +329,59 @@ public CompletableFuture<List<ITrafficStreamWithKey>> readNextTrafficStreamChunk
         }, kafkaExecutor);
     }
 
+    /**
+     * Keeps the Kafka consumer alive while waiting for synthetic closes to drain.
+     * Returns an empty list if the gate is active (caller should return immediately),
+     * or null if the gate is clear and real records should be fetched.
+     */
+    private List<ITrafficStreamWithKey> handleDrainGateIfActive(ITrafficSourceContexts.IReadChunkContext context) {
```
Aman-bv, line 337: Why did we add this, what is the purpose?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/traffic/source/TrafficStreamLimiter.java
```diff
@@ -88,6 +93,27 @@ public void doneProcessing(@NonNull WorkItem workItem) {
...
+        var inUse = maxConcurrentCost - available;
+        var queueDepth = workQueue.size();
+        heartbeatLogger.atInfo()
+            .setMessage("permits={}/{} inUse={} queueDepth={} utilization={}%")
+            .addArgument(available)
+            .addArgument(maxConcurrentCost)
+            .addArgument(inUse)
+            .addArgument(queueDepth)
+            .addArgument(() -> maxConcurrentCost > 0 ? (inUse * 100 / maxConcurrentCost) : 0)
+            .log();
```
Aman-bv, line 107: do we need this?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
```diff
@@ -77,7 +77,10 @@ public String getStatsString() {
     }
 
     /** Emit a periodic heartbeat log summarizing the accumulator state. */
-    public void logHeartbeat() {
+    /**
+     * Logs accumulator state and force-expires connections idle beyond 1.5x the connection timeout.
+     */
+    public void heartbeatAndExpireStaleConnections() {
```
Aman-bv, line 83: Why do we need the force expire, we already have an expiry config.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
```diff
@@ -208,7 +208,10 @@ private TrackedFuture<String, Channel> closeClientConnectionChannel(ConnectionRe
                                 .addArgument(session::calculateSizeSlowly)
                                 .log();
                         }
-                        session.schedule.clear();
+                        session.schedule.drainWithCancellation(
+                            new java.util.concurrent.CancellationException(
+                                "Connection closed with pending work for " + session.getChannelKeyContext()));
+                        session.scheduleSequencer.cancelAllWork();
```
Aman-bv, line 214: This was causing head of line blocking and freezing the offset.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
```diff
@@ -89,14 +92,8 @@ private long getUpdatePeriodMs() {
...
 
     private void updateContentTimeControllerWhenIdling() {
-        if (isWorkOutstanding()) {
-            return;
-        }
         var currentSourceTimeOp = timeShifter.transformRealTimeToSourceTime(Instant.now());
         if (currentSourceTimeOp.isEmpty()) {
-            // do nothing - the traffic source shouldn't be blocking initially.
-            // Leave it manage its own initialization since we don't have any better information about what a
-            // start time might be yet.
```
Aman-bv, line 99: Do we need to remove this logic and comment?

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
```diff
@@ -257,8 +257,9 @@ public void recordTargetResponseCode(int statusCode) {
 
     /** Emit a periodic heartbeat log summarizing the replay engine state. */
     public void logHeartbeat() {
+        var outstanding = totalCountOfScheduledTasksOutstanding.get();
```
Aman-bv, line 263: var not needed/unnecessary change

TrafficCapture/trafficReplayer/src/main/resources/log4j2.component.properties
```diff
@@ -1,2 +1,8 @@
 Log4jContextSelector=org.apache.logging.log4j.core.async.AsyncLoggerContextSelector
-AsyncLogger.WaitStrategy=Block
\ No newline at end of file
+# NEVER use Block — it stalls caller threads (including kafkaConsumerThread and Netty
+# event loops) when the ring buffer is full, causing offset stalls and consumer evictions.
+AsyncLogger.WaitStrategy=Timeout
+AsyncLogger.Timeout=10
+log4j2.asyncQueueFullPolicy=Discard
+log4j2.discardThreshold=WARN
+AsyncLogger.RingBufferSize=65536
```
Aman-bv: Lets not change anything here

coreUtilities/src/main/java/org/opensearch/migrations/utils/TrackedFuture.java
Aman-bv: During connection drain (cancelConnection → closeClientConnectionChannel), a CancellationException propagates through the tracked future chain. When `propagateCompletionToDependentFuture` calls `setParentDiagnosticFuture` on a future that already has a parent set (from a retry path), the ISE crashes the completion handler. This leaves the offset permanently stuck in the OffsetLifecycleTracker because `processCompletedTransaction` never runs.

Converting to a warning log lets the diagnostic chain degrade gracefully (retaining the original parent) instead of killing the completion pipeline. The at-least-once delivery guarantee is preserved — we still commit or skip based on `completionStatus`.

coreUtilities/src/main/java/org/opensearch/migrations/utils/TrackedFuture.java
Aman-bv: Good call — restored the comment in the latest push. It explains *why* the ancestry tree is cut at this point, which is non-obvious from the code alone.

TrafficCapture/dockerSolution/src/main/docker/kafka-consumer.properties
Aman-bv: Agreed — reverted in the latest push. These are deployment-specific tuning knobs that belong in the runtime configuration, not in the defaults shipped with the codebase.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/kafka/KafkaTrafficCaptureSource.java
Aman-bv: This is a refactored extraction of the inline drain-gate logic that was previously in `readNextTrafficStreamSynchronously`. The original code blocked the read loop when synthetic closes were outstanding, but had two problems:

1. **No timeout** — if the outstanding counter leaked (e.g. from the session-key mismatch bug fixed in the earlier commit), the read loop would block forever
2. **No consumer keepalive** — while blocking, `poll()` was never called, so the cooperative rebalance protocol couldn't complete and consumers got evicted after `max.poll.interval.ms`

This method adds a 5-minute circuit-breaker timeout and calls `touch()` (which triggers `poll()`) to keep the consumer alive during the drain phase. The extraction itself was done to reduce the cognitive complexity of the parent method (flagged by SonarQube).

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/traffic/source/TrafficStreamLimiter.java
Aman-bv: Yes — this is critical observability for diagnosing stuck-offset scenarios. When the semaphore permits are exhausted (utilization=100%), no new traffic streams can be admitted for processing, which means no offsets advance. Without this heartbeat, the only symptom visible externally is growing consumer lag with no indication of *why*. The limiter saturation was one of the contributing factors in the production incident that motivated this PR.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/CapturedTrafficToHttpTransactionAccumulator.java
Aman-bv: The existing expiry config is packet-timestamp-driven — `ExpiringTrafficStreamMap.expireOldEntries()` only fires when new packets arrive via `addObservation()`. When backpressure is engaged and no new records arrive for a connection, the expiry for that connection *never* fires (we observed connections stuck for 57 minutes despite a 6-minute timeout in production).

The wall-clock safety net in the heartbeat provides a fallback: if the packet-driven expiry hasn't run and a connection has been idle for 1.5x the configured timeout by wall clock, force-expire it. This ensures stuck connections don't permanently block offset advancement even when backpressure throttles new record delivery.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ClientConnectionPool.java
Aman-bv: Exactly right. `schedule.clear()` silently discarded pending work items without completing their futures. Those futures are tracked by the OffsetLifecycleTracker — if they never complete, their offset never advances, and the min-heap blocks all subsequent commits for that partition. `drainWithCancellation()` + `cancelAllWork()` properly completes each future with a CancellationException, which propagates through to `processCompletedTransaction` → `commitTrafficStreams` (now in a finally block), allowing the offset to advance.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
Aman-bv: Yes — the `isWorkOutstanding()` guard is the root cause of the stuck-offset bug. When work is outstanding (which is the *normal steady state* under load), this guard prevented `updateContentTimeControllerWhenIdling()` from advancing the content-time controller. This froze the backpressure gate (`stopReadsPast` never advanced), which prevented new records from being read, which prevented packet-timestamp-driven expiry from firing, which left connections stuck indefinitely.

The guard was removed intentionally so the gate always advances regardless of whether work is outstanding. The comment about `currentSourceTimeOp.isEmpty()` has been restored in the latest push — that one is still valid and explains the initialization behavior.

TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/ReplayEngine.java
Aman-bv: Pushing back on this one — `var outstanding` captures the atomic read once and reuses it 3 times in the method (line for tasksOutstanding, workOutstanding flag, and the stale-duration guard). Without it, we'd either call `.get()` on the volatile atomic 3 times (getting potentially inconsistent values within a single heartbeat log) or call `isWorkOutstanding()` which is less clear about *which* counter is being checked. The local var gives a consistent snapshot and makes the heartbeat internally coherent.

TrafficCapture/trafficReplayer/src/main/resources/log4j2.component.properties
Aman-bv: Agreed — reverted in the latest push. The async logger configuration should stay as-is in the codebase defaults.

Aman-bv:
Hey @gregschohn,
Can you please review this PR when you get a chance. Thanks!

PR #3235, opened 2026-07-19
Harden solr cloud vs standalone detection logic.

### LINKED ISSUE #3232: Harden SolrCloud vs. standalone detection in SolrBackupStrategy
`SolrBackupStrategy.isSolrCloud()` currently determines whether a Solr source is SolrCloud or standalone by making a single request to the Collections API and treating only an HTTP 200 as "SolrCloud." Every other outcome falls through to "assume standalone." This is in `CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java`.

This means a misconfigured or temporarily unreachable endpoint is silently classified as Solr standalone, and the migration keeps going against the wrong mode instead of surfacing the real problem.

Any response that isn't 200, plus any timeout/error, is coerced to standalone. A wrong endpoint therefore produces a false "standalone" result rather than a failure.

**Proposed changes**

* Distinguish "not SolrCloud" from "couldn't tell."
* Inspect the exceptions/status returned so that connectivity failures (unreachable host, TLS errors, etc.) are reported as errors instead of being swallowed into a standalone assumption.
* Increase the detection timeout (currently 5s) to reduce misclassification from transient network hiccups.
* Consider failing fast (or at least warning loudly) when the endpoint appears unreachable, rather than proceeding in a possibly-wrong mode.

Raised in review of PR #3209: https://github.com/opensearch-project/opensearch-migrations/pull/3209#discussion_r3572744632

jzonthemtn:
### Description
Hardens the SoleCloud vs Solr Standalone detection logic.

Detection now distinguishes three outcomes instead of two:
1. 200 → SolrCloud
2. Reachable non-200 (standalone Solr returns HTTP 400 "not running in SolrCloud mode") → standalone
3. 401/403 or connection failure (unreachable host, refused connection, TLS error, timeout) → throw `SolrTopologyDetectionException` instead of guessing standalone

Increased the probe timeout from 5s to 10s to avoid transient connection errors.

### Issues Resolved
Closes #3232 

### Testing
Added unit tests.

### Check List


CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -503,18 +518,35 @@ private void resolveCollections(String solrUrl) {
...
+        if (status == 401 || status == 403) {
+            // Access-controlled: can't read the Collections API to decide, so don't guess standalone.
+            throw new SolrTopologyDetectionException(
+                "Solr authentication/authorization failed (HTTP " + status + ") while detecting topology at "
+                    + solrUrl + "; cannot determine SolrCloud vs standalone");
+        }
+        // Reachable non-200 (standalone Solr returns HTTP 400 "not running in SolrCloud mode").
+        log.info("Solr topology detection: HTTP {} from Collections API at {} — treating as standalone",
+            status, solrUrl);
+        return false;
```
malliaridis: I would be explicit here too. In case of 404 (not found, in case of a changed base path) or 500 errors we would assume it's Solr standalone.

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -503,18 +518,35 @@ private void resolveCollections(String solrUrl) {
     // ---- Cloud vs standalone detection ----
 
     static boolean isSolrCloud(String solrUrl, SolrHttpClient httpClient) {
```
malliaridis: I do not like the return type of this function, because we either say `true` it is SolrCloud, or `false` it is Standalone`, rather than using enums. Of course you can say "we have only two states that can be represented with a boolean", but:
- in case of 500 errors we assume it is Solr standalone
- in case of 404 caused by changed base path (not `/solr/admin/collections` but `/tenant/admin/collections`, reverse proxies use case) we assume it is Solr standalone
- in case of 300-redirects it is standalone(?)

Instead, we should

- be explicit and have `isStandalone()` as well, making `isSolrCloud()` returning `false` not automatically assuming it is standalone (case where both functions return `false`)
- return an enum with 3 values, `Unknown`, `Standalone` or `SolrCloud`, where `Unknown` could be also extended with `Unavailable`
- return a `Boolean` or `null` and support that third state

malliaridis (commented):
The changes make sense and are hardening the SolrCloud vs Standalone detection logic. 

I have added a few more cases where this would still fail, and what I would do instead. However, this would require further changes at usage level too, and probably be a bit more complex to address.

malliaridis (approved):

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
jzonthemtn: @malliaridis I wrote up https://github.com/opensearch-project/opensearch-migrations/issues/3265 to capture the need to be able to change the `/solr` context path.

I pushed changes to this PR to add the enum.

malliaridis: With `UNKNOWN` defined as "reachable but can't identify" this looks solid.

One small note, if `401` and `403` responses are considered reachable, then I think returning `UNKNOWN` instead of throwing would be better. But if considered "unreachable" due to missing authentication / unauthorized, then this the current implementation better. :)

malliaridis (approved):

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -534,19 +549,57 @@ private void resolveCollections(String solrUrl) {
...
-    static boolean isSolrCloud(String solrUrl, SolrHttpClient httpClient) {
+    public enum SolrTopology {
+        SOLR_CLOUD,
+        STANDALONE,
+        /** Reachable, but the response does not identify either topology. */
+        UNKNOWN
+    }
+
+    static SolrTopology detectTopology(String solrUrl, SolrHttpClient httpClient) {
+        var listUrl = solrUrl + "/solr/admin/collections?action=LIST&wt=json";
```
gregschohn: Let's use the mode property in GET /solr/admin/info/system?wt=json instead

jzonthemtn: Change to hitting `GET /solr/admin/info/system?wt=json` and parse the `mode` property from the response, i.e:

```
{
  "responseHeader": {
    "status": 0,
    "QTime": 2
  },
  "mode": "std",
  "solr_home": "/var/solr/data"
}
```

jzonthemtn:
Now getting the Solr mode from `GET /solr/admin/info/system`.

jzonthemtn:
Marking as draft while I work through the `config-read` permission needed for the Solr endpoint.

jzonthemtn:
Keep the Collections API, but stop inferring from the status code. Parse the body for explicit markers instead:

```
  if (body.has("collections")) {
      return SolrTopology.SOLR_CLOUD;
  }
  if (body.path("error").path("msg").asText("").toLowerCase(Locale.ROOT).contains(NOT_CLOUD_MARKER)) {
      return SolrTopology.STANDALONE;
  }
```

docs/SolrBackupLayouts.md
```diff
@@ -40,6 +40,24 @@ A standalone backup does **not** record the core name, so the target index name
...
+## Solr permissions the migration needs
+
+When the migration connects to a live Solr (via `--source-username` / `--source-password`), it reads
+the Collections and Cores admin APIs to determine whether Solr is running in SolrCloud or standalone
+mode and to discover what to migrate. Under Solr's `RuleBasedAuthorizationPlugin` those need:
+
+| Permission | Used for |
+| --- | --- |
+| `collection-admin-read` | `/admin/collections?action=LIST` — topology detection and collection discovery |
+| `core-admin-read` | `/admin/cores?action=STATUS` — core discovery on standalone |
```
gregschohn: is there a link to some documentation that we can add to show a user how to make these adjustments and so that they can understand the implications?

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -573,31 +586,54 @@ static SolrTopology detectTopology(String solrUrl, SolrHttpClient httpClient) {
...
-        if (status == 200) {
-            return SolrTopology.SOLR_CLOUD;
+        if (status == 401) {
+            throw new SolrTopologyDetectionException(
+                "Solr authentication failed (HTTP 401) while detecting topology at " + solrUrl
+                    + "; cannot determine SolrCloud vs standalone. Check the source credentials.");
         }
-        if (status == 401 || status == 403) {
-            // Access-controlled: can't read the Collections API to decide, so don't guess standalone.
+        if (status == 403) {
```
gregschohn, line 795: Is this API call our only option?  The minimalist argument would probably ask what are the API calls that we NEED to make anyway for creating a backup - and separately for augmenting a backup for BYOS (which might be different).

For those APIs, is there already no way to infer from the results if it was cloud or standalone?  If not, do we even need to know beforehand or at all?  Is there post-processing on the bucket that we'd need to do?

Ideally, we'd be able to get a PIT backup triggered from a source w/out requiring any changes - or only requiring minimal changes - to the source.  Asking a user to open up some new APIs could be impractical for some locked down (in controlled pipelines and security hardened) environments

docs/SolrBackupLayouts.md
jzonthemtn: Removing the need for these permissions based on our conversation last week. Requiring users to grant new permissions on a legacy installation just to migrate is a potentially a large burden.

jzonthemtn:
> Marking as draft while I work through the `config-read` permission needed for the Solr endpoint.

Changing to not require any additional permissions.

jzonthemtn:
I reworked this some to simplify it.

Detection no longer happens up front. CREATE just attempts the SolrCloud BACKUP and treats a "not running in SolrCloud mode" rejection as the signal to use the replication handler, while IMPORT reads the topology from markers in the backup itself. (A marker being filenames specific to each one.) Anything ambiguous now fails loudly instead of silently defaulting to standalone.

There are no API requests made to determine if it's Solr Cloud or standalone. If an IMPORT doesn't have the markers, it falls back to asking for a `--solr-topology` param. This avoids needing Solr admin permissions.

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
jzonthemtn: Updated to not make API requests to determine Cloud vs standalone.

CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java
```diff
@@ -416,13 +450,43 @@ private String fetchViaSchemaApi(String solrUrl, String collection) {
...
+            var explicitlyCloud = cloudTopology != null;
+            try {
+                runCloudBackup(solrUrl, backupLocation, parsedUri);
+                cloudTopology = Boolean.TRUE;
+                logResolvedTopology();
+                return;
+            } catch (SolrSnapshotCreator.SolrBackupFailed | SolrHttpClient.SolrRequestException e) {
+                // Standalone's rejection arrives as HTTP 400, so it lands on either type.
+                // Only a positive match redirects; auth and repo errors surface as-is.
+                if (explicitlyCloud || !isNotSolrCloudRejection(e.getMessage())) {
```
gregschohn, line 474: can you document how broadly you've tested solr versions for the message match?  Should it take the status code in too?  The test tells me 6.6 -> 9.1 (I think) - but it would good to put that here or around line 488

gregschohn (approved):

PR #3248, opened 2026-07-21
Cleanup and update developer guide and deployment instructions

### LINKED ISSUE #3176: Orphaned/diverging quickstart.md
## Description
 
`deployment/k8s/quickstart.md` exists, but nothing links to it — not `README.md`, not `DEVELOPER_GUIDE.md`, not `deployment/k8s/README.md`. A developer can only find it by browsing the repo directly or searching GitHub.
 
That would just be a "hard to find" problem, except the file also seems to have drifted from the canonical Quick Start in `deployment/k8s/README.md`:
 
| | `deployment/k8s/quickstart.md` | `deployment/k8s/README.md` Quick Start |
|---|---|---|
| Cluster setup | `./minikubeLocal.sh --start` | `./localTesting.sh` (Minikube) or `./kindTesting.sh` (kind) |
| Image build | Implied separately, then `helm install` | Handled by `localTesting.sh`/`kindTesting.sh` |
| Helm install | Manual `helm install ma -n ma charts/aggregates/migrationAssistantWithArgo -f charts/aggregates/migrationAssistantWithArgo/valuesForLocalK8s.yaml` | Done automatically by the wrapper scripts |
| Source/target test clusters | Not set up at all | Set up as part of the wrapper scripts |
| kind support | Not mentioned | Documented |
 
A new contributor who finds `quickstart.md` first (e.g. via search) will end up with a different, incomplete environment (no test clusters) and no path to the kind-based flow.
 
## Proposed fix
 
- **Delete `quickstart.md`** and verify nothing outside this repo links to it (e.g. project wiki, blog posts). The Quick Start in `deployment/k8s/README.md` already supersedes it. Or instead
- **Keep it, but make it the canonical source** — move the merged Minikube + kind Quick Start content into `quickstart.md`, and have `deployment/k8s/README.md` link to it instead of duplicating it inline.

Either way, do a repo-wide search for references to `quickstart.md` (including the wiki, if accessible) before removing/changing it, to avoid leaving a dead link elsewhere.
 
## Acceptance criteria
 
- [ ] Confirm whether anything currently links to `deployment/k8s/quickstart.md` (wiki, other docs, external).
- [ ] Either remove `quickstart.md` or make it the single source of truth and have `deployment/k8s/README.md` reference it.
- [ ] No two documents in the repo describe two different ways to do the same local Quick Start.

malliaridis:
### Description

This PR focuses on cleaning up the deployment files and documentation that are either outdated or confusing. The main changes include:
- ~~Relocating local deployment scripts to clearly distinguish them from general-purpose files~~
- Updating links in READMEs to have a cleaner path to follow as a new developer
- Reorganize information provided by READMEs based on the new directories to provide only relevant information and keep the focus on the current context

### Issues Resolved

Closes #3176 

### Testing

_No tests have been executed_

### Check List


buildImages/scripts/README.md
```diff
@@ -85,7 +85,7 @@ helm template ma ../deployment/k8s/charts/aggregates/migrationAssistantWithArgo
 
 #### Port Forwarding
 - argo server only: `kubectl -n ma  port-forward service/argo-server 8001:2746`
-- all services: `../deployment/k8s/forwardAllServicePorts.sh`
+- all services: `../deployment/k8s-local/forwardAllServicePorts.sh`
```
gregschohn: Port forwarding works for remote K8s deployments too and is quite useful (when it isn't flaky)

deployment/k8s/aws/README.md
```diff
@@ -152,6 +152,53 @@ migration-assistant cleanup --stage prod
 Helm-uninstalls the release, deletes the CFN stack, and archives state
 to `~/.opensearch-migrate/<stage>/archive/<timestamp>/`.
 
+## Manual AWS Add-ons Setup
```
gregschohn: Why would anybody need to manually do these things?  I've tried to keep everything within scripts.
That lends itself to continuously testing them and is also self-documenting.
If we think that users might not understand how CSIs are getting set up, we should point them to the Cfn (or terraform) templates.
I'm fine explaining what is or can be added, but I would not frame it in the form of more manual commands for a user to run.  This should be cast in terms of our IAC.  Getting users/devs to work outside of IAC is a tax and antipattern.

deployment/k8s/charts/README.md
```diff
@@ -0,0 +1,92 @@
+# Migration Assistant Helm Charts
+
+## Migration Assistant environment
+
+The [Migration Assistant](aggregates/migrationAssistantWithArgo) helm
+chart consists of:
+* The Migration Console stateful set (a shell for users to run workflow commands to perform migration tasks)
+* Argo Workflows (used by the workflow commands to dynamically provision and manage resources that perform a migration)
+* Strimzi (to create Kafka clusters)
+* Observability services - Prometheus, Jaeger, and Grafana
```
gregschohn: We need to call Jaeger and Grafana as optional and only enabled for local deployments by default.
Notice that we don't test the utility of any of these services yet.

gregschohn (commented):
I've only started to go through this PR.  I don't like splitting k8s local and k8s - let's fold those back together.  It opens the door to treating different k8s installations differently.

Let's also REMOVE every notion of minikube from the docs, integ tests, and scripts.  I don't see a reason to support both & kind + EKS seems like it's just as valuable as kind + eks + minikube.  Doing that will shrink the surface/mass and should make a separate 'local' directory seem less critical.

buildImages/scripts/README.md
malliaridis: I only saw local setup guides mentioning the forwarding script, so I considered it a local script as well. The commands in there are `kubectl` commands, so should not matter which cluster we are using. 👍

deployment/k8s/aws/README.md
malliaridis: I agree. I didn't want to remove this section because I don't know if it is of any use, and I only moved it from another place into here, as it is AWS-specific. I can remove it completely if that's ok.

malliaridis:
If we remove minikube I agree that a separate k8s may be less valuable.

By removing minikube we can also merge in a separate PR `localTestingCommon.sh` with `kindTesting.sh`, as we do not reuse this logic anymore and would reduce the amount of scripts we have (plus any logic calling scripts from other scripts).

malliaridis:
Since the removal of minikube also requires the migration of other scripts and tests, I would address it in a separate PR. This Pr now focuses only on documentation updates and removes the mentioning of minikube in the dev docs.

jzonthemtn (approved):
I'm not an expert on the local testing stuff but the doc updates make sense.

deployment/k8s/tests/mountableTransformsImageVolumeTest.sh
gregschohn, line 1: why did you change the mode on this and some of the other shell scripts?
If it was inadvertent, please change them back.  A lot of the documentation (where we run these scripts and run them from each other) will break otherwise.

deployment/k8s/KindGuide.md
gregschohn, line 1: make this camelCase to match the pattern in the rest of the file

deployment/k8s/quickstart.md
gregschohn, line 1: What's the rationale for deleting this content?  If somebody is new to k8s, it seems like this was a reasonable place for them to get started.  I'm not going to argue that it's perfect, but I'd rather not throw non-k8s devs into the deep end.

deployment/README.md
gregschohn, line 1: This might be a style thing, but I'm a big fan (if not always proponent) of keeping a README files that can navigate the user through the repository.  What's added by removing this file?  We do have multiple deployment modes (and indeed, one is missing from this) and having a Table-of-Contents landing page above  those in the repo seems valuable.

TrafficCapture/dockerSolution/README.md
```diff
@@ -1,7 +1,10 @@
-# Docker Solution — Local CDC Dev Environment
+# Legacy Docker Solution — Local CDC Dev Environment
```
gregschohn, line 1: THANK YOU!

DEVELOPER_GUIDE.md
```diff
@@ -18,19 +18,23 @@
...
-## Kubernetes Quick Start
+## Quick Start
 
-* This [Kubernetes Guide](deployment/k8s/README.md) shows you how to create a minikube cluster locally and deploy the Migration Assistant to it.
-* This [AWS EKS Guide](deployment/k8s/aws/README.md) shows you how to deploy an EKS cluster and deploy the Migration Assistant to it.
+To get started with the Migration Assistant, you will need to deploy it to a Kubernetes environment.
 
-See the [project wiki](https://github.com/opensearch-project/opensearch-migrations/wiki)
-to learn more about how to use the migration console and its workflow commands.
+Please refer to the [Migration Assistant Deployment Guide](deployment/README.md)
```
gregschohn, line 31: Haven't you deleted this file?

deployment/k8s/README.md
```diff
@@ -31,260 +42,19 @@ helm install --create-namespace -n ma ma \
...
-If you want the kind cluster to run on OrbStack instead of Docker Desktop,
-switch the active Docker context before running the script so `docker` and
-`kind` target the same backend.
+Details regarding the deployment of the Migration Assistant solution can be found in
+[Migration Assistant Solution Deployment](../migration-assistant-solution/README.md).
 
-## Deploying
+### Local Deployment with kind
 
-### What Helm Manages (and what it doesn't manage)
```
gregschohn, line 154: If some of this is outdated, was never a good idea, etc, we should shape this more.
We could also move it into design docs if that seems like a more appropriate place.

However, this is a biproduct of very well-thought out discussions and designs about how to structure a migrations environment within k8s.  We went through several iterations and there's nothing to say that we shouldn't go through more.  We should explain helm responsibilities and how it's used.  It's a big part of the overall solution and it might look obvious now, but from a dead-start, it was not.

deployment/k8s/tests/mountableTransformsImageVolumeTest.sh
gregschohn, line 1: Why did the file mode change here.  A lot of things will break because of these mode changes

deployment/k8s/quickstart.md
gregschohn, line 1: What's the value in removing this file?  I'd like to keep the repo accessible for non-k8s devs.

deployment/k8s/README.md
```diff
@@ -1,27 +1,38 @@
...
+chart, which will install the migration console and resources required for it to perform migrations (e.g. Argo
+Workflows, metrics collectors, etc).
 
+**Notice**: The user is responsible for the cost of any underlying infrastructure required to operate the solution. We
+welcome feedback and contributions to optimize costs.
 
 ## Quick Start
 
-### EKS
+### AWS EKS Deployment
```
gregschohn, line 18: This should be aligned with the "GKE" section below.
The service is actually called 'Amazon Elastic Kubernetes Service" - Amazon, not AWS

deployment/k8s/README.md
```diff
@@ -31,260 +42,19 @@ helm install --create-namespace -n ma ma \
...
-This script will start minikube, build the images from source, and will install
-the Migration Assistant Helm chart alongside a test chart that creates a source
-and target cluster.
-
-```bash
-echo "Will start minikube, build images, and install the MA helm chart for those images"
-$(git rev-parse --show-toplevel)/deployment/k8s/localTesting.sh
-```
-
-### kind
```
gregschohn, line 51: Why are you dropping kind from this section?  It could use help, but I would group it with the others so that new devs/users know that this is also a first class citizen.

deployment/k8s/README.md
```diff
@@ -31,260 +42,19 @@ helm install --create-namespace -n ma ma \
...
-```
-
-To forward deployed services' ports from minikube to your localhost.  
-E.g. so that http://localhost:2746/ will load the argo web-ui, etc, run
-
-```bash
-$(git rev-parse --show-toplevel)/deployment/k8s/forwardAllServicePorts.sh
-```
-
-## Prerequisites
```
gregschohn, line 96: This looks like it is just removed.  Can we put it somewhere else?  Do you feel that it already is somewhere else?  If so, let's link to it.  This is a little bit tricky because to build a replayer, you need Java.  You don't need k8s/helm.  To build helm charts from scratch, you DO need Java.  I'd prefer to reduce the level of normalization (references) when information is pretty stable.

gregschohn (changes requested):

deployment/k8s/quickstart.md
malliaridis: The main reasons for removing this file was that it contained instructions that were already present elsewhere (README.md), and that this file is nowhere referenced. New devs would likely find it after already looking into the code and project structure, and would already have followed other instructions found by the README.md and developr guide.

In my opinion it should not be the responsibility of the project to explain what the tools that are required do. Providing links to the tools installation page is probably the max I would provide in a project.

However, if you believe it contains valuable information, I'd like to propose that we move these details to the DEVELOPER_GUIDE.md, as it is normally the entry point and could explain a bit that "we are using kubernetes..." "the developer tools used for kubernetes are... and do...", as well as "this is a gradle project..." and "you are running gradle tasks like this...".

If you would prefer keeping this file as is, I can add it back and add 1-2 references to it, so that devs can find it, and update the minikube instructions to use kind.

Just let me know what action you prefer. :)

deployment/k8s/KindGuide.md
malliaridis: I updated it to PascalCase, I think that is the convention most used by markdown files in this project. Hope that is the convention to use?

_There are a couple files not following a specific convention either, making it hard to figure out the right pattern._

deployment/README.md
malliaridis: I agree. I first wanted to maintain this for navigation purposes, then I moved things into deployment/k8s as we had there the navigation overview already, so it kinda duplicated the deployment options there.

deployment/k8s/README.md
malliaridis: This section moved to `deployment/k8s/charts/README.md`, as it is closer to the charts this way. It could be moved to design docs instead. What do you think?

deployment/k8s/tests/mountableTransformsImageVolumeTest.sh
malliaridis: Windows issue when moving files around, should be reverted now. Thanks for catching this.

deployment/k8s/README.md
malliaridis: The title was changed to "Local Deployment with kind" in this file and the detailed instructions are moved to a separate file to keep the file focused on providing an "Kubernetes Deployment Overview". The idea was to combine the `quickstart.md` with `KindGuide.md`, as they would contain almost the same content instruction-wise, and reduce the amount of duplicate documentation.

deployment/k8s/README.md
malliaridis: I was consuming the files from a first-developer's perspective, starting with the /README.md, moving forward to the /DEVELOPER_GUIDE.md, where the prerequisites contain more information, then moved forward to these files via references, where this would repeat itself. I found this list incomplete and therefore considered dropping it would be better, since the readers already consumed a "prerequisites" list from where they usually come from.

deployment/k8s/tests/mountableTransformsImageVolumeTest.sh
malliaridis: Windows issues, should not have been committed in the first place. Thanks

deployment/k8s/README.md
malliaridis: What exactly do you mean with "alignment"? Is this deployment following a terraform with charts deployment like GKE?

PR #3254, opened 2026-07-23
Enable the failed document stream only when a bucket is configured

### LINKED ISSUE #3252: Enable the failed document stream only when a bucket is configured
The RFS failed document stream (added in #3229) writes terminal document failures to a durable S3 stream. It is enabled whenever an S3 bucket resolves for the backfill, and there is currently no explicit way to turn it off. (For an AWS migration, the S3 bucket will always resolve meaning the failed document stream cannot be disabled.)

Each failed-document record includes the full original document source and this may not always be needed during a migration. It may also not always be allowed due to data governance or PII requirements.

jzonthemtn:
### Description
Adds an explicit opt-out for the failed document stream.

Failed document stream is currently only disabled if `failedDocumentStreamS3Bucket` is not set. However, when not set it defaults back to the `defaultS3Bucket`, which means it can never be disabled.

### Issues Resolved
Closes #3252 

### Testing
Added unit and integration tests.

### Check List
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.


DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
wrigleyDan: Maybe it's worth adding `--failed-document-stream-enabled` in the Javadoc because it is the most straightforward way to disable recording these.

wrigleyDan (approved):
Took me a little to understand how `--failed-document-stream-enabled` and `--failed-document-stream-s3-bucket configured` interact with each other because I had little to do with these before but I got it in the end.
Added one suggestion for the Javadoc. Other than that it looks good to me.

gregschohn (changes requested):
Please confirm the command line options for using the bucket - ideally we get to just one command that sets the bucket and uses it or just disables the feature rather than doing it with two options

jzonthemtn:
> Please confirm the command line options for using the bucket - ideally we get to just one command that sets the bucket and uses it or just disables the feature rather than doing it with two options

If a failed document stream bucket was not provided, it was falling back to the default bucket:

```
// migrationConfigTransformer.ts:451 (on main)
const bucket = userBucket ?? trimToUndefined(deploymentDefaults.defaultS3Bucket);
```

This caused the failed document stream to never be disabled -- this PR fixes that. If no bucket, then no failed document stream. (It removes that fallback to default.)

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -946,12 +932,23 @@ static String resolveSessionId(Args arguments, String workerId) {
         return "worker-" + workerId;
     }
 
+    static final String FAILED_DOCUMENT_STREAM_DISABLED_REASON =
+        "failed document stream disabled: no --failed-document-stream-s3-bucket configured";
+
+    static void logFailedDocumentStreamStatus(FailedDocumentStreamSink sink, String sessionId) {
+        if (sink == null) {
+            log.atInfo().setMessage(FAILED_DOCUMENT_STREAM_DISABLED_REASON).log();
```
gregschohn: This should be a warning since w/out it we cannot tell a user what landed and what didn't

gregschohn (approved):

PR #3255, opened 2026-07-24
Route GKE workload logs to Cloud Logging; fix GCP Terraform docs

tonypiazza:
## Description

Ships migration workload logs from GKE deployments to Google Cloud Logging, and
corrects drift in the GCP Terraform README.

### Logging (functional change)
- Adds a fluent-bit `stackdriver` output for the GKE values path, so workload
  logs (migration console, RFS/backfill workers, Argo workflow steps, and — for
  live migrations — capture proxy and traffic replayer) reach Google Cloud
  Logging. Authentication uses Workload Identity / ADC (no credential files).
- Grants the node service account `roles/logging.logWriter`.
- Adds a `gcp-metadata` ConfigMap carrying the cluster name and location the
  `k8s_container` resource requires, threaded from Terraform.
- Cloud Logging entries carry full resource labels (cluster, location,
  namespace, pod, container) and mapped severity, queryable per workload.

### Docs (README corrections)
- Removes documented Terraform variables that don't exist in the module
  (`cluster_type`/Autopilot, `bucket_name`, `cluster_name`) — the module is GKE
  Standard only and generates the bucket/cluster name as `os-migration-<random>`.
- Fixes stale `migration-snapshots-<project>` bucket paths in the Bring Your Own
  Snapshot section.
- Documents the observability behavior (logs shipped automatically; the bundled
  Grafana dashboard is opt-in via `conditionalPackageInstalls.grafana`).
- Widens the Terraform prerequisite to allow OpenTofu >= 1.6.

## Testing
- Validated end-to-end on a real GKE migration (backfill/RFS): workload logs land
  in Cloud Logging with correct resource labels and severity; zero collector
  errors. Backfill-only — capture proxy / traffic replayer use the same proven
  fluent-bit path but were not exercised by a live capture-and-replay run.
- `helm template` renders the GKE path; `terraform validate` passes.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -318,6 +318,19 @@ jobs:
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
 
+  gcp-terraform-test:
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+      - uses: hashicorp/setup-terraform@dfe3c3f87815947d99a8997f908cb6525fc44e9e # v4.0.1
+        with:
+          terraform_wrapper: false
+      - name: terraform test (GCP module)
+        # Plan-time tests only; no cloud credentials required.
+        working-directory: deployment/terraform/gcp
+        run: |
+          terraform init -backend=false
+          terraform test
 
   codeql-analyze:
     name: CodeQL (${{ matrix.language }})
@@ -462,6 +475,7 @@ jobs:
       - gradle-build-macos
       - codeql-analyze
       - sonarqube-analysis
+      - gcp-terraform-test
     if: always()
     runs-on: ubuntu-22.04
     steps:
--- a/buildImages/build.gradle
+++ b/buildImages/build.gradle
@@ -103,7 +103,8 @@ def buildKitProjects = [
                 imageName:  "migration_console_base",
                 buildArgs: [
                         BASE_IMAGE: RegistryImageBuildUtils.resolveBaseImageForContainer(
-                                intermediateRegistry, "migrations", "elasticsearch_test_console", "latest")
+                                intermediateRegistry, "migrations", "elasticsearch_test_console", "latest"),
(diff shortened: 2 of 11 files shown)


gregschohn (changes requested):
I want to see tests for these terraform configurations.  How should we know if something has broken in future updates to these files or in a TF version bump?

tonypiazza:
Good call — added `terraform test` coverage for the logging changes and wired it into CI so it actually gates future changes:

- **`tests/observability.tftest.hcl`** — plan-time, credential-free assertions that (1) the node SA keeps `roles/logging.logWriter`, (2) one IAM binding is planned per configured role, and (3) the cluster name/location are passed to the Helm release (what the fluent-bit stackdriver output needs for the `k8s_container` resource).
- **CI** — new `gcp-terraform-test` job runs `terraform test` on the GCP module (`init -backend=false`, no cloud creds) and is wired into `all-ci-checks-pass`. This also gates the pre-existing GCP terraform tests, which previously only ran manually — so a future edit or TF/provider bump that breaks the wiring now fails CI.

Full suite passes locally: 16 passed, 0 failed.

tonypiazza:
Added a third commit that's slightly outside the GCP scope, per discussion with @gregschohn: **Copy kafka-tools from a pinned image instead of downloading from Apache mirrors.**

The migration-console build was fetching the Kafka tools tarball from `dlcdn.apache.org` with a retry/fallback to `archive.apache.org`. That mirror ages out old releases and serves them slowly, so builds could crawl for ~17 min, and the primary-vs-backup split risks nondeterministic images. This replaces the download with a `COPY --from` off the pinned `apache/kafka:3.9.2` image, routed through the ECR pull-through cache like the other base images — deterministic, no retries, no Apache-mirror dependency. Stays on 3.9.2; no version bump.

Verified end-to-end on a Linux/buildkit environment: `migrationConsoleBase` builds green and the image has the kafka tools + msk-iam jar at `/root/kafka-tools/kafka` with `GetOffsetShell` runnable.

Bundled here rather than as a separate PR since it's a small, self-contained build fix.

gregschohn (approved):
Thanks for this - especially the build fix on top... way to take ownership!

PR #3263, opened 2026-07-27
Adding k6 loadtesting to k8s setup

awagen:
### Description
This PR adds the k6 setup into the k8s setup used in the project. 
It further integrates k6 runs as workflows to be started from within migration console shell (“workflow loadtest run …”) and managing of single runs or multiple via separate TUI ("workflow loadtest") within migrationConsole.

- the integration test supplied does not yet contain an actual system test, this will be added in another PR regarding scaling, which will then also contain options to scale all related resources of the workflow (current one just checks the run completes and that document counts in source and target agree)
- the charts live within deployments/k8s/charts/components/k6LoadTest. The installation is not baked into the normal migration console / workflow component installation, but must be separately installed via helm update (see README.md). These also bake each scenario into their own argo WorkflowTemplate to simplify starting different runs
- the related scenarios, variable presets are baked into a custom read-only docker image. The script paths are adjusted to match the image mount path
- an example setup of all involved components (capture-proxy, source, target, kafka, replayer and k6) can be found within the deployment/k8s/deployCdcWorkflow.sh
- the folder TrafficCapture/trafficLoadTest/ only contains example scripts of how to start up runs and all resources needed by the runs, such as setting presets, distinct scenarios and libs needed to run the scenarios. Those are baked into the custom k6_scripts read-only docker image.

### Check List


buildImages/backends/dockerHostedBuildkit.sh
```diff
@@ -44,16 +44,26 @@ ptc_rewrite_image() {
...
+  # is NOT reachable from inside the buildkit builder's own network namespace, so
+  # baking it into buildkitd.toml breaks all name resolution during the build
+  # (e.g. "Could not resolve host: cdn.amazonlinux.com"). Drop loopback
+  # (127.0.0.0/8) servers; if none remain, fall back to systemd-resolved's real
+  # uplink resolvers. If that still yields nothing, emit nothing so buildkit uses
+  # Docker's embedded resolver (127.0.0.11) on the user-defined network.
+  servers="$(printf '%s\n' "${dns_line}" | grep -oE '([0-9]{1,3}\.){3}[0-9]{1,3}' | grep -v '^127\.' || true)"
+
+  if [[ -z "${servers}" && -r /run/systemd/resolve/resolv.conf ]]; then
+    servers="$(grep -E '^nameserver' /run/systemd/resolve/resolv.conf | awk '{print $2}' | grep -v '^127\.' || true)"
```
gregschohn: I'm confused on this whole block.  Can you give me some more details on the problem that this is solving?

deployment/k8s/charts/components/k6LoadTest/files/k6/scripts/GENERATOR_logs_data_queries.js
```diff
@@ -10,9 +10,9 @@ import {
   makeAggSearch,
   makeSearchAfterSequence,
   scrollSequence,
-} from '../../query-utils.js';
+} from './LIB_query-utils.js';
```
gregschohn: nit: can we make the naming case-consistent at least for single names and projects.

TrafficCapture/trafficLoadTest/scenarios/ingest.js
```diff
@@ -8,9 +8,9 @@
  *   (EXECUTOR=ramping-arrival-rate): ramp / burst load shapes via RAMP_STAGES.
  *
  * Key environment variables (see k6-config/ingest-steady.env for load-profile defaults):
- *   SCENARIO           — document schema to use: "nyc_taxis" (default) or "logs_data"
+ *   SCHEMA           — document schema to use: "nyc_taxis" (default) or "logs_data"
```
gregschohn, line 11: i might have liked scenario better - is this JUST the metadata/schema or is it also the daata (and an inferred schema)?

deployment/k8s/charts/components/k6LoadTest/files/grafana/load-test.json
gregschohn, line 1: I'm uncomfortable with putting so much of the testing code directly within the definitions.  We're embracing OCI images as volumes to mount transforms and TLS materials.  Do you think it would make sense to keep the definitions in their own image that we build/publish locally, then have this chart mount that image with all of these files directly into the k6/k8s environment as necessary - as part of its helm installation.

That distinction would also make the contents of the chart a lot clearer - right now some stuff is load testing code and other stuff is for the k6 environment.  Lastly, needing to work around the template-processing that helm does is eventually going to be painful.  Do you want a helm variable to adjust a test as it is being deployed?  Probably not.

deployment/k8s/charts/components/k6LoadTest/templates/k6-testrun-examples.yaml
gregschohn, line 1: This comment is an extension of my above concern of putting too much into helm.

A configmap seems like an inferior choice to store another k8s resource definition that you might want to kick off.  Given that we have hard-dependencies on argo-workflows, this would be cleaner to do as separate workflows.  You could make a separate template for each one and kick them off w/ kubectl create workflow... (still somewhat verbose since we need to spell out the yaml/json)

deployment/k8s/deployWorkflowComponents.sh
gregschohn, line 1: Let's rename this script - it's name is way too general to be in this directory.

gregschohn: Can you convert this script to invoke a workflow from a configuration with a proxy and replayer between a source and a target.  You should be able to poll on the CR states for the proxy/replayer to know when they're ready.

Users should be able to express any setup of CDC with the workflow and we're not going to be recommmending them to do manual setup like this.

deployment/k8s/fillLocalRegistry.sh
```diff
@@ -11,6 +11,21 @@ MIGRATIONS_REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
 
 docker network inspect local-migrations-network >/dev/null 2>&1 || docker network create local-migrations-network
 
+# A container that exists but isn't running (exited, or stuck in a restart loop) would otherwise be
+# treated as "already there" by the checks below, silently reusing a broken container forever. Drop
+# it so it gets recreated with the current arguments.
+drop_if_not_running() {
```
gregschohn: Is this a fully generalized sort of fix or should a dev drop in and do some diagnosis into why the containers have gone dark?  
Prove to me that this method is required for non-buggy cases and won't create more complexity.

migrationConsole/lib/console_link/console_link/workflow/commands/k6.py
```diff
@@ -0,0 +1,501 @@
...
+            "scenario": meta.get("labels", {}).get("k6-scenario", "?"),
+            "phase": stage or "unknown",
+            "age": _age(meta.get("creationTimestamp")),
+        })
+    out.sort(key=lambda r: r["name"])
+    return out
+
+
+def _wait_for_testrun(namespace, name, timeout, interval):
+    """Poll a TestRun until it reaches a terminal stage; return the stage, or None on timeout."""
```
gregschohn: what is a stage here?  What are the possible values?

migrationConsole/lib/console_link/console_link/workflow/commands/loadtest.py
```diff
@@ -0,0 +1,501 @@
...
+            "age": _age(meta.get("creationTimestamp")),
+        })
+    out.sort(key=lambda r: r["name"])
+    return out
+
+
+def _wait_for_testrun(namespace, name, timeout, interval):
+    """Poll a TestRun until it reaches a terminal stage; return the stage, or None on timeout."""
+    deadline = time.time() + timeout
+    while time.time() < deadline:
```
gregschohn: Are we throwing some information to the logs periodically?  What if this test takes an hour, will we show a log line every minute or 5 with progress?  We have lots of metrics that we can chase down, right?

migrationConsole/lib/console_link/console_link/loadtest/tui/launch_modal.py
gregschohn, line 1: How should we justify tying load testing (which won't even always be installed) with `workflow manage`?  Is this something that has made your life much better?  Could it be just as well-managed as its own TUI?  (workflow test)?  I'm worried that the TUI is already complex enough & putting a new major feature in optionally is a lot of risk.

migrationConsole/lib/integ_test/integ_test/test_cases/k6_load_test_tests.py
```diff
@@ -0,0 +1,134 @@
+"""Load-test integration case (Shape A): a k6 load test layered on a live CDC migration.
+
+Instead of `generate-data`, this drives real load through the capture proxy with a short k6 run
+(k6-operator TestRun), then asserts the capture-and-replay pipeline moved that traffic to the
+target — i.e. the pipeline works *under load*, not just for a handful of hand-written docs.
+
+Requires the standalone k6LoadTest chart to be installed (operator + scenarios + RBAC). The test
+runner installs it when `--with-load-test` is passed; the case is explicit-selection only so it
```
gregschohn: are you filtering away these tests when --with-load-test isn't run?  I am fine always installing the helm chart for load tests for every integ test.  I am assuming that it wouldn't bump up the wall time on top of what we already have for MA.  I assume that the helm installs would run in parallel.

migrationConsole/lib/integ_test/integ_test/test_cases/k6_load_test_tests.py
```diff
@@ -0,0 +1,134 @@
...
+
+        logger.info("Waiting for capture-proxy to be ready...")
+        wait_for_proxy_ready(ns, timeout_seconds)
+
+        # Advance to the full migration (snapshot + backfill + replay). No pre-snapshot load, so
+        # everything k6 writes next reaches the target via replay only.
+        logger.info("Waiting for workflow to pause before full migration submit...")
+        self.argo_service.wait_for_suspend(workflow_name=self.workflow_name, timeout_seconds=600)
+        logger.info("Resuming workflow to submit full migration...")
+        self.argo_service.resume_workflow(workflow_name=self.workflow_name)
```
gregschohn, line 102: I don't think that these calls will do much any more - do we still have suspension points in an outer workflow?

TrafficCapture/trafficLoadTest/scripts/k6-run.sh
gregschohn, line 1: why are some of these scripts left here and others not?

TrafficCapture/trafficLoadTest/scripts/run_test_chaos.sh
gregschohn, line 1: why are some of these scripts left here and others not?

migrationConsole/lib/integ_test/integ_test/test_cases/k6_load_test_tests.py
```diff
@@ -0,0 +1,134 @@
...
+    wait_for_proxy_ready, wait_for_replayer_consuming, assert_replay_drained,
+)
+
+logger = logging.getLogger(__name__)
+
+# The k6 ingest scenario writes the nyc_taxis schema to this index by default.
+K6_INDEX = "nyc_taxis"
+
+
+class Test0050CdcK6LoadTest(MATestBase):
```
gregschohn: can you wire this into jenkins too as a new test.  It should have its own GHA trigger and should run with the EKS flag for now.  We'll set up some bigger tests that run nightly later, but if we should replace the current CDC tests with something a bit more rigorous.

gregschohn (commented):
I have some blocking comments on some of the other changes - like integrating this optionally with manage and some of the other script changes.

We should also consider leaving the test scripts where they were, outside of helm, and pulling them in at helm installation time so that we don't need to fight helm-isms constantly.

buildImages/backends/dockerHostedBuildkit.sh
awagen: I experienced an issue here (which I did not experience before) when I tried the local kind setup in a fresh ubuntu vm. DNS lookup did not work, leading to hosts not reachable, which seems be due to the dns_line extracted in lines 49/50 pulling an address only reachable from host, thus not reachable from within the buildkit container in some cases.

deployment/k8s/charts/components/k6LoadTest/files/k6/scripts/GENERATOR_logs_data_queries.js
awagen: Sure, I can add that. That was a bit of an artificial type prefix here that was used to distinguish files that would usually live in distinct folders, but due to the configmap flat key structure was flattened into a single folder. To distinguish that type from the actual file was the intention, but that is also valid in lower-casing  :)

TrafficCapture/trafficLoadTest/scenarios/ingest.js
awagen: I had specifically changed the name here cause it was mixing the data scenario (the data schema) with the actual load testing scenario, so the above SCHEMA reflects that it is indeed only a data type setting.

deployment/k8s/charts/components/k6LoadTest/files/grafana/load-test.json
awagen: Yes the pushing of all resources into configmaps isnt particulaly beautiful. I do like the suggestion of using OCI image for file mounts. Ill change that.

migrationConsole/lib/console_link/console_link/loadtest/tui/launch_modal.py
awagen: Good point regarding the separation. Initially workflow manage seemed like a central console were this aspect might fit, but I agree with you that load test management is both a separate aspect from migration workflows and separation makes sense here. 

On top of that, Id say a TUI is not strictly needed for loadtests, but I find it a nice addition to keep overview over running load test scenarios and manage them (start, stop, …). Its strictly convenience though.

TrafficCapture/trafficLoadTest/scripts/run_test_chaos.sh
awagen: This was due to k6 run definition related files having moved to the deployment/k8s load test subfolder, where the original folder only retained the scripts for the test runs. This is changed in upcoming commits, where k6 related files can be found in the trafficLoadTest subfolder, too.

TrafficCapture/trafficLoadTest/scripts/k6-run.sh
awagen: See comment below.

migrationConsole/lib/integ_test/integ_test/test_cases/k6_load_test_tests.py
awagen: This flag wont be needed anymore, as in the update is set depending on whether any test is a load test related test. It is still bound to those tests, as its depending on some public images / charts (k6 / k6-operator), and there could be scenarios (such as private vpc), where the pulling would not be possible.

deployment/k8s/fillLocalRegistry.sh
awagen: Not a general fix, and there are cases where this would not solve the issue, thus removing that one.

migrationConsole/lib/console_link/console_link/workflow/commands/loadtest.py
awagen: Maybe Im missing something, but where are you seeing the periodic logs? Here the function only returns a state that is logged once, the interval relates to the polling.

migrationConsole/lib/console_link/console_link/workflow/commands/k6.py
awagen: Adding those within the comment.

migrationConsole/lib/integ_test/integ_test/test_cases/k6_load_test_tests.py
awagen: We should touch base regarding access to / setup in jenkins to finalize the setup.

deployment/k8s/installK6Chart.sh
```diff
@@ -0,0 +1,129 @@
...
+  --set image.repository="$runner_repo" ${runner_tag:+--set image.tag="$runner_tag"} \
+  --set image.pullPolicy=IfNotPresent \
+  --set scriptsImage.repository="$scripts_repo" \
+  ${scripts_digest:+--set scriptsImage.digest="$scripts_digest"} \
+  ${scripts_tag:+--set scriptsImage.tag="$scripts_tag"} \
+  --set scriptsImage.pullPolicy=Always \
+  --timeout 300s 2>&1 | sed 's/^/  /'
+
+kubectl --context "$CONTEXT" -n "$NAMESPACE" rollout status deploy \
+  -l app.kubernetes.io/name=k6-operator --timeout=180s 2>&1 | tail -1 | sed 's/^/  /' || true
```
gregschohn: Let's not do `|| true` here

deployment/k8s/charts/components/k6LoadTest/templates/k6-workflowtemplates.yaml
```diff
@@ -0,0 +1,126 @@
...
+              (dict "name" "K6_OTEL_GRPC_EXPORTER_INSECURE" "value" "true")) | quote }}
+  templates:
+    - name: run
+      resource:
+        action: create
+        # Ties the TestRun's lifetime to the workflow: stopping a run means deleting the workflow.
+        setOwnerReference: true
+        # A TestRun that never reaches a terminal stage (for example runner pods that cannot pull an
+        # image) would otherwise block forever, since neither condition can match.
+        activeDeadlineSeconds: {{ $.Values.workflow.activeDeadlineSeconds | int }}
```
gregschohn: `activeDeadlineSeconds` is nested at the wrong level.  Argo Workflows  defines it on the containing Template.  `K8sResourceBuilder` explicitly hoists this field out of `resource`. 

Could this be moved to:
```
- name: run
  activeDeadlineSeconds: ...
  resource:
    action: create
```

libraries/testAutomation/testAutomation/test_runner.py
```diff
@@ -494,6 +534,13 @@ def run(self, skip_delete: bool = False, keep_workflows: bool = False,
 
                 self.k8s_service.wait_for_all_healthy_pods()
 
+                # Install the standalone k6 load-test chart so load-test cases (Test008*) can submit
+                # TestRuns. Kept separate from the migration chart and installed only for runs that
+                # ask for it, so a normal run has no operator/scenarios/RBAC, cannot trigger a load
+                # test, and does not pay for the k6-operator install.
+                if self._load_test_requested():
+                    self._install_load_test_chart()
```
gregschohn, line 584: We need tie into the uninstall above to uninstall the k6 chart when it was enabled...
```
# 3. helm uninstall (10m timeout set in helm_uninstall).
```

migrationConsole/lib/console_link/console_link/workflow/commands/loadtest.py
gregschohn: That's my question - For an hour-long run, what should a developer expect from the runner logs: periodic progress, failed checks, or request/response details? Which dashboard signals establish that the run is healthy, and can metrics distinguish concurrent runs of the same scenario?

buildImages/build.gradle
```diff
@@ -124,6 +124,17 @@ def buildKitProjects = [
                         ":TrafficCapture:dockerSolution:syncArtifact_migration_console_staging_noDockerBuild",
                         "buildKit_migrationConsoleBase" // All buildKit projects are configured at the root
                 ]
+        ],
+        [
```
gregschohn: Should `k6Scripts` be part of `buildKitProjects`? That group feeds `buildImagesToRegistry`, so this image will be built and pushed during normal production-image builds, even with `-PskipTestImages=true`. Since the README identifies this as a load-test image rather than a released artifact, I think it should be segregated similarly to the ES/Solr test images, or preferably placed behind a dedicated load-test image task. The EKS k6 pipeline can then invoke that task explicitly instead of making every production build include this image.

deployment/k8s/charts/components/k6LoadTest/values.yaml
```diff
@@ -0,0 +1,111 @@
...
+# node anti-affinity — it forces the runner pods onto distinct nodes. Only meaningful when
+# parallelism > 1, and it needs at least `parallelism` schedulable nodes or the surplus pods stay
+# Pending forever — so keep it false on single-node clusters (e.g. local minikube).
+#
+# There is no `cleanup` knob: the workflow owns its TestRun, so deleting the workflow takes the CR
+# and its pods with it. The operator's own `cleanup: post` is deliberately not used — it deletes the
+# CR the instant it finishes, which would race the workflow's own wait on that same status.
+testRun:
+  parallelism: 1
+  separate: false
```
gregschohn: Let's rename this to have more intent about anti-affinity

buildImages/backends/dockerHostedBuildkit.sh
gregschohn, line 1: Can we put this into a separate PR to give it more focused discussion and scrutiny.  I'm still not sure why we need it and that it won't break something else.

deployment/k8s/charts/components/k6LoadTest/templates/k6-workflowtemplates.yaml
```diff
@@ -0,0 +1,126 @@
...
+spec:
+  serviceAccountName: {{ $.Values.workflow.serviceAccountName }}
+  entrypoint: run
+  # Every per-run knob is a parameter with a default supplied here, so a caller overrides only what
+  # it cares about and never has to know the images. `runnerEnv` carries the WHOLE runner env as a
+  # one-line JSON array (JSON is valid YAML, so it splices into the manifest below): the console
+  # reads this default and replaces entries by name, which keeps one definition of the static vars
+  # and avoids two env entries of the same name.
+  arguments:
+    parameters:
```
gregschohn, line 89: How is this WorkflowTemplate associated with the migration or CaptureProxy that it should exercise? It appears that it is not: the template has no direct target parameter, and the effective destination comes from CAPTURE_PROXY_URL=https://capture-proxy:9201 hardcoded in every preset. 
`--target` can override it indirectly through the generic runnerEnv JSON...  I don't want to create a whole bunch of shadow parameters and then have to copy them in - what do you think the best way to maintain this is?  If the runnerEnv had the default mapping in place, that would have been more immediately obvious to me that we can control the target that way.

We'll need this once we start testing cloud throughput since we'll want to go through a LB endpoint rather than just the k8s endpoint.

deployment/k8s/installK6Chart.sh
gregschohn: don't do `|| true`

deployment/k8s/deployCdcLoadTestConfig.sh
```diff
@@ -0,0 +1,291 @@
...
+    [[ -z "$name" ]] && continue
+    endpoint="$(proxy_endpoint "$name")"
+    [[ -n "$endpoint" ]] && printf '  proxy %s endpoint: %s\n' "$name" "$endpoint"
+  done < <(K get captureproxies.migrations.opensearch.org -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}' 2>/dev/null || true)
+}
+
+cmd_down() {
+  say "Uninstall k6 load-test chart (operator + example TestRuns + RBAC)"
+  if command -v helm >/dev/null; then
+    helm --kube-context "$CONTEXT" uninstall k6-load-test -n "$NAMESPACE" 2>&1 | sed 's/^/  /' || true
```
gregschohn: Another spot w/ || true - please remove all of these or be very careful and make sure we're handling the silent failures appropriately

deployment/k8s/deployCdcLoadTestConfig.sh
gregschohn, line 1: This entrypoint structure seems different from our other deployment scripts: installation is encapsulated by installK6Chart.sh, but teardown reaches through that abstraction and runs Helm directly. It already means K6_RELEASE/--release can affect installation while down still uninstalls the hardcoded k6-load-test release. 

Could the chart script own both install and uninstall operations, with install remaining the default for compatibility, and have deployCdcLoadTestConfig.sh delegate both lifecycle operations to it?

migrationConsole/lib/console_link/console_link/loadtest/tui/app.py
gregschohn, line 1: Do you think that this has provided value?  If so, cool, if not, I don't know if it matters to try to keep it.

migrationConsole/lib/console_link/console_link/workflow/commands/loadtest.py
gregschohn, line 1: Thanks for separating this from `workflow manage`. I think it would be valuable to take that separation one step further and expose load-test management as its own CLI rather than registering it in the base `workflow` command.

The current chart-presence check hides and guards the command at runtime, but `workflow` still imports the implementation unconditionally and can probe Kubernetes while rendering its help. It also makes the base console aware of k6-specific templates, presets, labels, and lifecycle behavior.

Since the implementation is already largely self-contained, creating a separate `k6-loadtest` entrypoint appears fairly small: remove the import and registration here, add a dedicated `main()` and console-script entrypoint, and update its callers and tests. That would substantially reduce the chance of this optional test tooling destabilizing migration workflow commands. Ideally, it could eventually become an optional package as well, which would isolate its dependencies completely.

deployment/k8s/installK6Chart.sh
```diff
@@ -0,0 +1,129 @@
...
+  else
+    SCRIPTS_IMAGE="${REGISTRY_PREFIX}migrations/k6_scripts:latest"
+  fi
+fi
+split_ref "$SCRIPTS_IMAGE"
+scripts_repo="$ref_repo" scripts_tag="$ref_tag" scripts_digest="$ref_digest"
+
+# ── Install ────────────────────────────────────────────────────────────────────
+helm repo add grafana https://grafana.github.io/helm-charts >/dev/null 2>&1 || true
+# Vendor the k6-operator subchart: offline from Chart.lock if already vendored, else fetch it.
```
gregschohn: An agent came back with this...

"The chart has a remote k6-operator dependency, but this PR includes neither `Chart.lock` nor a vendored chart archive. That means installation resolves/downloads dependencies at runtime and cannot currently work offline as the comment suggests. We should commit the lock and vendored dependency, or explicitly document that installation requires network access?"

I'm not sure how to get helm to do network-free installs.  If you know, I'm happy for you to do that here & we can play with it.  If the current setup doesn't allow for it, you should remove the comment.

.claude/settings.local.json
```diff
@@ -11,7 +11,13 @@
       "Bash(cat:*)",
       "Bash(kubectl get:*)",
       "Bash(argo resume:*)",
-      "Bash(xargs ls)"
+      "Bash(xargs ls)",
```
gregschohn, line 14: were these required or intentional?  I'm honestly fine if they were...

.claude/settings.local.json
```diff
@@ -11,7 +11,13 @@
...
+      "Bash(helm version *)",
+      "Bash(helm template *)",
+      "Bash(uvx flake8 *)",
+      "Bash(grep -rni 'press .*`n`\\\\|\\\\bn\\\\b.*k6 panel\\\\|k6.*shortcut.*n' TrafficCapture/trafficLoadTest migrationConsole/lib/console_link/console_link/workflow)",
```
gregschohn: These ones seem super-specific though and maybe less valuable generally

deployment/k8s/charts/components/k6LoadTest/templates/k6-workflowtemplates.yaml
awagen: Thanks for pointing that out. Correcting.

deployment/k8s/installK6Chart.sh
awagen: That was referring to the situation where the k6-operator chart was already downloaded and available, such as if someone had checked it in (which we dont). So yeah, will remove the comment.

migrationConsole/lib/console_link/console_link/workflow/commands/loadtest.py
awagen: Good point. Id say the main system overview id keep in actual monitoring, e.g grafana based on otel metrics collected from all components. For the runner logs Id think a combination of workflow state in combination with e.g thresholds configured on the scenarios should do, which should lead to termination of the run itself and thus status change here. Im not sure its a good idea baking in further metrics queries here.

migrationConsole/lib/console_link/console_link/loadtest/tui/app.py
awagen: It is certainly an optional convenience. Given that some control and monitoring for the workflow handling is happening in a TUI, I find it convenient that there is an k6-specific overview and possibility to start runs and stop runs and have an overview over whats running in a TUI. Strictly needed, no. Id suggest to keep it right now, and as we use the functionalities more either appreciate the functionality or decide to remove it.

deployment/k8s/charts/components/k6LoadTest/templates/k6-workflowtemplates.yaml
awagen: Its true that the available settings here were implicit, e.g one would need to check the scenarios to get an overview. Your suggestion of extending the default parameters seems good to me for now, at least for general parameters that are general, and not scenario-specific. For scenario specific settings either we rely / document properly to check the scenario, or we add a template per scenario, containing all defaults per scenario, at the cost of some redundancy of the boilerplate around it.

buildImages/backends/dockerHostedBuildkit.sh
awagen: As per my description what I observed on linux environments, users are likely to experience issues there. Im good with separating this though, and file it via a separate PR, as it addresses a more general concern.

.claude/settings.local.json
awagen: I agree on the take on specifics, some of the entries here were not intentional. Question is whether we need this file in the first place or rely on users setting theirs (in case they are actually in non-yolo mode). Either way, for now Id just clean this up to cover general save calls rather than too specific ones.

deployment/k8s/charts/components/k6LoadTest/templates/k6-workflowtemplates.yaml
awagen: I opted to break this down more and provide defaults on per (scenario, profile) combination. This leads to a separate workflow template per pair, with values overview within values.yaml.  The target url is also set in values.yaml with a default, and can be overwritten per install and run. Also made the target parameter more explicit as workflow template param.

deployment/k8s/charts/components/k6LoadTest/values.yaml
awagen: Here I was just keeping the naming of the official TestRun parameter. Im not sure it pays to rename it from the original. Im open to wishes here though. You still prefer something more speaking like “useSeparatePodPerTestInstance”?

deployment/k8s/deployCdcLoadTestConfig.sh
awagen: Good catch, unifying both ways, install and uninstall, in the install chart with defaulting to install if uninstall not passed.

migrationConsole/lib/integ_test/integ_test/test_cases/k6_load_test_tests.py
awagen: Apparently cdc2E2eMigrationWithClusters.yaml has a relevant one, namely pause-for-pre-snapshot-data.

buildImages/build.gradle
awagen: It is a dedicated task now.

deployment/k8s/charts/components/k6LoadTest/values.yaml
awagen: Its renamed to useSeparatePodPerTestInstance now and later used as input to the original separate setting on k6.

gregschohn (approved):
Thanks for seeing this all the way through! 
I'm excited to do some load testing on the replayer now!

PR #3267, opened 2026-07-29
Track the live non-nested document count per shard in Document Backfill

AndreKurait:
### Description

Adds an accurate per-shard count of live, non-nested documents to Document Backfill, and fixes a resume-path defect found while building it.

The count is always computed and logged. `--shard-doc-count-s3-bucket` additionally writes it to S3 as gzipped NDJSON for programmatic consumption.

#### Why the number is not simply "documents we wrote"

A cluster exposes three different counts. Measured on a real OpenSearch 2.19 shard with 5 root documents and 9 nested children:

| surface | reports |
|---|---|
| `<index>/_count` | **5** — root documents only |
| `_cat/indices docs.count` | 14 — roots **and** nested children |
| Lucene `numDocs()` | 14 — likewise |

Nested children are separate Lucene documents carrying no stored `_id`, so `LuceneReader.getDocument` already skips them and they never reach the target. The count therefore has to follow `_count`. Verified identical on OpenSearch 2.19 (Lucene 9), ES 7.10 (Lucene 8) and ES 5.6 (Lucene 6).

#### Resume regression fixed here

The pre-pipeline reindexer checkpointed the Lucene doc number of the last document in each batch:

```java
var lastDoc = docsBatch.get(docsBatch.size() - 1);
...
new WorkItemCursor(lastDoc.progressCheckpointNum)
```

The pipeline rewrite replaced that with a running sum of documents written to the sink (`cumulativeOffset[0] += result.docsInBatch()`), introduced in f4186b147 "Add work coordination, transformation, and progress tracking to pipeline". Both paths coexisted for a while — the old reindexer kept using the Lucene doc number — so the change only took effect once the pipeline became the sole path.

That substituted a document count for a seek position. The two agree only when every source position yields an emitted document, which is not true of indices with nested documents: nested children occupy Lucene doc ids but carry no stored `_id`, so `LuceneReader.getDocument` returns null and nothing is emitted. The count then lags the true Lucene position and a resumed worker seeks backwards into a range it had already finished. On a shard of 3 roots interleaved with 4 children, a completed pass left a checkpoint of `3` while the last processed Lucene doc id was `9`, so a successor re-migrated the whole shard.

Re-migration is idempotent, so no documents were lost; the cost is wasted work plus the impossibility of deriving a trustworthy total from the checkpoint.

Every consumer already treats the checkpoint as a Lucene doc id (`readDocsFromSegment` computes `position - segmentDocBase`, `getSegmentsFromStartingSegment` binary-searches it against segment doc bases, `SolrBackupSource` does the same arithmetic, and `SnapshotExtractor.readDocuments` documents the parameter as "the Lucene doc index to resume from"). This restores the producer to match them, so Solr and delta paths need no change.

`ProgressCursor` / `WorkItemCursor` now carry the seek position and the emitted-document total as separate fields, so the accounting the pipeline rewrite wanted is still available without overloading the checkpoint.

#### Surviving lease handoff

A shard can outlive several leases, so the total lives in the work-coordination document as an optional `docsEmitted` field, stamped onto successors and read back on acquisition.

- **Absolute, never incremented.** Successor creation is not transactional and is retried — the bulk create tolerates 409s precisely because a predecessor may have died partway through — so an increment would inflate the total on retry.
- **The checkpoint batch is excluded.** Successors deliberately restart *at* the checkpoint (how 1:many document splits are handled), so they re-emit that batch. Without subtracting it the count grows on every lease expiry.
- **The crash-recovery path forwards it.** `acquireNextWorkItem` finishes successor creation for a worker that died mid-handoff; dropping the inherited total there would silently reset the shard's count to zero.

The field is optional on read (absent ⇒ 0), so work items created before this change keep working, and the painless lease script is untouched — no `scriptVersion` bump, no coordinated rollout.

#### Configuration

Records write to the S3 bucket, region and endpoint the config processor already resolves for the failed document stream (user value, else the snapshot repo's, else the deployment default), rather than introducing a second bucket to configure. The only new setting is `--shard-doc-count-s3-prefix`, added to `userSchemas` so it is settable through `documentBackfillConfig`; args reach the container as inline JSON so no per-flag plumbing was needed.

#### Performance

Counting rides the pass the backfill already performs, on a branch that already existed, so it adds no I/O.

An upfront counting pass was measured and rejected. On a 60k-document segment a stored-fields count-only pass costs **182 ms vs 235 ms** for the full read already being done — **77% overhead**, because block decompression dominates and restricting the field selector saves only 23%. That extrapolates to ~30 s at 10M docs/shard and ~5 min at 100M, all while holding a lease and producing no migration progress, and re-paid by every successor.

A `_primary_term` doc-values shortcut is ~400x cheaper (<1 ms) but was also rejected: it does not exist on Lucene 6, where `_version` doc-values and `_uid` postings both count children and would silently report 14 instead of 5.

### Issues Resolved

None filed; found and fixed while adding the count.

### Testing

Unit:
- `ProgressCursorPositionTest` — sparse (nested) positions, batching, mid-shard resume, count carried across a simulated lease expiry, positionless-source fallback, and the dense case that previously masked the defect.
- `DocsEmittedForSuccessorTest` — handoff arithmetic, a three-generation replay simulation, and the guard against a negative carried value.
- `S3ShardDocCountSinkTest` — record schema, gzip NDJSON layout, S3 key layout, flush-failure propagation, post-close rejection, idempotent close.

Integration (real containers, `isolatedTest`):
- `NestedDocCountTest` — ES 7.10 and OpenSearch 2.19 sources. Asserts the source really exhibits the `_count` vs `docs.count` divergence *before* asserting the migration matches `_count`, so it cannot pass vacuously against a fixture whose nested mapping silently produced no children.
- `LeaseExpirationTest` re-run to confirm the multi-generation resume path is unchanged.

Consumer note: records are at-least-once — the write is flushed before the work item is marked complete, and a failed flush aborts the shard so a successor re-emits. Prefer the highest `liveDocCount` per `(indexName, shardNumber)`, or the record with `shardComplete=true`.

### Check List


AndreKurait:
Opened #3268 as a deliberately minimal alternative to this PR, so the tradeoff can be compared directly rather than argued in the abstract. Only one of the two should be merged.

#3268 logs the **live Lucene** document count (what `_cat/indices docs.count` reports, nested children included) by summing `liveDocs.cardinality()` — a popcount per segment bitset, measured at **0.047 ms** on a 60k-document shard. 2 files, +217, no new CLI flags, no change to resume semantics.

This PR produces the **root-document** count instead (what `_count` reports, nested children excluded) and persists it across lease handoffs with optional S3 output. That number is strictly more expensive to obtain: identifying root documents requires a stored-fields walk, measured at **175 ms** on the same shard — ~3,700x the cardinality popcount — which is why it is folded into the existing read pass rather than computed separately.

Measured on a real OpenSearch 2.19 shard, cross-checked against the cluster (`docs.count`=59,988, `_count`=20,000):

| strategy | result | time |
|---|---|---|
| `liveDocs.cardinality()` (#3268) | 59,988 | 0.047 ms |
| stored-fields root walk (this PR) | 20,000 | 175 ms |

Happy to go either way — #3268 if a Lucene-level count in logs is enough, this one if the `_count`-comparable number or machine-readable output is needed.

AndreKurait:
Added a black-box regression guard and verified it in both directions, since unit coverage of the cursor could not have caught the original regression.

`LeaseExpirationTest.successorCheckpointIsALucenePositionNotADocumentCount` migrates a real nested shard (1,640 roots x 4 children = 8,200 Lucene docs) through a throttled target so the lease expires partway, then reads the successor checkpoint back out of the work-coordination index.

Same fixture, same 44 documents migrated:

| | successor checkpoint |
|---|---|
| `upstream/main` | **43** |
| with this PR | **214** |

43 is the document count. 214 is the Lucene position — 44 roots x (1 + 4 children) — which is where a successor must actually resume. On upstream the successor seeks to 43 and re-reads the whole shard.

Verified by cherry-picking the test alone onto `upstream/main`, where it fails with the intended assertion rather than incidentally:

```
org.opentest4j.AssertionFailedError: successor checkpoint (43) must exceed
documents written (44) on a nested shard; otherwise the checkpoint is a
document count and a successor will seek backwards and re-migrate finished
documents ==> expected: <true> but was: <false>
```

and passes on this branch.

One note for reviewers: the test reuses this class's existing container/toxiproxy/dedicated-coordinator harness, and `SourceTestBase.setupProcess` pins `--index-allowlist geonames`, so the nested index is named `geonames` — otherwise no work item is created and the test passes vacuously. I hit exactly that and it is called out in a comment.

Also addressed in the latest push: dropped the redundant `--shard-doc-count-s3-bucket`/`-region`/`-endpoint` args in favour of the destination the config processor already resolves, leaving only `--shard-doc-count-s3-prefix` (now wired through `userSchemas`), and corrected the commit message and description to identify this as a regression from f4186b147 rather than an inherent flaw.

AndreKurait:
Found and fixed a second instance of the same defect while auditing the fallback path: **the Solr adapter also lost its resume position.**

`SolrBackupSource.readDocumentsFromReader` seeks by Lucene doc id — it passes the checkpoint straight to `getSegmentsFromStartingSegment` and `readLiveDocsFromSegment` — but `toDocument` built its `Document` with an empty `sourceMetadata`, dropping `luceneDocNumber`. With no position on the document the pipeline falls back to counting emitted documents, that count is written as the work item's `startingDocId`, and it is read back as a Lucene doc id. Same count-as-position mismatch, reached via the fallback rather than directly.

Notably the synthesized-id path already used the doc number (`"solr_doc_" + change.getLuceneDocNumber()`), so the value was in hand and simply not propagated.

Two cases added to `SolrBackupSourceTest`. Verified as real regression guards by reverting the one-line fix: both **fail** without it and **pass** with it.

Worth flagging for review because it means the original checkpoint regression had a wider blast radius than the snapshot path alone — any Solr migration whose shard outlived its lease would resume at the wrong offset.

jzonthemtn:
I kicked off the failed checks again. I think this PR was probably made earlier when Jenkins was down for a bit.

The impacts with the pluggable RFS sources is small and manageable. Should definitely merge this first though and then I will update the design/PR on #3269.

jzonthemtn (approved):

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -882,9 +907,47 @@ private static CompletionStatus prepareAndMigrate(
...
+        return (index, shard, total, luceneTotal, thisGen, priorGen) -> {
+            var countRecord = ShardDocCountRecord.builder()
+                .workerId(workerId)
+                .indexName(index)
+                .shardNumber(shard)
+                .liveDocCount(total)
+                .liveLuceneDocCount(luceneTotal)
+                .docsThisGeneration(thisGen)
+                .docsPriorGenerations(priorGen)
+                .shardComplete(true)
```
jzonthemtn, line 946: `shardComplete` is `true` at both construction sites. `false` is reachable only by an omitted builder call. Re-emitted duplicates also carry `true`, so the flag can't discriminate. Drop it, or `@Builder.Default` it to `true`?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -882,9 +907,47 @@ private static CompletionStatus prepareAndMigrate(
...
+     * successor re-emits rather than the record being lost.
+     */
+    static DocumentMigrationBootstrap.ShardDocCountReporter shardDocCountReporter(ShardDocCountSink sink) {
+        if (sink == null) {
+            return (index, shard, total, luceneTotal, thisGen, priorGen) -> { };
+        }
+        var workerId = ProcessHelpers.getNodeInstanceName();
+        return (index, shard, total, luceneTotal, thisGen, priorGen) -> {
+            var countRecord = ShardDocCountRecord.builder()
+                .workerId(workerId)
```
jzonthemtn, line 939: Does `sessionId` need set here, too?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/DocumentMigrationBootstrap.java
```diff
@@ -217,21 +232,35 @@ CompletionStatus runPartitionMigration(
...
                 .addArgument(durationMs)
                 .log();
-
-            var error = migrationError.get();
-            if (error != null) {
-                context.recordPipelineError();
-                throw new RfsException("Partition migration failed for " + wi, error);
-            }
+            shardDocCountReporter.report(wi.getIndexName(), wi.getShardNumber(), shardMigratedDocs,
+                Math.max(0L, shardLiveLuceneDocs), totalDocsMigrated.get(), docsAlreadyEmitted);
```
jzonthemtn, line 263: -1 in the log via line 247/248, but Math.max(0L, ...) in the record, so a mismatch between the two. Not sure what the impact of that might be.

jzonthemtn:
I think the Jenkins failures are due to problems in the changes.

jzonthemtn (changes requested):

PR #3268, opened 2026-07-29
Log per-shard document counts, and fix the resume checkpoint

AndreKurait:
### Description

Logs the live Lucene document count for each shard the backfill reads — the number `_cat/indices docs.count` reports.

Backfill already logged `maxDoc`, which includes deleted documents, so no logged value could be compared against the source cluster. This adds the live count beside it.

Nested child documents are included, because each is its own Lucene document. That is the same reason `docs.count` exceeds `_count` on an index containing nested documents.

**This is a deliberately minimal alternative to #3267.** Both PRs contain the same resume-checkpoint fix and both report the same two numbers; they differ only in how those numbers are surfaced. Only one should be merged.

| | this PR | #3267 |
|---|---|---|
| resume checkpoint fix | yes | yes |
| reports `_count`-comparable root count | yes | yes |
| reports `docs.count`-comparable Lucene count | yes | yes |
| output | one log line per shard | log line + optional S3 NDJSON |
| survives lease handoff | no — per-generation subtotal | yes, running total in the work-coordination doc |
| new CLI flags | none | `--shard-doc-count-s3-*` (wired through `orchestrationSpecs`) |
| diff | 3 files, ~+280 | 17 files, ~+1500 |

The substantive difference is durability. This PR logs what each lease generation saw, so on a shard that outlives its lease you get several partial lines and have to sum them yourself. #3267 keeps one authoritative running total per shard in the coordination document, and can write it to S3 for programmatic consumption — at the cost of a coordination-document field, new CLI surface, and a larger review.

Choose this one if per-shard log lines are enough. Choose #3267 if you need one authoritative number per shard or machine-readable output.

#### Why the Lucene count is the cheap one

Measured on a real OpenSearch 2.19 shard (60,569 `maxDoc` / 59,988 live / 20,000 roots / 9 nested children per root on average):

| strategy | result | time |
|---|---|---|
| `liveDocs.cardinality()` summed over segments | 59,988 | **0.047 ms** |
| `reader.numDocs()` | 59,988 | 0.001 ms |
| stored-fields walk to identify root documents | 20,000 | 175 ms |

`cardinality()` is a popcount over each segment's bitset — no stored-field or doc-values access — so it is ~3,700x cheaper than the walk required to exclude nested children. Verified against the live cluster: `_cat/indices docs.count` = 59,988 and `_count` = 20,000.

Because it is effectively free it runs unconditionally rather than behind a flag. The root count is likewise free: it is the number of documents the migration already wrote.

The emitted line is:

```
Shard doc count: index=..., shard=0, migratedDocs=5, liveLuceneDocs=14
```

### Issues Resolved

None filed.

#### Resume checkpoint fix

Included here as well, since counting per shard is only meaningful if resume is correct. The pre-pipeline reindexer checkpointed the Lucene doc number of the last document in a batch; the pipeline rewrite (f4186b147) replaced that with a running sum of documents written, which diverges on nested indices and makes a successor seek backwards and re-migrate finished documents. See #3267 for the full analysis — the commit is identical in both PRs.

### Testing

`ProgressCursorPositionTest` covers the checkpoint fix (sparse nested positions, mid-shard resume, dense case that masked the regression).

`ShardLiveDocCountTest` (`isolatedTest`, ES 7.10 and OpenSearch 2.19 sources) asserts the count equals the source's `_cat/indices docs.count` for a nested index that also carries deleted documents, exercising a real cluster, a real snapshot, and the real unpack + reader path including the soft-delete wrapper — not a hand-built Lucene index.

It also asserts the count strictly exceeds `_count`, so the test fails if nested children are ever silently excluded, and asserts up front that the fixture actually produced nested children so it cannot pass vacuously.

Both parameterizations pass locally.

### Check List


AndreKurait:
Restructured so this is a like-for-like comparison against #3267 rather than a different feature.

Both PRs now contain the same resume-checkpoint fix and both report the same two per-shard numbers. The only difference is durability:

- **this PR** logs one line per shard per lease generation: `Shard doc count: index=..., shard=0, migratedDocs=5, liveLuceneDocs=14`. On a shard that outlives its lease you get several partial lines and sum them yourself. 3 files.
- **#3267** keeps one authoritative running total per shard in the work-coordination document (surviving lease handoff) and can write it to S3 as NDJSON. Costs a coordination-document field, `--shard-doc-count-s3-prefix` wired through `orchestrationSpecs`, and a much larger review. 17 files.

`migratedDocs` matches the source's `<index>/_count` (nested children carry no stored `_id` and never reach the target). `liveLuceneDocs` counts them, matching `_cat/indices docs.count`. Verified on real clusters: a shard with 5 roots and 9 nested children reports 5 and 14 against a source reporting `_count`=5 and `docs.count`=14.

The Lucene number is essentially free — summing `liveDocs.cardinality()` over segments while the reader is already open is a popcount per bitset, measured at **0.047 ms** on a 60k-document shard, versus **175 ms** for the stored-fields walk that would be needed to derive the root count independently. The root count is free too, since it is what the migration already wrote.

Still recommend picking one, not both.

AndreKurait:
Simplified the logging per review: the separate `Shard doc count` line is gone and the count now rides the existing per-partition stats line, with the scope of each figure named in the key.

```
Partition migration stats: index=..., shard=0, docsThisGeneration=5,
  bytesThisGeneration=369, batchesThisGeneration=1, durationMs=204,
  shardLiveLuceneDocs=14
```

Two reasons for dropping the migrated-doc count from the shard-scoped reporting here rather than presenting it beside the Lucene count:

1. `totalDocsMigrated` is a fresh counter per `migrateOneShard` call, and each lease generation is a separate process — so it is strictly per-generation. This PR has no coordination-document accumulation, so there is nothing to make it whole-shard.
2. Presenting a per-generation number next to a whole-shard one on a line called "shard doc count" invites summing them, and summing `docsThisGeneration` across generations over-counts: successors deliberately restart *at* the checkpoint and re-process that batch.

So `shardLiveLuceneDocs` is the only whole-shard figure, everything else says `ThisGeneration`, and `-1` is logged when the source cannot report the count cheaply rather than a `0` that would read as an empty shard.

Also squashed to two commits (checkpoint fix, then the count) and synced the shared cursor-fix commit with #3267 so the two PRs carry an identical version of it.

Re-verified on real clusters after the change: both ES 7.10 and OpenSearch 2.19 sources pass, with the Lucene count matching `_cat/indices docs.count` on a nested index carrying deleted documents.

AndreKurait:
Found and fixed a second instance of the same defect while auditing the fallback path: **the Solr adapter also lost its resume position.**

`SolrBackupSource.readDocumentsFromReader` seeks by Lucene doc id — it passes the checkpoint straight to `getSegmentsFromStartingSegment` and `readLiveDocsFromSegment` — but `toDocument` built its `Document` with an empty `sourceMetadata`, dropping `luceneDocNumber`. With no position on the document the pipeline falls back to counting emitted documents, that count is written as the work item's `startingDocId`, and it is read back as a Lucene doc id. Same count-as-position mismatch, reached via the fallback rather than directly.

Notably the synthesized-id path already used the doc number (`"solr_doc_" + change.getLuceneDocNumber()`), so the value was in hand and simply not propagated.

Two cases added to `SolrBackupSourceTest`. Verified as real regression guards by reverting the one-line fix: both **fail** without it and **pass** with it.

Worth flagging for review because it means the original checkpoint regression had a wider blast radius than the snapshot path alone — any Solr migration whose shard outlived its lease would resume at the wrong offset.

jzonthemtn:
Compared to #3267, this PR has the least impact on #3269.

Both PRs give the same shard document counts and the same resume-checkpoint fix, so this one answers "did the migration move the right number of documents?" just as well. The shard count is read off the shard, not accumulated, so lease handoff can't skew it either way. #3267 adds a durable emitted-doc total and an S3 NDJSON sink, which matter if shards outlive their leases or we plan to build something on top of the counts. Nothing is right now, so I'd hold off on that until it's needed.

Issue #3065, opened 2026-06-04
Add dlq redrive command to resubmit failed documents to OpenSearch


jzonthemtn:
When documents fail during RFS backfill they are written to the S3 DLQ. Currently operators can inspect failures with dlq list and dlq count, but there is no way to resubmit those documents to OpenSearch without re-running the full migration. A `redrive` command would let operators recover from transient failures or re-attempt documents after fixing a root cause (e.g. an index mapping issue).

**Background**

Each DLQ record contains a `requestItem` field that is the complete bulk operation spec (action metadata + source document) as originally sent to OpenSearch. This can be replayed directly against the `/{index}/_bulk` endpoint without any
transformation. Records also carry failureClass (NON_RETRYABLE or RETRYABLE_EXHAUSTED) and targetIndex, enabling targeted filtering before resubmission.

**Proposed command**

`console dlq redrive [OPTIONS]`

Options:

  - `--failure-class` - Filter to NON_RETRYABLE, RETRYABLE_EXHAUSTED, or both (default: both)
  - `--index` - Limit redrive to a single target index
  - `--limit` - Maximum number of documents to redrive
  - -`-dry-run` - Print what would be redriven without submitting
  - `--session` - Override the session ID (default: current session from ConfigMap)

**Example usage**

```
# Preview what would be redriven
console dlq redrive --dry-run

# Redrive only exhausted-retry documents
console dlq redrive --failure-class RETRYABLE_EXHAUSTED

# Redrive failures in one index after fixing its mapping
console dlq redrive --index orders-2024
```

**Proposal**

1. Calls `dlq.list_records(cfg) `to stream deduplicated records from S3 (gzipped NDJSON), applying any `--index` or `--failure-class filters`.
2. Reconstructs bulk NDJSON from each record's requestItem field and batches into bulk request bodies.
3. `POST`s each batch to `/{index}/_bulk` on the target OpenSearch cluster.
4. Parses bulk responses: documents that succeed are reported as recovered. Documents that fail again are reported as still-failed (they remain in the original DLQ and are not deleted).
5. Prints a summary of recovered, still-failed, and skipped document counts.

Note: NON_RETRYABLE documents (e.g. `mapper_parsing_exception`) are unlikely to succeed unless the index mapping has been changed beforehand.

**Acceptance criteria**

- [ ] `console dlq redrive `is available as a subcommand of the existing dlq group in cli.py
[...]

PR #3269, opened 2026-07-30
Pluggable document Sources for RFS

jzonthemtn:
### Description
Implements Pluggable Document Sources for RFS as described in https://github.com/opensearch-project/opensearch-migrations/issues/3065#issuecomment-5118437294.

For context:
* https://github.com/opensearch-project/opensearch-migrations/issues/3065#issuecomment-5118437294

### Issues Resolved
None

### Testing
TBD

### Check List


docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,355 @@
+# Pluggable Document Sources for RFS
```
jzonthemtn, line 1: This file will not be committed and is here for easier commenting and discussion through the design process.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+whose scope is narrower.
+
+Related: [#3065](https://github.com/opensearch-project/opensearch-migrations/issues/3065) (redrive),
+[#3063](https://github.com/opensearch-project/opensearch-migrations/issues/3063),
+[#3064](https://github.com/opensearch-project/opensearch-migrations/issues/3064),
+[#3066](https://github.com/opensearch-project/opensearch-migrations/issues/3066).
+
+## The idea
+
+RFS is already a general-purpose document pump: a source-agnostic pipeline that reads documents, transforms them, and writes them to a target with work coordination, leases, retries, and resumable progress. What it lacks is a way to *plug in* a new source. The reading abstraction exists and is clean; the wiring that decides which source to build is a hard-coded branch on source type, and each branch reaches directly into CLI arguments.
```
malliaridis: For how I understand the proposal, we want to make it easier to extend implementations of `DocumentSource`, without impacting the entire stack (e.g. new CLI arguments for every source, additional fields in the workflow configurations, etc.).

That makes sense.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+    registry picks one                                                      |
+    by spec.kind                                                       failures
+           ^                                                                |
+           |             +---------------------+                            |
+           +-------------|  S3 failure stream  |<---------------------------+
+                         +---------------------+
+             redrive: yesterday's failures are just another source
+```
+
+## The interfaces
```
malliaridis: What is basically proposed in this section is the introduction of clean interfaces, specifically for `DocumentSource`. I agree with the interfaces, and it is a nice and clean abstraction layer, but that does not resolve or address the main issue.

Using interfaces makes implementation pluggable and reusable, so definitely a good "optimization" before addressing the main issue.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+    default void validate(DocumentSourceSpec spec, SourceRuntime runtime) { }
+
+    /** True if construction is expensive enough to skip when no work remains. */
+    default boolean deferUntilWorkAvailable() { return false; }
+
+    DocumentSource create(DocumentSourceSpec spec, SourceRuntime runtime) throws IOException;
+}
+```
+
+Providers are discovered with `ServiceLoader`, so each reader module ships its own
```
malliaridis: Discovering providers via `ServiceLoader` makes it extendable, allowing users (or us) to add new `DocumentSource` implementations with ease.

However, if we plan to take full advantage of the "pluggability" the `ServiceLoader` provides, and with that I mean shipping JARs with various implementations outside the main migration assistant distro, getting these JAR files into our pods to be discoverable will be a big downside. So as long as the services are shipped in the same distribution as the migration assistant, I don't see a problem with that.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+- **Follow-up, any time after Phase 1** — remove the uncoordinated path, leaving `migratePartition` as
+  the pipeline's only entry point: `readCollectionMetadata`, `createCollection`,
+  `migrateAll`/`migrateCollection`, and `partitionConcurrency` (dead once `migrateCollection` goes).
+  Three end-to-end tests move to `migratePartition` with explicit index creation, and the duplicate Solr
+  schema-to-mappings conversion retires — it survives via `SolrBackupIndexMetadataFactory`. Leave
+  `CollectionMetadata` and `EsMetadataMigrationPipeline.migrateAll()` alone; both are still used.
+
+Phase 4 must also correct #3065, whose premise predates two changes: the stored request is the
+pre-transform document, so it is not replayable as sent, and there is no `dlq` command group — only
+`failed-document-stream`.
```
malliaridis: These instructions / phases are clear, helping me understand the actions that should be taken way better.

All phases are understandable, and they add value to the main goal.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+
+This is the sharpest compatibility hazard in the refactor. `Partition` currently extends `Comparable`
+with a default `compareTo` that compares `name()` lexicographically. `EsShardPartition.name()` is
+`snapshot/index/shardNumber`, so that default orders an 11-shard index as `0, 1, 10, 11, 2, ...`
+while the source actually returns `0, 1, 2, ..., 10, 11`.
+
+Nothing sorts `Partition` objects today, so it is dormant — and making ordering explicit is what would
+wake it up. The risk is not "sorted differently"; it is sorted differently *from the work items already
+written to the coordination index*. A build sorting by the inherited comparator would remap ordinals on
+any index with 11 or more shards, and is invisible in any test with fewer.
```
malliaridis: This reminds me of the comparison issue we discussed [here](https://github.com/opensearch-project/opensearch-migrations/pull/3187#discussion_r3612688892). Seems like the ordering issue is a problem elsewhere too? Not sure if considering this a bug instead of "canoncal order" would make things easier for us?

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+between them.
+
+The requirement is therefore *agreement*, not any particular order. A consistently different order is
+only a relabeling; coverage stays complete.
+
+The damage comes from *disagreement*: between the process that created the work items and the ones
+resolving them, between two workers, or across time as the partition set mutates or a new build
+changes the order while old work items still encode the old one. Then some partitions are read twice
+and others never at all — nothing fails, and the run reports success with documents missing.
+Determinism plus immutability is how workers agree without coordinating.
```
malliaridis: Perhaps the name comparison issue (see other comment) is actually one of the reasons we have dead letters?

malliaridis (commented):
There is a lot to comprehend, so I needed multiple iterations over the document and each section. This review may not be as value as that of others, but I thought I post it anyways.

In general I like the intent and the direction we are heading to.

I have only one concern, and that's about the "RFS" meaning. Right now, RFS stands for "replay-from-snapshot". With the implementation of this proposal, we stop replaying from snapshots and instead "replay-from-sources". I don't believe we can consider all these sources "snapshots", as a snapshot is more like "a point-in-time, immutable capture of 'something'". So I believe it is worth taking that into account while reviewing this proposal. It is likely not an issue introduced with this proposal, but it is becoming very obvious here.

_P.S. if you want me to leave a few comments on something specific, let me know and I will take a closer look on that :)_

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+```
+
+## The interfaces
+
+The contract is three interfaces: enumeration, ordering, streaming. Separating them lets each be
+stated, reused and tested independently; composing them keeps existing call sites unchanged. Today's
+`DocumentSource` carries a fourth method, `readCollectionMetadata`, which is not part of the contract —
+see the note under "Composition".
+
+### 1. Enumeration — how many partitions there are, and what each is called
```
gregschohn: Partitions don't need to be ordered.  They're converted into integers, so we need to make sure that they're unique & an easy way to do that is to order them and take their index.  However, that ordering only needs to be stable for a single version of RFS.  We'll never have different versions of RFS coordinating together.

We'll run partitions as completely independent units.  We only do the integral ordinal assignment to succinctly identify/address them.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+     */
+    Flux<Document> readDocuments(Partition partition, long startingDocOffset);
+}
+```
+
+The offset's units are source-defined, but it must obey one law, because the pipeline advances it
+additively as `startingDocOffset + docsInBatch`:
+
+> After reading *n* emitted documents from offset *s*, resuming from *s + n* must not omit any document
+> not yet emitted. Repetition is allowed.
```
gregschohn: Let's be more careful - we shouldn't be able to repeat any docs in the stream.  It's ok to send a document twice if we already sent up to X, but the last work item was marked at X-N.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+## Why ordering is correctness-critical
+
+Work items are created by partition *position* (`ShardWorkPreparer` iterates `listPartitions` and
+assigns ordinals) and later resolved by position (`DocumentMigrationBootstrap.resolvePartition` does
+`partitions.get(shardIdx)`). Those are two independent `listPartitions` calls, potentially in
+different processes on different hosts, minutes or hours apart. The ordinal is the only thing carried
+between them.
+
+The requirement is therefore *agreement*, not any particular order. A consistently different order is
+only a relabeling; coverage stays complete.
```
gregschohn: As per above, I'm not too concerned.  These labels must be unique.  They don't even really need to be ordered.  If squeezing them into an ordinal pigeonhole is too much work, we should reevaluate how we serialize the labels in the work-item documents.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+`partition.name()` alongside it.
+
+**Also:** stop `Partition` extending `Comparable<Partition>`. A throwing `compareTo` would only turn
+accidental sorting into a runtime failure; removing the interface makes it a compile error. Safe today
+— nothing holds partitions in a sorted collection, the snapshot caches are `HashMap`, and no
+implementor overrides `compareTo`.
+
+## Immutability
+
+A deterministic comparator only guarantees the same order for the same *set*. A partition appearing or
```
gregschohn: We iterate over all of the partitions up-front an divide work up at that point.  We won't have two different processes doing this unless one failed and another took over, in which case, we can discard everything from the first setup (though I don't think that we do today).  Also note that we'll NEVER have two different VERSIONS of the software doing this iteration.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+A new `:RfsSourceSpi` module on `RfsPipeline` + `RfsCommon` resolves both — `RfsCommon` declares
+`api project(':transformation')`, so `Version` comes through, and `SnapshotReader` and `SolrReader`
+already depend on both. `RootDocumentMigrationContext` stays out: it lives in `:RFS`, which declares
+`api project(':RfsPipeline')`. No substitute is needed, since the only thing a provider uses is
+`IRfsContexts.IDeltaStreamContext` — already in `RfsCommon`, and the type `LuceneSnapshotSource`
+declares today. The SPI must not grow a shadow of the RFS tracing hierarchy.
+
+```java
+/** Declarative description of where documents come from. */
+public record DocumentSourceSpec(
```
gregschohn: I'd expect to see a simple interface (which we probably already have) and different implementations for each source.  [collection|index]AllowList is something that would be spelled out differently for each one of those sources.  Maybe for a raw S3 ingest there is a filter option like 'prefix' that's totally different than what it should be for an ES snapshot.

Kind makes sense in the base implementation, but I wouldn't go any further - all of the other details seems like they would only need to be understood by the provider & I'm expecting to have 3+ providers - ES, Solr, S3-DLQ.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+    String s3Region,
+    URI endpoint,
+    VersionStrictness versionStrictness,
+    Supplier<IRfsContexts.IDeltaStreamContext> deltaStreamContextFactory,
+    SourceRecordReporter recordReporter
+) {}   // both non-null: supply no-op implementations, never null
+
+/** Side channel for records a source read but could not convert. Must be thread-safe. */
+public interface SourceRecordReporter extends AutoCloseable {
+    Mono<Void> recordUnredrivable(UnredrivableRecord record);
```
gregschohn: what does unredrivable mean?  Is the terminology right? How do you know that you won't ever be able to redrive it, even w/ a new transform or a target tweak?

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+    String kind,                      // "es-snapshot" | "solr-backup" | "failed-document-stream"
+    String uri,                       // file:// | s3:// | gs://
+    Version version,                  // nullable for kinds not discriminated by version
+    String name,                      // snapshot name / backup name / session id
+    List<String> collectionAllowlist,
+    Map<String, String> options       // kind-specific
+) {}
+
+/** Shared services a provider may need. Deliberately excludes anything target-shaped. */
+public record SourceRuntime(
```
gregschohn: There could be instances of these for each type of provider?  If so, maybe there's some shared code, especially for lucene indices.  I'd work from final interfaces inward rather than trying to make it generalizable from go.  These helper configs seem like they'll wiggle quite a bit as we refine.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+    @Override
+    default void close() throws Exception {}
+}
+
+/** Identity is objectKey + recordOrdinal; the key already embeds session=, index= and worker=. */
+public record UnredrivableRecord(
+    String collection, String partition, String objectKey, long recordOrdinal, String reason
+) {}
+
+public interface DocumentSourceProvider {
```
gregschohn: I'm very interested to think about what the implicit contract looks like for this.  Is it thread safe?  Can it block?  (I'm pretty sure that the answer to both are yes).  As a thought exercise, if the source was a bulk-getter, say against a live OpenSearch cluster, how well would we be able to manage throughput and performance with this contract.  Maybe it doesn't matter too much as long as we can run enough document sources/backfills in parallel with appropriate limits.

docs/PluggableDocumentSourcesDesign.md
gregschohn: We already do a mixture of both of these today.  Recall that some users might not be using the K8s distributions and might be building their own backfillers or their own images - in those cases, the pluggable API can make sense.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+    default boolean deferUntilWorkAvailable() { return false; }
+
+    DocumentSource create(DocumentSourceSpec spec, SourceRuntime runtime) throws IOException;
+}
+```
+
+Providers are discovered with `ServiceLoader`, so each reader module ships its own
+`META-INF/services` entry and adding a source means adding a module rather than editing the core. The
+application module applies no shade plugin, so service files merge safely. The repo already uses this
+pattern for `IJsonTransformerProvider`.
```
gregschohn: nit: this is a bit more info.  A link to Java's SPI loading mechanism would probably serve just as well.

Similar note - not sure what the next paragraph says.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+registered.
+
+Discovery needs an assembled-application test: a missing `META-INF/services` entry compiles and
+unit-tests clean, then finds zero providers at runtime.
+
+### Construction stays lazy
+
+Source construction is deferred until inside the work coordinator's lifetime. The Solr path relies on
+this to check whether the coordinator reports work already complete *before* any S3 setup, so a pod
+restart after completion does not redo the bucket listing and metadata download. The registry
```
gregschohn: check before download makes sense - there are multiple levels of metadata, right?  Partitions and within those doc metadata.  I preseume that we only download when we're really sure - but we'll also find out pretty quickly too (in wall-time relative to what we need to download).

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+Unchanged from today.
+
+```java
+public interface DocumentStream {
+    /**
+     * Stream documents for a partition, starting from the given offset.
+     * Returns a cold Flux — subscription triggers the read. Replayable: repeated
+     * subscription yields the same sequence.
+     */
+    Flux<Document> readDocuments(Partition partition, long startingDocOffset);
```
gregschohn: startingDocOffset implies that we know the overall count.  This could be expensive to compute for some data sources.  It COULD even be infinite for open-ended ones.  What we really want to be able to do is to bisect partitions.  

When RFS starts, we start at the beginning and work as many docs as we can.  As the lease expires, we want the next runner to pick up where it left off.  To do that, we write the last document index that made it all the way through.  That allows the next runner to skip past those documents...  That ASSUMES that we have a relatively inexpensive way to find the n'th document in a stream.

If it would be easier for this to be a pagination token, we could adjust the API here to take in a string or json blob instead.  That might point to a specific segment, a prefix path, etc.  That alone would still get us the ability to recover where we left off.

Eventually, I really would like to be able to cut one partition (shard) of data into multiple processes for upload.  The pagination system as just described (get the point where we were) would hold back the process of fanning out.  However, we could eventually add another method like getBisectionPoint(startingCursor, endingCursor) and it could return a midpoint so that [startingCursor, midpoint), [midpoint, endingCursor) would hit all of the items.  That wouldn't need to be perfect, just enough to help run a divide-and-conquer algorithm.

Let me know what you think about changing the API contract here.  It's definitely going to be a bit more work in the existing codebase, but I think it might obviate even more work in future code.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+   deduplicating on them does not inflate its counts.
+
+Item 6 is not optional. It is the specific regression that would otherwise ship silently.
+
+## First new source: failed-document-stream redrive
+
+The stream is already laid out as `session=<id>/index=<target>/worker=<id>/*.ndjson.gz`, which maps
+onto the interfaces directly: collections are `index=` prefixes, partitions are `worker=` prefixes,
+documents are records. Records deserialize back into `BulkOperationSpec` with the same
+`ObjectMapper` — id, body, operation type, routing and `_type` all return typed, and unknown
```
gregschohn: This would be a great spot for a transform to take the source record - the full ndjson record entirely - and transform it into a complete OS operations: Add: This Document... That first transform would go from S3->original request stream and the second transform would still be chained on after that.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+
+Deserializing a record into `BulkOperationSpec` proves the record round-trips, not that everything
+survives the pipeline. `Document` is narrower, and `BulkOperationConverter` builds a fresh operation
+from it, so anything not carried or regenerated is gone.
+
+| Stored property | Redrive behavior |
+|---|---|
+| id, original body, index vs delete, `_type`, routing | carried in `Document` (id, source, operation, hints) |
+| target index | carried as the collection name |
+| `op_type`, `version`, `version_type`, `if_seq_no`, `if_primary_term`, `pipeline`, `require_alias` | regenerated by the transform |
```
gregschohn: Where are these values from?  I understand some - but seq_no/primary_term/pipeline - how are those currently specified?

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+The provider checks one precondition in `validate(spec, runtime)`: the session has a completion manifest
+(below). A stream still being written can gain a `worker=` prefix sorting before existing ones, shifting
+every ordinal.
+
+Two more belong at the composition layer, since a source cannot see the sink's configuration:
+
+- The failure-stream output location must differ from the input, or the run reads and writes the same
+  place and never terminates. Compare normalized bucket, prefix and session, not session name alone.
+- The run needs its own `--session-name` so work coordination does not collide with the original
+  backfill.
```
gregschohn: This is 100% true.  I would call these out as outside the bounds of this design and simply caveat this for now by saying that calling code will need to do these things.

gregschohn: There probably is a risk that a user could try redriving docs that failed previously but had a subsequent or colliding version on the target.  Imagine a case where a user was demuxing two indices into one.  There could be id collisions.  In those cases, we need to make sure that users REALLY, REALLY understand what's happening.  Maybe this makes redriving even more sensitive and we warnings that are even more pronounced.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+
+Source behavior, not a precondition: the `index=unknown-index/` bucket — failures emitted before the
+target index was known — is excluded from `listCollections()` and reported as unredrivable.
+
+### Unredrivable records
+
+`requestItem` is nullable, and older or malformed records may not convert to a `Document`. The dividing
```
gregschohn: This is even before the initial transform, right?  I agree that if the ndjson is corrupted, we cannot proceed and need to fail + log loudly.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+- **Corruption fails loudly** — unreadable gzip or NDJSON, or an unsupported manifest schema version.
+  The latter fails in `validate`, before any work starts.
+- **Valid but unredrivable records are skipped, counted, and reported** with object key and record
+  ordinal: null `requestItem`, an unparseable operation, an unknown operation-type discriminator, a
+  delete with no id, or an index operation with no body — deletes legitimately have none. Unknown
+  *fields* are ignored, for forward compatibility.
+- Null-id index operations follow the policy under "Decisions taken here".
+
+Records are categorized four ways: redriven, failed again at the target, skipped as unredrivable, and
+unreadable. **Phase 3 emits these as durable, categorized diagnostics and metrics; aggregating them into
```
gregschohn: i'm not sure what the definitions here are - esp between unreadable and unrederivable.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+a run state is #3066's `CompletedWithErrors` work.** Until that lands the run does not report all four,
+and the console feature in Phase 4 depends on it.
+
+Reporting needs a channel — `Flux<Document>` cannot carry a skip event. `SourceRuntime` carries a
```
gregschohn: what is a skip event?

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+The bootstrap owns the reporter, not `DocumentSource`: terminal flush, then close after the attempt.
+Flush failure is fatal; close failure after a successful flush is logged, since the reports are already
+durable. The no-op returns `Mono.empty()` throughout.
+
+Reports duplicate, because under-advancing offsets cause records to be re-read. Deduplicate on write by
+`objectKey` plus `recordOrdinal`, or skipped counts inflate on every retry.
+
+These are record categories, not work-item states; `CompletionStatus` stays as it is.
+
+### Session closure: a completion manifest
```
gregschohn: I think that I understand some of this section, but not most of it.  1) Is it required for pluggable sources, even if the S3-DLQ is the first new one and 2) a diagram of how you're seeing this fit together would be helpful

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+unredrivable. That is where operator judgement belongs — producing a manifest, not replacing one.
+
+**A seal is permanent.** The manifest controls enumeration and ordering; a live listing is used only to
+validate integrity, excluding the manifest object itself. A missing or extra object means the session
+was not closed when sealed, and the run fails. There is no re-seal — a conditional publish cannot
+replace an existing key, and rewriting one would silently remap ordinals. Correction means copying the
+objects into a new source session and sealing that. A fresh `--session-name` is not sufficient; it
+changes the coordination namespace, not the source manifest.
+
+### Reading within a partition
```
gregschohn, line 313: See the earlier comment about moving away from ordinals if that would make life simpler

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+
+## Sequencing
+
+- **Phase 1** — introduce the interfaces, registry and conformance test; port the two existing sources.
+  No functional or argument change (log output gains `partition.name()`). Independently reviewable and
+  mergeable.
+- **Phase 2** — expose `--source-kind` and the `--source-*` argument group, with inference from
+  today's arguments as the default.
+- **Phase 3** — add the failed-document-stream source, plus the manifest writer on the
+  `NoWorkLeftException` path, the console `seal` command, and the durable `SourceRecordReporter`.
```
gregschohn: I think I might be understanding that I'm misunderstanding what seal is supposed to be doing.  Can you define this term/feature more completely earlier in the doc?

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+
+Phase 4 must also correct #3065, whose premise predates two changes: the stored request is the
+pre-transform document, so it is not replayable as sent, and there is no `dlq` command group — only
+`failed-document-stream`.
+
+## Open questions
+
+- **Post-redrive reconciliation.** Source records are never deleted, so a redrive leaves the original
+  session holding every failure plus a new session holding the ones that failed again — with nothing
+  marking which are resolved. An operator who redrove everything still sees the original count. This
```
gregschohn: The snapshotmigration CR status seems like the most natural place to track this.  
As the first run proceeds, the count of failed docs is appended.
Each time that we do these reconciliation amendments with new RFS workflows, they should append/amend the list to show how many were remedied and how many were not.

docs/PluggableDocumentSourcesDesign.md
```diff
@@ -0,0 +1,554 @@
...
+  hits #3063 (error summary), #3064 (counts in CR status) and #3066 (does a successful redrive clear
+  `CompletedWithErrors`, and does a redrive with its own failures enter it?). Design it here, or as
+  its own issue across those three?
+
+## Decisions taken here
+
+Recorded with rationale so they can be challenged rather than rediscovered.
+
+- **All redrives use the coordinated path.** No size threshold, no lightweight in-process mode. One
```
gregschohn: I think that this is fair.  A trickier question is whether or not we should try to REUSE the coordinating node when an RFS ended up in a CompletedWithErrors state.  I think that would be a good later optimization.  That would drop the console command from minutes to a few seconds for a small number of failed docs.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Makes sense. I will remove `PartitionOrdering` and add partition names to work items.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Tightened.

> Resuming from the cursor emitted with document `d` must yield every document the source would have emitted after `d`, and must not re-emit `d` or anything before it.

Duplicate delivery is still expected because a lease that expires after documents were sent but before the cursor was recorded will re-send them. That is the watermark lagging, not the source re-reading and the first draft mixed those two up.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Ok. Names go into the work item, and I took out the ordering.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Confirmed and no discard today. But `createUnassignedWorkItem` is an upsert and the same partition always yields the same id, so a re-run recreates what's there and fills in the rest. Dropped the follow-up from the doc rather than adding it.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Makes sense. `DocumentSourceSpec` is now a marker interface with just `kind()`, a record per source, and `DocumentSourceProvider<S extends DocumentSourceSpec>` so `validate` and `create` get their own type rather than a cast. Kind selects the provider; the provider parses the rest via `parseSpec`.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: SourceRuntime stays single. The split is spec vs. runtime, and I had two fields on the wrong side. S3 region, endpoint, and version strictness are user config, so I moved them to the typed specs. Shared Lucene code is already there in `SearchSnapshotExtractor`, which both readers depend on.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: I was imagining a record that couldn't be converted, without considering that a modified transform might convert it fine. There shouldn't be any such thing as an "unredrivable" record, so I removed that text.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Yes to both and I updated the text to say concurrent subscription across partitions must be safe.

No read-side rate control. A worker migrates exactly one partition, so parallelism means more workers. `batchConcurrency` bounds writes, not reads. A live-cluster bulk-getter would need to throttle itself in its implementation.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Both done. ServiceLoader is just a link now. Fixed the next paragraph, too: 

> The registry maps normalized `kind()` to provider, built and checked once at startup - a duplicate or blank kind fails there, and the error names the provider class - while source construction stays lazy.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Yes, there are three levels: 

* `listCollections` reads snapshot metadata
* `listPartitions` reads shard metadata
* `readDocuments` unpacks the shard files <--- expensive but runs only once a work item is held.

I added those the levels in the doc.

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: `applyTransformation` correlates `originalSource` by document id, which is unset until transform 1 runs, so a re-failure during redrive would store the post-transform body instead of the source document. 

That's different than what we do now. Is preserving the pre-transform body worth keeping for a redrive?

docs/PluggableDocumentSourcesDesign.md
jzonthemtn: Removed.

It was meant to show that every property stored in the failure stream has somewhere to go when a record passes through the narrow `Document` type. With the transform-chain suggestion nothing narrows, so this isn't needed. 

seq_no/primary_term/pipeline are OpenSearch bulk API action-line parameters that RFS models but never populates, so a transform is the only thing that can set them.

jzonthemtn:
Phases 1 and 2 are now implemented, minus the plumbing for the workflow path.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/provider/EsSnapshotSourceProvider.java
```diff
@@ -0,0 +1,106 @@
...
+
+    public static final String KIND = "es-snapshot";
+
+    @Override
+    public String kind() {
+        return KIND;
+    }
+
+    @Override
+    public EsSnapshotSpec parseSpec(JsonNode config) {
```
gregschohn: Is this how we pass configs around elsewhere in the repo?  I know that I've done this at some points (used lower level parsed jackson output), but I also recall finding different ways.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/provider/EsSnapshotSourceSpec.java
```diff
@@ -0,0 +1,118 @@
...
+ * @param emitDocType                carry the ES {@code _type} through to the sink
+ * @param previousSnapshotName       previous snapshot for a delta read, else null
+ * @param deltaMode                  which delta changes to emit, else null
+ * @param enableSourcelessMigrations reconstruct {@code _source} for indices that disabled it
+ */
+public record EsSnapshotSpec(
+    String repoUri,
+    String snapshotName,
+    Version version,
+    List<String> indexAllowlist,
```
gregschohn, line 35: Given this field, would it be more fair to call this something like an EsSnapshotReaderSpec or to include the base interface name in this one (starts to get long)?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -260,7 +257,7 @@ public static class Args {
         @Parameter(required = false,
             names = { "--source-version", "--sourceVersion" },
             converter = VersionConverter.class,
-            description = ("Version of the source cluster. Required when --source-type is SNAPSHOT."))
+            description = ("Version of the source cluster. Required; a SOLR flavor selects the Solr backup path."))
```
gregschohn: nit: I've never liked the word 'flavor' for forks.  Do we show "ES.."/"OS..." already or no?  This is probably largely a moot point since we have the userSchema wrapping this for most users

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -462,13 +467,83 @@ public String getTransformerConfigParameterArgPrefix() {
...
         }
     }
 
+    private record SupersededArg(String flag, Predicate<Args> wasGiven) {}
+
+    /**
+     * Arguments whose values now live in {@code --source-config}. Working directories are absent on
+     * purpose: they configure {@link SourceRuntime}, not the spec, and are still required.
+     */
+    private static final List<SupersededArg> SUPERSEDED_BY_SOURCE_CONFIG = List.of(
```
gregschohn: what's the purpose of this class?  Why not just absorb all of these command line options and force their complete migration into the json configs?

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -462,13 +467,83 @@ public String getTransformerConfigParameterArgPrefix() {
...
+        }
+    }
+
+    /**
+     * Requirements that outlive the source spec. Mirrors the inferred paths: a snapshot unpacks
+     * segments and so needs a working directory, a Solr backup reads in place and does not, and a
+     * Solr backup cannot coordinate work on the target.
+     */
+    private static void validateRuntimeArgs(Args args, DocumentSourceProvider<?> provider) {
+        if (provider instanceof SolrBackupSourceProvider) {
```
gregschohn: should the provider have access to all of the RFS arguments?  Should coordinator host be something in a base class for all configs?  This isn't too bad here, but it does beg the question, what's unique about this setting that it should live at a global level outside the provider's config

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -780,73 +840,102 @@ static OptionalInt classifySnapshotReadFailure(Exception e, Args arguments) {
...
 
-    private static MigrationSourceFactory buildElasticsearchSourceFactory(
+    /** Which source to open: a registered kind plus that provider's own config. */
+    record SourceSelection(String kind, JsonNode config) {}
+
+    /**
+     * Takes {@code --source-kind} and {@code --source-config} when given, else infers the selection
+     * from the per-source arguments so existing invocations keep working.
+     */
+    static SourceSelection selectSource(Args arguments, boolean emitDocType) {
```
gregschohn, line 963: 1) This seems like it could be a factory class.  2) (related to my other comment about collapsing command line options into the json configs) - I expect to see each provider manage its own configuration opaquely to the outer invoking program.

One case in point is emitDocType - why is this not rolled up into the ES source provider config.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -780,73 +840,102 @@ static OptionalInt classifySnapshotReadFailure(Exception e, Args arguments) {
...
+    static JsonNode readSourceConfig(String sourceConfig) {
+        if (sourceConfig == null) {
+            throw new ParameterException("--source-config is required when --source-kind is given.");
+        }
+        var mapper = ObjectMapperFactory.createDefaultMapper();
+        try {
+            if (sourceConfig.startsWith("@")) {
+                var path = Paths.get(sourceConfig.substring(1));
+                return mapper.readTree(Files.readString(path));
+            }
```
gregschohn, line 1004: Elsewhere in the tree, we have a separate argument to pull this from a file.  This has the advantage of less ambiguity.  If you want to keep this, we should open an issue to collapse all of the ...-file arguments to use this

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
```diff
@@ -780,73 +840,102 @@ static OptionalInt classifySnapshotReadFailure(Exception e, Args arguments) {
...
+            arguments.experimental.enableSourcelessMigrations
+        ).toJson());
+    }
+
+    /** Reads {@code --source-config}, either inline JSON or {@code @path} naming a JSON file. */
+    static JsonNode readSourceConfig(String sourceConfig) {
+        if (sourceConfig == null) {
+            throw new ParameterException("--source-config is required when --source-kind is given.");
+        }
+        var mapper = ObjectMapperFactory.createDefaultMapper();
```
gregschohn, line 999: Can we create a mapper with FAIL_ON_UNKNOWN_PROPERTIES=true so that mistaken keys (& ideally incorrect types) cause errors up-front rather than being silently ignored?

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/bulkload/pipeline/provider/ProviderSpecRoundTripTest.java
```diff
@@ -0,0 +1,134 @@
...
+    private final SolrBackupSourceProvider solrProvider = new SolrBackupSourceProvider();
+
+    @Test
+    void esSnapshotSpec_roundTripsEveryFieldWhenAllAreSet() {
+        var original = new EsSnapshotSpec(
+            "s3://bucket/snapshots", "nightly", Version.fromString("ES 7.10.2"),
+            List.of("logs", "metrics"), "us-east-2", "http://localhost:4566",
+            true, 4242L, true, true, "previous", DeltaMode.UPDATES_AND_DELETES, true);
+
+        assertThat(esProvider.parseSpec(original.toJson()), equalTo(original));
```
gregschohn, line 34: Is there a need for toJson() on the config objects?  Once you flatten the ambiguous parameters at the global level, we should never need to show a different/merged config.

That would largely moot the need for this test - though maybe toJson could be run for this test or the test might not be needed.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/DocumentSourceDiscoveryTest.java
```diff
@@ -0,0 +1,73 @@
...
+import static org.hamcrest.Matchers.instanceOf;
+
+/**
+ * Guards the assembled application: a missing {@code META-INF/services} entry unit-tests clean and
+ * then finds zero providers at runtime, so discovery itself has to be asserted.
+ */
+class DocumentSourceDiscoveryTest {
+
+    @Test
+    void serviceLoaderFindsEveryShippedProvider() {
```
gregschohn: I'm not sure that this test is valuable if the goal is to test the final package.  The test jar closure will be different than the release/production one since Gradle manages those configs separately.

That's fine because we should have integ tests for each of these sources, which should exercise that.

RFS/src/main/java/org/opensearch/migrations/bulkload/pipeline/adapter/OpenSearchDocumentSink.java
```diff
@@ -89,27 +90,32 @@ public Mono<Void> createCollection(CollectionMetadata metadata) {
     }
 
     @Override
-    public Mono<BatchResult> writeBatch(String collectionName, List<Document> batch) {
-        long bytesInBatch = batch.stream()
+    public Mono<BatchResult> writeBatch(String collectionName, List<PositionedDocument> batch) {
+        var documents = batch.stream()
+            .map(PositionedDocument::document)
+            .collect(Collectors.toList());
```
gregschohn: nit - you probably don't gain much by creating copies of this list since the operations are trivial on it

RFS/src/main/java/org/opensearch/migrations/bulkload/worker/ShardWorkPreparer.java
```diff
@@ -116,21 +115,22 @@ private static void prepareShardWorkItems(
             })
             .forEach(indexName -> {
                 var partitions = documentSource.listPartitions(indexName);
-                var shardCount = partitions.size();
                 log.atInfo()
-                    .setMessage("Index {} has {} shards")
+                    .setMessage("Index {} has {} partitions")
```
gregschohn: this class is still called ShardWorkPreparer - this could be a good spot to include "(shards)" alongside to tie the terms together

RfsPipeline/src/main/java/org/opensearch/migrations/bulkload/pipeline/source/PartitionEnumerator.java
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Enumerates the collections and partitions a source can offer.
+ *
+ * <p>Partitions are addressed by name, not position: a work item records the name and resolves it
+ * later, in another process. Enumeration order carries no meaning.
+ */
+public interface PartitionEnumerator {
+
+    /** All collection names available from this source. Deterministic. */
```
gregschohn: what does deterministic mean?  Do you mean stable order?  Aren't all of these going to be deterministic for the same input?

RfsPipeline/src/testFixtures/java/org/opensearch/migrations/bulkload/pipeline/source/DocumentSourceContractTest.java
```diff
@@ -0,0 +1,193 @@
...
+    protected abstract DocumentSource newSource() throws Exception;
+
+    /** A collection the source offers that has at least one partition with at least two documents. */
+    protected abstract String collectionUnderTest();
+
+    @Test
+    void listCollections_isDeterministic() throws Exception {
+        try (var source = newSource()) {
+            var first = source.listCollections();
+            var second = source.listCollections();
```
gregschohn: if the ordering wasn't deterministic, there's a good chance that this test might not find it.  Is there something else that we could do to catch when the order or contents are different between different runs?  Hashing all of the contents after a sort or a timestamp/path as a key that maps to a hashed value of what this list returned?

This is also something that we should eventually catch w/ enough logging and integ tests.

gregschohn:
I only have one major comment in the last review - consolidating the command line options solely into the json configurations.

I have another followup too and it can be for a follow-up.  This PR introduces a lot of new top-level directories.  We should consolidate everything snapshot related (including RFS & Metadata Migrate) under a directory, similar to TrafficCapture for live-traffic.  From there, we can figure out if we want more layers of projects or not.

I don't want that to disrupt other work in this stream, but we should get to it sooner rather than later.

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/bulkload/pipeline/provider/EsSnapshotSourceSpec.java
jzonthemtn: Renamed to `EsSnapshotSourceSpec` and `SolrBackupSourceSpec` for alignment). Went with `Source` over `Reader` since there is already `SnapshotReader` classes and wanted to try to avoid confusion between the two.

DocumentsFromSnapshotMigration/src/test/java/org/opensearch/migrations/DocumentSourceDiscoveryTest.java
jzonthemtn: Agreed. Removed.

RFS/src/main/java/org/opensearch/migrations/bulkload/pipeline/adapter/OpenSearchDocumentSink.java
jzonthemtn: Fixed.

RfsPipeline/src/main/java/org/opensearch/migrations/bulkload/pipeline/source/PartitionEnumerator.java
jzonthemtn: Replaced with an explicit contract (unique names, unspecified order, same set across calls and fresh sources).

jzonthemtn:
> I only have one major comment in the last review - consolidating the command line options solely into the json configurations.
> 
> I have another followup too and it can be for a follow-up. This PR introduces a lot of new top-level directories. We should consolidate everything snapshot related (including RFS & Metadata Migrate) under a directory, similar to TrafficCapture for live-traffic. From there, we can figure out if we want more layers of projects or not.
> 
> I don't want that to disrupt other work in this stream, but we should get to it sooner rather than later.

Yes, that's a good point. I captured it in a new issue in #3334 so it is not forgotten.

PR #3274, opened 2026-08-02
Properly seed GHA's Gradle test Docker cache

gregschohn:
## Summary

- add dedicated, writable Docker cache primer jobs before the Gradle test matrix
- keep Gradle cache image groups in the existing private-ECR image manifest instead of adding CI-specific manifest files
- make all Gradle test shards wait for and restore both completed caches
- prefer `mirror.gcr.io` for Docker Hub images, with retries and canonical-registry fallback

## Root cause

The Gradle test jobs restore the Docker image cache in read-only mode. The job that previously wrote that cache was removed when the Docker Compose E2E test moved to Jenkins, leaving no reliable producer for the images used by Gradle Testcontainers tests.

The old cache key also hashed unrelated Dockerfiles and Compose files instead of the images the cache contained. The new keys are derived from the repository's existing image manifest.

## Image inventory

The existing private-ECR manifest remains the single inventory file. Its original `charts` and `images` sections are unchanged and continue to drive isolated-network deployment mirroring. A separate `gradleTestImages` section identifies the 21 images relevant to Gradle cache priming and divides them into infrastructure and search caches.

Only five Gradle images overlap the 58 deployment images. Pulling all deployment images into every Gradle shard would waste cache quota and restore time, so the primer reads only the CI section. The full Elasticsearch/OpenSearch version matrix remains excluded because it is too large for the GitHub Actions cache quota.

## Validation

- parsed `.github/workflows/CI.yml` and the image manifest with Ruby YAML
- confirmed the deployment manifest still resolves to the same 15 charts and 58 images
- ran `bash -n` on the seed and mirroring scripts
- exercised mirror success, retagging, retry/fallback, canonical pulls, and invalid groups with a mocked Docker command
- ran `git diff --check`

### DIFF
--- a/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/infra/mirror/private-ecr-manifest.yaml
+++ b/deployment/k8s/charts/aggregates/migrationAssistantWithArgo/infra/mirror/private-ecr-manifest.yaml
@@ -1,6 +1,7 @@
 # Version-locked list of container images and Helm charts required by the
 # Migration Assistant EKS deployment. This file is packaged with the Helm chart
 # and consumed by aws-bootstrap.sh when mirroring public artifacts into ECR.
+# The gradleTestImages section is CI-only and is ignored by ECR mirroring.
 
 charts:
   # Format: "name|version|repository"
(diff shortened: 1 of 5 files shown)


deployment/k8s/charts/aggregates/migrationAssistantWithArgo/infra/mirror/private-ecr-manifest.yaml
```diff
@@ -108,3 +109,27 @@ images:
   - "mirror.gcr.io/library/amazoncorretto:11-al2023-headless"
   - "mirror.gcr.io/library/amazoncorretto:8-alpine"
   - "mirror.gcr.io/library/amazonlinux:2023"
+
+gradleTestImages:
```
jzonthemtn: Since this is just for CI, I _think_ you can put this into a file like `.github/gradle-test-images.txt` and reference it from there in case you don't want `gradleTestImages` in this file and next to important non-CI stuff.

jzonthemtn (approved):

PR #3277, opened 2026-08-02
Verify EKS CloudWatch dashboard lifecycle

gregschohn:
## Summary

- wait for both ACK Dashboard resources to report `ACK.ResourceSynced=True`
- verify that both CloudWatch dashboards exist and have nonempty bodies
- delete the Dashboard resources before normal EKS teardown and verify both Kubernetes finalization and AWS-side deletion

## Why

The existing EKS integration test did not fail when dashboard reconciliation or deletion failed. These assertions cover the CloudWatch dashboard permissions used by the ACK controller and catch orphaned dashboards during cleanup.

The eventual-consistency waits run after bootstrap and during post-build cleanup, so they do not extend the Helm install path.

## Validation

- `git diff --check`
- compiled both changed Jenkins shared-library scripts with Groovy 3.0.24


vars/verifyEksDashboards.groovy
```diff
@@ -0,0 +1,95 @@
...
+
+        dashboards.each { dashboard ->
+            sh """
+                set -eu
+                dashboard_body="\$(aws cloudwatch get-dashboard \
+                  --dashboard-name '${dashboard.name}' \
+                  --region '${region}' \
+                  --query DashboardBody \
+                  --output text)"
+                if [ -z "\$dashboard_body" ] || [ "\$dashboard_body" = "None" ]; then
```
jzonthemtn: I think this check will pass if the dashboard is empty (contents are `{}`). Not sure if that's a concern or not.

vars/verifyEksDashboards.groovy
```diff
@@ -0,0 +1,95 @@
...
+            """
+        }
+        return
+    }
+
+    sh """
+        set -eu
+        kubectl --context='${kubeContext}' --namespace ma delete \
+          ${resources} \
+          --wait=false
```
jzonthemtn, line 75: `--ignore-not-found` to avoid script returning  an exit code  just in case? Might not be necessary if nothing else could have already removed it.

jzonthemtn (approved):

vars/verifyEksDashboards.groovy
gregschohn: I don't think that anybody should be deleting this earlier... if there was an unknown deletion, I'd like to treat that as an error.

PR #3284, opened 2026-08-03
Make Solr context path configurable via `--solr-context-path`

### LINKED ISSUE #3265: Solr context path is hardcoded to `/solr`, breaking reverse-proxied and non-default-context deployments
### Description

Every Solr URL the migration tooling builds appends a literal `/solr` to the user-supplied `--source-host`. Solr's context path is configurable, and reverse proxies commonly rewrite it. Against any such deployment there is no way to make the tooling produce a correct URL, as `--source-host` only supplies scheme/host/port, and the path segment is not configurable.

### Affected call sites

`CreateSnapshot/src/main/java/org/opensearch/migrations/SolrBackupStrategy.java`

| Line | Purpose |
|---|---|
| 387 | config file fetch — `/solr/{collection}/admin/file` |
| 404 | schema fetch — `/solr/{collection}/schema` |
| 560 | cloud-vs-standalone detection — `/solr/admin/collections?action=LIST` |
| 611 | collection discovery — `/solr/admin/collections?action=LIST` |
| 623 | core discovery — `/solr/admin/cores?action=STATUS` |

`SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/`

| File | Line | Purpose |
|---|---|---|
| `SolrSnapshotCreator.java` | 73 | `action=BACKUP` |
| `SolrSnapshotCreator.java` | 148 | `action=REQUESTSTATUS` |
| `SolrStandaloneBackupCreator.java` | 61 | `replication?command=backup` |
| `SolrStandaloneBackupCreator.java` | 84 | `replication?command=details` |

### Impact

A user whose Solr is served at a non-`/solr` context path cannot run the tooling at all.

### Proposed fix

Make the context path configurable and thread it through both modules, such as by one of the following:

1. Add an explicit option (e.g. `--solr-context-path`, default `/solr`) and build URLs from it everywhere.
2. Honour any path already present on `--source-host` (so `--source-host http://host/tenant` resolves to `http://host/tenant/admin/collections`) and stop appending `/solr` unconditionally. Lower ceremony, but changes the meaning of existing `--source-host` values that include a trailing path.

Raised by @malliaridis while reviewing #3235.

jzonthemtn:
### Description
Make Solr context path configurable via `--solr-context-path`. Previously, `/solr` was hardcoded so it didn't support Solr installations running at a context path other than `/solr`.

### Issues Resolved
#3265 

### Testing
Added unit and integration tests.

### Check List
- [X] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable.


migrationConsole/lib/console_link/console_link/models/snapshot.py
```diff
@@ -18,9 +18,28 @@
...
 
 SOLR_NO_DELETE_MSG = "Solr backups are managed as files; no delete API available."
-SOLR_LIST_COLLECTIONS_API = "/solr/admin/collections?action=LIST&wt=json"
+SOLR_LIST_COLLECTIONS_PATH = "/admin/collections?action=LIST&wt=json"
 SOLR_NO_SNAPSHOT_REPO_MSG = "Solr does not use snapshot repositories."
 CREATE_SNAPSHOT_COMMAND = "/root/createSnapshot/bin/CreateSnapshot"
+DEFAULT_SOLR_CONTEXT_PATH = "/solr"
+
+
+def normalize_solr_context_path(context_path: Optional[str]) -> str:
```
gregschohn: How is solr_context_path being threaded into here?

There are 2 spots that I am expecting to see that I don't.
1) When checking the status of backup to see its progress.  We do that for ES snapshots.  Do we have an analog for solr?  If so, we'll need to make sure that the check, spun up from a workflow, through a console container with a services.yaml config, has the right context path set.
2) When running console commands with a workflow that has this path set, how does the value make it in?

gregschohn (commented):
Thanks - just one comment that isn't too major, but I think it might be a blocker for E2E functionality.

migrationConsole/lib/console_link/console_link/models/snapshot.py
jzonthemtn: Both valid because the value actually never reached the console.

There is an analog for Solr --- `_solr_backup_status` polls REQUESTSTATUS per collection for SolrCloud, or the replication handler for standalone, and emits the same field names as the ES snapshot status (percent completed, total and successful shards, data processed, ETA) so the workflow's `checkScript` parses it
unchanged. Non-deep checks return SUCCESS/IN_PROGRESS/FAILED like ES.

What was missing was the wiring to it. `solrContextPath` sat on `createSnapshotConfig`, which only reaches the Java container. The monitor cronjob's services.yaml comes from `getConsoleConfig` and never carried it, so status polling always hit `/solr`.

Same root cause with the same fix.

Moved it to the source cluster. It's a connection property, it covers the external-backup path (no `createSnapshotConfig` there), and it drops the duplication. `Cluster` carries it now instead of `SNAPSHOT_SCHEMA`.

gregschohn (approved):
Thanks for these changes and the additions!  LGTM

PR #3289, opened 2026-08-04
Fail integ tests fast when a workflow blocks on an unexpected approval gate

gregschohn:
## Problem

When a `ValidatingAdmissionPolicy` rejects a migration resource apply, the workflow **parks** rather than failing. `reconcile*Resource` runs `tryApply` with `continueOn: {failed: true}`, and `waitForFix` then blocks on the `<kind>.<name>.vapretry` ApprovalGate until someone approves it.

**That is intended behavior and this PR does not change it.** `K8S_USER_APPROVAL_WAIT_RETRY_STRATEGY` is documented as "intentionally open-ended in elapsed time"; the parked branch is a sibling step group so unrelated work keeps running; and the operator recovers in place with `workflow reset <resource>` followed by `workflow approve` (`RESETTABLE_PLURALS` deliberately excludes `approvalgates`, so the gate survives the reset and the retry loop can re-create the resource fresh). Making VAP rejection fail the workflow would turn a recoverable state into a dead end.

Integration tests are the problem. Nobody is going to approve anything, so:

1. The workflow stays `Running` forever. `ENDING_ARGO_PHASES` cannot see a workflow blocked on a gate, so `wait_for_ending_phase` has exactly one exit: `TimeoutError` at `MIGRATION_COMPLETION_TIMEOUT_SECONDS` (1800s), or up to 4800s for the CDC replayer wait.
2. The resulting message says nothing about the gate — just a list of running nodes.
3. The harness polls the **outer** test wrapper (configure/monitor/evaluate), but the reconcile steps that can park live in the **inner** `migration-workflow` that `configureAndSubmitWorkflow.sh` submits. Even a node-level check on the polled workflow would miss every park.

## Change

A `ParkedGateWatcher` that the existing wait loops call on each iteration:

- Fails on **any** gate the workflow is blocked on that the test didn't ask for — runtime (`.vapretry`) or step. Both block on `waitForUserApproval` and neither clears without an approval the test has no way to give.
- Checks **both** the polled workflow and the inner `migration-workflow`.
- Throttled to one API check per 30s, so a loop polling every 5s doesn't hammer the API.
- Requires the same gate on **two consecutive** checks before failing. A single sighting only logs a warning — a `Failed` tryApply whose retry loop is about to move on, or a step gate the test is about to approve, shouldn't fail a test.
- Wired into `wait_for_suspend`, `wait_for_ending_phase`, Test0003's suspend-or-end loop, and the CDC `_raise_if_cdc_dependency_error` hook (a VAP-denied CaptureProxy or Replayer leaves the workflow parked, not errored, so none of the existing dependency checks fire and the pod simply never appears).
- Timeout messages raised for any other reason now carry a single-sighting observation as `unconfirmed_parked_gates`.
- The failure names the gate and the matching recovery: `workflow reset` + `approve retry` for an `Impossible` denial, `approve change` for a `Gated` one, `approve step` for a step gate. `ParkedApprovalGateError` subclasses `AssertionError`, since the workflow behaved correctly — the migration config or the test's expectations did not.
- `IntegrationTestArgoService.expected_parked_gate_names` opts specific gates out. Test0003 is the only user: it runs `skip-approvals=false` and deliberately waits at five gates it approves by name, so it declares those. Every other test runs `skip-approvals=true`, where no step gates exist at all and one waiting is unexpected by construction.

Those five are the complete set of step gates this configuration can produce, so the opt-out isn't papering over a gate the test currently misses — there are exactly five `waitForUserApproval` step call sites in the templates (the six `waitForFix` sites in `resourceManagement.ts` are the runtime `.vapretry` gates, a separate mechanism):

| Gate | Created by | Suppressed when |
|---|---|---|
| `begin` | `fullMigration.ts` `approveBegin` | `requireBeginApproval` false |
| `captureproxysetup.<proxy>` | `setupCapture.ts` `approveProxySetup` | `skipProxyApproval` / global `skipApprovals` |
| `evaluatemetadata.<suffix>` | `metadataMigration.ts` `approveEvaluate` | `skipEvaluateApproval` |
| `migratemetadata.<suffix>` | `metadataMigration.ts` `approveMigrate` | `skipMigrateApproval` |
| `documentbackfill.<crdName>` | `fullMigration.ts` `approveBackfill` | `documentBackfillConfig.skipApproval` |

Notably the CDC replay path (`runSingleReplay`) adds no step gate at all — the replayer is gated only by its runtime `trafficreplay.<name>.vapretry` gate. What the opt-out buys is the guarantee going the other way: any gate *beyond* the declared five now fails Test0003 promptly, so if a future change adds a step gate the test doesn't know to approve, it surfaces as a named failure instead of a 1800s hang.

## Two notes on scope

**Why the predicate isn't "an ApprovalGate that is `Pending` with a non-empty denial reason".** `statusSchemaFor` in `generateMigrationResources.ts` gives ApprovalGate a status of only `phase` ∈ {Created, Pending, Approved, Error}; there is no denial-reason field on the gate, and the VAP message lands on the Argo `tryApply` **node's** message instead. `Pending` is also the steady state for every gate the workflow hasn't reached yet (gates are pre-created, and `resetGate` patches them back to `Pending`), so phase alone can't separate "blocking now" from "not reached yet".

On the `Error` state specifically: it's in the enum but **nothing ever writes it**. Every write to a gate's phase is `Created` (`makeApprovalGateResource`), `Pending` (`patchApprovalGatePhase`, six call sites, all literal `"Pending"`), or `Approved` (`approve.py`). The only `patch*Error` template is `patchCaptureProxyError`, which sets the *CaptureProxy's* phase, not a gate's. And `waitForUserApproval` has only `successCondition: status.phase == Approved` with no `failureCondition` — unlike the non-gate resource waits, which do check for `Error` — so a gate that somehow reached `Error` would hang exactly like one at `Pending`.

So the check runs against the node graph: a `Running` approval node, plus (for a runtime gate) a `Failed` sibling under the same `boundaryID` carrying a parsed reason. That keys on "blocked at an approval node," which is independent of gate phase and therefore covers the `Error` case too. `workflow approve` already does exactly this traversal, so this reuses it via a new public `approve.waiting_gates` accessor rather than adding a second implementation.

**No inner-workflow teardown.** As suggested, it isn't needed: `teardownMaApp` runs `workflow reset -> helm uninstall -> PVCs -> namespace`, and the k8s-local path does `kind delete cluster`. The leaked inner workflow goes away with the cluster either way, so this PR is detection + fail-fast only.

## Testing

- 21 unit tests in `test_parked_gate_detection.py` covering throttling, the consecutive-observation requirement, transient-observation reset, expected-gate opt-out for both kinds, step-gate inclusion, error swallowing, and message formatting.
- 7 tests in `test_integration_test_argo_service.py` asserting both wait loops fail fast (one sleep, not the 600s budget) on a runtime gate and on an unexpected step gate, honor the opt-out, watch both workflow names, and report unconfirmed sightings in the timeout message.
- 1 test in `test_workflow_cli_commands.py` pinning `waiting_gates`' step/change/retry categorization.
- `test_cdc_base.py` updated for the new `parked_gate_watcher` parameter and asserts the watcher covers outer + inner.
- Full integ_test suite: 114 passed. console_link `test_workflow_cli_commands.py`: 76 passed. flake8 clean.

Detection never raises on its own — a missing, still-starting, or unreadable workflow yields no gates, so a wait loop can't fail for a reason unrelated to what it's checking.

### DIFF
--- a/migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
+++ b/migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
@@ -14,7 +14,7 @@ from console_link.models.cluster import Cluster, SIGV4_SIGNING_ENDPOINT_KEY
 from console_link.workflow.commands.crd_utils import CRD_GROUP, CRD_VERSION
 
 from ..cluster_version import CDC_MIGRATION_COMBINATIONS
-from ..parked_gate_detection import INNER_WORKFLOW_NAME, ParkedGateWatcher
+from ..parked_gate_detection import INNER_WORKFLOW_NAME, ParkedApprovalGateError, ParkedGateWatcher
 from .ma_argo_test_base import MATestBase, MigrationType, MATestUserArguments  # noqa: F401 (re-exported)
 
 logger = logging.getLogger(__name__)
@@ -170,7 +170,13 @@ def _raise_if_cdc_dependency_error(namespace: str, workflow_names: Optional[Sequ
     # A CaptureProxy or Replayer whose apply was VAP-denied leaves the workflow
     # parked, not errored, so nothing above fires and the pod never appears.
     if parked_gate_watcher:
-        parked_gate_watcher.check()
+        try:
+            parked_gate_watcher.check()
+        except ParkedApprovalGateError:
+            # Every other raise above dumps first; match that, since the gate name
+            # alone doesn't show which resource's apply was denied.
+            _dump_argo_workflow_diagnostics(namespace)
+            raise
 
 
 def _raise_if_kafka_cluster_error(namespace: str) -> None:
--- a/migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
+++ b/migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
@@ -14,7 +14,7 @@ from console_link.models.cluster import Cluster, SIGV4_SIGNING_ENDPOINT_KEY
 from console_link.workflow.commands.crd_utils import CRD_GROUP, CRD_VERSION
 
 from ..cluster_version import CDC_MIGRATION_COMBINATIONS
-from ..parked_gate_detection import INNER_WORKFLOW_NAME, ParkedApprovalGateError, ParkedGateWatcher
+from ..parked_gate_detection import INNER_WORKFLOW_NAME, ParkedGateWatcher
 from .ma_argo_test_base import MATestBase, MigrationType, MATestUserArguments  # noqa: F401 (re-exported)
 
 logger = logging.getLogger(__name__)
@@ -170,13 +170,7 @@ def _raise_if_cdc_dependency_error(namespace: str, workflow_names: Optional[Sequ
     # A CaptureProxy or Replayer whose apply was VAP-denied leaves the workflow
     # parked, not errored, so nothing above fires and the pod never appears.
     if parked_gate_watcher:
-        try:
-            parked_gate_watcher.check()
-        except ParkedApprovalGateError:
-            # Every other raise above dumps first; match that, since the gate name
-            # alone doesn't show which resource's apply was denied.
-            _dump_argo_workflow_diagnostics(namespace)
-            raise
+        parked_gate_watcher.check()
 
 
 def _raise_if_kafka_cluster_error(namespace: str) -> None:
--- a/migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
+++ b/migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
@@ -14,6 +14,7 @@ from console_link.models.cluster import Cluster, SIGV4_SIGNING_ENDPOINT_KEY
 from console_link.workflow.commands.crd_utils import CRD_GROUP, CRD_VERSION
 
 from ..cluster_version import CDC_MIGRATION_COMBINATIONS
+from ..parked_gate_detection import INNER_WORKFLOW_NAME, ParkedGateWatcher
 from .ma_argo_test_base import MATestBase, MigrationType, MATestUserArguments  # noqa: F401 (re-exported)
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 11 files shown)


migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
```diff
@@ -157,14 +158,19 @@ def _get_capture_proxy_readiness_status(namespace: str) -> tuple[str, str, str,
...
     phase, message, _, _ = _get_capture_proxy_readiness_status(namespace)
     if phase == "Error":
         _dump_proxy_readiness_diagnostics(namespace)
         raise RuntimeError(f"captureproxy/{PROXY_RESOURCE_NAME} entered Error phase: {message}")
     _raise_if_kafka_cluster_error(namespace)
     _raise_if_migration_resource_error(namespace)
     _raise_if_argo_workflow_error(namespace, workflow_names)
+    # A CaptureProxy or Replayer whose apply was VAP-denied leaves the workflow
+    # parked, not errored, so nothing above fires and the pod never appears.
+    if parked_gate_watcher:
```
jzonthemtn, line 172: Minor -- I think `check()` raises without dumping anything first.

```
  if parked_gate_watcher:
      try:
          parked_gate_watcher.check()
      except ParkedApprovalGateError:
          _dump_argo_workflow_diagnostics(namespace)
          raise
```

jzonthemtn (approved):
+1 lgtm

migrationConsole/lib/integ_test/integ_test/test_cases/cdc_base.py
gregschohn: You were right that it wasn't dumping — thanks. I took your fix as written first (456e58cd), then Greg pushed back that it only covered one site, and he was right, so I've replaced it with a fix at the top level (1f32e6aa).

`check()` is called from four places: here, `wait_for_suspend`, `wait_for_ending_phase`, and `_wait_until_suspended_or_ended`. Only this one had a dump. In the other three the dump sits on the timeout path *below* the wait loop, which `check()` raising skips straight past — so a per-site fix here would have left three sites silently undumped.

`ParkedApprovalGateError` propagates uncaught from all four, so the failure dump in `ma_workflow_test` teardown should already have covered every one of them, with more output than the two `kubectl get`s here (five calls including `print_workflow_details` and `save_namespace_diagnostics`). Except it didn't fire, for a worse reason: teardown quiesced the workflow *before* dumping, and the `check()` inside `wait_for_ending_phase` raised there too — so on a parked gate the dump was skipped, and the evidence is unrecoverable once the namespace is reset.

So teardown now dumps first, then quiesces, and isolates each step so nothing can preempt the dump, the workflow delete, the CRD reset, or `cleanup()`. Dumping ahead of `stop_workflow` also preserves state that stopping destroys: the suspended node and which resource's apply was denied — the thing you actually wanted from the CI log. Errors are re-raised at the end, a lone one unwrapped so a parked gate still surfaces as `ParkedApprovalGateError`.

That covers all four sites at once, so I reverted this hunk as redundant. 20 new tests in `test_ma_workflow_teardown.py`; 6 of them fail against the previous teardown.

PR #3315, opened 2026-08-22
Add global serial replay ordering example test

gregschohn:
## Summary
Add GlobalSerialReplayE2ETest as a concrete example of replaying traffic while preserving global request order.

## Details
This test builds randomly interleaved traffic across many connections, replays it with MAX_CONCURRENT_REQUESTS=1, and verifies that requests arrive at the target in the original source order.

## Testing
Not run locally in this session.


TrafficCapture/trafficReplayer/src/test/java/org/opensearch/migrations/replay/e2etests/GlobalSerialReplayE2ETest.java
```diff
@@ -0,0 +1,196 @@
...
+import org.junit.jupiter.api.parallel.ResourceLock;
+
+@WrapWithNettyLeakDetection(disableLeakChecks = true)
+@Tag("longTest")
+class GlobalSerialReplayE2ETest {
+
+    private static final String NODE_ID = "global-order-test-node";
+    private static final int REQUEST_COUNT = 4_096;
+    private static final int CONNECTION_COUNT = 32;
+    private static final long CONNECTION_RANDOM_SEED = 0x5EEDL;
```
jzonthemtn, line 44: Just commenting I like this 5eed.

jzonthemtn (approved):
lgtm

PR #3316, opened 2026-08-23
Decouple source auth from capture proxies and snapshot migrations

gregschohn:
## Summary

- Stop source-cluster auth changes from being projected into `CaptureProxy` and `SnapshotMigration` custom resources.
- Remove those fields from the generated VAP checks and CLI/web dry-run policy model.
- Exclude source auth from proxy and direct snapshot-migration identity checksums.
- Keep source auth protected on `DataSnapshot`, which connects to the source cluster.

## Why

A capture proxy forwards traffic to the configured source endpoint at the network layer; it does not use the source cluster HTTP auth configuration. Snapshot migration stages consume a snapshot artifact and target-cluster credentials, so direct source-auth coupling there is also unnecessary. Treating these fields as immutable caused harmless auth changes to fail with `Impossible: sourceAuthType cannot be changed. Delete and recreate.`

## Resource audit

| Resource | Source auth behavior after this change |
| --- | --- |
| `CaptureProxy` | Not projected, checksummed, or admission-restricted |
| `DataSnapshot` | Still projected and immutable because snapshot creation/preparation connects to the source |
| `SnapshotMigration` | Not directly projected, checksummed, or admission-restricted; a changed managed-snapshot checksum can still correctly propagate after snapshot replacement |
| `CapturedTraffic` | No source-cluster auth coupling (unchanged) |
| `TrafficReplay` | No source-cluster auth coupling; target auth remains protected (unchanged) |
| `KafkaCluster` | No source-cluster auth coupling (unchanged) |

## Implementation

- Remove all six source-auth identity fields from CaptureProxy and SnapshotMigration CRD projections and rendered apply manifests.
- Keep endpoint, version, insecure-transport, and source-label identity behavior unchanged.
- Build proxy and snapshot-migration checksums from a source identity that explicitly excludes auth.
- Preserve the DataSnapshot source-auth fields and checksum behavior.
- Update resolved migration resources so preflight and generated MigrationRun data match the actual CR specs.

## Testing

- `mise x node@24.19.0 -- npm test` from `orchestrationSpecs`: 418 passed, 2 skipped.
- Type checks passed for schemas, config-processor, and migration-workflow-templates.
- Added regression coverage for generated VAPs, dry-run decisions, rendered manifests, managed snapshot propagation, proxy checksums, and external-snapshot migration stability.


orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
```diff
@@ -1001,7 +1004,7 @@ export class MigrationConfigTransformer extends StreamSchemaTransformer<
                 snapshotConfigChecksum,
                 resourceName: crdName(m.sourceLabel, m.targetConfig.label, m.label, m.migrationLabel),
                 configChecksum: cs(
-                    sourceConnectionIdentity,
+                    sourceConnectionIdentityWithoutAuth,
```
jzonthemtn, line 1007: Source auth is excluded from this checksum, but it is still passed into `sourceConfig` and put into the metadata container. If snapshot migration never accesses the source, should that credential propagation be removed, too?

jzonthemtn (approved):
Approved - one question but it's not a blocker.

orchestrationSpecs/packages/config-processor/src/migrationConfigTransformer.ts
gregschohn: Thanks.  I've stripped those other vestigial bits too!

PR #3317, opened 2026-08-23
Use Gradle-managed Node for SolrTransformations post-install

gregschohn:
## Summary
Make the SolrTransformations Gradle post-install step use the Node binary managed by the Gradle node plugin.

## Why
The SolrTransformations runPostInstallScripts task is a plain Exec task, so its npm rebuild esbuild call can fall back to whatever npm is on the host PATH instead of the pinned Gradle-managed Node toolchain. Prepending Gradle's resolved Node bin directory keeps that rebuild step on the same toolchain as the rest of the build.

This matters because the installTransformDeps task already runs under the Gradle-managed Node setup, but the follow-on rebuild step did not inherit that same toolchain deterministically.

## Testing
Not re-run after narrowing the PR to the Gradle change only.

### DIFF
--- a/TrafficCapture/SolrTransformations/build.gradle
+++ b/TrafficCapture/SolrTransformations/build.gradle
@@ -1,4 +1,5 @@
 import com.github.gradle.node.npm.task.NpmTask
+import com.github.gradle.node.NodeExtension
 
 plugins {
     id 'org.opensearch.migrations.java-library-conventions'
@@ -14,6 +15,8 @@ node {
     npmWorkDir = project.layout.buildDirectory.dir("npm")
 }
 
+def nodeExtension = extensions.getByType(NodeExtension)
+
 tasks.register('installTransformDeps', NpmTask) {
     description = 'Install TypeScript transform dependencies'
     args = ['install', '--ignore-scripts']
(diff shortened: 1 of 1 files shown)


orchestrationSpecs/packages/argo-workflow-builders/jest.config.js
```diff
@@ -6,7 +6,12 @@ module.exports = {
     moduleFileExtensions: ["ts", "js", "json", "yaml", "yml", "sh"],
 
     transform: {
-        "^.+\\.(ts|js)$": "ts-jest"
+        "^.+\\.(ts|js)$": ["ts-jest", {
+            tsconfig: {
+                allowJs: true,
+                noEmit: true
```
jzonthemtn: I'm not sure the `noEmit: true` is stopping the Jest TS emit.

I think it is overridden to always be `false` in `config-set.js`:

```
  _overriddenCompilerOptions = {
      inlineSourceMap: false,
      declaration: false,
      isolatedDeclarations: false,
      noEmit: false,
      // ...
  };
```

jzonthemtn (approved):
Approved, but take a look at `noEmit`. It might always be overridden to be `false` regarldless of it manually being set to true.

PR #3318, opened 2026-08-23
Move onto the stable OpenTelemetry semconv artifact

gregschohn:
### Description

`opentelemetry-semconv` has been pinned at `1.23.1-alpha` since before the library split into stable and incubating artifacts at 1.30.0. That split deleted both catch-all classes we import — `ResourceAttributes` and `SemanticAttributes` are gone from `io.opentelemetry.semconv` — so the stale pin is the only reason `coreUtilities` still compiles.

Our own Gradle build never notices. Anything that repackages `coreUtilities` and resolves a current semconv underneath the already-compiled jars does: `NoClassDefFoundError` out of `initializeOpenTelemetryForCollectors` before the SDK is built, which takes the process down at startup rather than degrading telemetry. This moves to `1.43.0` and onto the classes that replaced them.

The resource attributes are a no-op on the wire — `service.name` and `service.instance.id` were both promoted into the stable artifact as `ServiceAttributes`, so the emitted keys are byte-identical.

`exception.escaped` needed an actual decision. The spec dropped it and there is no stable equivalent, and it turns out we were getting nothing for it: the `awsxray` exporter's cause translator reads `exception` span events for `exception.type` / `exception.message` / `exception.stacktrace` and never looks for `escaped` — I checked contrib `v0.119.0`, `v0.124.0`, and `v0.159.0`, spanning the ADOT `v0.43.3` we pin through today. The attribute has been serialized, shipped, and discarded on every propagating exception. It does reach the Jaeger path, but only as an unindexed log field.

What the conventions model instead is span status `ERROR` plus a low-cardinality `error.type`, which is what both backends consume. The `isPropagating` distinction is preserved: propagating sets the status, non-propagating still just records the event.

### One behavior change worth a close look

Setting the status closes a gap that predates this PR. Nothing in the tree ever called `setStatus`, and these spans carry no `http.status_code`, so the X-Ray translator's fault branch — which only fires on status `ERROR` once it has failed to find an HTTP code — never fired. An escaping exception produced a `cause` payload with the stack trace attached to a segment marked neither error nor fault, so **it never counted toward fault rates and never lit up the service map**.

Those exceptions now register. Anyone alarming on X-Ray fault rate for these services should expect the number to climb off zero to whatever was always true. Happy to split this into a follow-up if you'd rather land the semconv bump on its own.

`error.type` also becomes an indexed annotation under our `index_all_attributes: true`, queryable as `error_type` once X-Ray sanitizes the dot — an event attribute never was.

### ADOT collector bump

Also bumps the ADOT collector from `v0.43.3` to `v0.49.0` (upstream collector/contrib `v0.156.0`) at all four places it is named, which have to move together or the private mirror won't carry the tag the chart asks for:

| File | Before | After |
|---|---|---|
| `TrafficCapture/.../otelCollector/Dockerfile` | `v0.43.3` | `v0.49.0` |
| `.../migrationAssistantWithArgo/values.yaml` (`otelCollectorImage`) | `v0.43.3` | `v0.49.0` |
| `.../infra/mirror/private-ecr-manifest.yaml` | `v0.43.3` | `v0.49.0` |
| `.../values.yaml` (`opentelemetry-operator` → `manager.collectorImage.tag`) | `0.43.2` | `v0.49.0` |

That last one also fixes a latent bug: the operator chart templates `--collector-image={{ repository }}:{{ tag }}` verbatim, and public ECR has no bare `0.43.2` tag — only `v`-prefixed ones plus `latest`. It's been unresolvable all along, masked because our Collector CRs set `image:` explicitly.

The ADOT `v0.49.0` release is dependency bumps, a Go toolchain update, and the removal of the `eks` resourcedetection detector for Fargate container insights — nothing touching the components we configure.

Left alone deliberately: `extraOtelConfiguration.version: "0.124.0"`, which is the Operator's own upstream collector version rather than an ADOT image.

### Issues Resolved

None filed; found while investigating a downstream repackaging that crash-looped on the missing class.

### Testing

- 3 new tests in `ScopedInstrumentationExceptionSignalingTest` — propagating sets `ERROR` + `error.type`, non-propagating leaves status `UNSET` with no `error.type`, and a message-less exception falls back to the type name instead of an empty status description.
- `:coreUtilities:test --tests 'org.opensearch.migrations.tracing.*'` — 11/11 pass.
- `:TrafficCapture:transformationShim:test --tests '*tracing*'` — 17/17 pass, including the two existing tests that drive `addTraceException(e, true)`.
- `compileJava compileTestJava compileTestFixturesJava` across all modules — clean, confirming `1.43.0` resolves and nothing else in the tree touched the deleted classes (a repo-wide grep found only these two importers).
- Ran all four `otel-config-*.yaml` from the docker solution against both `v0.43.3` and `v0.49.0`: all 8 start clean and reach `Everything is ready` with no config errors or deprecation warnings. Also built the `otelCollector` image on the new base and confirmed the container reports `healthy` and `/healthcheck` still exits 0.

### Check List


coreUtilities/src/main/java/org/opensearch/migrations/tracing/IScopedInstrumentationAttributes.java
```diff
@@ -85,13 +86,25 @@ default void close() {
...
+        span.recordException(e);
         if (isPropagating) {
-            span.recordException(e, Attributes.of(SemanticAttributes.EXCEPTION_ESCAPED, true));
-        } else {
-            span.recordException(e);
+            // An exception that escapes this scope means the operation failed, which the semantic
+            // conventions model as span status ERROR plus error.type -- the removed exception.escaped
+            // attribute recorded the same thing in a form no backend consumes.
+            span.setStatus(StatusCode.ERROR, describeThrowable(e));
+            span.setAttribute(ErrorAttributes.ERROR_TYPE, e.getClass().getCanonicalName());
```
jzonthemtn: Personal preference - I think `getName()` will always return something, while `getCanonicalName()` can return `null` so I have leaned toward using it.

jzonthemtn (approved):
Minor comment - definitely not a blocker.

coreUtilities/src/main/java/org/opensearch/migrations/tracing/IScopedInstrumentationAttributes.java
gregschohn: Good catch, and it turned out to matter more than a preference. `getCanonicalName()` is null for anonymous and local classes, and `setAttribute` silently drops a null, so an exception from an anonymous subclass would have produced a span marked ERROR with no `error.type` at all. Worse, OTel's own `recordException` uses `getCanonicalName()` too, so `exception.type` is missing from those same spans — I confirmed that with a scratch test, the event carries only `exception.message` and `exception.stacktrace`. Nothing left to identify the failure by.

Switched to `getName()` in 65abc35, plus the same fix in the status description, where `getSimpleName()` returns empty rather than null for an anonymous class. Added a test covering it.

PR #3319, opened 2026-08-23
Scope migration resource logs by CR UID

gregschohn:
## Summary

- add `migrations.opensearch.org/migration-resource-uid=<metadata.uid>` as the authoritative ownership label for migration workload pods
- propagate the UID through direct Argo pods, controller-created workloads, Strimzi pod templates, and monitor CronJob/Job/Pod templates
- make `workflow log resource` select only the chosen CR's UID instead of intersecting inherited relationship labels
- add `workflow log ... --show-pods` pod attribution for historical logs
- retain `workflow log filter` for intentional cross-resource log investigation

## Problem

Migration CR labels such as `source`, `snapshot`, `target`, and `from-snapshot-migration` describe relationships. They are not unique workload ownership keys. Nested resources legitimately share them.

For example, `DataSnapshot/source-snap1` and a later `SnapshotMigration` can both carry `source=source` and `snapshot=snap1`. The previous `workflow log resource` implementation selected every propagated CR label, which allowed metadata migration or document backfill pods to appear under the DataSnapshot that produced the input snapshot.

Kafka required an additional `strimzi.io/cluster` special case, and workflow-name filtering did not consistently apply to controller-created or long-lived workloads. Adding more negative selectors would preserve those ambiguities and require resource-type-specific knowledge in every log client.

## Implementation

The owning migration CR UID is now applied to:

- CreateSnapshot, MetadataMigration, and S3 traffic loader Argo pods
- CaptureProxy, TrafficReplay, and RFS bulk-loader Deployment pod templates
- the RFS coordinator StatefulSet pod template
- Strimzi KafkaNodePool broker pod templates and Kafka entity-operator pod templates
- snapshot and RFS monitor CronJob, Job, and Pod metadata

The ownership label is also present on the workload controller objects where applicable. Monitor reconciliation patches the label onto existing CronJobs and verifies it at all three CronJob/Job/Pod template levels.

`workflow log resource` reads `metadata.uid` from the selected CR and uses only that globally unique value. It deliberately does not append an Argo workflow selector because controller-created workloads and monitor jobs can outlive the workflow that created them. A missing UID fails explicitly rather than falling back to broad relationship labels.

Historical `workflow log` output accepts `--show-pods`. Each merged line is prefixed with `[pod/<name>]`, while timestamp ordering and the default output remain unchanged. Stern already identifies pod/container sources in follow mode.

Pods created from older templates do not acquire this label retroactively; their owning workload must reconcile or be recreated before UID-scoped resource lookup can find them.

## Integration Coverage

The dedicated Kind workflow test now installs minimal test-only `DataSnapshot` and `SnapshotMigration` CRDs and submits one real Argo workflow with two pods:

- both pods deliberately share the legacy `source` and `snapshot` relationship labels
- each pod carries the UID of a different migration CR
- `workflow log resource datasnapshot.<name> --show-pods` must include the DataSnapshot pod name and message
- the related SnapshotMigration pod name and message must both be absent

This exercises CR lookup, Kubernetes pod selection, actual `kubectl logs`, timestamp merging, and CLI rendering instead of only asserting selector construction.

The integration fixture now allocates an ephemeral local Argo port so an existing developer forward on `2746` cannot silently direct tests to another cluster.

## Validation

- `npm run -w @opensearch-migrations/migration-workflow-templates type-check-all`
- `npm run -w @opensearch-migrations/migration-workflow-templates test-all -- --runInBand`
  - template package: 61 tests
  - Argo workflow builders: 142 passed, 2 skipped
  - schemas: 101 tests
- `pytest tests/workflow-tests/test_workflow_cli_commands.py -q`
  - 78 passed
- `pytest tests/workflow-tests/test_workflow_integration.py -q`
  - 18 passed against the dedicated Kind/Argo cluster
- `bash -n` for both modified monitor scripts
- verified the installed Strimzi CRDs accept label maps at both pod-template paths


migrationConsole/lib/console_link/console_link/workflow/commands/log.py
```diff
@@ -248,6 +243,8 @@ def _output_options(func):
                         help='Stream live logs via stern instead of showing history')(func)
     func = click.option('--timestamps', is_flag=True, default=False,
                         help='Show timestamps in log output')(func)
+    func = click.option('--verbose', is_flag=True, default=False,
```
jzonthemtn: `--prefix-pod` maybe? I could see there being a `--verbose` in the future that is more of the traditional verbosity control.

orchestrationSpecs/packages/migration-workflow-templates/resources/scripts/applyRfsMonitorCronJob.sh
```diff
@@ -27,6 +27,7 @@ CLAIMED_AT="$(date -u +%s)"
 
 WORKFLOW_UID_LABEL="${RFS_MONITOR_WORKFLOW_UID_LABEL}"
 SESSION_LABEL="${RFS_MONITOR_SESSION_LABEL}"
+RESOURCE_UID_LABEL="migrations.opensearch.org/migration-resource-uid"
```
jzonthemtn: This is defined a lot. It might just be across different file types and there's nothing that can be done to simplify it.

jzonthemtn: I'd say that's it - at least one .ts, a couple in .py, and a couple in .sh.

migrationConsole/lib/console_link/console_link/workflow/commands/log.py
gregschohn: Renamed this to `--show-pods` in 0918d5d35. The plural reflects that merged history can include lines from multiple pods, and this keeps `--verbose` available for conventional diagnostic verbosity later.

orchestrationSpecs/packages/migration-workflow-templates/resources/scripts/applyRfsMonitorCronJob.sh
gregschohn: Consolidated this in 0918d5d35. The authored literal now has exactly two language roots: the Python log command definition and the TypeScript resource-label definition. Python tests import their root; TypeScript injects its root into the monitor containers. Each shell script requires the injected `MIGRATION_RESOURCE_UID_LABEL` with `: "${MIGRATION_RESOURCE_UID_LABEL:?}"`, so an unset or empty value fails immediately instead of silently rendering an incorrect workload.

jzonthemtn: Nice, that sounds good.

jzonthemtn (approved):

PR #3331, opened 2026-08-25
fix: dense_vector→knn_vector NPE and total_fields.limit stripped on metadata migration

epugh:
## Summary

- **Bug 1 (`es-vector-knn-metadata.js`)**: When a `dense_vector` field has no `index_options` (the common case — just `{"type": "dense_vector", "dims": 768}`), the transformer emitted `knn_vector` with `parameters: {encoder: {name:"sq"}, m: null, ef_construction: null}`. OpenSearch 3.0+'s lucene engine mapper throws a `NullPointerException` on the null values, surfacing as `"Failed to parse mapping [_doc]: null"` on index creation. Fix: only include `parameters`/`encoder` when the source field actually had `index_options`; otherwise omit `parameters` and let OpenSearch use its defaults.

- **Bug 2 (`OpenSearchIndexCreator.java`)**: `INTERNAL_SETTINGS_EXACT` contained `"mapping"` which stripped the entire `index.mapping.*` subtree from settings — including the user-configured `index.mapping.total_fields.limit`. This reset the limit to the OpenSearch default (1000), causing index creation to fail for any source index with more than 1000 mapped fields. Fix: removed `"mapping"` from the exact-match set (it was too broad).

Both bugs hit simultaneously for any wide-schema ES 8.x index with simple `dense_vector` fields targeting OpenSearch 3.0+. Validated end-to-end against the Chorus ecommerce index (5,373 fields, `dense_vector` without `index_options`) migrating from Elasticsearch 8.5 → OpenSearch 3.7.

Here is a link to the example ES schema: https://github.com/querqy/chorus-elasticsearch-edition/blob/main/elasticsearch/schema.json

## Test plan

- [ ] New JS unit tests in `es-vector-knn-metadata.test.js` cover: with `index_options`, without `index_options` (the bug case), dynamic templates, `index.knn` flag, and all similarity mappings
- [ ] New Java unit test `testCreate_preservesMappingTotalFieldsLimit` verifies `index.mapping.total_fields.limit` survives the settings strip while `uuid` and `creation_date` are still removed
- [ ] End-to-end validated: metadata migration of Chorus ecommerce index (ES 8.5 → OS 3.7) now succeeds with 0 issues and 19,364 documents indexed, with no manual pre-creation workaround


RFS/src/test/java/org/opensearch/migrations/bulkload/version_os_2_11/IndexCreator_OS_2_11Test.java
```diff
@@ -268,6 +268,43 @@ void testCreate_withRetryToRemoveRemovedTokenFilters() throws Exception {
...
+            "      },\n" +
+            "      \"number_of_shards\": \"1\",\n" +
+            "      \"number_of_replicas\": \"0\",\n" +
+            "      \"uuid\": \"should-be-stripped\",\n" +
+            "      \"creation_date\": \"1234567890\"\n" +
+            "    }\n" +
+            "  }\n" +
+            "}";
+
+        var result = create(client, rawJson, "indexName");
```
jzonthemtn, line 296: This doesn't look like it tests the code changed by the PR. `create` goes through IndexCreator_OS_2_11, while the fix is in `OpenSearchIndexCreator.stripInternalSettings()`. I think this is also why the test is failing.

transformation/standardJavascriptTransforms/src/es-vector-knn-metadata.js
```diff
@@ -13,24 +13,23 @@ function main(context) {
...
+                name:       'hnsw',
+                engine:     'lucene',
+                space_type: mapSimilarity(similarity),
             };
+            // Only include explicit HNSW parameters when the source field had index_options.
+            // Omitting parameters lets OpenSearch use its own defaults, which avoids
+            // NPE in OS 3.0+ when m/ef_construction would otherwise be null.
+            if (opts) {
+                const { m, ef_construction: efConstr } = opts;
+                const params = { encoder: { name: 'sq' } };
```
jzonthemtn, line 27: `index_options` doesn’t necessarily mean scalar quantization. Elasticsearch supports plain type: "hnsw" as well as int8_hnsw, int4_hnsw, and bbq_hnsw. This adds OpenSearch’s `sq` encoder for all of them. 

Could the encoder be driven by `opts.type` instead? For example, plain hnsw should have no encoder, while int8_hnsw could map to the closest supported `sq` configuration.

How should int4_hnsw and bbq_hnsw be handled? Fall back to `sq` with a logged warning?

This might belong better in a separate issue/PR because it might be starting to get out of scope from the original change and using `sq` was already here before this PR.

jzonthemtn:
@epugh Check your commit signing.

PR #3292, opened 2026-08-05
[codex] Remove AppRegistry from solution templates

gregschohn:
## Summary

- remove Service Catalog AppRegistry resources from the ECS and EKS solution templates
- stop exporting `MIGRATIONS_APP_REGISTRY_ARN` and remove the AppRegistry CDK dependency
- remove the downstream ECS CDK consumer that imported the application and associated migration stacks
- retain AWS Solutions adoption tracking through the existing `SO0290` solution metadata and `AwsSolution/<solution-id>/<version>` user agent
- update unit assertions, snapshots, and stale fixtures for all four generated templates

## Why

The legacy AppRegistry integration is no longer required for AWS Solutions adoption tracking. Keeping it in the templates adds obsolete resources and can prevent otherwise valid solution deployments.

## Validation

- `npm run test:lint`
- `npm run test:jest -- --runInBand` (7 tests and 4 snapshots passed)
- `STACK_NAME_SUFFIX="" ./gradlew :deployment:migration-assistant-solution:cdkSynthMinified`
- verified all four minified ECS/EKS templates contain no `AWS::ServiceCatalogAppRegistry` resources or `MIGRATIONS_APP_REGISTRY_ARN`
- `npm run test:lint` in `deployment/cdk/opensearch-service-migration`
- `npm run build` in `deployment/cdk/opensearch-service-migration`
- `npm run test:jest -- --runInBand` in `deployment/cdk/opensearch-service-migration` (109 tests passed)
- `cargo test domain::cfn` in `deployment/k8s/aws/cli` (13 tests passed)

`npm run build` continues to hit the existing `@types/eslint__js` stub-package TypeScript error; the release workflow's Gradle synthesis path passes.

PR #3335, opened 2026-08-26
Stop tagging deployed resources

gregschohn:
### Description

Deploying either CDK app currently requires `<service>:TagResource` and `<service>:UntagResource` on every taggable resource type. Some organizations deny those actions via SCPs, which fails the deployment outright — `iam:UntagRole` and `ecr:UntagResource` are the usual first casualties. This removes the tagging.

Untag permission is not avoidable by "allow tag, deny untag":

- `AWS::IAM::Role` is the one resource type whose provider declares `iam:UntagRole` on its **create**, update *and* delete handlers, so even a first-time create needs it.
- For everything else it is needed on every **update**, because changing a tag's value is an untag followed by a tag. A version-valued tag guarantees that on every upgrade.

Nothing read these tags back:

- The app-level `migration_deployment` tag dates from the Copilot era, when it was used to find orphaned resources across regions by hand. The README text describing that workflow and the `cdk bootstrap --tags` call that propagated it are both long gone; the only remaining references were the definition itself and one unit-test assertion.
- The subnet `Name` tags in the solution stacks were cosmetic. `vpcName` on the two create-VPC stacks went with them, since CDK only renders it as a `Name` tag and neither app looks a VPC up by name (both `Vpc.fromLookup` call sites use `vpcId`).
- AWS Solutions attribution does not depend on tags — it rides on the `MIGRATIONS_USER_AGENT` / boto3 `user_agent_extra` header. The AppRegistry resources that *did* track deployed resources were removed earlier, in #3292.

Two tag families are load bearing and are kept:

| Tag | Read by |
|---|---|
| `kubernetes.io/cluster/<name>`, `kubernetes.io/role/elb`, `kubernetes.io/role/internal-elb` on subnets | AWS Load Balancer Controller, to discover where it may place load balancers |
| `aws-cdk:subnet-name`, `aws-cdk:subnet-type` | `Vpc.fromLookup()`, to classify existing subnets as public/private/isolated |

### Implementation notes

`RemoveResourceTags` is an aspect rather than a set of deletions because CDK's own L2 constructs tag themselves: `Vpc` adds a `Name` tag, `Subnet` adds the `aws-cdk:subnet-*` pair, `eks.Cluster` tags the subnets it is given. The tags have to be stripped from the `TagManager` after those aspects have run, which is what the explicit priority above `AspectPriority.MUTATING` buys — without it, aspects inherited from an ancestor run before a node's own aspects and the node-local tags survive.

Two things that look simpler but do not work:

- `addPropertyDeletionOverride('Tags')` is silently a no-op. `CfnResource._toCloudFormation` resolves the raw overrides before merging them, and the resolve drops the `undefined` that signals the deletion.
- Walking the tree before `synth()` finds nothing, because `Tags.of(x).add(...)` only registers an aspect; the `TagManager` is not mutated until that aspect visits during synthesis.

The aspect uses `TagManager.of()` so it covers both L1 tagging conventions — the `tags` property on older generated resources and `cdkTagManager` on newer ones. Missing the latter left a `Name` tag on `AWS::EC2::EgressOnlyInternetGateway`.

The `aws-cdk-lib` peer range moves from `^2.150.0` to `^2.172.0`, the first release with `AspectOptions.priority`. The devDependency was already pinned well above that.

### Issues Resolved

None filed; came in as a customer deployment failure.

### Testing

Unit tests, per package:

- `deployment/migration-assistant-solution`: new `remove-resource-tags.test.ts` asserts that every tag rendered by all four stack/`createVPC` combinations is on an `AWS::EC2::` resource and matches a functional prefix. Snapshots regenerated. `npx jest` → 11 passed, `npx eslint .` clean.
- `deployment/cdk/opensearch-service-migration`: new `remove-resource-tags.test.ts` makes the same assertion across every stack a fully-enabled `StackComposer` produces, plus a case pinning the keep-list behavior. `createApp.test.ts` now asserts the aspect is registered at the right priority instead of asserting the tag. `npx jest` → 111 passed, `npx eslint .` and `npx tsc --noEmit` clean. Confirmed both new tests fail if the aspect is not applied.

Synthesized all four vended templates (`npx cdk synthesize --all`) and audited every `Tags`/`tags` property:

| Template | Tags remaining |
|---|---|
| `Migration-Assistant-Infra-Import-VPC` | none |
| `Migration-Assistant-Infra-Import-VPC-eks` | none |
| `Migration-Assistant-Infra-Create-VPC` | `aws-cdk:subnet-name`, `aws-cdk:subnet-type` on `AWS::EC2::Subnet` |
| `Migration-Assistant-Infra-Create-VPC-eks` | the above, plus `kubernetes.io/*` on `AWS::EC2::Subnet` / `RouteTable` / `NatGateway` / `EIP` |

No IAM, ECR, Logs, EKS, EFS, ECS or S3 resource is tagged in any of the four.

Not yet verified by me: an actual EKS deployment. That is the one thing worth watching on this PR.

### Caveats

- Applying this to an **already-deployed** stack is itself a tag removal, so that one update still needs untag permission. It is clean on fresh deploys, which is the case that was failing.
- `cdk bootstrap` tags the `CDKToolkit` stack's own ECR repo and IAM roles. That is outside these templates and unaffected here.
- The `kubernetes.io/cluster/<name>` tag interpolates the `Stage` parameter, so its key is a CloudFormation intrinsic. CDK stores intrinsic-keyed tags separately from the ones `removeTag` can reach, so it would survive the aspect regardless — it is in the keep-list because it should be kept, not because the keep-list is what saves it.

### Check List
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable. — n/a, no documented behavior changes


jzonthemtn:
A couple questions on this --

* I don't think the PR removes _all_ tags (or does it?), so will the remaining tags' creation also be denied by the SCP?
* Should the all-or-nothing creation of tags be an option instead?

My personal bias toward having verbose tags is showing here.

gregschohn:
@jzonthemtn, Thanks for the feedback!

The underlying issue caused me to research how CloudFormation handles tag reconciliation.  There are several different ways (services and strategies) that can automatically add tags to newly created resources.  When CloudFormation does its reconciliation as a stack is being redeployed, it will drop those previously unknown tags.

This PR would actually CREATE an additional issue for users that were trying to upgrade but don't have untag permissions.  I'm parking this PR for now.  There's some value to tagging resources with the MA version, it could cause more pain, and it wouldn't fix the underlying issue that was observed.

PR #3338, opened 2026-08-27
Propagate deployer tags to EKS Auto Mode resources, and prove they land

gregschohn:
### Description

CloudFormation stack tags only reach resources that CloudFormation creates. EKS Auto Mode provisions nodes, EBS volumes, ENIs and load balancers later and on its own, so stack tags never touch them. For a deployer whose policy requires tags on `ec2:RunInstances`, that means **no nodes can launch at all**.

This adds `--tags <Key=Value,...>` to `aws-bootstrap.sh` — equivalent to the CloudFormation console's Tags step — and threads the same tag set through every mechanism Auto Mode actually reads:

| Resources | Mechanism |
|---|---|
| Everything CloudFormation creates | `aws cloudformation deploy --tags` (stack tags) |
| EC2 instances, ENIs, root/ephemeral volumes | custom `NodeClass` `spec.tags` |
| PVC-provisioned EBS volumes | StorageClass `tagSpecification_N` parameters |
| The capture proxy's NLB, its target group, listeners and frontend security group | `service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags`, emitted by `setupCapture.ts` from the `provider-config` ConfigMap |

**The prerequisite that makes the rest work.** The `AmazonEKS*` managed policies on the cluster IAM role only permit the `eks:`-prefixed tags Auto Mode adds for itself, so without an extra grant a single user tag makes every `RunInstances` fall outside any Allow — an implicit deny. Adding tags naively therefore makes things *worse*. `EksInfra` now attaches `AutoModeTagPropagationPolicy` (Compute / Storage / Networking / LoadBalancer), each statement conditioned on the request carrying this cluster's own `eks:eks-cluster-name` matched against the session tag on the assumed cluster role, so it grants nothing outside the cluster. See the [tag propagation docs](https://docs.aws.amazon.com/eks/latest/userguide/auto-cluster-iam-role.html#tag-prop).

**Why the built-in NodePools have to go.** The built-in `default` NodeClass is reconciled by EKS and cannot carry user tags, and the built-in `system` / `general-purpose` NodePools are hard-wired to it. With `--tags`, the script creates a tagged NodeClass plus a small fallback NodePool (weight 1, capped at 16 vCPU / 64Gi so it never outranks the chart's `general-work-pool`), asserts an `EC2`-type access entry with `AmazonEKSAutoNodePolicy` for the node role, then sets `computeConfig.nodePools: []`. Note that `nodeRoleArn` must be omitted from that call — EKS rejects it alongside an empty pool list, which is [what the docs prescribe](https://docs.aws.amazon.com/eks/latest/userguide/create-node-pool.html#_cluster_without_built_in_node_pools) for a cluster with no built-in pools.

`nodeClassRef` is parameterised across the migration assistant, `testClusters` and `buildImages` charts so all of them follow the tagged NodeClass; `default` remains the default everywhere, so **behaviour without `--tags` is unchanged** — verified by rendering every chart from `main` and from this branch and diffing (byte-identical for chart defaults, `valuesForLocalK8s`, `valuesGke`, `valuesTraceJaeger`, and `valuesEks` without tags). The non-EKS solution stack snapshot is untouched.

`--tags` conflicts with `--use-general-node-pool` (which it deletes) → hard error. `--disable-general-purpose-pool` is redundant → downgraded to a note.

### Verifying it, and proving the verification isn't vacuous

Two optional layers, both off by default:

**`--enforce-tags-on-create-for-tests`** attaches a Deny to the cluster role for every create proven to carry the tags, so an untagged create is *refused at the moment it happens* rather than discovered later — reproducing a deployer policy that requires tags on create. Six action groups × one statement per tag key, because IAM AND-s multiple context keys inside one condition operator, so listing every key in one `Null` block would only deny a create omitting *all* of them. Each action is scoped to only the resource types that receive its request tags; a shared `Resource` list breaks `RunInstances`, which also authorizes against `launch-template/*` and friends that never receive tags. The statements are attached as one test-only inline policy. IAM's 10,240-character quota is aggregate across all inline policies on a role, so a sufficiently large test tag set can exceed the remaining capacity and is rejected; CI deliberately uses two tags.

**`--verify-resource-tags`** (on the test invocation the EKS CDC pipelines already make) runs `resource_tag_verifier.py`, which never searches *by* tag — an untagged resource is exactly the one a tag search cannot find, and scanning a region only works if the region is empty and stays empty. Three oracles that know a resource exists independently of its tags:

- **CloudFormation** — `ListStackResources` plus the tagging API, filtered on the CFN-injected `aws:cloudformation:stack-id`.
- **The Kubernetes API** — `Node.spec.providerID` → instances → their volumes and ENIs; `PersistentVolume.spec.csi.volumeHandle` → PVC volumes.
- **CloudTrail** — the catch-all, scoped to the cluster IAM role, checking *calls* rather than resources so it catches a create whose resource type nobody enumerated.

Expected tags are read from the stack's own tags, so nothing about them is configured in CI and the check cannot drift from the deployment it checks. The report states how each result was established. It only says "an untagged create would have been REFUSED" after reading the cluster role and confirming the complete test-only policy is attached; otherwise CloudTrail results are labeled observational. A unit test also keeps the expected action set aligned with `aws-bootstrap.sh`.

### Issues Resolved

None filed; reported by a customer who could not launch any EKS Auto Mode nodes because a tag-enforcing policy rejected the untagged `RunInstances` calls.

### Testing

**Live, on EKS.** `eks-cdc-full-e2e-test` and `eks-cdc-aoss-e2e-test` are wired to deploy with two tags (one static, one stage-derived — a single tag cannot catch a bug that keeps only the first entry of a comma-separated list), with enforcement on, and to verify afterwards. Both pass. From the full-e2e run, 63 resources checked:

```
Enforced by the TEST-ONLY IAM policy -- an untagged create would have been REFUSED, not just noticed:
  ec2:RunInstances                          (7 create(s) observed)
  ec2:CreateVolume                          (19 create(s) observed)
  elasticloadbalancing:CreateLoadBalancer   (5 create(s) observed)
  elasticloadbalancing:CreateTargetGroup    (5 create(s) observed)
  elasticloadbalancing:CreateListener       (5 create(s) observed)
PASSED: every resource checked carries the expected tags
```

with all five migration tests (`0031, 0032, 0033, 0040, 0042`) green on a cluster where both built-in NodePools are disabled and all compute comes from the tagged NodeClass.

**Unit.** 10 CDK, 210 orchestrationSpecs (14 snapshots), 103 Python. Notable coverage: every resource type in both EKS templates must appear in a hand-kept tag-coverage map, so adding a type fails the build until someone classifies it; the `AutoModeTagPropagationPolicy` scoping cannot be widened or dropped unnoticed; `test_runner.py` must contain no relative imports (it is run as a script).

**Rendering and parsing.** `helm template` across all charts as above; the generated NodeClass/NodePool manifest asserted field-by-field and re-checked with a single subnet and a single tag; the `--tags` parser exercised against 9 cases including `k=a=b`, empty values, blank/trailing commas and the reserved prefixes. That last set caught a real bug in development: `printf '%s' | tr` dropped the last tag of every value because `read` never saw a terminating newline, so `--tags A=1,B=2` yielded only `A=1`. `bash -n` clean under bash 3.2 (macOS), and no bash-4-only constructs.

### Findings worth knowing before writing a tag-enforcing policy

These came out of getting the above green and are the practically useful part for anyone in the same position:

- **An SCP that denies `ec2:RunInstances` on `aws:RequestTag` with `Resource: "*"` blocks EKS Auto Mode even when the instances are tagged correctly**, because that call also authorizes against the launch template, subnet, security group and AMI, none of which receive request tags. It must be scoped to `instance/*`, `volume/*`, `network-interface/*`.
- **`ec2:CreateFleet` and `ec2:CreateLaunchTemplate` carry no user tags at all** — the instances they produce inherit tags from the launch template's `launchTemplateData`. Requiring request tags on either blocks Auto Mode outright.
- **The load balancing controller's shared backend `k8s-traffic-*` security group cannot be tagged.** Per-Service annotations reach the frontend SG only, so `ec2:security-group` is out of reach for a per-Service tagging scheme.
- **An AWS Organizations tag policy does not enforce missing tags** — "Basic compliance rules do not enforce tag compliance on resources that are created without tags." It only rejects non-compliant *values*, and reports `TagPolicyViolation` rather than `AccessDenied`. Mandatory-tag enforcement needs an SCP.

### Not included

The newer Rust `migration-assistant` CLI is a full reimplementation of the bootstrap flow rather than a wrapper around `aws-bootstrap.sh`, so `--tags` is not supported there yet. It's listed in that README's set of not-yet-implemented flags so it is warned about rather than silently ignored. Note that the CLI does `helm upgrade --install`, so StorageClass `parameters` immutability will need handling there when `--tags` lands — `aws-bootstrap.sh` is install-only and hard-stops on an existing release, so it cannot hit that.

### Check List


jzonthemtn:
Check how this handles tags from an existing stack. It looks like inherited tags skip validation. I don’t think I see a test covering a run of `aws-bootstrap.sh` without `--tags` against an already-tagged stack (to test with inherited tags).

Also, quotes in tag values, such as `a"b`, will produce invalid YAML.

gregschohn:
> Check how this handles tags from an existing stack. It looks like inherited tags skip validation. I don’t think I see a test covering a run of `aws-bootstrap.sh` without `--tags` against an already-tagged stack (to test with inherited tags).
> 
> Also, quotes in tag values, such as `a"b`, will produce invalid YAML.

Thanks.  I think that I've addressed each of those in the latest commit.

jzonthemtn (approved):

PR #3341, opened 2026-08-28
Add AWS Terraform deployment support

tonypiazza:
## Description

Adds a Terraform module at `deployment/terraform/aws/` that provisions the AWS EKS infrastructure for the Migration Assistant, as an alternative to the existing CDK-based AWS Solution. It mirrors the conventions of the existing GCP Terraform module and is meant to become the recommended path for customers while the CDK deployment stays supported.

The change stands alongside what already exists rather than replacing it: a new directory plus a few doc-pointer updates. It leaves the CDK solution, the Helm chart, and all runtime behavior untouched.

## What it provisions

- **Networking:** a dual-stack VPC with public/private subnet pairs across two AZs, NAT + internet/egress-only gateways, and the S3 gateway + interface VPC endpoints. Can attach to an existing VPC instead (`create_vpc = false`), with opt-in endpoints.
- **EKS:** an EKS Auto Mode cluster (`system` + `general-purpose` node pools), cluster/node roles, and an account read-only access entry.
- **IAM + identity:** EKS Pod Identity associations for the Migration Assistant service accounts, a workload role, and the S3 snapshot role assumed by Amazon OpenSearch Service.
- **ECR:** a private repository for mirrored or locally built images.
- **Optional Helm install:** when `deploy_helm = true`, installs the repo's chart with the EKS values overlay.

## Notable capabilities

- **Isolated / air-gapped mode** (`isolated = true`): no NAT gateway and the full service-endpoint set (`s3, ecr.api, ecr.dkr, logs, monitoring, elasticfilesystem, sts, eks-auth`) so a migration runs with no public data path. This matches the endpoint set used by the existing isolated-VPC deployment path.
- **Per-leg private connectivity** (`source_connectivity` / `target_connectivity`): `privatelink` (interface VPC endpoint to a provider's endpoint service, with optional Route 53 private DNS) or `vpc_peering`, each defaulting to `none`.
- **Least-privilege IAM**: scoped to module-owned resources where the ARNs are knowable (the ECR repo, the MA log group for the stage/region, `MA-*` dashboards, and `migrations-*` for the snapshot role). Actions that cannot be resource-scoped are split into their own statements and annotated inline with why. The workload S3 statement is deliberately left on `*` because the console reads/writes user-configured buckets (notably the failed-document-stream bucket) and the default bucket is created by the Helm chart, not Terraform. Accepting explicit bucket ARNs to scope that down is a noted follow-up.
- **Optional permissions boundary** (`permissions_boundary_arn`) applied to every created role, for accounts that require one.

## Non-regression

All new behavior is opt-in and defaults to today's values, so a standard deployment is unaffected. The isolated flag, connectivity modes, snapshot-role toggle, and permissions boundary all default off/none. The module does not touch the CDK deployment or its resources.

## Testing

Plan-time tests run with mocked AWS/Helm providers (no cloud credentials):

- `terraform fmt -check -recursive`: clean
- `terraform validate`: success
- `terraform test`: 17/17 passing (new-VPC topology, isolated topology, existing-VPC additive behavior, EKS Auto Mode + Pod Identity, per-leg connectivity, snapshot-role toggle, Helm version precondition)

Applied to a clean AWS account with default configuration, then again with `deploy_helm = true` and `migration_assistant_version = 3.3.4`: 58 and 59 resources respectively, applying and destroying cleanly. `kubernetes_version = 1.35` (the module default) is a real EKS version in us-east-1, which had only ever been exercised against mocks. EKS Auto Mode cold-started 3 nodes and the chart came up 24/24 pods with 0 restarts. Pod Identity works end to end: `aws sts get-caller-identity` from the console pod returns the expected assumed role. The ECR repo-ARN scoping is tight in both directions, with the named lookup succeeding and the enumerating form denied.

That run surfaced a teardown bug, fixed in e4611913b: `helm_release` declared no dependency on networking, so `terraform destroy` tore down the VPC endpoints while `helm uninstall` was still running, which failed the chart's credential-dependent uninstall hooks and aborted the destroy partway. Destroy now completes in a single pass.

Separately validated on real AWS by another tester: an end-to-end backfill migration over private networking succeeded. Live capture-and-replay has not yet been exercised on this module.

Not yet exercised: private ECR pulls (all chart images come from `public.ecr.aws`, which needs no IAM), CloudWatch logging, whether the scoped IAM statements suffice for a full migration, and `isolated = true`, the connectivity modes, and `permissions_boundary_arn` against real AWS.

## Docs

The module README covers all variables, isolated mode, private connectivity (including the operator responsibilities the module intentionally does not automate: endpoint-service acceptance, reciprocal peering routes, TLS hostname matching), outputs, and CDK coexistence. Top-level deployment docs are updated to point at both the Terraform and CDK paths.

### DIFF
--- a/deployment/terraform/aws/iam.tf
+++ b/deployment/terraform/aws/iam.tf
@@ -1,0 +1,281 @@
+resource "aws_ecr_repository" "migration" {
+  name         = local.ecr_name
+  force_delete = true
+
+  image_scanning_configuration {
+    scan_on_push = true
+  }
+
+  tags = local.common_tags
+}
+
+data "aws_iam_policy_document" "pod_identity_assume_role" {
+  statement {
+    effect  = "Allow"
+    actions = ["sts:AssumeRole", "sts:TagSession"]
+
+    principals {
+      type        = "Service"
+      identifiers = ["pods.eks.amazonaws.com"]
+    }
+  }
+}
+
+resource "aws_iam_role" "migration_pods" {
+  name                 = "ma-${var.stage}-${var.region}-migrations"
+  description          = "Migration Assistant role assumed by workloads through EKS Pod Identity"
+  assume_role_policy   = data.aws_iam_policy_document.pod_identity_assume_role.json
+  permissions_boundary = var.permissions_boundary_arn
+
+  tags = local.common_tags
+}
+
+data "aws_iam_policy_document" "migration_pods" {
+  # ecr:GetAuthorizationToken is an account-level action that does not support
+  # resource scoping; it must be granted on "*".
+  statement {
+    sid       = "EcrAuth"
+    effect    = "Allow"
+    actions   = ["ecr:GetAuthorizationToken"]
+    resources = ["*"]
+  }
+
+  # Pull and push are scoped to this module's ECR repository (mirrored/built MA
+  # images). Replaces the previous AmazonEC2ContainerRegistryFullAccess managed
+  # policy, which granted these on every repo in the account.
+  statement {
+    sid    = "EcrImages"
+    effect = "Allow"
+    actions = [
+      "ecr:BatchGetImage",
+      "ecr:GetDownloadUrlForLayer",
+      "ecr:DescribeRepositories",
+      "ecr:BatchCheckLayerAvailability",
+      "ecr:CompleteLayerUpload",
+      "ecr:InitiateLayerUpload",
+      "ecr:PutImage",
+      "ecr:UploadLayerPart",
+    ]
+    resources = [aws_ecr_repository.migration.arn]
+  }
+
+  # EFS filesystems are created at runtime (by the app / EKS), so their ARNs are
+  # not known here; left on "*".
+  statement {
+    sid       = "ElasticFileSystem"
+    effect    = "Allow"
+    actions   = ["elasticfilesystem:ClientMount", "elasticfilesystem:ClientWrite"]
+    resources = ["*"]
+  }
+
+  # The source/target clusters are supplied as runtime migration config; their ARNs
+  # are not known to Terraform, so es:/aoss: access is granted on "*".
+  statement {
+    sid       = "OpenSearch"
+    effect    = "Allow"
+    actions   = ["es:ESHttp*", "aoss:APIAccessAll"]
+    resources = ["*"]
+  }
+
+  # secretsmanager:ListSecrets is account-level and cannot be scoped. Get/Describe
+  # target user-supplied secrets whose names are not known here, so also on "*".
+  statement {
+    sid    = "Secrets"
+    effect = "Allow"
+    actions = [
+      "secretsmanager:GetSecretValue",
+      "secretsmanager:DescribeSecret",
+      "secretsmanager:ListSecrets",
+    ]
+    resources = ["*"]
+  }
+
+  # s3:ListAllMyBuckets is an account-level action that cannot be resource-scoped;
+  # it only functions on "*".
+  statement {
+    sid       = "S3ListAllBuckets"
+    effect    = "Allow"
+    actions   = ["s3:ListAllMyBuckets"]
+    resources = ["*"]
+  }
+
+  # Object and bucket operations remain on "*" rather than scoped to migrations-*.
+  # The migration console reads/writes user-configured buckets whose names are not
+  # under our control, notably the failed-document-stream bucket, which is supplied
+  # per-migration and may use any name. The bucket the module itself relies on is
+  # created/deleted by the Helm chart's defaultBucketConfiguration (app-managed,
+  # named migrations-default-<account>-<stage>-<region>), not by Terraform. This
+  # matches the CloudFormation deployment. A future enhancement can accept explicit
+  # bucket ARNs (e.g. from the installer) to scope this down.
+  statement {
+    sid    = "Snapshots"
+    effect = "Allow"
+    actions = [
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:ListBucket",
+      "s3:DeleteObject",
+      "s3:DeleteObjectVersion",
+      "s3:ListBucketVersions",
+      "s3:ListBucketMultipartUploads",
+      "s3:AbortMultipartUpload",
+      "s3:CreateBucket",
+      "s3:DeleteBucket",
+    ]
+    resources = ["*"]
+  }
+
+  # logs:DescribeLogGroups only supports "*" (it enumerates across groups); the
+  # rest are scoped to the Migration Assistant log group for this stage/region.
+  statement {
+    sid       = "CloudWatchLogsDescribe"
+    effect    = "Allow"
+    actions   = ["logs:DescribeLogGroups"]
+    resources = ["*"]
+  }
+
+  statement {
+    sid    = "CloudWatchLogs"
+    effect = "Allow"
+    actions = [
+      "logs:PutLogEvents",
+      "logs:DescribeLogStreams",
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+    ]
+    resources = [
+      "arn:${data.aws_partition.current.partition}:logs:${var.region}:${data.aws_caller_identity.current.account_id}:log-group:/migration-assistant-${var.stage}-${var.region}*",
+    ]
+  }
+
+  # cloudwatch:ListMetrics/GetMetricData do not support resource-level scoping.
+  statement {
+    sid       = "CloudWatchMetrics"
+    effect    = "Allow"
+    actions   = ["cloudwatch:ListMetrics", "cloudwatch:GetMetricData"]
+    resources = ["*"]
+  }
+
+  # X-Ray Put* actions do not support resource-level scoping.
+  statement {
+    sid       = "XRayTraces"
+    effect    = "Allow"
+    actions   = ["xray:PutTraceSegments", "xray:PutTelemetryRecords"]
+    resources = ["*"]
+  }
+
+  statement {
+    sid       = "PassSnapshotRoles"
+    effect    = "Allow"
+    actions   = ["iam:PassRole"]
+    resources = ["arn:${data.aws_partition.current.partition}:iam::${data.aws_caller_identity.current.account_id}:role/*"]
+  }
+
+  statement {
+    sid    = "MigrationDashboards"
+    effect = "Allow"
+    actions = [
+      "cloudwatch:PutDashboard",
+      "cloudwatch:GetDashboard",
+      "cloudwatch:DeleteDashboards",
+    ]
+    resources = ["arn:${data.aws_partition.current.partition}:cloudwatch::${data.aws_caller_identity.current.account_id}:dashboard/MA-*"]
+  }
+
+  # The private CA is created at runtime (by the app) for the capture-proxy TLS
+  # path; its ARN is not known here, so acm-pca actions are granted on "*".
+  statement {
+    sid    = "PrivateCertificateAuthority"
+    effect = "Allow"
+    actions = [
+      "acm-pca:IssueCertificate",
+      "acm-pca:GetCertificate",
+      "acm-pca:DescribeCertificateAuthority",
+      "acm-pca:ListCertificateAuthorities",
+      "acm-pca:CreateCertificateAuthority",
+      "acm-pca:DeleteCertificateAuthority",
+      "acm-pca:UpdateCertificateAuthority",
+      "acm-pca:TagCertificateAuthority",
+    ]
+    resources = ["*"]
+  }
+}
+
+resource "aws_iam_role_policy" "migration_pods" {
+  name   = "MigrationsPodPolicy"
+  role   = aws_iam_role.migration_pods.id
+  policy = data.aws_iam_policy_document.migration_pods.json
+}
+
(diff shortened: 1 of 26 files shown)


deployment/terraform/aws/variables.tf
```diff
@@ -0,0 +1,270 @@
...
+    snapshots (trusted principal es.amazonaws.com). Default true preserves current
+    behavior. Set false when the migration SOURCE is not Amazon OpenSearch Service
+    (e.g. self-managed Elasticsearch/OpenSearch), which registers its S3 snapshot
+    repository without assuming an AWS role, so this role is unused.
+  EOT
+  type        = bool
+  default     = true
+}
+
+variable "isolated" {
```
jzonthemtn, line 54: What's the relationship between `isolated` and `cluster_endpoint_public_access`?  I think the way the comments are written, I would expect `isolated=true` to override `cluster_endpoint_public_access`.

deployment/terraform/aws/variables.tf
```diff
@@ -0,0 +1,270 @@
...
+
+  validation {
+    condition     = var.source_connectivity.mode != "vpc_peering" || (try(var.source_connectivity.peer_vpc_id, null) != null && try(var.source_connectivity.peer_cidr, null) != null)
+    error_message = "source_connectivity.peer_vpc_id and peer_cidr are required when mode = vpc_peering."
+  }
+}
+
+variable "target_connectivity" {
+  description = <<-EOT
+    Private connectivity for the target-write leg. Same modes as source_connectivity.
```
jzonthemtn, line 170: What do you think about narrowing the description to providers that expose a generic VPC endpoint service name?

AOS supports PrivateLink, but its managed domain endpoints require `aws_opensearch_vpc_endpoint` instead of the `aws_vpc_endpoint` used in this PR. "Almost always" suggests the PR supports AOS domains directly, which it currently doesn’t even if `mode=privatelink`.

I suppose there's the question of changing to `aws_opensearch_vpc_endpoint` instead, or offering it as an alternative.

deployment/terraform/aws/iam.tf
```diff
@@ -0,0 +1,281 @@
...
+    sid    = "CloudWatchLogs"
+    effect = "Allow"
+    actions = [
+      "logs:PutLogEvents",
+      "logs:DescribeLogStreams",
+      "logs:CreateLogGroup",
+      "logs:CreateLogStream",
+    ]
+    resources = [
+      "arn:${data.aws_partition.current.partition}:logs:${var.region}:${data.aws_caller_identity.current.account_id}:log-group:/migration-assistant-${var.stage}-${var.region}*",
```
jzonthemtn, line 146: I think this will block some metrics from EKS from getting to CloudWatch. 

```
"arn:${data.aws_partition.current.partition}:logs:${var.region}:${data.aws_caller_identity.current.account_id}:log-group:/metrics/default",
"arn:${data.aws_partition.current.partition}:logs:${var.region}:${data.aws_caller_identity.current.account_id}:log-group:/metrics/default:*",
```

jzonthemtn:
Looks good. Just going through it looks like it has parity with the CloudFormation templates. The comments I made were minor, other than the logging one which could be helpful.

I did not stand it up, but if that would be valuable let me know.

deployment/terraform/aws/iam.tf
```diff
@@ -0,0 +1,281 @@
...
+      "s3:GetObject",
+      "s3:PutObject",
+      "s3:ListBucket",
+      "s3:DeleteObject",
+      "s3:DeleteObjectVersion",
+      "s3:ListBucketVersions",
+      "s3:ListBucketMultipartUploads",
+      "s3:AbortMultipartUpload",
+      "s3:CreateBucket",
+      "s3:DeleteBucket",
```
jzonthemtn, line 122: The above comment mentions it, but probably a good idea to write up a new issue to address this in both the CF and TF at some point.

deployment/terraform/aws/iam.tf
```diff
@@ -0,0 +1,281 @@
...
+    resources = ["*"]
+  }
+
+  # secretsmanager:ListSecrets is account-level and cannot be scoped. Get/Describe
+  # target user-supplied secrets whose names are not known here, so also on "*".
+  statement {
+    sid    = "Secrets"
+    effect = "Allow"
+    actions = [
+      "secretsmanager:GetSecretValue",
```
jzonthemtn: Wondering if we could do something here to scope this just to certain secrets. Require the secrets have some prefix maybe that could then be scoped on?

I didn't check to see if CF does the same thing - if so probably like the previous comment, a follow-up issue to address both.

deployment/terraform/aws/variables.tf
tonypiazza: Good catch, and you found a real bug rather than just unclear wording. `isolated` only touched the NAT gateways, the private egress routes, and the endpoint set. It never reached `vpc_config`, so `isolated = true` gave you a cluster with no data-path egress and a Kubernetes API endpoint still reachable from `0.0.0.0/0`. The description promising a "fully private migration" made it worse by implying otherwise.

Fixed in 29c388738. `cluster_endpoint_public_access` now defaults to `null` and resolves in locals:

```hcl
endpoint_public_access = var.cluster_endpoint_public_access != null ? var.cluster_endpoint_public_access : !var.isolated
```

Unset follows `isolated`; an explicit `true` or `false` always wins.

On your expectation that `isolated` should override outright: I went with "isolated sets the default, explicit value wins" instead, for two reasons. Terraform can't distinguish an explicitly-set `true` from a defaulted one, so a hard override would mean silently discarding a value someone deliberately set. And the override has a real use case, since a private-only API endpoint means `kubectl` has to run from inside the VPC, and so does `terraform apply` itself when `deploy_helm = true`, because the Helm provider talks to the cluster endpoint directly. Being able to keep the API reachable while the data path stays air-gapped is worth preserving. The override affects only the control plane; NAT, routes, and the endpoint set are unchanged.

It keys off `var.isolated` rather than `create_vpc && var.isolated`, since the control plane is exposed the same way in a supplied VPC.

Worth flagging: for anyone already running `isolated = true`, this changes cluster reachability on the next apply. That's the intent, but it's a behavior change and not only a docs fix.

Three new run blocks cover it (`isolated_public_endpoint_override_is_honored`, `public_endpoint_can_be_disabled_without_isolated`, `both_endpoints_disabled_is_rejected`), alongside the existing isolated and default runs. I mutation-tested them by reverting the wiring, dropping the `isolated` coupling, and dropping the override branch, to confirm each assertion actually fails. The README documents the new default and the in-VPC access requirement.

This is plan-verified, not apply-verified. The module's tests are plan-time against mocked providers, and I have not applied `isolated = true` to a live account, so "EKS accepts `endpoint_public_access = false` alongside the default `public_access_cidrs`" rests on the API reference and on the provider sending those two fields independently, not on an observed apply. I tried to close that today and got blocked on account access. Flagging it rather than implying more coverage than exists.

deployment/terraform/aws/variables.tf
tonypiazza: Agreed, and thanks for being precise about why. `aws_opensearch_vpc_endpoint` takes
`domain_arn` plus `vpc_options`, while this module creates a generic `aws_vpc_endpoint`
from a `vpc_endpoint_service_name`. "Almost always" claimed something the code doesn't do.

Narrowed in b405cbee. Both descriptions and the README now use the positive form you
suggested: `privatelink` requires the provider to publish a VPC endpoint service name,
full stop. No managed-service framing in either direction. I also added the case the old
wording obscured, which I think is the one that would actually have saved you the
question: a cluster already inside the migration VPC needs no connectivity resources and
should leave `mode = "none"`.

On adding `aws_opensearch_vpc_endpoint`, I dug into this and came out less convinced than
when I read your comment. Two things:

When the CDK provisions the target domain it places it in the migration VPC
(`opensearch-domain-stack.ts:204-205`, `vpc` / `vpcSubnets`), so there's no endpoint
involved and the Terraform equivalent is `mode = "none"`. When you supply a pre-existing
`targetCluster`, the CDK provisions no private path at all. So an AOS VPC endpoint isn't a
CFN-parity gap in either direction. It'd be new capability, for an AOS domain in a VPC
that is neither the migration VPC nor peerable.

I also want to be careful about what belongs in this module. It deliberately knows nothing
about what the source and target are; they're runtime migration config. An `aos_domain`
mode would be the only place it encodes a specific managed service, and I'd rather not
open that door on speculation. It's also the shape our tests can't reach: the part that
decides whether it works is the cross-account `aws_opensearch_authorize_vpc_endpoint_access`
handshake, and these are plan-time tests against mocked providers.

So I'd rather not build it blind, and I'd leave it out for now.

Correcting myself on one point in case I said it elsewhere: I'd assumed
`aws_opensearch_vpc_endpoint` was the provider-side resource. It isn't.
`vpc_options.subnet_ids` are the consumer's and `vpc_id` is computed from them, so it's
consumer-side like ours. The provider-side counterpart is
`aws_opensearch_authorize_vpc_endpoint_access`.

deployment/terraform/aws/iam.tf
tonypiazza: You're right, and it's worse than blocked metrics: it blocked all of them. Fixed in 3acd5deea with exactly the two ARNs you wrote.

The mechanism, for the record. The `awsemf` exporter defaults `log_group_name` to `/metrics/default` (stream `otel-stream`) when it isn't set, and neither exporter in `valuesEks.yaml` sets it: `awsemf/app` and `awsemf/cadvisor` configure only namespace, dimension rollup, and metric declarations. The one `log_group_name` in the chart is on the fluent-bit logs pipeline, not the EMF exporters. So both `metrics/app` and `metrics/cadvisor` wrote to a group the scoped ARN didn't match, and `otel-collector` is in `pod_identity_service_accounts`, so the collector holds this role directly. `CreateLogGroup` and `PutLogEvents` denied, nothing reaches CloudWatch.

This regressed when I scoped the policy down. It was `"*"` before that.

I used your two-entry form rather than a `/metrics/default*` wildcard, since the wildcard would also match unrelated groups sharing the prefix.

Also added a test asserting the statement covers both that group and the migration-assistant group. Mutation-verified: removing either metrics ARN fails the first assertion, removing the migration-assistant ARN fails the second, so neither half can rot silently.

I considered setting `log_group_name` explicitly on both exporters instead, which would consolidate the groups, but that edits the shared chart overlay the CFN EKS path also consumes. Wider blast radius than an IAM fix warrants.

One gap you didn't mention: CFN grants `logs:PutRetentionPolicy` (`common-utilities.ts:211`) and the TF policy doesn't. `awsemf` only calls it when `log_retention` is configured and nothing in the chart sets that, so I left it out. Happy to add it for parity if you'd rather.

deployment/terraform/aws/iam.tf
tonypiazza: Agreed. Opened #3362 covering both paths.

Worth noting which direction the gap runs: CFN grants `s3:*` on `"*"` (`migration-console-stack.ts:145-151`, `reindex-from-snapshot-stack.ts:66-72`), while TF grants 10 enumerated actions on `"*"`. Same resource scope, narrower actions on the TF side, so resource scoping is the shared work.

The honest blocker, which is in the issue: the failed-document-stream bucket is supplied per-migration and can be named anything, so no static ARN list covers it. Real scoping means accepting explicit bucket ARNs as deployment config on both paths.

One detail that makes the CFN side more tractable than it looks: `migration-console-stack.ts:141-144` already resolves `ARTIFACT_S3_ARN` from SSM and then grants `"*"` anyway.

deployment/terraform/aws/iam.tf
tonypiazza: Good question, and checking CF changed my answer. **CF is tighter than TF here**, not the same. `getSecretAccessPolicy` (`common-utilities.ts:27-35`) scopes to the exact secret ARN and is used in three places (`migration-console-stack.ts:185,188`, `reindex-from-snapshot-stack.ts:125`). Only the traffic replayer falls back to `"*"`. So TF was the loosest of the three, and it also granted `ListSecrets`, which CF never does outside the EKS solution stack.

Two of those three actions turned out to have no caller at all. The only AWS Secrets Manager API call in the repo is `client.get_secret_value` at `console_link/models/cluster.py:275`. Nothing describes or enumerates: the `list_secrets` in `workflow/models/secret_store.py` is Kubernetes-backed (`list_namespaced_secret`, label selector) and never reaches AWS, and there's no Secrets Manager usage in the Java sources.

So 3f39e094f drops `DescribeSecret` and `ListSecrets`, leaving `GetSecretValue`, with a test pinning the statement to that one action (mutation-verified against both re-adding `ListSecrets` and swapping in `DescribeSecret`).

Caveat: I only grepped this repo, and the tests are plan-time, so a runtime denial wouldn't show up in CI.

On your prefix idea: that's the right shape and it's #3363. `GetSecretValue` stays on `"*"` for now because the secret ARN is runtime migration config the module never sees. Notably `cluster.py:275` fetches by an ARN it's handed and never discovers secrets, so prefix scoping would work fine. It needs a new module input plus a documented naming requirement, hence the separate issue. #3363 also flags that `eks-infra.ts:317-319` grants the same two unused actions, worth confirming against a real deployment before touching.

jzonthemtn (commented):
+1 lgtm!

jzonthemtn (approved):

PR #3352, opened 2026-09-02
Rework Kafka Record Management in TrafficReplayer to Support Horizontal Scaling

gregschohn:
## Review scope

This PR is currently a design review. The design documents below are the source of truth.

Please ignore every other change in the PR for now. The implementation is still in progress and
should not be reviewed as conforming to the design yet.

## Recommended reading order

1. [Capture and Replay Architecture](https://github.com/gregschohn/opensearch-migrations/blob/integrating3231/docs/captureAndReplayArchitecture.md)

   Start here. This is the top-level contract shared by the capture proxy, Kafka, and the replayer.
   It defines the protocol records, ordering and timing guarantees, failure behavior, expiration,
   target replay, tuple durability, Kafka commit rules, and the arguments for protocol soundness.

2. [Proxy Capture Protocol](https://github.com/gregschohn/opensearch-migrations/blob/integrating3231/docs/proxyCaptureProtocol.md)

   This refines the proxy side of the top-level contract. It covers Kafka group assignment,
   connection routing, heartbeats, capture-before-forward guarantees, bounded connection-record
   publication, per-connection publication ordering, shutdown, capture compromise, and proxy
   verification requirements.

3. [Replayer Processing and Commit Architecture](https://github.com/gregschohn/opensearch-migrations/blob/integrating3231/docs/replayerProcessingAndCommitArchitecture.md)

   This refines the replayer side. It describes the mutable-state owners, per-connection request
   ordering, target replay lifecycle, captured response handling, tuple output, whole-record
   accounting, per-partition backpressure, cancellation, reassignment, and Kafka commit behavior.

4. [Managed Fleet Capture Recovery](https://github.com/gregschohn/opensearch-migrations/blob/integrating3231/docs/managedFleetCaptureRecovery.md)

   Read this for Kubernetes and controller-managed deployments. It defines fleet-level readiness,
   replacement, capture-compromise handling, clock-health requirements, shutdown timing, and the
   boundaries that future controller automation must preserve.  This will not be implemented now
   but has been included to keep the rest of the designs honest.

5. [Bring-Your-Own Captured Traffic](https://github.com/gregschohn/opensearch-migrations/blob/integrating3231/docs/BringYourOwnCapturedTraffic.md)

   Read this when exporting or importing captured traffic. It defines the exact-fidelity archive
   format, timestamp modes, finalized-versus-range archive behavior, import orchestration, and the
   evidence required for imported traffic to follow the same replay protocol.  We do have users
   use kafka dumps for diagnostic purposes, so keeping this usable is valuable.

The top-level document should be reviewed first. The remaining documents can then be reviewed
independently according to the component or deployment mode of interest.


TrafficCapture/trafficReplayer/src/main/java/org/opensearch/migrations/replay/lifecycle/ReplayFatalFence.java
```diff
@@ -0,0 +1,39 @@
+package org.opensearch.migrations.replay.lifecycle;
+
+import java.util.Objects;
+
+/**
+ * Process-wide, one-way indication that the replayer can no longer safely accept commit work.
```
gregschohn: After fatal event-loop death is detected, the replayer emits best-effort diagnostics and immediately halts. Any concurrent external operation may or may not complete. Kafka’s committed offset determines the durable outcome after restart.

Remove ReplayFatalFence and its fatal-path commit coordination. It is an incomplete shutdown barrier and provides no external fencing guarantee. Retain only idempotent fatal detection, best-effort logs/metrics, and Runtime.halt(). Do not initiate cleanup or completion because of event-loop death; allow already-running commits and other operations to race naturally with process termination.

docs/captureAndReplayArchitecture.md
```diff
@@ -0,0 +1,1513 @@
...
+Every local connection registry, publisher lane, and broker-time baseline is keyed by
+`(writerNodeId, partition)`, not by one process-global current writer identity.
+
+### 2.2 Protocol records
+
+Every Kafka application-record value is one `CaptureRecord` protobuf envelope. Its `payload`
+`oneof` identifies exactly one of these payloads:
+
+- `TrafficStream` contains one or more `TrafficObservation` values for one connection.
+- `TrafficObservation` records captured network or connection-lifecycle activity. Each observation
```
jzonthemtn, line 269: Is `TrafficObservation` not inside of `TrafficStream` per line 268?

docs/captureAndReplayArchitecture.md
```diff
@@ -0,0 +1,1513 @@
...
+partition values that could disagree. The current archive import defined in §2.3 preserves the
+source partition layout.
+
+A Kafka topic must never contain records written under mutually wire-incompatible capture-protocol
+formats. Reusing an existing topic for a wire-incompatible format is explicitly prohibited, even
+after the old consumer group is empty or the deployment has entered a no-capture interval. A
+wire-incompatible format requires a new Kafka topic, and the complete capture and replay cohort
+using that format must use the new topic.
+
+### 2.3 Imported capture archives
```
jzonthemtn, line 302: I think this is an important part, but is it needed for a first implementation? Would would be the cost of leaving this out of these initial docs and designing/implementing bring-your-own later on? I'm asking purely from the perspective of simplification and risk-reduction on the first implementation.

docs/proxyCaptureProtocol.md
```diff
@@ -0,0 +1,1528 @@
...
+to Kafka for later replay. It covers the controller-less deployment. Managed-fleet terminal-failure
+behavior and fresh-run recovery are specified separately in
+[`managedFleetCaptureRecovery.md`](managedFleetCaptureRecovery.md).
+
+The design deliberately separates three concerns:
+
+1. Kafka group membership load-balances partitions for new connections.
+2. Periodic records from each writer and partition establish the broker-time baseline used to
+   expire incomplete state after a proxy becomes silent.
+3. The capture-before-forward contract prevents a strict-mode proxy from completing a source
```
jzonthemtn, line 16: I didn't find what `strict-mode` means, but I only looked "up" in the document preceding this one. It's used a few times in this doc but wasn't defined.

docs/proxyCaptureProtocol.md
```diff
@@ -0,0 +1,1528 @@
...
+and later traffic follow the configured one-way pass-through behavior immediately.
+
+In a controller-managed deployment, capture compromise and permission to begin pass-through are
+separate transitions:
+
+1. The proxy enters `COMPROMISED_AWAITING_CONTROLLER_ACK`. Capture remains permanently closed, and
+   no source-bound traffic that has not already passed the capture-before-forward gate may be
+   forwarded.
+2. The proxy immediately reports the compromise on an authenticated control path. The report
+   identifies the exact resource, process, Pod, capture activation, Kafka topic, and controller fence.
```
jzonthemtn, line 999: I'm not sure what "controller fence" is. Is it a Kafka thing I'm not familiar with, or something specific to this design?

docs/proxyCaptureProtocol.md
```diff
@@ -0,0 +1,1528 @@
...
+force every ordinary packet and connection lifecycle change through one global proxy lock.
+
+Each connection's event-loop owner also controls the lifetime of its current `TrafficStream`:
+
+1. Appending the first observation to a new record schedules one monotonic deadline at `F`.
+2. Before beginning every later `TrafficObservation`, the owner compares the local monotonic time
+   with that fixed deadline. If the deadline has passed, it detaches the old record before admitting
+   the new observation. A delayed timer therefore cannot place new activity into an expired record.
+3. Appending later observations before the deadline does not move or replace that deadline.
+4. If the record reaches its size limit, a complete Critical Mutation Traffic request requires
```
jzonthemtn, line 503: What is the "size limit"?

docs/proxyCaptureProtocol.md
```diff
@@ -0,0 +1,1528 @@
...
+
+- Kafka capture submission remains permanently closed;
+- the process emits a loud, persistent capture-gap alarm; and
+- it does not claim orderly connection or publisher retirement for the compromised capture
+  activation.
+
+In an unmanaged deployment, the Critical Mutation Traffic request waiting at the pre-forward check
+and later traffic follow the configured one-way pass-through behavior immediately.
+
+In a controller-managed deployment, capture compromise and permission to begin pass-through are
```
jzonthemtn, line 992: Line 71 - "The standalone design does not require transactions or a controller."

This says a "controller-managed deployment."

How does a proxy know which mode it is in?

docs/replayerProcessingAndCommitArchitecture.md
```diff
@@ -0,0 +1,1581 @@
...
+intake before a record needed to complete, close, or expire retained state. The design therefore
+accepts that extreme traffic density or record size may exhaust memory rather than introducing
+such a deadlock by construction.
+
+Periodic proxy publication prevents a quiet or low-volume connection from retaining one nonempty
+record until connection close. It does not create a replayer memory ceiling: a record can still be
+large, Kafka can return large batches, and final source-response and tuple work may remain active
+after retry policy stops waiting.
+
+The deterministic boundary requires a later record. In normal live operation, periodic
```
jzonthemtn, line 1106: `captureAndReplayArchitecture.md` section 10 calls the retry boundary "finite, deterministic."

Here says a completely quiet partition "cannot establish a record-based retry boundary," which is correct, since `W` is a predicate over future records rather than a duration.

I think `captureAndReplayArchitecture.md` needs updating here.

docs/replayerProcessingAndCommitArchitecture.md
```diff
@@ -0,0 +1,1581 @@
...
+request-processing completion.
+
+Each component releases data and I/O objects that it created or explicitly retained for its own
+work. If the implementation passes reference-counted buffers across an owner boundary, the
+lower-level interface must state exactly which operation retains and releases each reference. This
+medium-level design does not use a general ownership-transfer abstraction.
+
+### 4.4 Long-running activity reporting
+
+The activity monitor reports long-running work using tracing identity that includes:
```
jzonthemtn, line 488: Do we need to define what "long-running" means? Is there a threshold after which work becomes "long-running"?

docs/captureAndReplay/captureAndReplayArchitecture.md
```diff
@@ -0,0 +1,1513 @@
...
+partition values that could disagree. The current archive import defined in §2.3 preserves the
+source partition layout.
+
+A Kafka topic must never contain records written under mutually wire-incompatible capture-protocol
+formats. Reusing an existing topic for a wire-incompatible format is explicitly prohibited, even
+after the old consumer group is empty or the deployment has entered a no-capture interval. A
+wire-incompatible format requires a new Kafka topic, and the complete capture and replay cohort
+using that format must use the new topic.
+
+### 2.3 Imported capture archives
```
gregschohn: I definitely DID leave this out, then some sweeps over repo consistency made me remember that this is a supported feature today.  Moreover, this is a feature that we've used to diagnose proxy and replayer issues with users in the past.

Given the complexity of what we're bringing in, I don't want to let our highest fidelity diagnostic feature atrophy.  We might not have full E2E tests with this in place, but I want to make sure that 1) our designs can support this feature and 2) we know how to get it fully up to speed.

We don't yet have tests for this scenario & we should, but I'll probably make that a follow-up

docs/captureAndReplayArchitecture.md
gregschohn: Thanks - there's no reason to keep this separate from the grouped record.  I didn't mean for this to slip in.

docs/captureAndReplay/proxyCaptureProtocol.md
```diff
@@ -0,0 +1,1528 @@
...
+force every ordinary packet and connection lifecycle change through one global proxy lock.
+
+Each connection's event-loop owner also controls the lifetime of its current `TrafficStream`:
+
+1. Appending the first observation to a new record schedules one monotonic deadline at `F`.
+2. Before beginning every later `TrafficObservation`, the owner compares the local monotonic time
+   with that fixed deadline. If the deadline has passed, it detaches the old record before admitting
+   the new observation. A delayed timer therefore cannot place new activity into an expired record.
+3. Appending later observations before the deadline does not move or replace that deadline.
+4. If the record reaches its size limit, a complete Critical Mutation Traffic request requires
```
gregschohn: How many bytes can go out in a single Kafka record, which is finite.  This is currently controlled with maxTrafficBufferSize and has a range of 1 byte to 1MB.  The kafka topics are configured to reject records > 1MB.  We could bump that to 10MB, but performance degrades.  Since OpenSearch can put >> 1 or 10MB in a single message, we need to do chunking anyway and we always have.

I was purposefully trying to avoid putting too many values into the document so that the documents are easier to maintain.  Let me know if you have a specific recommendation for a change here.

docs/captureAndReplay/proxyCaptureProtocol.md
```diff
@@ -0,0 +1,1528 @@
...
+
+- Kafka capture submission remains permanently closed;
+- the process emits a loud, persistent capture-gap alarm; and
+- it does not claim orderly connection or publisher retirement for the compromised capture
+  activation.
+
+In an unmanaged deployment, the Critical Mutation Traffic request waiting at the pre-forward check
+and later traffic follow the configured one-way pass-through behavior immediately.
+
+In a controller-managed deployment, capture compromise and permission to begin pass-through are
```
gregschohn: It will be started with different flags/inputs.  In a case with no controller, we won't be able to get a strong signal on some things, so the user has to figure that out on their own.  Once we have a stronger controller, we can get these to coordinate together more.  Each individual proxy though by itself won't be able to make global guarantees.

docs/captureAndReplay/replayerProcessingAndCommitArchitecture.md
```diff
@@ -0,0 +1,1581 @@
...
+request-processing completion.
+
+Each component releases data and I/O objects that it created or explicitly retained for its own
+work. If the implementation passes reference-counted buffers across an owner boundary, the
+lower-level interface must state exactly which operation retains and releases each reference. This
+medium-level design does not use a general ownership-transfer abstraction.
+
+### 4.4 Long-running activity reporting
+
+The activity monitor reports long-running work using tracing identity that includes:
```
gregschohn: This is an existing feature and just for diagnostics now and before.  There might be a threshold hard coded in the replayer - but generally, we sort each activity's largest spans by time and put them in descending order.  This shows up in `longRunningActivity.log`.

docs/captureAndReplay/replayerProcessingAndCommitArchitecture.md
```diff
@@ -0,0 +1,1581 @@
...
+intake before a record needed to complete, close, or expire retained state. The design therefore
+accepts that extreme traffic density or record size may exhaust memory rather than introducing
+such a deadlock by construction.
+
+Periodic proxy publication prevents a quiet or low-volume connection from retaining one nonempty
+record until connection close. It does not create a replayer memory ceiling: a record can still be
+large, Kafka can return large batches, and final source-response and tuple work may remain active
+after retry policy stops waiting.
+
+The deterministic boundary requires a later record. In normal live operation, periodic
```
gregschohn: I think that both of these are true.  Duration is going to be in observed source-time.  That means that when we look at the deterministic records and their timestamps, was there a response within N seconds of the end of the request that fully contained a response...

In normal circumstances, somebody should always be writing to a partition.  If the proxy is shut down cleanly and there are no replacements, we'll get close connections and either way, all responses will be accounted for before or after that (source) time limit.

The issue is if the records in the kafka go completely dark.  In that case, we'll sit and wait for SOME additional signal to come in.  Notice that this implies that there needs to be at least one clean shut down after the last http transaction has gone through - even if it's from a different process (or 3 processes later if the next one crashed too).

This paragraph in particular is talking about the 'finalized' archive which would have come in from an import (BYO).  In that case, there aren't proxies actively appending data, so this just said that we'll assume that for those, once we hit the end of that imported range, all messages being accumulated will be finalized.  That won't push the source time further into the future but it WILL terminate all messages being accumulated.  That, in turn will push the incomplete/missing messages downward through the handlers so that they can get along with their work, knowing that nothing more will be coming.

docs/proxyCaptureProtocol.md
gregschohn: It's a construct from the managed fleet document.  It's defined as a guard.
I'm replacing with just controller.  That was TMI for this level of document.  Thanks

gregschohn (commented):
Thanks for the comments Jeff.  I think that I've addressed them.  I've also made a couple more passes, including one for overall readability, which changes a lot.  The commits for these files show what has changed...

f5753f261ac62273623126d49536add935cab431
2755e476494cdd362f066d38273377f6d0a5da6f

The last major protocol change
- Replay intake requires the source-responses for each 'read-ahead' message that the it demands.  If the source-response isn't available quickly, determined by the max timestamp coming in for the partition of that http transaction, then the replay intake will be satisfied from a pull perspective and will consider that as a completed message for calculating buffered messages since upon retry, it won't need the response anyway.  The idea was to not great a massive buffering event upon reaching a retry and instead just be ready for those and smooth out processing.

The older commit:
- asyncMessagePassingProgrammingGuide: corrected a stale threading model.
The replay-intake input queue is drained by the dedicated replay-intake
thread, not the Kafka executor; the Kafka queue is drained by the Kafka
thread between bounded client calls; wakeup() is a stateless prompt.
- Resolved the empty-poll demand contradiction in favor of the lower-level
model: empty poll results are never sent to replay intake, so they are no
longer listed as demand-recompute triggers in the architecture doc.
- Standardized ConnectionRequestStarted on the first accepted target write
(connection LLD section 8) everywhere, so cancellation between the turn
start and the first write cannot make the two models disagree.
- Corrected the claim that a `range` archive cannot establish a retry
boundary or broker-time expiration: higher-offset records inside the range
do establish both; only the end of the range supplies no further evidence.
- Fixed symbol collisions: the unseen prior heartbeat in architecture 8.1 is
now U (H remains the heartbeat publication interval), and the managed
fleet broker-skew bound is now K (N remains replayer demand, P * T).
- Stated F's five-second default where F is introduced, and reworded the
undefined "connection interval was reset" phrasing in proxy 1.1.
- Clarified that a late heartbeat record finishes and commits even when it
does not move the writer-partition baseline.
- Pushed mechanism out of the top-level architecture doc in seven places
(record flush interval, acknowledgement deadline, scale-down, publisher
lane retirement, partition revocation, shutdown, proxy retirement
timeline), keeping the concept and the invariant at the top and linking to
the companion document that owns the rules normatively.
- Replaced the architecture 8.1 flowchart with a gantt chart drawn on one
partition's broker-time axis, so the restart fallback's E + 2S wait is
visible as a length rather than only described.
- Documented redelivery in architecture 9: it is Kafka's ordinary
at-least-once behavior rather than a replayer operation, and Retain is
reached only through cancellation, process failure, or a protocol
violation. Relabeled the diagram's Retain edge accordingly.
- Rewrote the 7.4 retry-wait passage and the 7.5 tuple-output versus
request-completion passage of replayerProcessingAndCommitArchitecture for
readability.
- Reframed managedFleetCaptureRecovery (about 26% shorter) around its actual
purpose: fleet management is not being implemented soon, and the document
exists to confirm the architecture can support a future controller, so it
deliberately carries less rigor than the rest.
- Added docs/captureAndReplayReviewFindings-2026-09-18.md recording the
review verdict, the three inconsistencies above, the wording issues, and
the non-issues that were checked and cleared.

gregschohn:
Closing this in favor of PR-3394

PR #3354, opened 2026-09-02
Fix EKS bootstrap image and storage handling

gregschohn:
## Summary
- fix EKS `aws-bootstrap.sh` so `--build --use-public-images` builds from source while using the public BuildKit image instead of requiring a mirrored BuildKit image in the stack ECR
- make BuildKit bootstrap retries portable across macOS, WSL, and Linux by using `gtimeout`/`timeout` when available and otherwise relying on the total bootstrap deadline
- keep retrying BuildKit bootstrap until the configured total timeout expires, with diagnostics emitted after failed attempts
- disable the shared logs PVC for EKS installs so the chart does not create `logs-pvc` in EKS deployments
- keep the EKS Auto Mode `auto-ebs-sc` StorageClass as the default block storage class without forcing Prometheus to explicitly use it
- exclude generated output directories from repo-root Spotless scans so Gradle 8.14 validation passes on macOS CI

## Verification
- Installed MA into account `876908012510` / `us-east-2` with stack `MA-DEMO`
- Verified Helm releases in namespace `ma` are deployed
- Verified `migration-console-0` and `prometheus-kube-prometheus-stack-prometheus-0` are Running
- Ran local `./gradlew build --no-daemon --stacktrace` on macOS
- Ran `./gradlew spotlessMisc spotlessYaml --no-daemon --stacktrace`
- Ran `git diff --check -- build.gradle`
- Rendered the `migrationAssistantWithArgo` chart with EKS values
- CI rerun for #3354: `all-ci-checks-pass` is green, `gradle-build-macos` is green, and Jenkins is green except for the persistent `eks-cdc-k6-load-test` failure

### DIFF
--- a/deployment/k8s/aws/aws-bootstrap.sh
+++ b/deployment/k8s/aws/aws-bootstrap.sh
@@ -1781,10 +1781,14 @@ if [[ "$build" == "true" && -z "$ma_images_source" ]]; then
     # EKS/bootstrap builds are always Kubernetes-hosted; they cannot assume
     # direct access to a local Docker daemon from the target environment.
     source "${base_dir}/buildImages/backends/eksKubernetesBuildkit.sh"
-    # Use the ECR-mirrored buildkit image so the kubernetes driver doesn't pull
-    # from Docker Hub. mirror_images_to_ecr already copied this image above.
-    ECR_HOST="${MIGRATIONS_ECR_REGISTRY%%/*}"
-    export BUILDKIT_IMAGE="${ECR_HOST}/mirrored/docker.io/moby/buildkit:buildx-stable-1"
+    if [[ "$push_images_to_ecr" == "true" ]]; then
+      # Use the ECR-mirrored buildkit image so the kubernetes driver doesn't pull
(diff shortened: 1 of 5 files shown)


deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -196,6 +198,8 @@ while [[ $# -gt 0 ]]; do
       echo "                                            They are not part of the standard image set. Only the k6"
       echo "                                            load-test cases (Test008x) need them. Requires --build."
       echo "  --image-tag <tag>                         Override the image tag (default: git short SHA)"
+      echo "  --skip-image-build                        With --build, skip container image build/mirror steps and"
```
jzonthemtn: At first I thought `With --build` was supposed to be `With --skip-image-build` but then I realized what it was saying.

Very super minor, but I would find something like this clearer:

`Skip container image build/mirror steps and use images already present in the stack ECR repository. Requires --build.`

jzonthemtn (approved):

PR #3355, opened 2026-09-03
Add authenticated k6 runner with direct Valkey access

gregschohn:
## Summary

- support unauthenticated, HTTP Basic, and SigV4-authenticated k6 traffic without runtime code downloads
- build a pinned load-test-only k6 runner with the OSI-licensed xk6-redis extension
- connect k6 directly to Valkey and remove the Webdis proxy deployment
- carry the combined runner through the Helm chart, test automation, and Jenkins EKS load-test pipeline while keeping it out of production image aggregates
- allow restricted build environments to provide a Go module mirror through the GOPROXY Docker build argument

## Verification

- uncached ARM64 image build with `GOPROXY=https://goproxy.io`
- uncached AMD64 image build with `GOPROXY=https://goproxy.io`
- live k6-to-Valkey `PING`, `SET`, and `GET` smoke test
- final image reports k6 v2.2.0 and xk6-redis v0.4.1
- 105 focused migration-console load-test/chart tests pass
- 39 focused test-automation tests pass
- Helm lint and tag/digest chart renders pass
- Gradle load-test aggregate dry run selects `buildKit_k6Runner`
- modified shell scripts pass `bash -n`

All commits include DCO signoff.


TrafficCapture/trafficLoadTest/lib/sigv4.js
```diff
@@ -0,0 +1,139 @@
+// SPDX-License-Identifier: Apache-2.0
```
jzonthemtn, line 1: For `sigv4.js`:

* I didn't look at this all the way down, but I do see k6 has a `k6-jslib-aws` has a `SignatureV4` class that might can be used here instead.
* If not, I'd recommend this class have some tests.

TrafficCapture/trafficLoadTest/lib/valkey.js
```diff
@@ -0,0 +1,18 @@
...
+ * Execute one RESP command against Valkey.
+ *
+ * The registry and control bus are optional test aids. Preserve their existing graceful-degradation
+ * behavior when the store is unavailable instead of failing the traffic iteration.
+ */
+export async function valkeyCommand(command, ...args) {
+  try {
+    return await client.sendCommand(command, ...args);
+  } catch (_) {
+    return null;
```
jzonthemtn, line 16: Should it return `null` for every exception?

TrafficCapture/trafficLoadTest/Dockerfile
```diff
@@ -1,16 +1,33 @@
...
+# sum.golang.org still verifies their checksums.
+ARG GOPROXY
+
+RUN xk6 build v2.2.0 \
+      --with github.com/grafana/xk6-redis@v0.4.1 \
+      --output /tmp/k6
+
+FROM grafana/k6:2.2.0
+
+COPY --from=builder /tmp/k6 /usr/bin/k6
```
jzonthemtn, line 23: So this previously just pulled the k6 image from DockerHub. Now, it is building a new `k6` binary and copying it into the image overwriting the existing one, modifying the originally pulled image. k6 is AGPLv3, so I just wanted to make sure we're ok having that modified k6 image from a licensing perspective. I think so as long as this image is not distributed, but I'm not an expert on how the GPL/AGPL interacts with Apache license in cases like this.

TrafficCapture/trafficLoadTest/lib/sigv4.js
gregschohn: Good callout. It was painful to load a library into the build.  Limiting that to one (for the valkey lib) seems much safer than needing to depend upon two.

I've added tests.  Thanks

TrafficCapture/trafficLoadTest/lib/valkey.js
gregschohn: Thanks - I let autocoding miss that one.  I think that this was preserving the old behavior, but at least we now log too

jzonthemtn (approved):
looks good

PR #3356, opened 2026-09-04
Don't reverse-derive source fields from copy_to fan-in targets

ankitsinghdungriyal:
### Description

Sourceless reconstruction's last-resort pass (`SourceReconstructor`, pass 5) recovers a source field's value by probing one of its `copy_to` targets. When that target is a **fan-in target** — more than one source field copies into it — it holds the union of every contributor's values, with no record of who contributed what. Attributing that union back to a single source field invents data.

This is most damaging when the source field has a Lucene footprint of its own (`indexed` or `doc_values`): tiers 1–4 already failed for it, which means the document genuinely had no value there, and the union then lands in `_source` as exact, term-queryable content.

Concretely, on a mapping where several sibling fields share one search target, a document that populated only some of those siblings was reconstructed with the rest filled in from the union:

```
term bcc=<value only ever present in from>   source cluster: 0 hits   target cluster: 1 hit
```

Every sibling the document left empty inherits the values of the siblings it did populate. That is a silent correctness problem, not just noise — the fabricated values are exact and term-queryable, so they satisfy filters the original document would not have matched.

This change skips fan-in targets during reverse-derivation for source fields that have their own footprint. Source fields with **no** footprint keep the existing best-effort (lossy) recovery, because their absence from tiers 1–4 says nothing about whether the document had a value — that is the behaviour `NoStoredSourceMigrationTest.testCopyToWithMixedSourcePosture` covers, and it still passes.

### Issues Resolved

None filed — found while validating sourceless migration against indices whose mappings share `copy_to` targets across sibling fields. Happy to open an issue to track if preferred.

### Testing

Two new cases in `SourceReconstructorTest`:

- `reverseDerive_skipsFanInTarget_doesNotFabricateAbsentSiblings` — four indexed keyword fields fanning into one target, document populated for only two; asserts the absent two stay absent and the target is still stripped from `_source`. **Fails without this change**, printing the fabricated union:

  ```json
  {"from":["alice@example.com"],"to":["bob@example.com"],
   "cc":["alice@example.com","bob@example.com"],
   "bcc":["alice@example.com","bob@example.com"]}
  ```

- `reverseDerive_stillUsesFanInTarget_whenSourceHasNoLuceneFootprint` — guards against over-tightening: two `_source`-excluded, footprint-less text fields sharing a target must still be recovered through it.

Verified green (fix applied):

- `SourceReconstructorTest` — full class, including the 12 pre-existing `copy_to` / reverse-derive cases
- `NoStoredSourceMigrationTest.testCopyToWithMixedSourcePosture` — ES 7.10.2 → OS 3.5.0
- `NoStoredSourceMigrationTest.testCopyToFieldHandling` — ES 1.7.6 / 2.4.6 → OS 3.5.0
- `NoStoredSourceMigrationTest.testNormsFalseSourceDisabledRecovery` — ES 7.10.2 → OS 3.5.0
- `SourcelessMigrationTest.testSourcelessMigration_ES7_objectArrayWithTextFields`
- `./gradlew :SearchSnapshotExtractor:spotlessCheck`

End-to-end, the index that surfaced this migrated with `_source` matching the origin document exactly and per-field term queries returning identical hit counts on both clusters.

### Check List
- [x] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable. — n/a, no user-facing behaviour or API change

### DIFF
--- a/SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
+++ b/SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
@@ -608,8 +608,7 @@ public class SourceReconstructor {
                     // when the source has its own Lucene footprint (tiers 1-4 already failed, so the
                     // doc truly had no value); footprint-less sources keep best-effort recovery.
                     if (mappingContext.getCopyToSources(targetField).size() > 1
-                            && sourceMapping != null
-                            && (sourceMapping.indexed() || sourceMapping.docValues())) {
+                            && hasIndexedOrDocValuesFootprint(sourceMapping)) {
                         continue;
                     }
                     ProbeResult recovered = probeFieldValue(reader, docId, document, targetField,
@@ -617,17 +616,7 @@ public class SourceReconstructor {
                     if (recovered == null) {
                         continue;
                     }
-                    // Tiers 1/2/4 already shaped the value through the TARGET's mapping —
-                    // write it verbatim. Only the points/terms tier (Raw) still needs decoding,
-                    // and that decode runs through the SOURCE mapping so the JSON shape matches
-                    // the source field's declared type (date formatting, scaled_float division,
-                    // IP rendering, etc.).
-                    Object converted = switch (recovered) {
-                        case ProbeResult.Final f -> f.value();
-                        case ProbeResult.Raw r -> sourceMapping != null
-                                ? convertFallbackValue(r.raw(), sourceMapping)
-                                : null;
-                    };
+                    Object converted = convertProbeResult(recovered, sourceMapping);
                     if (converted != null) {
                         modified |= putNested(target, sourceField, converted);
                         break; // first-hit wins; stop iterating targets.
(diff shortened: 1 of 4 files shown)


gregschohn:
Thanks for this commit!  Can you please amend the pushed commit so that it includes a [DCO signoff](https://github.com/opensearch-project/opensearch-migrations/pull/3356/checks?check_run_id=101003818254)
```
Signed-off-by: Author Name <authoremail@example.com>
```

ankitsinghdungriyal:
> Thanks for this commit! Can you please amend the pushed commit so that it includes a [DCO signoff](https://github.com/opensearch-project/opensearch-migrations/pull/3356/checks?check_run_id=101003818254)
> 
> ```
> Signed-off-by: Author Name <authoremail@example.com>
> ```
Hey @gregschohn, I'd already included it there 
`Signed-off-by: ankitsinghdungriyal <ankit.dungriyal@smarsh.com>`

jzonthemtn:
> > Thanks for this commit! Can you please amend the pushed commit so that it includes a [DCO signoff](https://github.com/opensearch-project/opensearch-migrations/pull/3356/checks?check_run_id=101003818254)
> > ```
> > Signed-off-by: Author Name <authoremail@example.com>
> > ```
> 
> Hey @gregschohn, I'd already included it there `Signed-off-by: ankitsinghdungriyal <ankit.dungriyal@smarsh.com>`

@ankitsinghdungriyal Check your commit signing with gpg. Looks like that's what's missing.

SearchSnapshotExtractor/src/main/java/org/opensearch/migrations/bulkload/lucene/SourceReconstructor.java
```diff
@@ -603,6 +603,21 @@ private static boolean populateFromSegment(Map<String, Object> target, LuceneLea
                     continue;
                 }
                 for (String targetField : rankedTargets) {
+                    // A fan-in target (more than one source field copies into it) holds the UNION
```
jzonthemtn: A minor comment - this comment seems long - longer than the fix. :)

Possible to shorten/condense the comment?

AndreKurait:
Added an additive follow-up for a fan-in edge case: when all `copy_to` source fields have `index:false` and `doc_values:false` with `_source` disabled, the shared target is the only recoverable evidence. The fallback now emits that target directly (for example, `{"shared":"red"}`), preserving target-field search behavior without fabricating `alpha` or `beta`. Covered by an ES 7.10 → OpenSearch E2E test; both fan-in E2E cases and Spotless pass.

AndreKurait (approved):

PR #3372, opened 2026-09-12
Add configurable resources for metadata migration pods

### LINKED ISSUE #3347: Metadata Migration EKS Resource Configurations
### What is the bug?

There's no way to bump up the memory requirements for Metadata Migration.

_What versions of Elasticsearch / Opensearch._<br/>
_What version of the Migration Assistant (this can be determined by running `console --version` on the Migration Console)._

_Steps to reproduce the behavior._

_A clear and concise description of what you expected to happen._

_Add any other context about the problem, such as log files our console output._

Jeremydupras:
## Summary

Fixes #3347 by allowing metadata migration pod resources to be configured through workflow config.

This change:
- adds `metadataMigrationConfig.resources` to the user schema
- carries metadata resource settings through config transformation
- exposes metadata resources in the Argo-facing schema/config
- applies metadata resource settings to the metadata migration pod spec
- updates schema/template/config snapshots and validation coverage

## Details

Previously, metadata migration always used the built-in Java migration console resource defaults, so there was no way to increase memory requirements for metadata migration.

With this change, users can now set:

```json
"metadataMigrationConfig": {
  "resources": {
    "requests": {
      "cpu": "1500m",
      "memory": "3Gi"
    },
    "limits": {
      "cpu": "2000m",
      "memory": "4Gi"
    }
  }
}
```

Partial overrides are deep-merged with the existing metadata migration defaults.

## Testing

- updated schema snapshots
- updated workflow template snapshots
- added config transformer validation coverage for metadata resource overrides
- verified locally that submitted workflow config renders metadata pod resource limits from `metadataMigrationConfig.resources`

## Notes

This PR is scoped to metadata migration resource configurability only.
Local quickstart/test harness changes used during validation are not included.


jzonthemtn:
Hi @Jeremydupras, please check your git commit signing and sign-off for the DCO check.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
```diff
@@ -503,6 +503,19 @@ export const RESOURCE_REQUIREMENTS = z.object({
 
 export type ResourceRequirementsType = z.infer<typeof RESOURCE_REQUIREMENTS>;
 
+const CPU_MEMORY_RESOURCE_REQUIREMENTS = z.object({
```
jzonthemtn, line 506: This is `RESOURCE_REQUIREMENTS` minus ephemeral-storage, which is already .optional() in `CONTAINER_RESOURCES`, so the existing schema should cover metadata as-is. Was there a reason to exlucde ephemeral storage?

jzonthemtn:
@Jeremydupras Thanks for squaring away the DCO. Left one comment, otherwise looks good.

orchestrationSpecs/packages/schemas/src/userSchemas.ts
Jeremydupras: It was omitted to keep metadata migration narrower and simpler following the already established schema. Also the repo does not currently have a principled way to size or validate metadata scratch disk requirements. Using the already established RESOURCE_REQUIREMENTS could also be done which would allow the user to specify if they want ephemeral storage and exclude it if not. I can make the change if you'd like.

jzonthemtn: No, I don't think it's necessary. Just curious when I saw it.

jzonthemtn (approved):

Issue #3378, opened 2026-09-14
[RFC] Azure support for the OpenSearch Migration Assistant


tonypiazza:
## Summary

This proposes adding Azure support to the Migration Assistant: Azure Blob Storage (ABS) as a snapshot repository backend, AKS as an orchestration platform, and a Terraform module to provision both. It completes the third of the three major providers set out in #2845.

The work is divided into five PRs, sequenced so each is independently reviewable, each stays within one layer of the stack, and each is a no-op for existing AWS and GCP deployments.

I implemented the GCP and AWS paths referenced below and will implement and maintain the Azure path as well, so please assign this to me.

## Motivation

#2845 proposed making the Migration Assistant cloud-agnostic so that customers running OpenSearch outside AWS could use the same migration tooling. GCP support landed across [#3075], [#3205], [#3220], [#3255], and [#3282]; AWS Terraform support landed in [#3341]. Azure is the remaining provider named in that RFC, and it is the one where the groundwork is most complete.

## What Azure inherits

Most of #3075's review cycles went into replacing provider-specific parallel paths with generic ones. Azure plugs into those generic paths rather than adding new branches beside them:

| Existing generalization | What Azure needs now |
| --- | --- |
| [`RepoUri`][RepoUri] sealed interface with scheme-based dispatch | One record, one `parse` branch |
| Unified [`SnapshotCreator`][SnapshotCreator] (file/S3/GCS in one switch) | One `case` arm emitting `type: azure` |
| Provider-neutral CLI (`--repo-uri`, `--local-dir`, `--endpoint`) | No new flags |
| Unified workflow templates (no `*Gcs.ts` variants remain) | No new templates |
| Generic `REPO_ARTIFACTS_BUCKET`, shared `migrations-default-s3-config` | No new ConfigMap |
| [`SnapshotReaderGcs`][GcsRepo] as a separate Gradle module | Same pattern for `SnapshotReaderAzure` |
| `cloudProvider` as a resolved template input ([#3220]) | Already emits the Azure internal-LB annotation |
| Per-leg `modules/connectivity/` pattern, defaulting to public ([#3205]) | Same shape, Azure primitives |
| Terraform module layout and `helm.tf` destroy ordering ([#3341]) | Same structure, `azurerm` resources |

[...]

PR #3389, opened 2026-09-16
Make RepoUri validation exhaustive ahead of Azure support

tonypiazza:
### Description

First PR of the Azure support sequence in #3378. No Azure code, no new provider, no behavior change for AWS or GCP. This closes the generalization gaps that would otherwise let a new `RepoUri` variant pass through repo validation unnoticed.

**The problem.** `RepoUri` is a sealed interface, but seven validation sites test it with `instanceof` chains or scheme-prefix `startsWith` checks that have no exhaustiveness guarantee. A new variant added to the sealed hierarchy compiles fine and silently falls through every one of them, so its required arguments never get validated. The failure surfaces much later as a confusing runtime error from the storage client rather than as a `ParameterException` at startup.

**The fix.** Convert those seven sites to pattern-matching `switch` statements over the sealed type with no `default ->` arm. Adding a variant to `RepoUri` now fails to compile at each site that has to make a per-provider decision. PR 2 uses that list of compile errors as its work list, which is the reason this lands first and separately.

Sites converted:

| File | Sites |
| --- | --- |
| `MetadataMigration/.../cli/ClusterReaderExtractor.java` | 2 |
| `DocumentsFromSnapshotMigration/.../RfsMigrateDocuments.java` | 3 |
| `CreateSnapshot/.../CreateSnapshot.java` | 1 |
| `SolrReader/.../solr/SolrSnapshotCreator.java` | 1 |

Two details worth flagging for review:

- `ClusterReaderExtractor` needs an explicit `case null` arm. Its selector is nullable (`arguments.repoUri != null ? RepoUri.parse(...) : null`) because a null repo URI is legitimate on the remote-source path, and a `switch` over a null selector throws NPE where `instanceof` tolerated it silently. The other two sites reject null upfront (`RfsMigrateDocuments` throws when `args.repoUri == null`; `CreateSnapshot` declares `--repo-uri` as `required = true`), so neither needs the arm.
- Every exception message is byte-identical to before, including the inconsistent trailing period between `MetadataMigration` (no period) and `DocumentsFromSnapshotMigration` (period). The existing assertions pin those strings, and I did not normalize them here so this PR stays a pure refactor.

**Solr sites, added in review.** The first version of this PR scoped Solr out on the grounds that Solr rejects cloud providers other than S3 anyway. Two of those sites were worth doing, and one of them was hiding a real bug:

- `RfsMigrateDocuments.validateArgs` did not reject `gs://` for Solr sources at all. It parsed the URI, skipped the S3-only branch, and returned clean; the rejection happened later in `buildSolrSourceFactory` after the migration had already started. The exhaustive switch turns that into a startup `ParameterException`. The added test fails against the previous code, so this is a behavior fix rather than a refactor.
- `SolrSnapshotCreator.isCloudRepoUri` matched scheme prefixes with `startsWith` rather than going through `RepoUri`, which made it invisible to the sealed-type mechanism: a new variant would silently answer "not cloud" and get treated as a filesystem path. It now switches over `RepoUri`. It stays total on purpose, since callers pass locations Solr handles as plain paths (`SolrBackupStrategy` passes a bare path for file repos, and the legacy standalone replication handler accepts relative paths), so a value `RepoUri.parse` rejects still answers "not cloud" instead of throwing. There is a test pinning that.

`SolrBackupStrategy` itself is still untouched: its 11 `instanceof` uses sit downstream of `validateSolrRepoSupported`, which already rejects `gs://` up front, and PR 2 extends that single guard rather than adding provider arms to paths whose job is to reject the provider. Lifting the Solr provider limitation remains separate work, independent of this RFC.

**Verified the mechanism actually bites.** I temporarily added a fourth variant to the sealed `RepoUri` and compiled. It produced exhaustiveness errors at all 7 relevant sites, including both Solr ones above, and none of them are reachable without a compile failure. That probe was reverted; `RepoUri.java` is unchanged by this PR.

**Comment corrections.** `migrationConfigTransformer.ts` claimed GCS has no LocalStack equivalent. `fake-gcs-server` is exactly that equivalent; the real distinction is that the `localstack://` hostname-to-IP rewrite is a workaround for S3 *virtual-host-style addressing*, and object stores whose clients address buckets through the URL path need no equivalent. That mattered to get right before a third provider arrives, because the inaccurate comment invites reading the rewrite as a generic emulator feature that Azure would also need. `MetadataMigration/DEVELOPER_GUIDE.md` documented the `127.0.0.1` symptom without naming that cause, so it now says why.

`useLocalStack` keeps its name despite being S3-specific. It is projected as `repoUseLocalStack` and `snapshotRepoUseLocalStack` with `changeRestriction: "impossible"` in `migrationResourceProjections.ts`, so renaming it would force ValidatingAdmissionPolicy regeneration plus churn across `argoSchemaSnapshot` and the workflow `.snap.json` files, for no behavioral gain. Comments only here.

### Issues Resolved

Part of #3378.

### Testing

Six characterization tests added first, written to pass against the pre-refactor code so a red run afterward would unambiguously mean the refactor broke something. They cover the GCS validation rules that had no coverage at all before this PR: 2 in `ClusterReaderExtractorTest`, 2 in `RfsMigrateDocumentsHelpersTest`, 1 in `TestCreateSnapshotArgs`, 1 in `SolrSnapshotCreatorTest`. The `CreateSnapshot` one pins the asymmetry that snapshot *creation* writes to the repo and so needs no `--local-dir` for `gs://`, unlike the read paths. The `SolrSnapshotCreator` one pins the totality requirement described above.

One test is deliberately not a characterization test: `validateArgs_solr_rejectsGcsRepoUpFront` fails against the previous code, which is the evidence that the `gs://` gap was a real bug rather than a style issue.

Verified on this branch rebased onto main:

- `./gradlew clean build` green from a cold cache (`clean` deliberately, since a warm cache has masked a compile error in this repo before).
- Full `orchestrationSpecs` suite green: 437 tests across 4 packages, 18 snapshots all passing with none written or obsolete.
- The added tests confirmed executing by name under `--rerun-tasks` rather than being restored from cache.
- `grep -rn "instanceof RepoUri" --include="*.java" .` returns 11, down from 17, all remaining ones in `SolrBackupStrategy` behind `validateSolrRepoSupported`.
- `grep -rn "default ->"` across the four affected modules returns exactly 3, all pre-existing Solr sites with unchanged content. A fourth would silently defeat the entire point of this PR, so it is worth a reviewer's glance.
- Zero diff in any `.yaml`, `.json`, or `.snap.json`. No chart template, values file, CRD, or snapshot is touched, which is the evidence that AWS and GCP behavior is unchanged.

### Check List
- [ ] Public documentation issue/PR [created](https://github.com/opensearch-project/documentation-website/issues/new/choose), if applicable. N/A: the only docs touched are an internal developer guide and code comments.


DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
AndreKurait: let's fix this one too

AndreKurait (commented):
I'm aligned with stricter compile time assertions. A few more files need a change:

https://github.com/opensearch-project/opensearch-migrations/blob/86716d8e3207ff07bd1f293c860ebf93f6cb0ce5/SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrSnapshotCreator.java#L118-L120

DocumentsFromSnapshotMigration/src/main/java/org/opensearch/migrations/RfsMigrateDocuments.java
tonypiazza: Done in 9ccf993ba.

Worth flagging that this one was hiding a real bug, not just a missing compile-time assertion. `validateArgs` parsed the URI, matched neither branch for `gs://`, and returned clean. The rejection came later from `buildSolrSourceFactory`, after the migration had already started. It is now a startup `ParameterException`, and the added test fails against the previous code.

tonypiazza:
@AndreKurait Thanks, both are in.

`SolrSnapshotCreator.isCloudRepoUri` is d00c85bc5. That one was invisible to the sealed type entirely, since it matched scheme prefixes with `startsWith` rather than going through `RepoUri`: a new variant would have silently answered "not cloud" and had its location treated as a filesystem path. It now switches over `RepoUri`. I kept it total on purpose, because callers pass locations Solr handles as plain paths (`SolrBackupStrategy` passes a bare path for file repos, and the legacy standalone replication handler accepts relative paths), so a value `RepoUri.parse` rejects still answers "not cloud" instead of throwing. There is a test pinning that.

`RfsMigrateDocuments.validateArgs` is 9ccf993ba, with details in the inline reply. Short version: `gs://` was not rejected there at all, so that commit is a behavior fix rather than a refactor.

To confirm the mechanism actually bites, I temporarily added a fourth variant to the sealed `RepoUri` and compiled. It produced exhaustiveness errors at all 7 sites, including both of these. That probe is reverted and `RepoUri.java` is untouched by this PR.

On "a few more files": I grepped for `instanceof RepoUri` and for `startsWith("s3://")` / `startsWith("gs://")`, and these two were the only sites outside `SolrBackupStrategy`. If you had a third in mind, point me at it.

`SolrBackupStrategy`'s remaining 11 `instanceof` uses stay as they are. They all sit downstream of `validateSolrRepoSupported`, which already rejects `gs://` up front, so PR 2 extends that single guard instead of adding provider arms to paths whose job is to reject the provider.

SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrSnapshotCreator.java
```diff
@@ -114,9 +115,32 @@ public void createSnapshot() {
...
     static boolean isCloudRepoUri(String location) {
-        return location != null && (location.startsWith("s3://") || location.startsWith("gs://"));
+        if (location == null) {
+            return false;
+        }
+        RepoUri parsed;
+        try {
+            parsed = RepoUri.parse(location);
+        } catch (IllegalArgumentException e) {
+            return false;
```
AndreKurait, line 137: hmm, my hunch is that this should throw, i can be content with this either way though since this keeps existing behavior

AndreKurait (approved):

SolrReader/src/main/java/org/opensearch/migrations/bulkload/solr/SolrSnapshotCreator.java
tonypiazza: There is a concrete case that argues against throwing.

`SolrBackupStrategy` line 180 converts a `FileRepoUri` to its bare `path()` before passing it down as `backupLocation`, since Solr's backup APIs want a scheme-less path. For `--repo-uri file://relative/path` that yields `relative/path`, which `RepoUri.parse` rejects, so this catch is reached with a location Solr handles fine. Throwing would turn a working file-repo backup into a crash. I ran `parse` on the values line 180 actually produces: `file://relative/path` and `file://` both land in the catch, `file:///abs/path` parses clean.

Correcting my earlier comment: I also cited the legacy standalone replication handler, and that reason does not hold. `runStandaloneBackup` constructs `SolrStandaloneBackupCreator`, which never calls `isCloudRepoUri`. The line-180 case is the real one. Since this is approved and green, the comment correction rides in PR 2 rather than re-running the matrix here.

If you want the strictness anyway, the better fix is a `parse` that returns "not a repo URI" instead of throwing, so this site can classify without a try/catch. Bigger than this PR.

PR #3391, opened 2026-09-16
Apply metadata resource defaults during schema parsing

gregschohn:
## Summary

- resolve metadata migration resources to the Java migration console defaults when the block is omitted
- deep-merge partial metadata resource overrides during user-schema parsing
- align metadata resource behavior with proxy, replayer, and document backfill resources
- carry the required resource object through the transformer without downstream fallback defaults
- apply metadata pod resources through a strict Argo expression without emitting a duplicate static resource block
- preserve resource enforcement when resources come from either the direct container or `podSpecPatch.mainContainer`
- require `podSpecPatch` callers to use typed overlays so opaque strings cannot invalidate resource guarantees
- preserve volume, artifact, and pod-configuration type state across `addEnvVars`, resetting only the environment scope
- model direct environment values and ConfigMap/Secret `valueFrom` sources as an exclusive typed union
- add schema and workflow-builder coverage for default resolution, pod-spec-provided resources, builder-scope preservation, and environment-value rendering
- type raw config-processor fixtures as schema input so defaulted resources remain optional before parsing and required afterward

The user schema is now the single owner of metadata resource defaults. `ARGO_METADATA_OPTIONS` requires the completed resource object, and downstream workflow generation treats that contract as authoritative. Metadata uses `podSpecPatch` because it runs as a direct Argo container, but the patch now performs a strict resource lookup and supplies the only resource block emitted for that container.

Opaque `podSpecPatch` strings are no longer accepted by the safe builder API. A string can overwrite the main container after the builder has checked its resource state, so accepting one would make the compile-time guarantee unsound even when `addResources` was called directly. Callers now provide `PodSpecPatchOverlay` values, which the builder serializes internally; typed `terminationGracePeriodSeconds` support preserves the existing pod-level patch use case.

The builder tests directly extract the `ContainerBuilder` generic parameters before and after `addEnvVars`. They assert that container, volume, output, pod-configuration, and artifact scopes remain identical; the callback begins with an empty environment scope; and the output environment scope contains exactly the newly added variables with their expected value types. Negative checks use `@ts-expect-error` rather than `@ts-ignore` so they fail if the expected compiler errors disappear.

The environment renderer previously declared every entry as `BaseExpression`, even though `addEnvVar` legitimately accepts literal strings. That false object-only type allowed an unsafe `in` check and caused literal values to throw at runtime. Environment entries are now an exclusive union of direct literal/expression values and typed `configMapKeyRef` or `secretKeyRef` sources. The renderer narrows that union with a type predicate, while the builder rejects missing type tokens, ambiguous dual-source objects, and parameter-style `from` objects before rendering.

Config-processor fixtures that represent raw user configuration now use `z.input<typeof OVERALL_MIGRATION_CONFIG>`. This preserves the intended schema boundary: callers may omit defaulted resources, while parsed output continues to require the completed resource object.

The generated metadata WorkflowTemplate is 392 bytes smaller.

## Testing

- `npm test -w @opensearch-migrations/schemas -- --runInBand` (105 tests on the main-based branch)
- `npm run type-check -w @opensearch-migrations/schemas`
- `mise exec node@22 -- npm test -w @opensearch-migrations/config-processor -- --runInBand` (124 tests, 1 snapshot)
- `npm test -w @opensearch-migrations/argo-workflow-builders -- --runInBand` (153 passed, 2 skipped)
- `npm test -w @opensearch-migrations/migration-workflow-templates -- --runInBand` (68 tests, 14 snapshots)
- `mise exec node@22 -- npm run test` (453 passed, 2 skipped; 18 snapshots)
- `mise exec node@22 -- npm run type-check`


jzonthemtn (approved):
Looks good. Only concern is scope but those other changes are small.

PR #3393, opened 2026-09-17
Add standalone EKS access grant-only bootstrap mode

### LINKED ISSUE #3365: [BUG] aws-bootstrap.sh seems to run all the way even if the need is add IAM after initial bootstrap run
### What is the bug?
After initial bootstrapping if we want to only add a new IAM role to enable access to the EKS we need to run the bootstrap again, but the bootstrap does not seem to have an option to just add the IAM role without running the full bootstrap.

_What versions of Elasticsearch / Opensearch._<br/>
_What version of the Migration Assistant (this can be determined by running `console --version` on the Migration Console)._

_Steps to reproduce the behavior._

_A clear and concise description of what you expected to happen._

_Add any other context about the problem, such as log files our console output._

Jeremydupras:
### Description

This change adds a standalone `--grant-eks-access-only` mode to `deployment/k8s/aws/aws-bootstrap.sh` so operators can grant EKS cluster-admin access to an IAM principal after bootstrap without re-running image mirroring, Helm installation, or kubeconfig setup.

It also adds regression coverage for both the source bootstrap script and the assembled release artifact.

Key changes:
- add `--grant-eks-access-only` argument validation and help text
- skip release version resolution in grant-only mode
- skip non-AWS tool prerequisites (`jq`, `kubectl`, `helm`) in grant-only mode
- perform the EKS access-entry/policy association before kubeconfig and Helm work
- exit immediately after applying the requested access entry
- add regression coverage for:
  - missing principal guard
  - invalid flag combination guard
  - successful grant-only path without artifact/tool dependencies

### How To Run

Following the EKS deployment guide, download the bootstrap script and deploy into a new VPC or existing VPC first. After the cluster is up, run the new access-only mode to grant a second admin principal without re-running bootstrap.

Set the environment:

```bash
export STAGE=<stage>
export REGION=<region>
export STACK_NAME=MA-${STAGE}
export CLUSTER_NAME=migration-eks-cluster-${STAGE}-${REGION}
export BOOTSTRAP_PRINCIPAL_ARN=arn:aws:iam::<bootstrap-account-id>:role/<bootstrap-admin-role>
export NEW_PRINCIPAL_ARN=arn:aws:iam::<new-account-id>:role/<new-admin-role>
```

Bootstrap the cluster:

```bash
./aws-bootstrap.sh \
  --deploy-create-vpc-cfn \
  --stack-name "$STACK_NAME" \
  --stage "$STAGE" \
  --region "$REGION" \
  --eks-access-principal-arn "$BOOTSTRAP_PRINCIPAL_ARN"
```

Grant the second admin without re-running bootstrap:

```bash
./aws-bootstrap.sh \
  --grant-eks-access-only \
  --stage "$STAGE" \
  --region "$REGION" \
  --eks-access-principal-arn "$NEW_PRINCIPAL_ARN"
```

Verify the deployment and console access:

```bash
aws eks update-kubeconfig --region "$REGION" --name "$CLUSTER_NAME"
kubectl get pods -n ma
kubectl exec -it migration-console-0 -n ma -- /bin/bash
```

Verify the new access entry from the cluster-owning account:

```bash
aws eks list-access-entries \
  --cluster-name "$CLUSTER_NAME" \
  --region "$REGION"

aws eks list-associated-access-policies \
  --cluster-name "$CLUSTER_NAME" \
  --principal-arn "$NEW_PRINCIPAL_ARN" \
  --region "$REGION"
```

### Issues Resolved

Resolves https://github.com/opensearch-project/opensearch-migrations/issues/3365

### Testing

- `bash -n deployment/k8s/aws/aws-bootstrap.sh`
- Manual validation on EKS:
  - bootstrapped cluster successfully
  - ran `aws-bootstrap.sh --grant-eks-access-only` to add a second admin principal
  - verified both the bootstrap account and the newly granted account could access the cluster and `exec` into `migration-console-0`

### Check List
- [ ] Public documentation issue/PR created, if applicable


Jeremydupras:
This PR does not allow multiple IAM roles during bootstrap and each additional role would need to be run individually after bootstrap with the --grant-eks-access-only flag. A future improvement could be allowing multiple IAM roles during initial bootstrap and afterwards.

deployment/k8s/aws/aws-bootstrap.sh
```diff
@@ -574,8 +600,12 @@ if [[ "$create_vpc_endpoints" == "all" ]]; then
 fi
 
 # --- resolve version once ---
+# Grant-only mode never touches release artifacts, so avoid an unnecessary GitHub lookup.
+if [[ "$grant_eks_access_only" == "true" ]]; then
+  RELEASE_VERSION="grant-access-only"
```
gregschohn: nit: VERSION is already used in deployments - let's pick a name with more disambiguation

gregschohn (approved):
One minor nit, but other than that LGTM

deployment/k8s/aws/aws-bootstrap.sh
Jeremydupras: RELEASE_VERSION changed to BOOTSTRAP_ARTIFACT_VERSION

gregschohn (approved):
I would like to see BOOTSTRAP_ARTIFACT_VERSION split into two variables rather than overloading it for a couple purposes

